PR #72, opened 2017-06-21
Return filledAmount from fillOrderAsync

LogvinovLeon:
This PR:
* Returns filledAmunt from fillOrderAsync and updates tests


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -144,10 +145,11 @@ export class ExchangeWrapper extends ContractWrapper {
      *                                  execution the tokens cannot be transferred.
      * @param   takerAddress            The user Ethereum address who would like to fill this order.
      *                                  Must be available via the supplied Web3.Provider passed to 0x.js.
+     * @return                          The amount of the order (in taker tokens baseUnits) that was filled.
```
fabioberger: `The amount of the order that was filled (in taker token baseUnits).`

test/exchange_wrapper_test.ts
```diff
@@ -339,6 +339,24 @@ describe('ExchangeWrapper', () => {
                     expect(await zeroEx.token.getBalanceAsync(takerTokenAddress, takerAddress))
                         .to.be.bignumber.equal(fillableAmount.minus(partialFillAmount));
                 });
+                it('should return filled amount', async () => {
```
fabioberger: This test is more complex then this description. A simple `should return filled amount` test would be:

```
const signedOrder = await fillScenarios.createFillableSignedOrderAsync(
                        makerTokenAddress, takerTokenAddress, makerAddress, takerAddress, fillableAmount,
                    );
const partialFillAmount = new BigNumber(3);
const filledAmount = await zeroEx.exchange.fillOrderAsync(
    signedOrder, partialFillAmount, shouldCheckTransfer, takerAddress);
expect(filledAmount).to.be.bignumber.equal(partialFillAmount);
```

I would add that test and then rename this one to: `should return the partially filled amount if the fill amount specified is greater then the amount available`

fabioberger:
Agreed with @abandeali1. Actually all the `batch` methods should return the amount filled too.

abandeali1:
It's harder for the batch methods to return a single amount because it's not mandatory to use orders with the same `takerToken` (this is intentional and is also the reason they don't return a value on-chain). It could be nice to return a map of tokenAddress => filledAmount, but I don't think it's completely necessary.

fabioberger:
Let's work on returning something useful for batch methods in a subsequent PR.

PR #73, opened 2017-06-21
Auto-upload json docs on release

LogvinovLeon:
This PR:
* Auto-generates docs json upon release and uploads it to an S3 bucket


package.json
```diff
@@ -15,7 +15,9 @@
     "prebuild": "npm run clean",
     "build": "run-p build:*:prod",
     "prepublish": "run-p build:umd:prod build:commonjs:dev",
-    "postpublish": "publish-release --assets _bundles/index.js,_bundles/index.min.js",
+    "postpublish": "run-s release docs:json upload_docs_json",
+    "release": "publish-release --assets _bundles/index.js,_bundles/index.min.js",
+    "upload_docs_json": "aws s3 cp docs/index.json s3://0xjs-docs-jsons/$(git describe --tags).json",
```
fabioberger: Is `aws` already a dependency in this project? I don't think it should be. If it isn't, won't this fail for anyone who doesn't have it installed locally? Can you start a quip doc in the documentation folder called `0x.js developer setup` and add instructions there on installing `aws`?

fabioberger: Also, we need to make these files public:

```
--grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers
```

PR #79, opened 2017-06-22
Add kovan integration tests

LogvinovLeon:
This PR:
* Adds integration tests that check that contracts are deployed on kovan network

### DIFF
--- a/src/globals.d.ts
+++ b/src/globals.d.ts
@@ -87,10 +87,10 @@ declare module 'ethereumjs-abi' {
     const soliditySHA3: (argTypes: string[], args: any[]) => Buffer;
 }
 
+// truffle-hdwallet-provider declarations
 declare class HDWalletProvider {
     constructor(mnemonic: string, rpcUrl: string);
 }
-
 declare module 'truffle-hdwallet-provider' {
     export = HDWalletProvider;
 }
--- a/src/globals.d.ts
+++ b/src/globals.d.ts
@@ -86,3 +86,11 @@ declare module 'es6-promisify' {
 declare module 'ethereumjs-abi' {
     const soliditySHA3: (argTypes: string[], args: any[]) => Buffer;
 }
+
+declare class HDWalletProvider {
+    constructor(mnemonic: string, rpcUrl: string);
+}
+
+declare module 'truffle-hdwallet-provider' {
+    export = HDWalletProvider;
+}
--- a/test/artifacts_test.ts
+++ b/test/artifacts_test.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs';
 import * as _ from 'lodash';
 import * as chai from 'chai';
 import * as BigNumber from 'bignumber.js';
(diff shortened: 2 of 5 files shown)


test/artifacts_test.ts
```diff
@@ -0,0 +1,54 @@
...
+chaiSetup.configure();
+const expect = chai.expect;
+
+// Those tests are slower cause they're talking to a remote node
+const TIMEOUT = 10000;
+
+describe('Artifacts', () => {
+    describe('contracts are deployed on kovan', () => {
+        const kovanRpcUrl = 'https://kovan.0xproject.com';
+        const mnemonic = 'concert load couple harbor equip island argue ramp clarify fence smart topic';
```
fabioberger: Is there any way to extract this from the `package.json`?

test/artifacts_test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+// Those tests are slower cause they're talking to a remote node
+const TIMEOUT = 10000;
+
+describe('Artifacts', () => {
+    describe('contracts are deployed on kovan', () => {
+        const kovanRpcUrl = 'https://kovan.0xproject.com';
```
fabioberger: Let's more this to `constants.ts`

src/globals.d.ts
```diff
@@ -86,3 +86,11 @@ declare module 'es6-promisify' {
 declare module 'ethereumjs-abi' {
     const soliditySHA3: (argTypes: string[], args: any[]) => Buffer;
 }
+
+declare class HDWalletProvider {
+    constructor(mnemonic: string, rpcUrl: string);
+}
+
```
fabioberger: Let's remove this space and add a comment above these two declarations similar to the other declarations in this file.

test/artifacts_test.ts
```diff
@@ -0,0 +1,54 @@
...
+                makerTokenAddress,
+                takerTokenAddress,
+                salt: ZeroEx.generatePseudoRandomSalt(),
+                feeRecipient: ZeroEx.NULL_ADDRESS,
+                expirationUnixTimestampSec: new BigNumber(2524604400),
+            };
+            const orderHash = await zeroEx.getOrderHashHexAsync(order);
+        }).timeout(TIMEOUT);
+    });
+});
```
fabioberger: Instead of these indirect tests, why don't we call:

```
await (zeroEx.exchange as any)._getExchangeContractAsync()
```

```
await (zeroEx.token as any)._getProxyAddressAsync();
```

```
await (zeroEx.tokenRegistry as any)._getTokenRegistryContractAsync();
```

This would make this tests more concise and would test only the things we care about testing.

test/artifacts_test.ts
```diff
@@ -0,0 +1,28 @@
...
+import {ZeroEx} from '../src';
+import {web3Factory} from './utils/web3_factory';
+import {constants} from './utils/constants';
+
+// Those tests are slower cause they're talking to a remote node
+const TIMEOUT = 10000;
+
+describe('Artifacts', () => {
+    describe('contracts are deployed on kovan', () => {
+        const kovanRpcUrl = constants.KOVAN_URL;
```
fabioberger: Let's name the constant `KOVAN_RPC_URL`

PR #81, opened 2017-06-25
Add Wrapped ETH Support

fabioberger:
This PR:
- Adds support for calling `deposit` and `withdraw` on an Ether Wrapper ERC20 token contract.
- Adds tests for testing both methods and that the correct error messages are thrown if the caller has an insufficient balance or ETH/WETH.
- Adds a missing return comment for `getBalanceAsync`.


src/contract_wrappers/ether_token_wrapper.ts
```diff
@@ -0,0 +1,76 @@
...
+     * @param   amountInWei  Amount of ETH in Wei the caller wishes to withdraw.
+     * @param   withdrawer   The hex encoded user Ethereum address that would like to make the withdrawl.
+     */
+    public async withdrawAsync(amountInWei: BigNumber.BigNumber, withdrawer: string): Promise<void> {
+        assert.isBigNumber('amountInWei', amountInWei);
+        await assert.isSenderAddressAsync('withdrawer', withdrawer, this._web3Wrapper);
+
+        const wethContractAddress = await this.getContractAddressAsync();
+        const WETHBalanceInBaseUnits = await this._tokenWrapper.getBalanceAsync(wethContractAddress, withdrawer);
+        assert.assert(WETHBalanceInBaseUnits.gte(amountInWei), ZeroExError.INSUFFICIENT_WETH_BALANCE_FOR_WITHDRAWL);
```
LogvinovLeon: WITHDRAWAL

test/ether_token_wrapper_test.ts
```diff
@@ -0,0 +1,99 @@
...
+import promisify = require('es6-promisify');
+import {web3Factory} from './utils/web3_factory';
+import {ZeroEx, ZeroExError, Token} from '../src';
+import {BlockchainLifecycle} from './utils/blockchain_lifecycle';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const blockchainLifecycle = new BlockchainLifecycle();
+
+describe.only('EtherTokenWrapper', () => {
```
LogvinovLeon: remove only

src/web3_wrapper.ts
```diff
@@ -34,9 +34,14 @@ export class Web3Wrapper {
             return undefined;
         }
     }
+    public toWei(ethAmount: BigNumber.BigNumber): BigNumber.BigNumber {
+        const balanceWei = this.web3.toWei(ethAmount, 'ether');
+        return balanceWei;
+    }
     public async getBalanceInEthAsync(owner: string): Promise<BigNumber.BigNumber> {
```
LogvinovLeon: Why do we have this function? Wouldn't it be better if we only operate on wei?

fabioberger: Yeah, that makes sense to me. This was a legacy method ported from OTC. I'll refactor it to return wei.

PR #87, opened 2017-07-01
Wrap all event args in a newer version of BigNumber and test it

LogvinovLeon:
This PR:
* Wraps all event args in a newer version of BigNumber
* Adds a test for it
* Slightly refactor subscription tests


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -583,14 +585,30 @@ export class ExchangeWrapper extends ContractWrapper {
...
                 await promisify(event.stopWatching, event)();
             },
         };
         return zeroExEvent;
     }
+    private _wrapEventCallback(eventCallback: EventCallback): EventCallback {
+        const bignumberWrappingEventCallback = (err: Error, event: ContractEvent) => {
+            if (_.isNull(err)) {
+                const wrapIfBigNumber = (value: ContractEventArg): ContractEventArg => {
+                    return value instanceof String ? value : new BigNumber(value);
```
fabioberger: Should is `_.isString` here instead of `instanceof`.

fabioberger: Also, is it always true that either a string or number is returned here? We just assume if it's not a string, it should be wrapped in a BigNumber.

fabioberger: Also, what if a value is returned that is too large to represent with a `number`. Wouldn't it be returned as a string?

test/exchange_wrapper_test.ts
```diff
@@ -723,33 +716,43 @@ describe('ExchangeWrapper', () => {
...
+                                                                         indexFilterValues);
+                zeroExEvent.watch((err: Error, event: ContractEvent) => {
+                    const args = event.args as LogFillContractEventArgs;
+                    expect(args.filledValueM.isBigNumber).to.be.true();
+                    expect(args.filledValueT.isBigNumber).to.be.true();
+                    expect(args.feeMPaid.isBigNumber).to.be.true();
+                    expect(args.feeTPaid.isBigNumber).to.be.true();
+                    done();
+                });
+                await zeroEx.exchange.fillOrderAsync(
```
fabioberger: To really test this out, fill a really big order (an amount that wouldn't fit in a JS `number`). See if this still works for that edge-case.

src/contract_wrappers/exchange_wrapper.ts
LogvinovLeon: Its either a string or a bignumber.
There is no reliable way to check for a big-number that's why I check for a string

test/exchange_wrapper_test.ts
LogvinovLeon: All numbers from web3 are big numbers. Even zeroes

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -583,14 +585,30 @@ export class ExchangeWrapper extends ContractWrapper {
...
         const zeroExEvent = {
-            watch: event.watch.bind(event),
+            watch,
             stopWatchingAsync: async () => {
                 await promisify(event.stopWatching, event)();
             },
         };
         return zeroExEvent;
     }
+    private _wrapEventCallback(eventCallback: EventCallback): EventCallback {
```
fabioberger: `getBigNumberWrappingEventCallback `

PR #89, opened 2017-07-03
Add zeroEx.proxy

LogvinovLeon:
This PR:
* Makes proxy a public property on a zeroEx instance
* Exposes `zeroEx.proxy.isAuthorizedAsync`
* Adds `zeroEx.proxy.getAuthorizedAddressesAsync`

### DIFF
--- a/src/contract_wrappers/proxy_wrapper.ts
+++ b/src/contract_wrappers/proxy_wrapper.ts
@@ -22,6 +22,16 @@ export class ProxyWrapper extends ContractWrapper {
         const isAuthorized = await proxyContractInstance.authorized.call(exchangeContractAddress);
         return isAuthorized;
     }
+    /**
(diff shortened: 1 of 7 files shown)


src/contract_wrappers/proxy_wrapper.ts
```diff
@@ -22,6 +22,16 @@ export class ProxyWrapper extends ContractWrapper {
         const isAuthorized = await proxyContractInstance.authorized.call(exchangeContractAddress);
         return isAuthorized;
     }
+    /**
+     * Get the list of all Exchange contract addresses authorized by the Proxy contract.
+     * @param   exchangeContractAddress     The hex encoded address of the Exchange contract to use.
```
fabioberger: Let's use `call` instead of `use` in all methods where we are doing a `call`.

PR #90, opened 2017-07-03
Add implementation and tests for zeroEx.token.subscribeAsync

LogvinovLeon:
This PR:
* Refactors out some event helper functions to eventUtils
* Adds initial implementation and tests for `zeroEx.token.subscribeAsync`
* Adds parameter assertions for `subscribeAsync`
* Adds schemas for `orderHash`, `subscriptionOpts` and `indexFiltervalues`
* Removes `isValidOrderHash` assertion and use schema instead
* Makes `invalidateContractInstancesAsync` private on `token` and `exchange`
* Adds `assert.doesBelongToSrtingEnum`
* Defines `StringEnum` type
* Adds `zeroEx.etherToken._invalidateContractInstance`
* Fixes the bug when `zeroEx.setProviderAsync` didn't invalidate etherToken contract instance
* Adds `exchangeContractAddressAsync` to `orderSchema`


src/schemas/subscription_opts_schema.ts
```diff
@@ -0,0 +1,21 @@
...
+    ],
+};
+
+export const subscriptionOptsSchema = {
+    id: '/subscriptionOpts',
+    properties: {
+        fromBlock: {$ref: '/blockParam'},
+        toBlock: {$ref: '/blockParam'},
+    },
+    required: ['fromBlock', 'toBlock'],
```
fabioberger: These are not both required. They are both optional.

src/types.ts
```diff
@@ -289,8 +306,14 @@ export const ExchangeEvents = strEnum([
...
 
+export const TokenEvents = strEnum([
+    'Transfer',
+    'Approval',
+]);
+export type TokenEvents = keyof typeof TokenEvents;
+
 export interface IndexedFilterValues {
-    [index: string]: any;
+    [index: string]: ContractEventArg;
```
fabioberger: Nice! 💃

src/utils/assert.ts
```diff
@@ -27,6 +28,16 @@ export const assert = {
         const web3 = new Web3();
         this.assert(web3.isAddress(value), this.typeAssertionMessage(variableName, 'ETHAddressHex', value));
     },
+    doesBelongToStringEnum(variableName: string, value: string, stringEnum: StringEnum): void {
+        const doesBelongToStringEnum = !_.isUndefined(stringEnum[value]);
+        const enumValues = _.keys(stringEnum);
+        const enumValuesAsStrings = _.map(enumValues, enumValue => '\'' + enumValue + '\'');
```
fabioberger: Let's use template string here too:

```
`'${enumValue}'`
```

src/utils/assert.ts
```diff
@@ -27,6 +28,16 @@ export const assert = {
...
+        const enumValuesAsString = enumValuesAsStrings.join(', ');
+        assert.assert(
+            doesBelongToStringEnum,
+            `Expected ${variableName} to be one of: ${enumValuesAsString}, encountered: ${value}`,
```
fabioberger: Nice! This is a great error message 🍡

src/utils/event_utils.ts
```diff
@@ -0,0 +1,44 @@
...
+                    const isWeb3BigNumber = web3BigNumber.toString() === value.constructor.toString();
+                    return isWeb3BigNumber ?  new BigNumber(value) : value;
+                };
+                event.args = _.mapValues(event.args, wrapIfBigNumber);
+            }
+            eventCallback(err, event);
+        };
+        return bignumberWrappingEventCallback;
+    },
+    wrapEventEmitter(event: ContractEventObj): ContractEventEmitter {
```
fabioberger: Let's put this method at the top of this module.

src/utils/event_utils.ts
```diff
@@ -0,0 +1,44 @@
...
+import * as BigNumber from 'bignumber.js';
+import promisify = require('es6-promisify');
+
+export const eventUtils = {
+    /**
+     * Wraps eventCallback function so that all the BigNumber arguments are wrapped in a newer version of BigNumber.
+     * @param eventCallback     Event callback function to be wrapped
+     * @return Wrapped event callback function
+     */
+    getBigNumberWrappingEventCallback(eventCallback: EventCallback): EventCallback {
```
fabioberger: This should have an underscore in-front of its name since it is only called from within this module.

test/token_wrapper_test.ts
```diff
@@ -231,4 +241,104 @@ describe('TokenWrapper', () => {
...
+                    tokenAddress, TokenEvents.Transfer, subscriptionOpts, indexFilterValues);
+                eventSubscriptionToBeStopped.watch((err: Error, event: ContractEvent) => {
+                    done(new Error('Expected this subscription to have been stopped'));
+                });
+                await eventSubscriptionToBeStopped.stopWatchingAsync();
+                await zeroEx.token.transferAsync(tokenAddress, coinbase, addressWithoutFunds, transferAmount);
+                done();
+            })().catch(done);
+        });
+        it('Should wrap all event args BigNumber instances in a newer version of BigNumber', (done: DoneCallback) => {
```
fabioberger: Do we already have this test somewhere else? If so, this seems duplicitous since `tokenWrapper` and `exchangeWrapper` use the same methods found in `eventUtils`.

PR #93, opened 2017-07-04
Testrpc snapshots

LogvinovLeon:
This PR:
* Starts using testrpc snapshots in CI builds, which speeds them up https://quip.com/RAMmADOXkqYB
* Changes prepublish script to prepublishOnly https://iamakulov.com/notes/npm-4-prepublish/

Those changes speed up an average test run from 8mins to 5mins.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,13 +1,15 @@
 machine:
   node:
     version: 6.1.0
+  environment:
+      CONTRACTS_COMMIT: cd04d7c
 
 test:
   override:
-    - npm run testrpc:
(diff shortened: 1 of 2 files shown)


circle.yml
```diff
@@ -1,13 +1,15 @@
 machine:
   node:
     version: 6.1.0
+  environment:
+      CONTRACTS_COMMIT: cd04d7c
```
fabioberger: Let's name this `CONTRACTS_COMMIT_HASH`

PR #100, opened 2017-07-07
Improve signOrderHashAsync

fabioberger:
This PR:
Modifies signOrderHashAsync to parse the signatureHex string as V + R + S AND R + S + V and check both for a valid signature in order to fix the issue of different clients (and client versions) returning it in one of the two formats.

### DIFF
--- a/src/0x.ts
+++ b/src/0x.ts
@@ -221,39 +221,27 @@ export class ZeroEx {
 
         const signature = await this._web3Wrapper.signTransactionAsync(signerAddress, msgHashHex);
 
-        let signatureData;
-        const [nodeVersionNumber] = findVersions(nodeVersion);
-        // Parity v1.6.6 and earlier returns the signatureData as vrs instead of rsv as Geth does
-        // Later versions return rsv but for the time being we still want to support version < 1.6.6
-        // Date: May 23rd 2017
-        const latestParityVersionWithVRS = '1.6.6';
-        const isVersionBeforeParityFix = compareVersions(nodeVersionNumber, latestParityVersionWithVRS) <= 0;
-        if (isParityNode && isVersionBeforeParityFix) {
-            const signatureBuffer = ethUtil.toBuffer(signature);
-            let v = signatureBuffer[0];
-            if (v < 27) {
-                v += 27;
+        // HACK: There is no consensus on whether the signatureHex string should be formatted as
+        // v + r + s OR r + s + v, and different clients (even different versions of the same client)
+        // return the signature params in different orders. In order to support all client implementations,
+        // we parse the signature in both ways, and evaluate if either one is a valid signature.
+        const ecSignatureVRS = this.parseSignatureHexAsVRS(signature);
+        if (ecSignatureVRS.v === 27 || ecSignatureVRS.v === 28) {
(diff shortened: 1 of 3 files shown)


src/0x.ts
```diff
@@ -221,39 +221,27 @@ export class ZeroEx {
...
-        if (isParityNode && isVersionBeforeParityFix) {
-            const signatureBuffer = ethUtil.toBuffer(signature);
-            let v = signatureBuffer[0];
-            if (v < 27) {
-                v += 27;
+        // HACK: There is no consensus on whether the signatureHex string should be formatted as
+        // v + r + s OR r + s + v, and different clients (even different versions of the same client)
+        // return the signature params in different orders. In order to support all client implementations,
+        // we parse the signature in both ways, and evaluate if either one is a valid signature.
+        const ecSignatureVRS = this.parseSignatureHexAsVRS(signature);
```
LogvinovLeon: Please refactor those two functions to a `sig_utils` module.
Or if you insist in keeping them on a main class at least prefix them with an `_`

src/0x.ts
```diff
@@ -293,4 +281,28 @@ export class ZeroEx {
...
     }
+    private parseSignatureHexAsVRS(signatureHex: string): ECSignature {
+        const signatureBuffer = ethUtil.toBuffer(signatureHex);
+        let v = signatureBuffer[0];
+        if (v < 27) {
+            v += 27;
+        }
+        const r = signatureBuffer.slice(1, 33);
+        const s = signatureBuffer.slice(33, 65);
+        const ecSignature: ECSignature = {
```
LogvinovLeon: Can we have a function that converts signatureData to ECSignature? That way we can avoid code duplication.

src/0x.ts
```diff
@@ -221,39 +221,27 @@ export class ZeroEx {
...
-            const signatureBuffer = ethUtil.toBuffer(signature);
-            let v = signatureBuffer[0];
-            if (v < 27) {
-                v += 27;
+        // HACK: There is no consensus on whether the signatureHex string should be formatted as
+        // v + r + s OR r + s + v, and different clients (even different versions of the same client)
+        // return the signature params in different orders. In order to support all client implementations,
+        // we parse the signature in both ways, and evaluate if either one is a valid signature.
+        const ecSignatureVRS = this.parseSignatureHexAsVRS(signature);
+        if (ecSignatureVRS.v === 27 || ecSignatureVRS.v === 28) {
```
LogvinovLeon: `_.includes([27, 28], ecSignatureVRS.v)`

LogvinovLeon: Also refactor `[27, 28]` to a variable called: `validVRange`

src/0x.ts
fabioberger: Since I do not see us reusing these methods anywhere else in the codebase, we will keep them here for now. I will add the prefix however, good catch.

src/0x.ts
fabioberger: I don't consider this a serious form of code duplication. It doesn't include any logic, only instantiating an instance of the return type. It's also only in two places so it's fine for now.

src/0x.ts
fabioberger: Actually will move the private methods out so they don't show up in ZeroEx's method autocomplete list.

PR #103, opened 2017-07-11
Jsonschema types

LogvinovLeon:
This PR:
* Fixes failing tests because the jsonschema type definitions does not have `$ref` field

I don't know, why it was not failing before.

### DIFF
--- a/src/globals.d.ts
+++ b/src/globals.d.ts
@@ -26,6 +26,54 @@ declare namespace Chai {
 }
 /* tslint:enable */
 
+// jsonschema declarations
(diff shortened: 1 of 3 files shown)


src/globals.d.ts
```diff
@@ -26,6 +26,54 @@ declare namespace Chai {
 }
 /* tslint:enable */
 
+// jsonschema declarations
+declare interface Schema {
```
fabioberger: Did you write this def from scratch? If you got it from somewhere, please add the source above it in a comment.

PR #106, opened 2017-07-11
Single exchange

LogvinovLeon:
This PR:
* Reverts to using single exchange contract


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -49,7 +49,7 @@ import {signedOrderSchema, orderSchema} from '../schemas/order_schemas';
 import {constants} from '../utils/constants';
 import {TokenWrapper} from './token_wrapper';
 import {decorators} from '../utils/decorators';
-import {ExchangeArtifactsByName} from '../exchange_artifacts_by_name';
+import * as ExchangeArtifacts from '../artifacts/Exchange.json';
```
fabioberger: Let's name this `ExchangeArtifact`, since it's a single artifact.

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -626,18 +608,24 @@ export class ExchangeWrapper extends ContractWrapper {
         await Promise.all(stopWatchingPromises);
         this._exchangeLogEventEmitters = [];
     }
+    /**
+     * Returns the Ethereum address of the Exchange contract being used.
```
fabioberger: This is a public method. Let's use `@returns`.

PR #109, opened 2017-07-17
Fees balance allowance validation

LogvinovLeon:
This PR:
* Performs correct balance and allowance validation before submitting an order in case ZRX in one of the tokens traded.


src/utils/order_validation_utils.ts
```diff
@@ -0,0 +1,89 @@
+import {ExchangeContractErrs, SignedOrder} from '../types';
+import {TokenWrapper} from '../contract_wrappers/token_wrapper';
+
+export class OrderValidationUtils {
+    public static async validateFillOrderBalancesAndAllowancesAndThrowIfInvalidAsync(
```
fabioberger: Let's remove `And` from the name, it makes it longer without adding any new information.

src/utils/order_validation_utils.ts
```diff
@@ -0,0 +1,89 @@
...
+        tokenWrapper: TokenWrapper, signedOrder: SignedOrder, fillTakerAmount: BigNumber.BigNumber,
```
fabioberger: We are passing `TokenWrapper` into each call in this class. Why don't we instantiate the class with it, and store it as an instance variable?

src/utils/order_validation_utils.ts
```diff
@@ -0,0 +1,89 @@
...
+        } else {
+            const takerZRXBalance = await tokenWrapper.getBalanceAsync(zrxTokenAddress, senderAddress);
+            const takerZRXAllowance = await tokenWrapper.getProxyAllowanceAsync(zrxTokenAddress, senderAddress);
+
+            if (fillTakerAmount.greaterThan(takerBalance)) {
+                throw new Error(ExchangeContractErrs.InsufficientTakerBalance);
+            }
+            if (fillTakerAmount.greaterThan(takerAllowance)) {
+                throw new Error(ExchangeContractErrs.InsufficientTakerAllowance);
+            }
```
fabioberger: We can remove some code duplication here by setting the amount being checked against the balance/allowance to either be equal to `fillTakerAmount.plus(signedOrder.takerFee)` or `fillTakerAmount ` and then only having the conditional checks for balance/allowance once instead of twice.

test/order_validation_test.ts
```diff
@@ -0,0 +1,203 @@
+import * as chai from 'chai';
+import {chaiSetup} from './utils/chai_setup';
+import * as Web3 from 'web3';
+import * as BigNumber from 'bignumber.js';
```
fabioberger: Always put external imports before internal ones.

test/order_validation_test.ts
```diff
@@ -0,0 +1,203 @@
...
+                const newAllowanceWhichIsLessThanFees = makerFee.minus(1);
+                await zeroEx.token.setProxyAllowanceAsync(zrxTokenAddress, takerAddress,
+                    newAllowanceWhichIsLessThanFees);
+                return expect(zeroEx.exchange.fillOrderAsync(
+                    signedOrder, fillTakerAmount, shouldCheckTransfer, takerAddress,
+                )).to.be.rejectedWith(ExchangeContractErrs.InsufficientTakerFeeAllowance);
+            });
+        });
+        describe('should throw on insufficient balance or allowance when makerToken is ZRX',
+        () => {
```
fabioberger: Let's indent this. Right now it looks like a separate thing but it's actually a callback arg of `describe`.

PR #118, opened 2017-08-15
Update artifacts with the latest mainnet version

LogvinovLeon:
This PR:
* Makes 0x.js operational on mainnet
* Makes adjustments for last contracts changes:
    * Removes url from token registry, tests and schemas
    * Rearranges arguments in `isRoundingError`


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -714,7 +714,7 @@ export class ExchangeWrapper extends ContractWrapper {
         await assert.isUserAddressAvailableAsync(this._web3Wrapper);
         const exchangeInstance = await this._getExchangeContractAsync();
         const isRoundingError = await exchangeInstance.isRoundingError.call(
-            demoninator, numerator, makerTokenAmount,
+            numerator, demoninator, makerTokenAmount,
```
fabioberger: Did these really get swapped?

package.json
```diff
@@ -40,7 +40,7 @@
     "run_mocha": "mocha lib/test/**/*_test.js --timeout 3000 --bail"
   },
   "config": {
-    "artifacts": "Proxy Exchange TokenRegistry Token Mintable EtherToken",
```
LogvinovLeon: We don't use it anywhere in our code. The testrpc coinbase already has tokens.

src/contract_wrappers/exchange_wrapper.ts
LogvinovLeon: No. We just did it wrong before.
I checked the contracts and it's: `function isRoundingError(uint numerator, uint denominator, uint target)`

PR #120, opened 2017-08-17
Remove unnecessary assertions & add regression tests

fabioberger:
This PR addresses https://github.com/0xProject/0x.js/issues/119

- Removes the unnecessary assertion that a user address must be available from three 0x.js methods:
    - `getBalanceAsync`
    - `getAllowanceAsync`
    - `_isRoundingErrorAsync`
- Added regression tests for `getBalanceAsync` and `getAllowanceAsync` to ensure this type of error does not re-occur. This was also an opportunity to support writing tests with a web3 provider that does not have any addresses which will be used to make future tests more robust.

### DIFF
--- a/src/subproviders/empty_wallet_subprovider.ts
+++ b/src/subproviders/empty_wallet_subprovider.ts
@@ -1,0 +1,25 @@
+import * as _ from 'lodash';
+import Web3 = require('web3');
+
+/*
+ * This class implements the web3-provider-engine subprovider interface and returns
+ * that the provider has no addresses when queried.
+ * Source: https://github.com/MetaMask/provider-engine/blob/master/subproviders/subprovider.js
+ */
+export class EmptyWalletSubProvider {
+    public handleRequest(payload: any, next: () => void, end: (err: Error|null, result: any) => void) {
+        switch (payload.method) {
+            case 'eth_accounts':
+                end(null, []);
+                return;
+
+            default:
+                next();
+                return;
+        }
+    }
+    // Required to implement this method despite not needing it for this subprovider
+    public setEngine(engine: any) {
+        // noop
+    }
+}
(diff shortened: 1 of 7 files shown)


src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,25 @@
+import * as _ from 'lodash';
```
LogvinovLeon: Unused import

src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,25 @@
...
+import Web3 = require('web3');
```
LogvinovLeon: `import * as Web3 from 'web3';`

src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,25 @@
...
+
+/*
+ * This class implements the web3-provider-engine subprovider interface and returns
+ * that the provider has no addresses when queried.
+ * Source: https://github.com/MetaMask/provider-engine/blob/master/subproviders/subprovider.js
+ */
+export class EmptyWalletSubProvider {
+    public handleRequest(payload: any, next: () => void, end: (err: Error|null, result: any) => void) {
```
LogvinovLeon:
```
export interface JSONRPCPayload {
    params: any[];
    method: string;
}
```

test/utils/web3_factory.ts
```diff
@@ -7,17 +7,21 @@ import ProviderEngine = require('web3-provider-engine');
 import RpcSubprovider = require('web3-provider-engine/subproviders/rpc');
 import * as Web3 from 'web3';
 import {constants} from './constants';
+import {EmptyWalletSubProvider} from '../../src/subproviders/empty_wallet_subprovider';
 
 export const web3Factory = {
-    create(): Web3 {
-        const provider = this.getRpcProvider();
+    create(hasAddresses: boolean = true): Web3 {
```
LogvinovLeon: I'd rather prefer `withAddresses`
`web3Factory.create(withAddresses)`, but it's subjective. Feel free to ignore

PR #139, opened 2017-08-25
Gas price config

LogvinovLeon:
This PR:
* Allows the user to specify gas price for transactions when instantiating `ZeroEx`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+vTBD - _TBD_
+    * Added a possibility to specify `gasPrice` when instantiating `ZeroEx` (#139)
+
 v0.11.0 - _August 24, 2017_
 ------------------------
     * Added `zeroEx.token.setUnlimitedProxyAllowanceAsync` (#137)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # CHANGELOG
 
 vTBD - _TBD_
-    * Added a possibility to specify `gasPrice` when instantiating `ZeroEx` (#139)
+    * Added the ability to specify `gasPrice` when instantiating `ZeroEx` (#139)
 
 v0.11.0 - _August 24, 2017_
 ------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 vTBD - _TBD_
+    * Added the optional `zeroExConfig`  parameter to the constructor of `ZeroEx` (#139)
     * Added the ability to specify `gasPrice` when instantiating `ZeroEx` (#139)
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # CHANGELOG
 
+vTBD - _TBD_
+    * Added a possibility to specify `gasPrice` when instantiating `ZeroEx` (#139)
```
fabioberger: `Added the ability to specify `gasPrice` when instantiating `ZeroEx` (#139)`

src/0x.ts
```diff
@@ -162,15 +162,16 @@ export class ZeroEx {
      * Instantiates a new ZeroEx instance that provides the public interface to the 0x.js library.
      * @param   provider    The Web3.js Provider instance you would like the 0x.js library to use for interacting with
      *                      the Ethereum network.
+     * @param   gasPrice    The gas price to use for sending transactions. Defaults to 21 GWei.
      * @return  An instance of the 0x.js ZeroEx class.
      */
-    constructor(provider: Web3Provider) {
+    constructor(provider: Web3Provider, gasPrice?: BigNumber.BigNumber) {
```
fabioberger: So I really do feel strongly that we should make this a config obj called `configs`. I am 100% sure this won't be the last config param we will add to this library and it'll be cleaner to write:

```
const configs = {
  gasPrice,
  requestTimeoutLimit,
  verbose,
};
const zeroEx = new ZeroEx(provider, configs);
```

Then:

```
const zeroEx = new ZeroEx(provider, gasPrice, requestTimeoutLimit, verbose);
```

Especially as the number of configs grows.

src/0x.ts
```diff
@@ -162,15 +162,16 @@ export class ZeroEx {
...
         this._web3Wrapper = new Web3Wrapper(provider);
-        this.token = new TokenWrapper(this._web3Wrapper);
-        this.proxy = new TokenTransferProxyWrapper(this._web3Wrapper);
-        this.exchange = new ExchangeWrapper(this._web3Wrapper, this.token);
-        this.tokenRegistry = new TokenRegistryWrapper(this._web3Wrapper);
-        this.etherToken = new EtherTokenWrapper(this._web3Wrapper, this.token);
+        this.token = new TokenWrapper(this._web3Wrapper, gasPrice);
```
fabioberger: The above change will also keep all of these sub-classes cleaner too!

LogvinovLeon: I'll leave those subclasses alone for now, cause they're private and it's easy to change their interface later.

fabioberger: @LogvinovLeon if you don't do it now, you will do it once you add another param. Are you sure you don't want to take care of it now?

LogvinovLeon:
Why would we want to pass the whole library config object down to specific classes?
It's like passing the whole redux state to pure components.
Later - it there is enough configs - we can rethink it.

PR #142, opened 2017-08-30
Feature/support web3@1.0 provider

LogvinovLeon:
This PR:
* Starts supporting web3@1.0 providers and adds a test for it.


package.json
```diff
@@ -85,6 +85,7 @@
     "types-bn": "^0.0.1",
     "types-ethereumjs-util": "^0.0.5",
     "typescript": "^2.4.1",
+    "web3_beta": "https://github.com/ethereum/web3.js#1.0",
```
fabioberger: isn't there a more succinct way of pointing to a github repo from package.json? Something like `ethereum/web3.js#1.0`

src/0x.ts
```diff
@@ -163,6 +163,12 @@ export class ZeroEx {
      * @return  An instance of the 0x.js ZeroEx class.
      */
     constructor(provider: Web3Provider, config?: ZeroExConfig) {
+        if (_.isUndefined((provider as any).sendAsync)) {
```
fabioberger: If the user passes in something that is **not** a provider, it also won't have the `sendAsync` method. It will also not have the `send` method. We need to also check if it has the `send` method, otherwise we want to throw a helpful error message about how an invalid provider was supplied. Maybe:

```
Invalid `provider` supplied to ZeroEx constructor. A provider must have a `send` or `sendAsync` method in order to be valid.
```

src/0x.ts
```diff
@@ -163,6 +163,12 @@ export class ZeroEx {
...
+            // Web3@1.0 provider doesn't support synchronous http requests,
+            // so it only has `send` method, instead of `send and `sendAsync` in web3@0.x.x
```
fabioberger: `so it only has an async `send` method, instead of a `send and `sendAsync` in web3@0.x.x`

src/0x.ts
```diff
@@ -163,6 +163,12 @@ export class ZeroEx {
...
+            // That's why we reassign the send method.
```
fabioberger: We re-assign the `send` method so that Web3@1.0 providers work with 0x.js

test/web3_beta_test.ts
```diff
@@ -0,0 +1,16 @@
...
+import {web3Factory} from './utils/web3_factory';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+describe('ZeroEx with beta web3', () => {
+    const web3_beta_provider = web3Factory.getProviderBeta();
+    const zeroEx = new ZeroEx(web3_beta_provider);
+    it('is able to make a call using a beta provider', async () => {
+        await zeroEx.tokenRegistry.getTokenAddressesAsync();
```
fabioberger: This throws if you comment out the new logic in the constructor?

LogvinovLeon: Yes

src/utils/assert.ts
```diff
@@ -61,6 +61,10 @@ export const assert = {
     isBoolean(variableName: string, value: boolean): void {
         this.assert(_.isBoolean(value), this.typeAssertionMessage(variableName, 'boolean', value));
     },
+    isWeb3Provider(variableName: string, value: Web3.Provider): void {
+        const isWeb3Provider = _.isFunction((value as any).send) || _.isFunction((value as any).sendAsync);
```
fabioberger: Does the `Web3.Provider` type not have the `send` or `sendAsync` methods defined on it? Why is `any` necessary here?

LogvinovLeon: Currently it's an empty interface. I can update web3 types

PR #144, opened 2017-08-30
[WIP] ExchangeWrapper fillOrderAsync optimizations

BMillman19:
This PR:
* just a few comments for discussion

### DIFF
--- a/src/contract_wrappers/exchange_wrapper.ts
+++ b/src/contract_wrappers/exchange_wrapper.ts
@@ -147,33 +147,41 @@ export class ExchangeWrapper extends ContractWrapper {
     public async fillOrderAsync(signedOrder: SignedOrder, fillTakerTokenAmount: BigNumber.BigNumber,
                                 shouldThrowOnInsufficientBalanceOrAllowance: boolean,
                                 takerAddress: string): Promise<BigNumber.BigNumber> {
+        // validate input types
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         assert.isBigNumber('fillTakerTokenAmount', fillTakerTokenAmount);
         assert.isBoolean('shouldThrowOnInsufficientBalanceOrAllowance', shouldThrowOnInsufficientBalanceOrAllowance);
-        // this is already being done in validateFillOrderThrowIfInvalidAsync
-        await assert.isSenderAddressAsync('takerAddress', takerAddress, this._web3Wrapper);
 
-        // we can chain this with gas cost estimate and parallelize the resulting promise with validateFillOrderThrowIfInvalidAsync below
-        const exchangeInstance = await this._getExchangeContractAsync();
-
-        // we can do this optionally
-        await this.validateFillOrderThrowIfInvalidAsync(signedOrder, fillTakerTokenAmount, takerAddress);
-
+        // extract orderAdresses and orderValues from the provided order
         const [orderAddresses, orderValues] = ExchangeWrapper._getOrderAddressesAndValues(signedOrder);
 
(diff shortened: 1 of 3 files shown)


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -183,6 +188,7 @@ export class ExchangeWrapper extends ContractWrapper {
             },
         );
         this._throwErrorLogsAsErrors(response.logs);
+        // this may throw and error if logs is empty
```
LogvinovLeon: We always emit at least one event from Fill.
Either LogFill or LogError.
Here we have an even stronger warranty, cause the line before this one throws on LogError.

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -150,13 +150,18 @@ export class ExchangeWrapper extends ContractWrapper {
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         assert.isBigNumber('fillTakerTokenAmount', fillTakerTokenAmount);
         assert.isBoolean('shouldThrowOnInsufficientBalanceOrAllowance', shouldThrowOnInsufficientBalanceOrAllowance);
+        // this is already being done in validateFillOrderThrowIfInvalidAsync
```
fabioberger: We'd still want to do it here too in case this method gets refactored and no longer calls `validateFillOrderThrowIfInvalidAsync `

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -150,13 +150,18 @@ export class ExchangeWrapper extends ContractWrapper {
...
 
+        // we can chain this with gas cost estimate and parallelize the resulting promise with validateFillOrderThrowIfInvalidAsync below
         const exchangeInstance = await this._getExchangeContractAsync();
+
+        // we can do this optionally
         await this.validateFillOrderThrowIfInvalidAsync(signedOrder, fillTakerTokenAmount, takerAddress);
 
         const [orderAddresses, orderValues] = ExchangeWrapper._getOrderAddressesAndValues(signedOrder);
 
+        // can we do this optionally as well? we can have the caller pass in custom gas and gasPrice
```
fabioberger: Mh... don't think we want them to supply a gas amount.

src/contract_wrappers/exchange_wrapper.ts
BMillman19: hmm, lemme think on this, might be a way to save some rpcs

src/contract_wrappers/exchange_wrapper.ts
BMillman19: there is some possible optimization here, i suspect the estimated gas cost will be very similar between different fills

abandeali1: I agree that this `estimateGas` call isn't optimal. I think we should allow passing in the gas limit/price and fallback to `estimateGas`. Theres an generally an upper limit to the amount of gas a fill require, but it could be different with more complex tokens (e.g MiniMe tokens).

src/contract_wrappers/exchange_wrapper.ts
BMillman19: i think we should err on the side of caution and guard against an empty log here for a couple of reasons:
* its safer not to make assumptions about the implementation of the Exchange contract
* in the event that we do index an empty array, we want to give the user of the library a more descriptive error message

BMillman19:
Messed up this PR by rebasing..., please see https://github.com/0xProject/0x.js/pull/149 instead

PR #147, opened 2017-08-31
Fix incorrect types related to contract calls in types.ts

BMillman19:
This PR:
* Looked through various type definitions related to interactions with contracts
* Discovered incorrect types where Promises were expected but unwrapped types were defined (i.e Promise<number> vs number)
* This PR fixes these types, all tests pass

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -51,7 +51,7 @@ export interface ExchangeContract extends ContractInstance {
         call: () => Promise<string>;
     };
     getUnavailableTakerTokenAmount: {
-        call: (orderHash: string) => BigNumber.BigNumber;
+        call: (orderHash: string) => Promise<BigNumber.BigNumber>;
     };
     isRoundingError: {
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -60,63 +60,63 @@ export interface ExchangeContract extends ContractInstance {
...
     filled: {
-        call: (orderHash: string) => BigNumber.BigNumber;
+        call: (orderHash: string) => Promise<BigNumber.BigNumber>;
     };
     cancelled: {
-        call: (orderHash: string) => BigNumber.BigNumber;
+        call: (orderHash: string) => Promise<BigNumber.BigNumber>;
     };
     getOrderHash: {
         call: (orderAddresses: OrderAddresses, orderValues: OrderValues) => string;
```
LogvinovLeon: This one?

BMillman19: good catch

PR #148, opened 2017-09-01
Add 'content-type' header with value 'application/json' for http requ… 

BMillman19:
…ests made JSONRPC endpoints

This PR:
* the parity ethereum client rejects http requests that do not include a json content-type https://github.com/paritytech/parity/wiki/FAQ%3A-CLI%2C-Mining%2C-and-Networks#none-of-my-jsonrpc-requests-work-they-all-fail-with-no-output
* other clients may adopt a similar error in the future

### DIFF
--- a/test/utils/rpc.ts
+++ b/test/utils/rpc.ts
@@ -40,6 +40,9 @@ export class RPC {
             method: 'POST',
             uri: `http://${this.host}:${this.port}`,
             body: payload,
+            headers: {
+                'content-type': 'application/json'
+            },
         };
         const bodyString = await request(opts);
         const body = JSON.parse(bodyString);


test/utils/rpc.ts
```diff
@@ -40,6 +40,9 @@ export class RPC {
             method: 'POST',
             uri: `http://${this.host}:${this.port}`,
             body: payload,
+            headers: {
```
BMillman19: are there other instances in the library where we are forming raw http requests outside of web3?

LogvinovLeon: Not for now.
But there will be on the other branch. Thanks for pointing this one out, cause now I'll not forget to refactor this one.

PR #149, opened 2017-09-01
[WIP] Parallelization of fillOrderAsync

BMillman19:
This PR:
* This is a continuation of https://github.com/0xProject/0x.js/pull/144, please see for previous discussion 
Made some optimizations relating to parallelizing RPCs in fillOrderAsync
* Similar changes can be made in other methods in ExchangeWrapper
* There are some comments denoting RPCs we can possibly make optional
* Need to test against an ethereum client other than testrpc in order to quantify the gains
* All tests pass

### DIFF
--- a/src/contract_wrappers/exchange_wrapper.ts
+++ b/src/contract_wrappers/exchange_wrapper.ts
@@ -147,28 +147,41 @@ export class ExchangeWrapper extends ContractWrapper {
     public async fillOrderAsync(signedOrder: SignedOrder, fillTakerTokenAmount: BigNumber.BigNumber,
                                 shouldThrowOnInsufficientBalanceOrAllowance: boolean,
                                 takerAddress: string): Promise<BigNumber.BigNumber> {
+        // validate input types
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         assert.isBigNumber('fillTakerTokenAmount', fillTakerTokenAmount);
         assert.isBoolean('shouldThrowOnInsufficientBalanceOrAllowance', shouldThrowOnInsufficientBalanceOrAllowance);
-        await assert.isSenderAddressAsync('takerAddress', takerAddress, this._web3Wrapper);
 
-        const exchangeInstance = await this._getExchangeContractAsync();
-        await this.validateFillOrderThrowIfInvalidAsync(signedOrder, fillTakerTokenAmount, takerAddress);
-
+        // extract orderAdresses and orderValues from the provided order
         const [orderAddresses, orderValues] = ExchangeWrapper._getOrderAddressesAndValues(signedOrder);
 
-        const gas = await exchangeInstance.fillOrder.estimateGas(
-            orderAddresses,
-            orderValues,
-            fillTakerTokenAmount,
-            shouldThrowOnInsufficientBalanceOrAllowance,
-            signedOrder.ecSignature.v,
-            signedOrder.ecSignature.r,
-            signedOrder.ecSignature.s,
-            {
-                from: takerAddress,
-            },
-        );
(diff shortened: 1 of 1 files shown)


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -147,28 +147,41 @@ export class ExchangeWrapper extends ContractWrapper {
     public async fillOrderAsync(signedOrder: SignedOrder, fillTakerTokenAmount: BigNumber.BigNumber,
                                 shouldThrowOnInsufficientBalanceOrAllowance: boolean,
                                 takerAddress: string): Promise<BigNumber.BigNumber> {
+        // validate input types
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         assert.isBigNumber('fillTakerTokenAmount', fillTakerTokenAmount);
         assert.isBoolean('shouldThrowOnInsufficientBalanceOrAllowance', shouldThrowOnInsufficientBalanceOrAllowance);
-        await assert.isSenderAddressAsync('takerAddress', takerAddress, this._web3Wrapper);
```
fabioberger: Why was this assertion removed?

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -147,28 +147,41 @@ export class ExchangeWrapper extends ContractWrapper {
...
-            signedOrder.ecSignature.r,
-            signedOrder.ecSignature.s,
-            {
-                from: takerAddress,
-            },
-        );
+        // build a function that returns a promise that makes an RPC to retrieve the Exchange contract (may be memoized), 
+        // and then makes and RPC to estimate gas cost with the Exchange contract instance
+        const getExchangeInstanceAndGasAsync = async (): Promise<[ExchangeContract, number]> => {
+            const exchangeInstance = await this._getExchangeContractAsync();
```
fabioberger: This method is already memoized and returns the exchange contract instance if it's already been instantiated. This is also used by the `validateFillOrderThrowIfInvalidAsync ` method so I feel this should be called before the `Promise.all`. This will also make this method cleaner, and remove the nested array returns of the `Promise.all`.

src/contract_wrappers/exchange_wrapper.ts
BMillman19: It was my mistake to include this change in this PR, as it contributes to overall optimization of this method but does not relate to parallelizng RPCs, I still believe it should be deduplicated with the isSenderAddressAsync call in validateFillOrderThrowIfInvalidAsync. We should try to save as many RPCs as possible especially if the web3Provider is working with a remote ethereum client

fabioberger: Hm... My issue with that is `validateFillOrderThrowIfInvalidAsync` is called internally within `fillOrderAsync`. As part of the implementation, it is subject to change and could be removed in a subsequent refactor. I feel strongly that all public interface assertions should exist in the public method itself. I think a more appropriate refactor would be to call `this._orderValidationUtils.validateFillOrderThrowIfInvalidAsync` directly from this method rather then `validateFillOrderThrowIfInvalidAsync`.

BMillman19: I think I figured it out, `isSenderAddressAsync` is doing the task of talking to an ethereum client and verifying that it exists,  that should be left in `validateFillOrderThrowIfInvalidAsync` as is. We should change this line to `assert.isETHAddressHex()` I believe that is more in line with what the rest of the assertions are doing, as well as saves us 1 RPC =)

As a side note I noticed that `isETHAddressHex` instantiates an instance of a web3 object, is this a free operation? we may be better off passing in an existing instance of web3 or wrapper, similar to the interface of  `isSenderAddressAsync`

PR #151, opened 2017-09-04
Remove truffle contracts dependency

LogvinovLeon:
This PR:
* Removes `truffle-conract` 🥇 💯 👍 
* Starts using custom contract abstraction
* Stops awaiting transactions to be mined and returns transaction hash immidiately
* Adds `zeroEx.awaitTransactionMinedAsync`
* Removes `web3_beta` and tests, but still leaves support for the `web3_beta` provider, cause having both `web3`'s breaks the instalation.


src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+        this.defaults = defaults;
+        this.populateEvents();
+        this.populateFunctions();
+    }
+    private populateFunctions(): void {
+        const functionsAbi = _.filter(this.abi, abiPart => abiPart.type === 'function');
+        _.forEach(functionsAbi, (functionAbi: Web3.MethodAbi) => {
+            const cbStyleFunction = this.contract[functionAbi.name];
+            if (functionAbi.constant) {
+                this[functionAbi.name] = promisify(cbStyleFunction, this.contract);
```
fabioberger: Let's remove this one and force people to use `.callAsync`

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+                const cbStyleCallFunction = this.contract[functionAbi.name].call;
+                this[functionAbi.name].call = promisify(cbStyleCallFunction, this.contract);
```
fabioberger: Re-name to `callAsync`

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+            } else {
+                this[functionAbi.name] = this.promisifyWithDefaultParams(cbStyleFunction);
```
fabioberger: Let's only export `.sendTransactionAsync`

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+    constructor(web3ContractInstance: Web3.ContractInstance, defaults: Partial<Web3.TxData>) {
+        this.contract = web3ContractInstance;
+        this.address = web3ContractInstance.address;
+        this.abi = web3ContractInstance.abi;
+        this.defaults = defaults;
+        this.populateEvents();
+        this.populateFunctions();
+    }
+    private populateFunctions(): void {
+        const functionsAbi = _.filter(this.abi, abiPart => abiPart.type === 'function');
```
fabioberger: Let's add `function` and `event` to a type instead of using a string literal. We might end up using these in other parts of the codebase later and it would be great if we can easily grep for them.

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+            } else {
+                this[functionAbi.name] = this.promisifyWithDefaultParams(cbStyleFunction);
+                const cbStyleEstimateGasFunction = this.contract[functionAbi.name].estimateGas;
+                this[functionAbi.name].estimateGas =
+                    promisify(cbStyleEstimateGasFunction, this.contract);
+            }
+        });
+    }
+    private populateEvents(): void {
+        const eventsAbi = _.filter(this.abi, abiPart => abiPart.type === 'event');
```
fabioberger: Add `event` to type (see comment above).

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+        _.forEach(eventsAbi, (eventAbi: Web3.EventAbi) => {
+            this[eventAbi.name] = this.contract[eventAbi.name];
+        });
+    }
+    private promisifyWithDefaultParams(fn: (...args: any[]) => void): (...args: any[]) => Promise<any> {
+        const promisifiedWithDefaultParams = (...args: any[]) => {
+            const promise = new Promise((resolve, reject) => {
+                const lastArg = args[args.length - 1];
+                let txData: Partial<Web3.TxData> = {};
+                if (_.isObject(lastArg) && !_.isArray(lastArg) && !lastArg.isBigNumber) {
```
fabioberger: Let's make this validation stricter. We should make sure that it's an object that conforms to the txData schema.

src/contract.ts
```diff
@@ -0,0 +1,68 @@
...
+                    txData = args.pop();
+                }
+                txData = {
+                    ...txData,
+                    ...this.defaults,
```
fabioberger: The defaults should not overwrite the txData.

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -182,10 +181,7 @@ export class ExchangeWrapper extends ContractWrapper {
                 gas,
             },
         );
-        this._throwErrorLogsAsErrors(response.logs);
```
fabioberger: Remove this method

src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -730,7 +721,9 @@ export class ExchangeWrapper extends ContractWrapper {
         if (!_.isUndefined(this._exchangeContractIfExists)) {
             return this._exchangeContractIfExists;
         }
-        const contractInstance = await this._instantiateContractIfExistsAsync((ExchangeArtifacts as any));
+        const contractInstance = await this._instantiateContractIfExistsAsync<ExchangeContract>(
+            (ExchangeArtifacts as any),
```
fabioberger: Why did we not have to cast as `Artifact` here?

src/web3_wrapper.ts
```diff
@@ -41,6 +49,27 @@ export class Web3Wrapper {
...
+                                                                                       address?: string): Promise<A> {
+        if (_.isUndefined(address)) {
+            const networkIdIfExists = await this.getNetworkIdIfExistsAsync();
+            if (_.isUndefined(networkIdIfExists)) {
+                throw new Error(ZeroExError.NoNetworkId);
+            }
+            if (_.isUndefined(artifact.networks[networkIdIfExists])) {
+                throw new Error(ZeroExError.ContractNotDeployedOnNetwork);
+            }
+            address = artifact.networks[networkIdIfExists].address.toLowerCase();
```
fabioberger: Let's not overwrite function args.

LogvinovLeon:
Also remember to update the list of public types in the docs

src/contract.ts
```diff
@@ -0,0 +1,68 @@
+import * as Web3 from 'web3';
+import * as _ from 'lodash';
+import promisify = require('es6-promisify');
```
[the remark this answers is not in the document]
LogvinovLeon: Unfortunately. Whenever the package exports a function (non module object) we need to do it like that.
https://stackoverflow.com/a/35706271/3546986

src/globals.d.ts
```diff
@@ -69,3 +69,18 @@ declare class HDWalletProvider {
 declare module 'truffle-hdwallet-provider' {
     export = HDWalletProvider;
 }
+
+// abi-decoder declarations
+interface DecodedLogArg {
+    name: string;
+    value: string|BigNumber.BigNumber;
```
fabioberger: Can this really only be string or bignumber? What about boolean?

src/index.ts
```diff
@@ -30,5 +30,7 @@ export {
     ContractEventArgs,
     Web3Provider,
     ZeroExConfig,
-    TransactionReceipt,
+    TransactionReceiptWithDecodedLogs,
+    LogWithDecodedArgs,
+    DecodedLogArgs,
```
fabioberger: please put the relevant ones in public types list in `website`.

test/exchange_wrapper_test.ts
```diff
@@ -185,8 +186,9 @@ describe('ExchangeWrapper', () => {
                         makerTokenAddress, takerTokenAddress, makerAddress, takerAddress, fillableAmount,
                     );
                     const partialFillAmount = new BigNumber(3);
-                    await zeroEx.exchange.fillOrderAsync(
+                    const txHash = await zeroEx.exchange.fillOrderAsync(
                         signedOrder, partialFillAmount, shouldThrowOnInsufficientBalanceOrAllowance, takerAddress);
+                    await zeroEx.awaitTransactionMinedAsync(txHash);
                     expect(await zeroEx.token.getBalanceAsync(makerTokenAddress, makerAddress))
                         .to.be.bignumber.equal(fillableAmount.minus(partialFillAmount));
                     expect(await zeroEx.token.getBalanceAsync(takerTokenAddress, makerAddress))
```
fabioberger: Can we add a test that makes sure that the decodedLogs that are returned are of the expected format and values?

src/globals.d.ts
LogvinovLeon: We already had this discussion.
The result of it is `ContractEventArg` type in `types.ts`, which I can't import in `global.ts`.
We have a rule to not define unneccessary types before they're needed.
Currently we will decode only our events and our events have only those types

src/index.ts
LogvinovLeon: Will do after this is finalized and merged

test/0x.js_test.ts
```diff
@@ -204,4 +207,29 @@ describe('ZeroEx library', () => {
...
         });
     });
+    describe('#awaitTransactionMinedAsync', () => {
+        beforeEach(async () => {
+            await blockchainLifecycle.startAsync();
+        });
+        afterEach(async () => {
+            await blockchainLifecycle.revertAsync();
+        });
+        it('return transaction receipt with decoded logs', async () => {
```
fabioberger: `returns`

PR #156, opened 2017-09-06
Custom abi decoder

LogvinovLeon:
This PR:
* Removed abi-decoder as a dependency
* Reimplments a subset of it in TS
* Handles the case, when the log can't be decoded

### DIFF
--- a/src/utils/abi_decoder.ts
+++ b/src/utils/abi_decoder.ts
@@ -56,7 +56,7 @@ export class AbiDecoder {
     }
     private padZeros(address: string) {
         let formatted = address;
-        if (!_.includes(formatted, '0x')) {
+        if (_.startsWith(formatted, '0x')) {
             formatted = formatted.slice(2);
         }
 
--- a/src/utils/abi_decoder.ts
+++ b/src/utils/abi_decoder.ts
@@ -61,6 +61,6 @@ export class AbiDecoder {
         }
 
         formatted = _.padStart(formatted, 40, '0');
-        return '0x' + formatted;
+        return `0x${formatted}`;
     }
 }
--- a/src/utils/abi_decoder.ts
+++ b/src/utils/abi_decoder.ts
@@ -5,13 +5,13 @@ import * as SolidityCoder from 'web3/lib/solidity/coder';
 
 export class AbiDecoder {
     private savedABIs: Web3.AbiDefinition[] = [];
-    private methodIDs: {[signatureHash: string]: Web3.EventAbi} = {};
+    private methodIds: {[signatureHash: string]: Web3.EventAbi} = {};
     constructor(abiArrays: Web3.AbiDefinition[][]) {
         _.map(abiArrays, this.addABI.bind(this));
     }
     public decodeLog(logItem: Web3.LogEntry): DecodedArgs|undefined {
-        const methodID = logItem.topics[0];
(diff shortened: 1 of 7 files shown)


src/utils/abi_decoder.ts
```diff
@@ -0,0 +1,66 @@
+import * as Web3 from 'web3';
+import * as _ from 'lodash';
+import {AbiType, DecodedLogArgs, DecodedArgs} from '../types';
+import * as SolidityCoder from 'web3/lib/solidity/coder.js';
```
fabioberger: do you need the `.js` here?

src/utils/abi_decoder.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export class AbiDecoder {
+    private savedABIs: Web3.AbiDefinition[] = [];
+    private methodIDs: {[signatureHash: string]: Web3.EventAbi} = {};
+    constructor(abiArrays: Web3.AbiDefinition[][]) {
+        _.map(abiArrays, this.addABI.bind(this));
+    }
+    public decodeLog(logItem: Web3.LogEntry): DecodedArgs|undefined {
+        const methodID = logItem.topics[0];
```
fabioberger: rename this to `methodID`. Our convention is camelCase except for acronyms.

src/utils/abi_decoder.ts
```diff
@@ -0,0 +1,66 @@
...
+            _.map(event.inputs, (param: Web3.EventParameter) => {
+                let value;
+                if (param.indexed) {
+                    value = logItem.topics[topicsIndex];
+                    topicsIndex++;
+                } else {
+                    value = decodedData[dataIndex];
+                    dataIndex++;
+                }
+                if (param.type === 'address') {
```
fabioberger: Is this an `ABIType`? If so, let's use that type rather then string literal. Same goes for the subsequent ones.

src/utils/abi_decoder.ts
```diff
@@ -0,0 +1,66 @@
...
+                const signature = `${abi.name}(${_.map(abi.inputs, input => input.type).join(',')})`;
+                const signatureHash = new Web3().sha3(signature);
+                this.methodIDs[signatureHash] = abi;
+            }
+        });
+        this.savedABIs = this.savedABIs.concat(abiArray);
+    }
+    private padZeros(address: string) {
+      let formatted = address;
+      if (formatted.indexOf('0x') !== -1) {
```
fabioberger: Let's use `!_.includes(formatted, '0x')` instead of `indexOf`

src/utils/abi_decoder.ts
```diff
@@ -0,0 +1,66 @@
...
+        formatted = formatted.slice(2);
+      }
+
+      formatted = _.padStart(formatted, 40, '0');
+      return '0x' + formatted;
```
fabioberger: template string instead of concatenation.

src/utils/abi_decoder.ts
LogvinovLeon: It's not an `ABIType`.
```
export enum AbiType {
    Function = 'function',
    Constructor = 'constructor',
    Event = 'event',
    Fallback = 'fallback',
}
```
it's rather `SolidityType`, which we don't have defined yet and I'm rather lazy to define all the `ufixed128x19`s.

PR #163, opened 2017-09-12
Standard relayer api

LogvinovLeon:
This PR:
* Adds a `Relay` constructor on `ZeroEx`. When instantiated - will contain methods to interact with relays using standard relayer API.


src/0x.ts
```diff
@@ -42,6 +43,10 @@ bigNumberConfigs.configure();
  */
 export class ZeroEx {
     /**
+     * Relay constructor. When instantiated - will contain methods to interact with relays using standard relayer API.
```
fabioberger: `Relay constructor. Relay is a client for sending requests to Relayers that implement the Standard Relayer API.`

src/relay/index.ts
```diff
@@ -0,0 +1,40 @@
+import 'isomorphic-fetch';
+import {StandardRelayerApi, SignedOrder} from '../types';
+
+export class Relay {
+    private url: string;
+    private version: number;
+    constructor(url: string) {
```
fabioberger: Let's already add an options obj to the constructor with the `version` as one of it's optional params.

src/relay/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        this.url = url;
+        this.version = 0;
+    }
+    public async getTokenPairsAsync(): Promise<StandardRelayerApi.RelayerApiTokenTradeInfo[]> {
+        const tokenPairs = await this._requestAsync('token_pairs', 'GET');
```
fabioberger: I'd prefer we add the `/` at the beginning of the supplied path. It makes it very obvious that it's a path.

src/relay/index.ts
```diff
@@ -0,0 +1,40 @@
...
+        const fees = await this._requestAsync(`fees`, 'POST');
+        return fees;
+    }
+    public async submitOrderAsync(signedOrder: SignedOrder)
+    : Promise<void> {
+        await this._requestAsync(`order`, 'POST');
+    }
+    private async _requestAsync(path: string, method: string): Promise<any> {
+        const url = `${this.url}/v${this.version}/${path}`;
+        const response = await fetch(url, {
```
fabioberger: What if the response is a non-200? Right now we return the json response regardless... Remember that `fetch` doesn't throw ever. It always returns a response object.

src/types/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './types';
```
fabioberger: Let's always put imports before exports. And new line between. At first I thought this was an import from where it was placed.

src/types/relay.ts
```diff
@@ -0,0 +1,43 @@
...
+    tokenB: RelayerApiTokenTradeInfo;
+}
+
+export type OrderState = 'OPEN'|'EXPIRED'|'CLOSED'|'UNFUNDED';
+
+export interface RelayerApiOrderResponse {
+    signedOrder: SignedOrder;
+    state: OrderState;
+    pending: {
+        fillAmount: number;
```
fabioberger: Should be a bigNumber.

src/types/relay.ts
```diff
@@ -0,0 +1,43 @@
...
+        cancelAmount: number;
```
fabioberger: Should be a bigNumber.

src/types/relay.ts
```diff
@@ -0,0 +1,43 @@
...
+    };
+    remainingTakerTokenAmount: number;
```
fabioberger: Should be a bigNumber.

src/types/relay.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export interface RelayerApiFeesRequest {
+    maker: string;
+    taker: string;
+    makerTokenAddress: string;
+    takerTokenAddress: string;
+    makerTokenAmount: string;
+    takerTokenAmount: string;
```
fabioberger: So maybe I misunderstand but I feel like this client should also convert the received amounts and convert them to BigNumbers before returning the response to the caller. The fact that amounts are sent over-the-wire as strings is an implementation detail that shouldn't leak out of this interface.

test/relay_api_test.ts
```diff
@@ -0,0 +1,85 @@
...
+import {Web3Wrapper} from '../src/web3_wrapper';
+import {SchemaValidator, schemas} from '0x-json-schemas';
+
+chaiSetup.configure();
+chai.config.includeStack = true;
+const expect = chai.expect;
+
+// We need to have a mock API server in order to test that
+describe('Relay API', () => {
+    const relayUrl = 'https://example.com';
```
fabioberger: What about `https://api.relayer.com`

test/relay_api_test.ts
```diff
@@ -0,0 +1,85 @@
...
+    const relay = new ZeroEx.Relay(relayUrl);
+    const schemaValidator = new SchemaValidator();
+    const responsePayload = {example: 'data'};
```
fabioberger: Let's stub out the actual responses we expect, and make sure they are properly converted to usable objects (i.e BigNumbers instead of strings for amounts).

test/relay_api_test.ts
LogvinovLeon: `https://api.relayer.com` seems confusing. It's not clear - if it's a real domain.
example.com is a standard way to express, that it doesn't matter and it's an example domain.
No one will expect to get real api responces from this one.

fabioberger: come on...

src/relay/index.ts
```diff
@@ -0,0 +1,74 @@
+import 'isomorphic-fetch';
+import * as _ from 'lodash';
+import * as BigNumber from 'bignumber.js';
+import {StandardRelayerApi, SignedOrder} from '../types';
+
+export class Relay {
+    private url: string;
+    private version: number;
+    constructor(url: string, opts?: StandardRelayerApi.RelayOpts) {
```
fabioberger: All of the args in this class are part of the public interface and need assertions.

src/types/index.ts
fabioberger: Please add a space between the imports and exports.

test/fixtures/standard_relayer_api/order/0xdeadbeef.json
```diff
@@ -0,0 +1,27 @@
...
+            "s": "0x40349190569279751135161d22529dc25add4f6069af05be04cacbda2ace2254"
+        }
+    },
+    "state": "OPEN",
+    "pending": {
+        "fillAmount": "50000000000000000",
+        "cancelAmount": "50000000000000000"
+    },
+    "remainingTakerTokenAmount": "10000000000000000"
+}
```
fabioberger: Should this one be removed? Along with the .ts version?

test/relay_api_test.ts
```diff
@@ -0,0 +1,76 @@
...
+            const params = {
+                maker: '0x9e56625509c2f60af937f23b7b532600390e8c8b',
+                taker: '0xa2b31dacf30a9c50ca473337c01d8a201ae33e32',
+                makerTokenAddress: '0x323b5d4c32345ced77393b3530b1eed0f346429d',
+                takerTokenAddress: '0xef7fff64389b814a946f3e92105513705ca6b990',
+                makerTokenAmount: '10000000000000000000',
+                takerTokenAmount: '30000000000000000000',
+            };
+            fetchMock.post(`${relayUrl}/v0/fees`, JSON.stringify(feesResponseJSON));
+            const fees = await relay.getFeesAsync(params);
```
fabioberger: The params object passed to these methods should also have BigNumbers, and not string representations. The method should then convert it to the string rep. before sending it over-the-wire.

test/relay_api_test.ts
```diff
@@ -7,11 +7,13 @@ import {chaiSetup} from './utils/chai_setup';
...
 import {feesResponse} from './fixtures/standard_relayer_api/fees';
-import {orderResponse} from './fixtures/standard_relayer_api/order/0xdeadbeef';
+import {
+    orderResponse,
+} from './fixtures/standard_relayer_api/order/0xabc67323774bdbd24d94f977fa9ac94a50f016026fd13f42990861238897721f';
 import {ordersResponse} from './fixtures/standard_relayer_api/orders';
 import {tokenPairsResponse} from './fixtures/standard_relayer_api/token_pairs';
 import * as feesResponseJSON from './fixtures/standard_relayer_api/fees.json';
-import * as orderResponseJSON from './fixtures/standard_relayer_api/order/0xdeadbeef.json';
+import * as orderResponseJSON from './fixtures/standard_relayer_api/order/0xabc67323774bdbd24d94f977fa9ac94a50f016026fd13f42990861238897721f.json'; // tslint:disable-line
```
fabioberger: Let's use `disable-next-line` since this line is already super long.

PR #170, opened 2017-09-26
Add validateOrderFillableOrThrowAsync Method

fabioberger:
**Problem**: Relayers want a validation method with which to prune their orderbook. During pruning, they do not know who the order's eventual taker is going to be.

This PR:
* Adds a validation method that validates whether an order is fillable, without knowing the eventual taker. It validates that the order has not already been completely filled/cancelled, hasn't expired and that the maker still has a sufficient balance/allowance.
* The caller can optionally pass in a `opts.expectedFillTakerTokenAmount` to the validation. If omitted, the method ensures that the total remaining `takerTokenAmount` is fillable. If supplied, it validates that the order is fillable up to the passed `expectedFillTakerTokenAmount`. If a relayer wants to keep orders on their books that are partially unfillable, they could first query the makers balance/allowance and then call this method with `expectedFillTakerTokenAmount` set to the remaining maker balance that is also covered by a sufficient allowance to 0x.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 v0.17.0 - _September 26, 2017_
+    * Added `zeroEx.exchange.validateOrderFillableOrThrowAsync` to simplify orderbook pruning (#170)
     * Made `zeroEx.exchange.getZRXTokenAddressAsync` public (#171)
 
 v0.16.0 - _September 20, 2017_
--- a/src/contract_wrappers/exchange_wrapper.ts
+++ b/src/contract_wrappers/exchange_wrapper.ts
@@ -633,7 +633,7 @@ export class ExchangeWrapper extends ContractWrapper {
      *                          to validate for.
      */
     public async validateOrderFillableOrThrowAsync(
-        signedOrder: SignedOrder, opts: ValidateOrderFillableOpts,
+        signedOrder: SignedOrder, opts?: ValidateOrderFillableOpts,
     ): Promise<void> {
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         const zrxTokenAddress = await this._getZRXTokenAddressAsync();
--- a/src/contract_wrappers/exchange_wrapper.ts
+++ b/src/contract_wrappers/exchange_wrapper.ts
@@ -637,7 +637,7 @@ export class ExchangeWrapper extends ContractWrapper {
         signedOrder: SignedOrder, opts?: ValidateOrderFillableOpts,
     ): Promise<void> {
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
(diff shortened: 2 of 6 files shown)


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -624,6 +625,24 @@ export class ExchangeWrapper extends ContractWrapper {
...
     }
     /**
+     * Checks if order is still fillable and throws an error otherwise.
+     * @param   signedOrder     An object that conforms to the SignedOrder interface. The
+     *                          signedOrder you wish to validate.
+     * @param   opts            An object that conforms to the ValidateOrderFillableOpts
+     *                          interface. Allows specifying a specific fillTakerTokenAmount
+     *                          to validate for.
+     */
+    public async validateOrderFillableThrowIfNotFillableAsync(
```
LogvinovLeon: `validateOrderFillableOrThrowAsync`

src/utils/order_validation_utils.ts
```diff
@@ -12,6 +13,26 @@ export class OrderValidationUtils {
         this.tokenWrapper = tokenWrapper;
         this.exchangeWrapper = exchangeWrapper;
     }
+    public async validateOrderFillableThrowIfNotFillableAsync(
+        signedOrder: SignedOrder, zrxTokenAddress: string, expectedFillTakerTokenAmount?: BigNumber.BigNumber,
+    ): Promise<void> {
+        const orderHash = utils.getOrderHashHex(signedOrder);
+        const unavailableTakerTokenAmount = await this.exchangeWrapper.getUnavailableTakerAmountAsync(orderHash);
+        if (signedOrder.makerTokenAmount.eq(unavailableTakerTokenAmount)) {
```
LogvinovLeon: Can we make it that single error type is thrown only from one function?
And then this will look like:
```
this.validateIfExpired()
this.validateIfFillAmountZero();
```

I think - it will be more clear like that.

LogvinovLeon:
`(-0.6%)` is no good. Can we please add tests for this method?

test/order_validation_test.ts
```diff
@@ -54,6 +54,27 @@ describe('OrderValidation', () => {
     afterEach(async () => {
         await blockchainLifecycle.revertAsync();
     });
+    describe('validateOrderFillableOrThrowAsync', () => {
```
LogvinovLeon: Add success test

LogvinovLeon:
Add success test and a regression test and after that - good to go! 👍

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 v0.17.0 - _September 26, 2017_
+    * Added `zeroEx.exchange.validateOrderFillableOrThrowAsync` to simplify orderbook pruning (#170)
```
LogvinovLeon: This version is already published

PR #171, opened 2017-09-26
Make getZRXTokenAddressAsync public

LogvinovLeon:
This PR:
* Makes `zeroEx.exchange.getZRXTokenAddressAsync` public


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -708,6 +708,15 @@ export class ExchangeWrapper extends ContractWrapper {
             throw new Error(errMessage);
         }
     }
+    /**
+     * Returns an address of ZRX token used by the exchange contract.
```
fabioberger: `Returns the ZRX token address used...`

PR #172, opened 2017-09-27
Add OrderTransactionOpts to enable optional validation to exchange_wr…

BMillman19:
…apper

* Added OrderTransactionOpts type for specifying extra options when calling certain exchange_wrapper methods involving transactions
* Added tests

### DIFF
--- a/test/exchange_wrapper_test.ts
+++ b/test/exchange_wrapper_test.ts
@@ -114,18 +114,14 @@ describe('ExchangeWrapper', () => {
                         .to.be.rejectedWith(ExchangeContractErrs.OrderFillAmountZero);
                 });
                 it('should validate when orderTransactionOptions specify to validate', async () => {
-                    return expect(zeroEx.exchange.batchFillOrKillAsync(orderFillOrKillRequests, takerAddress,
-                        {
-                            shouldValidate: true,
-                        },
-                    )).to.be.rejectedWith(ExchangeContractErrs.OrderFillAmountZero);
+                    return expect(zeroEx.exchange.batchFillOrKillAsync(orderFillOrKillRequests, takerAddress, {
+                        shouldValidate: true,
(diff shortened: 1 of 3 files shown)


test/exchange_wrapper_test.ts
```diff
@@ -93,14 +94,51 @@ describe('ExchangeWrapper', () => {
...
+                });
+                it('should validate when orderTransactionOptions are not present', async () => {
+                    return expect(zeroEx.exchange.batchFillOrKillAsync(orderFillOrKillRequests, takerAddress))
+                        .to.be.rejectedWith(ExchangeContractErrs.OrderFillAmountZero);
+                });
+                it('should validate when orderTransactionOptions specify to validate', async () => {
+                    return expect(zeroEx.exchange.batchFillOrKillAsync(orderFillOrKillRequests, takerAddress,
+                        {
+                            shouldValidate: true,
+                        },
```
fabioberger: Nit: No need to put the start and end curly brackets on a new line when using an anonymous object as a method param.

test/exchange_wrapper_test.ts
```diff
@@ -323,6 +468,26 @@ describe('ExchangeWrapper', () => {
                     expect(cancelledAmount).to.be.bignumber.equal(cancelAmount);
                 });
             });
+            describe('order transaction options', () => {
+                it('should validate when orderTransactionOptions are not present', async () => {
+                    return expect(zeroEx.exchange.cancelOrderAsync(signedOrder, new BigNumber(0)))
```
fabioberger: Nit: Let's not in-line instantiating new objects in method params. Instead declare outside with a descriptive variable name of what it represents.

Our convention is also not to pass numbers/strings/booleans directly at params but to always assign them to variables with readable names. Otherwise the reader needs to look at the method definition to figure out what the value represents (saves future devs time).

PR #195, opened 2017-10-18
Reduce final bundle size by 11% (82kB)

LogvinovLeon:
Before: 785 kB http://goal-keeper-technology-17752.bitballoon.com/
After: 703 kB http://immigration-officer-porcupine-23410.bitballoon.com/
This PR:
* Changes how we use lodash, so that we don't include the whole bundle
* Updates ethereumjs-blockstream, cause the new version is significantly smaller
* Updates 0x-json-schemas, cause the new version doesn't depend on lodash
* Removes parts of the artifacts that we don't use

There are things to be done and I can bring it down by ~150kB, but it'll take more time and this was just a low-hanging fruit

This article has `lodash-es` magic explanation: https://medium.com/@martin_hotell/tree-shake-lodash-with-webpack-jest-and-typescript-2734fa13b5cd


LogvinovLeon:
This breaks ts declarations. I'll revert it for now and unpublish the version and I'll return to it later

PR #202, opened 2017-10-25
Bignumber types

LogvinovLeon:
This PR:
* Uses the new version of BigNumber that contains native types
* This allows us to write types as `BigNumber` instead of `BigNumber.BigNumber`
* This forces us to use a hack with chai namespace

### DIFF
--- a/src/globalsAugment.d.ts
+++ b/src/globalsAugment.d.ts
@@ -1,0 +1,18 @@
+import BigNumber from 'bignumber.js';
+
+declare global {
+    /* tslint:disable */
+    namespace Chai {
+        interface NumberComparer {
+            (value: number|BigNumber, message?: string): Assertion;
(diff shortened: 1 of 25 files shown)


src/globalsAugment.d.ts
```diff
@@ -0,0 +1,18 @@
+import BigNumber from 'bignumber.js';
```
fabioberger: Add a comment at the top of this file about how it is a hack in order to import the `bignumber.js` definitions from the module within a global definition.

src/globalsAugment.d.ts
```diff
@@ -0,0 +1,18 @@
...
+
+declare global {
+    /* tslint:disable */
```
fabioberger: The comment associated to this hack should also be here.

PR #205, opened 2017-11-09
Order watcher

LogvinovLeon:
This PR:
* Adds `EventWatcher` class
* Adds `OrderStateWatcher` class
* Adds four public methods:
    * zeroEx.orderStateWatcher.addOrder(signedOrder)
    * zeroEx.orderStateWatcher.removeOrder(orderHash)
    * zeroEx.orderStateWatcher.subscribe(callback)
    * zeroEx.orderStateWatcher.unsubscribe()

**Should not be merged** cause it will be used as a base branch for future order watcher related PRs.

### DIFF
--- a/src/order_watcher/event_watcher.ts
+++ b/src/order_watcher/event_watcher.ts
@@ -1,7 +1,12 @@
 import * as Web3 from 'web3';
 import * as _ from 'lodash';
 import {Web3Wrapper} from '../web3_wrapper';
-import {BlockParamLiteral, EventCallback, EventWatcherCallback} from '../types';
+import {
+    BlockParamLiteral,
+    EventCallback,
+    EventWatcherCallback,
+    ZeroExError,
+} from '../types';
 import {AbiDecoder} from '../utils/abi_decoder';
 import {intervalUtils} from '../utils/interval_utils';
 import {assert} from '../utils/assert';
@@ -28,6 +33,9 @@ export class EventWatcher {
     }
     public subscribe(callback: EventWatcherCallback): void {
         assert.isFunction('callback', callback);
+        if (!_.isUndefined(this._callbackIfExistsAsync)) {
+            throw new Error(ZeroExError.SubscriptionAlreadyPresent);
+        }
         this._callbackIfExistsAsync = callback;
         this._intervalIdIfExists = intervalUtils.setAsyncExcludingInterval(
             this._pollForBlockchainEventsAsync.bind(this), this._pollingIntervalMs,
--- a/src/order_watcher/event_watcher.ts
+++ b/src/order_watcher/event_watcher.ts
@@ -5,18 +5,18 @@ import {BlockParamLiteral, EventCallback, MempoolEventCallback} from '../types';
 import {AbiDecoder} from '../utils/abi_decoder';
 import {intervalUtils} from '../utils/interval_utils';
 
-const DEFAULT_MEMPOOL_POLLING_INTERVAL = 200;
+const DEFAULT_EVENT_POLLING_INTERVAL = 200;
 
 export class EventWatcher {
     private _web3Wrapper: Web3Wrapper;
     private _pollingIntervalMs: number;
     private _intervalId: NodeJS.Timer;
-    private _lastMempoolEvents: Web3.LogEntry[] = [];
+    private _lastEvents: Web3.LogEntry[] = [];
     private _callbackAsync?: MempoolEventCallback;
     constructor(web3Wrapper: Web3Wrapper, pollingIntervalMs: undefined|number) {
         this._web3Wrapper = web3Wrapper;
         this._pollingIntervalMs = _.isUndefined(pollingIntervalMs) ?
-                                  DEFAULT_MEMPOOL_POLLING_INTERVAL :
+                                  DEFAULT_EVENT_POLLING_INTERVAL :
                                   pollingIntervalMs;
     }
     public subscribe(callback: MempoolEventCallback, numConfirmations: number): void {
@@ -27,7 +27,7 @@ export class EventWatcher {
     }
     public unsubscribe(): void {
         delete this._callbackAsync;
-        this._lastMempoolEvents = [];
+        this._lastEvents = [];
         intervalUtils.clearAsyncExcludingInterval(this._intervalId);
     }
     private async _pollForMempoolEventsAsync(numConfirmations: number): Promise<void> {
@@ -38,13 +38,13 @@ export class EventWatcher {
             // that's why we just ignore those cases.
             return;
         }
-        const removedEvents = _.differenceBy(this._lastMempoolEvents, pendingEvents, JSON.stringify);
-        const newEvents = _.differenceBy(pendingEvents, this._lastMempoolEvents, JSON.stringify);
+        const removedEvents = _.differenceBy(this._lastEvents, pendingEvents, JSON.stringify);
+        const newEvents = _.differenceBy(pendingEvents, this._lastEvents, JSON.stringify);
         let isRemoved = true;
         await this._emitDifferencesAsync(removedEvents, isRemoved);
         isRemoved = false;
         await this._emitDifferencesAsync(newEvents, isRemoved);
-        this._lastMempoolEvents = pendingEvents;
+        this._lastEvents = pendingEvents;
     }
     private async _getMempoolEventsAsync(numConfirmations: number): Promise<Web3.LogEntry[]> {
         let fromBlock: BlockParamLiteral|number;
--- a/src/order_watcher/event_watcher.ts
+++ b/src/order_watcher/event_watcher.ts
@@ -31,7 +31,7 @@ export class EventWatcher {
         intervalUtils.clearAsyncExcludingInterval(this._intervalId);
     }
     private async _pollForMempoolEventsAsync(numConfirmations: number): Promise<void> {
-        const pendingEvents = await this._getMempoolEventsAsync(numConfirmations);
+        const pendingEvents = await this._getEventsAsync(numConfirmations);
         if (pendingEvents.length === 0) {
             // HACK: Sometimes when node rebuilds the pending block we get back the empty result.
             // We don't want to emit a lot of removal events and bring them back after a couple of miliseconds,
@@ -46,7 +46,7 @@ export class EventWatcher {
         await this._emitDifferencesAsync(newEvents, isRemoved);
         this._lastEvents = pendingEvents;
     }
-    private async _getMempoolEventsAsync(numConfirmations: number): Promise<Web3.LogEntry[]> {
+    private async _getEventsAsync(numConfirmations: number): Promise<Web3.LogEntry[]> {
         let fromBlock: BlockParamLiteral|number;
         let toBlock: BlockParamLiteral|number;
         if (numConfirmations === 0) {
@@ -56,12 +56,12 @@ export class EventWatcher {
             toBlock = await this._web3Wrapper.getBlockNumberAsync();
             fromBlock = toBlock - numConfirmations;
         }
-        const mempoolFilter = {
+        const eventFilter = {
             fromBlock,
             toBlock,
         };
-        const pendingEvents = await this._web3Wrapper.getLogsAsync(mempoolFilter);
-        return pendingEvents;
+        const events = await this._web3Wrapper.getLogsAsync(eventFilter);
+        return events;
     }
     // TODO: Let's emit out own LogEntry type that has property isRemoved rather then removed
     private async _emitDifferencesAsync(logs: Web3.LogEntry[], isRemoved: boolean): Promise<void> {
--- a/src/order_watcher/event_watcher.ts
+++ b/src/order_watcher/event_watcher.ts
@@ -8,6 +8,10 @@ import {assert} from '../utils/assert';
 
 const DEFAULT_EVENT_POLLING_INTERVAL = 200;
 
+/*
+ * The EventWatcher watches for blockchain events at the specified block confirmation
+ * depth.
+ */
 export class EventWatcher {
     private _web3Wrapper: Web3Wrapper;
     private _pollingIntervalMs: number;
--- a/src/order_watcher/event_watcher.ts
+++ b/src/order_watcher/event_watcher.ts
@@ -53,19 +53,16 @@ export class EventWatcher {
         this._lastEvents = pendingEvents;
     }
     private async _getEventsAsync(): Promise<Web3.LogEntry[]> {
-        let fromBlock: BlockParamLiteral|number;
-        let toBlock: BlockParamLiteral|number;
+        let latestBlock: BlockParamLiteral|number;
         if (this._numConfirmations === 0) {
-            fromBlock = BlockParamLiteral.Pending;
-            toBlock = fromBlock;
+            latestBlock = BlockParamLiteral.Pending;
         } else {
             const currentBlock = await this._web3Wrapper.getBlockNumberAsync();
-            toBlock = currentBlock - this._numConfirmations;
-            fromBlock = toBlock;
+            latestBlock = currentBlock - this._numConfirmations;
         }
         const eventFilter = {
-            fromBlock,
-            toBlock,
+            fromBlock: latestBlock,
+            toBlock: latestBlock,
         };
         const events = await this._web3Wrapper.getLogsAsync(eventFilter);
         return events;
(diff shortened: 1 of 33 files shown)


src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+        delete this._orders[orderHash];
+        this._dependentOrderHashes[signedOrder.maker][signedOrder.makerTokenAddress].delete(orderHash);
+        // We currently do not remove the maker/makerToken keys from the mapping when all orderHashes removed
+    }
+    /**
+     * Starts an orderStateWatcher subscription. The callback will be called every time a watched order's
+     * backing blockchain state has changed. This is a call-to-action for the caller to re-validate the order.
+     * @param   callback            Receives the orderHash of the order that should be re-validated, together
+     *                              with all the order-relevant blockchain state needed to re-validate the order.
+     * @param   numConfirmations    Number of confirmed blocks deeps you want to run the orderWatcher from. Passing
```
BMillman19: "deep"

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+     *                              is 0 will watch the backing node's mempool, 3 will emit events when blockchain
```
BMillman19: "in"

src/order_watcher/order_state_watcher.ts
BMillman19: this method doesn't accept numConfirtmations

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+    public subscribe(callback: EventWatcherCallback): void {
+        assert.isFunction('callback', callback);
+        this._callbackIfExistsAsync = callback;
+        this._intervalIdIfExists = intervalUtils.setAsyncExcludingInterval(
+            this._pollForMempoolEventsAsync.bind(this), this._pollingIntervalMs,
+        );
+    }
+    public unsubscribe(): void {
+        delete this._callbackIfExistsAsync;
```
BMillman19: is calling this on undefined ok?

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+     * @param   orderHash     The orderHash of the order you wish to stop watching.
+     */
+    public removeOrder(orderHash: string): void {
+        assert.doesConformToSchema('orderHash', orderHash, schemas.orderHashSchema);
+        const signedOrder = this._orders[orderHash];
+        if (_.isUndefined(signedOrder)) {
+            return; // noop
+        }
+        delete this._orders[orderHash];
+        this._dependentOrderHashes[signedOrder.maker][signedOrder.makerTokenAddress].delete(orderHash);
```
BMillman19: we probably want to remove empty objects here or else they will keep piling up

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+                                blockNumber;
+        const methodOpts = {
+            defaultBlock,
+        };
+
+        for (const orderHash of orderHashes) {
+            const signedOrder = this._orders[orderHash] as SignedOrder;
+            const orderState = await this._orderStateUtils.getOrderStateAsync(signedOrder, methodOpts);
+            if (!_.isUndefined(this._callbackIfExistsAsync)) {
+                await this._callbackIfExistsAsync(orderState);
```
BMillman19: still confused why there is an await here, maybe you can explain to me irl

src/types.ts
```diff
@@ -390,16 +396,27 @@ export interface JSONRPCPayload {
 }
 
 /*
+ * pollingIntervalMs: How often to poll the Ethereum node for new events.
+ * numConfirmations: How many confirmed blocks deep you wish to listen for events at.
+ */
+export interface OrderStateWatcherConfig {
+    pollingIntervalMs?: number;
```
BMillman19: should these also follow the "ifExists" convention

src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+                isValid: false,
+                orderHash,
+                error: err.message,
+            };
+            return orderState;
+        }
+    }
+    public async getOrderRelevantStateAsync(
+        signedOrder: SignedOrder, methodOpts?: MethodOpts): Promise<OrderRelevantState> {
+        const zrxTokenAddress = await this.exchangeWrapper.getZRXTokenAddressAsync();
```
BMillman19: we can parallelize almost all of these

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+                if (isOrderWatched) {
+                    await this._emitRevalidateOrdersAsync([orderHash], blockNumber);
+                }
+                break;
+
+            case ExchangeEvents.LogError:
+                return; // noop
+
+            default:
+                throw utils.spawnSwitchErr('decodedLog.event', decodedLog.event);
```
abandeali1: Why do we throw an error here rather than return noop? I'm assuming most logs we encounter will *not* be relevant to us and will end up hitting the default case.

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+    private async _emitRevalidateOrdersAsync(orderHashes: string[], blockNumber: number): Promise<void> {
+        const defaultBlock = this._numConfirmations === 0 ?
+                                BlockParamLiteral.Pending :
+                                blockNumber;
+        const methodOpts = {
+            defaultBlock,
+        };
+
+        for (const orderHash of orderHashes) {
```
abandeali1: Use `Promise.all` instead of `for...of` with `await`. This actually seems like it could be a big bottleneck.

src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+} from '../types';
+import {ZeroEx} from '../0x';
+import {TokenWrapper} from '../contract_wrappers/token_wrapper';
+import {ExchangeWrapper} from '../contract_wrappers/exchange_wrapper';
+import {utils} from '../utils/utils';
+import {constants} from '../utils/constants';
+
+export class OrderStateUtils {
+    private tokenWrapper: TokenWrapper;
+    private exchangeWrapper: ExchangeWrapper;
```
abandeali1: Add `_` to private variable names for consistency.

src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+                orderHash,
+                error: err.message,
+            };
+            return orderState;
+        }
+    }
+    public async getOrderRelevantStateAsync(
+        signedOrder: SignedOrder, methodOpts?: MethodOpts): Promise<OrderRelevantState> {
+        const zrxTokenAddress = await this.exchangeWrapper.getZRXTokenAddressAsync();
+        const orderHash = ZeroEx.getOrderHashHex(signedOrder);
```
abandeali1: Mildly annoying that we have to calculate the `orderHash` twice, but I can't think of a great way to reuse it here.

src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+            signedOrder.makerTokenAddress, signedOrder.maker, methodOpts,
+        );
+        const makerFeeBalance = await this.tokenWrapper.getBalanceAsync(
+            zrxTokenAddress, signedOrder.maker, methodOpts,
+        );
+        const makerFeeProxyAllowance = await this.tokenWrapper.getProxyAllowanceAsync(
+            zrxTokenAddress, signedOrder.maker, methodOpts,
+        );
+        const filledTakerTokenAmount = await this.exchangeWrapper.getFilledTakerAmountAsync(orderHash, methodOpts);
+        const canceledTakerTokenAmount = await this.exchangeWrapper.getCanceledTakerAmountAsync(orderHash, methodOpts);
```
abandeali1: Can we `await Promise.all` all 6 of these individual function calls?

src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,99 @@
...
+            throw new Error(ExchangeContractErrs.InsufficientMakerBalance);
+        }
+        if (orderRelevantState.makerProxyAllowance.eq(0)) {
+            throw new Error(ExchangeContractErrs.InsufficientMakerAllowance);
+        }
+        if (!signedOrder.makerFee.eq(0)) {
+            if (orderRelevantState.makerFeeBalance.eq(0)) {
+                throw new Error(ExchangeContractErrs.InsufficientMakerFeeBalance);
+            }
+            if (orderRelevantState.makerFeeProxyAllowance.eq(0)) {
```
abandeali1: It seems pretty gameable to only check for 0 balances/allowances, even though technically all orders are partially fillable with any balance/allowance.

Maybe we can let the user define some threshold (probably a percentage)?

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+import * as _ from 'lodash';
+import {Web3Wrapper} from '../web3_wrapper';
+import {BlockParamLiteral, EventCallback, EventWatcherCallback} from '../types';
+import {AbiDecoder} from '../utils/abi_decoder';
+import {intervalUtils} from '../utils/interval_utils';
+import {assert} from '../utils/assert';
+
+const DEFAULT_EVENT_POLLING_INTERVAL = 200;
+
+export class EventWatcher {
```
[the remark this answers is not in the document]
fabioberger: Since this isn't public, I only added a comment to the top of the class.

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+    private _web3Wrapper: Web3Wrapper;
+    private _pollingIntervalMs: number;
+    private _intervalIdIfExists?: NodeJS.Timer;
+    private _lastEvents: Web3.LogEntry[] = [];
+    private _callbackIfExistsAsync?: EventWatcherCallback;
+    private _numConfirmations: number;
+    constructor(web3Wrapper: Web3Wrapper, pollingIntervalMs: undefined|number, numConfirmations: number) {
```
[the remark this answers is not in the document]
fabioberger: It's no longer optional.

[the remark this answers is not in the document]
fabioberger: It will be in a subsequent PR.

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+        this._web3Wrapper = web3Wrapper;
+        this._numConfirmations = numConfirmations;
+        this._pollingIntervalMs = _.isUndefined(pollingIntervalMs) ?
+                                  DEFAULT_EVENT_POLLING_INTERVAL :
+                                  pollingIntervalMs;
+    }
+    public subscribe(callback: EventWatcherCallback): void {
```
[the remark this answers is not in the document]
fabioberger: Let's throw in this case since it would be a developer error to try subscribing twice.

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+        assert.isFunction('callback', callback);
+        this._callbackIfExistsAsync = callback;
```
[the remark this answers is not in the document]
fabioberger: I like this idea, since it also reduces cleanup by not having to manage the instance variable.

src/order_watcher/event_watcher.ts
fabioberger: Yup!

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+    private async _pollForMempoolEventsAsync(): Promise<void> {
+        const pendingEvents = await this._getEventsAsync();
+        if (pendingEvents.length === 0) {
+            // HACK: Sometimes when node rebuilds the pending block we get back the empty result.
+            // We don't want to emit a lot of removal events and bring them back after a couple of miliseconds,
+            // that's why we just ignore those cases.
+            return;
+        }
+        const removedEvents = _.differenceBy(this._lastEvents, pendingEvents, JSON.stringify);
```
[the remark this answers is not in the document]
fabioberger: Hm... yeah, I'm not sure we should be comparing objects by JSON.stringifying them. Doesn't this rely on the ordering of the object properties? @LogvinovLeon

src/types.ts
fabioberger: No since it's a public config variable and the user will only define it if they want it to exist.

src/utils/order_state_utils.ts
fabioberger: This is getting heavily refactored by @LogvinovLeon. He should look into this.

src/order_watcher/event_watcher.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+        const eventFilter = {
+            fromBlock: latestBlock,
+            toBlock: latestBlock,
+        };
+        const events = await this._web3Wrapper.getLogsAsync(eventFilter);
+        return events;
+    }
+    private async _emitDifferencesAsync(logs: Web3.LogEntry[], isRemoved: boolean): Promise<void> {
+        for (const log of logs) {
```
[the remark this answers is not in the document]
fabioberger: @LogvinovLeon does this work with the interface being more similar to a stream?

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+        this._orderStateUtils = orderStateUtils;
+    }
+    /**
+     * Add an order to the orderStateWatcher
+     * @param   signedOrder     The order you wish to start watching.
+     */
+    public addOrder(signedOrder: SignedOrder): void {
+        assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
+        const orderHash = ZeroEx.getOrderHashHex(signedOrder);
+        assert.isValidSignature(orderHash, signedOrder.ecSignature, signedOrder.maker);
```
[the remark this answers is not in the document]
fabioberger: Added a comment explaining that order signature is verified before order is added.

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+     * @param   numConfirmations    Number of confirmed blocks deeps you want to run the orderWatcher from. Passing
+     *                              is 0 will watch the backing node's mempool, 3 will emit events when blockchain
+     *                              state relevant to a watched order changed 3 blocks ago.
+     */
+    public subscribe(callback: OnOrderStateChangeCallback): void {
+        assert.isFunction('callback', callback);
+        if (!_.isUndefined(this._callbackIfExistsAsync)) {
+            throw new Error(ZeroExError.SubscriptionAlreadyPresent);
+        }
+        this._callbackIfExistsAsync = callback;
```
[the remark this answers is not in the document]
fabioberger: We don't, it throws on attempting to do so.

src/order_watcher/order_state_watcher.ts
```diff
@@ -0,0 +1,190 @@
...
+        this._eventWatcher.unsubscribe();
+    }
+    private async _onEventWatcherCallbackAsync(log: LogEvent): Promise<void> {
+        const maybeDecodedLog = this._abiDecoder.tryToDecodeLogOrNoop(log);
+        const isDecodedLog = !_.isUndefined((maybeDecodedLog as LogWithDecodedArgs<any>).event);
+        if (!isDecodedLog) {
+            return; // noop
+        }
+        const blockNumberBuff = ethUtil.toBuffer(maybeDecodedLog.blockNumber);
+        const blockNumber = ethUtil.bufferToInt(blockNumberBuff);
```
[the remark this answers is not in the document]
fabioberger: It is received as a hex encoded string. Added a comment.

src/order_watcher/order_state_watcher.ts
fabioberger: We currently are only able to decode the logs included in this switch statement. Thus, if we cannot decode the log, this code is not run. **If** we add more event types to our decoder, this default block will notify us that we should consider how to handle this new event type.

src/order_watcher/order_state_watcher.ts
fabioberger: @LogvinovLeon thoughts?

src/utils/order_state_utils.ts
fabioberger: This should be done in a separate refactor since this will be a change in convention.

src/utils/order_state_utils.ts
fabioberger: @LogvinovLeon consider this after you're subsequent refactor.

src/utils/order_state_utils.ts
fabioberger: That is the idea! They will need to define a stricter definition within the callback. Our definition of validity in this context is the more relaxed definiton.

PR #206, opened 2017-11-10
Allow timeout for await transaction mined

lukeautry:
This PR:
* Currently awaitTransactionMinedAsync will poll forever
* Allow an optional 'timeoutMs' argument which will reject the promise after n ms

### DIFF
--- a/src/0x.ts
+++ b/src/0x.ts
@@ -289,10 +289,11 @@ export class ZeroEx {
      * Waits for a transaction to be mined and returns the transaction receipt.
      * @param   txHash            Transaction hash
      * @param   pollingIntervalMs How often (in ms) should we check if the transaction is mined.
+     * @param   timeoutMs         How long (in ms) to poll for transaction mined until aborting.
      * @return  Transaction receipt with decoded log args.
      */
     public async awaitTransactionMinedAsync(
-        txHash: string, pollingIntervalMs: number = 1000): Promise<TransactionReceiptWithDecodedLogs> {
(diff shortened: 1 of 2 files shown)


LogvinovLeon:
Makes sense. Will do.

src/0x.ts
```diff
@@ -311,6 +312,16 @@ export class ZeroEx {
                 }
             }, pollingIntervalMs);
         });
+
+        if (timeoutMs) {
+            return Promise.race([
```
LogvinovLeon: As far as I understand - this will reject but will not cancel the polling and we will continue to send requests

lukeautry: That's a good point - adjusted the approach. Instead of using promise race, just set a "timeoutExceeded" bool to check with the interval function. If timeout has been exceeded, clear the interval and reject the promise.

src/0x.ts
```diff
@@ -294,9 +294,19 @@ export class ZeroEx {
...
+        let timeoutExceeded = false;
+        if (timeoutMs) {
+            setTimeout(() => timeoutExceeded = true, timeoutMs);
+        }
+
         const txReceiptPromise = new Promise(
             (resolve: (receipt: TransactionReceiptWithDecodedLogs) => void, reject) => {
             const intervalId = intervalUtils.setAsyncExcludingInterval(async () => {
+                if (timeoutExceeded) {
+                    clearInterval(intervalId);
```
LogvinovLeon: Nit: `intervalUtils.clearAsyncExcludingInterval`

LogvinovLeon:
Published as v0.22.6 https://0xproject.com/docs/0xjs/0.22.6#awaitTransactionMinedAsync

PR #207, opened 2017-11-10
Additional order watcher tests

fabioberger:
This PR:
* Adds additional tests for the order watcher. Most notable tests for the `numConfirmations` config and making sure the watcher doesn't emit events when state changes for unwatched orders.


test/utils/rpc.ts
```diff
@@ -26,6 +26,13 @@ export class RPC {
         const didRevert = await this.sendAsync(payload);
         return didRevert;
     }
+    public async mineBlockAsync(): Promise<void> {
```
LogvinovLeon: `Promise<bool>`

PR #208, opened 2017-11-11
Clean up subscription state.

dekz:
This PR:

Higher level classes which implemented `activeSubscriptions` could result in mismatched state from the base.

In the case of an exception, keep the state correct between contract wrapper, exchange wrapper and token wrapper.

### DIFF
--- a/src/contract_wrappers/contract_wrapper.ts
+++ b/src/contract_wrappers/contract_wrapper.ts
@@ -38,19 +38,10 @@ export class ContractWrapper {
         this._onLogAddedSubscriptionToken = undefined;
         this._onLogRemovedSubscriptionToken = undefined;
     }
-    protected _subscribe<ArgsType extends ContractEventArgs>(
-        address: string, eventName: ContractEvents, indexFilterValues: IndexedFilterValues, abi: Web3.ContractAbi,
-        callback: EventCallback<ArgsType>): string {
-        const filter = filterUtils.getFilter(address, eventName, indexFilterValues, abi);
-        if (_.isUndefined(this._blockAndLogStreamer)) {
-            this._startBlockAndLogStream();
-        }
-        const filterToken = filterUtils.generateUUID();
-        this._filters[filterToken] = filter;
-        this._filterCallbacks[filterToken] = callback;
-        return filterToken;
-    }
-    protected unsubscribeAll(): void {
+    /**
+     * Cancels all existing subscriptions
+     */
+    public unsubscribeAll(): void {
         const filterTokens = _.keys(this._filterCallbacks);
         _.each(filterTokens, filterToken => {
             this._unsubscribe(filterToken);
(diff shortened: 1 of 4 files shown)


src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -825,13 +827,7 @@ export class ExchangeWrapper extends ContractWrapper {
...
         return ZRXtokenAddress;
     }
-    /**
-     * Cancels all existing subscriptions
-     */
-    public unsubscribeAll(): void {
-        _.forEach(this._activeSubscriptions, this._unsubscribe.bind(this));
-        this._activeSubscriptions = [];
-    }
+
```
fabioberger: Nit: remove this new line.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+    ZeroExError,
+    Token,
+    SubscriptionOpts,
+    TokenEvents,
+    ContractEvent,
+    TransferContractEventArgs,
+    ApprovalContractEventArgs,
+    TokenContractEventArgs,
+    LogWithDecodedArgs,
+    LogEvent,
```
fabioberger: Let's remove any type imports we aren't using in this file.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+        coinbase = userAddresses[0];
+        addressWithoutFunds = userAddresses[1];
+    });
+    beforeEach(async () => {
+        await blockchainLifecycle.startAsync();
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+
```
fabioberger: Nit: remove new line.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+            const token = tokens[0];
+            tokenAddress = token.address;
+        });
+        afterEach(() => {
+            zeroEx.token.unsubscribeAll();
+        });
+        it('Should receive the Error when an error occurs', (done: DoneCallback) => {
+            (async () => {
+                const callback = (err: Error, logEvent: LogEvent<ApprovalContractEventArgs>) => {
+                    expect(err).to.not.be.undefined();
```
fabioberger: We should check that the error is not `null` here.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+                    expect(logEvent).to.be.undefined();
+                    done();
+                };
+                Sinon.stub((zeroEx as any)._web3Wrapper, 'getBlockAsync')
```
fabioberger: We need to unstub this method.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+                 .throws("JSON RPC error")
```
fabioberger: When putting subsequent function calls on new lines, make sure to indent it.

fabioberger: This is also missing a semi-colon. Our convention is to always add semi-colons.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+                zeroEx.token.subscribe(
+                    tokenAddress, TokenEvents.Approval, indexFilterValues, callback);
+                await zeroEx.token.setAllowanceAsync(tokenAddress, coinbase, addressWithoutFunds, allowanceAmount);
+            })().catch(done);
+         });
+        it('Should allow unsubscribeAll to be called multiple times', (done: DoneCallback) => {
+            (async () => {
+                const callback = (err: Error, logEvent: LogEvent<ApprovalContractEventArgs>) => { };
```
fabioberger: Let's use `_.noop` rather then an empty function body for clarity.

test/subscription_test.ts
```diff
@@ -0,0 +1,93 @@
...
+                zeroEx.token.subscribe(
+                    tokenAddress, TokenEvents.Approval, indexFilterValues, callback);
+                await zeroEx.token.setAllowanceAsync(tokenAddress, coinbase, addressWithoutFunds, allowanceAmount);
```
fabioberger: Is this line necessary for this test?

src/contract_wrappers/contract_wrapper.ts
```diff
@@ -50,6 +50,12 @@ export class ContractWrapper {
         this._filterCallbacks[filterToken] = callback;
         return filterToken;
     }
+    protected unsubscribeAll(): void {
```
BMillman19: lets just make this public and then remove the overrides from the subclasses

fabioberger: Good catch!

PR #219, opened 2017-11-13
Rounding validation

LogvinovLeon:
This PR:
* Adds a validations that the order values are within the range without rounding errors

### DIFF
--- a/packages/0x.js/src/utils/order_state_utils.ts
+++ b/packages/0x.js/src/utils/order_state_utils.ts
@@ -117,6 +117,13 @@ export class OrderStateUtils {
                 throw new Error(ExchangeContractErrs.InsufficientMakerFeeAllowance);
             }
         }
+        const minimumFillableTakerTokenAmountWithingNoRoundingErrorRange = signedOrder.takerTokenAmount
+                                                                           .times(1000)
+                                                                           .dividedBy(signedOrder.makerTokenAmount);
+        if (orderRelevantState.remainingFillableTakerTokenAmount
+            .lessThan(minimumFillableTakerTokenAmountWithingNoRoundingErrorRange)) {
+            throw new Error(ExchangeContractErrs.OrderFillRoundingError);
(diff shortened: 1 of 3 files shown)


packages/0x.js/src/utils/order_state_utils.ts
```diff
@@ -113,6 +117,13 @@ export class OrderStateUtils {
                 throw new Error(ExchangeContractErrs.InsufficientMakerFeeAllowance);
             }
         }
+        const minimumFillableTakerTokenAmountWithingNoRoundingErrorRange = signedOrder.takerTokenAmount
```
fabioberger: `minFillableTakerTokenAmountWithinNoRoundingErrorRange`

packages/0x.js/src/utils/order_state_utils.ts
```diff
@@ -113,6 +117,13 @@ export class OrderStateUtils {
...
+                                                                           .times(1000)
```
fabioberger: Why 1000? Can you set to constant with a descriptive name?

packages/0x.js/src/utils/order_state_utils.ts
```diff
@@ -113,6 +117,13 @@ export class OrderStateUtils {
...
+                                                                           .dividedBy(signedOrder.makerTokenAmount);
+        if (orderRelevantState.remainingFillableTakerTokenAmount
+            .lessThan(minimumFillableTakerTokenAmountWithingNoRoundingErrorRange)) {
+            throw new Error(ExchangeContractErrs.OrderFillRoundingError);
```
fabioberger: I think we need a more specific error message name here. Perhaps: `CannotFillOrderWithoutCausingRoundingError`?

LogvinovLeon: Should it include some info about intervals?
Cause the above is not true. You can.

PR #222, opened 2017-11-14
Remove support for Async callback types when used in Subscribe functions

dekz:
This PR:
* Removes in: Exchange subscribe, Token subscribe and OrderWatcher subscribe.


packages/0x.js/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -52,8 +52,7 @@ export class ContractWrapper {
             throw new Error(ZeroExError.SubscriptionNotFound);
         }
         if (!_.isUndefined(err)) {
-            const callback = this._filterCallbacks[filterToken];
-            callback(err, undefined);
+            this._filterCallbacks[filterToken](err, undefined);
```
fabioberger: I actually prefer us to first set it to a variable called `callback`, for readability.

packages/0x.js/src/index.ts
```diff
@@ -6,8 +6,6 @@ export {
     ECSignature,
     ZeroExError,
     EventCallback,
-    EventCallbackAsync,
-    EventCallbackSync,
```
fabioberger: Can you also open a PR in the website repo, removing these from the `public0xjsTypes` array in `utils/constants.ts`.

packages/0x.js/CHANGELOG.md
```diff
@@ -1,13 +1,17 @@
 # CHANGELOG
 
+v0.25.0 - _November 14, 2017_
```
fabioberger: Let's just leave this as `vx.x.x`

fabioberger:
💯

packages/0x.js/src/index.ts
LogvinovLeon: No. We shouldn't remove them cause they're still present in the old versions

PR #224, opened 2017-11-15
Add connect to monorepo

BMillman19:
This PR:
* Add connect into the monorepo


packages/connect/test/ws_orderbook_channel_test.ts
```diff
@@ -0,0 +1,45 @@
...
+        baseTokenAddress: '0x323b5d4c32345ced77393b3530b1eed0f346429d',
+        quoteTokenAddress: '0xef7fff64389b814a946f3e92105513705ca6b990',
+        snapshot: true,
+        limit: 100,
+    };
+    const emptyOrderbookChannelHandler = {
+        onSnapshot: () => { return; },
+        onUpdate: () => { return; },
+        onError: () => { return; },
+        onClose: () => { return; },
```
fabioberger: You could use `_.noop` instead of `{ return; }` here but nbd :)

PR #228, opened 2017-11-17
Add postpublish scripts for sub-packages

fabioberger:
This PR:
* Adds a postpublish script for each sub-package. For most, it simply adds a release entry on Github's releases page. For 0x.js, it also generates and uploads the docs JSON.
* Adds a lerna:publish command which cleans and builds the packages before calling `lerna publish`.


packages/0x.js/package.json
```diff
@@ -92,14 +87,12 @@
     "bignumber.js": "~4.1.0",
     "bn.js": "4.11.8",
     "compare-versions": "^3.0.1",
-    "es6-promisify": "^5.0.0",
```
BMillman19: don't we still rely on `es6-promisify` in this package?

packages/0x.js/scripts/postpublish.js
```diff
@@ -0,0 +1,43 @@
...
+const subPackageName = packageJSON.name;
+
+let tag;
+let version;
+postpublish_utils.getLatestTagAndVersionAsync(subPackageName)
+    .then(function(result) {
+        console.log('POSTPUBLISH: Releasing...');
+        tag = result.tag;
+        version = result.version;
+         const releaseName = subPackageName + ' v' + version;
```
BMillman19: we should put this release name formatting into a postpublish_utils function since its repeated across all the scripts

packages/0x.js/scripts/postpublish.js
```diff
@@ -0,0 +1,43 @@
...
+const packageJSON = require('../package.json');
+
+const cwd = __dirname + '/..';
+const subPackageName = packageJSON.name;
+
+let tag;
+let version;
+postpublish_utils.getLatestTagAndVersionAsync(subPackageName)
+    .then(function(result) {
+        console.log('POSTPUBLISH: Releasing...');
```
BMillman19: do you still want to keep the console.log statements in here?

scripts/postpublish_utils.js
```diff
@@ -0,0 +1,47 @@
...
+                const latestVersion = sortedVersions[0];
+                const latestTag = subPackagePrefix + latestVersion;
+                return {
+                    tag: latestTag,
+                    version: latestVersion
+                };
+            });
+    },
+    publishReleaseNotes: function(tag, releaseName, assets) {
+        console.log('POSTPUBLISH: Releasing ', releaseName, '...');
```
BMillman19: same here

packages/0x.js/package.json
fabioberger: We do. Good catch.

packages/0x.js/scripts/postpublish.js
fabioberger: Not this one, but the others yes. I think it's helpful to know where something went wrong.

packages/tslint-config/scripts/postpublish.js
```diff
@@ -0,0 +1,14 @@
+const postpublish_utils = require('../../../scripts/postpublish_utils');
```
LogvinovLeon: Can we have this script in one place and symlink to it?

fabioberger: We could... my hunch thought is that these will eventually diverge for different sub-packages.

PR #233, opened 2017-11-23
Forces the users of 0x.js to pass the network id

LogvinovLeon:
This PR:
* Forces the users of 0x.js to pass the networkId which makes some code simpler and non-async
* Adds some more linter rules
* Fixes the linter run command to lint all files and fixes the issues
* Adds artifacts for ZRX token and removes redundant parts of artifacts
* Adds `--project` config to tslint which allows to use type-dependent rules (like the ones that promises should be awaited.

Most of the changes in this PR are either import changes, because we added a rule to sort them and applied the auto-fix or artifacts clean-up


packages/0x.js/src/utils/order_state_utils.ts
```diff
@@ -1,28 +1,62 @@
...
+        }
+        const minFillableTakerTokenAmountWithinNoRoundingErrorRange = signedOrder.takerTokenAmount
+                                                                      .dividedBy(ACCEPTABLE_RELATIVE_ROUNDING_ERROR)
+                                                                      .dividedBy(signedOrder.makerTokenAmount);
+        if (orderRelevantState.remainingFillableTakerTokenAmount
+            .lessThan(minFillableTakerTokenAmountWithinNoRoundingErrorRange)) {
+            throw new Error(ExchangeContractErrs.OrderFillRoundingError);
+        }
+        // TODO Add linear function solver when maker token is ZRX #badass
+        // Return the max amount that's fillable
```
fabioberger: Remove outdated comment.

packages/assert/test/assert_test.ts
```diff
@@ -183,7 +184,7 @@ describe('Assertions', () => {
         it('should not throw for valid input', () => {
             const validInputs = [
                 42,
-                0.00,
+                0.00, // tslint:disable-line:number-literal-format
```
fabioberger: Let's move this to `0`

packages/0x.js/src/web3_wrapper.ts
```diff
@@ -18,17 +19,19 @@ interface RawLogEntry {
...
-    constructor(provider: Web3.Provider, defaults?: Partial<Web3.TxData>) {
+    constructor(provider: Web3.Provider, networkId: number, defaults?: Partial<Web3.TxData>) {
         if (_.isUndefined((provider as any).sendAsync)) {
             // Web3@1.0 provider doesn't support synchronous http requests,
             // so it only has an async `send` method, instead of a `send` and `sendAsync` in web3@0.x.x`
             // We re-assign the send method so that Web3@1.0 providers work with 0x.js
             (provider as any).sendAsync = (provider as any).send;
         }
         this.web3 = new Web3();
+        this.networkId = networkId;
```
fabioberger: What if someone updates the provider to one connected to a different network?

PR #234, opened 2017-11-23
Make DecodedLogEvent contain web3 log under a log subkey

LogvinovLeon:
This PR:
* Separates logs from log events
* Now log events contain log as a subkey which will simplify consumption of this type (inspired by OTC)

### DIFF
--- a/packages/0x.js/src/types.ts
+++ b/packages/0x.js/src/types.ts
@@ -40,7 +40,10 @@ export type OrderValues = [BigNumber, BigNumber, BigNumber,
                            BigNumber, BigNumber, BigNumber];
 
 export type LogEvent = Web3.LogEntryEvent;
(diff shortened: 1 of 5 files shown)


packages/0x.js/src/types.ts
```diff
@@ -40,7 +40,10 @@ export type OrderValues = [BigNumber, BigNumber, BigNumber,
                            BigNumber, BigNumber, BigNumber];
 
 export type LogEvent = Web3.LogEntryEvent;
-export type DecodedLogEvent<ArgsType> = Web3.DecodedLogEntryEvent<ArgsType>;
+export interface DecodedLogEvent<ArgsType> {
+    removed: boolean;
```
fabioberger: Now that this is our own type, can we re-name this to `isRemoved`?

PR #236, opened 2017-11-23
Fix/validate orders against latest block

fabioberger:
This PR:
* Makes it such that order validation happens against the `latest` block, not the `pending` block. This fixes a bug for users of Infura, which does not have reliable support for querying state for the `pending` block.


LogvinovLeon:
LGTM as soon as tests pass

PR #243, opened 2017-11-28
Refactor docs

fabioberger:
This PR:
* Refactors the docs to use a declarative structure where all the hard-coded information related to a package's documentation is in a single place and all sub-components render themselves accordingly.
* Prepared the docs to support any arbitrary number of additional docs for new packages.
* Combined the top-level react component for 0x.js and Smart contract docs into a single component. This removed a lot of duplicate code and forced the data to be extracted from the rendering layer.


packages/website/ts/index.tsx
```diff
@@ -78,16 +78,25 @@ const LazyPortal = createLazyComponent(
...
 );
 const LazySmartContractsDocumentation = createLazyComponent(
-    'SmartContractsDocumentation',
+    'Documentation',
     async () => System.import<any>(
         /* webpackChunkName: "smartContractDocs" */'ts/containers/smart_contracts_documentation',
     ),
 );
 
+const docs = class Documentation extends
```
LogvinovLeon: Is this absolutely necessary?

packages/website/ts/components/top_bar.tsx
```diff
@@ -272,25 +274,21 @@ export class TopBar extends React.Component<TopBarProps, TopBarState> {
         );
     }
     private renderDocsMenu() {
-        if (!this.isViewing0xjsDocs() && !this.isViewingSmartContractsDocs()) {
+        if (!this.isViewing0xjsDocs() && !this.isViewingSmartContractsDocs() || _.isUndefined(this.props.menu)) {
```
BMillman19: do we still need the first part of this check if we now have this.props.menu ?

packages/website/ts/pages/documentation/event_definition.tsx
```diff
@@ -53,9 +55,15 @@ export class EventDefinition extends React.Component<EventDefinitionProps, Event
     private renderEventCode() {
         const indexed = <span style={{color: CUSTOM_GREEN}}> indexed</span>;
         const eventArgs = _.map(this.props.event.eventArgs, (eventArg: EventArg) => {
+            const t = (
```
BMillman19: can we have a more descriptive variable name here?

packages/website/ts/pages/documentation/method_signature.tsx
```diff
@@ -16,45 +18,64 @@ const defaultProps = {
...
 };
 
-function renderParameters(method: TypescriptMethod|SolidityMethod, typeDefinitionByName?: TypeDefinitionByName) {
+function renderParameters(
+    method: TypescriptMethod|SolidityMethod, docsInfo: DocsInfo, typeDefinitionByName?: TypeDefinitionByName,
+) {
     const parameters = method.parameters;
     const params = _.map(parameters, (p: Parameter) => {
         const isOptional = p.isOptional;
+        const t = (
```
BMillman19: same here

packages/website/ts/components/top_bar.tsx
fabioberger: I think it's a stricter check. If is possible we have a future page with a menu passed in that is not a docs page.

PR #244, opened 2017-11-28
Add Connect Docs

fabioberger:
This PR:
* Adds the docs for 0x connect
* Refactors TopBar to also include 0x Connect
* Refactors SourceLink so that it doesn't hard-code for `0x.js`


packages/website/ts/containers/connect_documentation.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+const connectDocSections = {
+    introduction: 'introduction',
+    installation: 'installation',
+    httpClient: 'httpClient',
+    types: 'types',
+};
+
+const docsInfoConfig: DocsInfoConfig = {
+    packageName: '0x Connect',
```
BMillman19: is this supposed to be 0x.js?

packages/website/package.json
```diff
@@ -18,7 +18,7 @@
   "author": "Fabio Berger",
   "license": "Apache-2.0",
   "dependencies": {
-    "0x.js": "^0.27.1",
+    "0x.js": "0xproject/0x.js/packages/0x.js#0x.js@0.27.0",
```
LogvinovLeon: 0.27.1

packages/website/ts/containers/connect_documentation.tsx
fabioberger: Let me work on the naming here since both you and Leo got confused about this.

PR #250, opened 2017-12-04
Compare versions script

LogvinovLeon:
This PR:
* Makes all packages use the same versions of shared deps
* Adds a script to check that

Before:
screen shot 2017-12-02 at 17 31 03
After:
screen shot 2017-12-04 at 19 57 33


packages/monorepo-scripts/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@0xproject/monorepo-scripts",
+  "version": "0.0.0",
```
fabioberger: Let's add the `private` key here since we don't want to publish this to NPM.

packages/monorepo-scripts/src/deps_versions.ts
```diff
@@ -0,0 +1,59 @@
...
+    [depName: string]: Versions;
+}
+
+function log(...args: any[]) {
+    console.log(...args); // tslint:disable-line:no-console
+}
+
+function getDependencies(path: string): Dependencies {
+    const file = fs.readFileSync(path).toString();
+    const parsed = JSON.parse(file);
```
fabioberger: This could throw, let's catch it and throw a more helpful error msg.

packages/monorepo-scripts/src/deps_versions.ts
```diff
@@ -0,0 +1,59 @@
...
+    const dependencies = {
+        ...parsed.dependencies,
+        ...parsed.devDependencies,
+    };
+    return dependencies;
+}
+
+function getPackageName(path: string): string {
```
fabioberger: Let's not make this a stand alone function since it is only used in one place and doesn't generalize well.

packages/monorepo-scripts/src/deps_versions.ts
```diff
@@ -0,0 +1,59 @@
...
+    const [dotDot, packageName, packageJSON] = path.split('/');
+    return packageName;
+}
+
+const files = globSync('../*/package.json');
```
fabioberger: Nit: Set `'../*/package.json'` to constant at top of the file.

packages/monorepo-scripts/src/deps_versions.ts
LogvinovLeon: Yeah... Although it can only throw if we have a malformed package.json.
And it will throw smth like: 'Not a valid json line:column'
What can be better than that? NPM throws the same message.

fabioberger: Ok, I guess that is fine.

PR #252, opened 2017-12-06
Add Subproviders Subpackage

fabioberger:
This PR:
* Creates a new `Subproviders` package with the following subproviders:
    * LedgerSubprovider
    * RedundantRPCSubprovider
    * InjectedWeb3Subprovider
* Adds unit and integration tests for the LedgerSubprovider and unit tests for RedundantRPCSubprovider.
* Use these moved/refactored subproviders in Portal

Open question: We currently also build this package for UMD & CommonJS. Should we only build for CommonJS?


packages/website/package.json
```diff
@@ -18,6 +18,7 @@
   "author": "Fabio Berger",
   "license": "Apache-2.0",
   "dependencies": {
+    "@0xproject/subproviders": "0.0.0",
```
LogvinovLeon: ^0.0.1

packages/website/ts/globals.d.ts
```diff
@@ -157,3 +163,5 @@ declare interface Artifact {
         };
     };
 }
+
+// tslint:enable:max-classes-per-file
```
LogvinovLeon: No need

packages/website/ts/globals.d.ts
```diff
@@ -132,21 +133,26 @@ declare class Subprovider {}
...
+    }
     export = RpcSubprovider;
 }
-// tslint:disable-next-line:max-classes-per-file
-declare class HookedWalletSubprovider {
-    constructor(wallet: any);
-}
-declare module 'web3-provider-engine/subproviders/hooked-wallet' {
-    export = HookedWalletSubprovider;
+declare module 'web3-provider-engine' {
```
LogvinovLeon: We have it in at least two places in code. Maybe it's time for a separate typings package?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -0,0 +1,320 @@
...
+        }
+        const isHexPrefixed = data.slice(0, 2) === '0x';
+        if (!isHexPrefixed) {
+            return false;
+        }
+        const nonPrefixed = data.slice(2);
+        const isValid = nonPrefixed.match(HEX_REGEX);
+        return isValid;
+    }
+    private static validatePersonalMessage(msgParams: PartialTxParams) {
```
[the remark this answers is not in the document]
fabioberger: Maybe in a later iteration.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -0,0 +1,320 @@
...
+            default:
+                next();
+                return;
+        }
+    }
+    public async getAccountsAsync(): Promise<string[]> {
+        this._ledgerClientIfExists = await this.createLedgerClientAsync();
+
+        const accounts = [];
+        for (let i = 0; i < NUM_ADDRESSES_TO_FETCH; i++) {
```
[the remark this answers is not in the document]
fabioberger: Looks like there is no programmatic way to retrieve the extended public key from a Ledger Nano S of yet.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -0,0 +1,320 @@
...
+            const result = await this.emitPayloadAsync(payload);
+            this._nonceLock.signal();
+            return result;
+        } catch (err) {
+            this._nonceLock.signal();
+            throw err;
+        }
+    }
+    private async signTransactionWithoutSendingAsync(txParams: PartialTxParams): Promise<ResponseWithTxParams> {
+        await this._nonceLock.wait();
```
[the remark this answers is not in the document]
fabioberger: How so?

packages/subproviders/src/types.ts
```diff
@@ -0,0 +1,115 @@
...
+    gasPrice?: string;
+    gas: string;
+    to: string;
+    from?: string;
+    value?: string;
+    data?: string;
+    chainId: number; // EIP 155 chainId - mainnet: 1, ropsten: 3
+}
+
+export type DoneCallback = (err?: Error) => void;
```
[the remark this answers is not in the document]
fabioberger: It is the way Mocha expects their done callback. Mocha does not follow JS conventions here.

packages/website/ts/globals.d.ts
fabioberger: Will move after your PR get merged ;)

packages/subproviders/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@0xproject/subproviders",
+  "version": "0.0.1",
+  "main": "lib/src/index.js",
+  "types": "lib/src/index.d.ts",
+  "license": "Apache-2.0",
+  "scripts": {
+    "prebuild": "npm run clean",
```
BMillman19: does prebuild get called with `lerna run build` ?

packages/subproviders/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "clean": "shx rm -rf lib",
+    "build": "tsc",
+    "lint": "tslint --project . 'src/**/*.ts' 'test/**/*.ts'",
+    "run_mocha_unit": "mocha lib/test/unit/**/*_test.js --timeout 10000 --bail --exit",
+    "run_mocha_integration": "mocha lib/test/integration/**/*_test.js --timeout 10000 --bail --exit",
+    "test": "npm run test:unit",
+    "test:all": "run-s test:unit test:integration",
+    "test:unit": "run-s clean build run_mocha_unit",
+    "test:integration": "run-s clean build run_mocha_integration"
+  },
```
BMillman19: we need a test:circleci as well

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -0,0 +1,305 @@
...
+        switch (payload.method) {
+            case 'eth_coinbase':
+                try {
+                    accounts = await this.getAccountsAsync();
+                    end(null, accounts[0]);
+                } catch (err) {
+                    end(err);
+                }
+                return;
+
```
BMillman19: should we get rid of all the newlines?

packages/subproviders/src/subproviders/subprovider.ts
```diff
@@ -0,0 +1,45 @@
...
+} from '../types';
+/*
+ * A version of the base class Subprovider found in providerEngine
+ * This one has an async/await `emitPayloadAsync` and also defined types.
+ * Altered version of: https://github.com/MetaMask/provider-engine/blob/master/subproviders/subprovider.js
+ */
+export class Subprovider {
+    private engine: any;
+    private currentBlock: any;
+    private static getRandomId() {
```
BMillman19: some comments here about the approach of producing random Ids might be good

packages/subproviders/package.json
fabioberger: No, removed this. It's legacy from Dekz's repo.

packages/subproviders/package.json
fabioberger: Good catch

packages/subproviders/src/subproviders/ledger.ts
fabioberger: No, our convention is to keep new lines between switch cases **if** the cases are not wrapped in curly-brackets. I'm happy to revise this, but I feel it better elucidates the separation between the cases.

packages/subproviders/src/subproviders/ledger.ts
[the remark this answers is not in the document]
LogvinovLeon: Are you sure? https://github.com/LedgerHQ/ledger-node-js-api/blob/e7b12b1911f265c3e782bc5bf6d890aacfe74b2a/src/ledger-eth.js#L27

packages/subproviders/src/subproviders/ledger.ts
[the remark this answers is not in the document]
LogvinovLeon: We can lock inside of `populateMissingTxParamsAsync`

PR #253, opened 2017-12-06
Refactor web3Wrapper to a separate package

LogvinovLeon:
This PR:
* Is a lot of refactoring. No new logic
* Refactors web3Wrapper out
* Refactors promisify to utils package and removes dependency on es6-promisify.
* This made it more type-safe but generic.
* We still have a web3Wrapper in website cause it's different in a lot of ways.


packages/contracts/deploy/cli.ts
```diff
@@ -1,3 +1,5 @@
+import {TxData} from '@0xproject/types';
```
fabioberger: Where is this being used?

packages/contracts/deploy/src/utils/network.ts
```diff
@@ -1,15 +1,14 @@
-import promisify = require('es6-promisify');
+import {promisify} from '@0xproject/utils';
+import {Web3Wrapper} from '@0xproject/web3-wrapper';
```
fabioberger: Wait, first of all, we aren't using this here. Second, why do we need this module at all? Can't we just call `web3Wrapper`'s `getNetworkIdAsync` instead?

packages/web3-wrapper/src/index.ts
```diff
@@ -155,13 +130,17 @@ export class Web3Wrapper {
...
-        const web3ContractInstance = this.web3.eth.contract(abi).at(address);
+    public getContractFromAbi(abi: Web3.ContractAbi): Web3.Contract<any> {
+        const web3Contract = this.web3.eth.contract(abi);
+        return web3Contract;
+    }
+    public getContractInstance(abi: Web3.ContractAbi, address: string): Web3.ContractInstance {
+        const web3ContractInstance = this.getContractFromAbi(abi).at(address);
         return web3ContractInstance;
     }
-    private async getNetworkAsync(): Promise<number> {
```
fabioberger: Why did we get rid of this method? It's still a useful `web3` call.

packages/types/README.md
```diff
@@ -0,0 +1,10 @@
+Web3 wrapper
+------
+
+Wrapped version of web3 with nicer interface to be used across 0x projects and packages
```
fabioberger: Kek... This is not a README for `types`...

packages/utils/README.md
```diff
@@ -0,0 +1,10 @@
+utils
+------
+
+Utils to be shared across 0x projects and packages
+
+## Install
+
+```bash
+yarn add @0xproject/utils
+```
```
fabioberger: There are **many** more util methods shared between our packages, such as `consoleLog` and `spawnSwitchErr`. Can you please go through all the util methods in `0x.js`, `website` and `contracts` and extract all duplicate methods into this package?

fabioberger: This can be done in a subsequent PR.

packages/web3-wrapper/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@types/lodash": "^4.14.86",
+    "npm-run-all": "^4.1.2",
+    "shx": "^0.2.2",
+    "tslint": "5.8.0",
+    "typescript": "~2.6.1",
+    "web3-typescript-typings": "^0.7.2"
+  },
+  "dependencies": {
+    "@0xproject/utils": "^0.0.1",
+    "@0xproject/types": "^0.0.1",
```
fabioberger: Shouldn't this be a `devDep`?

packages/web3-wrapper/src/index.ts
```diff
@@ -74,37 +64,22 @@ export class Web3Wrapper {
...
         const balanceWei = this.web3.toWei(ethAmount, 'ether');
         return balanceWei;
     }
     public async getBalanceInWeiAsync(owner: string): Promise<BigNumber> {
         let balanceInWei = await promisify<BigNumber>(this.web3.eth.getBalance)(owner);
+        // Rewrap in a new BigNumber
         balanceInWei = new BigNumber(balanceInWei);
         return balanceInWei;
     }
+    public async getBalanceInEthAsync(owner: string): Promise<BigNumber> {
```
fabioberger: Um, this was probably taken from `website`. It should be removed. I'm not a fan of this method and feel it should be refactored to return wei, not ETH.

packages/web3-wrapper/src/index.ts
```diff
@@ -155,13 +130,17 @@ export class Web3Wrapper {
...
+    public getContractInstance(abi: Web3.ContractAbi, address: string): Web3.ContractInstance {
+        const web3ContractInstance = this.getContractFromAbi(abi).at(address);
         return web3ContractInstance;
     }
-    private async getNetworkAsync(): Promise<number> {
-        const networkId = await promisify<number>(this.web3.version.getNetwork)();
-        return networkId;
+    public async estimateGasAsync(data: string): Promise<number> {
+        const gas = await promisify<number>(this.web3.eth.estimateGas)({data});
+        return gas;
```
fabioberger: We should re-wrap `gas` in a new BigNumber.js before returning.

packages/website/ts/blockchain.ts
```diff
@@ -65,7 +65,7 @@ export class Blockchain {
     public nodeVersion: string;
     private zeroEx: ZeroEx;
     private dispatcher: Dispatcher;
-    private web3Wrapper: Web3Wrapper;
+    private web3Wrapper?: Web3Wrapper;
```
fabioberger: Um, why is this now optional?

packages/contracts/deploy/src/utils/network.ts
LogvinovLeon: current web3Wrapper assumes that networkId is already passed from outside.

packages/web3-wrapper/src/index.ts
LogvinovLeon: We don't use it now. All users pass networkId from outside.

packages/website/ts/blockchain.ts
LogvinovLeon: It always was. Do we set it in the constructor? No. Then it's optional. It's better to have it as a type so that if we ever change the code - we're more carefull

packages/web3-wrapper/src/index.ts
LogvinovLeon: Rewrap a number and make it a better number?

packages/web3-wrapper/src/index.ts
fabioberger: What about in `contracts`?

packages/web3-wrapper/src/index.ts
fabioberger: ah nvm

packages/contracts/deploy/src/utils/network.ts
fabioberger: Maybe `networkId` should be optionally passed in. It isn't actually used in the constructor... And the method for getting the networkId from web3 should still be there.

fabioberger: Happy to discuss this over Appear.in.

.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - run: wget https://s3.amazonaws.com/testrpc-shapshots/${CONTRACTS_COMMIT_HASH}.zip
       - run: unzip ${CONTRACTS_COMMIT_HASH}.zip -d testrpc_snapshot
       - run: node ./node_modules/lerna/bin/lerna.js bootstrap
-      - run: yarn lerna:run bootstrap
```
BMillman19: I agree that bootstrap should be removed but most if not all packages build as a part of their test:circleci commands

packages/contracts/deploy/src/utils/network.ts
```diff
@@ -1,15 +1,14 @@
...
     async getNetworkIdIfExistsAsync(port: number): Promise<number> {
         const url = `http://localhost:${port}`;
         const web3Provider = new Web3.providers.HttpProvider(url);
-        const defaults = {};
-        const web3Wrapper = new Web3Wrapper(web3Provider, defaults);
-        const networkIdIfExists = await web3Wrapper.getNetworkIdIfExistsAsync();
-        return networkIdIfExists;
+        const web3Wrapper = new Web3Wrapper(web3Provider);
+        const networkId = web3Wrapper.getNetworkIdAsync();
+        return networkId;
```
fabioberger: Let's get rid of this file completely. It's only called in two places, so let's refactor those to call `web3Wrapper` directly.

PR #255, opened 2017-12-07
Update WETH

abandeali1:
This PR:
* Updates EtherToken and its dependencies (see contracts with _v2 suffix)
* Changes the file structure of the contracts. I'd like feedback here, but this way of versioning allows for the most code reuse. Versions will be bumped of the contracts change after being deployed on mainnet. Anything that is completely unused can be removed or deprecated.
* Adds tests for `ERC20Token` (different from the old `StandardToken` in that it throws on insufficient balance/allowance) and `EtherToken_v2` (adds deposit/withdraw events and uses `ERC20Token`.
* Updates 0x.js and Truffle to latest versions


packages/contracts/contracts/test/Mintable_v2.sol
```diff
@@ -0,0 +1,18 @@
...
+
+/*
+ * Mintable
+ * Base contract that creates a mintable UnlimitedAllowanceToken
+ */
+contract Mintable_v2 is ERC20Token, SafeMath_v2 {
+    function mint(uint _value) 
+        public
+    {
+        require(_value <= 100000000000000000000);
```
LogvinovLeon: What's this value? Can we name it?

packages/contracts/contracts/tokens/ERC20Token.sol
```diff
@@ -0,0 +1,68 @@
+pragma solidity ^0.4.18;
```
LogvinovLeon: Why did we get rid of UnlimitedAllowance contract?

packages/contracts/contracts/tokens/ERC20Token.sol
```diff
@@ -0,0 +1,68 @@
...
+        view
+        returns (uint)
+    {
+        return allowed[_owner][_spender];
+    }
+
+    mapping (address => uint) balances;
+    mapping (address => mapping (address => uint)) allowed;
+    uint public totalSupply;
+}
```
LogvinovLeon: This will not lint ;)

packages/contracts/contracts/tokens/Token.sol
```diff
@@ -35,4 +35,4 @@ contract Token {
 
     event Transfer(address indexed _from, address indexed _to, uint _value);
     event Approval(address indexed _owner, address indexed _spender, uint _value);
-}
+}
```
LogvinovLeon: This either. Please configure your text editor to add newlines.

fabioberger:
From a high level, I think we should consider re-structuring the `contracts` package such that every contract is essentially it's own sub-package, and is associated with a [ethpm](https://github.com/ethpm/ethpm-spec) `package.json` detailing it's contract dependencies and the specific versions of each that it depends on. 

I'm not a fan of the `_v2` suffix with code duplication, instead, we should be "publishing" new versions of the contracts by hosting them on IPFS, and updating the dependencies on the select contracts we want them to change for. This way, we only ever have one version of each contract in the repo (the latest).

If someone wants to see an earlier version of the contract, they should be able to either pull it from IPFS **or** checkout a specific commit at which the contract was at the desired "version".

Thoughts?

abandeali1:
@fabioberger, I agree that using something like ethpm would be ideal. I'll look into this further and see what's possible.

packages/contracts/contracts/utils/SafeMath_v2.sol
```diff
@@ -0,0 +1,74 @@
...
+
+    function min256(uint256 a, uint256 b)
+        internal
+        pure
+        returns (uint256)
+    {
+        return a < b ? a : b;
+    }
+}
+
```
fabioberger: What changed in `SafeMath`?

fabioberger: Just another con to the suffix approach, you don't get clear diffs for contract changes ;) Glad we are looking into EthPM.

packages/contracts/test/ts/ether_token.ts
```diff
@@ -78,7 +81,8 @@ contract('EtherToken', (accounts: string[]) => {
             const initEthBalance = await getEthBalanceAsync(account);
             const ethTokensToWithdraw = initEthTokenBalance;
             expect(ethTokensToWithdraw).to.not.be.bignumber.equal(0);
-            const txHash = await zeroEx.etherToken.withdrawAsync(ethTokensToWithdraw, account);
+            const txHash = await zeroEx.etherToken.withdrawAsync(ethTokensToWithdraw, account,
+                                                                 {gasLimit: constants.MAX_ETHERTOKEN_WITHDRAW_GAS});
```
fabioberger: I'd prefer we don't inline objects:

```
const txHash = await zeroEx.etherToken.withdrawAsync(ethTokensToWithdraw, account, {
    gasLimit: constants.MAX_ETHERTOKEN_WITHDRAW_GAS,
});
```

packages/contracts/test/ts/unlimited_allowance_token_v2.ts
```diff
@@ -0,0 +1,129 @@
...
+            const initSpenderAllowance = initOwnerBalance;
+            await zeroEx.token.setAllowanceAsync(tokenAddress, owner, spender, initSpenderAllowance);
+            await zeroEx.token.transferFromAsync(tokenAddress, owner, spender, spender, amountToTransfer,
+                                                 {gasLimit: constants.MAX_TOKEN_TRANSFERFROM_GAS});
+
+            const newSpenderAllowance = await zeroEx.token.getAllowanceAsync(tokenAddress, owner, spender);
+            expect(newSpenderAllowance).to.be.bignumber.equal(0);
+        });
+    });
+});
```
fabioberger: The test duplication is also painful.

packages/contracts/contracts/tokens/ERC20Token.sol
```diff
@@ -0,0 +1,58 @@
...
+
+import "./Token_v2.sol";
+
+contract ERC20Token is Token_v2 {
+
+    function transfer(address _to, uint _value)
+        public
+        returns (bool) 
+    {
+        require(balances[msg.sender] >= _value && balances[_to] + _value >= balances[_to]);
```
BMillman19: doesn't everything in here need to be safe mathified

packages/contracts/contracts/tokens/EtherToken_v2.sol
```diff
@@ -0,0 +1,59 @@
...
+
+*/
+
+pragma solidity 0.4.18;
+
+import "./UnlimitedAllowanceToken_v2.sol";
+import "./../utils/SafeMath_v2.sol";
+
+contract EtherToken_v2 is UnlimitedAllowanceToken_v2, SafeMath_v2 {
+
```
BMillman19: Do you think there would be any value here in including the address of the old ether token?

packages/contracts/contracts/tokens/UnlimitedAllowanceToken_v2.sol
```diff
@@ -0,0 +1,47 @@
...
+    /// @param _from Address to transfer from.
+    /// @param _to Address to transfer to.
+    /// @param _value Amount to transfer.
+    /// @return Success of transfer.
+    function transferFrom(address _from, address _to, uint _value)
+        public 
+        returns (bool) 
+    {
+        uint allowance = allowed[_from][msg.sender];
+        require(balances[_from] >= _value && allowance >= _value && balances[_to] + _value >= balances[_to]);
```
BMillman19: safe math?

packages/contracts/contracts/tokens/ERC20Token.sol
abandeali1: Added it back in.

packages/contracts/contracts/utils/SafeMath_v2.sol
abandeali1: Just updating Solidity and changing functions to `pure` (otherwise there will be compiler warnings in the new EtherToken).

packages/contracts/test/ts/unlimited_allowance_token_v2.ts
abandeali1: Yeah, we need to figure out the best structure for testing within contract packages as well.

packages/contracts/contracts/tokens/ERC20Token.sol
abandeali1: It's an option, but this line checks for overflows anyways.

packages/contracts/contracts/tokens/EtherToken_v2.sol
abandeali1: I've thought about a `version` string. Maybe include the old address in comments?

Issue #254, opened 2017-12-07
With 0x.js v0.27.1 Webpack fails to minify `./node_modules/ethereum-address/index.js:3` when running a build


MarcZenn:
Here's an output of the command failure error: 

```
yarn build v0.27.5
$ node_modules/.bin/tsc -p ./ && node scripts/build.js
Creating an optimized production build...
ts-loader: Using typescript@2.5.3 and /Users/marczenn/Desktop/par-ui/tsconfig.json
Failed to compile.

Failed to minify the code from this file:

        ./node_modules/ethereum-address/index.js:3

Read more here: http://bit.ly/2tRViJ9

error Command failed with exit code 1.
```

MarcZenn:
I believe I found a way around this. Please see https://github.com/ccxt/ccxt/issues/299. 

It seems that because there isn't a transpiled version of `./node_modules/ethereum-address/index.js:3` You can use the following hack to circumvent the issue. 

```
      {
        test: require.resolve("ethereum-address"),
        loader: require.resolve("babel-loader"),
        options: { compact: true }
      },
```

This definitely is NOT the way to solve the problem but for now it can get you by.

fabioberger:
Yep, we can remove this dependency and re-implement it ourselves.

davidqhr:
Please remove this dependency.

Now, I have to compile this file in the ci process.
```
node_modules/.bin/babel node_modules/ethereum-address/index.js > node_modules/ethereum-address/index.js
```

davidqhr:
@LogvinovLeon 

I found ethereum-address is still a dependency of @0xproject/assert

[https://github.com/0xProject/0x.js/blob/a6f97181310c9acba176b2a28ffba1ca7d53727d/packages/assert/package.json#L42](https://github.com/0xProject/0x.js/blob/a6f97181310c9acba176b2a28ffba1ca7d53727d/packages/assert/package.json#L42)

PR #257, opened 2017-12-08
Remove ethereum-address dependency and add this logic to the repo

LogvinovLeon:
This PR:
* Resolves: #254


fabioberger:
Can we please wait for #253 to be merged and then move this to the `utils` subpackage and use it throughout all the 3 subpackages that rely on this dependency?

packages/assert/package.json
```diff
@@ -22,6 +22,9 @@
     "url": "https://github.com/0xProject/0x.js/issues"
   },
   "homepage": "https://github.com/0xProject/0x.js/packages/assert/README.md",
+  "dependencies": {
```
BMillman19: merge this and the other dependencies struct

packages/assert/package.json
```diff
@@ -39,7 +42,7 @@
   "dependencies": {
     "@0xproject/json-schemas": "^0.6.10",
     "bignumber.js": "~4.1.0",
-    "ethereum-address": "^0.0.4",
+    "js-sha3": "^0.7.0",
```
BMillman19: shouldnt this go in the utils package dependencies?

packages/website/ts/components/inputs/address_input.tsx
```diff
@@ -1,4 +1,4 @@
-import {isAddress} from 'ethereum-address';
+import {addressUtils} from '@0xproject/utils';
```
BMillman19: we need a corresponding change on line 71

PR #262, opened 2017-12-12
Fix portal subscriptions bug

LogvinovLeon:
This PR:
* Renames `blockAndLogStreamer` to `blockAndLogStreamerIfExists`
* Removes async suffix from non-async function in portal
* Stops unsubscribing on error, cause 0x.js already does that


packages/website/package.json
```diff
@@ -18,7 +18,7 @@
   "author": "Fabio Berger",
   "license": "Apache-2.0",
   "dependencies": {
-    "0x.js": "0xproject/0x.js/packages/0x.js#0x.js@0.27.1",
+    "0x.js": "^0.27.2",
```
fabioberger: Mh, is  this temporary?

packages/website/ts/blockchain.ts
```diff
@@ -517,8 +516,6 @@ export class Blockchain {
                 // to rollbar and stop watching when one occurs
                 // tslint:disable-next-line:no-floating-promises
                 errorReporter.reportAsync(err); // fire and forget
-                // tslint:disable-next-line:no-floating-promises
-                this.stopWatchingExchangeLogFillEventsAsync(); // fire and forget
```
fabioberger: Ah ha!

packages/website/package.json
fabioberger: I guess once we publish the new version, we can revert this to a hard-coded version.

PR #264, opened 2017-12-14
Create WETH Page

fabioberger:
This PR:
* Adds a permanent WETH page to Portal
* Removes the "Convert" button from the balances page
* Re-designs the conversion dialog to be specifically a wrap/unwrap dialog
* Uses a new WETH icon
* Refactors tokenAmountInput so it can be displayed with or without a label
* Adds the ability to clear trackedTokens local storage since we must force users to start tracking the new WETH and stop tracking the old.

screen shot 2017-12-14 at 10 58 24

screen shot 2017-12-14 at 10 58 33


packages/website/ts/components/eth_weth_conversion_button.tsx
```diff
@@ -12,12 +12,15 @@ import {errorReporter} from 'ts/utils/error_reporter';
...
 interface EthWethConversionButtonProps {
+    direction: Side;
     ethToken: Token;
     ethTokenState: TokenState;
     dispatcher: Dispatcher;
     blockchain: Blockchain;
     userEtherBalance: BigNumber;
-    onError: () => void;
+    isOutdatedWrappedEther: boolean;
+    onConversionSuccessful?: () => void;
```
LogvinovLeon: IfExists

packages/website/ts/components/inputs/token_amount_input.tsx
```diff
@@ -8,9 +8,9 @@ import {BalanceBoundedInput} from 'ts/components/inputs/balance_bounded_input';
 import {InputErrMsg, Token, TokenState, ValidatedBigNumberCallback, WebsitePaths} from 'ts/types';
 
 interface TokenAmountInputProps {
-    label: string;
     token: Token;
     tokenState: TokenState;
+    label?: string;
```
LogvinovLeon: IfExists

packages/website/ts/components/portal_menu.tsx
```diff
@@ -50,6 +50,14 @@ export class PortalMenu extends React.Component<PortalMenuProps, PortalMenuState
                 >
                     {this.renderMenuItemWithIcon('Trade history', 'zmdi-format-list-bulleted')}
                 </MenuItem>
+                <MenuItem
+                    style={this.props.menuItemStyle}
+                    className="py2"
+                    to={`${WebsitePaths.Portal}/weth`}
+                    onClick={this.props.onClick.bind(this)}
+                >
+                    {this.renderMenuItemWithIcon('ETH wrapper', 'zmdi-circle-o')}
```
abandeali1: Prefer `Wrapped ETH` for consistency.

fabioberger: Hm... This is the menu item name. It should be a call to action similar to the others: `Generate order`, `Fill order`...

packages/website/ts/components/eth_weth_conversion_button.tsx
fabioberger: I don't think it makes sense to add the `IfExists` prefix to prop names. You might or might not define it when instantiating the component. I'll add a defaultProp value instead.

PR #265, opened 2017-12-14
Implement first custom linter rule async-suffix

LogvinovLeon:
This PR:
* Implements first custom linter rule that checks that every function returning a promise has an `Async` suffix

### DIFF
--- a/packages/tslint-config/rules/asyncSuffixRule.ts
+++ b/packages/tslint-config/rules/asyncSuffixRule.ts
@@ -1,0 +1,10 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+
+import {AsyncSuffixWalker} from './walkers/async_suffix';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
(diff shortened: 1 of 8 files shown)


packages/tslint-config/rules/asyncSuffixRule.ts
```diff
@@ -0,0 +1,10 @@
+import * as Lint from 'tslint';
+import * as ts from 'typescript';
+
+import {AsyncSuffixWalker} from './walkers/async_suffix';
+
+export class Rule extends Lint.Rules.AbstractRule {
+    public apply(sourceFile: ts.SourceFile): Lint.RuleFailure[] {
+        return this.applyWithWalker(new AsyncSuffixWalker(sourceFile, this.getOptions()));
+    }
+}
```
fabioberger: Is it their convention to have these file names be camel-case?

Horbi:
0x000757d2AcC3F49bdf7AF5Ea1156b4969F4eBd97

packages/tslint-config/rules/asyncSuffixRule.ts
LogvinovLeon: It's the requirement

PR #266, opened 2017-12-15
Remove all unused vars and imports

LogvinovLeon:
This PR:
* Enables new linter rule
* Fixes all the errors (oh boy there was a lot)
* Fixes tests on 0x.js (have no idea how did this work before)
* Not all of the things here are stupid unused vars removals. Review carefully

### DIFF
--- a/packages/0x.js/contract_templates/contract.mustache
+++ b/packages/0x.js/contract_templates/contract.mustache
@@ -2,6 +2,7 @@
  * This file is auto-generated using abi-gen. Don't edit directly.
  * Templates can be found at https://github.com/0xProject/0x.js/tree/development/packages/abi-gen-templates.
  */
+// tslint:disable-next-line:no-unused-variable
 import {TxData, TxDataPayable} from '@0xproject/types';
 import {classUtils, promisify} from '@0xproject/utils';
 import {BigNumber} from 'bignumber.js';
--- a/packages/0x.js/test/remaining_fillable_calculator_test.ts
+++ b/packages/0x.js/test/remaining_fillable_calculator_test.ts
@@ -7,7 +7,6 @@ import { RemainingFillableCalculator } from '../src/order_watcher/remaining_fill
 import { ECSignature, SignedOrder } from '../src/types';
 
 import { chaiSetup } from './utils/chai_setup';
-import { TokenUtils } from './utils/token_utils';
 
 chaiSetup.configure();
 const expect = chai.expect;
--- a/packages/0x.js/test/remaining_fillable_calculator_test.ts
+++ b/packages/0x.js/test/remaining_fillable_calculator_test.ts
@@ -2,11 +2,11 @@ import BigNumber from 'bignumber.js';
 import * as chai from 'chai';
 import 'mocha';
 
-import { ZeroEx } from '../src/0x';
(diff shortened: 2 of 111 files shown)


packages/0x.js/contract_templates/contract.mustache
```diff
@@ -2,6 +2,7 @@
  * This file is auto-generated using abi-gen. Don't edit directly.
  * Templates can be found at https://github.com/0xProject/0x.js/tree/development/packages/abi-gen-templates.
  */
+// tslint:disable-next-line:no-unused-variable
```
fabioberger: We shouldn't be linting `.mustache` files.

packages/0x.js/test/remaining_fillable_calculator_test.ts
```diff
@@ -7,7 +7,6 @@ import { RemainingFillableCalculator } from '../src/order_watcher/remaining_fill
 import { ECSignature, SignedOrder } from '../src/types';
```
abandeali1: Remove spaces.

packages/0x.js/test/remaining_fillable_calculator_test.ts
```diff
@@ -7,7 +7,6 @@ import { RemainingFillableCalculator } from '../src/order_watcher/remaining_fill
...
 
 import { chaiSetup } from './utils/chai_setup';
```
abandeali1: Remove spaces.

packages/subproviders/src/subproviders/subprovider.ts
```diff
@@ -34,9 +33,6 @@ export class Subprovider {
     }
     public setEngine(engine: any): void {
         this.engine = engine;
-        engine.on('block', (block: any) => {
-            this.currentBlock = block;
-        });
```
fabioberger: Mh... we cannot remove this... Even though we don't use it, it could be used by a super class. I think this should be `protected` rather then `private`.

packages/tslint-config/tslint.json
```diff
@@ -51,6 +51,7 @@
     "no-string-throw": true,
     "no-submodule-imports": false,
     "no-unnecessary-type-assertion": true,
+    "no-unused-variable": [true, {"ignore-pattern": "^_\\d*"}],
```
fabioberger: What pattern are we ignoring here?

packages/0x.js/contract_templates/contract.mustache
LogvinovLeon: We're not. But generated files have unused import TxDataPayable if the generated contract doesn't have payable functions. It'll be too complicated to detect it so I just ignore it.

packages/subproviders/src/subproviders/subprovider.ts
LogvinovLeon: Could be. Dead code. We'll add it as soon as it will be used by a super class.

packages/tslint-config/tslint.json
LogvinovLeon: _, _1, _2, etc. There are some places in code where I needed two dummy vars

packages/0x.js/contract_templates/contract.mustache
fabioberger: Ah, I see!

PR #267, opened 2017-12-17
Refactor EtherToken wrapper to accept address in method args

fabioberger:
This PR:
- Modifies the `etherTokenWrapper` methods to accept an `etherTokenAddress` as the first arg. Since it is becoming apparent we will be updating the canonical WETH contract, we want users of 0x.js to be able to interact with n number of etherToken contracts without re-instantiating the library for each one. Because of this, we modify the interface for `etherTokenWrapper` methods to be similar to `tokenWrapper` methods.
- Unrelated: I also fixed an issue with 0x.js docs where `unsubscribeAll` was showing up as a method in each contractWrapper instance since it was declared in the base class. Since it is only used by `exchangeWrapper` and `tokenWrapper`, I made it `protected` and re-declared it in those two.

### DIFF
--- a/packages/0x.js/CHANGELOG.md
+++ b/packages/0x.js/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+v0.28.0 - _TBD_
+------------------------
+* Add `etherTokenAddress` arg to `depositAsync` and `withdrawAsync` methods on `EtherToken` wrapper. (#267)
+
 v0.27.1 - _November 28, 2017_
 ------------------------
     * Export `TransactionOpts` type
--- a/packages/0x.js/CHANGELOG.md
+++ b/packages/0x.js/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 v0.28.0 - _TBD_
 ------------------------
-* Add `etherTokenAddress` arg to `depositAsync` and `withdrawAsync` methods on `EtherToken` wrapper. (#267)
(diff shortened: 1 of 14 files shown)


packages/0x.js/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+v0.28.0 - _TBD_
+------------------------
+* Add `etherTokenAddress` arg to `depositAsync` and `withdrawAsync` methods on `EtherToken` wrapper. (#267)
```
LogvinovLeon: on `zeroEx.etherToken`. The fact that it's a wrapper is an internat knowledge, but changelog is for external users.

packages/0x.js/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -50,10 +50,7 @@ export class ContractWrapper {
         this._onLogAddedSubscriptionToken = undefined;
         this._onLogRemovedSubscriptionToken = undefined;
     }
-    /**
-     * Cancels all existing subscriptions
-     */
-    public unsubscribeAll(): void {
+    protected unsubscribeAll(): void {
```
LogvinovLeon: Good catch!

packages/0x.js/CHANGELOG.md
LogvinovLeon: Also - document two other changes. Removal of public method unsubscribeAll and removal of config parameter in top-level ZeroExConfig,

PR #268, opened 2017-12-17
Fix code selection bug in docs

fabioberger:
This PR:
* Fixes https://github.com/0xProject/wiki/issues/19
* Disables re-rendering markdownCodeBlock renderer if props haven't changed. Re-rendering this component causes a user's selection to become de-selected. This is annoying while trying to copy-paste code examples from wiki, 0x.js docs, etc...
* Adds a convenience `yarn lerna:rebuild` command that first calls `clean` followed by `build`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
     "scripts": {
         "testrpc": "testrpc -p 8545 --networkId 50 -m \"${npm_package_config_mnemonic}\"",
         "lerna:run": "lerna run",
+        "lerna:rebuild": "lerna run clean; lerna run build;",
         "lerna:publish": "yarn install; lerna run clean; lerna run build; lerna publish --registry=https://registry.npmjs.org/"
     },
     "config": {
--- a/packages/website/ts/pages/shared/markdown_code_block.tsx
+++ b/packages/website/ts/pages/shared/markdown_code_block.tsx
@@ -7,14 +7,23 @@ interface MarkdownCodeBlockProps {
     language: string;
 }
 
-export function MarkdownCodeBlock(props: MarkdownCodeBlockProps) {
-    return (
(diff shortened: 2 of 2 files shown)


packages/website/ts/pages/shared/markdown_code_block.tsx
```diff
@@ -7,14 +7,23 @@ interface MarkdownCodeBlockProps {
...
-            <HighLight
-                className={props.language || 'js'}
-            >
-                {props.literal}
-            </HighLight>
-        </span>
-    );
+interface MarkdownCodeBlockState {}
+
+export class MarkdownCodeBlock extends React.Component<MarkdownCodeBlockProps, MarkdownCodeBlockState> {
```
LogvinovLeon: Why not `React.PureComponent`? it will give the same behaviour.

package.json
```diff
@@ -7,6 +7,7 @@
     "scripts": {
         "testrpc": "testrpc -p 8545 --networkId 50 -m \"${npm_package_config_mnemonic}\"",
         "lerna:run": "lerna run",
+        "lerna:rebuild": "lerna run clean; lerna run build;",
```
LogvinovLeon: `lerna run clean && lerna run build`
As far as I'm aware `lerna run clean` asks for a user confirmation.
Maybe `yes | lerna run clean`?

fabioberger: No it doesn't. You are thinking of `lerna clean` ;)

packages/website/ts/pages/shared/markdown_code_block.tsx
fabioberger: Turns out you can only do this when all sub-components are also pure. This is not the case of `HighLight` component.

PR #271, opened 2017-12-18
Implement ledger optimization for address derivation

LogvinovLeon:
This PR:
* Uses HD library to derive addresses from pk and chain code instead of asking ledger for each address

### DIFF
--- a/packages/subproviders/src/subproviders/ledger.ts
+++ b/packages/subproviders/src/subproviders/ledger.ts
@@ -145,7 +145,10 @@ export class LedgerSubprovider extends Subprovider {
         for (let i = 0; i < NUM_ADDRESSES_TO_FETCH; i++) {
             const derivedHDNode = hdKey.derive(`m/${i + this._derivationPathIndex}`);
             const derivedPublicKey = derivedHDNode.publicKey;
-            const ethereumAddressUnprefixed = ethUtil.publicToAddress(derivedPublicKey, true).toString('hex');
+            const shouldSanitizePublicKey = true;
+            const ethereumAddressUnprefixed = ethUtil.publicToAddress(
+                derivedPublicKey, shouldSanitizePublicKey,
+            ).toString('hex');
             const ethereumAddressPrefixed = ethUtil.addHexPrefix(ethereumAddressUnprefixed);
             accounts.push(ethereumAddressPrefixed.toLowerCase());
         }
--- a/packages/subproviders/src/subproviders/ledger.ts
+++ b/packages/subproviders/src/subproviders/ledger.ts
@@ -2,6 +2,7 @@ import {assert} from '@0xproject/assert';
 import {addressUtils} from '@0xproject/utils';
 import EthereumTx = require('ethereumjs-tx');
 import ethUtil = require('ethereumjs-util');
+import HDNode = require('hdkey');
 import * as _ from 'lodash';
 import Semaphore from 'semaphore-async-await';
 import Web3 = require('web3');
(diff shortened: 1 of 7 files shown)


packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -127,21 +128,27 @@ export class LedgerSubprovider extends Subprovider {
...
 
-        // TODO: replace with generating addresses without hitting Ledger
+        let ledgerResponse;
+        try {
+            ledgerResponse = await this._ledgerClientIfExists.getAddress_async(
+                `${this._derivationPath}`, this._shouldAlwaysAskForConfirmation, SHOULD_GET_CHAIN_CODE,
+            );
+        } finally {
+            await this.destoryLedgerClientAsync();
+        }
```
fabioberger: We should still handle any potential errors here with a `catch` statement.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -127,21 +128,27 @@ export class LedgerSubprovider extends Subprovider {
     public async getAccountsAsync(): Promise<string[]> {
         this._ledgerClientIfExists = await this.createLedgerClientAsync();
 
-        // TODO: replace with generating addresses without hitting Ledger
+        let ledgerResponse;
+        try {
+            ledgerResponse = await this._ledgerClientIfExists.getAddress_async(
+                `${this._derivationPath}`, this._shouldAlwaysAskForConfirmation, SHOULD_GET_CHAIN_CODE,
```
fabioberger: No need for template string, can simply do `this._derivationPath`.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -127,21 +128,27 @@ export class LedgerSubprovider extends Subprovider {
...
-                    derivationPath, this._shouldAlwaysAskForConfirmation, SHOULD_GET_CHAIN_CODE,
-                );
-                accounts.push(result.address.toLowerCase());
-            } catch (err) {
-                await this.destoryLedgerClientAsync();
-                throw err;
-            }
+            const derivedHDNode = hdKey.derive(`m/${i + this._derivationPathIndex}`);
+            const derivedPublicKey = derivedHDNode.publicKey;
+            const ethereumAddressUnprefixed = ethUtil.publicToAddress(derivedPublicKey, true).toString('hex');
```
fabioberger: Assign `true` to an intermediate variable. What does it represent?

packages/subproviders/src/types.ts
```diff
@@ -11,7 +11,7 @@ export interface LedgerCommunicationClient {
  */
 export interface LedgerEthereumClient {
     getAddress_async: (derivationPath: string, askForDeviceConfirmation: boolean,
-                       shouldGetChainCode: boolean) => Promise<LedgerGetAddressResult>;
+                       shouldGetChainCode: true) => Promise<LedgerGetAddressResult>;
```
fabioberger: keck?

packages/subproviders/src/subproviders/ledger.ts
LogvinovLeon: We were not doing that in a previous implementation. How do we want to handle them?
Before we were just re-throwing them. This is the same. Otherwise - we'll need to have a mapping for all possible ledger errors.

packages/subproviders/src/types.ts
LogvinovLeon: It's important. Other types assume that ledger returns chainCode so this helps to enforce that. So that the code can't pass false but still expect a chainCode.

LogvinovLeon:
Yes, I've tried that with my test ledger on MEW

packages/subproviders/src/types.ts
fabioberger: Hm... is this the right way to handle this? This should describe the interface to the `LedgerEthereumClient` and it can receive either value. I think before the code that relies on this being `true`, you should assert it and throw with a helpful error message if it is set to false. This way, it's explicit that the future coder needs to refactor the logic if they want to set it to `false`.

LogvinovLeon: Debatable. It's like defining the type for the whole module when you just use one function (extrapolating here). We only use one version of this function. So we restrict it with the type and the user will get a helpfull message. Also - where should we assert it? In our code?
```
const stuff = true;
assert(stuff);
call(smth, stuff);
```
Seems strange to me.

LogvinovLeon: I'd prefer type system to assertions system.

PR #273, opened 2017-12-18
Add proper types for yargs and ethereumjs-abi

LogvinovLeon:
This PR:
* Adds proper types for yargs and ethereumjs-abi in contracts

### DIFF
--- a/packages/contracts/deploy/cli.ts
+++ b/packages/contracts/deploy/cli.ts
@@ -23,12 +23,12 @@ const DEFAULT_GAS_PRICE = ((10 ** 9) * 2).toString();
  * Compiles all contracts with options passed in through CLI.
  * @param argv Instance of process.argv provided by yargs.
  */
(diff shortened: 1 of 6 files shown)


packages/contracts/deploy/cli.ts
```diff
@@ -150,11 +150,11 @@ function deployCommandBuilder(yargsInstance: any) {
         })
         .command('compile',
                  'compile contracts',
-                 _.noop,
+                 _.identity,
                  onCompileCommand)
         .command('migrate',
                  'compile and deploy contracts using migration scripts',
-                 _.noop,
+                 _.identity,
```
fabioberger: Why does this need to return the first arg it receives?

fabioberger: Maybe add a short comment to the right of it. It seems non-obvious to me.

PR #278, opened 2017-12-19
Throw a better error message when taker is null|undefined or anything but not a string

LogvinovLeon:
This PR:
*  Improves an error message when taker is anything except string.
* Refactors our decorators module so that we can more easily improve other error messages and make the handlers composable.
* Adds a test for a new behaviour
https://airtable.com/tblFe0Q9JuKJPYbTn/viw2LUWhbLzAsfGXH/recJWzqENfEUCbA0D

### DIFF
--- a/packages/0x.js/src/utils/decorators.ts
+++ b/packages/0x.js/src/utils/decorators.ts
@@ -20,7 +20,7 @@ const schemaErrorTransformer = (error: Error) => {
     if (_.includes(error.message, constants.INVALID_TAKER_FORMAT)) {
         // tslint:disable-next-line:max-line-length
         const errMsg = 'Order taker must be of type string. If you want anyone to be able to fill an order - pass ZeroEx.NULL_ADDRESS';
-        throw new Error(errMsg);
+        return new Error(errMsg);
     }
     return error;
 };
--- a/packages/0x.js/src/utils/decorators.ts
+++ b/packages/0x.js/src/utils/decorators.ts
@@ -1,17 +1,37 @@
 import * as _ from 'lodash';
 
-import {AsyncMethod, ZeroExError} from '../types';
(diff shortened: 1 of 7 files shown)


packages/0x.js/src/utils/decorators.ts
```diff
@@ -1,17 +1,37 @@
...
-                             descriptor: TypedPropertyDescriptor<AsyncMethod>,
-    ): TypedPropertyDescriptor<AsyncMethod> {
+type ErrorTransformer = (err: Error) => Error;
+
+const contractCallErrorTransformer = (error: Error) => {
+    if (_.includes(error.message, constants.INVALID_JUMP_PATTERN)) {
+        return new Error(ZeroExError.InvalidJump);
+    }
+    if (_.includes(error.message, constants.OUT_OF_GAS_PATTERN)) {
+        return new Error(ZeroExError.OutOfGas);
```
fabioberger: Ditto here.

packages/0x.js/src/utils/decorators.ts
```diff
@@ -1,17 +1,37 @@
...
-     */
-    contractCallErrorHandler(target: object,
-                             key: string|symbol,
-                             descriptor: TypedPropertyDescriptor<AsyncMethod>,
-    ): TypedPropertyDescriptor<AsyncMethod> {
+type ErrorTransformer = (err: Error) => Error;
+
+const contractCallErrorTransformer = (error: Error) => {
+    if (_.includes(error.message, constants.INVALID_JUMP_PATTERN)) {
+        return new Error(ZeroExError.InvalidJump);
```
fabioberger: Should we be throwing this? Just in case this transformer is used in another chain in the future. It just so happens that it's the last transfer right now.

PR #285, opened 2017-12-20
Add new underscore-privates rule to @0xproject/tslint-config and fix …

BMillman19:
…lint errors

This PR:
* Added custom `underscore-privates` rule to @0xproject/tslint-config
* Fixed resulting lint errors
* Update changelog for @0xproject/tslint-config

https://airtable.com/tblFe0Q9JuKJPYbTn/viw2LUWhbLzAsfGXH/recq2QxxyfR0b4Zuv


packages/contracts/deploy/src/utils/contract.ts
```diff
@@ -23,7 +23,7 @@ export class Contract implements Web3.ContractInstance {
         this.populateFunctions();
```
fabioberger: Erh... these are missing the underscores. How did you do this find and replace? I think we should be using a abstract syntax tree parser that knows what a method declaration is and what a method call is.

fabioberger: Ah, this issue arose because of the use of `any` in our typings for `web3.ContractInstance`. We should move to using the `abi-gen` generated types here. Created a task for that. In the mean time, we should make sure this isn't happening anywhere else in the codebase.

PR #288, opened 2017-12-21
Fix doc anchor link collisions

fabioberger:
This PR:
* Fixes a bug where if two sections contain methods with the same name, the anchor link would always navigate to the first method. By namespacing the anchors below sectionNames, we alleviate this issue.


BMillman19:
lgtm code wise, can you list repro steps for the bug?

fabioberger:
Yep, you can observe it by clicking on `subscribeAll` in any of the sections other then `exchange` (the top section) in the 0x.js docs.

PR #290, opened 2017-12-21
Fix build minification issue

fabioberger:
This PR:
* Uses a custom fork of `truffle-hdwallet-provider` that upgrades the internal dep. of `web3-provider-engine` to `^13.0.1`, a version without ES7 features. This fixes webpacks build minification process.


fabioberger:
Turns out this isn't the underlying issue of the minification bug. Do we still want to use our own fork with the updated dependency? @LogvinovLeon @BMillman19

LogvinovLeon:
I don't see the point

PR #291, opened 2017-12-22
Use prettier

LogvinovLeon:
This PR:
* Adds prettier and uses it to reformat all files
* Runs prettier on CI

Prettier is the tool that auto-formats code. You no longer need to fix those annoying linter errors manually. You just hit save and the code is properly formatted. Like gofmt

TODO items for all:
You need to configure your setup to auto-format files.
The easiest way is to use one of the editors integrations https://prettier.io/docs/en/editors.html
You can also set up a pre-commit hook https://prettier.io/docs/en/precommit.html


packages/0x.js/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -34,12 +36,14 @@ export class ContractWrapper {
     protected _web3Wrapper: Web3Wrapper;
     private _networkId: number;
     private _abiDecoder?: AbiDecoder;
-    private _blockAndLogStreamerIfExists: BlockAndLogStreamer|undefined;
+    private _blockAndLogStreamerIfExists: BlockAndLogStreamer | undefined;
```
fabioberger: Are we leaving this for now? I'm ok with it. It is consistent with the styling in many other projects.

PR #295, opened 2017-12-24
Fix array typing generation when permitting multiple types

nadavhollander:
Incorrect typings were being generated when an array type consisted of elements separated by an | operator.  For instance, if the constituent element of the array was a `number|BigNumber`, the generated array type would be `number|BigNumber[]`.

Fixed by grouping all constituent elements in a type array with parentheses.


BMillman19:
Thanks dude!

fabioberger:
@LogvinovLeon mind adding a line to the changelog and publishing a new version?

PR #296, opened 2017-12-25
Prevent getFeesAsync method on HttpClient from mutating input

BMillman19:
This PR:
* Remove type conversion of the `feesRequest` in `getFeesAsync()`
* JSON.stringify in the `_reqestAsync()` method already takes care of this for us
* Remove BigNumber to string type conversion utility because it is no longer used
* Added a test


packages/connect/test/http_client_test.ts
```diff
@@ -124,6 +124,18 @@ describe('HttpClient', () => {
             const fees = await relayerClient.getFeesAsync(request);
             expect(fees).to.be.deep.equal(feesResponse);
         });
+        it('does not mutate input', async () => {
+            fetchMock.post(url, feesResponseJSON);
+            const makerTokenAmountBefore = new BigNumber(request.makerTokenAmount);
+            const takerTokenAmountBefore = new BigNumber(request.takerTokenAmount);
+            const saltBefore = new BigNumber(request.salt);
+            const expirationUnixTimestampSecBefore = new BigNumber(request.expirationUnixTimestampSec);
+            await relayerClient.getFeesAsync(request);
```
fabioberger: Are we sure there is no mutation in any of the other methods in `Connect`?

fabioberger:
@BMillman19 don't forget to add a CHANGELOG entry for `Connect`. It should be part of this PR.

PR #297, opened 2017-12-25
Refactor JSON parsing in HttpClient

BMillman19:
This PR:
* Move parsing logic into bank of parsers that `HttpClient` imports
* Refactor `typeConverters` to use pure, non-mutating functions
* Update `orderbookChannelMessageParser` to use the new parsers


packages/connect/src/http_client.ts
```diff
@@ -154,6 +166,6 @@ export class HttpClient implements Client {
             throw Error(response.statusText);
         }
         const json = await response.json();
-        return json;
+        return jsonParser(json);
```
fabioberger: Assign this to a variable before returning.

packages/connect/src/http_client.ts
```diff
@@ -115,10 +118,13 @@ export class HttpClient implements Client {
...
             payload: request,
         };
-        const fees = await this._requestAsync('/fees', HttpRequestType.Post, requestOpts);
-        assert.doesConformToSchema('fees', fees, schemas.relayerApiFeesResponseSchema);
-        typeConverters.convertStringsFieldsToBigNumbers(fees, ['makerFee', 'takerFee']);
-        return fees;
+        const result = await this._requestAsync(
+            '/fees',
+            HttpRequestType.Post,
+            relayerResponseJsonParsers.parseFeesResponseJson,
```
fabioberger: Why rather pass these methods in then simply calling them on the return value of `_requestAsync`. I would prefer the latter. It's clearer to me.

PR #300, opened 2017-12-28
Update subprovider to catch correct RPC method

cavanmflynn:

* Changed `personal_sign` to `eth_sign`. ZeroEx uses `web3.eth.sign`, rather than `web3.eth.personal.sign`, so the Ledger subprovider should be looking for `eth_sign`.
* Get the data to sign from the second parameter, rather than the first. The first parameter is the address.

### DIFF
--- a/packages/subproviders/src/subproviders/ledger.ts
+++ b/packages/subproviders/src/subproviders/ledger.ts
@@ -106,8 +106,8 @@ export class LedgerSubprovider extends Subprovider {
                 }
                 return;
 
-            case 'personal_sign':
-                const data = payload.params[0];
+            case 'eth_sign':
+                const data = payload.params[1];
                 try {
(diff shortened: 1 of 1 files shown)


dekz:
Sweet good pickup.

It should be possible for us to support both request methods here and support both web3.eth.sign and web3.personal.sign. Although 0x.js uses web3.eth.sign, perhaps other applications just using the subproviders may not.

```
case 'eth_sign':
case 'personal_sign':
```

Although it seems that the positional argument changes. 
`eth_sign` is address, message
`personal_sign` is message, address

I'd be happy to track personal_sign in another PR.

fabioberger:
Thanks for the submission @cavanmflynn! 

Going to merge this and then will update the tests which still call `eth_sign`.

PR #304, opened 2018-01-03
Added constructor ABIs to abi-gen

ritave:
Additionally if the constructor is not existent in JSON, meaning it's
implicite with no parameters, we're
explicitly creating one, with actual JSON parameters that it should have.

### DIFF
--- a/packages/abi-gen/src/index.ts
+++ b/packages/abi-gen/src/index.ts
@@ -14,6 +14,7 @@ import * as Web3 from 'web3';
 import { ContextData, ParamKind } from './types';
 import { utils } from './utils';
 
+const ABI_TYPE_CONSTRUCTOR = 'constructor';
 const ABI_TYPE_METHOD = 'function';
 const ABI_TYPE_EVENT = 'event';
 const MAIN_TEMPLATE_NAME = 'contract.mustache';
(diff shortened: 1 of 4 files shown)


packages/abi-gen/src/index.ts
```diff
@@ -75,6 +76,11 @@ for (const abiFileName of abiFileNames) {
         process.exit(1);
     }
 
+    let constructor = ABI.find((abi: Web3.AbiDefinition) => abi.type === ABI_TYPE_CONSTRUCTOR) as Web3.ConstructorAbi;
+    if (!constructor) {
```
LogvinovLeon: `if (_.isUndefined(constructor)) {`

packages/abi-gen/src/index.ts
```diff
@@ -75,6 +76,11 @@ for (const abiFileName of abiFileNames) {
         process.exit(1);
     }
 
+    let constructor = ABI.find((abi: Web3.AbiDefinition) => abi.type === ABI_TYPE_CONSTRUCTOR) as Web3.ConstructorAbi;
```
LogvinovLeon: `constructor` is a reserved name in JS. Let's name it `ctor`

PR #307, opened 2018-01-04
Base tsconfig.json

LogvinovLeon:
This PR:
* Imports web3-typscript-typings to the monorepo with the whole history
* Refactors common parts of tsconfig.json to a common tsconfig.json
* Enables strict mode and fixes some of the issues
* Some of the other issues are harder/impossible to fix - that's why in some packages some checks that are part of the strict mode are disabled


BMillman19:
There is a build error when the changes from https://github.com/0xProject/0x.js/pull/304 are included: 

```
src/utils.ts(66,9): error TS2322: Type '{ type: "constructor"; stateMutability: "nonpayable"; payable: false; inputs: never[]; }' is not assignable to type 'ConstructorAbi'.
  Types of property 'type' are incompatible.
    Type '"constructor"' is not assignable to type 'AbiType.Constructor'.
```

PR #318, opened 2018-01-12
Improve HttpClient errors

BMillman19:
This PR:

* Sanitize HttpClient api urls and improve error messaging

### DIFF
--- a/packages/connect/CHANGELOG.md
+++ b/packages/connect/CHANGELOG.md
@@ -2,8 +2,9 @@
 
 ## v0.x.x - _TBD, 2017_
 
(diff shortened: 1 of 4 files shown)


LogvinovLeon:
Can we also try to prevent people from passing a v0 suffix as we add it ourselves?

BMillman19:
@LogvinovLeon should we keep adding a v0 to the end of the passed in URL? an api url can implement standard relayer api v0 w/o having a v0 at the end of the path

LogvinovLeon:
I think we shouldn't

BMillman19:
Yea I agree, i will remove it

PR #320, opened 2018-01-15
Fix web3 typing defaultAccount not allowing `undefined`

ritave:
The default value for the defaultAccount property is undefined,
this happens in real-life scenarios, and so the type should
explicitly warn users about it

### DIFF
--- a/packages/web3-typescript-typings/CHANGELOG.md
+++ b/packages/web3-typescript-typings/CHANGELOG.md
@@ -3,4 +3,4 @@
 ## v0.9.3 - _January 11, 2018_
 
 * Add type for getData on a contract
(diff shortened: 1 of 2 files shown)


packages/web3-typescript-typings/CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 ## v0.9.3 - _January 11, 2018_
 
 * Add type for getData on a contract
+* Fixed the `defaultAccount` not allowing for `undefined` value
```
LogvinovLeon: Please add a PR number as we have here: https://github.com/0xProject/0x.js/blob/development/packages/0x.js/CHANGELOG.md

PR #322, opened 2018-01-16
Speed-up CI 

LogvinovLeon:
This PR:

* Reduces the CI tests running time from [~12-13 min](https://circleci.com/gh/0xProject/0x.js/2305) to [~7 min](https://circleci.com/workflow-run/45c09a14-7a8c-4a4c-b8f4-425f6eee9ab1) by parallelizing some of the tests
* Removes UMD tests
* Upgrades testrpc two patch versions forward, because there are some race-conditions fixes there
* Adds a fix to make tests more reliable on pending events (Ganache issue)
* Adds `--frozen-lockfile` flag to CI yarn config (apparently that wasn't the default) https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile

There was also an interesting issue that causes CI to fail randomly.
It was caused by [this](https://github.com/0xProject/0x.js/commit/cb11aec84df346d5180c7d5874859c1c34f0bf1c#diff-b3747316e1fbfb200f0590928ddc4010L44) line actually incrementing the port.
I've reported a TS issue and they've recommended the TSLint rule which I enabled.


fabioberger:
@LogvinovLeon did we decide to no longer support UMD bundles for 0x.js? If so, we need to also remove it from the 0x.js docs.

package.json
```diff
@@ -17,7 +17,7 @@
     "devDependencies": {
         "@0xproject/utils": "^0.1.0",
         "async-child-process": "^1.1.1",
-        "ethereumjs-testrpc": "6.0.3",
+        "ethereumjs-testrpc": "^6.0.3",
```
fabioberger: Hm, should we specify the specific minor and patch version we want or do we trust `TestRPC` to properly adhere to semantic versioning?

packages/0x.js/src/order_watcher/event_watcher.ts
```diff
@@ -51,6 +51,10 @@ export class EventWatcher {
     }
     private async _pollForBlockchainEventsAsync(callback: EventWatcherCallback): Promise<void> {
         const pendingEvents = await this._getEventsAsync();
+        if (_.isUndefined(pendingEvents)) {
+            // HACK: This should never happen, but happens frequently on CI due to a ganache bug
```
fabioberger: Hm, could we check if we are running in CI by checking an environment variable in this conditional block as well?  That would make it more explicit that this conditional only exists for CircleCI.

package.json
LogvinovLeon: I've added `--frozen-lockfile` flag so as long as someone doesn't manually update it we'll stay on that version

packages/0x.js/src/order_watcher/event_watcher.ts
LogvinovLeon: I'm not sure if it does and why this happens. Personally I'd rather have a warning here, but 0x.js didn't emit any output before and I'm not sure this is a good time to start.
This is smth that should never happen, but it isn't critical and we can recover from that.

package.json
fabioberger: That flag only applies for CI runs and not when people are running the tests locally. I'd prefer we explicitly version.

packages/0x.js/src/order_watcher/event_watcher.ts
fabioberger: Ok, then let's just move the `HACK` comment above the conditional and also link to the specific Ganache issue on Github.

packages/dev-utils/src/rpc.ts
```diff
@@ -41,7 +41,7 @@ export class RPC {
             method,
             params,
         });
-        this._url += 1;
+        this._id += 1;
```
fabioberger: @BMillman19 this happened during the big `underscore for private methods` refactor you undertook. Any ideas how this could have happened?

package.json
LogvinovLeon: Not really. We have a yarn.lock in the repo and unless someone force-updates or removes the file the version will stay locked. That's the whole point of yarn. Deterministic builds.

packages/0x.js/src/order_watcher/event_watcher.ts
LogvinovLeon: There is no specific ganache issue because I haven't investigated the cause yet.
I don't feel like reporting smth like: "Your thing returns undefined in random cases" is good enough to create an issue.

PR #324, opened 2018-01-18
Reject negative amounts in public methods

### LINKED ISSUE #212: 0x.js passes negative `takerTokenAmount` to the exchange contract
I was poking around in 0x.js and the contracts, and it looks like 0x.js will happily pass along a negative value for `takerTokenAmount` (via `zeroEx.exchange.fillOrderAsync`) to the contract's `exchange.fillOrder`.  When a negative BigNumber meets uint256, you end up trying to buy 10^17 of something (or the entire order, whichever's smaller), which is probably not what you wanted!

See input data field [11] in this transaction (`takerTokenAmount`), which was filled with a very small negative number: https://kovan.etherscan.io/tx/0x8c83193f05137eb682e0ab41dd2bfdae43db84b5704588761eb08d48547e588b

dekz:
Fixes #212 

Previously negative fill amounts could be passed into the public methods of exchange wrapper, be valid, and submitted to the contract. This would result in the user filling much more than they expected.

* added an assertion that the amount passed into assert `isValidBaseUnitAmount`, asserting that the value must not be negative.
* The batch fills use a schema and negative amounts fail the Number schema validation


packages/assert/src/index.ts
```diff
@@ -12,6 +12,8 @@ export const assert = {
     },
     isValidBaseUnitAmount(variableName: string, value: BigNumber) {
         assert.isBigNumber(variableName, value);
+        const isNegative = value.lessThan(0);
+        this.assert(!isNegative, `${variableName} should not be a negative number, found value: ${value.toNumber()}` );
```
fabioberger: `should not` -> `cannot`

dekz:
Hmm seems like I got a test failure on the contracts repo due to a port issue, I'll rebase from development.

PR #312, opened 2018-01-10
Better error handling on async/sync intervals

LogvinovLeon:
This PR:
* Resolves https://github.com/0xProject/0x.js/issues/308
* Makes `EventWatcher` and `OrderWatcher` callbacks failable
* Extends intervalUtils to also support sync intervals

PR #325, opened 2018-01-18
Small Fixes/Additions

fabioberger:
This PR:

- Fixes the copy on the Balances page of Portal to inform users to wrap/unwrap ETH/WETH on the Wrap ETH tab.
- Re-instates the mobile menu button on the topBar of Portal for small screens/mobile.
- Adds a link to our forum from the website footer.
- Adds the missing error callback to calls to `setAsyncExcludingInterval` missed by PR #312.


LogvinovLeon:
It's not the change introduced by this PR, but let's not use `clearInterval` directly and use it through `intervalUtils`. I think it improves grepability.

PR #327, opened 2018-01-19
Make the top level package a lerna managed package

BMillman19:
This PR:
* Adds `./` as a top managed package to lerna
* This resolves the issue we saw here: https://github.com/0xProject/0x.js/pull/326 where our top level @0xproject/utils dependency does not get automatically updated on publish
* This fix is proposed but warned against here: https://github.com/lerna/lerna/issues/1216
* w/o this change we just need to remember to manually update versions in our top level directory, also `yarn install` won't link to the local version of @0xproject/utils, causing local changes not to be reflected

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
     "lerna": "2.5.1",
-    "packages": ["packages/*"],
+    "packages": ["./", "packages/*"],
     "version": "independent",
     "npmClient": "yarn",
     "useWorkspaces": true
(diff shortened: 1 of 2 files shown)


fabioberger:
This is a short-term solution given that it won't be supported in Lerna V3. Maybe it's fine if we try and remember to do this manually, or otherwise have the Github bot remind us?

BMillman19:
Yea we can use a combination of greenkeeper bot and an extra check in the publishing process doc

PR #329, opened 2018-01-19
Implement stop-gap smart contract versioning

fabioberger:
This PR:
- Moves the contracts repo away from a suffix-based contract versioning system towards a copy-on-write versioning system that stores previous versions of smart contracts locally.


packages/contracts/src/previous/ERC20Token/ERC20Token_v1.sol
```diff
@@ -1,8 +1,8 @@
 pragma solidity ^0.4.11;
 
-import "./Token.sol";
+import { Token_v1 as Token } from "../Token/Token_v1.sol";
 
-contract StandardToken is Token {
+contract ERC20Token_v1 is Token {
```
fabioberger: This does not look right. What happened over here?

abandeali1: I just renamed it for consistency, seems much more descriptive.

PR #330, opened 2018-01-22
[WIP] Basic Parameter Decorator for Ethereum Address

dekz:
Wanted to get a PR for this open as soon as possible so we can discuss.

This PR creates a Parameter decorator (and a Method decorator) to transform values passed into public interfaces such as applying `lowerCase` of Ethereum address'. The aim is to gently aide the user rather than raise an assertion error.

Decorators are applied first before assertions.

Two decorators are required as the Parameter decorator doesn't apply, only the Method decorator is able to modify values when calling the method.

The process is as follows:
  1. Register the parameter to be transformed, and the transformation function
  2. Register the Method
  3. When the method is called the Method Decorator transforms the correct parameters and calls the method with transformed values.

Example Usage:
```typescript
    @decorators.asyncZeroExErrorHandler
    @decorators.parameterTransformer // Need to set the Method Decorator
    public async fillOrderAsync(
        signedOrder: SignedOrder,
        fillTakerTokenAmount: BigNumber,
        shouldThrowOnInsufficientBalanceOrAllowance: boolean,
        @decorators.ethereumAddress takerAddress: string, // Essentially tag the parameter by index
        orderTransactionOpts: OrderTransactionOpts = {},
    ): Promise<string> {
```

Todo:
  - [ ] make sure the naming is clear and makes sense
  - [ ] do Generics of ParameterTransformer<String> make sense here?
  - [ ] are there other examples of transformation/coercion?


packages/0x.js/src/utils/decorators.ts
```diff
@@ -85,7 +91,40 @@ const syncErrorHandlerFactory = (errorTransformer: ErrorTransformer) => {
...
+        return descriptor.value.apply(this, args);
+    };
+    return descriptor;
+};
+
 export const decorators = {
     asyncZeroExErrorHandler: asyncErrorHandlerFactory(zeroExErrorTransformer),
     syncZeroExErrorHandler: syncErrorHandlerFactory(zeroExErrorTransformer),
+    ethereumAddress: ethereumAddressParameterDecorator,
+    parameterTransformer: parameterTransformerMethodDecorator,
```
BMillman19: `parameterTransformer` seems to have the same method signature as `asyncZeroExErrorHandler` and `syncZeroExErrorHandler`, should we call it `parameterHandler` instead?

BMillman19:
One comment on naming inline. Other places we could possible use transformation / coercion is methods that accept `BigNumber` if a developer passes in a `number` or `string` they can be coerced into instances of `BigNumber`

dekz:
@BMillman19 I've only given that a little thought (actual type coercion).

Right now this does `string` -> `string` and is more of a formatter, something that TypeScript wouldn't pick up for us just based on types.

Would like to discuess the idea of `number` -> `BigNumber` as well (for javascript users) idea a bit more.

packages/0x.js/src/utils/decorators.ts
```diff
@@ -85,7 +91,40 @@ const syncErrorHandlerFactory = (errorTransformer: ErrorTransformer) => {
 // _.flow(f, g) = f ∘ g
 const zeroExErrorTransformer = _.flow(schemaErrorTransformer, contractCallErrorTransformer);
 
+const addParameterTransformer = (target: object, key: string | symbol, transformer: ParameterTransformer) => {
+    const param_key = `parameter_transformer_${key}`;
+    const transformableParams: ParameterTransformer[] = Reflect.getOwnMetadata(param_key, target, key) || [];
```
fabioberger: When exactly does `Reflect.getOwnMetadata` return `undefined`?

dekz:
We have agreed that a decorator in this case doesn't add anything over an explicit normalise call in the method body. Simple readable concepts is better than complex language features.

Closing this PR in favour of explicitly calling a normalise utility function.

PR #332, opened 2018-01-22
Contracts refactor. Bye, truffle!!! :clap:

LogvinovLeon:
This PR:

* Removes truffle
* Makes tests independent from each other
* Uses our deployer to migrate needed contracts in tests
* Moves a lot of types/utils around because now they're shared between 0x.js and contracts
* Prepares contracts to use abi-gen (doesn't use it yet)


abandeali1:
Is there a reason that `contracts/test/ts/utils` is different from `contracts/util`? Can we combine these?

packages/contracts/package.json
```diff
@@ -44,11 +47,12 @@
         "copyfiles": "^1.2.0",
         "dirty-chai": "^2.0.1",
         "mocha": "^4.0.1",
+        "npm-run-all": "^4.1.2",
         "solc": "^0.4.18",
         "truffle": "^4.0.1",
```
abandeali1: Forgot the most important part :)

packages/contracts/test/exchange/helpers.ts
```diff
@@ -1,52 +1,68 @@
...
 
     before(async () => {
-        const [tokenRegistry, exchange] = await Promise.all([TokenRegistry.deployed(), Exchange.deployed()]);
-        exchangeWrapper = new ExchangeWrapper(exchange);
-        const [repAddress, dgdAddress] = await Promise.all([
-            tokenRegistry.getTokenAddressBySymbol('REP'),
-            tokenRegistry.getTokenAddressBySymbol('DGD'),
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        maker = accounts[0];
+        feeRecipient = accounts[1] || accounts[accounts.length - 1];
```
fabioberger: What is this magic? In which scenarios is `accounts[1]` undefined?

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -38,49 +44,57 @@ contract('Exchange', (accounts: string[]) => {
...
-        [tokenRegistry, exchange] = await Promise.all([TokenRegistry.deployed(), Exchange.deployed()]);
-        exWrapper = new ExchangeWrapper(exchange);
-        const [repAddress, dgdAddress, zrxAddress] = await Promise.all([
-            tokenRegistry.getTokenAddressBySymbol('REP'),
-            tokenRegistry.getTokenAddressBySymbol('DGD'),
-            tokenRegistry.getTokenAddressBySymbol('ZRX'),
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        tokenOwner = maker = accounts[0];
+        taker = accounts[1] || accounts[accounts.length - 1];
+        feeRecipient = accounts[2] || accounts[accounts.length - 1];
```
fabioberger: I really dislike these. So one environment is TestRPC where we have 10 accounts. What is the second one? And should these tests work for both the case where `taker` === `feeRecipient` and when not? Sounds error prone to me. @abandeali1

packages/contracts/test/multi_sig_with_time_lock_except_remove_auth_addr.ts
```diff
@@ -0,0 +1,188 @@
...
+const PROXY_ABI = artifacts.TokenTransferProxyArtifact.networks[constants.TESTRPC_NETWORK_ID].abi;
+const MUTISIG_WALLET_WITH_TIME_LOCK_EXCEPT_REMOVE_AUTHORIZED_ADDRESS_ABI =
+    artifacts.MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddressArtifact.networks[constants.TESTRPC_NETWORK_ID]
+        .abi;
+
+chaiSetup.configure();
+const expect = chai.expect;
+const web3 = web3Factory.create();
+const web3Wrapper = new Web3Wrapper(web3.currentProvider);
+const blockchainLifecycle = new BlockchainLifecycle(devConstants.RPC_URL);
```
fabioberger: Again `devConstants.RPC_URL` is implicitly the same in `web3` instance.

packages/contracts/test/multi_sig_with_time_lock_except_remove_auth_addr.ts
```diff
@@ -0,0 +1,188 @@
...
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        owners = [accounts[0], accounts[1]];
+        authorizedAddress = `0x${crypto
+            .solSHA3([accounts[0]])
+            .slice(0, 20)
+            .toString('hex')}`;
+        unauthorizedAddress = `0x${crypto
+            .solSHA3([accounts[1]])
+            .slice(0, 20)
+            .toString('hex')}`;
```
fabioberger: These transformations should be in a helper function so that they aren't repeated. I know this is legacy code but if you could improve it, that would be awesome.

packages/contracts/test/zrx_token.ts
```diff
@@ -1,60 +1,69 @@
...
+    beforeEach(async () => {
+        await blockchainLifecycle.startAsync();
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
     describe('constants', () => {
         it('should have 18 decimals', async () => {
-            const decimals = new BigNumber(await zrx.decimals.call());
+            const decimals = new BigNumber(await zrx.decimals());
```
fabioberger: Hm, these methods on the contract wrapper don't have `async` suffixes?

packages/contracts/test/zrx_token.ts
```diff
@@ -73,8 +82,7 @@ contract('ZRXToken', (accounts: string[]) => {
             const receiver = spender;
             const initOwnerBalance = await zeroEx.token.getBalanceAsync(zrxAddress, owner);
             const amountToTransfer = new BigNumber(1);
-            const txHash = await zeroEx.token.transferAsync(zrxAddress, owner, receiver, amountToTransfer);
-            await zeroEx.awaitTransactionMinedAsync(txHash);
```
fabioberger: Why is it no longer necessary to await the tx to be mined?

packages/deployer/src/utils/contract.ts
```diff
@@ -28,16 +28,16 @@ export class Contract implements Web3.ContractInstance {
         _.forEach(functionsAbi, (functionAbi: Web3.MethodAbi) => {
             if (functionAbi.constant) {
                 const cbStyleCallFunction = this._contract[functionAbi.name].call;
-                this[functionAbi.name] = {
-                    callAsync: promisify(cbStyleCallFunction, this._contract),
```
fabioberger: Why did we get rid of `callAsync` and re-instate the use of `call`? It's still an async call no?

packages/types/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { BigNumber } from '@0xproject/utils';
+import { BigNumber } from 'bignumber.js';
```
fabioberger: Why this change?

packages/utils/src/abi_decoder.ts
```diff
@@ -1,9 +1,10 @@
-import { BigNumber } from '@0xproject/utils';
+import { AbiType, DecodedLogArgs, LogWithDecodedArgs, RawLog, SolidityTypes } from '@0xproject/types';
 import * as _ from 'lodash';
 import * as Web3 from 'web3';
 import * as SolidityCoder from 'web3/lib/solidity/coder';
 
-import { AbiType, ContractEventArgs, DecodedLogArgs, LogWithDecodedArgs, RawLog, SolidityTypes } from '../types';
+// tslint:disable-next-line:no-unused-variable
```
fabioberger: Can you add a line explaining why this tslint is being ignored here.

packages/contracts/test/exchange/wrapper.ts
abandeali1: Yeah you're right, I think a lot of these tests would probably fail if `taker` === `feeRecipient`

packages/contracts/test/ether_token.ts
```diff
@@ -1,43 +1,48 @@
...
+import { constants } from '../util/constants';
 
 import { chaiSetup } from './utils/chai_setup';
-
-const { EtherToken } = new Artifacts(artifacts);
+import { deployer } from './utils/deployer';
 
 chaiSetup.configure();
 const expect = chai.expect;
+const web3 = web3Factory.create();
```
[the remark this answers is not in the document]
LogvinovLeon: We do the same in 0x.js and I don't see how this is not future-proof. Let's not over-engineer before it's needed. I agree that that's kinda more explicit, but I'm generally against adding new params if they could as well be constants inside.

packages/contracts/test/exchange/helpers.ts
```diff
@@ -1,52 +1,68 @@
...
 import { chaiSetup } from '../utils/chai_setup';
+import { deployer } from '../utils/deployer';
 
 chaiSetup.configure();
 const expect = chai.expect;
 
-const { Exchange, TokenRegistry } = new Artifacts(artifacts);
+const web3 = web3Factory.create();
+const web3Wrapper = new Web3Wrapper(web3.currentProvider);
+const blockchainLifecycle = new BlockchainLifecycle(devConstants.RPC_URL);
```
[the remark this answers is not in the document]
LogvinovLeon: Ok, now I see. So should we remove this param here or should we add it to web3Factory?

packages/contracts/test/exchange/helpers.ts
LogvinovLeon: I wish I knew the answer. It was like that before.

packages/contracts/test/zrx_token.ts
LogvinovLeon: They don't for now, but they will with the upcomming abi-gen PR

packages/contracts/test/zrx_token.ts
LogvinovLeon: Tests run only against testrpc. On testrpc we have instant mining and that's why we await transactions only if we need a receipt.

packages/deployer/src/utils/contract.ts
LogvinovLeon: It is. Will fix that in abi-gen PR

packages/types/src/index.ts
LogvinovLeon: Because here we needed only the type and it helped to break a circular dependency.

packages/utils/src/abi_decoder.ts
LogvinovLeon: I wish I knew that. Looks like a tslint bug, cause the import is clearly not unused.

LogvinovLeon: There are some other linter false-positives that I'm investigating right now

PR #334, opened 2018-01-23
Remove Nonce Tracker to prevent faucet from becoming stuck

### LINKED ISSUE #333: Faucet stuck with invalid nonce
The Faucet can become stuck attempting to use same nonce over and over again.

```
Added 0x20c0418bf3dcbc66b461100a22852cb4abbdec0d to the ETH queue
Processing ETH 0x20c0418bf3dcbc66b461100a22852cb4abbdec0d
sending tx. from: 0xea95a7a25506c9f70fb7bc50877c435a609353b2 recipient: 0x20c0418bf3dcbc66b461100a22852cb4abbdec0d value: 100000000000000000
txData {"gasPrice":"0x4a817c800","nonce":"0x4e9c","gas":"0x5208","from":"0xea95a7a25506c9f70fb7bc50877c435a609353b2","to":"0x20c0418bf3dcbc66b461100a22852cb4abbdec0d","value":"0x16345785d8a0000"}
Unexpected err: [object Object] - {"code":-32010,"message":"Transaction nonce is too low. Try incrementing the nonce."}
```

Querying an Infura node for the latest nonce while this happens returns a valid value.

The Metamask Provider Engine `Nonce Tracker` caches the `getTransactionCount` response (converted into a nonce) until a later `sendTransaction` successfully sends. If the returned nonce isn't valid (potential syncing node, or cached an old getTransactionCount) then this value is [never cleared from the cache](https://github.com/MetaMask/provider-engine/blob/master/subproviders/nonce-tracker.js#L56).

I'll create a PR to remove this Nonce Tracker and open an issue on Provider Engine to discuss. We may end up implementing our own in `subproviders` or opening a PR on Provider Engine.

There is an issue with completely removing Nonce Tacker. Under high load we may attempt to re-use a nonce and that results in a error:
```
Unexpected err: [object Object] - {"code":-32010,"message":"Transaction with the same hash was already imported."}
```
But at least this error is recoverable as the user just has to retry again. Using the Nonce Tracker in its current form will get us stuck forever.

dekz:
fixes #333

### DIFF
--- a/packages/kovan-faucets/src/ts/handler.ts
+++ b/packages/kovan-faucets/src/ts/handler.ts
@@ -2,7 +2,6 @@ import * as express from 'express';
 import * as _ from 'lodash';
 import ProviderEngine = require('web3-provider-engine');
 import HookedWalletSubprovider = require('web3-provider-engine/subproviders/hooked-wallet');
-import NonceSubprovider = require('web3-provider-engine/subproviders/nonce-tracker');
 import RpcSubprovider = require('web3-provider-engine/subproviders/rpc');
 
 import { configs } from './configs';
@@ -76,7 +75,6 @@ export class Handler {
     // tslint:disable-next-line:prefer-function-over-method
     private _createProviderEngine(rpcUrl: string) {
         const engine = new ProviderEngine();
-        engine.addProvider(new NonceSubprovider());
         engine.addProvider(new HookedWalletSubprovider(idManagement));
         engine.addProvider(
             new RpcSubprovider({


BMillman19:
what are we losing by removing this subprovider?

dekz:
It just saves a few network calls when used in the Faucet. I'm in favour of removing it from the faucet completely. It is a useful provider for Relayers to reduce network calls and latency which is for the greater good.

packages/kovan-faucets/src/ts/handler.ts
```diff
@@ -76,7 +75,6 @@ export class Handler {
     // tslint:disable-next-line:prefer-function-over-method
     private _createProviderEngine(rpcUrl: string) {
         const engine = new ProviderEngine();
-        engine.addProvider(new NonceSubprovider());
```
fabioberger: I don't think this is the right solution to this issue. The reason we added the nonce subprovider was because we are submitting multiple txs to the network before waiting for them to confirm. Asking the network for the next nonce would return the same nonce until the txs confirm whereas this subprovider updates the nonce immediately. Have we tried this "fix" when requesting ETH multiple times in quick succession?

fabioberger: I think we need to do more investigating into why exactly the nonceTracker can end up with the wrong nonce in the first place. Is there really no way to force-clear the nonce tracker if it gets into an inconsistent state?

dekz: Yep there is more information in #333 including what happens when we attempt multiple requests in quick succession, we will get a number of errors from nonce re-use:

I.e if someone spams to the same account:

```
Unexpected err: [object Object] - {"code":-32010,"message":"Transaction with the same hash was already imported."}
```

It's difficult to reproduce the Infura Nonce value being different to what is acceptable when we sendTransaction. We are getting stuck with a cached nonce that is not valid on the next sendTransaction and it is the NonceSubprovider caching this incorrect value indefinitely. Is the test faucet really under that high of a load for the nonce optimisation?

If so, I think we either clone the Noncetracker and add it to our subproviders, maintaining the same functionality but clearing any Cached values on error, or try and get this fixed upstream. Discussion started [here](https://github.com/MetaMask/provider-engine/issues/213).

fabioberger: So, you only need to submit two requests to the faucet in less then 15 seconds in order to cause this issue to reproduce. This is not **heavy** load.

The issue here is Infura being janky. I think we can fix the issue of getting an incorrect nonce value cached by not using Infura. If we want to keep using Infura then we should move NonceTracker to subproviders.

This is my hunch, but if I find some extra time I'll try debugging to make sure.

dekz: For those following along, we're going to create a NonceTracker in the `subproviders` package. Allowing us to better handle error cases (in the occasional event of a faulty node) while continuing to reduce network usage of nonce requests.

I'll open this up in a separate PR.

fabioberger: nope.

PR #338, opened 2018-01-23
Fix getTransactionReceipt not returning null

ritave:
If the transaction is not yet put into blockchain, using `getTransactionReceipt` will return null.

### DIFF
--- a/packages/web3-typescript-typings/CHANGELOG.md
+++ b/packages/web3-typescript-typings/CHANGELOG.md
@@ -2,5 +2,6 @@
 
 ## v0.9.3 - _January 11, 2018_
 
+* Fix `getTransactionReceipt` not returning null (#338)
 * Add type for getData on a contract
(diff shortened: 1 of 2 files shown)


fabioberger:
@ritave can you also make an entry in the `CHANGELOG.md` file in the `web3-typescript-typings` package about this change? After that I will merge it!

ritave:
@fabioberger Normally I would, but after moving it into the monorepo there is no CHANGELOG file anymore!

fabioberger:
@ritave there is one, it's just nested within this sub-package's directory: https://github.com/0xProject/0x.js/blob/development/packages/web3-typescript-typings/CHANGELOG.md

ritave:
@fabioberger Ok, seems the `CHANGELOG.md` wasn't there when I was branching. Rebased on top of newest `development` and added the change

PR #339, opened 2018-01-24
Add support for all testnets to the faucet

BMillman19:
This PR:

* rename `kovan-faucets` package to `testnet-faucets`
* add query param `networkId` to the `/ether` and `/zrx` endpoints of the faucet


packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -17,62 +19,55 @@ import { ZRXRequestQueue } from './zrx_request_queue';
...
+        this._dispense(req, res, this._etherRequestQueueByNetworkId, 'ETH');
     }
     public dispenseZRX(req: express.Request, res: express.Response) {
+        this._dispense(req, res, this._zrxRequestQueueByNetworkId, 'ZRX');
+    }
+    private _dispense(
+        req: express.Request,
+        res: express.Response,
+        requestQueueByNetworkId: RequestQueueByNetworkId,
+        assetSymbol: string,
```
fabioberger: Since this can only be `ZRX` or `ETH`, let's just make it a string enum with these two values only.

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -17,62 +19,55 @@ import { ZRXRequestQueue } from './zrx_request_queue';
...
+    ) {
         const recipientAddress = req.params.recipient;
         if (_.isUndefined(recipientAddress) || !this._isValidEthereumAddress(recipientAddress)) {
-            res.status(400).send('INVALID_REQUEST');
+            res.status(400).send('INVALID_RECIPIENT_ADDRESS');
+            return;
+        }
+        const networkId = _.get(req.query, 'networkId', DEFAULT_NETWORK_ID);
```
fabioberger: What if this is `undefined`? It just makes the next `_.get` return undefined? If so, this is relying on a behavior of `_.get` that isn't explicitly mentioned in their docs. Maybe we should explicitly check for `undefined`?

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -17,62 +19,55 @@ import { ZRXRequestQueue } from './zrx_request_queue';
...
-                full: this._etherRequestQueue.isFull(),
-                size: this._etherRequestQueue.size(),
-            },
-            zrx: {
-                full: this._zrxRequestQueue.isFull(),
-                size: this._zrxRequestQueue.size(),
-            },
-        });
-        res.status(200).send(payload);
-    }
```
fabioberger: Hm, I think this endpoint could still be useful. Should we re-instate it? Not sure why it was considered deprecated and removed.

packages/testnet-faucets/src/ts/request_queue.ts
```diff
@@ -51,6 +51,6 @@ export class RequestQueue {
     }
     // tslint:disable-next-line:prefer-function-over-method
     protected async processNextRequestFireAndForgetAsync(recipientAddress: string) {
-        throw new Error('Expected processNextRequestFireAndForgetAsync to be implemented by a superclass');
+        throw new Error('Expected processNextRequestFireAndForgetAsync to be implemented by a subclass');
```
fabioberger: Good catch!

packages/testnet-faucets/src/ts/handler.ts
BMillman19: It's my understanding that it is undefined wit will just return DEFAULT_NETWORK_ID which is defined as 42 in this case

BMillman19: because the next `_.get()` call does not have a default return value it may be `undefined` which we check in the next line before returning `INVALID_NETWORK_ID`

packages/testnet-faucets/src/ts/handler.ts
BMillman19: sure!

packages/testnet-faucets/src/ts/handler.ts
fabioberger: Ah yes, you are right!

PR #346, opened 2018-01-26
Added CLI options for explicit specifying location of partials and main template

### LINKED ISSUE #344: Abi-Gen looks for wrongly named partials
Partials used in abi-gen need to have the `.mustache` suffix to work, while at the same time abi-gen expect Handlebar templates. This confuses users as well as IDEs and stops them from coloring properly.

ritave:
This PR makes the CLI more explicit in specifying the main template and the partial templates. Previously the process was very hidden and not understandable without reading source-code. This PR helps and CLI can now be understood without guessing.

The previous options still exists for backwards compatibility but is now deprecated and should be removed when not needed to ease up on logic

Fixes #344


LogvinovLeon:
Will accept and publish after that. LGTM

ritave:
@LogvinovLeon done

PR #349, opened 2018-01-30
Add shouldAddPersonalMessagePrefix param to signOrderHashAsync

fabioberger:
This PR:

- Adds the `shouldAddPersonalMessagePrefix` parameter to `signOrderHashAsync` so that the caller can decide on whether to add the personalMessage prefix before relaying the request to the signer. Parity Signer, Ledger and TestRPC add the prefix themselves, Metamask expects it to have already been added. Previously we would try and infer whether to include the prefix by looking at the nodeVersion, however this does not work for signers that are not also nodes (e.g Ledger).


packages/0x.js/CHANGELOG.md
```diff
@@ -1,10 +1,21 @@
 # CHANGELOG
 
-## v0.30.1 - _TBD, 2018_
+## v0.31.0 - _January 30, 2018_
+
+    * Add the `shouldAddPersonalMessagePrefix` parameter to `signOrderHashAsync` so that the
```
LogvinovLeon: PR number

PR #350, opened 2018-01-30
Add zeroEx.etherToken.getContractAddressIfExists

LogvinovLeon:
## Description

Add zeroEx.etherToken.getContractAddressIfExists

## Motivation and Context

N/A

## How Has This Been Tested?

By tests

## Types of changes

## Checklist:


packages/0x.js/src/contract_wrappers/ether_token_wrapper.ts
```diff
@@ -163,6 +163,15 @@ export class EtherTokenWrapper extends ContractWrapper {
     public unsubscribeAll(): void {
         super.unsubscribeAll();
     }
+    /**
+     * Retrieves the Ethereum address of the EtherToken contract deployed on the network
+     * that the user-passed web3 provider is connected to. If it's an unknown private network - undefined is returned.
+     * @returns The Ethereum address of the EtherToken contract or undefined.
+     */
+    public getContractAddressIfExists(): string | undefined {
```
abandeali1: I know that `getContractAddressIfExists` is more descriptive than `getContractAddress` in this case, but I also feel like the name should be standardized for all of our wrappers. Thoughts?

packages/0x.js/src/contract_wrappers/ether_token_wrapper.ts
```diff
@@ -163,6 +163,15 @@ export class EtherTokenWrapper extends ContractWrapper {
...
+        const contractAddress = artifacts.EtherTokenArtifact.networks[this._networkId].address;
```
fabioberger: Won't `artifacts.EtherTokenArtifact.networks[this._networkId]` return undefined if the `networkId` is not in the `networks` object? Then we are looking for `address` property on undefined and this will throw. This needs to be fixed. Let's add a regression test for this.

packages/0x.js/src/contract_wrappers/ether_token_wrapper.ts
```diff
@@ -163,6 +163,15 @@ export class EtherTokenWrapper extends ContractWrapper {
     public unsubscribeAll(): void {
         super.unsubscribeAll();
     }
+    /**
+     * Retrieves the Ethereum address of the EtherToken contract deployed on the network
+     * that the user-passed web3 provider is connected to. If it's an unknown private network - undefined is returned.
```
fabioberger: How about instead of `an unknown private network`, we said `it's not Kovan, Ropsten, Rinkeby, Mainnet or TestRPC (networkId: 50), it will return undefined (e.g a private network).`. This is more explicit.

packages/0x.js/src/contract_wrappers/ether_token_wrapper.ts
fabioberger: I prefer descriptive over standardised. It makes this much more readable for the next developer.

packages/0x.js/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@
 
     * Add `zeroEx.etherToken.getContractAddressIfExists` (#350)
 
+## v0.31.1 - _February 1, 2018_
+
+    * Fix the bug causing order watcher to throw is makerToken === zrx (#357)
```
fabioberger: Typo: `is` -> `if`

packages/0x.js/src/contract_wrappers/ether_token_wrapper.ts
fabioberger: Was this addressed?

PR #351, opened 2018-01-30
Portal Ledger Support, Lazy-loading token balances/allowances

fabioberger:
This PR:
- Adds Ledger support to Portal 🥇 
- Adds lazy-loading of token balances/allowances, making Portal loading times faster and reducing internal state dependencies.
- Removes the bulky loading animation and replaces it with a simple circular spinner (it doesn't show for too long now 💃 )
- Add Rinkeby and Ropsten support to Portal
- Fetches a reasonable default gasPrice for transactions from EthGasStation
- Fixed a bug so Portal properly detects when users reject a transaction request on Metamask, Parity Signer and Ledger
- Changes dev domain to `0xproject.localhost` since `.dev` is owned by Google and chrome complains about missing certs.
- Fix bug where negative amounts would cause our input fields to break
- Fix setState on unmounted component race conditions across the entire website

screen shot 2018-01-30 at 20 23 09

screen shot 2018-01-30 at 20 25 07


LogvinovLeon:
Can we use .local?

packages/website/package.json
```diff
@@ -8,9 +8,12 @@
         "clean": "shx rm -f public/bundle*",
         "lint": "tslint --project . 'ts/**/*.ts' 'ts/**/*.tsx'",
         "dev": "webpack-dev-server --content-base public --https",
-        "update_contracts": "for i in ${npm_package_config_artifacts}; do copyfiles -u 4 ../contracts/build/contracts/$i.json ../website/contracts; done;",
-        "deploy_staging": "npm run build; aws s3 sync ./public/. s3://staging-0xproject --profile 0xproject --region us-east-1 --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers",
-        "deploy_live": "npm run build; aws s3 sync ./public/. s3://0xproject.com --profile 0xproject --region us-east-1 --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers"
+        "update_contracts":
```
LogvinovLeon: Please undo those changes

packages/website/ts/blockchain.ts
```diff
@@ -121,6 +123,9 @@ export class Blockchain {
     constructor(dispatcher: Dispatcher, isSalePage: boolean = false) {
         this._dispatcher = dispatcher;
         this._userAddress = '';
+        this._defaultGasPrice = new BigNumber(30000000000);
```
LogvinovLeon: Can we use ZeroEx.toBaseUnitAmount? It's unclear if that's 3 Gwei or 0.3 Gwei or 30 GWei

packages/website/ts/blockchain.ts
```diff
@@ -121,6 +123,9 @@ export class Blockchain {
...
+        // tslint:disable-next-line:no-floating-promises
+        this._updateDefaultGasPriceAsync();
```
LogvinovLeon: FireAndForget?

packages/website/ts/components/dialogs/ledger_config_dialog.tsx
```diff
@@ -184,6 +202,8 @@ export class LedgerConfigDialog extends React.Component<LedgerConfigDialogProps,
         const selectAddressBalance = this.state.addressBalances[selectedRowIndex];
         this.props.dispatcher.updateUserAddress(selectedAddress);
         this.props.blockchain.updateWeb3WrapperPrevUserAddress(selectedAddress);
+        // tslint:disable-next-line:no-floating-promises
+        this.props.blockchain.fetchTokenInformationAsync();
```
LogvinovLeon: FireAndForget?

LogvinovLeon:
image

packages/website/ts/blockchain.ts
fabioberger: `toBaseUnitAmount` goes from unit amount to baseUnit, i.e Ether to Wei. If we go from gwei to wei, it's also weird because the assumption is we are passing in the unit amount, not some other intermediary unit. I'll use a named constant to make this more readable.

Something like:

```
3 * GWEI_IN_WEI
```

PR #352, opened 2018-01-30
Add order signing and dispensing ability to faucet

BMillman19:
## Description

- Add new endpoints to the testnet faucet that dispense orders signed by the dispenser address
- Extract common parameter extraction and sanitizing into middleware

## Motivation and Context

- An easy source of ZRX/WETH liquidity on testnets

## How Has This Been Tested?

- Manual testing in dev environment

## Types of changes

## Checklist:


packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -62,37 +77,83 @@ export class Handler {
...
     }
+    private async _dispenseOrder(req: express.Request, res: express.Response, requestedAssetType: RequestedAssetType) {
+        const zeroEx = _.get(this._zeroExByNetworkId, req.networkId);
+        if (_.isUndefined(zeroEx)) {
+            res.status(400).send('UNSUPPORTED_NETWORK_ID');
+            return;
+        }
+        res.setHeader('Content-Type', 'application/json');
+        const makerTokenAddress = await zeroEx.tokenRegistry.getTokenAddressBySymbolIfExistsAsync(requestedAssetType);
+        if (_.isUndefined(makerTokenAddress)) {
```
fabioberger: I feel like this conditional will only ever be hit if the developer messed up. For this reason, I think this should probably throw an error rather then returning one to the caller. Thoughts?

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -62,37 +77,83 @@ export class Handler {
...
+            res.status(400).send('INVALID_TOKEN');
+            return;
+        }
+        const takerTokenSymbol =
+            requestedAssetType === RequestedAssetType.WETH ? RequestedAssetType.ZRX : RequestedAssetType.WETH;
+        const takerTokenAddress = await zeroEx.tokenRegistry.getTokenAddressBySymbolIfExistsAsync(takerTokenSymbol);
+        if (_.isUndefined(takerTokenAddress)) {
```
fabioberger: Ditto above.

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -62,37 +77,83 @@ export class Handler {
...
+            makerFee: new BigNumber(0),
+            takerFee: new BigNumber(0),
+            makerTokenAmount,
+            takerTokenAmount,
+            makerTokenAddress,
+            takerTokenAddress,
+            salt: ZeroEx.generatePseudoRandomSalt(),
+            exchangeContractAddress: zeroEx.exchange.getContractAddress(),
+            feeRecipient: ZeroEx.NULL_ADDRESS,
+            expirationUnixTimestampSec: new BigNumber(Date.now() + TWO_DAYS_IN_MS),
```
fabioberger: Let's make this longer. Perhaps 5 days? Otherwise these might expire by the end of the hackathon, right around demo time 🌮

packages/testnet-faucets/src/ts/id_management.ts
```diff
@@ -18,4 +20,16 @@ export const idManagement = {
         const rawTx = `0x${tx.serialize().toString('hex')}`;
         callback(null, rawTx);
     },
+    signMessage(message: object, callback: Callback) {
+        const data = _.get(message, 'data');
```
fabioberger: Rename to `dataIfExists`

packages/testnet-faucets/src/ts/id_management.ts
```diff
@@ -18,4 +20,16 @@ export const idManagement = {
...
+        if (_.isUndefined(data)) {
+            callback(new Error('No data to sign'), null);
```
fabioberger: Let's make this `NO_DATA_TO_SIGN` so it's easier for the caller to check for it.

packages/testnet-faucets/src/ts/parameter_extractor.ts
```diff
@@ -0,0 +1,34 @@
...
+            res.status(400).send('UNSUPPORTED_NETWORK_ID');
+            return;
+        }
+        req.networkId = networkId;
+        next();
+    },
+};
+
+function _isValidEthereumAddress(address: string): boolean {
+    const lowercaseAddress = address.toLowerCase();
```
fabioberger: Why do we need to lowercase the address here?

packages/testnet-faucets/src/ts/parameter_extractor.ts
```diff
@@ -0,0 +1,34 @@
...
+    extract(req: Request, res: Response, next: NextFunction) {
+        const recipientAddress = req.params.recipient;
+        if (_.isUndefined(recipientAddress) || !_isValidEthereumAddress(recipientAddress)) {
+            res.status(400).send('INVALID_RECIPIENT_ADDRESS');
+            return;
+        }
+        const lowerCaseRecipientAddress = recipientAddress.toLowerCase();
+        req.recipientAddress = lowerCaseRecipientAddress;
+        const networkId = _.get(req.query, 'networkId', DEFAULT_NETWORK_ID);
+        const rpcUrl = _.get(rpcUrls, networkId);
```
fabioberger: Rename to `rpcUrlIfExists`

packages/testnet-faucets/src/ts/parameter_extractor.ts
```diff
@@ -0,0 +1,34 @@
...
+        if (_.isUndefined(rpcUrl)) {
+            res.status(400).send('UNSUPPORTED_NETWORK_ID');
+            return;
+        }
+        req.networkId = networkId;
```
fabioberger: Why not just re-assign these to `req.params.`? That way we don't need to add these parameters to the `Express.Request` type which feels hacky.

packages/testnet-faucets/src/ts/server.ts
```diff
@@ -20,8 +21,10 @@ app.get('/ping', (req: express.Request, res: express.Response) => {
     res.status(200).send('pong');
 });
 app.get('/info', handler.getQueueInfo.bind(handler));
-app.get('/ether/:recipient', handler.dispenseEther.bind(handler));
-app.get('/zrx/:recipient', handler.dispenseZRX.bind(handler));
+app.get('/ether/:recipient', parameterExtractor.extract, handler.dispenseEther.bind(handler));
```
fabioberger: Rename `parameterExtractor` to `parameterTransformer`

packages/testnet-faucets/tsconfig.json
```diff
@@ -3,5 +3,10 @@
     "compilerOptions": {
         "outDir": "lib"
     },
-    "include": ["./src/ts/**/*", "../../node_modules/web3-typescript-typings/index.d.ts"]
+    "include": [
+        "./src/ts/**/*",
+        "../../node_modules/types-bn/index.d.ts",
```
fabioberger: Why are we adding this?

packages/testnet-faucets/package.json
```diff
@@ -35,6 +36,8 @@
         "shx": "^0.2.2",
         "source-map-loader": "^0.1.6",
         "tslint": "5.8.0",
+        "types-bn": "^0.0.1",
```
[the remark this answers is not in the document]
BMillman19: I was getting compilation errors before but they seem to be resolved on the latest `development` branch

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -19,32 +28,38 @@ import { ZRXRequestQueue } from './zrx_request_queue';
...
 
-enum QueueType {
+enum RequestedAssetType {
     ETH = 'ETH',
+    WETH = 'WETH',
     ZRX = 'ZRX',
 }
 
-const DEFAULT_NETWORK_ID = 42; // kovan
+const TWO_DAYS_IN_MS = 1.728e8; // TODO: make this configurable
```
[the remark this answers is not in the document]
BMillman19: Not in this PR

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -19,32 +28,38 @@ import { ZRXRequestQueue } from './zrx_request_queue';
...
 
 export class Handler {
-    private _etherRequestQueueByNetworkId: RequestQueueByNetworkId = {};
-    private _zrxRequestQueueByNetworkId: RequestQueueByNetworkId = {};
+    private _zeroExByNetworkId: ItemByNetworkId<ZeroEx> = {};
+    private _etherRequestQueueByNetworkId: ItemByNetworkId<RequestQueue> = {};
+    private _zrxRequestQueueByNetworkId: ItemByNetworkId<RequestQueue> = {};
     constructor() {
         _.forIn(rpcUrls, (rpcUrl: string, networkId: string) => {
```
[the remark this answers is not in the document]
BMillman19: even if I define `rpcUrls` as an object with numbers as keys, they get converted into strings

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -62,37 +77,83 @@ export class Handler {
...
+        this._dispenseAsset(req, res, this._zrxRequestQueueByNetworkId, RequestedAssetType.ZRX);
+    }
+    public async dispenseWETHOrder(req: express.Request, res: express.Response) {
+        await this._dispenseOrder(req, res, RequestedAssetType.WETH);
     }
-    private _dispense(
+    public async dispenseZRXOrder(req: express.Request, res: express.Response, next: express.NextFunction) {
+        await this._dispenseOrder(req, res, RequestedAssetType.ZRX);
+    }
+    // tslint:disable-next-line:prefer-function-over-method
```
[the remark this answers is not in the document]
BMillman19: I'm using the style/convention already used in this file for `_createProviderEngine` below but I can update both to be private static, I think it makes more sense that way anyway so it is more explicit that the function does not have a dependency on `this`

packages/testnet-faucets/src/ts/handler.ts
BMillman19: sure, makes sense

packages/testnet-faucets/src/ts/global.d.ts
```diff
@@ -24,3 +24,13 @@ declare module 'ethereumjs-tx' {
     }
     export = EthereumTx;
 }
+
+// Define extra params on Request for parameter extraction in middleware
+/* tslint:disable */
```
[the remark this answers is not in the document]
BMillman19: If we are not setting new fields on the Express.Request object we don't need this anymore

packages/testnet-faucets/tsconfig.json
BMillman19: Same as above comment about bn types, was getting a compilation / build error before about missing bn types but looks like it is resolved now

BMillman19:
I updated our `types-ethereumjs-util` fork but still couldn't get around the types-bn dependency, everything else should be good to go tho

packages/testnet-faucets/tsconfig.json
fabioberger: So is this still needed? @BMillman19

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -1,50 +1,92 @@
...
+        engine.start();
+        return engine;
+    }
     constructor() {
         _.forIn(rpcUrls, (rpcUrl: string, networkId: string) => {
-            const providerObj = this._createProviderEngine(rpcUrl);
+            const providerObj = Handler._createProviderEngine(rpcUrl);
             const web3 = new Web3(providerObj);
+            const zeroExConfig = {
+                networkId: +networkId,
```
dekz: Are these networkIds numbers now in rpc_urls? The +networkId threw me for a bit.

BMillman19: yea although they are written as numbers in the literal representation they are converted to string under the hood because JavaScript, I'll change the literal representation back to strings so its not confusing

BMillman19:
looks like we have an implicit dependency on `types-bn` from 0x.js as well. i.e. we include the types in the `tsconfig.json` but they are not specified as dependencies in the `package.json`. The only reason this is working is because at least one of our packages in the monorepo is specifying `types-bn` as a dependency and it is being hoisted to the top level `node_modules` directory

PR #355, opened 2018-01-31
Nonce tracker subprovider

dekz:
## Description

Caches the nonce when a request to getTransactionCount is made
and increments the pending nonce after successful transactions.

If a nonce too low error occurs then clear the cache and perform a clean request for Transaction count.

TODO:
- [x] Attempt to reproduce with multiple instances running and interleaving faucet requests

## Motivation and Context

#333 

On Kovan our faucet can get stuck when using the NonceTracker in Provider Engine.

## How Has This Been Tested?

Used a fake Static Provider which tracked state and on subsequent calls would return a different value. If this value was cached it would never change unless a tx was made.

Also caused a `Nonce Too Low` error when calling sendRawTransaction and check the subsequent nonce value has changed.

## Types of changes


packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+import { JSONRPCPayload } from '../types';
+
+import { Subprovider } from './subprovider';
+
+const NONCE_TOO_LOW_ERROR_MESSAGE = 'Transaction nonce is too low';
+
+export class NonceTrackerSubprovider extends Subprovider {
+    private _nonceCache: { [address: string]: string } = {};
+    private static _reconstructTransaction(payload: JSONRPCPayload): EthereumTx {
+        const raw = payload.params[0];
```
BMillman19: what if params is empty?

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+        const transactionData = ethUtil.stripHexPrefix(raw);
```
BMillman19: I think we can reduce these two lines to one `ethUtil.toBuffer(raw)` https://github.com/ethereumjs/ethereumjs-util/blob/0bc0616c2de0de0723f87c3b64923bb9f962a8e1/index.js#L153-L175

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+    private static _determineAddress(payload: JSONRPCPayload): string {
+        switch (payload.method) {
+            case 'eth_getTransactionCount':
+                return payload.params[0].toLowerCase();
+            case 'eth_sendRawTransaction':
+                const transaction = NonceTrackerSubprovider._reconstructTransaction(payload);
+                return `0x${transaction.getSenderAddress().toString('hex')}`.toLowerCase();
+            default:
+                throw new Error('Invalid Method');
```
BMillman19: Let's include the `payload.method` in the error message

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+    constructor() {
```
BMillman19: lets get rid of this

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+        super();
+    }
+    // tslint:disable-next-line:async-suffix
+    public async handleRequest(
+        payload: JSONRPCPayload,
+        next: (callback?: (err: Error | null, result: any, cb: any) => void) => void,
```
BMillman19: we can pull out some of these function signatures as type declarations at the top of the file for readability

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+        end: (err: Error | null, data?: any) => void,
+    ): Promise<void> {
+        switch (payload.method) {
+            case 'eth_getTransactionCount':
+                const blockTag = providerEngineUtils.blockTagForPayload(payload);
+                if (!_.isNull(blockTag) && blockTag === 'pending') {
+                    const address = NonceTrackerSubprovider._determineAddress(payload);
+                    const cachedResult = this._nonceCache[address];
+                    if (cachedResult) {
```
BMillman19: `if(!_.isUndefined(cachedResult)` ?

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,107 @@
...
+                const transaction = NonceTrackerSubprovider._reconstructTransaction(payload);
+                return `0x${transaction.getSenderAddress().toString('hex')}`.toLowerCase();
+            default:
+                throw new Error('Invalid Method');
+        }
+    }
+    constructor() {
+        super();
+    }
+    // tslint:disable-next-line:async-suffix
```
BMillman19: how come we are disabling the linter here?

dekz: It's an async method without an async suffix. This is not our interface but our linting rules require this.

packages/subproviders/src/subproviders/nonce_tracker.ts
dekz: nice!

packages/subproviders/src/subproviders/nonce_tracker.ts
dekz: I do not believe this is possible as it would pass through sendTransaction validation before sendRawTransaction. This would imply an upstream provider has messed up their payload. 

Better to be safe than buggy though!

dekz:
@fabioberger getting a lint error on `globals.d.ts`

```
/src/ts/global.d.ts[25, 5]: A maximum of 1 class per file is allowed.
```

What am I missing here?

packages/testnet-faucets/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM node
 
 WORKDIR /src
 
+RUN apt-get -qq update && apt-get install -y libhidapi-dev libusb-1.0-0-dev
```
fabioberger: What is this for? Let's add a comment above it.

packages/testnet-faucets/Dockerfile
```diff
@@ -2,6 +2,8 @@ FROM node
...
+ENV CIRCLECI=true
```
fabioberger: Is this because of the `node_hid` issue?

packages/testnet-faucets/src/ts/global.d.ts
```diff
@@ -24,3 +20,21 @@ declare module 'ethereumjs-tx' {
...
+        public start(): void;
+        public stop(): void;
+    }
+    export = Web3ProviderEngine;
+}
+/* tslint:enable */
+
+declare module 'web3-provider-engine/subproviders/rpc';
+declare module 'web3-provider-engine/subproviders/nonce-tracker';
+declare module 'web3-provider-engine/subproviders/hooked-wallet';
```
fabioberger: Let's leave any `declare module` at the top of the file.

packages/subproviders/test/unit/nonce_tracker_subprovider_test.ts
```diff
@@ -0,0 +1,198 @@
...
+        const secondResponse = await promisify(provider.sendAsync, provider)(payload);
+        expect(secondResponse.result).to.be.eq('0x01');
+    });
+    it('does not cache the result for latest transaction count', async () => {
+        provider = new Web3ProviderEngine();
+        let called = false;
+        const nonceTrackerSubprovider = new NonceTrackerSubprovider();
+        provider.addProvider(nonceTrackerSubprovider);
+        provider.addProvider(
+            new FixtureSubprovider({
```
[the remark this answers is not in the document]
dekz: yes there is and it looks mucho better!

packages/testnet-faucets/Dockerfile
dekz: yep, added a comment for this.

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,97 @@
...
+    BlockParamLiteral,
+    ErrorCallback,
+    JSONRPCPayload,
+    NonceSubproviderErrors,
+    OptionalNextCallback,
+} from '../types';
+
+import { Subprovider } from './subprovider';
+
+const NONCE_TOO_LOW_ERROR_MESSAGE = 'Transaction nonce is too low';
```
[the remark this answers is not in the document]
dekz: Yeah let me quickly think this over, the reason I didn't move it is because its not an Error created by the nonce tracker, just a downstream error we'd like to handle.

packages/subproviders/test/unit/nonce_tracker_subprovider_test.ts
```diff
@@ -0,0 +1,151 @@
...
+    });
+    it('clears the cache on a Nonce Too Low Error', async () => {
+        provider = new Web3ProviderEngine();
+        const nonceTrackerSubprovider = new NonceTrackerSubprovider();
+        provider.addProvider(nonceTrackerSubprovider);
+        provider.addProvider(createFixtureSubprovider());
+        provider.addProvider(
+            new FixtureSubprovider({
+                eth_sendRawTransaction: (data: any, next: any, end: any) => {
+                    end(new Error('Transaction nonce is too low'));
```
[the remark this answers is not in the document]
dekz: Added a comment about why we don't export this due to it not being our error, and we can't re-throw

packages/subproviders/src/subproviders/nonce_tracker.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { BlockParamLiteral } from '@0xproject/types';
+
+import { ErrorCallback, JSONRPCPayload, NonceSubproviderErrors, OptionalNextCallback } from '../types';
+
+import { Subprovider } from './subprovider';
+
+// We do not export this since this is not our error, and we do not throw this error
+const NONCE_TOO_LOW_ERROR_MESSAGE = 'Transaction nonce is too low';
+export class NonceTrackerSubprovider extends Subprovider {
```
fabioberger: Let's add a comment above this class explaining that this was heavily inspired by the ProviderEngine `NonceSubprovider` but with the added feature that the cached nonce is cleared if it ever becomes invalid.

PR #357, opened 2018-02-01
BUGFIX: don't remove maker/zrx order from cache twice

lukeautry:
When removing an order from the watcher cache, we call _removeFromDependentOrderHashes for both the makerTokenAddress and the ZRX token address for fees. The problem is when the makerTokenAddress is the ZRX token address and you call this line:

```
this._dependentOrderHashes[makerAddress][tokenAddress].delete(orderHash);
```

This fails and results in the rest of the script terminating.

### DIFF
--- a/packages/0x.js/src/order_watcher/order_state_watcher.ts
+++ b/packages/0x.js/src/order_watcher/order_state_watcher.ts
@@ -133,8 +133,12 @@ export class OrderStateWatcher {
         delete this._orderStateByOrderHashCache[orderHash];
         const exchange = (this._orderFilledCancelledLazyStore as any)._exchange as ExchangeWrapper;
         const zrxTokenAddress = exchange.getZRXTokenAddress();
+        
         this._removeFromDependentOrderHashes(signedOrder.maker, zrxTokenAddress, orderHash);
-        this._removeFromDependentOrderHashes(signedOrder.maker, signedOrder.makerTokenAddress, orderHash);
+        if (zrxTokenAddress !== signedOrder.makerTokenAddress) {
+            this.removeFromDependentOrderHashes(signedOrder.maker, signedOrder.makerTokenAddress, orderHash);
+        }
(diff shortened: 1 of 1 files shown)


LogvinovLeon:
According to the Set docs Set.delete() just returns false and doesn't throw. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/delete
screen shot 2018-02-01 at 12 55 01

fabioberger:
@LogvinovLeon the issue is that you are trying to call `delete` on undefined.

fabioberger:
This LGTM. Do we make this same mistake elsewhere?

packages/0x.js/src/order_watcher/order_state_watcher.ts
```diff
@@ -133,8 +133,12 @@ export class OrderStateWatcher {
         delete this._orderStateByOrderHashCache[orderHash];
         const exchange = (this._orderFilledCancelledLazyStore as any)._exchange as ExchangeWrapper;
         const zrxTokenAddress = exchange.getZRXTokenAddress();
+        
         this._removeFromDependentOrderHashes(signedOrder.maker, zrxTokenAddress, orderHash);
-        this._removeFromDependentOrderHashes(signedOrder.maker, signedOrder.makerTokenAddress, orderHash);
+        if (zrxTokenAddress !== signedOrder.makerTokenAddress) {
+            this.removeFromDependentOrderHashes(signedOrder.maker, signedOrder.makerTokenAddress, orderHash);
```
BMillman19: theres an underscore missing here

lukeautry: How did this compile 😕

LogvinovLeon: It didn't. But I changed that before publishing

PR #358, opened 2018-02-01
Add build:watch command to all TS packages

LogvinovLeon:
## Description

This PR adds a build:watch command to all TS packages so that you can recompile/watch all packages by just running: `lerna --parallel build:watch`

## Motivation and Context

Dev velocity

## How Has This Been Tested?

On my computer

## Types of changes

## Checklist:


dekz:
👍 

Can we throw `lerna --parallel build:watch` in the main package.json scripts so a dev can just run `yarn run lerna:watch` for example?

BMillman19:
If we make a change in 0x.js will this automatically watch and rebuild dependent packages, like testnet-faucets for example?

fabioberger:
Same q as @BMillman19.

LogvinovLeon:
@BMillman19 @fabioberger Yes. It's not idead, cause sometimes building 0x.js involves multiple changes and it cascades multiple time, but after all - yes.

PR #359, opened 2018-02-01
TS 2.7

LogvinovLeon:
## Description

Upgrade to TS 2.7.1

## How Has This Been Tested?

I've built the monorepo

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/packages/0x.js/src/contract_wrappers/contract_wrapper.ts
+++ b/packages/0x.js/src/contract_wrappers/contract_wrapper.ts
@@ -34,8 +34,8 @@ export class ContractWrapper {
     protected _web3Wrapper: Web3Wrapper;
     private _networkId: number;
     private _abiDecoder?: AbiDecoder;
-    private _blockAndLogStreamerIfExists: BlockAndLogStreamer | undefined;
-    private _blockAndLogStreamInterval: NodeJS.Timer;
+    private _blockAndLogStreamerIfExists?: BlockAndLogStreamer;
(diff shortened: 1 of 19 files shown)


packages/0x.js/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -191,7 +191,7 @@ export class ContractWrapper {
         }
         this._blockAndLogStreamerIfExists.unsubscribeFromOnLogAdded(this._onLogAddedSubscriptionToken as string);
         this._blockAndLogStreamerIfExists.unsubscribeFromOnLogRemoved(this._onLogRemovedSubscriptionToken as string);
-        intervalUtils.clearAsyncExcludingInterval(this._blockAndLogStreamInterval);
+        intervalUtils.clearAsyncExcludingInterval(this._blockAndLogStreamIntervalIfExists as NodeJS.Timer);
```
BMillman19: we should just check for undefined here instead of casting

BMillman19:
Oh and also did we check if the auto generated classes generated from `contract_templates` are ok post-upgrade?

packages/0x.js/src/contract_wrappers/contract_wrapper.ts
LogvinovLeon: We shouldn't because we're sure that this is not undefined at that moment. It's checked above. It's just the Type system that's not expressive enough to prove it.

PR #362, opened 2018-02-02
Arbitrage contract with tests

LogvinovLeon:
## Description

This PR adds tha initial implementation of Arbitrage smart contract that shows how to use 0x Exchange contract from the smart-contract level.
it implements the atomic arbitrage between ED and 0x. You either make both trades or none.
This PR also contains tests for this.

## Motivation and Context

We want to have better smart contracts docs before Eth Denver

## How Has This Been Tested?

yarn test

## Types of changes

## Checklist:


packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    uint constant MAX_UINT = 2**256 - 1;
+
+    function Arbitrage(address _exchangeAddress, address _etherDeltaAddress, address _proxyAddress) {
+        exchange = Exchange(_exchangeAddress);
+        etherDelta = EtherDelta(_etherDeltaAddress);
+        proxyAddress = _proxyAddress;
+    }
+
+    function setAllowances(address tokenAddress) public onlyOwner {
```
abandeali1: Can we add some comments on when to set an allowance? It's also not obvious that you need to deposit/withdraw from the Arbitrage contract (and how to do that). While you can technically do this manually by setting an allowance and then calling `transfer` and `transferFrom` directly on the token, it would be more intuitive to have explicit `withdraw` and `deposit` methods.

Also, let's make this `external`.

packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+     * 10 nonce
+     * 11 amount
+
+     * signature
+     * exchange then etherDelta
+     */
+    function makeAtomicTrade(
+        address[8] addresses, uint[12] values,
+        uint8[2] v, bytes32[2] r, bytes32[2] s
+    ) public onlyOwner {
```
abandeali1: Let's make this `external` (saves on gas).

packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+import "./EtherDelta.sol";
+import "../tokens/WETH9.sol";
+
+contract Arbitrage is Ownable {
+
+    Exchange exchange;
+    EtherDelta etherDelta;
+    address proxyAddress;
+
+    uint constant MAX_UINT = 2**256 - 1;
```
abandeali1: I know our other contracts use `uint` instead of `uint256`, but I prefer the latter going forward to make things more explicit (applies to all of contract).

packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    function setAllowances(address tokenAddress) public onlyOwner {
+        Token token = Token(tokenAddress);
+        token.approve(address(etherDelta), MAX_UINT);
+        token.approve(proxyAddress, MAX_UINT);
+        token.approve(owner, MAX_UINT);
+    }
+
+    /*
+     * I 愛 the limitations on Solidity stack size!
```
fabioberger: @Recmo lol

packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+     * 11 amount
+
+     * signature
+     * exchange then etherDelta
+     */
+    function makeAtomicTrade(
+        address[8] addresses, uint[12] values,
+        uint8[2] v, bytes32[2] r, bytes32[2] s
+    ) public onlyOwner {
+        makeExchangeTrade(addresses, values, v, r, s);
```
fabioberger: It's kinda janky that we pass all args into these two methods. Maybe we should extract out the relevant parts into a new variable?

packages/contracts/contracts/tutorials/EtherDelta.sol
```diff
@@ -0,0 +1,168 @@
+pragma solidity ^0.4.19;
+
+import "../utils/SafeMath.sol";
+import "./AccountLevels.sol";
```
fabioberger: How about putting all EtherDelta contracts in a sub-folder? Just so it's clear what is and is not part of it.

packages/contracts/test/tutorials/arbitrage.ts
```diff
@@ -0,0 +1,211 @@
...
+        await weth.setBalance(edMaker, ZeroEx.toBaseUnitAmount(new BigNumber(2), 18), { from: coinbase });
+        // Set the allowance for the maker on EtherDelta side
+        await weth.approve(etherDelta.address, INITIAL_ALLOWANCE, { from: edMaker });
+        // Deposit maker funds into EtherDelta
+        await etherDelta.depositToken(weth.address, amountGive, { from: edMaker });
+
+        const amountGet = ZeroEx.toBaseUnitAmount(new BigNumber(1), 18);
+        // Fund the front runner on EtherDelta side
+        await zrx.setBalance(edFrontRunner, amountGet, { from: coinbase });
+        // Set the allowance for the maker on EtherDelta side
```
fabioberger: This comment is inaccurate. `maker` -> `front runner`

packages/contracts/test/tutorials/arbitrage.ts
```diff
@@ -0,0 +1,211 @@
...
+        let tokenGet: string;
+        let tokenGive: string;
+        let amountGet: BigNumber;
+        let amountGive: BigNumber;
+        let expires: BigNumber;
+        let nonce: BigNumber;
+        let edSignature: ECSignature;
+        before(async () => {
+            order = await orderFactory.newSignedOrderAsync();
+            const shouldAddPersonalMessagePrefix = false;
```
fabioberger: Put this line right above the call to `signOrderHashAsync`

packages/contracts/test/tutorials/arbitrage.ts
```diff
@@ -0,0 +1,211 @@
...
+                amountGive,
+                expires,
+                nonce,
+                edFillAmount,
+            ];
+            v = [order.params.v as number, edSignature.v];
+            r = [order.params.r as string, edSignature.r];
+            s = [order.params.s as string, edSignature.s];
+        });
+        it('should successfully execute the arbitrage if not front-runner', async () => {
```
fabioberger: `if no front-runner`

packages/contracts/contracts/tutorials/Arbitrage.sol
```diff
@@ -0,0 +1,103 @@
...
+        makeEtherDeltaTrade(addresses, values, v, r, s);
+    }
+
+    function makeEtherDeltaTrade(
+        address[8] addresses, uint[12] values,
+        uint8[2] v, bytes32[2] r, bytes32[2] s
+    ) internal {
+        uint amount = values[11];
+        etherDelta.depositToken(addresses[5], values[7]);
```
abandeali1: For the sake of a tutorial, I think automatically depositing before the EtherDelta trade and withdrawing after a trade is fine, but it should probably be optional if ever used in production (going to save a lot of gas by keeping tokens in ED).

packages/contracts/contracts/tutorials/EtherDelta.sol
```diff
@@ -0,0 +1,168 @@
...
+import "../tokens/Token.sol";
+
+contract EtherDelta is SafeMath {
```
abandeali1: Let's prettify this to keep formatting/indentation consistent.

Recmo:
The current implementation does an atomic trade of a 0x and ED order, this involves up to four different token addresses (excluding ZRX). The main use-case is arbitrage, where there are only two tokens involved. This would simplify the code quite a bit, which could be good for a tutorial. (You could even add an extra credit where the contract verifies that there is an arbitrate and rejects if there is no guaranteed profit).

packages/contracts/contracts/tutorials/Arbitrage.sol
LogvinovLeon: Unfortunately it's impossible (or at least I was unable to come out with a way to do that) because of the limit on the number of the local variables.

packages/contracts/contracts/tutorials/Arbitrage.sol
LogvinovLeon: This code doesn't pretend to be gas efficient/safe/nice. It's just a tutorial.

PR #363, opened 2018-02-02
Change the InjectedWeb3Subprovider to accept Web3.Provider

dekz:
## Description

Remove the requirement to pass in a Web3 object to InjectedWeb3Subprovider and accept a Web3 Provider.

## Motivation and Context

Allows us to use a controlled version of Web3.

## How Has This Been Tested?

Manually tested in Website which uses InjectedWeb3 provider.

## Types of changes

### DIFF
--- a/packages/subproviders/src/subproviders/injected_web3.ts
+++ b/packages/subproviders/src/subproviders/injected_web3.ts
@@ -4,7 +4,7 @@ import Web3 = require('web3');
 /*
  * This class implements the web3-provider-engine subprovider interface and forwards
  * requests involving user accounts (getAccounts, sendTransaction, etc...) to the injected
- * MetamaskInpageProvider instance in their browser.
+ * provider instance in their browser.
  * Source: https://github.com/MetaMask/provider-engine/blob/master/subproviders/subprovider.js
  */
 export class InjectedWeb3Subprovider {
@@ -41,6 +41,7 @@ export class InjectedWeb3Subprovider {
         }
     }
     // Required to implement this method despite not needing it for this subprovider
+    // This type is Web3ProviderEngine, but there is no need to import this for a noop.
     // tslint:disable-next-line:prefer-function-over-method
     public setEngine(engine: any) {
         // noop
--- a/packages/subproviders/src/subproviders/injected_web3.ts
+++ b/packages/subproviders/src/subproviders/injected_web3.ts
@@ -1,17 +1,16 @@
 import * as _ from 'lodash';
 import Web3 = require('web3');
-import Web3ProviderEngine = require('web3-provider-engine');
 
 /*
  * This class implements the web3-provider-engine subprovider interface and forwards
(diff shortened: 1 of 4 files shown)


packages/subproviders/src/subproviders/injected_web3.ts
```diff
@@ -1,17 +1,16 @@
 import * as _ from 'lodash';
 import Web3 = require('web3');
-import Web3ProviderEngine = require('web3-provider-engine');
 
 /*
  * This class implements the web3-provider-engine subprovider interface and forwards
  * requests involving user accounts (getAccounts, sendTransaction, etc...) to the injected
- * web3 instance in their browser.
+ * MetamaskInpageProvider instance in their browser.
```
fabioberger: Let's just say the `injected provider instance`, and not mention Metamask specifically. It could be any provider.

packages/subproviders/src/subproviders/injected_web3.ts
```diff
@@ -43,7 +42,7 @@ export class InjectedWeb3Subprovider {
     }
     // Required to implement this method despite not needing it for this subprovider
     // tslint:disable-next-line:prefer-function-over-method
-    public setEngine(engine: Web3ProviderEngine) {
+    public setEngine(engine: any) {
```
fabioberger: Why was this changed?

packages/website/ts/blockchain.ts
```diff
@@ -100,7 +100,7 @@ export class Blockchain {
             // We catch all requests involving a users account and send it to the injectedWeb3
             // instance. All other requests go to the public hosted node.
             provider = new ProviderEngine();
-            provider.addProvider(new InjectedWeb3Subprovider(injectedWeb3));
+            provider.addProvider(new InjectedWeb3Subprovider(injectedWeb3.currentProvider));
             provider.addProvider(new FilterSubprovider());
             provider.addProvider(new RedundantRPCSubprovider(publicNodeUrlsIfExistsForNetworkId));
             provider.start();
```
fabioberger: Can you also start a PR on wiki repo updating the wiki articles with this change?

packages/subproviders/src/subproviders/injected_web3.ts
dekz: Since we're not using the type in any way its only use was to be imported and then referenced here. By removing the type and using any in a method that is a noop, we can remove the import of the Web3 Provider Engine in this class.

packages/subproviders/src/subproviders/injected_web3.ts
```diff
@@ -42,8 +41,9 @@ export class InjectedWeb3Subprovider {
         }
     }
     // Required to implement this method despite not needing it for this subprovider
+    // This type is Web3ProviderEngine, but there is no need to import this for a noop.
```
fabioberger: // The `engine` argument type should be `Web3ProviderEngine` but we've decided to keep it as type `any` to // remove the providerEngine dependency given that this method is a noop anyways

PR #365, opened 2018-02-04
bugfix: undefined log params blow up tryToDecodeLogOrNoop

lukeautry:
Noticing this in the logs; seems to give the watcher trouble

```
(node:96) UnhandledPromiseRejectionWarning: BigNumber Error: new BigNumber() not a number: undefined
    at raise (/usr/src/app/node_modules/bignumber.js/bignumber.js:1190:25)
    at /usr/src/app/node_modules/bignumber.js/bignumber.js:1178:33
    at new BigNumber (/usr/src/app/node_modules/bignumber.js/bignumber.js:193:67)
    at /usr/src/app/node_modules/0x.js/node_modules/@0xproject/utils/lib/abi_decoder.js:48:46
    at arrayMap (/usr/src/app/node_modules/lodash/lodash.js:660:23)
    at Function.map (/usr/src/app/node_modules/lodash/lodash.js:9571:14)
    at AbiDecoder.tryToDecodeLogOrNoop (/usr/src/app/node_modules/0x.js/node_modules/@0xproject/utils/lib/abi_decoder.js:44:11)
    at OrderStateWatcher.&lt;anonymous&gt; (/usr/src/app/node_modules/0x.js/lib/src/order_watcher/order_state_watcher.js:216:60)
    at step (/usr/src/app/node_modules/0x.js/lib/src/order_watcher/order_state_watcher.js:32:23)
    at Object.next (/usr/src/app/node_modules/0x.js/lib/src/order_watcher/order_state_watcher.js:13:53)
```

I think it's probably happening at https://github.com/0xProject/0x.js/blob/development/packages/utils/src/abi_decoder.ts#L43 or https://github.com/0xProject/0x.js/blob/development/packages/utils/src/abi_decoder.ts#L49. 

Not sure on the best solution here, but this PR would at least handle an undefined value.

PR #366, opened 2018-02-05
Fix an exception when a signature collision happens

LogvinovLeon:
Fixes an exception in an order watcher when we have an event with the same signature as one of the known events, but it has different indexed fields and therefore fails to decode.

## Description

`Deposit(address a, uint b);`
and
`Deposit(address indexed a, uint b);`
have the same signature, but are encoded differently.

## How Has This Been Tested?

Manual tests

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

Closes #365

### DIFF
--- a/packages/0x.js/CHANGELOG.md
+++ b/packages/0x.js/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v0.x.x - _TBD, 2018_
 
-    * Fixed the bug causing order watcher to throw if there is the event with the same signature but different indexed fields (#)
+    * Fixed the bug causing order watcher to throw if there is the event with the same signature but different indexed fields (#366)
 
 ## v0.31.1 - _February 1, 2018_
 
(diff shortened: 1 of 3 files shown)


packages/0x.js/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # CHANGELOG
 
-## v0.x.0 - _TBD, 2018_
+## v0.x.x - _TBD, 2018_
 
+    * Fixed the bug causing order watcher to throw if there is the event with the same signature but different indexed fields (#366)
```
fabioberger: `if there is the event` -> `if there is an event`

packages/utils/CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # CHANGELOG
 
+## v0.x.x - _TBD, 2018_
+
+    * Fix a bug causing abi_decoder to throw on unknown events with the same signature (#366)
```
fabioberger: `Fix a bug related to event signature collisions (argument indexes aren't included in event signatures) in the abi_decoder. The decoder used to throw on unknown events with identical signatures as a known event (except indexes).`

PR #372, opened 2018-02-06
Add interfaces for old contracts

Recmo:
## Description

* Interface contracts are added implementing the ABI of the Token, TokenRegistery, TokenTransferProxy and Exchange contracts.
* The current version of Exchange is copied to previous tagged as V1.
* The current version of the contracts are updated to use interfaces.

## Motivation and Context

When importing old contracts in new code, the solidity versions may not match. For example, the Exchange contract imports the ZRXToken contract. The new exchange contract will be compiled with a recent version of Solidity (0.4.19). The old token contract was written in an older version of the language, and may not compile cleanly.

Instead of importing the full contract, we define the public part (the ABI) as an interface and import that instead.

## Types of changes

## Checklist:


packages/contracts/src/current/protocol/TokenRegistry/TokenRegistry.sol
```diff
@@ -16,9 +16,9 @@
 
 */
 
-pragma solidity ^0.4.11;
+pragma solidity ^0.4.19;
```
abandeali1: Let's keep the version and the imports the same until we actually make changes to the TokenRegistry.

packages/contracts/src/current/protocol/TokenTransferProxy/TokenTransferProxy.sol
```diff
@@ -16,10 +16,10 @@
 
 */
 
-pragma solidity ^0.4.11;
+pragma solidity ^0.4.19;
```
abandeali1: Same as above. When we do start making changes to the proxy, we should also keep a `TokenTransferProxy_v1` copy in the previous directory (just occurred to me, we should have done this for the Exchange as well).

packages/contracts/src/current/interfaces/IExchange_v1.sol
```diff
@@ -0,0 +1,226 @@
+pragma solidity ^0.4.19;
+
+contract IExchange_v1 {
```
abandeali1: We can leave out the `_v1` for all of the current packages. Also, I would prefer putting the interfaces in their respective folders, rather than creating a new `interfaces` folder (i.e put this in the `Exchange` folder.

packages/contracts/src/current/tokens/WETH9/WETH9.sol
```diff
@@ -13,7 +13,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-pragma solidity ^0.4.18;
+pragma solidity ^0.4.19;
```
abandeali1: Same as above.

packages/contracts/src/current/tokens/ZRXToken/ZRXToken.sol
```diff
@@ -16,9 +16,9 @@
 
 */
 
-pragma solidity ^0.4.11;
+pragma solidity ^0.4.19;
```
abandeali1: ZRXToken should keep the same version and imports, since we're not actually going to be updating it.

packages/contracts/src/current/interfaces/IExchange_v1.sol
Recmo: My intention for this was to have the old interface around. This could be useful if we want to try something where the new Exchange contract calls the existing Exchange contract.

fabioberger: So essentially only have interfaces for old versions of the contracts? That makes sense. But why keep the `IExchange_v1` interface around if no one is using it?

Recmo: Good catch! The whole `/interfaces/` folder should not be there. Must have messed up some of the rebasing.

There is an old version of the exchange in `previous/Exchange/Exchange_v1.sol`. I don't think it's likely that we will use it, but there might be cases where we want to access the old exchange from the new exchange.

Simple example where this is necessary: If the new exchange is backwards compatible (same order+sig would be accepted) then we need to check if it has been filled/cancelled by the other exchange. Otherwise an adversary could submit it to both and fill it twice. (note to self: this only solves half of the problem, submitting to new, then old, would still work).

PR #373, opened 2018-02-06
Lowercase addresses in the Public API

dekz:
## Description

Checksummed addresses are not supported, so we normalize these in all Public API's where we accept an address.

## Motivation and Context

Checksummed addresses would be passed into ZeroEx and would result in an error. This is input validation we can fix and normalize to lower case.

In some cases we were inconsistent in asserting the address is a Hex Address. Some public methods would do this some would not. This PR attempts to ensure all public methods check the addresses and lowercase them.

## Types of changes

## Checklist:


packages/0x.js/src/0x.ts
```diff
@@ -71,11 +71,12 @@ export class ZeroEx {
      * @return  Whether the signature is valid for the supplied signerAddress and data.
      */
     public static isValidSignature(data: string, signature: ECSignature, signerAddress: string): boolean {
+        const normalizedSignerAddress = signerAddress.toLowerCase();
```
fabioberger: @dekz I think we want to do these checks **after** we've asserted that it's an ETHAddressHex. Otherwise this could throw an unhelpful error if signerAddress is not a string :(

dekz: hmm let me think about this because its the `isETHAddressHex` which is asserting it is a lowercase address, so passing into `isETHAddressHex` first would raise the assertion error.

PR #375, opened 2018-02-07
Organize async task queues by network

BMillman19:
## Description

Generalized the `RequestQueue` into a `DispatchQueue` that serializes any arbitrary asynchronous tasks.

## Motivation and Context

Because the same address was being used to dispense both ETH and ZRX, sometimes the transaction nonces would get messed up. This fixes that problem by serializing transactions on a per network basis and only advancing the queue once the in-flight transaction resolves.

## How Has This Been Tested?

Local testing

## Types of changes

## Checklist:


packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -79,48 +61,76 @@ export class Handler {
...
     public async dispenseWETHOrder(req: express.Request, res: express.Response) {
         await this._dispenseOrder(req, res, RequestedAssetType.WETH);
     }
     public async dispenseZRXOrder(req: express.Request, res: express.Response, next: express.NextFunction) {
         await this._dispenseOrder(req, res, RequestedAssetType.ZRX);
     }
+    private _dispenseAsset(req: express.Request, res: express.Response, requestedAssetType: RequestedAssetType) {
+        const networkId = req.params.networkId;
+        const recipient = req.params.recipient;
+        const networkConfig = _.get(this._networkConfigByNetworkId, networkId);
```
fabioberger: Let's simply index in here rather then use `_.get`.

packages/testnet-faucets/src/ts/dispatch_queue.ts
```diff
@@ -0,0 +1,47 @@
...
+        this._queueIntervalIdIfExists = intervalUtils.setAsyncExcludingInterval(
+            async () => {
+                const task = this._queue.shift();
+                if (_.isUndefined(task)) {
+                    return Promise.resolve();
+                }
+                await task();
+            },
+            this._queueIntervalMs,
+            _.noop,
```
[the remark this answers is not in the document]
BMillman19: Hmm its a bit tricky because the error reporting is asynchronous, let's sync up if theres a way to fix this

packages/testnet-faucets/src/ts/handler.ts
```diff
@@ -79,48 +61,76 @@ export class Handler {
...
+            res.status(400).send('UNSUPPORTED_NETWORK_ID');
+            return;
+        }
+        let dispenserTask;
+        switch (requestedAssetType) {
+            case RequestedAssetType.ETH:
+                dispenserTask = dispenseAssetTasks.dispenseEtherTask(recipient, networkConfig.web3);
+                break;
+            case RequestedAssetType.WETH:
+                dispenserTask = dispenseAssetTasks.dispenseTokenTask(recipient, requestedAssetType, networkConfig.zeroEx);
```
dekz: Can these be defaulted to the same case statement body?
```typescript
case RequestedAssetType.WETH:
case RequestedAssetType.ZRX:
    dispenserTask = dispenseAssetTasks.dispenseTokenTask(recipient, requestedAssetType, networkConfig.zeroEx);
    break;
````

PR #376, opened 2018-02-07
Signature generalisation

Recmo:
## Description

* Make signatures opaque `bytes` with a one-byte flag
* Implement different signature types:
  * Current `eth_sign`
  * EIP712
  * `msg.sender`
  * External contract
* Make message hash compatible with EIP712
* Adjust internal and external interfaces for this change

## Motivation and Context

This tries to implement on:

* [ZEIP-1](https://github.com/0xProject/ZEIPs/issues/1): Off-chain order generation by smart contracts
* [ZEIP-7](https://github.com/0xProject/ZEIPs/issues/7): On-chain order generation by smart contracts

It does not try to implement, but is related to:

* [ZEIP-3](https://github.com/0xProject/ZEIPs/issues/3): Cross-relayer orders
* [ZEIP-9](https://github.com/0xProject/ZEIPs/issues/9): Allow generation of multiple orders with a single signature
* [ZEIP-10](https://github.com/0xProject/ZEIPs/issues/10): multihash/multiformat message format.

## Types of changes

* [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -104,34 +102,52 @@ contract MixinExchangeCore is
...
+        }
         require(isValidSignature(
-            order.maker,
-            order.orderHash,
-            v,
-            r,
-            s
+            order.orderHash ^ 0x1, // Domain separator maker/taker
+            order.taker,
+            takerSignature
```
abandeali1: - The taker signature should be optional (so if `takerSignature.length` is 0, assume taker is `msg.sender`).
- We also need to update a mapping to prevent replay attacks. 
- I personally prefer the more generalized approaches described in the ZEIP 18 comments. Maybe since we're pretty undecided on the best approach for this, we just make this a separate function for now?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidatorEcrecover.sol
```diff
@@ -19,26 +19,116 @@
...
-        uint8 v,
-        bytes32 r,
-        bytes32 s)
-        public
-        constant
-        returns (bool)
+        address signer,
+        bytes signature)
+        public view
+        returns (bool valid)
```
abandeali1: Change named return to `isValid`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -104,34 +102,52 @@ contract MixinExchangeCore is
...
+        } else {
+            order.taker = taker;
+        }
         require(isValidSignature(
-            order.maker,
-            order.orderHash,
-            v,
-            r,
-            s
+            order.orderHash ^ 0x1, // Domain separator maker/taker
```
abandeali1: If we have a different message format or add a salt, we can probably remove the separator.

packages/contracts/src/contracts/current/protocol/Exchange/ISigner_v1.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.4.19;
+
+contract ISigner {
+
+    function validate(bytes32 hash, bytes signature)
```
abandeali1: why not rename this `isValidSignature` for consistency?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+        bytes32 hash,
+        address signer,
+        bytes signature)
+        public view
+        returns (bool isValid)
+    {
+        // TODO: Domain separation: make hash depend on role. (Taker sig should not be valid as maker sig, etc.)
+        
+        require(signature.length >= 1);
+        SignatureType stype = SignatureType(uint8(signature[0]));
```
fabioberger: Nit: can we rename `stype` to `sigType`? Camel-case and less ambiguous.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+        
+        // Variables are not scoped in Solidity
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+        
+        // Zero is always an invalid signature
+        if (stype == SignatureType.Invalid) {
```
fabioberger: Why is this case necessary?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+            require(signature.length == 1);
+            isValid = false;
+            return;
+        
+        // Implicitely signed by caller
```
fabioberger: Typo: `Implicitly`

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+        } else if (stype == SignatureType.Caller) {
+            require(signature.length == 1);
+            isValid = signer == msg.sender;
+            return;
+        
+        // Signed using web3.eth_sign
+        } else if (stype == SignatureType.Ecrecover) {
```
fabioberger: Super cool that V2 will keep backward compatibility support for EC sigs 🥇

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+        
+        // Signature verified by signer contract
+        } else if (stype == SignatureType.Contract) {
+            isValid = ISigner(signer).isValidSignature(hash, signature);
+            return;
+        
+        // Anything else is illegal
+        } else {
+            revert();
+
```
fabioberger: Awkward new line. Can we remove?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+          
+        }
+        return false;
+    }
+    
+    function get32(bytes b, uint256 index)
+        private pure
+        returns (bytes32 result)
+    {
+        // require(b.length >= index + 32);
```
fabioberger: Is this meant to be commented out still?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: It's not necessary, but it seemed a good idea from a defense programing perspective. Default values are usually zero. Let's not make them accidentally be some sort of signature.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: It's commented for performance reasons. This check is more costly than the actual function. That should not concert us at the moment though, and it would be better to have more checks while we keep breaking stuff.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,117 @@
...
+        
+        // Zero is always an invalid signature
+        if (stype == SignatureType.Invalid) {
+            require(signature.length == 1);
+            isValid = false;
+            return;
+        
+        // Implicitely signed by caller
+        } else if (stype == SignatureType.Caller) {
+            require(signature.length == 1);
```
abandeali1: Can you give an example of how this would be used? It seems like `msg.sender` will always be `order.maker` in this case, which seems a bit strange to me.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
abandeali1: +1

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -177,8 +184,14 @@ contract MixinExchangeCore is
             orderHash: getOrderHash(orderAddresses, orderValues)
         });
 
-        require(order.maker == msg.sender);
-        require(order.makerTokenAmount > 0 && order.takerTokenAmount > 0 && takerTokenCancelAmount > 0);
+        require(order.makerTokenAmount > 0);
+        require(order.takerTokenAmount > 0);
+        require(takerTokenCancelAmount > 0);
+        require(isValidSignature(
+            keccak256(orderSchemaHash, order.orderHash),
```
abandeali1: It seems a bit strange to me that we are always including the orderSchemaHash, even when we aren't using an EIP712 signature. What's the logic for this?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,118 @@
...
+        bytes32 s;
+        
+        // Zero is always an invalid signature
+        if (signatureType == SignatureType.Invalid) {
+            require(signature.length == 1);
+            isValid = false;
+            return;
+        
+        // Implicitly signed by caller
+        } else if (signatureType == SignatureType.Caller) {
```
abandeali1: Can you provide an example of how this is used? It seems to me that in the current contracts, `msg.sender` will always be `order.maker`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,118 @@
...
+        require(signature.length >= 1);
+        SignatureType signatureType = SignatureType(uint8(signature[0]));
+        
+        // Variables are not scoped in Solidity
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+        
+        // Zero is always an invalid signature
+        if (signatureType == SignatureType.Invalid) {
```
abandeali1: When would someone pass in an intentionally invalid signature? Why not just revert if the signature is malformed?

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -113,26 +105,19 @@ contract MixinWrapperFunctions is
     /// @param orderAddresses Array of address arrays containing individual order addresses.
```
abandeali1: The length of the inputs here are no longer guaranteed to be 484 bytes.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -141,54 +126,38 @@ contract MixinWrapperFunctions is
...
-    /// @dev Synchronously executes multiple calls of fillOrderNoThrow in a single transaction.
-    /// @param orderAddresses Array of address arrays containing individual order addresses.
-    /// @param orderValues Array of uint256 arrays containing individual order values.
-    /// @param takerTokenFillAmounts Array of desired amounts of takerToken to fill in orders.
-    /// @param v Array ECDSA signature v parameters.
-    /// @param r Array of ECDSA signature r parameters.
-    /// @param s Array of ECDSA signature s parameters.
-    function batchFillOrdersNoThrow(
+
+    function fillOrdersUpTo(
```
abandeali1: Remove this function, we renamed to `marketFillOrders`.

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSignatureValidator.sol
```diff
@@ -20,19 +20,15 @@ pragma solidity ^0.4.19;
...
-    /// @param v ECDSA signature parameter v.
-    /// @param r ECDSA signature parameters r.
-    /// @param s ECDSA signature parameters s.
+    /// @dev Verifies that a signature is valid.
+    /// @param digest Message digest that is signed.
+    /// @param signer Address of signer.
+    /// @param signature Proof of signing.
     /// @return Validity of order signature.
     function isValidSignature(
+        bytes32 digest,
```
abandeali1: I don't really have a preference for `digest` over `hash`, but this arg is still named `hash` in `ISigner`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
Recmo: The reasoning is that EIP712 is not a standard for signing a hash, but a standard for generating a hash from structured data. Since none of the other signing methods have requirements on the hashing algorithm, we may as well use EIP712 everywhere.

In fact, we should probably use an EIP712 hash for `order.orderHash`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: The current PR excludes taker abstraction, so for a successful `fillOrder`, we always have `msg.sender == order.taker`. When taker abstract is added, this will no longer hold true, and this signature type can be used by either maker or taker if they call `fillOrder`.

In the current PR the `cancelOrder` function also takes a signature. If maker calls `cancelOrder` herself she can set the signature to `SignatureType.Caller`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -159,7 +165,8 @@ contract MixinExchangeCore is
     function cancelOrder(
         address[5] orderAddresses,
         uint256[6] orderValues,
-        uint256 takerTokenCancelAmount)
+        uint256 takerTokenCancelAmount,
+        bytes signature)
```
Recmo: This signature does not include a nonce, so `cancelOrders` with this signature mechanism are prone to replays.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: I want to reserve `0x00` as an invalid value, because it is the default value and might show up accidentally.

You are right, it should `revert()` instead of throwing.

Would there be value in adding a signature schema that rejects (returns false) always? It's pretty much free to add, does hurt and might be useful somehow?

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
Recmo: What's the significance of 484? I updated `fillOrderNoThrow` to take arbitrary lengths.

abandeali1: Sorry, it wouldn't let me comment on the correct line for some reason. The delegatecall takes the input length, and we're still passing in 484.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
abandeali1: In the current implementation we don't even need a signature here since `msg.sender` is required to be the maker.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,118 @@
...
+        require(b.length >= index + 32);
+        
+        // Arrays are prefixed by a 256 bit length parameter
+        index += 32;
+        
+        // Read the bytes32 from array memory
+        assembly {
+            result := mload(add(b, index))
+        }
+    }
```
abandeali1: Let's add an explicit return here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,118 @@
...
+            r = get32(signature, 2);
+            s = get32(signature, 34);
+            recovered = ecrecover(
+                keccak256("\x19Ethereum Signed Message:\n32", hash),
+                v,
+                r,
+                s
+            );
+            isValid = signer == recovered;
+            return;
```
abandeali1: Let's change all of these returns to `return isValid` for clarity.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
abandeali1: Ignore my previous comment, looks like you accidentally renamed `batchFillOrdersNoThrow` to `fillOrdersUpTo`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -141,54 +126,38 @@ contract MixinWrapperFunctions is
...
-        uint256[6][] orderValues,
-        uint256[] takerTokenFillAmounts,
-        uint8[] v,
-        bytes32[] r,
-        bytes32[] s)
-        external
+        uint[6][] orderValues,
+        uint[] takerTokenFillAmounts,
+        bytes[] signatures)
+        public /// Compiler crash when set to external
```
abandeali1: This compiles if set to external now.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
fabioberger: @Recmo our convention is the thumbs up a comment once it's been addressed.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -100,29 +96,39 @@ contract MixinExchangeCore is
...
+            require(order.takerTokenAmount > 0);
+            require(isValidSignature(
+                keccak256(orderSchemaHash, order.orderHash),
+                order.maker,
+                signature
+            ));
+        }
+        
+        // Validate taker
+        if (order.taker != address(0)) {
```
abandeali1: Is there an efficiency difference between the if block vs `require(order.taker == address(0) || order.taker == msg.sender)`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -177,8 +183,10 @@ contract MixinExchangeCore is
             orderHash: getOrderHash(orderAddresses, orderValues)
         });
 
+        require(order.makerTokenAmount > 0);
```
abandeali1: Why this change?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+        // This is always an implicit option, since a signer can create a
+        // signature array with invalid type or length. We may as well make
+        // it an explicit option. This aids testing and analysis. It is
+        // also the initialization value for the enum type.
+        if (signatureType == SignatureType.Illegal) {
+            revert();
+        
+        // Always invalid signature
+        // Like Illegal, this is always implicitely available and therefore
+        // offered dxplicitely. It can be implicitely creates by providing
```
abandeali1: Typo: explicitely

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+            require(signature.length == 1);
+            isValid = false;
+            return isValid;
+        
+        // Implicitly signed by caller
+        // The signer has initiated the call. In the case of non-contract
+        // accounts it means the transaction itself was signed.
+        // Example: lets say for a particular operation three signatures
+        // A, B  are required. To submit the transaction, A and B can give
+        // a signatue to C, who can then submit the transaction using
```
abandeali1: Typo: "signature"

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+            isValid = signer == msg.sender;
+            return isValid;
+        
+        // Signed using web3.eth_sign
+        } else if (signatureType == SignatureType.Ecrecover) {
+            require(signature.length == 66);
+            v = uint8(signature[1]);
+            r = get32(signature, 2);
+            s = get32(signature, 34);
+            recovered = ecrecover(
```
abandeali1: `address recovered`

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+                s
+            );
+            isValid = signer == recovered;
+            return isValid;
+            
+        // Signature using EIP712
+        } else if (signatureType == SignatureType.EIP712) {
+            v = uint8(signature[1]);
+            r = get32(signature, 2);
+            s = get32(signature, 35);
```
abandeali1: should be `s = get32(signature, 34);`

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -27,53 +28,41 @@ contract MixinWrapperFunctions is
     SafeMath
 {
   
-    /// @dev Fills an order with specified parameters and ECDSA signature. Throws if specified amount not filled entirely.
```
abandeali1: Why are all of these comments removed?

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -27,53 +28,41 @@ contract MixinWrapperFunctions is
...
-            s
+            signature
         ) == takerTokenFillAmount);
     }
 
     /// @dev Fills an order with specified parameters and ECDSA signature. Returns false if the transaction would otherwise revert.
     /// @param orderAddresses Array of order's maker, taker, makerToken, takerToken, and feeRecipient.
     /// @param orderValues Array of order's makerTokenAmount, takerTokenAmount, makerFee, takerFee, expirationTimestampInSec, and salt.
     /// @param takerTokenFillAmount Desired amount of takerToken to fill.
-    /// @param v ECDSA signature parameter v.
```
abandeali1: add an `@param signature` comment

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -113,26 +110,19 @@ contract MixinWrapperFunctions is
     /// @param orderAddresses Array of address arrays containing individual order addresses.
     /// @param orderValues Array of uint256 arrays containing individual order values.
     /// @param takerTokenFillAmounts Array of desired amounts of takerToken to fill in orders.
-    /// @param v Array ECDSA signature v parameters.
```
abandeali1: add an @param signature comment

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -141,54 +131,38 @@ contract MixinWrapperFunctions is
...
```
abandeali1: add an `@param signatures` comment

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -141,54 +131,38 @@ contract MixinWrapperFunctions is
...
                 takerTokenFillAmounts[i],
-                v[i],
-                r[i],
-                s[i]
+                signatures[i]
             );
         }
     }
 
-    /// @dev Synchronously executes multiple calls of fillOrderNoThrow in a single transaction.
```
abandeali1: Fix comments.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -141,54 +131,38 @@ contract MixinWrapperFunctions is
...
-        uint256[6][] orderValues,
-        uint256[] takerTokenFillAmounts,
-        uint8[] v,
-        bytes32[] r,
-        bytes32[] s)
+        uint[6][] orderValues,
+        uint[] takerTokenFillAmounts,
+        bytes[] signatures)
         external
+        returns (uint)
```
abandeali1: This doesn't actually return anything.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -197,17 +171,12 @@ contract MixinWrapperFunctions is
     /// @param orderAddresses Array of address arrays containing individual order addresses.
     /// @param orderValues Array of uint256 arrays containing individual order values.
     /// @param takerTokenFillAmount Desired total amount of takerToken to fill in orders.
-    /// @param v Array ECDSA signature v parameters.
```
abandeali1: add an `@param signatures` comment

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -230,17 +197,12 @@ contract MixinWrapperFunctions is
...
```
abandeali1: add an `@param signatures` comment

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
dekz: implicitly
a valid format

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+            return isValid;
+        
+        // Signed using web3.eth_sign
+        } else if (signatureType == SignatureType.Ecrecover) {
+            require(signature.length == 66);
+            v = uint8(signature[1]);
+            r = get32(signature, 2);
+            s = get32(signature, 34);
+            recovered = ecrecover(
+                keccak256("\x19Ethereum Signed Message:\n32", hash),
```
fabioberger: Let's add support for current Trezors which do not use ASCII characters for the message length encoding: https://github.com/ethereum/go-ethereum/issues/14794.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+                v,
+                r,
+                s
+            );
+            isValid = signer == recovered;
+            return isValid;
+            
+        // Signature using EIP712
+        } else if (signatureType == SignatureType.EIP712) {
+            v = uint8(signature[1]);
```
fabioberger: Do we want to assert the signature length?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -100,29 +96,39 @@ contract MixinExchangeCore is
             expirationTimestampInSec: orderValues[4],
             orderHash: getOrderHash(orderAddresses, orderValues)
         });
+        
+        // Validate order and maker only if first time seen
+        if (filled[order.orderHash] == 0 && cancelled[order.orderHash] == 0) {
```
LogvinovLeon: Do we have any data on how expensive is the signature validation? Two state reads also consume some gas.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+        uint8 v;
+        bytes32 r;
+        bytes32 s;
+        
+        // Always illegal signature
+        // This is always an implicit option, since a signer can create a
+        // signature array with invalid type or length. We may as well make
+        // it an explicit option. This aids testing and analysis. It is
+        // also the initialization value for the enum type.
+        if (signatureType == SignatureType.Illegal) {
```
LogvinovLeon: All those ifs add to understanding but increase the gas consumption. Will we comment them out in a production version?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,138 @@
...
+            v = uint8(signature[1]);
+            r = get32(signature, 2);
+            s = get32(signature, 35);
+            address recovered = ecrecover(hash, v, r, s);
+            isValid = signer == recovered;
+            return isValid;
+        
+        // Signature verified by signer contract
+        } else if (signatureType == SignatureType.Contract) {
+            isValid = ISigner(signer).isValidSignature(hash, signature);
```
LogvinovLeon: isValidSignature seems too generic. Can we change to `isValid0xOrderSignature` or smth longer?

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -27,53 +28,41 @@ contract MixinWrapperFunctions is
...
-        uint8 v,
-        bytes32 r,
-        bytes32 s)
+        bytes signature)
         public
         returns (bool success, uint256 takerTokenFilledAmount)
     {
         bytes4 FILL_ORDER_FUNCTION_SIGNATURE = bytes4(keccak256("fillOrder(address[5],uint256[6],uint256,uint8,bytes32,bytes32)"));
         
+        // Input size is padded to a 4 + n * 32 byte boundary
```
LogvinovLeon: If this is not a WIP any more we need more comments. Much more comments.
I'm unable to read that. Let's reference to some solidity documentation or other sources that can help people understand that. (Talking about the whole function. Let's also have it as a rule to have a very good and verbose explanation before every assembly trick. How much does it save, why is it absolutely neccessary? not against assembly in general, but I want other people to be able to understand that even if they're not so experienced.

LogvinovLeon: Feel free to ignore it for now. It's still a prototype.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
abandeali1: I think it's intended to be generic though. I could see people using the same function for other purposes as well?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
abandeali1: An ecrecover is at least 3000 gas (and your gas profiler actually showed it being one of the most expensive parts of the function). An sload is only 200 I believe.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -73,19 +73,15 @@ contract MixinExchangeCore is
...
-          bytes32 r,
-          bytes32 s)
-          public
-          returns (uint256 takerTokenFilledAmount)
+        address[5] orderAddresses,
+        uint[6] orderValues,
+        uint takerTokenFillAmount,
+        bytes signature)
+        public
+        returns (uint256 takerTokenFilledAmount)
```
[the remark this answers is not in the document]
abandeali1: Yeah, there was extra indentation before.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
Recmo: Not that I know of. It is like this mostly because I reverted the taker abstraction. With taker abstraction we need a more complex expression than `order.taker == msg.sender`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
Recmo: I used to do `assert(someThing && somethingElse && somethingOther)` a lot in a past project. Then one day we get a stack trace from a customer, and it says basically: "assertion failed on line ...". We wasted a lot of time trying to figure out which of the clauses failed and learned a hard lesson on never using `&&` in an assertion :)

If there is no performance difference, I prefer not using `&&` in assertions. Imho it makes it more readable and debuggable.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
Recmo: I added them again.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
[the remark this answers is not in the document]
Recmo: It's the convention I'm used to and also matches with what I see in the `tokenRegistery` contract.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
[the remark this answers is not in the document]
Recmo: Next hackathon idea: `atom-grammarly` plugin.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: Yes we do!

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
Recmo: Good question! `ecrecover` is 3000 gas + 36 for the hash + couple hundred overhead. State read is 200 gas. Let's ignore (partially) canceled orders for now and look at two scenarios:

1. Unfilled order: Read both filled and cancelled, then verify signature ~ 4000 gas.
2. Partially filled order: Read filled, short circuit, skip verify ~ 200 gas.

So depending on the ratio partial filled / new orders the expect gas cost is 4000 x + 200 (1-x) = 
3800 x + 200.

Status quo is a fixed cost of about 3500. Setting these equal and solving for x gives 87%.

So if less than 87% of the fill orders are new orders, this method saves gas.

## But!

We need to optimize for the case where the order is valid. In this case, we would need to read both state variables anyway to process the order, so reading it earlier should not add gas cost. Making the optimization essentially free.

At the moment we don't optimize for this, and we read the same state again in `getUnavailableTakerTokenAmount` on line 124(244) and again on line 138. This is of course 
redundant, as the state can not change in between.

I added a comment to remind us to optimize this.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: I was not planning on that, and we need some way to distinguish the cases. What do you suggest instead?

A simple trick is to sort the branches by popularity.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: As it is, it is pretty generic. Right now we only use it for maker signature, we will also use it for taker signature and maybe controller signature, etc. These would then sign different objects, not Orders's but FillOrder's and possibly other structures.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
Recmo: Yeah, I'm not happy with the complexity of this assembly at all. It's not even tested and probably contains bugs. It does show that there is a solution though. We can reimplement it using `byte[] input` so we can construct the input arguments in Solidity instead of assembly.

Recmo: Added a `// TODO`s

Recmo:
## TODO

[ ] Tezor signatures

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
EvilJordan: I'm glad you're adding support for this, but can't give a thumbs up because it's obnoxious it's different in the first place.

Recmo:
> I kind of like adding more explicit support for 0xProject/ZEIPs#7, since this requires one less external call than SignatureType.Contract. Thoughts on adding SignatureType.Approved?

How about this:

```
contract MixinSignatureValidator {

    enum SignatureType {
        // ....
        Presigned
   }

    mapping ((hash, signer) => bool) presigned;

    function preSign(bytes32 hash, address signer, bytes signature) {
         require(signature.type != Presigned); // Not really necessary
         if(isValidSignature(hash, signer, signature)) {
              presigned[hash, signer] = true;
         }
    }

    function isValidSignature(bytes32 hash, address signer, bytes signature) {
         // ...
         if(signature.type == Presign && presigned[hash, signer])
              return true;
    }

  // ...
}
```

Recmo:
> Also, it occurs to me that we should probably be returning the address of the signer rather than a bool in all cases. With taker abstraction, we don't always know the address of the taker in advance (and therefore can't pass in signer).

Hmm... So far my solution was to add `address taker, bytes takerSignature` to the `fillOrder` arguments. In the case of `SignatureType.Caller` this is redundant. In the case of `SignatureType.Ecrecover` it is a bit more complex, as an invalid signature would always be accepted, but result in a random address (which will likely not have tokens and thus ultimately fail the TX, but I prefer if it fails at signature verification stage already). In the case of `SignatureType.Contract` and the above proposed `SignatureType.Presign` we absolutely need a provided address, but this could be provided in the `bytes signature`.

I'm not sure which is more elegant. Can we revisit this discussion when we PR the taker abstraction?

abandeali1:
Re presigning an order, why not just use a mapping of orderHash => signer?

Recmo:
> Re presigning an order, why not just use a mapping of orderHash => signer?

I try to keep `isValidSignature` agnostic to its use case. Currently it supports multiple signers for the same message. This mapping would break that assumption by allowing only one signer per orderHash. Now that I think of it, that may also cause a race condition if we receive two `preSigns` for the same message hash with different signers.

> I also don't think that presigning should actually require a signature (we can simply calculate the orderHash and use that, since maker is part of the hash).

This would make it specific to the maker signatures, it would not work for taker signatures and potential other kind of signature uses.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
Recmo: Actually, I need to side with the Trezor guys here. The way `eth_sign` does it is a security hole. It allows two different messages to have the same hash.

EvilJordan: Oh, agreed. varInt is much better than an ASCII representation. It's just so strange there doesn't appear to be an EIP around it and everyone (other than Trezor) just supports both ways like it's no big deal.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -26,85 +27,129 @@ contract MixinWrapperFunctions is
...
+        // | 420    | (1)    | signature                    |
+        // | (2)    | (3)    | padding (zero)               |
+        // | (4)    |        | end of input                 |
+        //
+        // (1): len(signature)
+        // (2): 420 + len(signature)
+        // (3): (32 - len(signature)) mod 32
+        // (4): 420 + len(signature) + (32 - len(signature)) mod 32
+        //
+        // [1]: https://solidity.readthedocs.io/en/develop/abi-spec.html
```
Recmo: Looks like this is even more complex: http://solidity.readthedocs.io/en/develop/abi-spec.html#use-of-dynamic-types

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
fabioberger: We are really trying to push the community towards https://github.com/ethereum/EIPs/pull/712

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,163 @@
...
+            v = uint8(signature[1]);
+            r = get32(signature, 2);
+            s = get32(signature, 34);
+            recovered = ecrecover(hash, v, r, s);
+            isValid = signer == recovered;
+            return isValid;
+        
+        // Signature from Trezor hardware wallet
+        // It differs from web3.eth_sign in the encoding of message length
+        // (Bitcoin varint encoding vs ascii-decimal, the later is not
```
fabioberger: `later` -> `latter`

PR #377, opened 2018-02-07
Use SignedOrder and Order types from 0x.js

LogvinovLeon:
This PR:
* Moves contracts repo tests infra closer to 0x.js by using Order and SignedOrder types

## Description

This also allows to be more type-safe, cause it's now impossible to use an unsigned order in place where signed order is expected by acident.

We also re-use signing and hashing from 0x.js, so when working on v2 we'll need to change that only in one place.

## Motivation and Context

I was irritated, that I need to do `order.params.v as number` everywhere because we had a single class for an Order and SignedOrder and I needed to promise the compiler, that I've already signed the order.

## How Has This Been Tested?

All the old tests pass

## Types of changes

## Checklist:


packages/contracts/test/exchange/wrapper.ts
```diff
@@ -165,15 +160,15 @@ describe('Exchange', () => {
...
+                fillTakerTokenAmount: order.takerTokenAmount.div(2),
             });
 
             return expect(exWrapper.fillOrKillOrderAsync(order, taker)).to.be.rejectedWith(constants.REVERT);
         });
     });
 
     describe('batch functions', () => {
-        let orders: Order[];
+        let orders: SignedOrder[];
```
fabioberger: Rename to `signedOrders`

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -186,25 +181,25 @@ describe('Exchange', () => {
         describe('batchFillOrders', () => {
             it('should transfer the correct amounts', async () => {
                 const fillTakerTokenAmounts: BigNumber[] = [];
-                const makerToken = rep.address;
-                const takerToken = dgd.address;
+                const makerTokenAddress = rep.address;
+                const takerTokenAddress = dgd.address;
                 orders.forEach(order => {
```
fabioberger: `signedOrder`

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -223,25 +218,25 @@ describe('Exchange', () => {
         describe('batchFillOrKillOrders', () => {
             it('should transfer the correct amounts', async () => {
                 const fillTakerTokenAmounts: BigNumber[] = [];
-                const makerToken = rep.address;
-                const takerToken = dgd.address;
+                const makerTokenAddress = rep.address;
+                const takerTokenAddress = dgd.address;
                 orders.forEach(order => {
```
fabioberger: Ditto above. If it's a `signedOrder`, it should be named as such.

PR #385, opened 2018-02-09
Better validate ZeroExConfig on private networks

LogvinovLeon:
Validate that if 0x.js is connected to a private network - all the options addresses should be passed.
## Description

This PR uses two approaches.
Schema validation and the trick that allows to have dependent taged types in TS.

This is now the type for ZeroExConfig.

```typescript
export interface ZeroExPublicNetworkConfig {
    networkId: 1 | 3 | 4 | 42 | 50;
    gasPrice?: BigNumber;
    exchangeContractAddress?: string;
    zrxContractAddress?: string;
    tokenRegistryContractAddress?: string;
    tokenTransferProxyContractAddress?: string;
    orderWatcherConfig?: OrderStateWatcherConfig;
}

export interface ZeroExPrivateNetworkConfig {
    networkId: number;
    gasPrice?: BigNumber;
    exchangeContractAddress: string;
    zrxContractAddress: string;
    tokenRegistryContractAddress: string;
    tokenTransferProxyContractAddress: string;
    orderWatcherConfig?: OrderStateWatcherConfig;
}
export type ZeroExConfig = ZeroExPublicNetworkConfig | ZeroExPrivateNetworkConfig;
```

As you see one version is for private networks and other is for public ones, but the problem is that when you define `const x = 5` in TS it has the type `number` not `5`. You can always cast it, but it adds some overhead.

For now I did it like that, but I'm open to suggestions.
Maybe that's actually good. If you have networkId constant - then it's no problem defining it's type manually as `1` or `42`. If you're unsure which value does your var have - then you'd better first check it before instantiating 0x.js.

## Motivation and Context

This helps people to understand why they get ContractDoesNotExist error earlier.

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/src/utils/assert.ts
+++ b/packages/0x.js/src/utils/assert.ts
@@ -1,7 +1,7 @@
 import { assert as sharedAssert } from '@0xproject/assert';
 // We need those two unused imports because they're actually used by sharedAssert which gets injected here
 // tslint:disable-next-line:no-unused-variable
-import { Schema, SchemaValidator } from '@0xproject/json-schemas';
+import { Schema } from '@0xproject/json-schemas';
 // tslint:disable-next-line:no-unused-variable
 import { BigNumber } from '@0xproject/utils';
 import { Web3Wrapper } from '@0xproject/web3-wrapper';
--- a/packages/0x.js/src/utils/assert.ts
+++ b/packages/0x.js/src/utils/assert.ts
@@ -1,7 +1,7 @@
 import { assert as sharedAssert } from '@0xproject/assert';
 // We need those two unused imports because they're actually used by sharedAssert which gets injected here
 // tslint:disable-next-line:no-unused-variable
-import { Schema } from '@0xproject/json-schemas';
+import { Schema, SchemaValidator } from '@0xproject/json-schemas';
 // tslint:disable-next-line:no-unused-variable
 import { BigNumber } from '@0xproject/utils';
 import { Web3Wrapper } from '@0xproject/web3-wrapper';
--- a/packages/0x.js/test/expiration_watcher_test.ts
+++ b/packages/0x.js/test/expiration_watcher_test.ts
@@ -9,10 +9,10 @@ import * as Web3 from 'web3';
 import { ZeroEx } from '../src/0x';
 import { ExpirationWatcher } from '../src/order_watcher/expiration_watcher';
 import { DoneCallback, Token } from '../src/types';
-import { constants } from '../src/utils/constants';
 import { utils } from '../src/utils/utils';
 
 import { chaiSetup } from './utils/chai_setup';
+import { constants } from './utils/constants';
 import { FillScenarios } from './utils/fill_scenarios';
 import { reportNoErrorCallbackErrors } from './utils/report_callback_errors';
 import { TokenUtils } from './utils/token_utils';
--- a/packages/assert/CHANGELOG.md
+++ b/packages/assert/CHANGELOG.md
@@ -1,9 +1,5 @@
 # CHANGELOG
 
-## v0.1.0 - _TBD, 2018_
-
-    * Add schemaValidator as a field so that one can add custom schemas (#385)
-
 ## v0.0.18 - _February 9, 2017_
 
     * Fix publishing issue where .npmignore was not properly excluding undesired content (#389)
--- a/packages/assert/CHANGELOG.md
+++ b/packages/assert/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.1.0 - _TBD, 2018_
+
+    * Add schemaValidator as a field so that one can add custom schemas (#385)
+
 ## v0.0.18 - _February 9, 2017_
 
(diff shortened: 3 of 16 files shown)


packages/assert/src/index.ts
```diff
@@ -67,8 +69,7 @@ export const assert = {
         this.assert(isWeb3Provider, this.typeAssertionMessage(variableName, 'Web3.Provider', value));
     },
     doesConformToSchema(variableName: string, value: any, schema: Schema): void {
-        const schemaValidator = new SchemaValidator();
-        const validationResult = schemaValidator.validate(value, schema);
+        const validationResult = assert.schemaValidator.validate(value, schema);
```
BMillman19: we're using `this.property` everywhere else in the file

packages/assert/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.1.0 - _TBD, 2018_
```
BMillman19: 👍

packages/assert/src/index.ts
```diff
@@ -4,8 +4,10 @@ import * as _ from 'lodash';
 import * as validUrl from 'valid-url';
 
 const HEX_REGEX = /^0x[0-9A-F]*$/i;
+const schemaValidator = new SchemaValidator();
 
 export const assert = {
+    schemaValidator,
```
fabioberger: I really don't like this pattern. The reason why is because if a user of this package forgets to add the schemas he later tries to check for by calling `doesConformToSchema`, it will fail to do so.

fabioberger: I'd prefer we convert `assert` into a class and explicitly have it instantiated with all the schemas passed into it. If a user calls `doesConformToSchema` with a schema that was not passed in during instantiation, an error is thrown. Does this make sense?

packages/contracts/util/constants.ts
```diff
@@ -2,7 +2,7 @@ export const constants = {
     NULL_BYTES: '0x',
     INVALID_OPCODE: 'invalid opcode',
     REVERT: 'revert',
-    TESTRPC_NETWORK_ID: 50,
+    TESTRPC_NETWORK_ID: 50 as 50,
```
fabioberger: Hm, this is going to be really confusing for TS newbies :( Is there no way around this?

packages/assert/src/index.ts
LogvinovLeon: This is also what will happen now. It will just throw an error, that that exact schema is not found.

packages/contracts/util/constants.ts
LogvinovLeon: I'd really like to not do it that way, but I described the reasons in PR description. Maybe the gains here are smaller than losses and we can just abandon that PR?

packages/assert/src/index.ts
fabioberger: Let's leave the code better off then how we found it. Any reservations with making this a class?

packages/contracts/util/constants.ts
fabioberger: Let's remove this.

packages/assert/src/index.ts
```diff
@@ -67,8 +69,7 @@ export const assert = {
         this.assert(isWeb3Provider, this.typeAssertionMessage(variableName, 'Web3.Provider', value));
     },
     doesConformToSchema(variableName: string, value: any, schema: Schema): void {
```
fabioberger: Final design:
```
doesConformToSchema(variableName: string, value: any, schema: Schema, subSchemas?: Schema[]): void
```

packages/0x.js/src/utils/assert.ts
```diff
@@ -1,7 +1,7 @@
 import { assert as sharedAssert } from '@0xproject/assert';
 // We need those two unused imports because they're actually used by sharedAssert which gets injected here
 // tslint:disable-next-line:no-unused-variable
-import { Schema } from '@0xproject/json-schemas';
+import { Schema, SchemaValidator } from '@0xproject/json-schemas';
```
fabioberger: Remove this unused import.

packages/0x.js/test/expiration_watcher_test.ts
```diff
@@ -9,10 +9,10 @@ import * as Web3 from 'web3';
 import { ZeroEx } from '../src/0x';
 import { ExpirationWatcher } from '../src/order_watcher/expiration_watcher';
 import { DoneCallback, Token } from '../src/types';
-import { constants } from '../src/utils/constants';
 import { utils } from '../src/utils/utils';
 
 import { chaiSetup } from './utils/chai_setup';
+import { constants } from './utils/constants';
```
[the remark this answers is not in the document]
fabioberger: @LogvinovLeon?

packages/assert/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+    * Add schemaValidator as a field so that one can add custom schemas (#385)
+
```
fabioberger: Remove this changelog entry.

PR #389, opened 2018-02-09
Fix NPM Inclusion Issues

fabioberger:
## Description

Currently many of our packages are not correctly ignoring src files, configs, etc... from NPM package publishing. This PR fixes that by adding a `.npmignore` file to each sub-package in the repo with the correct files/folders excluded.

Additionally, this PR adds TSLint checking to the `deployer` sub-package and fixes the existing issues.

## Motivation and Context

Our packages are much larger then they need to be. This reduces the size of almost all of them. This was an oversight on our part.

## How Has This Been Tested?

Manually running `yarn pack` in each sub-package and inspecting the contents of the resulting package to ensure it only includes desired files/folders.

## Types of changes

## Checklist:


LogvinovLeon:
Is it possible to have one .npmignore file in the root directory? Why are we not ignoring tslint.json?

LogvinovLeon:
Why are we ignoring `yarn-error.log` only in one package?

fabioberger:
@LogvinovLeon Lerna by-design wants it to be in each individual sub-package: https://github.com/lerna/lerna/issues/296

fabioberger:
Will add `yarn-error.log` to all of them.

packages/0x.js/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.32.2 - _February 9, 2018_
```
BMillman19: These new CHANGELOG entries aren't published just yet so they should technically be labeled as TBD but if we're publishing right after this, should be ok

packages/abi-gen/CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
 # CHANGELOG
 
+## v0.2.1 - _February 7, 2018_
```
BMillman19: February 9, 2018

packages/dev-utils/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.0.12 - _February 7, 2018_
```
BMillman19: 9th

packages/json-schemas/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.7.10 - _February 9, 2017_
```
BMillman19: 2018

packages/0x.js/CHANGELOG.md
fabioberger: We are.

PR #394, opened 2018-02-13
Add page options to relevant HttpClient methods

BMillman19:
## Description

Added additional `page` and `per_page` properties to inputs on various `HttpClient` methods where paging is supported 

## Motivation and Context

Right now, requesting specific pages of information is not supported

## How Has This Been Tested?

Unit tests

## Types of changes

## Checklist:


packages/connect/src/http_client.ts
```diff
@@ -13,14 +13,20 @@ import {
...
     TokenPairsItem,
-    TokenPairsRequest,
+    TokenPairsRequestOpts,
 } from './types';
 import { relayerResponseJsonParsers } from './utils/relayer_response_json_parsers';
 
 const TRAILING_SLASHES_REGEX = /\/+$/;
+const DEFAULT_PAGED_REQUEST_OPTS: PagedRequestOpts = {
+    page: 0,
+    per_page: 100,
```
fabioberger: I'd prefer we use camelcase here `perPage` and then have a mapping `perPage: 'per_page'` which all the opts get looked up against before the query params get appended to the request URL.

packages/connect/src/http_client.ts
```diff
@@ -38,34 +44,37 @@ export class HttpClient implements Client {
     }
     /**
      * Retrieve token pair info from the API
-     * @param   request     A TokenPairsRequest instance describing specific token information
-     *                      to retrieve
+     * @param   requestOpts     An optional (TokenPairsRequestOpts & PagedRequestOpts) instance describing token information
+     *                          to retrieve with page information, defaults to { page: 0, per_page: 100 }
```
fabioberger: `Request options specifying the specific token pairs to retrieve as well as page options (page index and/or number of items per page to fetch). Defaults to { page: 0, per_page: 100 }.`

fabioberger: Don't think we need to put the type in the description. Big chance they become out of sync in the future.

packages/connect/src/http_client.ts
```diff
@@ -38,34 +44,37 @@ export class HttpClient implements Client {
...
      * @return  The resulting TokenPairsItems that match the request
      */
-    public async getTokenPairsAsync(request?: TokenPairsRequest): Promise<TokenPairsItem[]> {
-        if (!_.isUndefined(request)) {
-            assert.doesConformToSchema('request', request, clientSchemas.relayerTokenPairsRequestSchema);
+    public async getTokenPairsAsync(requestOpts?: TokenPairsRequestOpts & PagedRequestOpts): Promise<TokenPairsItem[]> {
+        if (!_.isUndefined(requestOpts)) {
+            assert.doesConformToSchema('requestOpts', requestOpts, clientSchemas.tokenPairsRequestOptsSchema);
+            assert.doesConformToSchema('requestOpts', requestOpts, clientSchemas.pagedRequestOptsSchema);
```
fabioberger: Looks like our schemas  do not throw if someone mis-spells a option (e.g `pages`) since it ignores irrelevant keys?

packages/connect/src/http_client.ts
```diff
@@ -38,34 +44,37 @@ export class HttpClient implements Client {
...
         };
-        const responseJson = await this._requestAsync('/token_pairs', HttpRequestType.Get, requestOpts);
+        const responseJson = await this._requestAsync('/token_pairs', HttpRequestType.Get, httpRequestOpts);
         const tokenPairs = relayerResponseJsonParsers.parseTokenPairsJson(responseJson);
         return tokenPairs;
     }
     /**
      * Retrieve orders from the API
-     * @param   request     An OrdersRequest instance describing specific orders to retrieve
+     * @param   requestOpts     An optional (OrdersRequestOpts & PagedRequestOpts) instance describing specific orders to retrieve
```
fabioberger: Ditto comment above about including the type.

packages/connect/src/http_client.ts
BMillman19: Going to add an item to discuss this in the backlog

packages/connect/src/schemas/paged_request_opts_schema.ts
```diff
@@ -0,0 +1,8 @@
+export const pagedRequestOptsSchema = {
+    id: '/PagedRequestOpts',
+    type: 'object',
+    properties: {
+        page: { type: 'number' },
+        per_page: { type: 'number' },
```
fabioberger: This needs to be `perPage`. This is the downside discussed re: allowing arbitrary additional properties in JS objects validated with schemas :(

BMillman19: damn nice catch

PR #396, opened 2018-02-14
[WIP] Reclaim storage on expired orders

Recmo:
## Types of changes

## Checklist:


abandeali1:
This feels like a bit of an edge case, and IMO could lead to less flexibility down the line. I could see the filled and cancelled mappings being a key part of different incentive structures, where they are used as proof of a certain type of transaction happening.

PR #408, opened 2018-02-17
[WIP] Check dependencies when compiling contracts

abandeali1:
## Description
Previously, contracts where not recompiled if the hash of the source matched the hash stored in the contract artifacts. This PR checks the contract's dependencies when determining if the contract should be recompiled. 

## Motivation and Context
Contracts are not correctly recompiling on the `v2-prototype` branch.

## How Has This Been Tested?
Tested manually for now.
## Types of changes

## Checklist:


.gitignore
```diff
@@ -71,3 +71,6 @@ packages/website/public/bundle*
 
 # generated binaries
 bin/
+
+# generated contract artifacts
+packages/contracts/src/artifacts
```
BMillman19: add newline at the end of the file

packages/deployer/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.2.0 - _??_
```
BMillman19: `_TBD, 2018_`

packages/deployer/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
...
+
+    * Check dependencies when determining if contracts should be recompiled.
```
BMillman19: add PR number here

packages/deployer/src/deployer.ts
```diff
@@ -36,7 +36,7 @@ export class Deployer {
      */
     public async deployAsync(contractName: string, args: any[] = []): Promise<Web3.ContractInstance> {
         const contractArtifact: ContractArtifact = this._loadContractArtifactIfExists(contractName);
-        const contractData: ContractData = this._getContractDataFromArtifactIfExists(contractArtifact);
```
BMillman19: should we rename the methods as well? `this._getContractNetworkDataFromArtifactIfExists`

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
     public async compileAllAsync(): Promise<void> {
         await this._createArtifactsDirIfDoesNotExistAsync();
         this._contractSourcesIfExists = await Compiler._getContractSourcesAsync(this._contractsDir);
```
fabioberger: If this can not exist, we should check for it's existence before using it. `IfExists` semantically means that this variable could be `undefined`.

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
-        const compiledContractPromises = _.map(contractBaseNames, async (contractBaseName: string): Promise<void> => {
-            return this._compileContractAsync(contractBaseName);
+        _.forIn(this._contractSourcesIfExists, (source, fileName) => {
+            this._contractSourceData[fileName] = Compiler._getContractSpecificSourceData(source);
         });
-        await Promise.all(compiledContractPromises);
-
+        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
+            ? _.keys(this._contractSourcesIfExists)
+            : Array.from(this._specifiedContracts.values());
```
fabioberger: Use `_.values` here. It gives you an array from the object values.

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
-     * @param contractBaseName Name of contract with '.sol' extension.
+     * @param fileName Name of contract with '.sol' extension.
      */
-    private async _compileContractAsync(contractBaseName: string): Promise<void> {
+    private async _compileContractAsync(fileName: string): Promise<void> {
         if (_.isUndefined(this._contractSourcesIfExists)) {
             throw new Error('Contract sources not yet initialized');
         }
+        const contractSpecificSourceData = this._contractSourceData[fileName];
+        const currentArtifact = (await this._getContractArtifactOrReturnAsync(fileName)) as ContractArtifact;
```
fabioberger: Rename: `_getContractArtifactOrReturnAsync` to `_getContractArtifactIfExistsAsync` and `currentArtifact` to `currentArtifactIfExists`

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
-                    oldNetwork.keccak256 !== sourceHash ||
-                    oldNetwork.optimizer_enabled !== this._optimizerEnabled) &&
-                isContractSpecified;
-        } catch (err) {
-            shouldCompile = isContractSpecified;
-        }
-
+        const shouldCompile =
+            _.isUndefined(currentArtifact) ||
+            currentArtifact.networks[this._networkId].optimizer_enabled !== this._optimizerEnabled ||
```
fabioberger: Are we absolutely sure that a key exists for `this._networkId` here? Otherwise this will give an indexing error. It would be safer to check that as well here.

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
         if (!shouldCompile) {
             return;
         }
 
-        const input = {
-            [contractBaseName]: source,
-        };
-        const solcVersion = Compiler._parseSolidityVersion(source);
-        const fullSolcVersion = binPaths[solcVersion];
+        const fullSolcVersion = binPaths[contractSpecificSourceData.solc_version];
```
fabioberger: Eeek. There should be no variables with underscores... Camel case everywhere **except** right before communicating with an external service that expects underscored names.

packages/deployer/src/compiler.ts
```diff
@@ -187,46 +205,78 @@ export class Compiler {
...
-        const contractData: ContractData = {
-            solc_version: solcVersion,
+        const contractNetworkData: ContractNetworkData = {
+            solc_version: contractSpecificSourceData.solc_version,
             keccak256: sourceHash,
+            source_tree_hash: sourceTreeHash,
             optimizer_enabled: this._optimizerEnabled,
             abi,
             unlinked_binary,
             updated_at,
```
fabioberger: These are fine since the artifacts file spec expects underscores.

packages/deployer/src/compiler.ts
```diff
@@ -238,8 +288,8 @@ export class Compiler {
         if (_.isUndefined(this._contractSourcesIfExists)) {
             throw new Error('Contract sources not yet initialized');
         }
-        const contractBaseName = path.basename(importPath);
-        const source = this._contractSourcesIfExists[contractBaseName];
+        const fileName = path.basename(importPath);
+        const source = this._contractSourcesIfExists[fileName];
```
fabioberger: Rename `source` to `sourceIfExists`

packages/deployer/src/compiler.ts
```diff
@@ -254,4 +304,25 @@ export class Compiler {
...
+        try {
+            const opts = {
+                encoding: 'utf8',
+            };
+            const contractArtifactString = await fsWrapper.readFileAsync(currentArtifactPath, opts);
+            contractArtifact = JSON.parse(contractArtifactString);
+            return contractArtifact;
+        } catch (err) {
+            utils.consoleLog(`Artifact for ${fileName} does not exist`);
+            return contractArtifact;
```
fabioberger: Let's just return `undefined` explicitly here so that it's clearer to the reader what is happening. This confused me for a second and could be brittle to future changes.

packages/deployer/src/cli.ts
```diff
@@ -15,7 +16,6 @@ const DEFAULT_NETWORK_ID = 50;
 const DEFAULT_JSONRPC_PORT = 8545;
 const DEFAULT_GAS_PRICE = (10 ** 9 * 2).toString();
 const DEFAULT_CONTRACTS_LIST = '*';
-
```
LogvinovLeon: Why is that removed? Looked better before

packages/deployer/src/compiler.ts
```diff
@@ -5,32 +5,39 @@ import solc = require('solc');
...
+    ContractSourceData,
     ContractSources,
+    ContractSpecificSourceData,
     ImportContents,
 } from './utils/types';
 import { utils } from './utils/utils';
 
-const SOLIDITY_FILE_EXTENSION = '.sol';
 const ALL_CONTRACTS_IDENTIFIER = '*';
+const SOLIDITY_VERSION_REGEX = /(?:solidity\s\^?)([0-9]{1,2}[.][0-9]{1,2}[.][0-9]{1,2})/;
```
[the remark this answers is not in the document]
LogvinovLeon: Maybe this will be easier?
```
/(?:solidity\s\^?)(\d+\.\d+\.\d+)/;
```

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
         });
-        await Promise.all(compiledContractPromises);
-
+        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
+            ? _.keys(this._contractSourcesIfExists)
+            : Array.from(this._specifiedContracts.values());
+        _.forEach(fileNames, fileName => {
+            this._setSourceTreeHash(fileName);
+        });
+        await Promise.all(_.map(fileNames, async fileName => this._compileContractAsync(fileName)));
```
LogvinovLeon: `await Promise.all(_.map(fileNames, this._compileContractAsync));`

packages/deployer/src/compiler.ts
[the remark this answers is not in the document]
abandeali1: Yep, I'll add a source to the one I copied.

packages/deployer/src/compiler.ts
abandeali1: It doesn't matter in this case because it will be overwritten, though.

packages/deployer/src/compiler.ts
abandeali1: Seems like this doesn't work with Sets.

packages/deployer/src/compiler.ts
abandeali1: A default networkId is always passed into the constructor.

packages/deployer/src/compiler.ts
abandeali1: Hmm, `source` should always exist here assuming that `this._contractSourcesIfExists` is initialized.

packages/deployer/src/compiler.ts
abandeali1: We need the arrow function for binding `this` here.

packages/deployer/src/compiler.ts
```diff
@@ -116,67 +150,51 @@ export class Compiler {
...
         this._contractSourcesIfExists = await Compiler._getContractSourcesAsync(this._contractsDir);
-        const contractBaseNames = _.keys(this._contractSourcesIfExists);
-        const compiledContractPromises = _.map(contractBaseNames, async (contractBaseName: string): Promise<void> => {
-            return this._compileContractAsync(contractBaseName);
+        _.forIn(this._contractSourcesIfExists, (source, fileName) => {
+            this._contractSourceData[fileName] = Compiler._getContractSpecificSourceData(source);
         });
-        await Promise.all(compiledContractPromises);
-
+        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
```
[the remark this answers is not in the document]
abandeali1: Doesn't work with Sets.

packages/deployer/src/compiler.ts
BMillman19: i think i prefer manually calling bind here because its more explicit what the intent is

packages/deployer/src/compiler.ts
fabioberger: Ah no, `_getContractSourcesAsync` returns an empty object, never `undefined`. So the variable here should simply be `this._contractSources` which may be an empty object.

packages/deployer/src/compiler.ts
abandeali1: Hmm, the way this is currently written seems pretty understandable/standard to me. Are we explicitly binding `this` in every single case where we need to throughout our codebase (rather than using arrow functions)?

packages/deployer/src/compiler.ts
fabioberger: Yep, it's checked above, my bad!

packages/deployer/src/compiler.ts
fabioberger: @abandeali1

PR #410, opened 2018-02-21
Translate Landing Page

fabioberger:
## Description

- Translates the landing page into Spanish, Chinese, Korean and Russian
- Move tsconfig dep to devDeps
- Add responsive breakpoints for `vertical-align` `middle` & `top` as well as `table-cell`
- Remove Segment.io from the app

## Motivation and Context

Be more international!

## Types of changes

## Checklist:


packages/website/ts/translations/chinese.ts
```diff
@@ -0,0 +1,77 @@
+import { Key } from 'ts/types';
```
BMillman19: can we make these language files normal json files? That way our internationalization pipeline be something like, new language gets added, then a json wtih all the translated strings gets created via dashboard or something, the dashboard writes json, and those files get imported into the source or pulled down by the app

PR #419, opened 2018-02-27
Improve an error message when an incorrect number of constructor param…

LogvinovLeon:
…s is passed

## Description

Currently if someone forgets and writes code like: `deployer.deployAsync('Exchange')` we just submit a malformed transaction and it fails on the blockchain. We can't detect that on the type-level cause it's dependent on the ABI. This PR checks the number of arguments and throws a somewhat more helpful error message. In future - we might want to also validate a schema generated from the ABI inputs, but I think that's a 80% solution.
screen shot 2018-02-27 at 13 49 42

## Motivation and Context

It was supper annoying for @dekz  when he forgot to pass constructor arguments and we've spent quite a bit of time figuring this out.

## How Has This Been Tested?

I tried to pass a wrong number of args.

## Types of changes

## Checklist:


packages/deployer/src/deployer.ts
```diff
@@ -49,6 +49,18 @@ export class Deployer {
             gas,
         };
         const abi = contractNetworkDataIfExists.abi;
+        const constructorAbi = _.find(abi, { type: AbiType.Constructor }) as Web3.ConstructorAbi;
+        const constructorArgs = _.isUndefined(constructorAbi) ? [] : constructorAbi.inputs;
+        if (constructorArgs.length !== args.length) {
+            const constructorSignature = `constructor(${_.map(constructorArgs, arg => `${arg.type} ${arg.name}`).join(
```
fabioberger: Mh, not the biggest fan of embedding interators *inside* of template strings. Can we move it to the line above for readability?

PR #430, opened 2018-03-03
Upgrade TypeDoc Dep & Add Missing @0xproject/types To Docs

fabioberger:
## Description

- A newer version of TypeDoc exists with support for TS 2.7.0. But during attempting to upgrade, I found a bug that needed to be fixed in TypeDoc (https://github.com/TypeStrong/typedoc/pull/714). So I upgraded us to our own fork of TypeDoc until this gets merged.
- Fixed docs to properly render doc JSON's generated by both the old and new versions of TypeDoc.
- Added `@0xproject/types` to TypeDoc generated docs JSON since types in `@0xproject/types` previously they were missing from our 0x.js documentation.
- Refactored docs so that the types section can render types from both the `0x.js` and `@0xproject/types` packages.
- Added `TransactionReceipt` to list of public types since it was missing! 🤕 

Also includes a small doc UI fix.

## Motivation and Context

More prep for moving docs out of website.

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


LogvinovLeon:
This is great. It feels like we just copy over the changes on postpublish scripts. Could you please remind me why is it that way? Can't we have some shared utils? Also - why are those not TS?

LogvinovLeon:
And the linter fails

fabioberger:
@LogvinovLeon yep, agreed. Scripts should be written in TS and transpiled into JS before the `yarn lerna:publish` command is run. Lerna looks for a JS file called `postpublish.js`. We could probably also reduce code duplicity here. Will create an Airtable task for this.

PR #432, opened 2018-03-04
Allow LedgerSubprovider to return any number of accounts

fabioberger:
## Description

Previously `getAccountsAsync` always returned 10 accounts. Now the caller can specify any number of accounts to fetch. Either way, only a single request is sent to the Ledger device, fetching the public key and chainCode. We then use the key path derivation scheme to generate the requested account addresses locally.

## Motivation and Context

People use more then the first 10 accounts on their Ledger (or they should!), so why not allow developers to fetch them as well.

## How Has This Been Tested?

Added unit and integration tests

## Types of changes

## Checklist:


LogvinovLeon:
LGTM

PR #438, opened 2018-03-06
Extract Docs Template To Separate Package

fabioberger:
## Description

Currently our documentation template react components lived in the website package. Now they have their own dedicated package `react-docs`.  In addition, another package was created called `react-shared` which contains the code shared between the docs package and the website.

I also added a stand-alone example of using `react-doc` to the package and deployed it: http://react-docs-example.s3-website-us-east-1.amazonaws.com/

Other improvements:
- Improved the Documentation component interface by removing a couple more unnecessary props.

## Motivation and Context

There aren't good theme in existence for Doxity and TypeDoc generated docs. So we wanted to make our template available for others in the community.

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


packages/react-docs/example/public/css/basscss_responsive_margin.css
```diff
@@ -0,0 +1,160 @@
...
+  .sm-mt0 { margin-top:    0 }
+  .sm-mr0 { margin-right:  0 }
+  .sm-mb0 { margin-bottom: 0 }
+  .sm-ml0 { margin-left:   0 }
+  .sm-mx0 { margin-left:   0; margin-right:  0 }
+  .sm-my0 { margin-top:    0; margin-bottom: 0 }
+
+  .sm-m1  { margin:        .5rem }
+  .sm-mt1 { margin-top:        .5rem }
+  .sm-mr1 { margin-right:        .5rem }
```
dekz: formatting here is a bit weird

dekz: formatting here is a bit weird

packages/react-docs/example/public/css/roboto.css
```diff
@@ -0,0 +1,83 @@
+@font-face {
+  font-family: 'Roboto';
+  src: url('../fonts/Roboto-Thin.ttf') format('truetype');
+  font-weight: 100;
+  font-style: normal;
+}
+
+/*@font-face {
```
dekz: a few font faces commented out. Remove these?

packages/react-docs/example/ts/docs.tsx
```diff
@@ -0,0 +1,123 @@
...
+    menuSubsectionToVersionWhenIntroduced: {},
+    sections: docSections,
+    visibleConstructors: [docSections.web3Wrapper],
+    typeConfigs: {
+        // Note: This needs to be kept in sync with the types exported in index.ts. Unfortunately there is
+        // currently no way to extract the re-exported types from index.ts via TypeDoc :(
+        publicTypes: ['TxData', 'TransactionReceipt', 'RawLogEntry'],
+        typeNameToExternalLink: {
+            Web3: 'https://github.com/ethereum/wiki/wiki/JavaScript-API',
+            Provider: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L123',
```
dekz: I think this should be [line 150](https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L150)

packages/react-docs/example/ts/docs.tsx
```diff
@@ -0,0 +1,123 @@
...
+            BigNumber: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L127',
+            DecodedLogEntryEvent: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L150',
```
dekz: this should be [line 123](https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L123)

packages/react-docs/example/ts/docs.tsx
```diff
@@ -0,0 +1,123 @@
...
+    }
+    private _onVersionSelected(semver: string) {
+        const selectedDocJSON = versionToDocJSON[semver];
+        this.setState({
+            selectedVersion: semver,
+            docAgnosticFormat: docsInfo.convertToDocAgnosticFormat(selectedDocJSON as TypeDocNode),
+        });
+    }
+    private _getSourceUrl() {
+        const sourceUrl = `${docsInfoConfig.packageUrl}/blob/@0xproject/web3-wrapper%40${
```
dekz: does this need to be url encoded in code or will it work without it and the browser will fix?

packages/react-docs/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@0xproject/react-docs",
+  "version": "0.0.1",
+  "description": "React documentation component for rendering TypeDoc & Doxity generated JSON",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "lint": "tslint --project . 'src/ts/**/*.ts' 'src/ts/**/*.tsx'",
+    "build": "tsc",
+    "build:example": "NODE_ENV=production webpack; exit 0;",
```
dekz: do we want to hide if webpacks to fail? `;` sucks on windows so trying to avoid these

packages/react-docs/.npmignore
```diff
@@ -0,0 +1,5 @@
+.*
```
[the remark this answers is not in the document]
BMillman19: same with screenshot.png

packages/react-docs/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "extends": "../../tsconfig",
+    "compilerOptions": {
+        "outDir": "./lib/",
+        "jsx": "react",
+        "baseUrl": "./",
+        "strictNullChecks": false,
```
LogvinovLeon: It requires a lot of work to fix that in the whole website package, but let's please use strict null checks in this one. It's super usefull. Let's be strict.

packages/react-docs/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+        "noImplicitThis": false,
```
LogvinovLeon: Same here

packages/react-docs/tslint.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": ["@0xproject/tslint-config"],
+    "rules": {
+        "no-implicit-dependencies": false,
```
LogvinovLeon: Why do we have that? Seems dangerous.

packages/react-shared/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "scripts": {
+    "lint": "tslint --project . 'src/ts/**/*.ts' 'src/ts/**/*.tsx'",
+    "build": "tsc",
+    "build:watch": "tsc -w",
+    "clean": "shx rm -rf lib"
+  },
+  "author": "Fabio Berger",
+  "license": "MIT",
```
LogvinovLeon: Apache 2.0

packages/react-docs/example/public/css/roboto.css
fabioberger: These are here because we might want to include them at a later date. Added a comment about this exception to the `no commented out code` policy.

packages/react-docs/example/ts/docs.tsx
fabioberger: Browser can fix.

packages/react-docs/package.json
fabioberger: Luckily we don't need it in this repo 💃 Removed.

packages/react-docs/tsconfig.json
fabioberger: Agreed, this should be a subsequent PR. It's actually a non-trivial amount of work since we do heavily rely on it in these components :/

fabioberger:
@dekz finally fixed the browser history issue with rendering the relevant hash element.

PR #441, opened 2018-03-09
Add web3 provider logging to GA

tomhschmidt:
## Description

Added web3 provider logging to GA

## How Has This Been Tested?

Visit localhost:3572 with Metamask installed and check out Google Analyics. https://analytics.google.com/analytics/web/?authuser=1#report/content-pages/a98720122w145076439p149780311/%3F_u.date00%3D20180308%26_u.date01%3D20180308%26_r.drilldown%3Danalytics.pagePath%3A%2F%26explorer-table.secSegmentId%3Danalytics.customDimension1%26explorer-table.plotKeys%3D%5B%5D/

## Types of changes

## Checklist:


packages/website/ts/index.tsx
```diff
@@ -46,6 +48,12 @@ const LazyConnectDocumentation = createLazyComponent('Documentation', async () =
 );
 
 ReactGA.initialize('UA-98720122-1');
+window.onload = () => {
+    const providerName = (window as any).web3
+        ? utils.web3ProviderToString((window as any).web3.currentProvider)
```
dekz: I think you want to wrap this and check if web3 is defined or not. `!_.isUndefined((window as any).web3` before trying to access `currentProvider` and ``currentProvider.constructor.name`

packages/website/ts/utils/utils.ts
```diff
@@ -309,6 +309,17 @@ export const utils = {
             containerId,
         });
     },
+    web3ProviderToString(provider: Web3Provider): string {
+        let parsedProviderName = provider.constructor.name;
+        if (provider.constructor.name === 'MetamaskInpageProvider') {
+            parsedProviderName = 'METAMASK';
+        } else if (provider.constructor.name === 'EthereumProvider') {
+            parsedProviderName = 'MIST';
+        } else if (provider.constructor.name === 'o') {
```
dekz: Is this because its minified? Will this change on different versions? What does the Cipher Browser constructor look like?

tomhschmidt: Yeah, I just re-tested with a fresh parity install and it's different :/ -- turns out there's just a isParity var. Going to use that instead

tomhschmidt: Not sure about Cipher. Is there any easy way to test it other than spinning up a new web page for it to hit?

packages/website/ts/utils/utils.ts
```diff
@@ -309,6 +309,17 @@ export const utils = {
...
     },
+    web3ProviderToString(provider: Web3Provider): string {
+        let parsedProviderName = provider.constructor.name;
+        if (provider.constructor.name === 'MetamaskInpageProvider') {
+            parsedProviderName = 'METAMASK';
+        } else if (provider.constructor.name === 'EthereumProvider') {
+            parsedProviderName = 'MIST';
+        } else if ((provider as any).isParity) {
+            parsedProviderName = 'PARITY';
+        }
```
dekz: Maybe add a default case of `UNKNOWN` to imply there is a web3 but we don't know what it is (yet)

tomhschmidt: I originally had that, but took it out and replaced it with just the constructor name. I feel like gives us more info and still implies that "UNKNOWN" case

dekz: Yeah I'm just worried about anyone else who is minified, might end up with a bunch of `o`, `a` etc.

tomhschmidt: I feel you. I think we can just map those to "UNKNOWN" offline and we might also capture some understandable constructor names (e.g. MetamaskInpageProvider). I'm down to change this though if ppl think it's better to have 'UNKNOWN' be mapped explicitly.

fabioberger:
@BMillman19 @dekz please review PR's more closely. There were several non-trivial issues with this one.

PR #443, opened 2018-03-09
Move `react-doc` example to separate sub-package

fabioberger:
## Description

In order to have a realistic getting-started example of how to use the react-docs package, it needed to be in a separate package and import the react-docs package as a dependency. This PR does this.

## Types of changes

## Checklist:


packages/react-docs-example/README.md
```diff
@@ -0,0 +1,55 @@
+## @0xproject/react-docs-example
+
+An [example documentation page](http://react-docs-example.s3-website-us-east-1.amazonaws.com/) built using [react-docs](https://github.com/0xProject/0x-monorepo/tree/development/packages/react-docs) rendering the [@0xproject/web3-wrapper](https://github.com/0xProject/0x-monorepo/tree/development/packages/web3-wrapper) Typescript package. This is a great starter project for hosting your own Solidity or Typescript documentation page.
+
+## Usage
+
+This package is intended as a boilerplate for creating and hosting your own documentation page. Copy it's contents into a new directory to get started.
```
LogvinovLeon: Maybe clone instead of copy? Or even fork?

packages/react-docs-example/package.json
```diff
@@ -0,0 +1,59 @@
...
+        "An example app using react-docs",
+    "scripts": {
+        "lint": "tslint --project . 'ts/**/*.ts' 'ts/**/*.tsx'",
+        "build": "tsc",
+        "build:example": "NODE_ENV=production webpack",
+        "build:watch": "tsc -w",
+        "clean": "shx rm -rf lib; shx rm -f public/bundle*",
+        "dev": "webpack-dev-server --open",
+        "deploy_example":
+            "npm run build:example; aws s3 sync ./public/. s3://react-docs-example --profile 0xproject --region us-east-1 --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers"
```
LogvinovLeon: If it's intended to be forked by other people - I'll remove 0x-specific configs like region of profile name, but I also see it being easier for us and they'll probably be smart enough to figure it out.

packages/react-docs-example/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+    "compilerOptions": {
```
LogvinovLeon: Why don't we extend the base config here?

packages/react-docs-example/package.json
fabioberger: Yeah, think we'll keep it for now. Mentioned they'd need to update it in the README.

PR #444, opened 2018-03-09
Remove custom web3Wrapper from website

fabioberger:
## Description

- Removes custom web3Wrapper from website (@0xproject/web3-wrapper)
- Refactors Portal to pass around Ether amounts in wei, and to only convert to Ether before displaying an amount.
- Removed unnecessary `exchangeAddress` instance variable from `blockchain.ts`.

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/blockchain.ts
+++ b/packages/website/ts/blockchain.ts
@@ -531,16 +531,14 @@ export class Blockchain {
             !_.isUndefined(this.networkId),
             'Cannot call fetchTokenInformationAsync if disconnected from Ethereum node',
         );
-        utils.assert(this._doesUserAddressExist(), BlockchainCallErrs.UserHasNoAssociatedAddresses);
 
         this._dispatcher.updateBlockchainIsLoaded(false);
 
         const tokenRegistryTokensByAddress = await this._getTokenRegistryTokensByAddressAsync();
 
-        const trackedTokensByAddress = trackedTokenStorage.getTrackedTokensByAddress(
-            this._userAddressIfExists,
-            this.networkId,
-        );
+        const trackedTokensByAddress = _.isUndefined(this._userAddressIfExists)
+            ? {}
+            : trackedTokenStorage.getTrackedTokensByAddress(this._userAddressIfExists, this.networkId);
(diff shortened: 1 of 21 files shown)


packages/website/ts/blockchain.ts
```diff
@@ -543,6 +554,13 @@ export class Blockchain {
 
         this._dispatcher.updateBlockchainIsLoaded(true);
     }
+    public async getFirstAccountIfExistsAsync() {
```
dekz: This looks like it could be useful inside of web3 wrapper itself. Further down this pattern is used:
```typescript
const addresses = await this._web3Wrapper.getAvailableAddressesAsync()
const userAddressIfExists = addresses[0] || '';
```
Since this isn't a UI layer, how do you feel about not returning an empty string but null or undefined?

BMillman19: the 'IfExists' portion of the method name suggests that `undefined` will be returned if we don't find an address

packages/website/ts/components/token_balances.tsx
```diff
@@ -241,7 +248,7 @@ export class TokenBalances extends React.Component<TokenBalancesProps, TokenBala
                                 <img style={{ width: ICON_DIMENSION, height: ICON_DIMENSION }} src={ETHER_ICON_PATH} />
                             </TableRowColumn>
                             <TableRowColumn>
-                                {this.props.userEtherBalance.toFixed(PRECISION)} ETH
+                                {userEtherBalanceInEth.toFixed(PRECISION)} ETH
```
dekz: I think we have PRECISION defined in a few different files, always set to `5`. Move this into constants as `ETH_DISPLAY_PRECSION`

packages/website/ts/blockchain.ts
fabioberger: Once we made this method return undefined when no addresses are found, it was no longer needed. Now we simply do:

```
const userAddresses = await this._web3Wrapper.getAvailableAddressesAsync();
this._userAddressIfExists = userAddresses[0];
```

PR #447, opened 2018-03-10
Refactor Website Analytics

fabioberger:
…e, combining the provider type util function, moving GA id to configs and using utils.onPageLoadAsync

## Description

- Move all calls to ReactGA into a single module
- Merge disparate provider name logic to use single utils method
- Move Google analytics id to configs
- Re-use utils.onPageLoadAsync

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/utils/utils.ts
+++ b/packages/website/ts/utils/utils.ts
@@ -4,10 +4,11 @@ import { BigNumber } from '@0xproject/utils';
 import deepEqual = require('deep-equal');
 import * as _ from 'lodash';
 import * as moment from 'moment';
-import { Order, ScreenWidths, Side, SideToAssetToken, Token, TokenByAddress } from 'ts/types';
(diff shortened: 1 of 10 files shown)


packages/website/ts/utils/utils.ts
```diff
@@ -287,4 +277,25 @@ export const utils = {
...
+
+            case 'EthereumProvider':
+                parsedProviderName = Providers.Mist;
+                break;
+
+            default:
+                parsedProviderName = constructorName;
+                break;
+        }
+        if ((provider as any).isParity) {
```
dekz: If we're moving away from constructor names, metamask has the equivalent.

```
window.web3.currentProvider.isMetaMask
true
```

fabioberger: Yep, they both work. I guess potentially the `isMetaMask` flag is less likely to change then the constructor now. Will update.

PR #469, opened 2018-03-20
Generalized Asset Transfer Proxy

hysz:
## Description

A generalized asset transfer proxy, with an implementation for ERC20 tokens. This has been hooked up to the exchange's settlement layer.

Included in this PR:
- Asset transfer proxy, which delegates transfers to specialized transfer proxies (currently ERC20)
- Specialized transfer proxies can be hotswapped (registered/unregistered on the fly).
- Settlement layer now uses the asset transfer proxy to delegate transfers
- Order now includes makerAssetProxyId and takerAssetProxyId
- Updated fillOrderNoThrow to account for changes to the Order structure
- The exchange constructor now takes the asset transfer proxy, rather than a token transfer proxy.
- Authorizable contract (implementation abstracted from TokenTransferProxy)
- Memory util, which includes copying an address to/from a position in a byte array

Todo:
- I was unable to import the LibErrors due to a conflicting declaration error. I am looking into this now.
-  All existing tests pass, but I need to add new tests specific to the transfer proxy and its edge cases.

## Motivation and Context

We would like to add support for transferring new types of assets. At present, we only support ERC20 tokens. These changes lay the foundation for extending the exchange to new asset classes, like ERC721.

## How Has This Been Tested?

All existing tests pass.

## Types of changes

## Checklist:


packages/contracts/src/utils/types.ts
```diff
@@ -38,6 +38,8 @@ export interface DefaultOrderParams {
     takerTokenAmount: BigNumber;
     makerFeeAmount: BigNumber;
     takerFeeAmount: BigNumber;
+    makerAssetProxyId: number;
```
hysz: Is there a better type to use than `number`? This will get downcast to `uint8`.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/ERC20TransferProxy.sol
```diff
@@ -0,0 +1,57 @@
...
+        bytes assetMetadata,
+        address from,
+        address to,
+        uint256 amount)
+        public
+        onlyAuthorized
+        returns (bool success)
+    {
+        address token = decodeERC20Metadata(assetMetadata);
+        return TRANSFER_PROXY.transferFrom(token, from, to, amount);
```
hysz: @abandeali1 can I remove the `ITokenTransferProxy` contract and call the ERC20 transfer directly?

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
```diff
@@ -48,6 +50,8 @@ contract LibOrder {
         uint256 takerFeeAmount;
         uint256 expirationTimeSeconds;
         uint256 salt;
+        uint8 makerAssetProxyId;
```
hysz: @abandeali1 - I added these new fields to the end of the struct so it would be easier to extend functions that access fields in assembly, like in `MixinWrapperFunctions.sol`. Although I can reorder so they're higher, closer to the maker/taker token addresses if that would make more sense.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetProxyEncoderDecoder.sol
```diff
@@ -0,0 +1,98 @@
...
+    /// @dev Encodes order's taker metadata into a byte array for the respective asset proxy.
+    /// @param takerAssetProxyId Id of the asset proxy.
+    /// @param takerTokenAddress Address of the asset.
+    /// @param orderHash Hash of order.
+    /// @return assetMetadata Byte array encoded for the respective asset proxy.
+    function encodeTakerMetadata(uint8 takerAssetProxyId, address takerTokenAddress, bytes32 orderHash)
+        public pure
+        returns (bytes assetMetadata)
+    {
+        if(AssetProxyId(takerAssetProxyId) == AssetProxyId.ERC20) {
```
LogvinovLeon: space

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+import "./AssetProxyEncoderDecoder.sol";
+import "./IAssetProxy.sol";
+import "../../utils/Authorizable/Authorizable.sol";
+
+contract AssetTransferProxy is
+    AssetProxyEncoderDecoder,
+    Authorizable
+{
+    // Mapping from Asset Proxy Id's to their respective Asset Proxy
+    mapping (uint8 => IAssetProxy) public assetProxies_;
```
LogvinovLeon: Why not `assetProxies`?

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+
+    // Logs registration of new asset proxy
+    event LogAssetProxyRegistration(
+        AssetProxyId id,
+        address assetClassAddress,
+        bool overwrite,
+        bool did_overwrite
```
LogvinovLeon: Camel case

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+        bool overwrite)
+        public
+        onlyAuthorized
+    {
+        // Convert assetProxyId to mapping id
+        require(uint256(assetProxyId) < 256);
+        uint8 id = uint8(assetProxyId);
+
+        // Ensure an existing asset proxy is not unintentionally overwritten
+        bool will_overwrite = (assetProxies_[id] != address(0x0));
```
LogvinovLeon: camelCase

packages/contracts/src/contracts/current/utils/Authorizable/Authorizable.sol
```diff
@@ -0,0 +1,94 @@
...
+    /// @dev Authorizes an address.
+    /// @param target Address to authorize.
+    function addAuthorizedAddress(address target)
+        public
+        onlyOwner
+        targetNotAuthorized(target)
+    {
+        authorized[target] = true;
+        authorities.push(target);
+        LogAuthorizedAddressAdded(target, msg.sender);
```
LogvinovLeon: emit

packages/contracts/src/contracts/current/utils/Authorizable/Authorizable.sol
```diff
@@ -0,0 +1,94 @@
...
+    {
+        delete authorized[target];
+        for (uint i = 0; i < authorities.length; i++) {
+            if (authorities[i] == target) {
+                authorities[i] = authorities[authorities.length - 1];
+                authorities.length -= 1;
+                break;
+            }
+        }
+        LogAuthorizedAddressRemoved(target, msg.sender);
```
LogvinovLeon: emit

packages/contracts/src/utils/types.ts
LogvinovLeon: In JS - no.

LogvinovLeon: Maybe a union type? 1|2|3?

packages/contracts/package.json
```diff
@@ -18,7 +18,7 @@
...
         "clean": "shx rm -rf ./lib",
-        "generate_contract_wrappers": "node ../abi-gen/lib/index.js --abis 'src/artifacts/@(DummyToken|TokenTransferProxy|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken|Arbitrage|EtherDelta|AccountLevels).json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers/generated --backend ethers && prettier --write 'src/contract_wrappers/generated/**.ts'",
+        "generate_contract_wrappers": "node ../abi-gen/lib/index.js --abis 'src/artifacts/@(DummyToken|TokenTransferProxy|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken|Arbitrage|EtherDelta|AccountLevels|AssetTransferProxy|ERC20TransferProxy).json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers/generated --backend ethers && prettier --write 'src/contract_wrappers/generated/**.ts'",
```
LogvinovLeon: Extract the list of abis to config

packages/contracts/src/contracts/current/utils/Authorizable/Authorizable.sol
```diff
@@ -0,0 +1,94 @@
...
+        public
+        onlyOwner
+        targetAuthorized(target)
+    {
+        delete authorized[target];
+        for (uint i = 0; i < authorities.length; i++) {
+            if (authorities[i] == target) {
+                authorities[i] = authorities[authorities.length - 1];
+                authorities.length -= 1;
+                break;
```
fabioberger: Will this approach scale? Will we reach some gasLimit if we have a bunch of `authorized` addresses?

packages/contracts/src/utils/types.ts
fabioberger: I would use a custom enum type:

```
enum AssetProxyIds {
  InValid,
  ERC20,
  ERC721,
  ...
}
```

These map to 0, 1, 2, ... so the first enum value should be set to `Invalid` if the id's start with `1` in the smart contracts.

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
```diff
@@ -0,0 +1,86 @@
...
+
+pragma solidity ^0.4.21;
+
+contract Memory {
+
+    /// @dev Gets an address from a position in a byte array.
+    /// @param b Byte array containing an address.
+    /// @param index Index in byte array of address.
+    /// @return address from byte array.
+    function getAddress(
```
Recmo: I prefer calling them `read...` and `write...`. `get...` is used often for accessors for state variables.

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
```diff
@@ -0,0 +1,86 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.4.21;
+
+contract Memory {
```
Recmo: We can call it `LibMemory` to emphasize that it is just a collection of pure functions. (Like `LibOrder`)

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetProxy.sol
```diff
@@ -0,0 +1,38 @@
...
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
+    /// @return Success of transfer.
+    function transferFrom(
+        bytes assetMetadata,
+        address from,
+        address to,
+        uint256 amount)
+        public
+        returns (bool success);
```
Recmo: We should also think about batch versions of these commands. We don't have to solve it in this PR though.

Recmo: And balance + allowance checks.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetProxyEncoderDecoder.sol
```diff
@@ -0,0 +1,98 @@
...
+contract AssetProxyEncoderDecoder is
+    Memory
+{
+    // Supported asset proxies
+    enum AssetProxyId {
+        ERC20,
+        ERC721,
+        ENS,
+        OWNABLE,
+        ACCEPT_OWNERSHIP
```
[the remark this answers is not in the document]
hysz: Yeah I think that's fine.

I put the full list in so that we wouldn't have to redeploy the AssetTransferProxy contract each time we add one of these proxies. Although after chatting with Amir, I think we're going to change AssetTransferProxy's API to not depend on this enum at all.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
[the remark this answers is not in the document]
hysz: The trailing underscore denotes a state variable, although I see it's not used elsewhere in our contracts so I'll remove it.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+
+    // Logs registration of new asset proxy
+    event LogAssetProxyRegistration(
+        AssetProxyId id,
+        address assetClassAddress,
+        bool overwrite,
+        bool did_overwrite
+    );
+
+    // Logs unregistration of an existing asset proxy
```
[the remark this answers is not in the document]
hysz: Lol +1

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+    event LogAssetProxyUnregistration(
+        AssetProxyId id,
+        address assetClassAddress
+    );
```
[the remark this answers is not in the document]
hysz: Good call +1

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+        returns (bool)
+    {
+        // Lookup asset proxy
+        require(assetMetadata.length >= 1);
+        uint8 id = uint8(assetMetadata[0]);
+        IAssetProxy assetProxy = assetProxies_[id];
+        require(assetProxy != address(0x0));
+
+        // Delegate transfer to asset proxy
+        return assetProxy.transferFrom(assetMetadata, from, to, amount);
```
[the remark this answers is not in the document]
hysz: Using a byte array allows us to encode arbitrary metadata. For ERC20 all we need is a the id and address, but this could be different for other standards.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+
+        // Return asset proxy
+        return assetProxy;
+    }
+
+    /// @dev Unregisters an asset proxy.
+    /// @param assetProxyId Id of the asset proxy to unregister.
+    function unregisterAssetProxy(AssetProxyId assetProxyId)
+        public
+        onlyAuthorized
```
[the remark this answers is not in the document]
hysz: I like that. I'll update the API to use this.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+    {
+        // Convert assetProxyId to mapping id
+        require(uint256(assetProxyId) < 256);
+        uint8 id = uint8(assetProxyId);
+
+        // Ensure an existing asset proxy is not unintentionally overwritten
+        bool will_overwrite = (assetProxies_[id] != address(0x0));
+        require(overwrite || !will_overwrite);
```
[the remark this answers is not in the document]
hysz: I like that. I'll update the API to use this.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetProxy.sol
hysz: I was actually just chatting with @abandeali1 about doing something like this in the next iteration. I'll make a note.

packages/contracts/src/contracts/current/utils/Authorizable/Authorizable.sol
```diff
@@ -0,0 +1,94 @@
...
+
+    /// @dev Removes authorizion of an address.
+    /// @param target Address to remove authorization from.
+    function removeAuthorizedAddress(address target)
+        public
+        onlyOwner
+        targetAuthorized(target)
+    {
+        delete authorized[target];
+        for (uint i = 0; i < authorities.length; i++) {
```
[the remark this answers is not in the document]
hysz: Good stuff! :)

I've added this method to the contract, but left the old one as well for backwards-compatibility.

packages/contracts/src/contracts/current/utils/Authorizable/Authorizable.sol
hysz: This already existed in the codebase; I just abstracted it out of ITokenTransferProxy so it could be used by other contracts. I didn't look too closely at it, although @Recmo had the same thought and posted a great solution above 
 which I have incorporated.

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
hysz: I agree. Wasn't sure on the convention here, since other utilities (like SafeMath) don't have the Lib prefix. @abandeali1 thoughts?

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
hysz: +1 agreed

packages/contracts/src/utils/types.ts
hysz: I like the Enum approach. It can match the Enum used in the contracts.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
hysz: The trailing underscore denotes a state variable, although I see it's not used elsewhere in our contracts so I'll remove it.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -99,6 +101,18 @@ contract MixinExchangeCore is
         }
         require(takerTokenFillAmount > 0);
 
+        // Validate maker asset proxy
+        if (!isValidAssetProxyId(order.makerAssetProxyId)) {
```
abandeali1: These checks should be made in `AssetTransferProxy` using the mapping rather than the enum (I'd remove the enum altogether). The goal is to be able to add new proxies _without_ swapping out the Exchange contract.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -68,17 +69,20 @@ contract MixinSettlementProxy is
         )
     {
         makerTokenFilledAmount = getPartialAmount(takerTokenFilledAmount, order.takerTokenAmount, order.makerTokenAmount);
+        bytes32 orderHash = getOrderHash(order);
+
         require(
             TRANSFER_PROXY.transferFrom(
-                order.makerTokenAddress,
+                encodeMetadata(order.makerAssetProxyId, order.makerTokenAddress),
```
abandeali1: I think we'll find that this will need to be made more generic as we start experimenting with different proxies for different standards.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -87,9 +91,10 @@ contract MixinSettlementProxy is
         if (order.feeRecipientAddress != address(0)) {
             if (order.makerFeeAmount > 0) {
                 makerFeeAmountPaid = getPartialAmount(takerTokenFilledAmount, order.takerTokenAmount, order.makerFeeAmount);
+
                 require(
                     TRANSFER_PROXY.transferFrom(
-                        ZRX_TOKEN,
+                        encodeERC20Metadata(ZRX_TOKEN),
```
abandeali1: I'm don't think we want to hard code ERC20 encoding here. I could see us having different proxies that are compatible with the same tokens.

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
abandeali1: I like the `Lib` prefix. Maybe `LibBytes` is a better name?

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
```diff
@@ -0,0 +1,86 @@
...
+    {
+        require(b.length >= index + 20); // 20 is length of address
+
+        // Add offset to index:
+        // 1. Arrays are prefixed by 32-byte length parameter (add 32 to index)
+        // 2. Account for size difference between address length and 32-byte storage word (subtract 12 from index)
+        index += 20;
+
+        // stores additional bytes that occupy the 32-byte word where we'll be storing the address
+        bytes32 neighbors;
```
abandeali1: Any reason this is declared outside of the assembly?

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,122 @@
...
+    function registerAssetProxy(
+        AssetProxyId assetProxyId,
+        address newAssetProxyAddress,
+        address currentAssetProxyAddress)
+        public
+        onlyAuthorized
+    {
+        // Convert assetProxyId to mapping id
+        require(uint256(assetProxyId) < 256);
+        uint8 id = uint8(assetProxyId);
```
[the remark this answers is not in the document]
hysz: +1 This will also allow us to add new proxies without redeploying the contract.

packages/contracts/src/contracts/current/utils/Memory/Memory.sol
hysz: I tried to do this but got a compiler error. Taking another look, I see you actually can declare assembly-local variables using `let`.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetProxyEncoderDecoder.sol
```diff
@@ -0,0 +1,58 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.4.21;
+
+contract IAssetProxyEncoderDecoder {
+
+    // Supported asset proxies
+    enum AssetProxyId {
```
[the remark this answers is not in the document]
hysz: +1 I've removed this Enum and updated the encoder/decoder to be more flexible.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
hysz: +1 I've removed this Enum and updated the encoder/decoder to be more flexible.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetProxyEncoderDecoder.sol
```diff
@@ -0,0 +1,59 @@
...
+contract AssetProxyEncoderDecoder is
+    LibBytes,
+    IAssetProxyEncoderDecoder
+{
+
+    /// @dev Encodes an order's metadata into a byte array for the respective asset proxy.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param tokenAddress Address of the asset.
+    /// @return assetMetadata Byte array encoded for the respective asset proxy.
+    function encodeMetadata(
```
hysz: @abandeali1 We'll be able to remove this function after moving encoded metadata into the order. The settlement layer will simply pass along the encoded metadata right through to the destination proxy.

abandeali1:
Can you change the base branch of this PR to `v2-proxy` and rebase? Looks good to merge after that.

hysz:
@abandeali1 Updated to `v2-proxy`

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /// @dev Registers a new asset proxy.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param assetProxyAddress address of the asset proxy contract.
+    /// @param overwrite Any existing asset proxy registered with the same assetProxyId will be overwritten.
+    function registerAssetProxy(
+        AssetProxyId assetProxyId,
+        address assetProxyAddress,
+        bool overwrite)
```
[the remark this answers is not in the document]
hysz: Updated to take the current address and overwrite only if it's the same, as per @Recmo's suggestion.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,132 @@
...
+    Authorizable
+{
+    // Mapping from Asset Proxy Id's to their respective Asset Proxy
+    mapping (uint8 => IAssetProxy) public assetProxies_;
+
+    // Logs registration of new asset proxy
+    event LogAssetProxyRegistration(
+        AssetProxyId id,
+        address assetClassAddress,
+        bool overwrite,
```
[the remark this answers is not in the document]
hysz: This boolean has been removed.

packages/contracts/src/contract_wrappers/generated/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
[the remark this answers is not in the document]
hysz: Reverted. I must have remove the generated folder, not realizing there was a tracked file.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
hysz: Updated to pass the erc20 proxy id into constructor.

PR #470, opened 2018-03-21
Fix a bug causing fillOrdersUpTo validation to fail because of some e…

### LINKED ISSUE #466: fillOrdersUpToAsync validation fails if fill amount for any order is zero
## Expected Behavior

Calling `zeroEx.exchange.fillOrdersUpToAsync(...)` should not throw in validation phase if any of the passed signed orders don't get filled at all.

## Current Behavior

Currently, when validation is enabled (default), `fillOrdersUpToAsync` calls `validateFillOrderThrowIfInvalidAsync`: https://github.com/0xProject/0x-monorepo/blob/7d9c357/packages/0x.js/src/contract_wrappers/exchange_wrapper.ts#L276

That piece of code throws if `fillTakerTokenAmount` is `0`: https://github.com/0xProject/0x-monorepo/blob/7d9c357/packages/0x.js/src/utils/order_validation_utils.ts#L161

However, this is not how the `fillOrdersUpTo` function of the smart contract works. The smart contract simply breaks the loop whenever all `fillTakerTokenAmount` becomes `0` and leaves rest of the orders alone: https://github.com/0xProject/0x-monorepo/blob/12d62e1157168ec9e8cd23749523b2bcda8eefe5/packages/contracts/src/contracts/current/protocol/Exchange/Exchange.sol#L394

Validation should work the same way. Passing extraneous orders to the function seems like a perfectly valid use case.

## Possible Solution

Break out of the validation loop whenever `fillTakerTokenAmount` becomes zero, or parameterize `validateFillOrderThrowIfInvalidAsync` so that it doesn't check for the above condition.

## Context

I'm trying to simulate market orders client-side.

## Your Environment

|             Package | Version |
| ------------------: | :------ |
|             `0x.js` | 0.32.4  |
| `Exchange Contract` | v1      |

LogvinovLeon:
…xtra orders being passed

## Description

Fixes #466 

## How Has This Been Tested?

Added a regression test

## Types of changes

## Checklist:


packages/0x.js/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## v0.34.0 - _TBD_
 
     * Update Kovan EtherToken artifact address to match TokenRegistry.
+    * Fix the bug causing `zeroEx.exchange.fillOrdersUpToAsync` validation to fail if there were some extra orders passed (#TBD)
```
fabioberger: Please update the PR number.

PR #478, opened 2018-03-23
Remove partial cancels

abandeali1:
## Description

This PR removes the ability to partially cancel an order.

## Motivation and Context

Allowing partial cancels makes it difficult for a relayer to use a flat fee model. For example, a market maker can place a very large order and then cancel the majority of it in order to reduce fees.

Instead of partially cancelling an order, a trader must now cancel the entire order and then create a new order for the remainder (which is still as gas efficient as before).

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -85,9 +83,15 @@ contract MixinExchangeCore is
         // Compute the order hash
         bytes32 orderHash = getOrderHash(order);
 
+        // Check if order has been cancelled
+        if (cancelled[orderHash]) {
+            LogError(uint8(Errors.ORDER_FULLY_FILLED_OR_CANCELLED), orderHash);
```
hysz: Now that it's easier to differentiate between the fully filled and cancelled states, would it be worthwhile to separate `ORDER_FULLY_FILLED_OR_CANCELLED` into `ORDER_FULLY_FILLED` and `ORDER_CANCELLED`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -41,7 +41,7 @@ contract MixinExchangeCore is
 {
     // Mappings of orderHash => amounts of takerTokenAmount filled or cancelled.
     mapping (bytes32 => uint256) public filled;
-    mapping (bytes32 => uint256) public cancelled;
+    mapping (bytes32 => bool) public cancelled;
```
Recmo: I can update [PR428](https://github.com/0xProject/0x-monorepo/pull/428) to optimize this into a single var.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -106,10 +110,10 @@ contract MixinExchangeCore is
         }
 
         // Validate order availability
-        uint256 remainingTakerTokenAmount = safeSub(order.takerTokenAmount, getUnavailableTakerTokenAmount(orderHash));
+        uint256 remainingTakerTokenAmount = safeSub(order.takerTokenAmount, filled[orderHash]);
         takerTokenFilledAmount = min256(takerTokenFillAmount, remainingTakerTokenAmount);
         if (takerTokenFilledAmount == 0) {
```
Recmo: If we change this to `remainingTakerTokenAmount == 0` we no don't depend on `require(takerTokenFillAmount > 0)` to keep this condition correct.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -144,47 +148,39 @@ contract MixinExchangeCore is
...
 
-        uint256 remainingTakerTokenAmount = safeSub(order.takerTokenAmount, getUnavailableTakerTokenAmount(orderHash));
-        takerTokenCancelledAmount = min256(takerTokenCancelAmount, remainingTakerTokenAmount);
-        if (takerTokenCancelledAmount == 0) {
-            LogError(uint8(Errors.ORDER_FULLY_FILLED_OR_CANCELLED), orderHash);
-            return 0;
+        if (cancelled[orderHash]) {
+            LogError(uint8(Errors.ORDER_CANCELLED), orderHash);
+            return false;
         }
```
Recmo: I'm trying to understand what a return value `true` means. It doesn't mean "the order is unfillable", because we return false when the order is expired or already cancelled. Besides, after calling cancel the order is always unfillable.

It could mean "the order changed from fillable to unfillable". But then we also need to handle `Errors.ORDER_FULLY_FILLED`. This would require an extra state read though.

abandeali1: As is it's "the order changed from not cancelled to cancelled". Definitely a bit awkward though.

Recmo: Maybe we can adjust the docs to read:

```
/// After calling, the order can not be filled anymore.
/// @returns true if the order state changed to Cancelled. false if the transaction was already cancelled or othwerise unfillable.
```

Success just seems a bit confusing, this call never fails in the sense of taking an order down.

PR #482, opened 2018-03-26
Move common types out of web3 types

LogvinovLeon:
## Description
Moves types that are common ethereum types to our types package out of web3 types.

## Motivation and Context

This allows most of the packages to drop web3 as a dependency.

## How Has This Been Tested?

It compiles

## Types of changes

## Checklist:

### DIFF
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -1,16 +1,15 @@
-import { TxData, TxDataPayable } from '@0xproject/types';
+import { ContractAbi, DataItem, TxData, TxDataPayable } from '@0xproject/types';
 import { Web3Wrapper } from '@0xproject/web3-wrapper';
 import * as ethersContracts from 'ethers-contracts';
 import * as _ from 'lodash';
-import * as Web3 from 'web3';
 
 export class BaseContract {
     protected _ethersInterface: ethersContracts.Interface;
     protected _web3Wrapper: Web3Wrapper;
-    public abi: Web3.ContractAbi;
+    public abi: ContractAbi;
     public address: string;
     protected static _transformABIData(
-        abis: Web3.DataItem[],
+        abis: DataItem[],
         values: any[],
         transformation: (type: string, value: any) => any,
     ): any {
@@ -46,20 +45,20 @@ export class BaseContract {
         // 2. Global config passed in at library instantiation
         // 3. Gas estimate calculation + safety margin
         const removeUndefinedProperties = _.pickBy;
-        const txDataWithDefaults = {
+        const txDataWithDefaults = ({
             to: this.address,
             ...removeUndefinedProperties(this._web3Wrapper.getContractDefaults()),
             ...removeUndefinedProperties(txData as any),
             // HACK: TS can't prove that T is spreadable.
             // Awaiting https://github.com/Microsoft/TypeScript/pull/13288 to be merged
-        };
+        } as any) as TxData;
         if (_.isUndefined(txDataWithDefaults.gas) && !_.isUndefined(estimateGasAsync)) {
             const estimatedGas = await estimateGasAsync(txData);
             txDataWithDefaults.gas = estimatedGas;
         }
         return txDataWithDefaults;
     }
-    constructor(web3Wrapper: Web3Wrapper, abi: Web3.ContractAbi, address: string) {
+    constructor(web3Wrapper: Web3Wrapper, abi: ContractAbi, address: string) {
         this._web3Wrapper = web3Wrapper;
         this.abi = abi;
         this.address = address;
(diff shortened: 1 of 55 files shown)


packages/base-contract/src/index.ts
```diff
@@ -46,20 +45,20 @@ export class BaseContract {
...
         const removeUndefinedProperties = _.pickBy;
-        const txDataWithDefaults = {
+        const txDataWithDefaults = ({
             to: this.address,
             ...removeUndefinedProperties(this._web3Wrapper.getContractDefaults()),
             ...removeUndefinedProperties(txData as any),
             // HACK: TS can't prove that T is spreadable.
             // Awaiting https://github.com/Microsoft/TypeScript/pull/13288 to be merged
-        };
+        } as any) as TxData;
```
fabioberger: Why not: `const txDataWithDefaults: TxData = ...`?

packages/react-docs-example/ts/docs.tsx
```diff
@@ -57,8 +57,8 @@ const docsInfoConfig: DocsInfoConfig = {
                 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L314',
             LogEntry: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L366',
             FilterObject: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L109',
-            ['Web3.BlockParam']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L278',
-            ['Web3.ContractAbi']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L47',
+            ['BlockParam']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L278',
+            ['ContractAbi']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L47',
```
fabioberger: If these are no longer using `web3-typescript-typings`, remove them entirely from this list and add them under `publicTypes` above.

packages/react-docs-example/ts/json/0.1.12.json
```diff
@@ -2288,7 +2288,7 @@
                                                     },
                                                     {
                                                         "type": "reference",
-                                                        "name": "Web3.BlockParam"
+                                                        "name": "BlockParam"
```
fabioberger: Don't update these. They are frozen to what `web3-wrapper` looked like at these earlier versions. They can stay unchanged.

packages/react-docs-example/ts/json/0.2.0.json
```diff
@@ -2177,7 +2177,7 @@
                                             },
                                             "type": {
                                                 "type": "reference",
-                                                "name": "Web3.BlockParam"
+                                                "name": "BlockParam"
```
fabioberger: Ditto all these. Revert.

packages/website/ts/containers/web3_wrapper_documentation.ts
```diff
@@ -59,8 +59,8 @@ const docsInfoConfig: DocsInfoConfig = {
                 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L314',
             LogEntry: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L366',
             FilterObject: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L109',
-            ['Web3.BlockParam']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L278',
-            ['Web3.ContractAbi']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L47',
+            ['BlockParam']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L278',
+            ['ContractAbi']: 'https://github.com/0xProject/web3-typescript-typings/blob/f5bcb96/index.d.ts#L47',
```
fabioberger: Ditto above. Remove entirely here and add to `publicTypes` array above.

fabioberger: What above the others in the array? Were any of them moved?

packages/base-contract/src/index.ts
LogvinovLeon: Because this type is not directly convertible to TxData.
In line 51 we convert txData to any and loose the info that from is defined. If I try to directly assign - it complaints that from might be undefined. So I first convert to any and then to txData.

PR #485, opened 2018-03-28
Add metacoin example project

LogvinovLeon:
## Description

Thia PR adds an example Metacoin project that showcases 0x dev-tools (deployer, sol-cov, dev-utils, utils, abi-gen, contract_templates, GanacheSubprovider)

## Motivation and Context

It's good to create a minimal starter project for smart contracts devs. I've found quite a lot of places that could be improved and improved then when tried to use those tools as an external dev.

## How Has This Been Tested?

tests

## Types of changes

## Checklist:

*   [] Added new entries to the relevant CHANGELOGs.


packages/base-contract/src/index.ts
```diff
@@ -1,40 +1,31 @@
...
-            BaseContract._transformTypedData(abis[i].type, value, transformation),
-        );
+        return _.map(values, (value: any, i: number) => formatABIDataItem(abis[i], value, formatter));
     }
     protected static _lowercaseAddress(type: string, value: string): string {
         return type === 'address' ? value.toLowerCase() : value;
     }
-    protected static _bigNumberToString(type: string, value: string): string {
-        return _.isObject(value) && (value as any).isBigNumber ? value.toString() : value;
+    protected static _bigNumberToString(type: string, value: any): any {
```
fabioberger: Why not `value: BigNumber): BigNumber`?

packages/base-contract/src/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { DataItem } from '@0xproject/types';
+import * as _ from 'lodash';
+
+// tslint:disable-next-line:completed-docs
+export function formatABIDataItem(abi: DataItem, value: any, formatter: (type: string, value: any) => any): any {
```
fabioberger: Why `value: any`? We assume it's iterable in the code body. It should be at least `any[]`?

packages/deployer/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.3.4 - _TBD, 2018_
+
+    * Don't try to write contract artifact if the error occured (#485)
```
fabioberger: `if the error` -> `if an error`

packages/deployer/src/compiler.ts
```diff
@@ -177,10 +177,15 @@ export class Compiler {
         );
 
         if (!_.isUndefined(compiled.errors)) {
+            const isError = (errorOrWarning: string) => !errorOrWarning.includes('Warning');
```
fabioberger: Let's not have `Warning` be a string literal. Is these part of an error message we throw? If so, a constant should be used in both places linking them conceptually.

packages/metacoin/contracts/Metacoin.sol
```diff
@@ -0,0 +1,25 @@
...
+    struct TransferData {
+        address to;
+        uint256 amount;
+    }
+
+    function Metacoin() public {
+        balances[msg.sender] = 10000;
+    }
+
+    function transfer(TransferData transferData) public returns (bool success) {
```
fabioberger: 🤙

packages/metacoin/package.json
```diff
@@ -0,0 +1,60 @@
+{
```
fabioberger: Should `Metacoin` be in the monorepo?! Could see both sides of the argument. What I'm worried about is people forking the monorepo and building their project inside it 🔫  But maybe that's a stupid worry...

packages/metacoin/package.json
```diff
@@ -0,0 +1,60 @@
...
+        "coverage:report:text": "istanbul report text",
+        "coverage:report:html": "istanbul report html && open coverage/index.html",
+        "coverage:report:lcov": "istanbul report lcov",
+        "test:circleci": "yarn test:coverage",
+        "compile": "node ../deployer/lib/src/cli.js compile --contracts Metacoin --contracts-dir contracts --artifacts-dir artifacts"
+    },
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/0xProject/0x-monorepo.git"
+    },
```
fabioberger: Let's remove this.

packages/metacoin/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "author": "",
```
fabioberger: Put your name here 😉

packages/metacoin/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "license": "Apache-2.0",
+    "bugs": {
+        "url": "https://github.com/0xProject/0x-monorepo/issues"
+    },
+    "homepage": "https://github.com/0xProject/0x-monorepo/packages/metacoin/README.md",
```
fabioberger: Remove `homepage` and `bugs`. These are more for published packages.

packages/metacoin/package.json
```diff
@@ -0,0 +1,60 @@
...
+        "shx": "^0.2.2",
+        "tslint": "5.8.0",
+        "chai": "^4.0.1",
+        "chai-as-promised": "^7.1.0",
+        "chai-bignumber": "^2.0.1",
+        "typescript": "2.7.1"
+    },
+    "publishConfig": {
+        "access": "private"
+    }
```
fabioberger: remove. We will never publish. Public or private.

packages/metacoin/test/utils/web3_wrapper.ts
```diff
@@ -0,0 +1,30 @@
...
+    new GanacheSubprovider({
+        logger: {
+            log: (arg: any) => {
+                fs.appendFileSync('ganache.log', `${arg}\n`);
+            },
+        },
+        verbose: env.parseBoolean(EnvVars.SolidityCoverage),
+        port: 8545,
+        networkId: 50,
+        mnemonic: 'concert load couple harbor equip island argue ramp clarify fence smart topic',
```
fabioberger: It would be great if we have a `config.ts` file with all the configurable aspects of the Metacoin project. This is really easy for people to edit and to see all that can be easily configured in one place.

packages/base-contract/src/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (abi.type.match(trailingArrayRegex)) {
+        const arrayItemType = abi.type.replace(trailingArrayRegex, '');
+        return _.map(value, val => {
+            const arrayItemAbi = {
+                ...abi,
+                type: arrayItemType,
+            };
+            return formatABIDataItem(arrayItemAbi, val, formatter);
+        });
+    } else if (abi.type === 'tuple') {
```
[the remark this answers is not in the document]
LogvinovLeon: This is not the abi type, but the data type. We don't have a type for that. This will be the list of all solidity types which is infinite. `int`, `int[]`, `int[][]` etc.

packages/deployer/src/compiler.ts
```diff
@@ -177,10 +177,15 @@ export class Compiler {
         );
```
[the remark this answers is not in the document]
LogvinovLeon: https://stackoverflow.com/a/7229996/3546986

packages/deployer/src/compiler.ts
```diff
@@ -177,10 +177,15 @@ export class Compiler {
...
 
         if (!_.isUndefined(compiled.errors)) {
+            const isError = (errorOrWarning: string) => !errorOrWarning.includes('Warning');
+            const errors = _.filter(compiled.errors, isError);
             _.forEach(compiled.errors, errMsg => {
                 const normalizedErrMsg = getNormalizedErrMsg(errMsg);
                 this._solcErrors.add(normalizedErrMsg);
             });
+            if (!_.isEmpty(errors)) {
+                return;
```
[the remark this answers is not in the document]
LogvinovLeon: Yes. They're printed after the compilation of all contracts is done. That is strange. Might change that in future PR's

packages/base-contract/src/index.ts
LogvinovLeon: Because the real semantic of that function is: "Get a value and if it's a BigNumber - return a string. Otherwise - noop." With TS 2.7 I can't express such a logic in a type system. With TS 2.8 cond types that's possible. I'll change that as soon as we migrate.

packages/base-contract/src/utils.ts
LogvinovLeon: We only assume it's iterable in one of the branches. It might be a `BigNumber` and `BigNumber` is not a subtype of `any[]`

packages/deployer/src/compiler.ts
LogvinovLeon: It's a part of an error message solidity compiler throws. The only way to separate Warnings from errors is like that

packages/metacoin/package.json
LogvinovLeon: I'm ... not sure. I think it might make sense to move it out later. It's much easier to develop like that. Let's keep it for now and move out before we push it hard to the public.

packages/metacoin/package.json
LogvinovLeon: screen shot 2018-04-02 at 11 01 42

packages/deployer/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.3.5",
+        "changes": [
+            {
+                "note": "Don't try to write contract artifact if the error occured",
```
fabioberger: `if the error` -> `if an error`

PR #487, opened 2018-03-30
Implement wallet locked and uninstalled states

BMillman19:
## Description

screen shot 2018-03-29 at 4 57 55 pm
screen shot 2018-03-29 at 4 58 01 pm
screen shot 2018-03-29 at 4 58 07 pm

Wallet states for unlocked and uninstalled metamask. Also provides a way to set up your ledger.

## Motivation and Context

N/A

## How Has This Been Tested?

N/A

## Types of changes

## Checklist:


packages/website/ts/components/wallet/wallet_disconnected_item.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const styles: Styles = {
+    button: {
+        border: colors.walletBorder,
+        borderStyle: 'solid',
+        borderWidth: 1,
+        height: 80,
+    },
+    hrefAdjustment: {
+        paddingTop: 20, // For some reason when we set the href prop of a FlatButton material-ui reduces the top padding
```
fabioberger: 🤔

fabioberger: Let's probably convert this to a `HACK` comment. Is there an outstanding Material-ui issue for this that we could link to?

packages/website/ts/components/wallet/wallet_disconnected_item.tsx
```diff
@@ -0,0 +1,79 @@
...
+            <div className="mx-auto">
+                <div className="table" style={{ height: ITEM_HEIGHT }}>
+                    <div className="table-cell align-middle">
+                        <ProviderButton isExternallyInjectedProvider={isExternallyInjectedProvider} />
+                        <div className="flex flex-center py2" style={{ paddingBottom: BUTTON_BOTTOM_PADDING }}>
+                            <div className="mx-auto">
+                                <img src="/images/ledger_icon.png" style={{ width: LEDGER_ICON_WIDTH }} />
+                                <a
+                                    className="px1"
+                                    href="javascript:;"
```
fabioberger: Hm... do you want to use an `a tag` here for it's styling? Why not a div?

packages/website/ts/components/wallet/wallet_disconnected_item.tsx
BMillman19: Couldn't find anything related in the issues or PRs, we are a good number of versions behind, we might want to consider upgrading material ui but not sure the refactor would be worth it. Plus the devs seems to be working towards a v1 release

packages/website/ts/components/wallet/wallet_disconnected_item.tsx
BMillman19: I retooled this a bit, now the entire div should be clickable to activate the ledger prompt

PR #488, opened 2018-03-30
Monitor different state layers with OrderWatcher

avonian:
Allow instantiation of stand-alone OrderWatchers that can monitor different blockchain state layers (e.g. pending or latest)

## Description

Modified `OrderStateWatcherConfig` to support a `stateLayer` (BlockParamLiteral) setting so that OrderStateWatcher instances can monitor different blockchain state layers.

Also modified ZeroEx class so that it does not instantiate a default `orderStateWatcher`, instead, users can now create as many stand-alone watchers as they need by using the new `ZeroEx.createOrderStateWatcher(OrderStateWatcherConfig)` method.

In the future this will allow to create watchers that go beyond simply `pending` and `latest` state layers, to arbitrary confirmation depths.

## Motivation and Context

OrderStateWatcher subscriptions for mined blocks (rather than simply mempool transactions) can provide reliable order state updates that are unlikely to get reversed, and it seems they could be a reliable source of truth for relayers looking to update orderbooks.

## How Has This Been Tested?

Ran various tests on testRPC as well as Kovan without issues, in fact, where previously I would observe many inconsistencies operating OrderStateWatcher at the `pending` layer, running watchers at the `latest` layer seems to consistently provide accurate snapshots of order state as I expect to see it.

I've updated the `order_state_watcher_test` which also passes.

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/CHANGELOG.md
+++ b/packages/0x.js/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## v0.34.0 - _TBD_
 
     * Fix the bug causing `zeroEx.exchange.fillOrdersUpToAsync` validation to fail if there were some extra orders passed (#470)
+    * Remove automatic instantiation of `zeroEx.orderStateWatcher`
+    * Add `zeroEx.createOrderStateWatcher` to allow creating arbitrary number of OrderStateWatchers
+    * Added `stateLayer` setting to `OrderStateWatcherConfig` so OrderStateWatcher can be set to monitor different blockchain state layers
 
 ## v0.33.2 - _March 18, 2018_
 
--- a/packages/0x.js/CHANGELOG.md
+++ b/packages/0x.js/CHANGELOG.md
@@ -3,9 +3,9 @@
 ## v0.34.0 - _TBD_
 
     * Fix the bug causing `zeroEx.exchange.fillOrdersUpToAsync` validation to fail if there were some extra orders passed (#470)
-    * Remove automatic instantiation of `zeroEx.orderStateWatcher`
-    * Add `zeroEx.createOrderStateWatcher` to allow creating arbitrary number of OrderStateWatchers
-    * Added `stateLayer` setting to `OrderStateWatcherConfig` so OrderStateWatcher can be set to monitor different blockchain state layers
+    * Remove automatic instantiation of `zeroEx.orderStateWatcher` (#488)
+    * Add `zeroEx.createOrderStateWatcher` to allow creating arbitrary number of OrderStateWatchers (#488)
+    * Added `stateLayer` setting to `OrderStateWatcherConfig` so OrderStateWatcher can be set to monitor different blockchain state layers (#488)
 
 ## v0.33.2 - _March 18, 2018_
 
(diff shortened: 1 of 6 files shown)


packages/0x.js/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 ## v0.34.0 - _TBD_
 
     * Fix the bug causing `zeroEx.exchange.fillOrdersUpToAsync` validation to fail if there were some extra orders passed (#470)
+    * Remove automatic instantiation of `zeroEx.orderStateWatcher`
+    * Add `zeroEx.createOrderStateWatcher` to allow creating arbitrary number of OrderStateWatchers
+    * Added `stateLayer` setting to `OrderStateWatcherConfig` so OrderStateWatcher can be set to monitor different blockchain state layers
```
fabioberger: @avonian can you please add the PR numbers to the right of these entries? Copy format ` (#470)` like above.

packages/0x.js/src/types.ts
```diff
@@ -175,6 +175,7 @@ export interface OrderStateWatcherConfig {
     eventPollingIntervalMs?: number;
     expirationMarginMs?: number;
     cleanupJobIntervalMs?: number;
+    stateLayer: BlockParamLiteral;
```
fabioberger: Add description for `stateLayer` in the above comment. Let's mention that it defaults to `pending`.

fabioberger: Oh, it should also be `optional`.

packages/0x.js/src/0x.ts
```diff
@@ -336,6 +324,20 @@ export class ZeroEx {
         const txReceipt = await txReceiptPromise;
         return txReceipt;
     }
+    /**
+     * Instantiates and returns a new OrderStateWatcher instance.
+     * @param   config      The configuration object. Look up the type for the description.
+     * @return  An instance of the 0x.js OrderStateWatcher class.
```
fabioberger: Let's add `Defaults to watching the pending state`

packages/0x.js/src/order_watcher/event_watcher.ts
```diff
@@ -22,8 +22,10 @@ export class EventWatcher {
     private _pollingIntervalMs: number;
     private _intervalIdIfExists?: NodeJS.Timer;
     private _lastEvents: LogEntry[] = [];
-    constructor(web3Wrapper: Web3Wrapper, pollingIntervalIfExistsMs: undefined | number) {
+    private _stateLayer: BlockParamLiteral;
+    constructor(web3Wrapper: Web3Wrapper, pollingIntervalIfExistsMs: undefined | number, stateLayer: BlockParamLiteral = BlockParamLiteral.Pending) {
```
LogvinovLeon: Maybe we should default to Latest for now? It's more stable

packages/0x.js/src/order_watcher/order_state_watcher.ts
```diff
@@ -76,6 +76,7 @@ export class OrderStateWatcher {
     private _balanceAndProxyAllowanceLazyStore: BalanceAndProxyAllowanceLazyStore;
     private _cleanupJobInterval: number;
     private _cleanupJobIntervalIdIfExists?: NodeJS.Timer;
+    private _stateLayer: BlockParamLiteral;
```
LogvinovLeon: We never use that variable outside of the constructor. Let's make it local.

PR #489, opened 2018-03-30
Automate NPM Publish Process

fabioberger:
## Description

This PR automates a lot of the manual steps involved during publishing. You now run:

```
yarn lerna:publish
```

And it:
- Updates CHANGELOG files to accurately reflect the correct date and version to be published.
- Commits and pushes the CHANGELOG updates to `development` on Github.
- Calls `lerna publish` under the hood, passing it the appropriate new package versions from the CHANGELOG.
- Publishes the CHANGELOG notes in the Github releases created for each package.

In addition, there is also a script to migrate our existing CHANGELOG.md files to CHANGELOG.json files (MD is not good enough for structured data -- found lots of inconsistencies). We still have MD versions of the CHANGELOG but they are auto-generated from the JSON.

Other small changes:
- Improved README
- Fixed some dep versions to avoid multiple  #versions.

## Motivation and Context

Publishing packages was becoming too tedious impacting the frequency with which we were publishing updates.

## How Has This Been Tested?

Manually. There is a `IS_DRY_RUN` flag in the `publish` script.

## Types of changes

## Checklist:


packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+(async () => {
+    const allLernaPackages = lernaGetPackages(MONOREPO_ROOT_PATH);
+    const publicLernaPackages = _.filter(allLernaPackages, pkg => !pkg.package.private);
+    _.each(publicLernaPackages, lernaPackage => {
+        const changelogMdIfExists = getChangelogMdIfExists(lernaPackage.package.name, lernaPackage.location);
+        if (_.isUndefined(changelogMdIfExists)) {
+            throw new Error(`${lernaPackage.package.name} should have CHANGELOG.md b/c it's public. Add one.`);
+        }
+
+        const lines = (changelogMdIfExists as any).split('\n');
```
LogvinovLeon: `as string`. And it's strange, that TS doesn't detect it already being checked for undefined

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+        const changelogs: Changelog[] = [];
+        let changelog: Changelog = {
+            version: '',
+            changes: [],
+        };
+        for (const line of lines) {
+            if (_.startsWith(line, '## ')) {
```
LogvinovLeon: Assign `##` to a named constant

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+                let version = line.substr(4).split(' - ')[0];
+                if (version === '0.x.x') {
```
LogvinovLeon: I don't get that at all. What if it's 1.x.x? What if it's 0.x.y?

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+                    version,
+                    changes: [],
+                };
+                if (!_.isUndefined(date)) {
+                    changelog.timestamp = date.unix();
+                }
+                if (!_.includes(dateStr, 'TBD')) {
+                    changelog.isPublished = true;
+                }
+                (changelogs as any).push(changelog);
```
LogvinovLeon: Why a conversion here? It's already of a correct type.

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+            } else if (_.includes(line, '* ')) {
+                const note = line.split('* ')[1].split(' (#')[0];
```
LogvinovLeon: Could we use regexes with capture groups instead of that magic with splits and magic numbers?
Or at least provide an example format of the data being parsed above in a comment. 
` * Some note (#123)`

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+                    pr = prChunk.split(')')[0];
+                }
+                const changes = {
+                    note,
+                    pr,
+                };
+                changelog.changes.push(changes);
+            }
+        }
+        const changelogJson = JSON.stringify(changelogs, null, '\t');
```
LogvinovLeon: Why do we use tab as a separator?

packages/monorepo-scripts/src/convert_changelogs.ts
```diff
@@ -0,0 +1,83 @@
...
+});
+
+function getChangelogMdIfExists(packageName: string, location: string): string | undefined {
+    const changelogPath = path.join(location, 'CHANGELOG.md');
+    let changelogMd: string;
+    try {
+        changelogMd = fs.readFileSync(changelogPath, 'utf-8');
+        return changelogMd;
+    } catch (err) {
+        // If none exists, create new, empty one.
```
LogvinovLeon: We don't create a new one so please remove that comment

packages/monorepo-scripts/src/globals.d.ts
```diff
@@ -1,6 +1,7 @@
 declare module 'async-child-process';
 declare module 'publish-release';
 declare module 'es6-promisify';
+declare module 'semver-diff';
```
LogvinovLeon: 1. Define types.
2. Move those types to `typescript-typings`

packages/monorepo-scripts/src/postpublish_utils.ts
```diff
@@ -88,23 +91,52 @@ export const postpublishUtils = {
...
+        return 'N/A';
+    },
+    updateChangelogIsPublished(packageName: string) {
+        const changelogJSONPath = path.join(constants.monorepoRootPath, 'packages', packageName, 'CHANGELOG.json');
+        const changelogJSON = fs.readFileSync(changelogJSONPath, 'utf-8');
+        const changelogs = JSON.parse(changelogJSON);
+        const latestLog = changelogs[0];
+        latestLog.isPublished = true;
+        changelogs[0] = latestLog;
+        fs.writeFileSync(changelogJSONPath, JSON.stringify(changelogs, null, '\t'));
```
LogvinovLeon: Why tab?

packages/monorepo-scripts/src/publish.ts
```diff
@@ -0,0 +1,209 @@
...
+    minor: SemVerIndex.Minor,
+    major: SemVerIndex.Major,
+};
+
+(async () => {
+    const updatedPublicPackages = await getPublicLernaUpdatedPackagesAsync();
+    const updatedPackageNames = _.map(updatedPublicPackages, pkg => pkg.name);
+
+    const allLernaPackages = lernaGetPackages(constants.monorepoRootPath);
+    const relevantLernaPackages = _.filter(allLernaPackages, pkg => {
```
LogvinovLeon: If relevant package is a package that is updated why don't we call it updatedLernaPackages and updatedPackageNames. I don't fully understand that step.

LogvinovLeon:
And it doesn't build. I've already fixed that in my PR, but feel free to duplicate. You'll need to add a .gitkeep file in solc_bin directory to keep it empty but existing.

package.json
```diff
@@ -12,13 +12,12 @@
...
         "lerna:rebuild": "lerna run clean; lerna run build;",
         "lerna:publish":
-            "yarn install; lerna run clean; lerna run build; lerna publish --registry=https://registry.npmjs.org/"
+            "yarn install; yarn lerna:rebuild; node ./packages/monorepo-scripts/lib/publish.js"
     },
     "config": {
         "mnemonic": "concert load couple harbor equip island argue ramp clarify fence smart topic"
     },
     "devDependencies": {
-        "@0xproject/utils": "^0.4.2",
```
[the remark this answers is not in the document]
BMillman19: we used to need it for the post-publish scripts we had at the top level

package.json
```diff
@@ -12,13 +12,12 @@
         "lerna:run": "lerna run",
         "lerna:rebuild": "lerna run clean; lerna run build;",
         "lerna:publish":
-            "yarn install; lerna run clean; lerna run build; lerna publish --registry=https://registry.npmjs.org/"
+            "yarn install; yarn lerna:rebuild; node ./packages/monorepo-scripts/lib/publish.js"
```
BMillman19: is there we can call `yarn publish` in the context of the `monorepo-scripts` package? if not, the two implementations may diverge

packages/monorepo-scripts/package.json
```diff
@@ -7,6 +7,8 @@
     "scripts": {
         "build:watch": "tsc -w",
         "deps_versions": "node ./lib/deps_versions.js",
+        "publish": "yarn build; node ./lib/publish.js",
+        "convert_changelogs": "yarn build; node ./lib/convert_changelogs.js",
```
BMillman19: where are we running this?

packages/monorepo-scripts/src/convert_changelogs.ts
fabioberger: I think we only use this in one place. Keep in mind, this script is temporary. We will use it once to migrate over to JSON and never again. From then on, MD changelogs are generated from the JSON counterparts.

package.json
fabioberger: Can you elaborate?

packages/monorepo-scripts/package.json
fabioberger: This script is temporary. We will use it once to migrate over to JSON and never again. From then on, MD changelogs are generated from the JSON counterparts.

packages/monorepo-scripts/src/globals.d.ts
fabioberger: I can only do this once there is an older version of `typescript-typings` published... `monorepo-scripts` will otherwise cause a dependency cycle. This is a bit hacky but don't know another way around it.

packages/monorepo-scripts/src/postpublish_utils.ts
fabioberger: Why not?

dekz: I think prettier will reject the files if they have tabs in them, and it's inconsistent with our other json files (like package.json) which has 4 space indentation.

packages/monorepo-scripts/src/publish.ts
```diff
@@ -0,0 +1,209 @@
...
+import * as path from 'path';
+import { exec as execAsync, spawn } from 'promisify-child-process';
+import semverDiff = require('semver-diff');
+import semverSort = require('semver-sort');
+
+import { constants } from './constants';
+import { Changelog, Changes, SemVerIndex, UpdatedPackage } from './types';
+import { utils } from './utils';
+
+const IS_DRY_RUN = process.env.IS_DRY_RUN === 'false';
```
dekz: I find this a bit confusing, when dry run then it should be a "test" run with no dangerous side effects.

```
const IS_DRY_RUN = process.env.IS_DRY_RUN === 'true'
```

and in the package.json

```
"script:publish:dry": "IS_DRY_RUN=true yarn script:publish"
```

Setting DRY_RUN to false feels like it shouldn't perform a dry run, and therefor perform the dangerous side effects.

packages/website/package.json
```diff
@@ -38,7 +38,7 @@
         "less": "^2.7.2",
         "lodash": "^4.17.4",
         "material-ui": "^0.17.1",
-        "moment": "^2.18.1",
+        "moment": "2.13.0",
```
dekz: What's the reason for the downgrade? I believe latest moment.js now includes types in the package.

packages/monorepo-scripts/src/postpublish_utils.ts
fabioberger: Yep, that's correct. Calling Prettier on all generated files now.

packages/monorepo-scripts/src/publish.ts
fabioberger: Yep, agreed. Flipped it.

packages/website/package.json
fabioberger: Yep, we can also consolidate versions to a higher one. Going with `2.21.0` since thats what another dep uses, so we only include a single version :)

packages/monorepo-scripts/src/globals.d.ts
LogvinovLeon: Gotcha. True

package.json
BMillman19: there is a `script:publish` command in the `monorepo-scripts` package, we have to remember to modify both this one and that one

PR #491, opened 2018-04-01
Fix build errors

abandeali1:
## Description

* Adds missing import for `Web3Provider` type
* Creates the `solc_bin` directory in the deployer package if it does not already exist

## Types of changes

## Checklist:

### DIFF
--- a/packages/deployer/CHANGELOG.md
+++ b/packages/deployer/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.3.2 - _TBD_
+
+    * Create solc_bin directory if does not exist before attempting to compile (#491)
+
 ## v0.3.1 - _March 18, 2018_
 
     * Add TS types for `yargs`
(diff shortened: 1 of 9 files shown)


packages/deployer/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v0.3.2 - _April, 2018_
```
fabioberger: Please put `TBD` as date here.

packages/deployer/src/compiler.ts
```diff
@@ -32,6 +32,7 @@ import {
 import { utils } from './utils/utils';
 
 const ALL_CONTRACTS_IDENTIFIER = '*';
+const SOLC_BIN_DIR = path.join(__dirname, '../../solc_bin');
```
fabioberger: The point of using `path.join` is so that the delimiter used is OS specific. Please split `../../solc_bin` into, `'..', '..', 'solc_bin'` so that `path.join` results in the proper delimiter (i.e `\` on windows) for the entire path. Will also open an Airtable issue to do this everywhere, we've been making this mistake in many places.

PR #492, opened 2018-04-02
[WIP] Add market quote

dekz:
## Description

* adds `marketFillOrdersQuote` for an on-chain/off-chain query, given a set of orders, how much will be filled and how much fees are required.

## Motivation and Context

It is difficult on-chain to compute the amount of fees required to fill the orders. 

The complexity is compounded when buying ZRX (fee abstraction) and there are fees required to be paid. Without an on-chain query we would have to iterate or assume no fees. 

With an on-chain quote we can perform the following in wrapper contracts:
```
1. Quote the desired tokens, computing how many fees are required to sufficiently buy X desired tokens
2. Quote for fee abstraction of a set of ZRX orders
3. market fill ZRX orders for desiredTokenFees+tokenFeeAbstractionFees
4. marketFill desiredTokens, having sufficient fees
```

## How Has This Been Tested?

Off-chain tests via exchange wrapper
On-chain tests via trade widget forwarding contract 

## Types of changes

## Checklist:


dekz:
We had previously discussed a forwarding contract having a supply of ZRX tokens and topping those tokens up after market filling the `desiredTokens`. That solution couldn't handle fees on the ZRX orders and could result in the pool of tokens being drained.

This solution also has a small problem. The original quotes in step 1) is for 100% of the available takerToken. For example, assume a DAI/WETH token pair. 
```
Step 1 calculates a quote for exchanging 1 WETH for 100 DAI, with a total fee of 0.1 ZRX. 
Step 2 calculates a quote for exchange 0.1 ZRX on a ZRX/WETH token pair, with a total fee of 0.1 ZRX
Step 3 market fills 0.2ZRX on a ZRX/WETH token pair, leaving 0.9 WETH
Step 4 market fills DAI/WETH but only has 0.9 WETH remaining in takerTokens
```

So the result is total ZRX fee abstraction, but may result in a couple of fractions of ZRX remaining in the forwarding contract (as it was originally quoted to fill 1WETH, not 0.9 WETH).

These fractions are small but are worth calling out the limitations

PR #493, opened 2018-04-02
Deployer: Support for External Codebases + Overloaded Function Names 

hysz:
## Description

1. The ability to import code from multiple source directories
2. Namespacing source directories to avoid filename clashes
3. Support for compiling overloaded functions

## Motivation

These changes were made as part of the ERC721 project, where we depend on the Zeppelin library.

(i)
In order to import external libraries, like Zeppelin, we need support for importing from multiple locations. Filenames could clash between codebases, so it's important that directories are namespaced. For example, 0x has a version SafeMath.sol that is different from Zeppelin's. 

(ii)
Solidity allows for function overloading, but Typescript does not. In particular, ERC721 overloads the transferFrom function. We needed a way to rename overloaded functions in our artifacts, so that compiling to Typescript doesn't break.

## How Has This Been Tested?

All existing tests in the deployer and contracts packages pass. New tests were written for function overloading.

## Types of changes

## Checklist:


dekz:
`--contract-dirs` yesssss!

packages/deployer/src/cli.ts
```diff
@@ -67,6 +67,32 @@ async function onDeployCommandAsync(argv: CliOptions): Promise<void> {
...
     await commands.deployAsync(argv.contract as string, deployerArgs, deployerOpts);
 }
+/**
+ * Creates a set of contracts to compile.
+ * @param contractDirectoriesList Comma separated list of contract directories
+ * @return Set of contract directories
+ */
+function getContractDirectoriesFromList(contractDirectoriesList: string): Set<ContractDirectory> {
+    const directories = new Set();
+    const namespacedDirectories: string[] = contractDirectoriesList.split(",");
```
dekz: You'll likely get build failures from tslint/prettier as it prefers single quotes for strings over double quotes, and some spacing around directives `if(` etc.

Our configuration file for prettier is:
```
{
    "tabWidth": 4,
    "printWidth": 120,
    "trailingComma": all,
    "singleQuote": true
}
```

Or you can `yarn run prettier` in the top level directory and it should fix up most for you!

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+            {
+                throw new Error("Failed to rename overloaded function '" + unoverloadedName +  "' to '" + overloadedName + "' already exists.");
+            }
+
+            // Replace function name in ABI if overloaded
+            if(overloadedName != originalName) {
+                (abi[i] as any).name = overloadedName;
+                console.log("Renamed overloaded function '" + originalName + "' to '" + overloadedName + "'");
+            } else {
+                functionList[originalName] = 1;
```
dekz: I'm not exactly clear on what this does, this is just to keep a count of the overloaded functions?

Maybe its clearer (but less efficient) to increment in the above conditional and on line 251 functionList[originalName] + 1.

You can also save a few `"` and `+` by using back ticks.
```typescript
originalName+"_"+(++functionList[originalName])
```
becomes
```typescript
`${originalName}_${functionList[originalName]+1}`
```

packages/deployer/src/utils/compiler.ts
```diff
@@ -7,6 +7,51 @@ import { constants } from './constants';
...
+* @param  contractBaseDir Base contracts directory of search tree.
+* @return sourceFileId A system-wide unique identifier for the source file.
+*/
+export function constructDependencyFileId(dependencyFilePath: string, sourceFilePath: string): string {
+    if(dependencyFilePath.substr(0,1) == '/') {
+        // Path of the form /namespace/path/to/dependency.sol
+        return dependencyFilePath;
+    } else {
+        // Dependency is relative to the source file: ./dependency.sol, ../../some/path/dependency.sol, etc.
+        // Join the two paths to construct a valid sourec file id: /namespace/path/to/dependency.sol
```
dekz: typo `source`

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
         const sourceMapRuntime = compiled.contracts[contractIdentifier].srcmapRuntime;
         const sources = _.keys(compiled.sources);
         const updated_at = Date.now();
+
+        // There is no function overloading in typescript, so we must change the Typescript ABI interface.
+        // Overloaded function names are incremented as follows: functionName, functionName_2, functioname_3, ...
+        // If functionName_N already exists then compilation will fail.
+        const functionList:FunctionList = {};
+        for(let i = 0; i < abi.length; ++i) {
+            const type:string = abi[i].type;
```
fabioberger: Now I know you're not using `prettier` plugin 😉 Missing a space after variable name, before type. Please install and use `onSave` or run before committing.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
         const sourceMapRuntime = compiled.contracts[contractIdentifier].srcmapRuntime;
         const sources = _.keys(compiled.sources);
         const updated_at = Date.now();
+
+        // There is no function overloading in typescript, so we must change the Typescript ABI interface.
+        // Overloaded function names are incremented as follows: functionName, functionName_2, functioname_3, ...
+        // If functionName_N already exists then compilation will fail.
+        const functionList:FunctionList = {};
+        for(let i = 0; i < abi.length; ++i) {
```
fabioberger: use `_.each`

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +115,43 @@ export class Compiler {
...
-        this._contractSources = await Compiler._getContractSourcesAsync(this._contractsDir);
-        _.forIn(this._contractSources, this._setContractSpecificSourceData.bind(this));
-        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
-            ? _.keys(this._contractSources)
+        this._contractSources = {};
+        for(let contractDir of Array.from(this._contractDirs.values())) {
+            let sources = await Compiler._getContractSourcesAsync(contractDir.path, contractDir.path);
+            _.forIn(sources, (source, sourceFilePath) => {
+                // Construct a unique ID for this source file
+                const sourceFileId:string = constructSourceFileId(contractDir.namespace, sourceFilePath);
```
dekz: You can skip over most of these types as they are inferred by the return type of the function or the value assigned.
```typescript
const sourceFileId = constructSourceFileId(contractDir.namespace, sourceFilePath);
```

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
         const sources = _.keys(compiled.sources);
         const updated_at = Date.now();
+
+        // There is no function overloading in typescript, so we must change the Typescript ABI interface.
+        // Overloaded function names are incremented as follows: functionName, functionName_2, functioname_3, ...
+        // If functionName_N already exists then compilation will fail.
+        const functionList:FunctionList = {};
+        for(let i = 0; i < abi.length; ++i) {
+            const type:string = abi[i].type;
+            if(type !== "function") continue;
```
fabioberger: It's time we move the `AbiType` type currently in `@0xproject/abi-gen` to the `@0xproject/types` package. Use it here instead of using a string literal.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+
+            // Construct possible function names
+            const originalName:string = (abi[i] as any).name;
+            const overloadedName:string = (originalName in functionList) ? originalName+"_"+(++functionList[originalName]) : originalName;
```
fabioberger: Use template strings: \`${originalName}_${++functionList[originalName]}\`

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+            const unoverloadedName:string = originalName.indexOf("_") > 0 ? originalName.substr(0, originalName.lastIndexOf("_")) : originalName;
```
fabioberger: Use `_.includes` over `indexOf`

fabioberger: Is there no chance a user uses `_` in their method names? If so, we might need a more robust check here, perhaps a regex checking for number followed by underscore?

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+
+            // Fail if an overloaded function was renamed to a function that already exists
+            if((overloadedName != originalName && overloadedName in functionList) ||
```
fabioberger: Yikes. We should aways be doing strict equality checks: `!==`. Feel like our linter checks for this. Is you don't want to install the editor plugin, you can run it with: `yarn lint` in a package, or `yarn lerna:run lint` if you want to lint **all** packages.

packages/deployer/src/compiler.ts
fabioberger: Rather then use `in`, let's use `_.has` for consistency. Unfortunately there are 50 ways to do the same thing in JS, to simplify our lives, we always try and do it the same way :)

```
_.has(functionList, originalName)
```

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
         const sourceMapRuntime = compiled.contracts[contractIdentifier].srcmapRuntime;
         const sources = _.keys(compiled.sources);
         const updated_at = Date.now();
+
+        // There is no function overloading in typescript, so we must change the Typescript ABI interface.
+        // Overloaded function names are incremented as follows: functionName, functionName_2, functioname_3, ...
+        // If functionName_N already exists then compilation will fail.
+        const functionList:FunctionList = {};
```
fabioberger: Rename `FunctionList` to `FunctionNameToSeenCount`.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+            if(type !== "function") continue;
+
+            // Construct possible function names
+            const originalName:string = (abi[i] as any).name;
+            const overloadedName:string = (originalName in functionList) ? originalName+"_"+(++functionList[originalName]) : originalName;
+            const unoverloadedName:string = originalName.indexOf("_") > 0 ? originalName.substr(0, originalName.lastIndexOf("_")) : originalName;
+
+            // Fail if an overloaded function was renamed to a function that already exists
+            if((overloadedName != originalName && overloadedName in functionList) ||
+                (unoverloadedName != originalName && unoverloadedName in functionList))
```
fabioberger: This is a lot of conditionals to process. Let's assign them to an understandable variable name to reduce the cognitive burden on the reader:

```
const didOverloadedFunctionNameCreateCollision = (overloadedName != originalName && overloadedName in functionList) || (unoverloadedName != originalName && unoverloadedName in functionList);
if (didOverloadedFunctionNameCreateCollision) {
  ...
```

This way a casual reader knows **what** the conditional checks, and can dig into the nuts and bolts if he wants, but ins't forced to just to understand what the code is doing. 

It also removes the need for the code comment.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+            {
+                throw new Error("Failed to rename overloaded function '" + unoverloadedName +  "' to '" + overloadedName + "' already exists.");
```
fabioberger: Template string.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+            }
+
+            // Replace function name in ABI if overloaded
+            if(overloadedName != originalName) {
```
fabioberger: Hm... not sure this logic is clear to me. When would `overloadedName` to equal to `originalName`? Shouldn't `originalName` always be the **original** name? This might be a naming issue.

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
...
+                (abi[i] as any).name = overloadedName;
+                console.log("Renamed overloaded function '" + originalName + "' to '" + overloadedName + "'");
+            } else {
+                functionList[originalName] = 1;
+            }
+        }
+
```
fabioberger: This function is becoming very long and windy. Can we perform some functional decomposition here?

packages/deployer/src/utils/compiler.ts
```diff
@@ -7,6 +7,51 @@ import { constants } from './constants';
...
+/**
+* Constructs a system-wide unique identifier for a source file.
+* @param directoryNamespace Namespace of the source file's root contract directory.
+* @param sourceFilePath Path to a source file, relative to contractBaseDir.
+* @return sourceFileId A system-wide unique identifier for the source file.
+*/
+export function constructSourceFileId(directoryNamespace: string, sourceFilePath: string): string {
+    let namespacePrefix:string = "";
+    if(directoryNamespace != "") {
+        namespacePrefix = "/" + directoryNamespace;
```
fabioberger: Template strings. Please check all new code for this. Strict equality, etc... too. Will stop commenting on them.

packages/deployer/src/utils/compiler.ts
```diff
@@ -7,6 +7,51 @@ import { constants } from './constants';
...
+    }
+    return namespacePrefix + "/" + sourceFilePath.replace(/^\/+/g, '');
```
fabioberger: We assign to a variable before returning. This adds clarity of intent.

fabioberger: Assign regex to named constant for clarity.

packages/deployer/src/utils/compiler.ts
```diff
@@ -7,6 +7,51 @@ import { constants } from './constants';
...
+}
+
+/**
+* Constructs a system-wide unique identifier for a dependency file.
+* @param dependencyFilePath Path from a sourceFile to a dependency.
+* @param  contractBaseDir Base contracts directory of search tree.
+* @return sourceFileId A system-wide unique identifier for the source file.
+*/
+export function constructDependencyFileId(dependencyFilePath: string, sourceFilePath: string): string {
+    if(dependencyFilePath.substr(0,1) == '/') {
```
fabioberger: `_.startsWith` is what you're looking for.

packages/deployer/src/utils/compiler.ts
```diff
@@ -7,6 +7,51 @@ import { constants } from './constants';
...
+
+/**
+* Constructs a system-wide unique identifier for a contract.
+* @param directoryNamespace Namespace of the source file's root contract directory.
+* @param sourceFilePath Path to a source file, relative to contractBaseDir.
+* @return sourceFileId A system-wide unique identifier for contract.
+*/
+export function constructContractId(directoryNamespace: string, sourceFilePath: string): string {
+    let namespacePrefix:string = "";
+    if(directoryNamespace != "") {
```
fabioberger: `!_.isEmpty`

packages/deployer/src/utils/types.ts
```diff
@@ -18,6 +18,11 @@ export interface ContractNetworks {
     [key: number]: ContractNetworkData;
 }
 
+export interface ContractDirectory {
+    path: string,
+    namespace: string,
```
fabioberger: Should be semi-colons. Prettier to the rescue!

packages/deployer/test/compiler_utils_test.ts
```diff
@@ -47,28 +47,28 @@ describe('Compiler utils', () => {
...
                 },
             );
-            expect(parseDependencies(exchangeSource)).to.be.deep.equal(['Token.sol', 'Ownable.sol']);
+            expect(parseDependencies(exchangeSource, "/main/TokenTransferProxy.sol")).to.be.deep.equal(['/base/Token.sol', '/base/Ownable.sol']);
         });
         // TODO: For now that doesn't work. This will work after we switch to a grammar-based parser
         it.skip('correctly parses commented out dependencies', async () => {
             const contractWithCommentedOutDependencies = `// import "./TokenTransferProxy.sol";`;
-            expect(parseDependencies(contractWithCommentedOutDependencies)).to.be.deep.equal([]);
+            expect(parseDependencies(contractWithCommentedOutDependencies, "")).to.be.deep.equal([]);
```
fabioberger: Without looking up the definition of `parseDependencies`, the next reader has no idea what parameter was set to an empty string. Solution: assign clearly named variable to empty string, pass variable here.

packages/deployer/src/cli.ts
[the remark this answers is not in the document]
Recmo: It's how `solc` is doing it, which seems like a good benchmark:

```
  --allow-paths path(s)
                       Allow a given path for imports. A list of paths can be
                       supplied by separating them with a comma.
```

packages/deployer/src/compiler.ts
```diff
@@ -207,6 +237,36 @@ export class Compiler {
         const sourceMapRuntime = compiled.contracts[contractIdentifier].srcmapRuntime;
         const sources = _.keys(compiled.sources);
         const updated_at = Date.now();
+
+        // There is no function overloading in typescript, so we must change the Typescript ABI interface.
```
Recmo: There are some options to overloading in TS. The documentation gives a simple example:

http://www.typescriptlang.org/docs/handbook/functions.html#overloads

I don't know TS well enough to tell if this is to be preferred over the current solution. It seems like you'd have to manually do the type-pattern-matching, but that's doable.

@fabioberger @LogvinovLeon what's your take on this?

packages/deployer/src/cli.ts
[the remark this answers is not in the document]
hysz: It's also how the Deployer parses other list parameters. Namely, the `specifiedContracts`.

packages/deployer/src/compiler.ts
```diff
@@ -87,13 +96,14 @@ export class Compiler {
         }
         return sources;
     }
+
```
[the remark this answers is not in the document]
hysz: Space removed.
* I installed the `Prettier` plugin and it doesn't complain about this. Is there a config value we should set for this?

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +115,43 @@ export class Compiler {
     public async compileAsync(): Promise<void> {
         await createDirIfDoesNotExistAsync(this._artifactsDir);
         await createDirIfDoesNotExistAsync(SOLC_BIN_DIR);
-        this._contractSources = await Compiler._getContractSourcesAsync(this._contractsDir);
-        _.forIn(this._contractSources, this._setContractSpecificSourceData.bind(this));
-        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
-            ? _.keys(this._contractSources)
+        this._contractSources = {};
+        for(let contractDir of Array.from(this._contractDirs.values())) {
```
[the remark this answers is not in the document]
hysz: 1. Moved the `contractDirs` to outside the loop.
2. I'll keep this in mind for the other loops. Although in this case I'll leave it as `for...of`, as there is a call to `_getContractSourcesAsync` within the loop.

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +115,43 @@ export class Compiler {
...
+
+                // Record the file's source and data
+                if(!_.isUndefined(this._contractSources[sourceFileId])) {
+                    throw new Error("Found duplicate source files with ID '" + sourceFileId + "'");
+                }
+                this._contractSources[sourceFileId] = source;
+
+                // Create a mapping between the contract id and its source file id
+                const contractId = constructContractId(contractDir.namespace, sourceFilePath);
+                this._contractIds[contractId] = sourceFileId;
```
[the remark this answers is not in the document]
hysz: Good call. I've added a check.

packages/deployer/test/compiler_utils_test.ts
hysz: Updated earlier tests as well.

packages/deployer/src/compiler.ts
hysz: @dekz @fabioberger - I moved this code block to a separate function. Also used your notes to rewrite it in a (hopefully) more clear way.

packages/deployer/package.json
```diff
@@ -6,7 +6,7 @@
     "types": "lib/src/index.d.ts",
     "scripts": {
         "build:watch": "tsc -w",
-        "build": "yarn clean && copyfiles 'test/fixtures/contracts/**/*' ./lib && tsc && copyfiles -u 3 './lib/src/monorepo_scripts/**/*' ./scripts",
+        "build": "yarn clean && copyfiles 'test/fixtures/contracts/**/*' 'test/fixtures/overloaded_contracts/**/*' ./lib && tsc && copyfiles -u 3 './lib/src/monorepo_scripts/**/*' ./scripts",
```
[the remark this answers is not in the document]
hysz: I guess that's not necessary. I've moved them into a subdirectory of `contracts`

packages/deployer/src/compiler.ts
```diff
@@ -105,27 +115,48 @@ export class Compiler {
...
         await createDirIfDoesNotExistAsync(this._artifactsDir);
         await createDirIfDoesNotExistAsync(SOLC_BIN_DIR);
-        this._contractSources = await Compiler._getContractSourcesAsync(this._contractsDir);
+        this._contractSources = {};
+        const contractDirs = Array.from(this._contractDirs.values());
+        for (const contractDir of contractDirs) {
+            const sources = await Compiler._getContractSourcesAsync(contractDir.path, contractDir.path);
+            _.forIn(sources, (source, sourceFilePath) => {
+                const sourceFileId = constructUniqueSourceFileId(contractDir.namespace, sourceFilePath);
+
```
BMillman19: remove all the newlines in this block

packages/deployer/src/compiler.ts
```diff
@@ -105,27 +115,48 @@ export class Compiler {
...
+                    throw new Error(`Found duplicate source files with ID '${sourceFileId}'`);
+                }
+                this._contractSources[sourceFileId] = source;
+
+                // Create a mapping between the contract id and its source file id
+                const contractId = constructContractId(contractDir.namespace, sourceFilePath);
+                if (!_.isUndefined(this._contractIds[contractId])) {
+                    throw new Error(`Found duplicate contract with ID '${contractId}'`);
+                }
+                this._contractIds[contractId] = sourceFileId;
```
BMillman19: we may want to consider writing to the instance variables after this loop because it may terminate prematurely in an error case

packages/deployer/src/compiler.ts
```diff
@@ -105,27 +115,48 @@ export class Compiler {
...
-    private async _compileContractAsync(fileName: string): Promise<void> {
+    private async _compileContractAsync(sourceFileId: string): Promise<void> {
         if (_.isUndefined(this._contractSources)) {
             throw new Error('Contract sources not yet initialized');
         }
-        const contractSpecificSourceData = this._contractSourceData[fileName];
-        const currentArtifactIfExists = await getContractArtifactIfExistsAsync(this._artifactsDir, fileName);
-        const sourceHash = `0x${contractSpecificSourceData.sourceHash.toString('hex')}`;
-        const sourceTreeHash = `0x${contractSpecificSourceData.sourceTreeHash.toString('hex')}`;
+        const contractSpecificSourceDataByFileId = this._contractSourceDataByFileId[sourceFileId];
```
BMillman19: check for existence here also rename to `contractSpecificSourceDataIfExists` since it may be of type `ContractSourceData` or `undefined`

packages/deployer/src/utils/compiler.ts
```diff
@@ -1,12 +1,107 @@
...
+import { ContractArtifact, ContractSources, FunctionNameToSeenCount } from './types';
 
+/**
+ * Constructs a system-wide unique identifier for a source file.
+ * @param directoryNamespace Namespace of the source file's root contract directory.
+ * @param sourceFilePath Path to a source file, relative to contractBaseDir.
+ * @return sourceFileId A system-wide unique identifier for the source file.
+ */
+export function constructUniqueSourceFileId(directoryNamespace: string, sourceFilePath: string): string {
+    let namespacePrefix = '';
```
BMillman19: this can just be reduced to `const namespacePrefix = !_.isEmpty(directoryNamespace) ? `/${directoryNamespace}` : ''`

packages/deployer/src/utils/compiler.ts
```diff
@@ -1,12 +1,107 @@
...
+    const sourceFileName = path.basename(sourceFilePath, constants.SOLIDITY_FILE_EXTENSION);
+    const contractId = `${namespacePrefix}${sourceFileName}`;
+    return contractId;
+}
+/**
+ *  There is no function overloading in typescript, so we must change the Typescript ABI interface.
+ * Overloaded function names are incremented as follows: functionName, functionName_2, functioname_3, ...
+ * If functionName_N already exists then compilation will fail.
+ */
+export function renameOverloadedFunctionNames(contractAbi: ContractAbi) {
```
BMillman19: we might want to return a new `ContractAbi` instance instead of mutating the input, if we choose to stick with this option we might want to document it in the comments above or make it part of the function name

packages/deployer/src/compiler.ts
```diff
@@ -68,14 +76,15 @@ export class Compiler {
                         encoding: 'utf8',
                     };
                     const source = await fsWrapper.readFileAsync(contentPath, opts);
-                    sources[fileName] = source;
-                    logUtils.log(`Reading ${fileName} source...`);
+                    const sourceFilePath = contentPath.substr(contractBaseDir.length);
```
abandeali1: Nit: `slice` is a bit more standard than `substr`and will do the same thing.

packages/deployer/src/compiler.ts
hysz: I didn't actually realize we owned the code that produces Typescript from the artifacts. This could be interesting.

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +114,47 @@ export class Compiler {
...
+        }
         _.forIn(this._contractSources, this._setContractSpecificSourceData.bind(this));
-        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
-            ? _.keys(this._contractSources)
+        const specifiedContractIds = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
+            ? _.keys(contractIds)
             : Array.from(this._specifiedContracts.values());
-        for (const fileName of fileNames) {
-            await this._compileContractAsync(fileName);
+        for (const contractId of specifiedContractIds) {
```
abandeali1: Replace with:
```
await Promise.all(
    _.map(specifiedContractIds, contractId => this._compileContractAsync(contractIds[contractId]))
);
```
This way you won't have to wait for the Promises to resolve sequentially, which should speed things up a bit.

packages/deployer/src/compiler.ts
```diff
@@ -193,20 +225,24 @@ export class Compiler {
...
+        const contractName = path.basename(sourceFileId, constants.SOLIDITY_FILE_EXTENSION);
+        const contractIdentifier = `${sourceFileId}:${contractName}`;
         if (_.isUndefined(compiled.contracts[contractIdentifier])) {
             throw new Error(
-                `Contract ${contractName} not found in ${fileName}. Please make sure your contract has the same name as it's file name`,
+                `Contract ${contractName} not found in ${sourceFileId}. Please make sure your contract has the same name as it's file name`,
             );
         }
-        const abi: ContractAbi = JSON.parse(compiled.contracts[contractIdentifier].interface);
+        const abiMaybeWithOverloadedFunctions: ContractAbi = JSON.parse(
```
abandeali1: I'm wondering if this is something that should be fixed at the `abi-gen` level, rather than the ABI level. This change makes our compiler output non-standard ABIs. Thoughts @LogvinovLeon @Recmo @fabioberger ?

packages/deployer/src/utils/compiler.ts
```diff
@@ -1,12 +1,100 @@
...
+        const functionName = (abiItem as MethodAbi).name;
+        return functionName;
+    });
+
+    // Rename overloaded functions
+    const outputContractAbi = _.cloneDeep(contractAbi);
+    const functionNameToSeenCount: FunctionNameToSeenCount = {};
+    _.forEach(outputContractAbi, abiItem => {
+        const type = abiItem.type;
+        if (type !== AbiType.Function) {
```
BMillman19: instead of duplicating this check here and line 61, we can just create a filtered list of `MethodAbi` instances using `_.filter` and transform that list into a list of functionNames and then perform the logic here. That will also get rid of the need to cast `abiItem` as `MethodAbi` as well

BMillman19: we can also return w/o performing a clone if we see that the list is empty after filtering

packages/deployer/package.json
```diff
@@ -47,12 +47,15 @@
     },
     "homepage": "https://github.com/0xProject/0x-monorepo/packages/deployer/README.md",
     "devDependencies": {
+        "@0xproject/dev-utils": "^0.2.1",
         "@0xproject/monorepo-scripts": "^0.1.16",
         "@0xproject/tslint-config": "^0.4.14",
         "@types/require-from-string": "^1.2.0",
         "@types/semver": "^5.5.0",
         "@types/yargs": "^11.0.0",
+        "@types/request-promise-native": "^1.0.2",
```
fabioberger: This should be removed. Probably an incorrect merge.

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +114,47 @@ export class Compiler {
...
-            ? _.keys(this._contractSources)
+        const specifiedContractIds = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
+            ? _.keys(contractIds)
             : Array.from(this._specifiedContracts.values());
-        for (const fileName of fileNames) {
-            await this._compileContractAsync(fileName);
-        }
+        await Promise.all(
+            _.map(specifiedContractIds, async contractId => this._compileContractAsync(contractIds[contractId])),
+        );
```
fabioberger: I'm kinda surprised this was running sequentially before. @abandeali1 @LogvinovLeon can you guys confirm that there wasn't a good reason for it?

packages/deployer/src/compiler.ts
fabioberger: Agreed.

packages/deployer/test/overloaded_test.ts
```diff
@@ -0,0 +1,128 @@
...
+        const artifactString = await fsWrapper.readFileAsync(passingArtifactPath, opts);
+        const artifact: ContractArtifact = JSON.parse(artifactString);
+        const abi = artifact.networks[0].abi;
+
+        // Assert test, test_2 and test_3 functions exist
+        let testExists = false;
+        let test2Exists = false;
+        let test3Exists = false;
+        for (const abiItem of abi) {
+            if (abiItem.type !== 'function') {
```
fabioberger: Use Type, not string literal.

packages/abi-gen/src/index.ts
```diff
@@ -126,18 +125,21 @@ for (const abiFileName of abiFileNames) {
     }
 
     const methodAbis = ABI.filter((abi: AbiDefinition) => abi.type === ABI_TYPE_METHOD) as MethodAbi[];
-    const methodsData = _.map(methodAbis, methodAbi => {
-        _.map(methodAbi.inputs, (input, i: number) => {
+    const methodAbisSanitized = abiUtils.renameOverloadedMethods(methodAbis) as MethodAbi[];
+    const methodsData = _.map(methodAbis, (methodAbi, i: number) => {
+        _.map(methodAbi.inputs, (input, j: number) => {
```
BMillman19: conceptually `map` should be used for side-effect free conversion from an array of one type into another type. if you want to iterate  over a data source and mutate it you should use `forEach` instead

packages/base-contract/src/index.ts
```diff
@@ -49,10 +53,41 @@ export class BaseContract {
         }
         return txDataWithDefaults;
     }
+    protected _lookupEthersInterface(functionSignature: string): ethersContracts.Interface {
+        const ethersInterface = this._ethersInterfacesByFunctionSignature[functionSignature];
+        if (_.isUndefined(ethersInterface)) {
+            throw new Error(`Failed to lookup method with function signature '${functionSignature}'`);
+        }
+
```
BMillman19: newline

packages/deployer/src/compiler.ts
abandeali1: I can't see a good reason for it, looks like it was changed to a `for..of` loop in the large recent refactor.

packages/utils/src/abi_utils.ts
```diff
@@ -0,0 +1,74 @@
...
+        ]);
+        // Group method Abis by name (overloaded methods will be grouped together, in alphabetical order)
+        const methodAbisByName = _.transform(
+            methodAbisByOriginalIndexOrdered,
+            (result: { [key: string]: Array<{ index: number; methodAbi: MethodAbi }> }, entry) => {
+                (result[entry.methodAbi.name] || (result[entry.methodAbi.name] = [])).push(entry);
+            },
+            {},
+        );
+        // Rename overloaded methods to overloadedMethoName_1, overloadedMethoName_2, ...
```
dekz: typo `overloadedMethodName_1`

packages/deployer/src/cli.ts
```diff
@@ -45,7 +45,7 @@ async function onDeployCommandAsync(argv: CliOptions): Promise<void> {
     const web3Wrapper = new Web3Wrapper(web3Provider);
     const networkId = await web3Wrapper.getNetworkIdAsync();
     const compilerOpts: CompilerOptions = {
-        contractsDir: argv.contractsDir,
+        contractDirs: getContractDirectoriesFromList(argv.contractsDir),
```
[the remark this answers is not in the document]
fabioberger: We should add a regression test for this.

packages/deployer/src/compiler.ts
```diff
@@ -105,25 +113,47 @@ export class Compiler {
...
+                if (!_.isUndefined(contractIds[contractId])) {
+                    throw new Error(`Found duplicate contract with ID '${contractId}'`);
+                }
+                contractIds[contractId] = sourceFileId;
+            });
+        }
         _.forIn(this._contractSources, this._setContractSpecificSourceData.bind(this));
-        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
-            ? _.keys(this._contractSources)
+        const specifiedContractIds = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
```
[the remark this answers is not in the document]
abandeali1: I don't believe `_.has` works on Sets.

packages/utils/src/abi_utils.ts
```diff
@@ -0,0 +1,74 @@
...
+        }
+        return param.type;
+    },
+    getFunctionSignature(abi: MethodAbi): string {
+        const functionName = abi.name;
+        const parameterTypeList = abi.inputs.map((param: DataItem) => this.parseFunctionParam(param));
+        const functionSignature = `${functionName}(${parameterTypeList})`;
+        return functionSignature;
+    },
+    renameOverloadedMethods(inputContractAbi: ContractAbi): ContractAbi {
```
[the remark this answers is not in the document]
hysz: Thanks for taking the time to write that out! 👍

packages/deployer/src/cli.ts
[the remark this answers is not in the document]
hysz: Looks like there are not explicit tests for the command line interface.

Presumably, this is because compiling other dependent packages (Deployer, Contracts, Metacoin) would have failed if the CLI is wrong. Although, in this particular case, I had been working on all of those packages and would have built them manually at some point. So, we definitely do need explicit explicit test frameworks for all of our CLI's.

packages/deployer/src/compiler.ts
```diff
@@ -68,14 +75,15 @@ export class Compiler {
                         encoding: 'utf8',
                     };
                     const source = await fsWrapper.readFileAsync(contentPath, opts);
-                    sources[fileName] = source;
-                    logUtils.log(`Reading ${fileName} source...`);
+                    const sourceFilePath = contentPath.slice(contractBaseDir.length);
```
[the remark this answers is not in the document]
hysz: > Will dirPath always begin with contractBaseDir

That's correct, although it's inferred from how the private function is called and should be within the call itself. I'll add a check.

packages/deployer/src/compiler.ts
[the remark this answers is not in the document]
hysz: Can confirm `_.has` did not work.

packages/metacoin/contracts/Metacoin.sol
```diff
@@ -22,4 +27,14 @@ contract Metacoin {
...
+
+    function transfer(TransferData transferData, uint32 callback) public returns (int) {
+        transfer(transferData);
+        return callback;
+    }
+
+    function transfer(NestedTransferData nestedTransferData) public returns (int) {
+        return transfer(nestedTransferData.transferData, nestedTransferData.callback);
+    }
+
```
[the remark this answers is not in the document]
hysz: There was not an existing test framework in `abi-gen` so for this PR tests were placed in `metacoin`. Verified this was okay (as a temporary solution) with @abandeali1 and @LogvinovLeon.

PR #494, opened 2018-04-03
Generalized Asset Transfer Proxy + ERC721 Proxy + Deployer Updates

hysz:
## Description

Overview:
1. Completed Asset Transfer Proxy
2. ERC721 Transfer Proxy
3. Updates to the deployer, allowing us to import the Zeppelin Library.

Complete list of changes (Asset Transfer Proxy):
- Asset transfer proxy, which delegates transfers to specialized transfer proxies (ERC20/ERC721)
- Specialized transfer proxies can be hotswapped (registered/unregistered on the fly).
- Settlement layer now uses the asset transfer proxy to delegate transfers
- Order now includes makerAssetProxyData and takerAssetProxyData
- Complete re-write of the fillOrderNoThrow, to account for changes to the Order structure
- The exchange constructor now takes the asset transfer proxy, rather than a token transfer proxy.
- Authorizable contract (implementation abstracted from TokenTransferProxy)
- Memory util, which includes copying an address or uint256 to/from a position in a byte array
- Test cases for adding/removing proxies and transferring ERC20/ERC721 tokens
- Test cases for exchanging ERC20/ERC721 tokens

## Motivation and Context

We would like to add support for transferring new types of assets. At present, we only support ERC20 tokens. These changes lay the foundation for extending the exchange to new asset classes, like ERC721.

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -68,43 +77,39 @@ contract MixinSettlementProxy is
...
-        require(
-            TRANSFER_PROXY.transferFrom(
-                order.makerTokenAddress,
-                order.makerAddress,
-                takerAddress,
-                makerTokenFilledAmount
-            )
+        bytes32 orderHash = getOrderHash(order);
+
+        TRANSFER_PROXY.transferFrom(
```
dekz: I may have lost it in the details, how are we preventing partial fills of ERC721 orders?

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,116 @@
...
+        returns (bool)
+    {
+        return (assetProxies[assetProxyId] != address(0x0));
+    }
+
+    /// @dev Deregisters an asset proxy.
+    /// @param assetProxyId Id of the asset proxy to deregister.
+    function deregisterAssetProxy(uint8 assetProxyId)
+        public
+        onlyAuthorized
```
abandeali1: `onlyOwner`

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,116 @@
...
+        require(assetProxy != address(0x0));
+
+        // Return asset proxy
+        return assetProxy;
+    }
+
+    /// @dev Returns true if an asset proxy is registered.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @return True if an asset proxy is registered. False, otherwise.
+    function isAssetProxyRegistered(uint8 assetProxyId)
```
abandeali1: If we allow `getAssetProxy` to return 0, this function probably isn't necessary.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,116 @@
...
+        // Delegate transfer to asset proxy
+        assetProxy.transferFrom(assetMetadata, from, to, amount);
+    }
+
+    /// @dev Registers a new asset proxy.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param newAssetProxyAddress Address of the asset proxy contract to register.
+    /// @param currentAssetProxyAddress Address of existing asset proxy to overwrite.
+    function registerAssetProxy(
+        uint8 assetProxyId,
```
abandeali1: There's an attack vector here: the owner of this contract can deregister a proxy and then register a new proxy to the same ID, causing existing orders to be valid for a different proxy. To avoid this, we can store a nonce that increments every time a new proxy is stored, and then uses the nonce as the new proxy's ID. This should guarantee that no ID is every reused.

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
```diff
@@ -33,7 +33,9 @@ contract LibOrder {
         "uint256 makerFeeAmount",
         "uint256 takerFeeAmount",
         "uint256 expirationTimeSeconds",
-        "uint256 salt"
+        "uint256 salt",
+        "bytes makerAssetProxyData",
```
abandeali1: I prefer `makerAssetData` and `takerAssetData`. These fields could become more generic in the future, where some of the data ends up being used outside of a proxy.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -68,43 +77,39 @@ contract MixinSettlementProxy is
...
         );
-        require(
-            TRANSFER_PROXY.transferFrom(
-                order.takerTokenAddress,
-                takerAddress,
-                order.makerAddress,
-                takerTokenFilledAmount
-            )
+
+        TRANSFER_PROXY.transferFrom(
```
abandeali1: Why were these requires removed? We can't assume every token transfer will revert on failure, so we should add back `bool` returns to all instances of `transferFrom` and keep the requires.

packages/contracts/test/asset_transfer_proxy/core.ts
```diff
@@ -0,0 +1,289 @@
...
+    let makerAddress: string;
+    let takerAddress: string;
+    let zrx: DummyTokenContract;
+    let dmyBalances: Balances;
+    let tokenTransferProxy: TokenTransferProxyContract;
+    let assetTransferProxy: AssetTransferProxyContract;
+    let erc20TransferProxyV1: ERC20TransferProxy_v1Contract;
+    let erc20TransferProxy: ERC20TransferProxyContract;
+    let erc721TransferProxy: ERC721TransferProxyContract;
+    const nilAddress = "0x0000000000000000000000000000000000000000";
```
abandeali1: We can use `ZeroEx.NULL_ADDRESS` here

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/proxies/ERC721TransferProxy.sol
```diff
@@ -0,0 +1,94 @@
...
+    {
+        // Decode metadata
+        address token;
+        uint256 tokenId;
+        (token, tokenId) = decodeMetadata(assetMetadata);
+
+        // There exists only 1 of each token.
+        require(amount == 1);
+
+        // Delegate call to ERC721 contract
```
abandeali1: This comment is a bit confusing, since we're not actually making a `delegatecall` here.

packages/contracts/src/contracts/current/protocol/TokenTransferProxy/TokenTransferProxy.sol
```diff
@@ -16,71 +16,23 @@
 
 */
 
-pragma solidity ^0.4.11;
+pragma solidity ^0.4.21;
```
abandeali1: This file should not be changed - we want to test against the exact same bytecode as the existing deployed version.

packages/contracts/src/utils/asset_transfer_proxy_utils.ts
```diff
@@ -0,0 +1,63 @@
+import { BigNumber } from '@0xproject/utils';
+import * as Web3 from 'web3';
+import { AssetProxyId } from './types'
+var ethersUtils = require('ethers-utils');
+
+export function zeroPad(value: string, width: number): string {
+    return "0".repeat(width - value.length) + value;
+}
+
+export function encodeAssetProxyId(assetProxyId: AssetProxyId, encoded_metadata: {value: string})
```
abandeali1: `encoded_metadata` should be camel-cased.

packages/contracts/src/utils/asset_transfer_proxy_utils.ts
```diff
@@ -0,0 +1,63 @@
+import { BigNumber } from '@0xproject/utils';
+import * as Web3 from 'web3';
+import { AssetProxyId } from './types'
+var ethersUtils = require('ethers-utils');
+
+export function zeroPad(value: string, width: number): string {
```
abandeali1: `ethereumjs-util` has some helper methods for padding already (for Buffers). I also find it more intuitive/readable to use Buffers for these types of things in general, since that's closer to what's happening at the EVM level.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetProxy.sol
```diff
@@ -0,0 +1,34 @@
...
+
+pragma solidity ^0.4.21;
+
+contract IAssetProxy {
+
+    /// @dev Transfers assets.
+    /// @param assetMetadata Byte array encoded for the respective asset proxy.
+    /// @param from Address to transfer token from.
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
```
Recmo: `/// @dev Either succeeds or throws`

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetTransferProxy.sol
```diff
@@ -0,0 +1,80 @@
...
+
+*/
+
+pragma solidity ^0.4.21;
+
+import "./IAssetProxy.sol";
+import "../../utils/Authorizable/IAuthorizable.sol";
+
+contract IAssetTransferProxy is
+    IAuthorizable
```
Recmo: `is IAssetProxy`

I'd also rename this to IAssetProxyDispatcher`.

The cool thing is that the Dispatcher is itself a valid `IAssetProxy` so we could nest them.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetTransferProxy.sol
```diff
@@ -0,0 +1,80 @@
...
+        uint8 id,
+        address assetClassAddress
+    );
+
+    /// @dev Delegates transfer to the corresponding asset proxy.
+    /// @param assetMetadata Byte array encoded for the respective asset proxy.
+    /// @param from Address to transfer token from.
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
+    function transferFrom(
```
Recmo: Remove this, inherit from `IAssetProxy` instead.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -20,22 +20,22 @@ pragma solidity ^0.4.21;
...
+import "../AssetTransferProxy/IAssetTransferProxy.sol";
 
 /// @dev Provides MixinSettlement
 contract MixinSettlementProxy is
     MSettlement,
     LibPartialAmount
 {
-
-    ITokenTransferProxy TRANSFER_PROXY;
+    IAssetTransferProxy TRANSFER_PROXY;
```
Recmo: We can lower this to `IAssetProxy` the `MixinSettlementProxy` doesn't use any of the management functions.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
Recmo: The transfer method in `IAssetProxy` does not return a value, it either succeeds or throws. The implementation for ERC20 or the old TokenProxy does the return value checking.

I think it's cleaner this way because you can not forget to check the return value.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -60,85 +61,172 @@ contract MixinWrapperFunctions is
...
+        // |          | 0x120  |         |   10. expirationTimeSeconds                 |
+        // |          | 0x140  |         |   11. salt                                  |
+        // |          | 0x160  |         |   12. Offset to makerAssetProxyMetadata (*) |
+        // |          | 0x180  |         |   13. Offset to takerAssetProxyMetadata (*  |
+        // |          | 0x1A0  | 32      | makerAssetProxyMetadata Length              |
+        // |          | 0x1C0  | **      | makerAssetProxyMetadata Contents            |
+        // |          | 0x1E0  | 32      | takerAssetProxyMetadata Length              |
+        // |          | 0x200  | **      | takerAssetProxyMetadata Contents            |
+        // |          | 0x220  | 32      | signature Length                            |
+        // |          | 0x240  | **      | signature Contents                          |
```
Recmo: Ouch! XD. Good job!

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -60,85 +61,172 @@ contract MixinWrapperFunctions is
...
+
+            /////// Write signature ///////
+            // Write offset to paramsArea
+            mstore(paramsAreaOffset, sub(dataAreaEnd, paramsAreaStart))
+
+            // Calculate length of signature
+            sourceOffset := signature
+            bytesLen := mload(sourceOffset)
+            sourceOffset := add(sourceOffset, 0x20)
+            bytesLenPadded := add(div(bytesLen, 0x20), gt(mod(bytesLen, 0x20), 0))
```
Recmo: Took me a while to wrap my head around, so this is actualy len in terms of 32-byte words, not bytes. (probably should be named `lenPaddedWords`.

The computation is correct, but a bit complex, it can be simplified to `div(add(bytesLen, 0x1F), 0x20)`. (Proof left as an exercise to the reader).

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -60,85 +61,172 @@ contract MixinWrapperFunctions is
...
+
+            // Write length of signature
+            mstore(dataAreaEnd, bytesLen)
+            dataAreaEnd := add(dataAreaEnd, 0x20)
```
Recmo: You can consider this just another full-word to write and make it part of the loop below:

* Remove `sourceOffset := add(sourceOffset, 0x20)`
* `bytesLenPadded := div(add(bytesLen, 0x3F), 0x20)`

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -60,85 +61,172 @@ contract MixinWrapperFunctions is
...
+
+            // Write contents of signature
+            for {let i := 0} lt(i, bytesLenPadded) {i := add(i, 1)} {
+                mstore(dataAreaEnd, mload(sourceOffset))
+                dataAreaEnd := add(dataAreaEnd, 0x20)
+                sourceOffset := add(sourceOffset, 0x20)
+            }
```
Recmo: These four copies of the same `bytes` copy operation are really breaking the DRY. Is there some way we can turn them into a function?

packages/contracts/src/contracts/current/utils/LibBytes/LibBytes.sol
```diff
@@ -0,0 +1,126 @@
...
+            result := and(mload(add(b, index)), 0xffffffffffffffffffffffffffffffffffffffff)
+        }
+        return result;
+    }
+
+    /// @dev Writes an address into a specific position in a byte array.
+    /// @param input Address to put into byte array.
+    /// @param b Byte array to insert address into.
+    /// @param index Index in byte array of address.
+    function writeAddress(
```
Recmo: I'd put the `address input` as last argument. Basically from most generic to most specific.

packages/contracts/src/contracts/current/utils/LibBytes/LibBytes.sol
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return result;
+    }
+
+    /// @dev Writes a uint256 into a specific position in a byte array.
+    /// @param input uint256 to put into byte array.
+    /// @param b Byte array to insert <input> into.
+    /// @param index Index in byte array of <input>.
+    function writeUint256(
+        uint256 input,
```
Recmo: Same here.

packages/contracts/src/utils/asset_transfer_proxy_utils.ts
Recmo: There's also [`String.prototype.padEnd`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padEnd). (After the [leftPad disaster](https://www.theregister.co.uk/2016/03/23/npm_left_pad_chaos/) the JS community woke up and realized that this stuff is needs to become part of the language standard :) )

packages/contracts/package.json
```diff
@@ -27,8 +27,9 @@
         "test:circleci": "yarn test:coverage"
     },
     "config": {
-        "abis": "src/artifacts/@(DummyToken|TokenTransferProxy|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken).json",
-        "contracts": "Exchange,DummyToken,ZRXToken,Token,WETH9,TokenTransferProxy,MultiSigWallet,MultiSigWalletWithTimeLock,MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress,MaliciousToken,TokenRegistry"
+        "abis": "src/artifacts/@(DummyToken|TokenTransferProxy|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken|AssetTransferProxy|ERC20TransferProxy_v1|ERC20TransferProxy|ERC721TransferProxy|DummyERC721Token).json",
+        "contracts": "Exchange,DummyToken,ZRXToken,Token,WETH9,TokenTransferProxy,MultiSigWallet,MultiSigWalletWithTimeLock,MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress,MaliciousToken,TokenRegistry,AssetTransferProxy,ERC20TransferProxy_v1,ERC20TransferProxy,ERC721TransferProxy,DummyERC721Token",
+        "dirs": "src/contracts,zeppelin:../../node_modules/zeppelin-solidity"
```
Recmo: What do we use from zeppelin?

Truffle by default adds just `node_modules` as a dir. This way you need to `import "zeppeling-solidity/something/something.sol" and it is clear what is from zeppelin.

Recmo:
Do we need the flexibility that this `Authorized` mechanism provides? Or is it enough to use `Ownable`?

packages/contracts/package.json
hysz: The ERC721 contracts come from Zeppelin. The library is quite extensive, so we may find it useful in other ways as well.

The updates I made to the deployer allow the directories to be namespaced. In this case, `0x` is in the global namespace and `zeppelin` is in its own namespace. If you check out the ERC721 Transfer Proxy, you'll see that we import contracts from zeppelin like this: `import "/zeppelin/contracts/token/ERC721/ERC721Token.sol";`

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,116 @@
...
+    // Mapping from Asset Proxy Id's to their respective Asset Proxy
+    mapping (uint8 => IAssetProxy) public assetProxies;
+
+    /// @dev Delegates transfer to the corresponding asset proxy.
+    /// @param assetMetadata Byte array encoded for the respective asset proxy.
+    /// @param from Address to transfer token from.
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
+    function transferFrom(
+        bytes assetMetadata,
```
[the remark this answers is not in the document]
hysz: I don't have a strong opinion, although maintaining consistency sounds like a good approach.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
```diff
@@ -0,0 +1,116 @@
...
+    /// @dev Registers a new asset proxy.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param newAssetProxyAddress Address of the asset proxy contract to register.
+    /// @param currentAssetProxyAddress Address of existing asset proxy to overwrite.
+    function registerAssetProxy(
+        uint8 assetProxyId,
+        address newAssetProxyAddress,
+        address currentAssetProxyAddress)
+        public
+        onlyAuthorized
```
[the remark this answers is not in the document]
hysz: Note that the successor proxy -- `TokenTransferProxy` -- was `Authorizable`. Was there a specific use case for that is no longer relevant?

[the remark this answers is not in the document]
abandeali1: This contract should be both `Authorizable` and `Ownable`, but the modifier should be `onlyOwner` in this case. The Exchange contract won't be the address registering/deregistering proxies here.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/AssetTransferProxy.sol
hysz: We chatted about this offline and decided to keep the current implementation for the time being. Ultimately, this is more of a trust issue than a security vulnerability. Two events would need to take place for such an attack to play out: (i) The contract owner would have to swap proxies, and (ii) the enduser would have to set allowances on both proxies. The assumptions that we are not a bad actor and the enduser knows what they're signing are fundamental to our protocol. If either of these failed, then our entire system would be vulnerable.

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/IAssetTransferProxy.sol
hysz: I really like this idea +1

packages/contracts/src/contracts/current/protocol/AssetTransferProxy/proxies/ERC721TransferProxy.sol
```diff
@@ -0,0 +1,94 @@
...
+        public
+        onlyAuthorized
+    {
+        // Decode metadata
+        address token;
+        uint256 tokenId;
+        (token, tokenId) = decodeMetadata(assetMetadata);
+
+        // There exists only 1 of each token.
+        require(amount == 1);
```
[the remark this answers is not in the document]
hysz: Good call

packages/contracts/src/utils/asset_proxy_utils.ts
```diff
@@ -0,0 +1,50 @@
+import { BigNumber } from '@0xproject/utils';
+import BN = require('bn.js');
+import ethUtil = require('ethereumjs-util');
+
+import { AssetProxyId } from './types';
+
+export function encodeAssetProxyId(assetProxyId: AssetProxyId): Buffer {
+    const formattedAssetProxyId = new BN(assetProxyId);
+    const encodedAssetProxyId = ethUtil.toUnsigned(formattedAssetProxyId);
```
abandeali1: Can you not just use `ethUtil.toBuffer` here?

packages/contracts/src/utils/asset_proxy_utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    if (!ethUtil.isValidAddress(address)) {
+        throw new Error(`Invalid Address: ${address}`);
+    }
+    const encodedAddress = ethUtil.toBuffer(address);
+    return encodedAddress;
+}
+
+export function encodeUint256(value: BigNumber): Buffer {
+    const formattedValue = new BN(value.toString(10));
+    const encodedValue = ethUtil.toUnsigned(formattedValue);
```
abandeali1: Same as above.

packages/contracts/src/utils/crypto.ts
```diff
@@ -31,6 +31,8 @@ export const crypto = {
                 argTypes.push('address');
             } else if (_.isString(arg)) {
                 argTypes.push('string');
+            } else if (arg instanceof Buffer) {
```
abandeali1: I just discovered that `_.isBuffer` exists :)

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,330 @@
...
+                newAddress,
+                currentAddress,
+                { from: owner },
+            );
+            // Verify new asset proxy has replaced initial version
+            proxyAddress = await assetProxyDispatcher.getAssetProxy.callAsync(AssetProxyId.ERC20);
+            expect(proxyAddress).to.be.equal(newAddress);
+        });
+
+        it('should throw if registering with incorrect "old_address" field', async () => {
```
abandeali1: `oldAddress`

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,330 @@
...
+                AssetProxyId.ERC20,
+                ZeroEx.NULL_ADDRESS,
+                erc20TransferProxy.address,
+                { from: owner },
+            );
+            const newProxyAddress = await assetProxyDispatcher.getAssetProxy.callAsync(AssetProxyId.ERC20);
+            expect(newProxyAddress).to.be.equal(ZeroEx.NULL_ADDRESS);
+        });
+
+        it('should throw if requesting address is not authorized', async () => {
```
abandeali1: `not owner`

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,330 @@
...
+        const erc20TransferProxyV1Instance = await deployer.deployAsync(ContractName.ERC20V1Proxy, [
+            tokenTransferProxy.address,
+        ]);
+        erc20TransferProxyV1 = new ERC20Proxy_v1Contract(
+            erc20TransferProxyV1Instance.abi,
+            erc20TransferProxyV1Instance.address,
+            provider,
+        );
+        await erc20TransferProxyV1.addAuthorizedAddress.sendTransactionAsync(assetProxyDispatcher.address, {
+            from: accounts[0],
```
abandeali1: Let's use the `owner` variable here.

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,330 @@
...
+            // Perform a transfer from makerAddress to takerAddress
+            const balances = await dmyBalances.getAsync();
+            const amount = new BigNumber(10);
+            return expect(
+                assetProxyDispatcher.transferFrom.sendTransactionAsync(
+                    encodedProxyMetadata,
+                    makerAddress,
+                    takerAddress,
+                    amount,
+                    { from: notOwner },
```
abandeali1: This should be `{ from: exchangeAddress }` to make sure the tx is failing for the right reason.

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+        });
+
+        it('should successfully decode metadata encoded by typescript helpers', async () => {
+            const metadata = encodeERC20V1ProxyData(zrx.address);
+            const address = await erc20TransferProxyV1.decodeMetadata.callAsync(metadata);
+            expect(address).to.be.equal(zrx.address);
+        });
+
+        it('should successfully encode/decode metadata padded with zeros', async () => {
+            const testAddress = '0x0000000000000000056000000000000000000010';
```
abandeali1: Let's declare this at the top if we're going to reuse it.

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+        const erc20TransferProxyV1Instance = await deployer.deployAsync(ContractName.ERC20V1Proxy, [
+            tokenTransferProxy.address,
+        ]);
+        erc20TransferProxyV1 = new ERC20Proxy_v1Contract(
+            erc20TransferProxyV1Instance.abi,
+            erc20TransferProxyV1Instance.address,
+            provider,
+        );
+        await erc20TransferProxyV1.addAuthorizedAddress.sendTransactionAsync(assetProxyDispatcherAddress, {
+            from: accounts[0],
```
abandeali1: use `owner`

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+            from: makerAddress,
+        });
+        await zrx.approve.sendTransactionAsync(erc20TransferProxy.address, INITIAL_BALANCE, {
+            from: takerAddress,
+        });
+        await zrx.approve.sendTransactionAsync(erc20TransferProxy.address, INITIAL_BALANCE, {
+            from: makerAddress,
+        });
+        // Deploy ck and set initial balances
+        const ckInstance = await deployer.deployAsync(ContractName.DummyERC721Token, constants.DUMMY_ERC721TOKEN_ARGS);
```
abandeali1: I know we use token symbols as variable names in our other tests, but we should probably us something more generic like `erc721Instance`. Took me a while to figure out what `ck` stood for :)

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+            // Perform a transfer from makerAddress to takerAddress
+            const balances = await dmyBalances.getAsync();
+            const amount = new BigNumber(0);
+            expect(
+                erc20TransferProxy.transferFrom.sendTransactionAsync(
+                    encodedProxyMetadata,
+                    makerAddress,
+                    takerAddress,
+                    amount,
+                    { from: notOwner },
```
abandeali1: `{ from: assetProxyDispatcherAddress }`

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+                ),
+            ).to.be.rejectedWith(constants.REVERT);
+        });
+
+        it('should throw if transferring >1 amount of a token', async () => {
```
abandeali1: missing a space after `>`

packages/contracts/test/asset_proxy_dispatcher/proxies.ts
```diff
@@ -0,0 +1,378 @@
...
+            // Perform a transfer from makerAddress to takerAddress
+            const balances = await dmyBalances.getAsync();
+            const amount = new BigNumber(500);
+            expect(
+                erc20TransferProxy.transferFrom.sendTransactionAsync(
+                    encodedProxyMetadata,
+                    makerAddress,
+                    takerAddress,
+                    amount,
+                    { from: notOwner },
```
abandeali1: `{ from: assetProxyDispatcherAddress }`

packages/contracts/test/exchange/core.ts
```diff
@@ -119,8 +217,36 @@ describe('Exchange', () => {
...
+            zrx.approve.sendTransactionAsync(erc20TransferProxy.address, INITIAL_ALLOWANCE, {
+                from: takerAddress,
+            }),
             zrx.setBalance.sendTransactionAsync(makerAddress, INITIAL_BALANCE, { from: tokenOwner }),
             zrx.setBalance.sendTransactionAsync(takerAddress, INITIAL_BALANCE, { from: tokenOwner }),
+            ck.setApprovalForAll.sendTransactionAsync(erc721TransferProxy.address, true, { from: makerAddress }),
+            ck.setApprovalForAll.sendTransactionAsync(erc721TransferProxy.address, true, { from: takerAddress }),
+            ck.mint.sendTransactionAsync(
+                makerAddress,
+                new BigNumber('0x1010101010101010101010101010101010101010101010101010101010101010'),
```
abandeali1: Set these to constants if we're going to reuse them.

packages/contracts/test/exchange/helpers.ts
```diff
@@ -38,12 +40,31 @@ describe('Exchange', () => {
             deployer.deployAsync(ContractName.DummyToken, constants.DUMMY_TOKEN_ARGS),
             deployer.deployAsync(ContractName.DummyToken, constants.DUMMY_TOKEN_ARGS),
         ]);
+        const assetProxyDispatcher = await deployer.deployAsync(ContractName.AssetProxyDispatcher);
```
abandeali1: Do we actually need to deploy and configure these contracts for the purposes of these tests?

packages/contracts/test/exchange/core.ts
```diff
@@ -749,4 +875,260 @@ describe('Exchange', () => {
...
+            const takerTokenFillAmount = signedOrder.takerTokenAmount;
+            const res = await exWrapper.fillOrderAsync(signedOrder, takerAddress, { takerTokenFillAmount });
+            // Verify post-conditions
+            const newOwnerMakerToken = await ck.ownerOf.callAsync(makerTokenId);
+            expect(newOwnerMakerToken).to.be.bignumber.equal(takerAddress);
+            const newOwnerTakerToken = await ck.ownerOf.callAsync(takerTokenId);
+            expect(newOwnerTakerToken).to.be.bignumber.equal(makerAddress);
+        });
+
+        it('should successfully exchange a single token between the maker and taker (via filleOrderNoThrow)', async () => {
```
abandeali1: The `fillOrderNoThrow` tests should be in the wrapper tests.

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -6,15 +6,21 @@ import * as chai from 'chai';
 import * as _ from 'lodash';
 import * as Web3 from 'web3';
 
+import { AssetProxyDispatcherContract } from '../../src/contract_wrappers/generated/asset_proxy_dispatcher';
+import { DummyERC721TokenContract } from '../../src/contract_wrappers/generated/dummy_e_r_c721_token';
 import { DummyTokenContract } from '../../src/contract_wrappers/generated/dummy_token';
+import { ERC20ProxyContract } from '../../src/contract_wrappers/generated/e_r_c20_proxy';
+import { ERC721ProxyContract } from '../../src/contract_wrappers/generated/e_r_c721_proxy';
```
abandeali1: It seems that a lot of these imports are unused.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC721Proxy.sol
```diff
@@ -0,0 +1,96 @@
...
+    function transferFrom(
+        bytes assetMetadata,
+        address from,
+        address to,
+        uint256 amount)
+        external
+        onlyAuthorized
+    {
+        // No work to do if amount is zero
+        if (amount == 0) return;
```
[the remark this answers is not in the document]
abandeali1: We think we should check if `amount == 0` in the Exchange contract before attempting to transfer, typically. Having this require makes it easier to reason about `transferFrom` - if it did not revert, you know that the transfer was successful.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC721Proxy.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @param assetMetadata ERC721-encoded byte array.
+    /// @return tokenAddress Address of ERC721 token.
+    /// @return tokenId Id of ERC721 token.
+    function decodeMetadata(bytes memory assetMetadata)
+        public pure
+        returns (address tokenAddress, uint256 tokenId)
+    {
+        require(assetMetadata.length == 53);
+        tokenAddress = readAddress(assetMetadata, 1);
+        tokenId = uint256(readBytes32(assetMetadata, 21));
```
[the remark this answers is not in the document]
abandeali1: I think if we're going to use `readAddress`, we might as well have functions for all the types.

packages/contracts/src/utils/asset_proxy_utils.ts
abandeali1: Sorry for not being clear. I meant why use `toUnsigned` over `toBuffer` here?

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
```diff
@@ -33,7 +33,9 @@ contract LibOrder {
         "uint256 makerFee",
         "uint256 takerFee",
         "uint256 expirationTimeSeconds",
-        "uint256 salt"
+        "uint256 salt",
+        "bytes makerAssetData",
+        "bytes takerAssetData"
     );
 
     struct Order {
```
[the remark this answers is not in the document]
hysz: Cool. I've removed the fields from the Order struct and updated the Exchange accordingly.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -225,7 +321,7 @@ contract MixinWrapperFunctions is
         returns (FillResults memory fillResults)
     {
         for (uint256 i = 0; i < orders.length; i++) {
-            require(orders[i].takerTokenAddress == orders[0].takerTokenAddress);
+            require(areBytesEqual(orders[i].takerAssetData, orders[0].takerAssetData));
```
abandeali1: Hmm, this is tricky. It's possible that someone wants to trade the same token using different proxies, in which case this `require` would fail. This will also be much more expensive than the old `require`. I think we can remove these completely and document that that they should only be used with orders for the same takerToken (for marketSells) or makerToken (for marketBuys). This can always be determined off-chain, but the requires were nice for preventing human error.

packages/contracts/src/contracts/current/protocol/Exchange/IExchange.sol
```diff
@@ -35,8 +35,8 @@ contract IExchange {
         address indexed maker,
         address taker,
         address indexed feeRecipient,
-        address makerToken,
-        address takerToken,
+        bytes makerAssetData,
```
dekz: This converts to `keccak256(assetProxyId, tokenAddress)` right? Which adds little bit of complexity to watch for Token pair trades (especially if there is a mix of old proxy and new proxy). There are multiple combinations of ZRX/WETH to watch for (due to multiple asset proxies). In the case of ERC721 this is `keccak256(assetProxyId, tokenAddress, tokenId)`, which makes it difficult to watch all ERC721 exchanges (since one needs the tokenId).

The combinations also increase if we throw in the fee ZRX_PROXY_DATA in there as well. 

This is not insurmountable at all, just something we would have to document well and provide guides on. What is the gas cost/tradeoff of keeping both here?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -20,22 +20,24 @@ pragma solidity ^0.4.21;
...
+import "../AssetProxyDispatcher/IAssetProxy.sol";
 
 /// @dev Provides MixinSettlement
 contract MixinSettlementProxy is
     MSettlement,
     LibPartialAmount
 {
-
-    ITokenTransferProxy TRANSFER_PROXY;
+    IAssetProxy TRANSFER_PROXY;
```
dekz: Is this always an`IAssetProxyDispathcher`, or are there times when this is the lower level `IAssetProxy`? If it's always the `IAssetProxyDispathcher` I would declare the type as so and rename to `PROXY_DISPATCHER`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -20,22 +20,24 @@ pragma solidity ^0.4.21;
...
+    // TODO: Is there a way to encode ZRX_PROXY_DATA into takerAssetData/makerAssetData?
+    //       As it stands, only one of the ERC20 proxies can be used.
+    bytes ZRX_PROXY_DATA;
```
dekz: At launch will it be the new ERC20 proxy, or the old? If it is the old proxy would this force all of the new order makers to set the allowance? 

We can't infer the assetProxyId for ZRX based off the assetProxyId in the respective maker and taker metadata since it is now more than ERC20. I guess we're in a safe position setting this to the new ERC20 proxy since there are currently no fees being forced by relayers.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC20Proxy.sol
```diff
@@ -0,0 +1,80 @@
...
+import "../../../utils/Authorizable/Authorizable.sol";
+import { Token_v1 as ERC20Token } from "../../../../previous/Token/Token_v1.sol";
+
+contract ERC20Proxy is
+    LibBytes,
+    Authorizable,
+    IAssetProxy
+{
+
+    /// @dev Transfers ERC20 tokens.
```
dekz: Can we add a note on all of these functions that `throw` if the transfer was not successful (standard ERC20 `transferFrom` doesn't).

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
dekz: I would vote for keeping this assertion in on chain as it prevents human error (and accidental purchase of the wrong tokens). Can we just compare the takerTokenAddress part out of this?

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
```diff
@@ -65,15 +65,15 @@ contract LibOrder {
...
                 order.feeRecipientAddress,
                 order.makerTokenAmount,
                 order.takerTokenAmount,
                 order.makerFee,
                 order.takerFee,
                 order.expirationTimeSeconds,
-                order.salt
+                order.salt,
+                order.makerAssetData,
+                order.takerAssetData
```
dekz: I find it strange that the maker sets the `asset transfer proxy` for the taker. Locking in the `takerTokenAddress` and the `assetId` (in the case of 721 makes sense). But it definitely feels weird that a maker decides to use the OLD proxy, forcing a Taker (who has set an allowance on the NEW proxy) to do the same.

I think it's good that we've tried to re-use the old proxy to avoid users having to re-set allowances. I think it works in theory, but will it actually work in practice, or will it become a more confusing user onboarding?

fabioberger:
@hysz can you fix the merge conflicts?

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
abandeali1: I don't think we should be doin g any decoding in the exchange contract, since the point is to have `makerAssetData` and `takerAssetData` be unstructured (except for the first byte).

I also think that it generally makes sense to have the maker decide the assetProxy for the taker. The example I like to think of here is if we implemented a deposit contract as a proxy, the maker probably would only want the taker to use the deposit contract if they too were using it. However, if the proxies are not sufficiently differentiated then it doesn't really make sense (like the old/new ERC20 proxies). Maybe there's a way for the maker to specify multiple acceptable proxies? It increases complexity for sure, but the flexibility seems like it could be worth it.

packages/contracts/src/contracts/current/protocol/Exchange/IExchange.sol
hysz: Can confirm that the `makerAssetData`/`takerAssetData` fields are logged in place of `makerToken` and `takerToken`.

@abandeali1 I'm open to moving the tokens back out of the byte arrays. I think I'm just missing the full scope of this approach. From my understanding, it seems we can better generalize the exchange by abstracting away proxy-specific data. When we start exchanging other assets, like ENS names, fields like makerToken and takerToken don't appear to provide value at the exchange level.

dekz: I must've gotten confuse din the solidity docs, maybe the hashing only applies to indexed data?

This event is more for the general ecosystem to read. Is the concern the fact we don't parse this unstructured byte array in the Exchange contract and that this would cause unnecessary struct knowledge just to log an event with particular fields?

packages/contracts/src/contracts/current/protocol/Exchange/LibOrder.sol
hysz: We chatted about this offline, and decided the old ERC20 proxy will be removed for simplicity.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
hysz: We chatted about this offline, and decided the old ERC20 proxy will be removed for simplicity. The new ERC20 proxy will be used here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinWrapperFunctions.sol
hysz: That would be difficult because the takerAssetData is an unstructured byte array. It is not intended to be decoded outside of its destination proxy.

That said, this comparison should be fine now that we have agreed to remove the old ERC20 proxy. @abandeali1 - are you okay with leaving this comparison in?

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/AssetProxyDispatcher.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        // transferFrom will either succeed or throw.
+        assetProxy.transferFrom(assetMetadata, from, to, amount);
+    }
+
+    /// @dev Adds a new asset proxy.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param newAssetProxy Asset proxy contract to add.
+    /// @param currentAssetProxy Existing asset proxy to overwrite.
+    function addAssetProxy(
```
dekz: It might be YAGNI, but how do we feel about an human+contract readable string associated with what the Proxy is? Similar to the Token Registry + Symbol? I.e the ERC20 Token Proxy would have `id: 1, name: 'ERC20'`. 

I haven't fully formed this thought yet, but a forwarding contract might want overloaded functions (i.e buy anything), but needs to perform transfer logic. So it might need a way to look up, `if (lookup(firstByte(makerMetadata)) == 'ERC721')`, therefor call `safeTransfer`.

OTOH the forwarding contract could always set allowances before transfer (erc721) and use the DelegateProxy. 

Maybe it's YAGNI...

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlementProxy.sol
```diff
@@ -20,22 +20,24 @@ pragma solidity ^0.4.21;
...
 /// @dev Provides MixinSettlement
 contract MixinSettlementProxy is
     MSettlement,
     LibPartialAmount
 {
-
-    ITokenTransferProxy TRANSFER_PROXY;
+    IAssetProxy TRANSFER_PROXY;
+    // TODO: Is there a way to encode ZRX_PROXY_DATA into takerAssetData/makerAssetData?
+    //       As it stands, only one of the ERC20 proxies can be used.
```
fabioberger: This comment is outdated.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC20Proxy.sol
```diff
@@ -0,0 +1,80 @@
...
+        address token = decodeMetadata(assetMetadata);
+        bool success = ERC20Token(token).transferFrom(from, to, amount);
+        require(success == true);
+    }
+
+    /// @dev Encodes ERC20 byte array.
+    /// @param assetProxyId Id of the asset proxy.
+    /// @param tokenAddress Address of the asset.
+    /// @return assetMetadata ERC20-encoded byte.
+    function encodeMetadata(
```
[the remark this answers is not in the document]
dekz: I think this is useful for off-chain consumers (0x.js will likely do this natively, but it's possible to get this result directly off a node!). It also helps libraries test their implementations of the encoding.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC721Proxy.sol
```diff
@@ -0,0 +1,93 @@
...
+        // There exists only 1 of each token.
+        require(amount == 1);
+
+        // Decode metadata
+        address token;
+        uint256 tokenId;
+        (token, tokenId) = decodeMetadata(assetMetadata);
+
+        // Either succeeds or throws.
+        ERC721Token(token).transferFrom(from, to, tokenId);
```
dekz: Should we call safeTransferFrom here? At this current moment we assume `taker` is a user controlled address, but I expect to see contracts integrating with 0x (forwarding contract) and argue it's best to go with the safe way to transfer, lest ERC721 tokens get stuck! 

Thoughts?

packages/contracts/src/contracts/current/test/DummyERC721Token/DummyERC721Token.sol
```diff
@@ -0,0 +1,47 @@
...
+contract DummyERC721Token is
+    Ownable,
+    ERC721Token
+{
+    function DummyERC721Token(
+        string name,
+        string symbol)
+        public
+        ERC721Token(name, symbol)
+    {}
```
fabioberger: Do we want to keep the constructor if it's body is empty?

packages/contracts/src/contracts/current/utils/LibBytes/LibBytes.sol
```diff
@@ -0,0 +1,183 @@
...
+        require(b.length >= index + 20); // 20 is length of address
+
+        // Add offset to index:
+        // 1. Arrays are prefixed by 32-byte length parameter (add 32 to index)
+        // 2. Account for size difference between address length and 32-byte storage word (subtract 12 from index)
+        index += 20;
+
+        // Read address from array memory
+        assembly {
+            // 1. Add index to to address of bytes array
```
fabioberger: Nit: `to to address`

packages/contracts/src/utils/types.ts
```diff
@@ -45,8 +51,10 @@ export interface DefaultOrderParams {
     takerTokenAddress: string;
     makerTokenAmount: BigNumber;
     takerTokenAmount: BigNumber;
-    makerFee: BigNumber;
-    takerFee: BigNumber;
+    makerFeeAmount: BigNumber;
+    takerFeeAmount: BigNumber;
```
fabioberger: This change seems out of place. Was this a merge conflict issue?

packages/contracts/test/asset_proxy_dispatcher/auth.ts
```diff
@@ -14,19 +14,19 @@ chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
-describe('TokenTransferProxy', () => {
+describe('AssetProxyDispatcher - Auth', () => {
```
fabioberger: Why the `- Auth`?

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,297 @@
...
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+    describe('addAssetProxy', () => {
+        it('should record proxy upon registration', async () => {
+            await assetProxyDispatcher.addAssetProxy.sendTransactionAsync(
+                AssetProxyId.ERC20,
+                erc20Proxy.address,
+                ZeroEx.NULL_ADDRESS,
```
fabioberger: Set this to a descriptive variable name so next dev doesn't need to guess what this param refers to:
```
const prevProxyAddress = ZeroEx.NULL_ADDRESS;
```

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,297 @@
...
+            const proxyAddress = await assetProxyDispatcher.getAssetProxy.callAsync(AssetProxyId.ERC20);
+            expect(proxyAddress).to.be.equal(erc20Proxy.address);
+        });
+
+        it('should be able to record multiple proxies', async () => {
+            // Record first proxy
+            await assetProxyDispatcher.addAssetProxy.sendTransactionAsync(
+                AssetProxyId.ERC20,
+                erc20Proxy.address,
+                ZeroEx.NULL_ADDRESS,
```
fabioberger: Ditto for rest of file.

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
```diff
@@ -0,0 +1,297 @@
...
+                { from: owner },
+            );
+            const proxyAddress = await assetProxyDispatcher.getAssetProxy.callAsync(AssetProxyId.ERC20);
+            expect(proxyAddress).to.be.equal(erc20Proxy.address);
+        });
+
```
fabioberger: Omit new lines.

packages/contracts/test/exchange/wrapper.ts
fabioberger: Let's also keep `erc` as one word in future file names.

packages/contracts/test/multi_sig_with_time_lock_except_remove_auth_addr.ts
```diff
@@ -1,3 +1,9 @@
+/*
+ *
+ * @TODO:   Before deploying, the MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress contract must be updated
+ *          to have a mapping of all approved addresses. These tests muts be updated appropriately.
```
fabioberger: Typo: `muts`

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/AssetProxyDispatcher.sol
```diff
@@ -0,0 +1,85 @@
...
+{
+    // Mapping from Asset Proxy Id's to their respective Asset Proxy
+    mapping (uint8 => IAssetProxy) public assetProxies;
+
+    /// @dev Transfers assets. Either succeeds or throws.
+    /// @param assetMetadata Byte array encoded for the respective asset proxy.
+    /// @param from Address to transfer token from.
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
+    function transferFrom(
```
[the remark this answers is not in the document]
dekz: This method is external. I believe the current convention is to just document whether it throws or not in the comments.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/AssetProxyDispatcher.sol
```diff
@@ -0,0 +1,85 @@
...
+    /// @param currentAssetProxy Existing asset proxy to overwrite.
+    function addAssetProxy(
+        uint8 assetProxyId,
+        IAssetProxy newAssetProxy,
+        IAssetProxy currentAssetProxy)
+        external
+        onlyOwner
+    {
+        // Ensure the existing asset proxy is not unintentionally overwritten
+        require(currentAssetProxy == assetProxies[assetProxyId]);
```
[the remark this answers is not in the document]
dekz: This is a good question. Do we have a shut off valve for individual proxies in the event of deprecation, or emergency? 

It feels dangerous to replace a proxy and re-use an ID, assigning a new id every time feels safer.

packages/contracts/package.json
```diff
@@ -71,6 +72,7 @@
         "ethereumjs-util": "^5.1.1",
         "ethers-contracts": "^2.2.1",
         "lodash": "^4.17.4",
-        "web3": "^0.20.0"
+        "web3": "^0.20.0",
+        "zeppelin-solidity": "^1.8.0"
```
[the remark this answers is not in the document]
hysz: So the ERC721 implementation is quite large (split across several contracts) and has additional dependencies within Zeppelin. I was actually relieved to learn there was a node module we could just import, rather than cluttering our own contracts. I could probably derive a barebones ERC721 implementation that satisfies our tests and proxy, although I'm not sure what we would gain. What's the drive behind not wanting this dependency?

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/AssetProxyDispatcher.sol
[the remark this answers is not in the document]
hysz: @fabioberger @dekz - I've updated the documentation to make this more clear. 

Ideally, we don't want to recycle ID's at all. @abandeali1 had an interesting idea of internally recording nonces for each proxy, and this might be something we investigate further in the next version.

packages/contracts/package.json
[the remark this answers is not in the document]
fabioberger: I'm ok with keeping it.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC20Proxy.sol
```diff
@@ -0,0 +1,80 @@
...
+    Authorizable,
+    IAssetProxy
+{
+
+    /// @dev Transfers ERC20 tokens. Either succeeds or throws.
+    /// @param assetMetadata ERC20-encoded byte array.
+    /// @param from Address to transfer token from.
+    /// @param to Address to transfer token to.
+    /// @param amount Amount of token to transfer.
+    function transferFrom(
```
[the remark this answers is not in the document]
fabioberger: I'm personally a fan of verbosity in public interfaces where we are able to reduce the effort required for a developer to immediately understand what a method call does (without having to lookup it's implementation and read the comments above it's declaration). The nice thing too is that a developer cannot forget to add this useful comment when re-using the method in a new location.

I would also be against adding error-handling behavior to the function name if I felt this behavior were likely to change in the future, requiring yet another re-name. This seems unlikely here.

I do see now though that this might be a larger naming convention discussion and am happy to table it for now.

packages/contracts/src/contracts/current/test/DummyERC721Token/DummyERC721Token.sol
fabioberger: @hysz are you saying that if a contract wants the constructor of it's base contract to be called, it must implement a constructor itself? If so, it is worth adding a comment above this explaining this quirk. That was non-obvious to me and could see a future developer remove this thinking it was superfluous.

packages/contracts/test/asset_proxy_dispatcher/dispatcher.ts
fabioberger: Ok, let's leave it for now.

packages/contracts/src/contracts/current/protocol/AssetProxyDispatcher/proxies/ERC20Proxy.sol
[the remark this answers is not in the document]
fabioberger: Lastly, let us remember that Solidity dev tools are still lacking (e.g goto definition, inline method definitions, etc...), so it's much more cumbersome to see a method comment if it's not duplicated above every call-site.

packages/contracts/src/contracts/current/utils/LibBytes/LibBytes.sol
```diff
@@ -0,0 +1,183 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.4.21;
+
+contract LibBytes {
```
fabioberger: Do we have test coverage for this library @hysz? I think it would be worthwhile since we are likely to re-use these in many places.

If this is the only change blocking this PR, we could also create an Asana task for it. But it should get done before audits.

hysz: I think this is really important and just created a set explicit test cases for this library. Thanks for pointing this out 👍

PR #498, opened 2018-04-03
Sender abstraction

abandeali1:
## Motivation and Context

Implements sender abstraction in its simplest form (ZEIP #18)

## How Has This Been Tested?

Need to add more tests.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
@@ -18,10 +18,12 @@
 
 pragma solidity ^0.4.21;
 pragma experimental ABIEncoderV2;
+pragma experimental "v0.5.0";
 
 import "./mixins/MExchangeCore.sol";
 import "./mixins/MSettlement.sol";
 import "./mixins/MSignatureValidator.sol";
+import "./mixins/MTransactions.sol";
 import "./LibOrder.sol";
 import "./LibErrors.sol";
 import "./LibPartialAmount.sol";
@@ -35,6 +37,7 @@ contract MixinExchangeCore is
     MExchangeCore,
     MSettlement,
     MSignatureValidator,
+    MTransactions,
     SafeMath,
     LibErrors,
     LibPartialAmount
@@ -113,10 +116,16 @@ contract MixinExchangeCore is
             require(order.takerAssetAmount > 0);
             require(isValidSignature(orderHash, order.makerAddress, signature));
         }
+        
+        // Validate sender
+        if (order.senderAddress != address(0)) {
+            require(order.senderAddress == msg.sender);
+        }
 
-        // Validate taker
+        // Validate transaction signed by taker
+        address takerAddress = getSignerAddress();
         if (order.takerAddress != address(0)) {
-            require(order.takerAddress == msg.sender);
+            require(order.takerAddress == takerAddress);
         }
         require(takerAssetFillAmount > 0);
 
@@ -146,21 +155,10 @@ contract MixinExchangeCore is
 
         // Settle order
         (fillResults.makerAssetFilledAmount, fillResults.makerFeePaid, fillResults.takerFeePaid) =
-            settleOrder(order, msg.sender, fillResults.takerAssetFilledAmount);
+            settleOrder(order, takerAddress, fillResults.takerAssetFilledAmount);
 
         // Log order
-        emit Fill(
-            order.makerAddress,
-            msg.sender,
-            order.feeRecipientAddress,
-            fillResults.makerAssetFilledAmount,
-            fillResults.takerAssetFilledAmount,
-            fillResults.makerFeePaid,
-            fillResults.takerFeePaid,
-            orderHash,
-            order.makerAssetData,
-            order.takerAssetData
-        );
+        emitFillEvent(order, takerAddress, orderHash, fillResults);
         return fillResults;
     }
 
@@ -178,8 +176,16 @@ contract MixinExchangeCore is
         // Validate the order
         require(order.makerAssetAmount > 0);
         require(order.takerAssetAmount > 0);
-        require(order.makerAddress == msg.sender);
 
+        // Validate sender
+        if (order.senderAddress != address(0)) {
+            require(order.senderAddress == msg.sender);
+        }
+        
+        // Validate transaction signed by maker
+        address makerAddress = getSignerAddress();
+        require(order.makerAddress == makerAddress);
+        
         if (block.timestamp >= order.expirationTimeSeconds) {
             emit ExchangeError(uint8(Errors.ORDER_EXPIRED), orderHash);
             return false;
@@ -232,5 +238,28 @@ contract MixinExchangeCore is
         );
         isError = errPercentageTimes1000000 > 1000;
         return isError;
+    }
+
+    /// @dev Logs a Fill event with the given arguments.
+    ///      The sole purpose of this function is to get around the stack variable limit.
+    function emitFillEvent(
+        Order memory order,
+        address takerAddress,
+        bytes32 orderHash,
+        FillResults memory fillResults)
+        internal
+    {
+        emit Fill(
+            order.makerAddress,
+            takerAddress,
+            order.feeRecipientAddress,
+            fillResults.makerAssetFilledAmount,
+            fillResults.takerAssetFilledAmount,
+            fillResults.makerFeePaid,
+            fillResults.takerFeePaid,
+            orderHash,
+            order.makerAssetData,
+            order.takerAssetData
+        );
     }
 }
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
@@ -18,7 +18,6 @@
 
 pragma solidity ^0.4.21;
 pragma experimental ABIEncoderV2;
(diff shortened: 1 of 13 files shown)


packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        // Execute transaction
+        transactions[transactionHash] = true;
+        require(address(this).delegatecall(data));
+
+        // Reset current transaction signer
+        currentSigner = address(0);
+    }
+
+    function getSignerAddress()
```
fabioberger: Let's add a comment above this method since the logic involved is not straight forward:

```
/**
 * Returns the signerAddress. If the fill was submitted on-behalf of the taker / cancel was submitted on-behalf
 * of a maker (executeTransaction entrypoint) then it returns the signer address.
 * If the taker submitted the fill directly / maker submitted the cancel directly, it returns msg.sender.
 */
```

Think this comment can still be improved. The `currentSigner` being a "global" variable set in `executeTransaction` makes the code harder to follow. I know there is no way around that unfortunately, so let's make sure to leave sufficient comments.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -113,10 +116,16 @@ contract MixinExchangeCore is
             require(order.takerAssetAmount > 0);
             require(isValidSignature(orderHash, order.makerAddress, signature));
         }
+        
+        // Validate sender
+        if (order.senderAddress != address(0)) {
```
fabioberger: Let's set this conditional to a helpful variable name. Perhaps:

```
const isDirectTakerFill = order.senderAddress != address(0);
```

It's helpful to have a human-readable variable name for conditionals that elucidate the intent behind the conditional.

If we don't want to create intermediate variables because of stack limit, let's add a more helpful comment.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -113,10 +116,16 @@ contract MixinExchangeCore is
...
+            require(order.senderAddress == msg.sender);
+        }
```
fabioberger: So this conditional enforces that there are two valid orders that can be filled directly by the taker:
- One with `senderAddress` set to `NULL_ADDRESS`
- One with `senderAddress` set to `msg.sender`

?

If so, do we want to constrain this further?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -0,0 +1,85 @@
...
+import "./mixins/MSignatureValidator.sol";
+import "./mixins/MTransactions.sol";
+
+contract MixinTransactions is
+    MSignatureValidator,
+    MTransactions
+{
+
+    // Mapping of transaction hash => executed
+    mapping (bytes32 => bool) public transactions;
```
fabioberger: Add comment about how this prevents replay attacks?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -0,0 +1,85 @@
...
+            data
+        );
+
+        // Validate transaction has not been executed
+        require(!transactions[transactionHash]);
+
+        // TODO: is SignatureType.Caller necessary if we make this check?
+        if (signer != msg.sender) {
+            // Validate signature
+            require(isValidSignature(transactionHash, signer, signature));
```
fabioberger: Does this mean we need to check the signature for all calls through `executeTransaction`? Because when calling fill/cancel directly, we only check it for the first fill.

packages/contracts/test/exchange/transactions.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            it('should cancel the order when signed by maker and called by sender', async () => {
+                const data = exchange.cancelOrder.getABIEncodedTransactionData(order);
+                const signedTx = makerTransactionFactory.newSignedTransaction(data);
+                await exchangeWrapper.executeTransactionAsync(signedTx, senderAddress);
+                const res = await exchangeWrapper.fillOrderAsync(signedOrder, takerAddress);
+                const newBalances = await erc20Wrapper.getBalancesAsync();
+                expect(newBalances).to.deep.equal(erc20Balances);
+            });
+        });
```
fabioberger: Let's add tests for replay attack protection.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -18,10 +18,12 @@
 
 pragma solidity ^0.4.21;
 pragma experimental ABIEncoderV2;
+pragma experimental "v0.5.0";
```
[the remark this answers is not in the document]
abandeali1: Whoops, this shouldn't be there.

packages/contracts/test/exchange/transactions.ts
```diff
@@ -0,0 +1,195 @@
...
+    });
```
hysz: It could also be worthwhile testing that the `currentSigner` global variable is reset after each transaction.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -178,8 +176,16 @@ contract MixinExchangeCore is
...
         require(order.takerAssetAmount > 0);
-        require(order.makerAddress == msg.sender);
 
+        // Validate sender
+        if (order.senderAddress != address(0)) {
+            require(order.senderAddress == msg.sender);
+        }
+        
+        // Validate transaction signed by maker
+        address makerAddress = getSignerAddress();
```
hysz: Should we add something like this in `cancelOrdersUpTo`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
abandeali1: + 1. Also renamed the function to `getCurrentContextAddress`, which is a bit more accurate.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
abandeali1: Yeah, we're pretty much right at the stack limit unfortunately, but I can add a more descriptive comment.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
abandeali1: Hmm, I'm not sure what you mean. What other constraints were you thinking about?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
abandeali1: Yeah, there is no such thing as a partial transaction, so signatures are always checked.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
abandeali1: `cancelOrdersUpTo` is tricky. Since it doesn't actually take an order, the sender is never validated. Basically, there would be no reason to ever call it via `executeTransaction` at the moment.

It is a bit weird that there is this constraint on `cancelOrder`, but that makers can still cancel via `cancelOrdersUpTo`.  I wonder of there's a way around this?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -0,0 +1,85 @@
...
+
+            // Set the current transaction signer
+            currentSigner = signer;
+        }
+
+        // Execute transaction
+        transactions[transactionHash] = true;
+        require(address(this).delegatecall(data));
```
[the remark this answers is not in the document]
abandeali1: This is definitely missing a bunch of tests, but I was hoping that combinatorial testing would catch some of them :)

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
fabioberger: I mean should we enforce that if the user is calling `fillOrder` directly that `senderAddress` **must** be `msg.sender`. This would involve first checking if `getCurrentContextAddress` returns `msg.sender`, and if so, enforce that `senderAddress` is too.

Thoughts?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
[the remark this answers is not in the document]
fabioberger: Let's not rely on combinatorial testing for this. We should explicitly whitelist the methods we expect here. Combinatorial testing is not a silver bullet, we still need thorough unit tests.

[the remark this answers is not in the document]
abandeali1: Yep, just trying to get the contracts feature complete at this point. Added a bunch of test cases to our internal doc.

PR #500, opened 2018-04-04
Run all tests against in-process Ganache

fabioberger:
## Description

Currently, if a user wants the monorepo tests, we require them to:

1. Start TestRPC in a separate terminal windeow
2. CD into the `contracts` package
3. Run the `yarn migrate` command 
4. Run `yarn lerna:run test` from the top-level directory.

After this PR get's merged, all they need to do is run:

```
yarn lerna:test
```

Presto!

We now use an in-process ganache provider (which we've already been using for `contracts` tests), for all the tests!

## Motivation and Context

As more outside contributors start developing within the monorepo, it's paramount that it's as easy as possible to get started. This PR is one more step in that direction.

## How Has This Been Tested?

Ran all the monorepo packages tests.

## Types of changes

## Checklist:


packages/0x.js/package.json
```diff
@@ -33,7 +33,7 @@
         "upload_docs_json": "aws s3 cp generated_docs/index.json $S3_URL --profile 0xproject --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers --content-type application/json"
     },
     "config": {
-        "artifacts": "TokenTransferProxy Exchange TokenRegistry Token EtherToken",
+        "artifacts": "Exchange DummyToken ZRXToken Token WETH9 TokenTransferProxy MultiSigWallet MultiSigWalletWithTimeLock MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress MaliciousToken TokenRegistry Arbitrage EtherDelta AccountLevels",
```
LogvinovLeon: Why?

packages/0x.js/test/token_wrapper_test.ts
```diff
@@ -515,3 +519,14 @@ describe('TokenWrapper', () => {
     });
 });
 // tslint:disable:max-file-line-count
+
+function addEmptyWalletSubprovider(provider: Web3.Provider): Web3.Provider {
```
LogvinovLeon: Let's make a name more generic. I'd love to use that function in other places.

LogvinovLeon: I think it doesn't make sense for now, but I'll generaize it later to prepend any subprovider.

packages/0x.js/test/0x.js_test.ts
```diff
@@ -1,28 +1,46 @@
...
 
 import { ApprovalContractEventArgs, LogWithDecodedArgs, Order, TokenEvents, ZeroEx } from '../src';
 
+import { runMigrationsAsync } from './migrations/migrate';
 import { chaiSetup } from './utils/chai_setup';
 import { constants } from './utils/constants';
 import { TokenUtils } from './utils/token_utils';
 import { web3, web3Wrapper } from './utils/web3_wrapper';
 
+const artifactsDir = path.resolve('test', 'artifacts');
```
LogvinovLeon: Could we create a test util and import deployer from there as we do in other packages?

packages/0x.js/test/artifacts/AccountLevels.json
```diff
@@ -0,0 +1,40 @@
+{
```
LogvinovLeon: Why do we commit those?

packages/0x.js/test/0x.js_test.ts
```diff
@@ -31,7 +49,7 @@ describe('ZeroEx library', () => {
             expect((zeroEx.exchange as any)._exchangeContractIfExists).to.not.be.undefined();
             expect((zeroEx.tokenRegistry as any)._tokenRegistryContractIfExists).to.not.be.undefined();
 
-            const newProvider = web3Factory.getRpcProvider();
+            const newProvider = web3.currentProvider;
```
LogvinovLeon: Please try to minimize the usage of web3 when possible. Just export the provider from that module above. Don't use the web3 object to just carry the provider.

packages/0x.js/test/token_wrapper_test.ts
```diff
@@ -191,7 +193,8 @@ describe('TokenWrapper', () => {
             let zeroExWithoutAccounts: ZeroEx;
             before(async () => {
                 const hasAddresses = false;
-                const web3WithoutAccounts = web3Factory.create({ hasAddresses });
+                const emptyWalletProvider = addEmptyWalletSubprovider(web3.currentProvider);
+                const web3WithoutAccounts = new Web3(emptyWalletProvider);
```
LogvinovLeon: WHY??? We never use that web3 object. Just store that provider.

packages/assert/package.json
```diff
@@ -9,7 +9,7 @@
         "build": "tsc && copyfiles -u 3 './lib/src/monorepo_scripts/**/*' ./scripts",
         "clean": "shx rm -rf lib test_temp scripts",
         "lint": "tslint --project . 'src/**/*.ts' 'test/**/*.ts'",
-        "run_mocha": "mocha lib/test/**/*_test.js",
+        "run_mocha": "mocha lib/test/**/*_test.js --exit",
```
LogvinovLeon: Do you know why that happens? Ideally - we would not need that flag if we properly stop the provider in the global after all hook but I might be too lazy to do that ;)

packages/subproviders/test/unit/redundant_rpc_subprovider_test.ts
```diff
@@ -3,11 +3,13 @@ import * as chai from 'chai';
 import * as _ from 'lodash';
 import Web3 = require('web3');
 import Web3ProviderEngine = require('web3-provider-engine');
+import RpcSubprovider = require('web3-provider-engine/subproviders/rpc');
 
 import { RedundantRPCSubprovider } from '../../src';
 import { DoneCallback } from '../../src/types';
 import { chaiSetup } from '../chai_setup';
 import { reportCallbackErrors } from '../utils/report_callback_errors';
+import { subprovider as ganacheSubprovider } from '../utils/subprovider';
```
LogvinovLeon: Why don't we just export it as ganacheSubprovider?

packages/subproviders/test/unit/redundant_rpc_subprovider_test.ts
```diff
@@ -18,6 +20,8 @@ describe('RedundantRpcSubprovider', () => {
         provider = new Web3ProviderEngine();
         const endpoints = ['http://localhost:8545'];
         const redundantSubprovider = new RedundantRPCSubprovider(endpoints);
+        // Hack: Hot-swap rpc with ganacheSubprovider
```
LogvinovLeon: What do you think about changing RedundantRPCSuprovider to redundant subprovider that accepts providers/subproviders instead of urls?

packages/subproviders/test/unit/redundant_rpc_subprovider_test.ts
```diff
@@ -38,6 +42,11 @@ describe('RedundantRpcSubprovider', () => {
         provider = new Web3ProviderEngine();
         const endpoints = ['http://does-not-exist:3000', 'http://localhost:8545'];
         const redundantSubprovider = new RedundantRPCSubprovider(endpoints);
+        // Hack: Hot-swap rpcs with [nonExistentSubprovider, ganacheSubprovider]
+        const nonExistentSubprovider = new RpcSubprovider({
```
LogvinovLeon: Same

packages/contracts/migrations/index.ts
```diff
@@ -1,20 +0,0 @@
-import { Deployer } from '@0xproject/deployer';
```
dekz: Isn't this file still useful? For manual test deployments onto testrpc (portal etc)

packages/0x.js/test/utils/web3_wrapper.ts
```diff
@@ -1,6 +1,10 @@
...
 import * as Web3 from 'web3';
 
-export const web3 = web3Factory.create();
-export const web3Wrapper = new Web3Wrapper(web3.currentProvider);
+import { constants } from './constants';
+
+const web3 = web3Factory.create({ shouldUseInProcessGanache: true });
+const web3Wrapper = new Web3Wrapper(web3.currentProvider);
+
+export { web3, web3Wrapper };
```
dekz: I think exporting provider here helps reduce Web3 imports for `web3.currentProvider`, like in `packages/deployer/test/util/provider.ts`

packages/0x.js/package.json
fabioberger: So here is the issue:

We need a canonical 0x protocol smart contract migration. This migration will be used to:
1. Deploy 0x protocol on new blockchains (mainnt, testnets, cosmos?)
2. To setup Ganache with the 0x smart contracts for testing purposes (0x.js tests need this)

Before this PR, the migrations were in `contracts` which satisfied 1, but not 2. Since we now use in-process ganache, we cannot simply call the migrations via an NPM script. We need access to the migrations source code.

Maybe the only viable solution here is to have the migrations live in their own package. Deployments to blockchains can happen from that package, and it can also be included into 0x.js. What do you think?

packages/assert/package.json
fabioberger: Oh really? So we can just stop it in a global after hook?

packages/0x.js/test/artifacts/AccountLevels.json
fabioberger: Because they are needed to deploy the contracts before running the 0x.js tests.

packages/contracts/migrations/index.ts
fabioberger: Yep agreed. Gonna move migrations into a separate package since it's used by 0x.js tests and independently when deploying to mainnet/testnets.

packages/0x.js/package.json
fabioberger: Ended up moving migrations into separate package.

packages/migrations/src/index.ts
```diff
@@ -0,0 +1,3 @@
+import { runMigrationsAsync } from './migration';
```
LogvinovLeon: `export { runMigrationsAsync } from './migration';`

packages/migrations/src/utils/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const constants = {
+    NULL_BYTES: '0x',
```
LogvinovLeon: Don't we have that in 3 other places?

packages/sol-cov/src/coverage_subprovider.ts
```diff
@@ -56,8 +56,8 @@ export class CoverageSubprovider extends Subprovider {
      * @param next Callback to call if this subprovider decides not to handle the request
      * @param end Callback to call if subprovider handled the request and wants to pass back the request.
      */
-    // tslint:disable-next-line:prefer-function-over-method
-    public handleRequest(payload: JSONRPCRequestPayload, next: NextCallback, end: ErrorCallback) {
+    // tslint:disable-next-line:prefer-function-over-method async-suffix
+    public async handleRequest(payload: JSONRPCRequestPayload, next: NextCallback, end: ErrorCallback) {
```
LogvinovLeon: 👍

packages/subproviders/test/integration/ledger_subprovider_test.ts
```diff
@@ -1,7 +1,7 @@
 import { JSONRPCResponsePayload } from '@0xproject/types';
 import { promisify } from '@0xproject/utils';
 import Eth from '@ledgerhq/hw-app-eth';
-// HACK: This depdency is optional and tslint skips optional depdencies
+// HACK: This dependency is optional and tslint skips optional dependencies
```
LogvinovLeon: Shouldn't we check that optional dependency if loaded? We can't just use it unconditionally

packages/0x.js/package.json
```diff
@@ -114,6 +114,9 @@
         "uuid": "^3.1.0",
         "web3": "^0.20.0"
     },
+    "optionalDependencies": {
```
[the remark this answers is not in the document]
fabioberger: Because `@0xproject/migrations` is a private package and it doesn't make sense to publish it.

packages/subproviders/test/integration/ledger_subprovider_test.ts
dekz: We could check it and throw though the tests will fail if it isn't found anyway?

packages/0x.js/test/exchange_wrapper_test.ts
```diff
@@ -977,7 +976,7 @@ describe('ExchangeWrapper', () => {
                 );
                 zeroEx.exchange.subscribe(ExchangeEvents.LogFill, indexFilterValues, callbackNeverToBeCalled);
 
-                const newProvider = web3Factory.getRpcProvider();
+                const newProvider = provider;
```
dekz: We can remove this assignment and just use provider.

packages/0x.js/test/token_wrapper_test.ts
```diff
@@ -424,7 +425,7 @@ describe('TokenWrapper', () => {
                 );
                 zeroEx.token.subscribe(tokenAddress, TokenEvents.Transfer, indexFilterValues, callbackNeverToBeCalled);
                 const callbackToBeCalled = reportNodeCallbackErrors(done)();
-                const newProvider = web3Factory.getRpcProvider();
+                const newProvider = provider;
```
dekz: Can remove this assignment

packages/0x.js/test/token_wrapper_test.ts
```diff
@@ -515,3 +516,14 @@ describe('TokenWrapper', () => {
     });
 });
 // tslint:disable:max-file-line-count
+
+function addEmptyWalletSubprovider(provider: Provider): Provider {
+    const providerEngine = new Web3ProviderEngine();
+    providerEngine.addProvider(new EmptyWalletSubprovider());
+    const currentSubproviders = (provider as any)._providers;
```
dekz: Hmm this seems a bit hacky, could we export more out of the `utils/web3_wrapper` to avoid this? Like a `export transportProviderEngineProvider`, bit long in the tooth but it might help avoid an `any` and accessing `_providers`.

Do you think this test is executing some logic that would otherwise be skipped if there was a non-empty account provider? If not it feels like a "just make sure nothing blows up" test case. This would also be caught in all of the tests against testnets where there are no accounts?

packages/dev-utils/test/blockchain_lifecycle_test.ts
```diff
@@ -8,8 +8,8 @@ import { BlockchainLifecycle, web3Factory } from '../src';
 const expect = chai.expect;
 
 describe('BlockchainLifecycle tests', () => {
-    const web3Provider = web3Factory.getRpcProvider();
-    const web3Wrapper = new Web3Wrapper(web3Provider);
+    const provider = web3Factory.getRpcProvider({ shouldUseInProcessGanache: true });
```
dekz: Seems like we have `shouldUseInProcessGanache: true` spread across 4 different packages, is `dev-utils` the rightful place for this to live? Looks like all other projects use the web3Factory from dev-utils, might as well have one place for the Ganache-ness to live.

packages/migrations/package.json
```diff
@@ -0,0 +1,36 @@
...
+        "build": "tsc",
+        "clean": "shx rm -rf lib",
+        "lint": "tslint --project . 'src/**/*.ts'",
+        "migrate": "run-s build compile script:migrate",
+        "script:migrate": "node ./lib/migrate.js",
+        "copy_artifacts": "copyfiles 'src/artifacts/**/*' ./lib",
+        "compile": "node ../deployer/lib/src/cli.js compile --contracts ${npm_package_config_contracts} --contracts-dir ../contracts/src/contracts --artifacts-dir src/artifacts"
+    },
+    "config": {
+        "contracts": "Exchange,DummyToken,ZRXToken,Token,WETH9,TokenTransferProxy,MultiSigWallet,MultiSigWalletWithTimeLock,MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress,MaliciousToken,TokenRegistry,Arbitrage,EtherDelta,AccountLevels"
```
dekz: Hmm a developer needs to keep this config and the compile command in sync with the contracts repo. It's also in the migration package as an enum. hmm I don't have a solution but I think it's worth pointing out.

packages/migrations/src/migrate.ts
```diff
@@ -0,0 +1,25 @@
...
+import { logUtils } from '@0xproject/utils';
+import * as path from 'path';
+
+import { runMigrationsAsync } from './migration';
+
+(async () => {
+    const deployerOpts = {
+        jsonrpcUrl: 'http://localhost:8545',
+        artifactsDir: path.resolve('src', 'artifacts'),
+        networkId: 50,
```
dekz: This appears to be a testrpc only migrate. When we deploy to kovan/mainnet will we create a seperate migrate file or refactor to provide this with parameters?

packages/0x.js/package.json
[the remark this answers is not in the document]
dekz: Does publishing the migrations package help us with developer on-boarding, such as removing the need to download a snapshot, they can deploy the exchange contracts etc themselves?

Does it allow us to get an entire InProcess TestRPC in the CodeSandbox?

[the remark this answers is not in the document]
fabioberger: It could... Any team could import it and then call it with their own deployer setup (own RPC endpoint, networkId, etc...). Think we could consider doing that at some point I guess. Hadn't thought about external developers using it.

packages/migrations/src/utils/constants.ts
fabioberger: Consolidated it into `@0xproject/utils`

packages/subproviders/test/integration/ledger_subprovider_test.ts
fabioberger: Yeah, tests fail anyways with a helpful error message.

packages/0x.js/test/token_wrapper_test.ts
fabioberger: I agree that this is a bit hacky. We do have a precendent of accessing private methods in tests so it's not out-of-line with our conventions. Think it could be a good candidate for a refactor, but fine for this PR. 

If you feel strongly otherwise, please let me know! Still open to changing my mind if there is a really good reason to.

packages/dev-utils/test/blockchain_lifecycle_test.ts
fabioberger: It's only in one place per package. Think that's fine.

packages/migrations/package.json
fabioberger: Agreed that it's a bit brittle at the moment. Will need to be refactored eventually.

packages/migrations/src/migrate.ts
fabioberger: Yes correct. Other migrations will require this to be further refactored.

PR #501, opened 2018-04-04
Use Provider over web3Wrapper

LogvinovLeon:
## Description

This PR:
* Makes our auto-generated contract wrappers accept `Provider` instead of `Web3Wrapper`.
* Moves the Provider type from `Web3.Provider` to `@0xproject/types:Provider`
* Moves even more types from globals to our type registry.
* Tries to use `Provider` instead of `web3` when `web3` is not neccessary as it's a simpler thing.
* Changes deployer config key from `web3Provider` to `provider` to be consistent with other tools.
* Adds some functions to the `web3Wrapper` and adds some basic tests for `web3Wrapper`

## Motivation and Context

Some of the feedback I've gotten when trying to migrate other teams to our tooling is that they're too coupled together and it's impossible to use our templates without using web3Wrapper. This PR solves that.

## How Has This Been Tested?

All tests pass and some new tests added.

## Types of changes

## Checklist:


packages/0x.js/src/0x.ts
```diff
@@ -131,13 +129,7 @@ export class ZeroEx {
...
         assert.isBigNumber('amount', amount);
         assert.isNumber('decimals', decimals);
-
-        const unit = new BigNumber(10).pow(decimals);
-        const baseUnitAmount = amount.times(unit);
-        const hasDecimals = baseUnitAmount.decimalPlaces() !== 0;
-        if (hasDecimals) {
-            throw new Error(`Invalid unit amount: ${amount.toString()} - Too many decimal places`);
-        }
+        const baseUnitAmount = Web3Wrapper.toBaseUnitAmount(amount, decimals);
```
fabioberger: 🕺

packages/0x.js/src/globals.d.ts
```diff
@@ -1,10 +1,3 @@
-declare module 'web3_beta';
```
fabioberger: 🙏

packages/0x.js/src/index.ts
```diff
@@ -32,6 +31,7 @@ export {
     ContractEventArg,
     LogWithDecodedArgs,
     Order,
+    Provider,
```
fabioberger: Don't forget to add `Provider` to the public types list in `zero_ex_js_documentation.ts`

packages/base-contract/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "0.2.0",
+        "changes": [
+            {
+                "note": "Contract wrappers now accept Provider and defaults instead of our Web3Wrapper",
```
fabioberger: `our Web3Wrapper` -> `Web3Wrapper`

packages/base-contract/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
...
+                "pr": 501
+            }
+        ],
+        "timestamp": 1522673609
```
fabioberger: Omit timestamp. Don't think this would cause anything to break, it just gets overwritten at publish time.

packages/deployer/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.4.0",
+        "changes": [
+            {
+                "note": "Changed the config key from `web3Provider` to `provider` to be consistent with other tools",
```
fabioberger: `config key from` -> `config key`

packages/monorepo-scripts/src/find_unused_dependencies.ts
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import * as depcheck from 'depcheck';
+import * as depcheckAsync from 'depcheck';
```
fabioberger: Is it really already Promisified?!

packages/typescript-typings/types/lerna-get-packages/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+        version: string;
+        name: string;
+        main?: string;
+        config?: {
+            additionalTsTypings?: string[];
+        };
+    };
+}
+declare function lernaGetPackages(path: string): LernaPackage[];
+// lerna-get-packages declarations
```
fabioberger: Remove this comment. Superfluous now.

packages/typescript-typings/types/web3_beta/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'web3_beta';
```
fabioberger: No one uses this anymore. Remove.

packages/web3-wrapper/package.json
```diff
@@ -41,6 +46,13 @@
         "copyfiles": "^1.2.0",
         "npm-run-all": "^4.1.2",
         "shx": "^0.2.2",
+        "chai": "^4.0.1",
+        "mocha": "^4.0.1",
+        "nyc": "^11.0.1",
+        "ganache-core": "0xProject/ganache-core",
```
fabioberger: Can we use the `ganacheSubprovider` and import `@0xproject/subproviders` rather then using `ganache-core` directly?

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -96,15 +137,15 @@ export class Web3Wrapper {
      * @returns Ethereum node's version string
      */
     public async getNodeVersionAsync(): Promise<string> {
-        const nodeVersion = await promisify<string>(this._web3.version.getNode)();
+        const nodeVersion = await this._sendRawPayloadAsync<string>({ method: 'web3_clientVersion' });
```
fabioberger: 😮 🍾 💃

packages/web3-wrapper/package.json
LogvinovLeon: We can't use the web3Factory because that will introducce a cycle in dependencies graph.
If we use Ganache subprovider then we need to create a provider engine, add a subprovider, start it. That doesn't make sense. Here I just get a provider and I'm good to go.

PR #507, opened 2018-04-06
Mnemonic wallet subprovider and multiple accounts in ledger

dekz:
## Description

Add a Mnemonic wallet suprovider to support users who use mnemonics.

Support multiple accounts (determined from `from` fields) in mnemonic and Ledger subproviders.

## Motivation and Context

We have a fork of the [truffle-hdwallet-provider](https://github.com/0xProject/truffle-hdwallet-provider). We had to add personal sign support to this as the repository hasn't been touched in over a year. Rather than maintain this outside of subproviders, let's just have a subprovider for it.

This PR adds it as a subprovider by leveraging the PrivateKeyWalletSubprovider.

Our ledger subprovider doesn't currently support multiple accounts being requested from the same Subprovider, we use an index to determine which account can sign, and this index must be updated out of band. This can result in the wrong account signing the data when a different `from` field is supplied.

## How Has This Been Tested?

Unit tests, ledger integration tests
[Manual tests in portal with ledger](https://kovan.etherscan.io/tx/0x8aac89484575241fe19f48c6644f5a55938b582675d4d11f3413a81263fda80b)

## Types of changes

## Checklist:

### TODO
- [x] iterator pattern for searching through HDKeys
- [x] Slurp up IS_CHILD and path and initial hd key into an object
- [x] fix address search limit in ledger to be 1000
- [x] Update website


packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+     * If you've added this Subprovider to your  app's provider, you can simply send
+     * an `eth_sendTransaction` JSON RPC request, and * this method will be called auto-magically.
+     * If you are not using this via a ProviderEngine instance, you can call it directly.
+     * @param txParams Parameters of the transaction to sign
+     * @return Signed transaction hex string
+     */
+    public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
+        const accounts = await this.getAccountsAsync();
+        const hdKey = this._findHDKeyByPublicAddress(txParams.from || accounts[0]);
+        const privateKeyWallet = new PrivateKeyWalletSubprovider(hdKey.privateKey.toString('hex'));
```
BMillman19: hm this seems a bit weird to me, should this class be a subclass of `PrivateKeyWalletSubprovider` instead?

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+ * This class implements the [web3-provider-engine](https://github.com/MetaMask/provider-engine) subprovider interface.
+ * This subprovider intercepts all account related RPC requests (e.g message/transaction signing, etc...) and handles
+ * all requests with accounts derived from the supplied mnemonic.
+ */
+export class MnemonicWalletSubprovider extends BaseWalletSubprovider {
+    private _derivationPath: string;
+    private _hdKey: HDNode;
+    private _derivationPathIndex: number;
+    constructor(mnemonic: string, derivationPath: string = DEFAULT_DERIVATION_PATH) {
+        assert.isString('mnemonic', mnemonic);
```
fabioberger: Add assertion for `derivationPath`

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+        super();
+        this._hdKey = HDNode.fromMasterSeed(bip39.mnemonicToSeed(mnemonic));
```
fabioberger:
```
const seed = bip39.mnemonicToSeed(mnemonic);
this._hdKey = HDNode.fromMasterSeed(seed);
```

for clarity.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * This class implements the [web3-provider-engine](https://github.com/MetaMask/provider-engine) subprovider interface.
+ * This subprovider intercepts all account related RPC requests (e.g message/transaction signing, etc...) and handles
+ * all requests with accounts derived from the supplied mnemonic.
+ */
+export class MnemonicWalletSubprovider extends BaseWalletSubprovider {
+    private _derivationPath: string;
+    private _hdKey: HDNode;
+    private _derivationPathIndex: number;
+    constructor(mnemonic: string, derivationPath: string = DEFAULT_DERIVATION_PATH) {
```
fabioberger: Add docComment above the constructor, mention that `derivationPath` defaults to `44'/60'/0'`

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+     * Set the final derivation path index. If a user wishes to sign a message with the
+     * 6th address in a derivation path, before calling `signPersonalMessageAsync`, you must
+     * call this method with pathIndex `6`.
+     * @param pathIndex Desired derivation path index
+     */
+    public setPathIndex(pathIndex: number) {
+        this._derivationPathIndex = pathIndex;
+    }
+    /**
+     * Retrieve the account associated with the supplied private key.
```
fabioberger: This comment was copy-pasted. please fix.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+     * This method is implicitly called when issuing a `eth_accounts` JSON RPC request
+     * via your providerEngine instance.
+     * @return An array of accounts
```
fabioberger: Add comment above what the default number of accounts is set to.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
fabioberger: Agreed that this is a bit strange.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+     * Signs a transaction with the from account (if specificed in txParams) or the first account.
+     * If you've added this Subprovider to your  app's provider, you can simply send
+     * an `eth_sendTransaction` JSON RPC request, and * this method will be called auto-magically.
+     * If you are not using this via a ProviderEngine instance, you can call it directly.
+     * @param txParams Parameters of the transaction to sign
+     * @return Signed transaction hex string
+     */
+    public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
+        const accounts = await this.getAccountsAsync();
+        const hdKey = this._findHDKeyByPublicAddress(txParams.from || accounts[0]);
```
fabioberger: Can't we just get this from the set `pathIndex` and `derivationPath`. We should probably just assert that `txParams.from` should be equal to the address pointed to from the `derivationPath` + `pathIndex`.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
```diff
@@ -0,0 +1,123 @@
...
+     * derived from the set path.
+     * If you've added the PKWalletSubprovider to your app's provider, you can simply send an `eth_sign`
+     * or `personal_sign` JSON RPC request, and this method will be called auto-magically.
+     * If you are not using this via a ProviderEngine instance, you can call it directly.
+     * @param data Message to sign
+     * @return Signature hex string (order: rsv)
+     */
+    public async signPersonalMessageAsync(data: string): Promise<string> {
+        const accounts = await this.getAccountsAsync();
+        const hdKey = this._findHDKeyByPublicAddress(accounts[0]);
```
fabioberger: Ditto above. What is the point of setting the `pathIndex` if we never use it.

packages/subproviders/src/subproviders/mnemonic_wallet_subprovider.ts
dekz: Hmm I always viewed that index as an offset into the set of available addresses.

From a perspective of automated trading, I can see a use case of using multiple accounts at the same time, derived from the mnemonic seed, and the `from` in `txParams` indicates the account to use (remember this can be through provider engine, which supports multiple accounts). Rather than the user having to set the index then make request to `sendTransaction`. 

Requiring them to match might make sense in a dApp where the user has 'selected a single account', but if this was to be used in our tests (with multiple accounts off the mnemonic) then it limits the overall functionality of the Mnemonic Wallet Subprovider.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -89,35 +84,12 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
-            const derivedPublicKey = derivedHDNode.publicKey;
-            const shouldSanitizePublicKey = true;
-            const ethereumAddressUnprefixed = ethUtil
-                .publicToAddress(derivedPublicKey, shouldSanitizePublicKey)
-                .toString('hex');
-            const ethereumAddressPrefixed = ethUtil.addHexPrefix(ethereumAddressUnprefixed);
-            accounts.push(ethereumAddressPrefixed.toLowerCase());
-        }
+    public async getAccountsAsync(
+        numberOfAccounts: number = walletUtils.DEFAULT_NUM_ADDRESSES_TO_FETCH,
```
fabioberger: This is leaking `walletUtils` into the public, exported interface. We don't **need** `DEFAULT_NUM_ADDRESSES_TO_FETCH` to be in `walletUtils`. It could be set to a different amount between the MnemonicWalletSubprovider and the LedgerSubprovider. Let's declare these in their respective packages. This is important for how it renders in the docs.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -89,35 +84,12 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
+    ): Promise<string[]> {
+        const initialDerivedKey = await this._initialDerivedKeyAsync();
+        const derivedKeys = walletUtils.calculateDerivedHDKeys(initialDerivedKey, numberOfAccounts);
+        const accounts = _.map(derivedKeys, 'address');
```
fabioberger: Our convention is to never use the string form of calling `_.map`. Use: `_.map(derivedKeys, derivedKey => derivedKey.address)` sometimes we shorthand to single character variables in this case: `_.map(derivedKeys, k => k.address)`

fabioberger: This way we preserve type safety, which the string form doesn't.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -130,7 +102,10 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
      */
     public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
         LedgerSubprovider._validateTxParams(txParams);
-        this._ledgerClientIfExists = await this._createLedgerClientAsync();
+        const initialDerivedKey = await this._initialDerivedKeyAsync();
```
fabioberger: Why not set the initialDerivedKey in the constructor? We only need to compute this at instantiation, and if the `baseDerivationPath` is updated, no?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -165,27 +140,28 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
         }
     }
     /**
-     * Sign a personal Ethereum signed message. The signing address will be the one
-     * retrieved given a derivationPath and pathIndex set on the subprovider.
+     * Sign a personal Ethereum signed message. The signing account will be the account
+     * associated with the provided address.
```
fabioberger: Think it's easy to confuse devs with the address/accounts naming confusion. Let's try to stick to one or the other. Here it might make sense to pass in `account: string`, and always refer to it as `account`. Thoughts?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -211,6 +183,7 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
         }
         const ledgerEthereumClient = await this._ledgerEthereumClientFactoryAsync();
         this._connectionLock.release();
+        this._ledgerClientIfExists = ledgerEthereumClient;
```
fabioberger: Why was this added? Was this missing, i.e a bug?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -223,4 +196,39 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
+        const hdKey = new HDNode();
+        hdKey.publicKey = new Buffer(ledgerResponse.publicKey, 'hex');
+        hdKey.chainCode = new Buffer(ledgerResponse.chainCode, 'hex');
+        return {
+            hdKey,
+            address: ledgerResponse.address,
+            isChildKey: true,
+            derivationBasePath: this._derivationBasePath,
+            derivationPath: `${this._derivationBasePath}/0`,
+            derivationIndex: 0,
```
fabioberger: Reference a single constant set to `0` in both the `derivationPath` and `derivationIndex`.

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -223,4 +196,39 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
+        };
+    }
+    private _findDerivedKeyByPublicAddress(initalHDKey: DerivedHDKey, address: string): DerivedHDKey {
```
fabioberger: Why `PublicAddress` and not simply `address`?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -223,4 +196,39 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
+        if (_.isUndefined(address) || !addressUtils.isAddress(address)) {
+            throw new Error(WalletSubproviderErrors.FromAddressMissingOrInvalid);
```
fabioberger: This error doesn't seem to match up with this helper function... This helper function could be used in many places, and should throw an internal error which is handled by the caller and possibly re-thrown if the `address` fed in was the `from` address of a request. 

Does this make sense?

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -223,4 +196,39 @@ export class LedgerSubprovider extends BaseWalletSubprovider {
...
+        }
+        const matchedDerivedKey = walletUtils.findDerivedKeyByAddress(address, initalHDKey, this._addressSearchLimit);
```
fabioberger: Should be renamed `findDerivedKeyForAddressIfExists`. `By` makes it sound like a mapping, and since it might not be found and returns undefined, this needs to be elucidated by the function name.

Ditto for the variable name here.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+ * This subprovider intercepts all account related RPC requests (e.g message/transaction signing, etc...) and handles
+ * all requests with accounts derived from the supplied mnemonic.
+ */
+export class MnemonicWalletSubprovider extends BaseWalletSubprovider {
+    private _addressSearchLimit: number;
+    private _derivationBasePath: string;
+    private _derivedKey: DerivedHDKey;
+    /**
+     * Instantiates a MnemonicWalletSubprovider. Defaults to derivationPath set to `44'/60'/0'/0`.
+     * This is the default in TestRPC/Ganache, this can be overridden if desired.
```
fabioberger: `it can be...`

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+     * @param mnemonic The mnemonic seed
+     * @param derivationBasePath The derivation path, defaults to `44'/60'/0'/0`
+     * @param addressSearchLimit The limit on address search attempts before raising `WalletSubproviderErrors.AddressNotFound`
+     * @return MnemonicWalletSubprovider instance
+     */
+    constructor(
+        mnemonic: string,
+        derivationBasePath: string = DEFAULT_DERIVATION_PATH,
+        addressSearchLimit: number = walletUtils.DEFAULT_ADDRESS_SEARCH_LIMIT,
```
fabioberger: The two above should be nested within a config object. Otherwise, I **must** set the first in order to set the second. This is cumbersome.

Also make `DEFAULT_ADDRESS_SEARCH_LIMIT` local to this file.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+    ) {
+        assert.isString('mnemonic', mnemonic);
+        assert.isString('derivationPath', derivationBasePath);
+        assert.isNumber('addressSearchLimit', addressSearchLimit);
+        super();
+        const seed = bip39.mnemonicToSeed(mnemonic);
+        const hdKey = HDNode.fromMasterSeed(seed);
+        this._derivationBasePath = derivationBasePath;
```
fabioberger: `_baseDerivationPath`

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+        this._derivedKey = {
+            address: walletUtils.addressOfHDKey(hdKey),
+            // HACK this isn't the base path for this root key, but is is the base path
+            // we want all of the derived children to spawn from
+            derivationBasePath: this._derivationBasePath,
```
fabioberger: Can you elaborate on this hack? Not sure I follow.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+     * This method is implicitly called when issuing a `eth_accounts` JSON RPC request
+     * via your providerEngine instance.
+     * @param numberOfAccounts Number of accounts to retrieve (default: 10)
+     * @return An array of accounts
+     */
+    public async getAccountsAsync(
+        numberOfAccounts: number = walletUtils.DEFAULT_NUM_ADDRESSES_TO_FETCH,
+    ): Promise<string[]> {
+        const derivedKeys = walletUtils.calculateDerivedHDKeys(this._derivedKey, numberOfAccounts);
+        const accounts = _.map(derivedKeys, 'address');
```
fabioberger: Ditto comment above.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    private _privateKeyWalletForAddress(address: string): PrivateKeyWalletSubprovider {
+        const derivedKey = this._findDerivedKeyByPublicAddress(address);
+        const privateKeyHex = derivedKey.hdKey.privateKey.toString('hex');
+        const privateKeyWallet = new PrivateKeyWalletSubprovider(privateKeyHex);
+        return privateKeyWallet;
+    }
+    private _findDerivedKeyByPublicAddress(address: string): DerivedHDKey {
+        if (_.isUndefined(address) || !addressUtils.isAddress(address)) {
+            throw new Error(WalletSubproviderErrors.FromAddressMissingOrInvalid);
```
fabioberger: This error is thrown at the wrong level. Needs to live at the function that receives the `from` address.

packages/subproviders/src/subproviders/private_key_wallet.ts
```diff
@@ -17,6 +15,11 @@ import { Subprovider } from './subprovider';
 export class PrivateKeyWalletSubprovider extends BaseWalletSubprovider {
     private _address: string;
     private _privateKeyBuffer: Buffer;
+    /**
+     * Instantiates a PrivateKeyWalletSubprovider.
+     * @param privateKey The private key of the ethereum address
```
fabioberger: `The corresponding private key to an Ethereum address`

packages/subproviders/src/subproviders/private_key_wallet.ts
```diff
@@ -42,26 +45,33 @@ export class PrivateKeyWalletSubprovider extends BaseWalletSubprovider {
      */
     public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
         PrivateKeyWalletSubprovider._validateTxParams(txParams);
+        if (!_.isUndefined(txParams.from) && txParams.from !== this._address) {
+            throw new Error(`${WalletSubproviderErrors.AddressNotFound}: ${txParams.from}`);
```
fabioberger: I think we need to throw a better error message here. Perhaps: 

```
Requested to sign transaction with account: ${txParams.from} using PrivateKeyWalletSubprovider instantiated with account: ${this._address}
```

packages/subproviders/src/subproviders/private_key_wallet.ts
```diff
@@ -42,26 +45,33 @@ export class PrivateKeyWalletSubprovider extends BaseWalletSubprovider {
...
         const tx = new EthereumTx(txParams);
         tx.sign(this._privateKeyBuffer);
         const rawTx = `0x${tx.serialize().toString('hex')}`;
         return rawTx;
     }
     /**
      * Sign a personal Ethereum signed message. The signing address will be
-     * calculated from the private key.
-     * If you've added the PKWalletSubprovider to your app's provider, you can simply send an `eth_sign`
+     * calculated from the private key, if an address is provided it must match the address calculated from the private key.
```
fabioberger: Nit: can add some more of these words to the previous line.

packages/subproviders/src/subproviders/private_key_wallet.ts
```diff
@@ -42,26 +45,33 @@ export class PrivateKeyWalletSubprovider extends BaseWalletSubprovider {
...
-    public async signPersonalMessageAsync(dataIfExists: string): Promise<string> {
-        if (_.isUndefined(dataIfExists)) {
+    public async signPersonalMessageAsync(data: string, address: string): Promise<string> {
+        if (_.isUndefined(data)) {
             throw new Error(WalletSubproviderErrors.DataMissingForSignPersonalMessage);
         }
-        assert.isHexString('data', dataIfExists);
-        const dataBuff = ethUtil.toBuffer(dataIfExists);
+        if (_.isUndefined(address) || address !== this._address) {
+            throw new Error(`${WalletSubproviderErrors.FromAddressMissingOrInvalid}: ${address}`);
```
fabioberger: Ditto above.

packages/subproviders/src/types.ts
```diff
@@ -96,19 +92,28 @@ export interface ResponseWithTxParams {
...
     CannotDetermineAddressFromPayload = 'CANNOT_DETERMINE_ADDRESS_FROM_PAYLOAD',
 }
+export interface DerivedHDKey {
+    address: string;
+    derivationIndex: number;
+    derivationBasePath: string;
+    derivationPath: string;
+    hdKey: HDNode;
+    isChildKey: boolean;
+}
```
fabioberger: This is more like: `DerivedHdKeyInfo` no? It's not the key itself, it's a bunch of info about the HDKey.

packages/subproviders/src/utils/wallet_utils.ts
```diff
@@ -0,0 +1,88 @@
+import ethUtil = require('ethereumjs-util');
+import HDNode = require('hdkey');
+import * as _ from 'lodash';
+
+import { DerivedHDKey, WalletSubproviderErrors } from '../types';
+
+const DEFAULT_ADDRESS_SEARCH_OFFSET = 0;
+const BATCH_SIZE = 10;
```
fabioberger: Unused constant. Does our lint rules not catch this?

packages/subproviders/src/utils/wallet_utils.ts
```diff
@@ -0,0 +1,88 @@
...
+const DEFAULT_ADDRESS_SEARCH_LIMIT = 1000;
+
+class DerivedHDKeyIterator implements IterableIterator<DerivedHDKey> {
+    private _initialDerivedKey: DerivedHDKey;
+    private _searchLimit: number;
+    private _index: number;
+
+    constructor(initialDerivedKey: DerivedHDKey, searchLimit: number = DEFAULT_ADDRESS_SEARCH_OFFSET) {
```
fabioberger: Why is `searchLimit` defaulted to `DEFAULT_ADDRESS_SEARCH_OFFSET`? Aren't those two completely different things?

fabioberger: Did you mean `DEFAULT_ADDRESS_SEARCH_LIMIT`?

packages/subproviders/src/types.ts
```diff
@@ -50,6 +51,7 @@ export interface LedgerSubproviderConfigs {
  *                                   before fetching their addresses
  */
 export interface AccountFetchingConfigs {
+    addressSearchLimit?: number;
```
fabioberger: Add a comment above for this new config setting. Method the default.

packages/subproviders/src/utils/wallet_utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    DEFAULT_NUM_ADDRESSES_TO_FETCH: 10,
+    calculateDerivedHDKeys(initialDerivedKey: DerivedHDKey, numberOfKeys: number): DerivedHDKey[] {
+        const derivedKeys: DerivedHDKey[] = [];
+        const derivedKeyIterator = new DerivedHDKeyIterator(initialDerivedKey, numberOfKeys);
+        for (const key of derivedKeyIterator) {
+            derivedKeys.push(key);
+        }
+        return derivedKeys;
+    },
+    findDerivedKeyByAddress(
```
fabioberger: `findDerivedKeyByAddressIfExists`

packages/subproviders/src/utils/wallet_utils.ts
```diff
@@ -0,0 +1,88 @@
...
+        const derivedKeyIterator = new DerivedHDKeyIterator(initialDerivedKey, searchLimit);
+        for (const key of derivedKeyIterator) {
+            if (key.address === address) {
+                matchedKey = key;
+                break;
+            }
+        }
+        return matchedKey;
+    },
+    addressOfHDKey(hdKey: HDNode): string {
```
fabioberger: Let's try and remain consistent between how we capitalize `hdkey`. It should either always be: `HDKey` or `hdKey`/`HdKey`.

packages/subproviders/src/utils/wallet_utils.ts
```diff
@@ -0,0 +1,88 @@
...
+        const shouldSanitizePublicKey = true;
+        const derivedPublicKey = hdKey.publicKey;
+        const ethereumAddressUnprefixed = ethUtil
+            .publicToAddress(derivedPublicKey, shouldSanitizePublicKey)
+            .toString('hex');
+        const address = ethUtil.addHexPrefix(ethereumAddressUnprefixed);
+        return address;
```
fabioberger: Do we need to lowercase?

packages/subproviders/test/unit/private_key_wallet_subprovider_test.ts
```diff
@@ -128,6 +128,23 @@ describe('PrivateKeyWalletSubprovider', () => {
                 });
                 provider.sendAsync(payload, callback);
             });
+            it('should throw if `address` param is not an address from private key when calling personal_sign', (done: DoneCallback) => {
```
fabioberger: `is not the address corresponding to the private key...`

packages/typescript-typings/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.1.1",
+        "changes": [
+            {
+                "note": "Add types for HDKey",
+                "pr": 507
+            }
+        ]
+    },
```
fabioberger: 🙏

packages/website/ts/components/dialogs/ledger_config_dialog.tsx
```diff
@@ -199,7 +199,6 @@ export class LedgerConfigDialog extends React.Component<LedgerConfigDialogProps,
     }
     private _onAddressSelected(selectedRowIndexes: number[]) {
         const selectedRowIndex = selectedRowIndexes[0];
-        this.props.blockchain.updateLedgerDerivationIndex(selectedRowIndex);
```
fabioberger: Did you manually test signing an order with a given Ledger address on website? Would be good to manually check that it's setting the correct address as the `from` parameter in the requests.

dekz: I manually [tested allowance changes](https://kovan.etherscan.io/tx/0x8aac89484575241fe19f48c6644f5a55938b582675d4d11f3413a81263fda80b) in portal with the 4th address on the ledger which should cover `sendTransaction`. But I'll quickly test `signPersonalMessage` through signing an order just incase.

packages/subproviders/src/subproviders/ledger.ts
dekz: I think I prefer it to be lazy, if we do it in `setPath` then it becomes async and could potentially throw (from a ledger error). Whereas this will throw when the user is trying to do an action and the developer is ready for any errors.

packages/subproviders/src/subproviders/ledger.ts
dekz: When I think account I picture a publicKey privateKey combination. Where as address is just a string encoded public key. It doesn't help that the JSON RPC documentation also uses both :P 

```
Parameters
account, message

DATA, 20 Bytes - address
DATA, N Bytes - message to sign
```

packages/subproviders/src/subproviders/ledger.ts
dekz: It could've been a bug if a developer asked for a lock and forgot to assign it or assigned it to a local variable. I moved it in here to make sure this doesn't happen in the future.

packages/subproviders/src/subproviders/private_key_wallet.ts
dekz: Sure, since this code is re-used by the Mnemonic wallet as well I think we should use the message you provided without mention of PrivateKeyWalletSubprovider.

packages/subproviders/src/utils/wallet_utils.ts
dekz: hmm previously when we lowercased this we called it:
```
const hdKey = new HDNode();
```
hdKey for camelCase parameters
HDNode/HDKey for types.

packages/subproviders/src/subproviders/ledger.ts
fabioberger: Haha, yeah it doesn't. We need to be better then JSON RPC 😛

packages/subproviders/src/subproviders/ledger.ts
dekz: The problem with only assigning it in that method is that ts forces us to check if it is undefined, even when that method throws.

```typescript
await this._createLedgerClientAsync();
await this._createLedgerClientAsync();
        if (_.isUndefined(this._ledgerClientIfExists)) {
            throw new Error('aaa');
        }

        let ledgerResponse;
        try {
            ledgerResponse = await this._ledgerClientIfExists.getAddress(
...
```
versus
```typescript
        const ledgerClient = await this._createLedgerClientAsync();

        let ledgerResponse;
        try {
            ledgerResponse = await ledgerClient.getAddress(
```

Thoughts?

packages/subproviders/CHANGELOG.json
```diff
@@ -10,6 +10,15 @@
...
                 "note": "Add private key subprovider and refactor shared functionality into a base wallet subprovider",
                 "pr": 506
+            },
+            {
+                "note": "Add mnemonic wallet subprovider, deprecating our truffle-hdwallet-provider fork.",
+                "pr": 507
+            },
+            {
+                "note":
+                    "Support multiple addresses in ledger and mnemonic wallets. Renamed derivationPath to baseDerivationPath.",
```
fabioberger: Make these separate entries ;)

Can we also add one to the effect of:
```
Refactors LedgerSubprovider such that explicitly setting the `pathIndex` is no longer required. Simply set the request `from` address as desired.
```

packages/subproviders/src/subproviders/ledger.ts
```diff
@@ -17,13 +17,16 @@ import {
...
 import { BaseWalletSubprovider } from './base_wallet_subprovider';
 
-const DEFAULT_DERIVATION_PATH = `44'/60'/0'`;
-const DEFAULT_NUM_ADDRESSES_TO_FETCH = 10;
+const DEFAULT_BASE_DERIVATION_PATH = `44'/60'/0'`;
 const ASK_FOR_ON_DEVICE_CONFIRMATION = false;
 const SHOULD_GET_CHAIN_CODE = true;
+const DEFAULT_NUM_ADDRESSES_TO_FETCH = 10;
+const DEFAULT_ADDRESS_SEARCH_LIMIT = 1000;
+const INITIAL_KEY_DERIVATION_INDEX = 0;
```
fabioberger: Remove.

packages/subproviders/src/subproviders/ledger.ts
fabioberger: Ok, then let's remove this assignment. It's not clean to have both. I do think this is a candidate for a future refactor.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,147 @@
...
+import { DerivedHDKeyInfo, MnemonicWalletSubproviderConfigs, PartialTxParams, WalletSubproviderErrors } from '../types';
+import { walletUtils } from '../utils/wallet_utils';
+
+import { BaseWalletSubprovider } from './base_wallet_subprovider';
+import { PrivateKeyWalletSubprovider } from './private_key_wallet';
+
+const DEFAULT_BASE_DERIVATION_PATH = `44'/60'/0'/0`;
+const DEFAULT_NUM_ADDRESSES_TO_FETCH = 10;
+const DEFAULT_ADDRESS_SEARCH_LIMIT = 1000;
+const INITIAL_KEY_DERIVATION_INDEX = 0;
```
fabioberger: Remove.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,147 @@
...
+    private _mnemonic: string;
+
+    /**
+     * Instantiates a MnemonicWalletSubprovider. Defaults to baseDerivationPath set to `44'/60'/0'/0`.
+     * This is the default in TestRPC/Ganache, it can be overridden if desired.
+     * @param config Configuration for the mnemonic wallet, must contain the mnemonic
+     * @return MnemonicWalletSubprovider instance
+     */
+    constructor(config: MnemonicWalletSubproviderConfigs) {
+        const mnemonic = config.mnemonic;
```
fabioberger: Feel this a superfluous assignment.

packages/subproviders/src/subproviders/mnemonic_wallet.ts
```diff
@@ -0,0 +1,147 @@
...
+     * via your providerEngine instance.
+     * @param numberOfAccounts Number of accounts to retrieve (default: 10)
+     * @return An array of accounts
+     */
+    public async getAccountsAsync(numberOfAccounts: number = DEFAULT_NUM_ADDRESSES_TO_FETCH): Promise<string[]> {
+        const derivedKeys = walletUtils.calculateDerivedHDKeyInfos(this._derivedKeyInfo, numberOfAccounts);
+        const accounts = _.map(derivedKeys, k => k.address);
+        return accounts;
+    }
+
```
fabioberger: Remove spaces. Why do these get added in? Thought we had a TSLint rule for these.

fabioberger: Let's add an Airtable task for creating one.

PR #508, opened 2018-04-06
Transaction Watcher initial interface

avonian:
## Description

Looking to build a TransactionWatcher (inspired by OrderWatcher) which could be used as a higher-level API for watching events. This can be especially useful for handling failed transactions (e.g. out of gas).

## Motivation and Context

The current recommended approach for detecting failed transactions is to poll the blockchain for transaction receipts. After discussing the problem with 0x team members, it seems this would be a good use case for a TransactionWatcher (similar to OrderWatcher).

## How Has This Been Tested?

Work in progress.

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/src/order_watcher/transaction_watcher.ts
+++ b/packages/0x.js/src/order_watcher/transaction_watcher.ts
@@ -1,0 +1,96 @@
+import { schemas } from '@0xproject/json-schemas';
+import { BlockParamLiteral, LogWithDecodedArgs, SignedOrder, TransactionReceiptWithDecodedLogs } from '@0xproject/types';
+import { AbiDecoder, intervalUtils } from '@0xproject/utils';
+import { Web3Wrapper } from '@0xproject/web3-wrapper';
+import * as _ from 'lodash';
+
+import { ZeroEx } from '../0x';
+import { ExchangeWrapper } from '../contract_wrappers/exchange_wrapper';
+import {
+    DepositContractEventArgs,
+    EtherTokenEvents,
+    WithdrawalContractEventArgs,
+} from '../contract_wrappers/generated/ether_token';
+import {
+    ExchangeEvents,
+    LogCancelContractEventArgs,
+    LogFillContractEventArgs,
+} from '../contract_wrappers/generated/exchange';
+import {
+    ApprovalContractEventArgs,
+    TokenEvents,
+    TransferContractEventArgs,
+} from '../contract_wrappers/generated/token';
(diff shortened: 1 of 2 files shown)


packages/0x.js/src/order_watcher/transaction_watcher.ts
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public addTransaction(transactionHash: string): void {
+        assert.isHexString('transactionHash', transactionHash);
+        this._transactionHashes.push(transactionHash)
+    }
+    /**
+     * Removes a transaction from the transactionWatcher
+     * @param   transactionHash     The transactionHash of the transaction you wish to stop watching.
+     */
+    public removeTransaction(orderHash: string): void {
```
LogvinovLeon: Do we need that? Transaction can be auto-removed when it's settled or not settled for N blocks.

packages/0x.js/src/types.ts
```diff
@@ -283,4 +285,8 @@ export interface OrderStateInvalid {
 export type OrderState = OrderStateValid | OrderStateInvalid;
 
 export type OnOrderStateChangeCallback = (err: Error | null, orderState?: OrderState) => void;
+
+export type OnTransactionSuccessCallback = (transactionReceipt?: TransactionReceiptWithDecodedLogs) => void;
```
LogvinovLeon: Why is that parameter optional?

packages/0x.js/src/types.ts
```diff
@@ -283,4 +285,8 @@ export interface OrderStateInvalid {
...
+
+export type OnTransactionFailedCallback = (transactionReceipt?: TransactionReceipt) => void;
```
LogvinovLeon: And this one too

avonian:
I wanted to submit something quickly to at least get the ball rolling, look forward to everyone's feedback.

packages/0x.js/src/order_watcher/transaction_watcher.ts
avonian: Good point, will remove.

packages/0x.js/src/types.ts
avonian: Typical mistake copy and pasting, noted for fix.

packages/0x.js/src/types.ts
avonian: Typical mistake copy and pasting, noted for fix.

LogvinovLeon:
Reopen if still valid

PR #509, opened 2018-04-06
Use solc.compileStandardWrapper

LogvinovLeon:
## Description

This PR migrates deployer to use [`solc.compileStandardWrapper`](https://github.com/ethereum/solc-js#from-version-0411) as a first step towards using standard artifacts format.

Before - deployer identified contracts by "fileName". That wasn't the actual fileName, but the base name. So if you had a contracts in `contracts/tokens/Token.sol` all the mappings will have an index `Token.sol`. This PR changes that basic contract id to be a contract name. I think - that's less confusing. Yes, we make an assumption, that all the names of the contracts are unique. We've already made it before. 

It also highlights warnings in yellow and errors in red ;)

## Motivation and Context

This will simplify the migrations towards a new standard artifacts format.

## How Has This Been Tested?

Remigrated contracts and 0x.js tests passed.

## Types of changes

## Checklist:


packages/deployer/src/compiler.ts
```diff
@@ -162,49 +171,69 @@ export class Compiler {
...
 
-        logUtils.log(`Compiling ${fileName} with Solidity v${solcVersion}...`);
-        const source = this._contractSources[fileName];
-        const input = {
-            [fileName]: source,
-        };
-        const sourcesToCompile = {
-            sources: input,
+        logUtils.log(`Compiling ${contractName} with Solidity v${solcVersion}...`);
+        const source = this._contractSources[contractName].source;
```
fabioberger:
```
const contractSource = this._contractSources[contractName];
const source = contractSource.source;
const absoluteFilePath = contractSource.absoluteFilePath;
```

fabioberger: Are we guaranteed there is a corresponding entry in  `this._contractSources`?

packages/deployer/src/compiler.ts
```diff
@@ -162,49 +171,69 @@ export class Compiler {
...
+        const absoluteFilePath = this._contractSources[contractName].absoluteFilePath;
+        const standardInput: solc.StandardInput = {
+            language: 'Solidity',
```
fabioberger: Should we make an enum for this?

packages/deployer/src/compiler.ts
```diff
@@ -162,49 +171,69 @@ export class Compiler {
...
+            sources: {
+                [absoluteFilePath]: {
+                    urls: [`file://${absoluteFilePath}`],
```
fabioberger: Why the `file://` prefix? Also, this isn't windows friendly. Do we want to be?

packages/deployer/src/compiler.ts
```diff
@@ -162,49 +171,69 @@ export class Compiler {
...
+            ),
         );
 
         if (!_.isUndefined(compiled.errors)) {
-            const SOLIDITY_WARNING_PREFIX = 'Warning';
-            const isError = (errorOrWarning: string) => !errorOrWarning.includes(SOLIDITY_WARNING_PREFIX);
-            const isWarning = (errorOrWarning: string) => errorOrWarning.includes(SOLIDITY_WARNING_PREFIX);
+            const SOLIDITY_WARNING = 'warning';
+            const isError = (entry: any) => entry.severity !== SOLIDITY_WARNING;
+            const isWarning = (entry: any) => entry.severity === SOLIDITY_WARNING;
```
fabioberger: Why are these any?

packages/deployer/src/utils/compiler.ts
```diff
@@ -105,16 +104,15 @@ export function parseDependencies(source: string): string[] {
 
 /**
  * Callback to resolve dependencies with `solc.compile`.
- * Throws error if contractSources not yet initialized.
  * @param  contractSources Source codes of contracts.
  * @param  importPath Path to an imported dependency.
  * @return Import contents object containing source code of dependency.
  */
 export function findImportIfExist(contractSources: ContractSources, importPath: string): solc.ImportContents {
```
fabioberger: Can we just pass in `contractSource` instead of `contractSources`?

packages/deployer/src/compiler.ts
```diff
@@ -106,24 +112,27 @@ export class Compiler {
         await createDirIfDoesNotExistAsync(this._artifactsDir);
         await createDirIfDoesNotExistAsync(SOLC_BIN_DIR);
         this._contractSources = await Compiler._getContractSourcesAsync(this._contractsDir);
-        _.forIn(this._contractSources, this._setContractSpecificSourceData.bind(this));
-        const fileNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
+        for (const contractName of _.keys(this._contractSources)) {
+            const contractSource = this._contractSources[contractName];
+            this._setContractSpecificSourceData(contractSource.source, contractName);
+        }
+        const contractNames = this._specifiedContracts.has(ALL_CONTRACTS_IDENTIFIER)
```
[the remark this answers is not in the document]
LogvinovLeon: That's a set operation, hot just the property check

packages/deployer/src/compiler.ts
LogvinovLeon: Yes

packages/deployer/src/compiler.ts
LogvinovLeon: Nah. Let's do that when we have more than one value.

packages/deployer/src/compiler.ts
LogvinovLeon: Because if how solidity accept sources. It should be either a string with contents or the URL. So I pass the file URL and it'll read it from the FS. Making this tool windows-friendly is another PR. I'll do that when I'll create a windows VM.

packages/deployer/src/utils/compiler.ts
LogvinovLeon: This wil be out after the resolver PR

PR #510, opened 2018-04-09
Implement `zeroEx.exchange.getOrderStateAsync`

avonian:
## Description

Implemented `zeroEx.exchange.getOrderStateAsync` which returns the OrderState for a given `signedOrder`

## Motivation and Context

In the event that an OrderStateWatcher suffers an outage, it would be very useful for relayers to be able to update the exact state of orders (including things like remaining fillable amounts) when they bring their watchers back online.

Although 0x.js currently offers methods to check validity of an order, these methods don't provide any additional information. _getOrderStateAsync_ borrows the same mechanism that OrderStateWatcher uses when emitting events and returns the OrderState for a _signedOrder_ from the `latest` state layer.

## How Has This Been Tested?

It has been tested manually and test cases have been added to `exchange_wrapper_tests` for both a valid signedOrder and an invalid signedOrder.

## Types of changes

## Checklist:


avonian:
I just realized when I ran prettier I ran it on all files instead of just the TS files, and I failed to notice it modified package.json... I'll wait for your guys comments and I'll fix that in the next commit.

packages/0x.js/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ Edit the package's CHANGELOG.json file only.
 
 CHANGELOG
 
+## v0.36.0 - TBD
+
+    * Add `zeroEx.exchange.getOrderStateAsync` to allow obtaining current OrderState for a signedOrder
+
```
fabioberger: @avonian we actually don't manually update `CHANGELOG.md` (see comment at top of file). This is auto-generated from `CHANGELOG.json`. Please add the entry there!

packages/0x.js/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -873,6 +876,19 @@ export class ExchangeWrapper extends ContractWrapper {
...
+     * Gets the latest OrderState of a signedOrder
+     * @param   signedOrder   The signedOrder
+     */
+    public async getOrderStateAsync(signedOrder: SignedOrder): Promise<OrderState> {
+        const balanceAndProxyAllowanceLazyStore = new BalanceAndProxyAllowanceLazyStore(
+            this._tokenWrapper,
+            BlockParamLiteral.Latest,
+        );
+        const orderFilledCancelledLazyStore = new OrderFilledCancelledLazyStore(this);
+        const orderStateUtils = new OrderStateUtils(balanceAndProxyAllowanceLazyStore, orderFilledCancelledLazyStore);
```
fabioberger: It feels a bit hacky to use the LazyStore's here... We don't want any caching for this method.

The proper refactor would be passing a class that implements the abstract classes `BalanceAndAllowanceFetcher` and `OrderFilledCancelledFetcher` into `OrderStateUtils`, where their respective interfaces would be:

**balanceAndAllowanceFetcher**
```typescript
abstract class BalanceAndAllowanceFetcher {
  public abstract async getBalanceAsync(tokenAddress: string, userAddress: string): Promise<BigNumber>;
  public abstract async getProxyAllowanceAsync(tokenAddress: string, userAddress: string): Promise<BigNumber>;
}
``` 

**orderFilledCancelledFetcher**
```typescript
abstract class OrderFilledCancelledFetcher {
  public abstract async getFilledTakerAmountAsync(orderHash: string): Promise<BigNumber>;
  public abstract async getCancelledTakerAmountAsync(orderHash: string): Promise<BigNumber>;
}
```

The OrderWatcher can pass in the lazyStores, and ExchangeWrapper can pass in classes that simply fetch the latest information without caching. The OrderStateUtils doesn't care about the implementation, only that it receives the correct value.

Can you make this refactor happen? Once this is done, we can also instantiate these two fetchers once at ExchangeWrapper instantiation, rather then on every call to `getOrderStateAsync`?

avonian:
@fabioberger I've started the refactoring, but wanted to give you an update in order to get some guidance... so far I've created the two abstract fetcher classes and modified the lazy stores so they _implement_ them. I also updated order_state_utils to expect fetchers. Am I headed in the right direction? And if so, I have a few questions:

1. I wasn't sure whether to add the abstract fetcher declarations to _types.js_, or create a folder with class files, I opted for the latter but if you want I can change this.
2. What would you call the _other_ classes that fetch the latest information without caching? (the ones used by the exchange wrapper)
3. As a followup to the above, the methods in the abstract fetchers are abstractions of the same public methods that exist in the _token_wrapper_ and _exchange_wrapper_... given this, does it make sense to  make the fetchers abstract, forcing us to have another set of classes for the non-caching method invocations... or should we change the fetchers to be normal classes with the default implementations (non-caching), and modify the lazy stores _extend_ rather than _implement_?

fabioberger:
@avonian awesome work so far!

1. Let's put the abstract classes into `src/abstract`. Don't think they will be the last abstract classes we implement.
2. Even though currently the full set of methods of the abstract classes can be found in `token_wrapper` and `exchange_wrapper`, this might not always be the case. Either interface could diverge and I feel it would be a better separation of concerns to implement (an albeit simple) fetcher instance without caching (it just calls the underlying method). My suggested naming would be `SimpleBalanceAndProxyAllowanceFetcher` and `SimpleOrderFilledCancelledFetcher`.

fabioberger:
PS: don't forget to thumbs up comments once they've been addressed ;)

avonian:
@fabio that makes sense, just so I'm 100% clear from a directory structure perspective, which of the following?

**Option A**
> src/abstract.ts
> src/fetchers/simple_balance_and_allowance_fetcher.ts
> src/fetchers/simple_order_filled_cancelled_fetcher.ts

**Option B**
> src/abstract/balance_and_allowance_fetcher.ts
> src/abstract/order_filled_cancelled_fetcher.ts
> src/fetchers/simple_balance_and_allowance_fetcher.ts
> src/fetchers/simple_order_filled_cancelled_fetcher.ts

fabioberger:
Option B.

avonian:
@fabio ready for review :)

avonian:
@fabio ready for review :)

packages/0x.js/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -873,6 +876,19 @@ export class ExchangeWrapper extends ContractWrapper {
             throw new Error(errMessage);
         }
     }
+    /**
+     * Gets the latest OrderState of a signedOrder
+     * @param   signedOrder   The signedOrder
+     */
```
fabioberger: Missing `@returns` docComment.

packages/0x.js/src/fetchers/simple_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,26 @@
+import { BlockParamLiteral } from '@0xproject/types';
+import { BigNumber } from '@0xproject/utils';
+
+import {BalanceAndProxyAllowanceFetcher} from '../abstract/balance_and_proxy_allowance_fetcher';
+import {TokenWrapper} from '../contract_wrappers/token_wrapper';
+
+export class SimpleBalanceAndProxyAllowanceFetcher implements BalanceAndProxyAllowanceFetcher {
+    private _token: TokenWrapper;
```
fabioberger: Let's name this `_tokenWrapper`. We only call it `token` in 0x.ts because we can the public interface to be `zeroEx.token.getBalance()`, which is nicer then `zeroEx.tokenWrapper...`.

packages/0x.js/src/fetchers/simple_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,26 @@
...
+    private _defaultBlock: BlockParamLiteral;
+    constructor(token: TokenWrapper, defaultBlock: BlockParamLiteral) {
+        this._token = token;
+        this._defaultBlock = defaultBlock;
+    }
+    public async getBalanceAsync(tokenAddress: string, userAddress: string): Promise<BigNumber> {
+        const methodOpts = {
+            defaultBlock: this._defaultBlock,
+        };
+        return this._token.getBalanceAsync(tokenAddress, userAddress, methodOpts);
```
fabioberger: Our convention is to assign to a variable before returning. This helps with readability of exactly what is being returned.

packages/0x.js/src/fetchers/simple_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+    public async getProxyAllowanceAsync(tokenAddress: string, userAddress: string): Promise<BigNumber> {
+        const methodOpts = {
+            defaultBlock: this._defaultBlock,
+        };
+        return this._token.getProxyAllowanceAsync(tokenAddress, userAddress, methodOpts);
```
fabioberger: Ditto above.

packages/0x.js/src/fetchers/simple_order_filled_cancelled_fetcher.ts
```diff
@@ -0,0 +1,24 @@
+import { BlockParamLiteral } from '@0xproject/types';
+import { BigNumber } from '@0xproject/utils';
+
+import {OrderFilledCancelledFetcher} from '../abstract/order_filled_cancelled_fetcher';
+import {ExchangeWrapper} from '../contract_wrappers/exchange_wrapper';
+
+export class SimpleOrderFilledCancelledFetcher implements OrderFilledCancelledFetcher {
+    private _exchange: ExchangeWrapper;
```
fabioberger: `_exchangeWrapper`

packages/0x.js/src/fetchers/simple_order_filled_cancelled_fetcher.ts
```diff
@@ -0,0 +1,24 @@
...
+    constructor(exchange: ExchangeWrapper) {
```
dekz: Should this also accept a `BlockParamLiteral` like `BalanceAndProxyAllowanceFetcher `? It would require a refactor of `OrderFilledCancelledLazyStore`

avonian: I'm ok with changing the simple fetchers to accept state layer, should I also modify getOrderState with a second optional param for state layer?

fabioberger:
Good catch @avonian, we usually do a separate PR but since it's quite small, feel free to add it here.

fabioberger:
Please push your other changes, they aren't showing up for me.

packages/0x.js/src/fetchers/simple_order_filled_cancelled_fetcher.ts
avonian: I assumed the answer is yes and went ahead and modified getOrderState to support a state param.

avonian:
@fabioberger my bad I was waiting for dekz to chime in, you can review latest commit :)

PR #522, opened 2018-04-12
Forwarding contract

dekz:
## Description

This is the trade widgets forwarding contract moved into the contracts repo.

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,221 @@
...
+    function marketBuyTokens(
+        Order[] orders,
+        bytes[] signatures,
+        Order[] feeOrders,
+        bytes[] feeSignatures,
+        uint256 sellTokenAmount)
+        private
+        returns (Exchange.FillResults memory totalFillResult)
+    {
+        uint256 takerTokenBalance = sellTokenAmount;
```
abandeali1: Why do we need this to be 2 variables?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,221 @@
...
+        // Ensure the token abstraction was fair 
+        require(isAcceptableThreshold(sellTokenAmount, requestedTokensResult.takerTokenFilledAmount), ERROR_UNACCEPTABLE_THRESHOLD);
+        // Update our return FillResult with the market sell
+        addFillResults(totalFillResult, requestedTokensResult);
+        // Transfer all tokens to msg.sender
+        transferToken(makerTokenAddress, msg.sender, totalFillResult.makerTokenFilledAmount);
+        return totalFillResult;
+    }
+
+    function marketBuyNFT(
```
abandeali1: Where is this used?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,221 @@
...
+            Exchange.FillResults memory feeTokensResult =
+                buyFeeTokens(feeOrders, feeSignatures, tokensSellQuote.takerFeePaid);
+            takerTokenBalance = safeSub(takerTokenBalance, feeTokensResult.takerTokenFilledAmount);
+            totalFillResult.takerFeePaid = feeTokensResult.takerFeePaid;
+        }
+
+        // Make our market sell to buy the requested tokens with the remaining balance
+        Exchange.FillResults memory requestedTokensResult = EXCHANGE.marketSellOrders(orders, takerTokenBalance, signatures);
+        // Ensure the token abstraction was fair 
+        require(isAcceptableThreshold(sellTokenAmount, requestedTokensResult.takerTokenFilledAmount), ERROR_UNACCEPTABLE_THRESHOLD);
```
abandeali1: Since there might be some ZRX leftover here, the contract owner should probably have a way to withdraw the extra ZRX.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: This will be public once I've written the tests, since we have to transfer the NFT's individually.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: I think that is fair, it will likely be minuscule, but if they use the forwarding contract enough it might add up. We could also deduct the current user ZRX balance from future fee abstraction and use it.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,221 @@
...
+        Order[] orders,
+        bytes[] signatures,
+        Order[] feeOrders,
+        bytes[] feeSignatures)
+        payable
+        public
+        returns (Exchange.FillResults memory totalFillResult)
+    {
+        require(msg.value > 0, ERROR_INVALID_INPUT);
+        address token = readAddress(orders[0].takerAssetData, 1);
```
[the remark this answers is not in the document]
dekz: As in similar to #515 ?

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// @param order An Order struct containing order specifications.
+    /// @param takerTokenFillAmount A number representing the amount of this order to fill.
+    /// @return Amounts filled and fees paid by maker and taker.
+    function fillOrderQuote(Order order, uint256 takerTokenFillAmount)
+        public
+        view
+        returns (Exchange.FillResults memory fillResults)
+    {
+        // Compute the order hash
+        bytes32 orderHash = EXCHANGE.getOrderHash(order);
```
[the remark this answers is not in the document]
dekz: I'm going to add this as a TODO optimization before launch, as the order structure is finalised.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: We modify `takerTokenBalance` when performing fee abstraction to use as the amount to sell in market sell.

`sellTokenAmount` is not always `msg.value` and is kept as is to compare with the final fill results to make sure it's fair.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+        require(token == address(ETHER_TOKEN), ERROR_INVALID_INPUT);
+        uint256 remainingTakerTokenAmount = deductFees(msg.value, feeProportion, feeRecipient);
+        ETHER_TOKEN.deposit.value(remainingTakerTokenAmount)();
+        marketBuyNFTTokens(orders, signatures, feeOrders, feeSignatures, remainingTakerTokenAmount);
+    }
+
+    function deductFees(
```
fabioberger: Rename to `deductFee`, singular since there is only a single fee being paid.

fabioberger: Let's rename this to `payAndDeductFee`, it wasn't clear to me that this method sent the fees. `deductFee` could also simply be the arithmetic deduction of the fee amount.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+        Exchange.FillResults memory requestedTokensResult;
+        if (tokensSellQuote.takerFeePaid > 0) {
+            // Fees are required for these orders
+            // Buy enough ZRX to cover the future market sell
+            Exchange.FillResults memory feeTokensResult = buyFeeTokens(feeOrders, feeSignatures, tokensSellQuote.takerFeePaid);
+            takerTokenBalance = safeSub(takerTokenBalance, feeTokensResult.takerAssetFilledAmount);
+            // Make our market sell to buy the requested tokens with the remaining balance
+            requestedTokensResult = EXCHANGE.marketSellOrders(orders, takerTokenBalance, signatures);
+            // It's possibile to over-buy fees by a small amount as the marketSellQuote was based on 100% of ETH
+            // and marketSell (after fee abstraction) less than 100%. If a user uses this enough it may be worth withdrawing
```
fabioberger: Not sure I fully understand why it's possible of over-buy fees. Is it a rounding issue?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -261,4 +263,17 @@ contract MixinExchangeCore is
...
     }
+
+    /// @dev Calculates the sum of values already filled and cancelled for a given order.
+    /// @param orderHash The Keccak-256 hash of the given order.
+    /// @return Sum of values already filled.
+    function getUnavailableTakerTokenAmount(bytes32 orderHash)
+        public view
+        returns (uint256)
+    {
+        if (cancelled[orderHash]) {
```
hysz: An order is also cancelled if its salt less than the maker's `makerEpoch`. There is a `getOrderStatus` in an one of my upcoming PR's that could be useful here.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+            balanceOf[msg.sender] += safeSub(feeTokensResult.makerAssetFilledAmount, requestedTokensResult.takerFeePaid);
+            require(balanceOf[msg.sender] >= 0);
```
fabioberger: When could this ever by negative?

fabioberger: Also, please add error message.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+        // Ensure the token abstraction was fair if fees were proportionally too high, we fail
+        require(isAcceptableThreshold(sellTokenAmount, requestedTokensResult.takerAssetFilledAmount), ERROR_UNACCEPTABLE_THRESHOLD);
+        // Transfer all tokens to msg.sender
+        transferToken(makerTokenAddress, msg.sender, requestedTokensResult.makerAssetFilledAmount);
+        return totalFillResult;
+    }
+
+    /// @dev Withdraws an amount of ZRX tokens from using this contract.
+    /// @param amount The amount of ZRX to withdraw.
+    function withdraw(uint amount) public {
```
fabioberger: Rename to `withdrawZRX` since this is what it does.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+        uint256 totalTakerAmountSpent;
+        for (uint256 i = 0; i < orders.length; i++) {
+            totalFeeAmount = safeAdd(totalFeeAmount, orders[i].takerFee);
+        }
+        if (totalFeeAmount > 0) {
+            // Fees are required for these orders. Buy enough ZRX to cover the future fill
+            Exchange.FillResults memory feeTokensResult = buyFeeTokens(feeOrders, feeSignatures, totalFeeAmount);
+            totalTakerAmountSpent = safeAdd(totalTakerAmountSpent, feeTokensResult.takerAssetFilledAmount);
+        }
+        address makerTokenAddress;
```
hysz: What's the benefit of declaring these variables outside the loop?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+        Order[] memory feeOrders,
+        bytes[] memory feeSignatures,
+        uint16  feeProportion,
+        address feeRecipient)
+        payable
+        public
+        returns (Exchange.FillResults memory totalFillResult)
+    {
+        require(msg.value > 0, ERROR_INVALID_INPUT);
+        address token = readAddress(orders[0].takerAssetData, 1);
```
hysz: We chatted previously about having a decode library. I created [a task](https://app.asana.com/0/636508596987970/651649276736967) for myself so this doesn't get lost.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,315 @@
...
+                signatures[i]
+            );
+
+            // We didn't buy the full amount when buying ZRX as some were taken for fees
+            singleFillResult.makerAssetFilledAmount = safeSub(singleFillResult.makerAssetFilledAmount, singleFillResult.takerFeePaid);
+            // Update amounts filled and fees paid by maker and taker
+            addFillResults(totalFillResult, singleFillResult);
+
+            // Stop execution if the entire amount of makerToken has been bought
+            if (totalFillResult.makerAssetFilledAmount == feeAmount) {
```
fabioberger: Couldn't `totalFillResult.makerAssetFilledAmount` > `feeAmount`, since we are over-buying?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: The original quote was for `100ETH` and `3 ZRX` in fees, we then buy the `3 ZRX` for the `100ETH` trade, leaving us with `99.998ETH` and giving us `3ZRX`.

We can only fill the marketSell for the remaining `99.998ETH`, which will be slightly less in fees `2.99X` so there is a little over bought due to the two steps being distinct.

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    // Revert Strings
+    string constant ERROR_FAILED_TO_FILL_ALL_ORDERS = "FAILED_TO_FILL_ALL_ORDERS";
+    string constant ERROR_UNACCEPTABLE_THRESHOLD = "UNACCEPTABLE_THRESHOLD";
+    string constant ERROR_INVALID_INPUT = "INVALID_INPUT";
+
+    Exchange EXCHANGE;
+    EtherToken ETHER_TOKEN;
+    ZRXToken ZRX_TOKEN;
+    mapping (address => uint256) public balanceOf; // Balance of ZRX tokens
```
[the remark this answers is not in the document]
dekz: I could go either way on this. It's about 5000-6000 gas to save the remaining ZRX. The acceptance threshold only permits a max 2% fee abstraction. 

2% of the original order fee isn't much at all when the order is low transaction value. If we assume relayers go with a fee percentage of 1% of the order, then its 2% of 1% of the token order.  Around $5000 TTV it could become whole ZRX.

There's also a UX problem of discovering that you have a balance left in this contract. 

Perhaps it makes sense as a future feature once we see how people are using this on mainnet?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: Unlikely to happen but it is output from another contract so it's best to check that.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,270 @@
...
+        ETHER_TOKEN.approve(proxyAddress, MAX_UINT);
+        ZRX_TOKEN.approve(proxyAddress, MAX_UINT);
+    }
+
+    /// @dev Buys the tokens, performing fee abstraction if required. This function is payable
+    ///      and will convert all incoming ETH into WETH and perform the trade on behalf of the caller.
+    ///      The caller is sent all funds from the buy of orders (less any fees required to be paid in ZRX)
+    ///      If the purchased token amount does not meet some threshold (98%) then this function reverts.
+    /// @param orders An array of Order struct containing order specifications.
+    /// @param signatures An array of Proof that order has been created by maker.
```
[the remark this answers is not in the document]
dekz: `Proof that order has been created by maker` is currently consistent language with the descriptions in Exchange contract

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
```diff
@@ -0,0 +1,139 @@
...
+        if (fillResults.takerAssetFilledAmount == 0) {
+            return fillResults;
+        }
+        // Validate fill order rounding
+        if (isRoundingError(fillResults.takerAssetFilledAmount, order.takerAssetAmount, order.makerAssetAmount)) {
+            fillResults.takerAssetFilledAmount = 0;
+            return fillResults;
+        }
+        fillResults.makerAssetFilledAmount = getPartialAmount(fillResults.takerAssetFilledAmount, order.takerAssetAmount, order.makerAssetAmount);
+        if (order.feeRecipientAddress != address(0)) {
```
fabioberger: This is out-of-date with V2. We no longer enforce that `feeRecipientAddress` is not NULL_ADDRESS. I guess I'm concerned with us keeping this logic up-to-date with Exchange in general but don't see a way around this atm.

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
```diff
@@ -0,0 +1,139 @@
...
+    /// @param orders An array of Order struct containing order specifications.
+    /// @param takerAssetFillAmount A number representing the amount of this order to fill.
+    /// @return Amounts filled and fees paid by maker and taker.
+    function expectedMaketSellFillResults(Order[] memory orders, uint256 takerAssetFillAmount)
+        public
+        view
+        returns (Exchange.FillResults memory fillResult)
+    {
+        for (uint256 i = 0; i < orders.length; i++) {
+            require(areBytesEqual(orders[i].makerAssetData, orders[0].makerAssetData), "all orders must be the same token pair");
```
fabioberger: This is only checking one side of the token pairing. Should we also be checking that the `takerAssetData` are the same?

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
```diff
@@ -0,0 +1,139 @@
...
+            addFillResults(fillResult, quoteFillResult);
+            if (fillResult.takerAssetFilledAmount == takerAssetFillAmount) {
+                break;
+            }
+        }
+        return fillResult;
+    }
+
+    /// @dev Calculates a quote total for buyTokens. This is useful for off-chain queries to 
+    ///      ensure all calculations are performed atomically for consistent results
```
fabioberger: Hm, doing this off-chain however is not guarenteed to give the same result as when the tx is actually submitted b/c of race conditions. When and how exactly do you expect this to be used?

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
```diff
@@ -0,0 +1,139 @@
...
+        Order[] memory orders,
+        uint256 zrxAmount)
+        public
+        view
+        returns (Exchange.FillResults memory totalFillResult)
+    {
+        address token = readAddress(orders[0].makerAssetData, 1);
+        require(token == address(ZRX_TOKEN), "order taker asset must be ZRX");
+        for (uint256 i = 0; i < orders.length; i++) {
+            require(areBytesEqual(orders[i].makerAssetData, orders[0].makerAssetData), "all orders must be the same token pair");
```
fabioberger: Ditto above. Only checking `makerAssetData`, not `takerAssetData`

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
...
+    // public static async getForwarderContractAsync(web3Wrapper: Web3Wrapper): Promise<ForwarderContract> {
+    //     const networkId = await web3Wrapper.getNetworkIdAsync();
+    //     const [abi, address] = await this._getContractAbiAndAddressFromArtifactsAsync(
+    //         artifacts.Forwarder,
+    //         web3Wrapper,
+    //         networkId,
+    //     );
+    //     const contractInstance = new ForwarderContract(web3Wrapper, abi, address);
+    //     return contractInstance;
+    // }
```
fabioberger: Should this be removed?

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
...
+        }
+        const abiAndAddress: [ContractAbi, string] = [artifact.networks[networkId].abi, contractAddress];
+        return abiAndAddress;
+    }
+    protected static _getContractAddress(artifact: Artifact, networkId: number, addressIfExists?: string): string {
+        let contractAddress;
+        if (_.isUndefined(addressIfExists)) {
+            contractAddress = artifact.networks[networkId].address;
+            if (_.isUndefined(contractAddress)) {
+                throw new Error('ContractDoesNotExist');
```
fabioberger: Our convention is all caps with underscores between words for error messages.

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
...
+            }
+        } else {
+            contractAddress = addressIfExists;
+        }
+        return contractAddress;
+    }
+
```
fabioberger: Remove space.

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
...
+        return quote;
+    }
+    public async buyNFTTokensAsync(
+        orders: SignedOrder[],
+        feeOrders: SignedOrder[],
+        from: string,
+    ): Promise<TransactionReceiptWithDecodedLogs> {
+        let fillAmountWei = _.reduce(
+            orders,
+            (prev: BigNumber, order: SignedOrder) => {
```
fabioberger: `prev` is a misnomer here, it's not the previous value, but rather the cumulative value.

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
...
+                return prev.plus(order.takerAssetAmount);
+            },
+            new BigNumber(0),
+        );
+        const feeParams = formatters.createMarketSellOrders(feeOrders, new BigNumber(0));
+        const params = formatters.createMarketSellOrders(orders, fillAmountWei);
+        const totalFees = _.reduce(
+            orders,
+            (prev: BigNumber, order: SignedOrder) => {
```
fabioberger: Ditto above.

packages/contracts/src/utils/forwarder_wrapper.ts
```diff
@@ -0,0 +1,185 @@
+import { Order, OrderTransactionOpts } from '0x.js';
+import { ContractAbi, TransactionReceiptWithDecodedLogs } from '@0xproject/types';
+import { AbiDecoder, BigNumber } from '@0xproject/utils';
+import { Web3Wrapper } from '@0xproject/web3-wrapper';
+import * as ethUtil from 'ethereumjs-util';
```
fabioberger: Unused import.

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
+import { TransactionReceiptWithDecodedLogs, ZeroEx } from '0x.js';
+import { BlockchainLifecycle, devConstants, web3Factory } from '@0xproject/dev-utils';
+import { BigNumber } from '@0xproject/utils';
+import { Web3Wrapper } from '@0xproject/web3-wrapper';
+import * as chai from 'chai';
+import ethUtil = require('ethereumjs-util');
```
fabioberger: Unused import.

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+    describe('buyTokens', () => {
+        it('should fill the order', async () => {
+            const fillAmount = signedOrder.takerAssetAmount.div(2);
+            const makerBalanceBefore = erc20Balances[makerAddress][defaultMakerAssetAddress];
+            const takerBalanceBefore = erc20Balances[takerAddress][defaultMakerAssetAddress];
+            const tx = await forwarderWrapper.buyTokensAsync([signedOrder], [], fillAmount, takerAddress);
```
fabioberger: Assign `[signedOrder]` and `[]` to well-named variable names to help code readability.

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+            const takerBalanceAfter = newBalances[takerAddress][defaultMakerAssetAddress];
+            const makerTokenFillAmount = fillAmount
+                .times(signedOrder.makerAssetAmount)
+                .dividedToIntegerBy(signedOrder.takerAssetAmount);
+
+            expect(makerBalanceAfter).to.be.bignumber.equal(makerBalanceBefore.minus(makerTokenFillAmount));
+            expect(takerBalanceAfter).to.be.bignumber.equal(takerBalanceBefore.plus(makerTokenFillAmount));
+            expect(newBalances[forwarderContract.address][weth.address]).to.be.bignumber.equal(new BigNumber(0));
+        });
+
```
fabioberger: Remove new line. Why is TSLint not catching this?

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+        it('should fill the order and perform fee abstraction', async () => {
+            const fillAmount = signedOrder.takerAssetAmount.div(4);
+            const takerBalanceBefore = erc20Balances[takerAddress][defaultMakerAssetAddress];
+            await forwarderWrapper.buyTokensAsync([orderWithFees], [feeOrder], fillAmount, takerAddress);
+            const newBalances = await erc20Wrapper.getBalancesAsync();
+            const takerBalanceAfter = newBalances[takerAddress][defaultMakerAssetAddress];
+
+            const acceptPerc = 98;
```
fabioberger: `acceptablePercent`

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+        });
+    });
+    describe('buyTokensFee', () => {
+        it('should fill the order and send fee to fee recipient', async () => {
+            const initEthBalance = await web3Wrapper.getBalanceInWeiAsync(feeRecipientAddress);
+            const fillAmount = signedOrder.takerAssetAmount.div(2);
+            const feeProportion = 150; // 1.5%
+            const txHash = await forwarderWrapper.buyTokensFeeAsync(
+                [signedOrder],
+                [],
```
fabioberger: Ditto above. We never pass literals as params.

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+                const txHash = await forwarderWrapper.buyTokensFeeAsync(
+                    [signedOrder],
+                    [],
+                    fillAmount,
+                    feeProportion,
+                    feeRecipientAddress,
+                    takerAddress,
+                );
+                expect.fail(); // Never reached
+            } catch (err) {
```
fabioberger: What is the error thrown? Anything worth checking?

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,306 @@
...
+        });
+    });
+    describe('Non fungible tokens', async () => {
+        it('buys non fungible tokens', async () => {
+            const makerAssetId = erc721MakerAssetIds[0];
+            signedOrder = orderFactory.newSignedOrder({
+                makerAssetAmount: new BigNumber(1),
+                makerAssetData: assetProxyUtils.encodeERC721ProxyData(erc721Token.address, makerAssetId),
+            });
+            const tx = await forwarderWrapper.buyNFTTokensAsync([signedOrder], [feeOrder], takerAddress);
```
fabioberger: Let's at least check the tx returned has status '1'. Is there nothing else worth checking? Like the actual balances post trade of the two participants perhaps.

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
dekz: It will fail as the takerAssetData is always WETH and won't have a balance of any other tokens to  fill the order.

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
dekz: This is useful off-chain and the atomicity guarantee is for the marketSell and fee abstraction. It can be calculated off-chain though performing it in 2 separate calls could be spread over different block numbers. This is here to aid that situation. It's use would be an accurate display to the end user.

packages/contracts/test/forwarder/forwarder.ts
dekz: It will be a revert, once ganache-cli supports returning error messages we can check the error message matches

packages/contracts/test/forwarder/forwarder.ts
dekz: I believe our tslint only errors on multiple consecutive lines, rather than breaking up with single lines to aid readability.

packages/contracts/src/utils/forwarder_wrapper.ts
dekz: This is removed for now with the above TODO remove, we're not using it in tests but we may have a more generic way of initializing these contracts from artifacts.

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    /// @dev Checks whether the amount of tokens sold against the amount of tokens requested
+    ///      is within a certain threshold. This ensures the caller gets a fair deal when
+    ///      performing any token fee abstraction. Threshold is 98%. If fee abstraction costs more than
+    ///      2% of the total transaction, we return false.
```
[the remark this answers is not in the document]
dekz: We don't have much real world data to go off, but it was an assumption that fees would be no greater than 2%. If this is unlikely happy to change it or we can redeploy easily if the landscape changes.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,310 @@
...
+    /// @param feeOrders An array of Order struct containing order specifications for fees.
+    /// @param feeSignatures An array of Proof that order has been created by maker for the fee orders.
+    /// @param feeProportion A proportion deducted off the incoming ETH and sent to feeRecipient. The maximum value for this
+    ///        is 1000, aka 10%. Supports up to 2 decimal places. I.e 0.59% is 59.
+    /// @param feeRecipient An address of the fee recipient whom receives feeProportion of ETH.
+    /// @return FillResults amounts filled and fees paid by maker and taker.
+    function buyTokensFee(
```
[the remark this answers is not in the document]
dekz: It's a little bit more expensive but I think it's worth combing the two.

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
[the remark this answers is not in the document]
fabioberger: Ok, let's perhaps move it up to 4-5%, just to be on the safe side. What do you think @abandeali1?

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
fabioberger: Not sure I understand your explanation. Could you elaborate?

packages/contracts/test/forwarder/forwarder.ts
fabioberger: I see... We should change that.

packages/contracts/src/contracts/current/forwarder/MixinForwarderQuote.sol
dekz: The underlying action here is multiple steps, buy tokens to pay fees and then buy the requested tokens. 

We could query those separately with their respective `calculateExpectedFillResult` in two calls to the blockchain. During this time, between the two calls, the blockchain progresses to the next block where one has been partially filled. Let's say it's the REP orders where a partial fill occurred. We would be showing the user a less accurate quote in the UI.

The rate will be accurate to a single block rather than potentially over multiple.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,243 @@
...
+import "../utils/Ownable/Ownable.sol";
+
+contract Forwarder is
+    MixinForwarderCore,
+    MixinForwarderQuote,
+    MixinERC721,
+    MixinERC20,
+    Ownable
+{
+    function Forwarder(
```
LogvinovLeon: Rebase of 0.23.0 PR and use constructor

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,243 @@
...
+    /// @return FillResults amounts filled and fees paid by maker and taker.
+    function buyNFTTokens(
+        Order[] memory orders,
+        bytes[] memory signatures,
+        Order[] memory feeOrders,
+        bytes[] memory feeSignatures,
+        uint16  feeProportion,
+        address feeRecipient)
+        payable
+        public
```
LogvinovLeon: Specify the return value

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,243 @@
...
+    {
+        require(msg.value > 0, "msg.value must be greater than 0");
+        address token = readAddress(orders[0].takerAssetData, 1);
+        require(token == address(ETHER_TOKEN), "order taker asset must be Wrapped ETH");
+        uint256 remainingTakerTokenAmount = payAndDeductFee(msg.value, feeProportion, feeRecipient);
+        ETHER_TOKEN.deposit.value(remainingTakerTokenAmount)();
+        marketBuyNFTTokens(orders, signatures, feeOrders, feeSignatures, remainingTakerTokenAmount);
```
LogvinovLeon: Don't forget to return fill results

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,243 @@
...
+        onlyOwner
+    {
+        address proxyAddress = EXCHANGE.getAssetProxy(assetProxyId);
+        if (proxyAddress != address(0)) {
+            ETHER_TOKEN.approve(proxyAddress, MAX_UINT);
+            ZRX_TOKEN.approve(proxyAddress, MAX_UINT);
+        }
+    }
+
+}
```
LogvinovLeon: \n

packages/contracts/src/contracts/current/forwarder/MixinERC721.sol
```diff
@@ -0,0 +1,24 @@
...
+
+    function transferNFTToken(
+        address token,
+        address to,
+        uint256 tokenId)
+        internal
+    {
+        ERC721Basic(token).transferFrom(address(this), to, tokenId);
+    }
+}
```
LogvinovLeon: \n

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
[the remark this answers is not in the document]
LogvinovLeon: Can't the app just redeploy that contract without that limit? This is just an example - no?

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,81 @@
...
+    /// @param numerator Numerator.
+    /// @param denominator Denominator.
+    /// @param target Value to calculate partial of.
+    /// @return Partial value of target.
+    function getPartialAmount(uint256 numerator, uint256 denominator, uint256 target)
+        internal
+        pure
+        returns (uint256 partialAmount)
+    {
+        partialAmount = safeDiv(safeMul(numerator, target), denominator);
```
LogvinovLeon: @Remco Do we have your chineese reminder solution somewhere in the codebase? Can we just use it?

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
dekz: Fill results here don't make as much sense here as they are NFTs and the total makerAssetTokenAmount would be the number of tokens. Perhaps it makes sense for the fees part.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    /// @dev Sets the allowances on the proxy for this contract
+    function setERC20ProxyApproval(uint8 assetProxyId)
+        public
+        onlyOwner
+    {
+        address proxyAddress = EXCHANGE.getAssetProxy(assetProxyId);
+        if (proxyAddress != address(0)) {
+            ETHER_TOKEN.approve(proxyAddress, MAX_UINT);
```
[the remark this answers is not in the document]
dekz: We can always redeploy the contract

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    /// @dev Buys feeAmount of ZRX fee tokens, taking into account the fees on buying fee tokens. This will guarantee
+    ///      At least feeAmount of ZRX fee tokens are purchased (sometimes slightly over due to rounding issues).
+    ///      It is possible that a request to buy 200 ZRX fee tokens will require purchasing 202 ZRX tokens
+    ///      As 2 ZRX is required to purchase the 200 ZRX fee tokens. This guarantees at least 200 ZRX for future purchases.
+    /// @param orders An array of Order struct containing order specifications for fees.
+    /// @param signatures An array of Proof that order has been created by maker for the fee orders.
+    /// @param feeAmount The number of requested ZRX fee tokens.
+    /// @return totalFillResults Amounts filled and fees paid by maker and taker. makerTokenAmount is the zrx amount deducted of fees
+    function buyFeeTokensInternal(
```
hysz: Why does buying fee tokens differ from buying an exact number of ERC20 tokens? This looks a lot like the market buy in the exchange core, which is called by `buyExactTokensInternal`.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactTokens.sol
```diff
@@ -0,0 +1,128 @@
...
+                totalFillResults.takerAssetFilledAmount = feeTokensResult.takerAssetFilledAmount;
+                totalFillResults.takerFeePaid = feeTokensResult.takerFeePaid;
+            }
+            // Make our market buy of the requested tokens with the remaining balance
+            requestedTokensResult = EXCHANGE.marketBuyOrders(orders, tokenAmount, signatures);
+            require(requestedTokensResult.makerAssetFilledAmount == tokenAmount, "traded amount did not meet acceptable threshold");
+        }
+        addFillResults(totalFillResults, requestedTokensResult);
+        // Transfer all tokens to msg.sender
+        transferToken(makerTokenAddress, msg.sender, totalFillResults.makerAssetFilledAmount);
```
hysz: Does sender abstraction allow us to perform these transfers in the Exchange? @abandeali1

packages/contracts/src/contracts/current/forwarder/MixinBuyExactTokens.sol
```diff
@@ -0,0 +1,128 @@
...
+        }
+        if (totalFeeAmount > 0) {
+            // Fees are required for these orders. Buy enough ZRX to cover the future fill
+            Exchange.FillResults memory feeTokensResult = buyFeeTokensInternal(feeOrders, feeSignatures, totalFeeAmount);
+            totalFillResults.takerFeePaid = feeTokensResult.takerFeePaid;
+            totalFillResults.takerAssetFilledAmount = feeTokensResult.takerAssetFilledAmount;
+        }
+        for (uint256 n = 0; n < orders.length; n++) {
+            // Fail if it wasn't fully filled otherwise we will keep WETH
+            Exchange.FillResults memory fillOrderResults = EXCHANGE.fillOrKillOrder(
```
hysz: Can we use `batchFillOrKillOrders` here?

packages/contracts/src/contracts/current/forwarder/MixinBuyExactTokens.sol
```diff
@@ -0,0 +1,128 @@
...
+    {
+        require(msg.value > 0, "msg.value must be greater than 0");
+        address token = readAddress(orders[0].takerAssetData, 1);
+        require(token == address(ETHER_TOKEN), "order taker asset must be Wrapped ETH");
+        uint8 proxyId = uint8(orders[0].makerAssetData[0]);
+        require(proxyId == 1 || proxyId == 2, "unsupported token proxy");
+
+        uint256 remainingTakerTokenAmount = msg.value;
+        ETHER_TOKEN.deposit.value(remainingTakerTokenAmount)();
+        if (proxyId == 1) {
```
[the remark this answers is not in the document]
dekz: Yeah the calculation is different for ERC721 (we sum all fees) versus ERC20 (we calculate expected results) and the scenario for buying ERC20 ZRX is different (we buy from the orders rather than feeOrders as its cheaper than 2 fills)

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
dekz: The exchange rate is different when buying ZRX and there are ZRX fees. If we don't take this into account than fees are removed and a 100 BUY of ZRX returns 98 ZRX (after fees). If this happens theres not enough to fill other orders (ERC721 for example) are we are short

packages/contracts/src/contracts/current/forwarder/MixinBuyExactTokens.sol
```diff
@@ -0,0 +1,128 @@
...
+            totalFillResults = buyExactTokensInternal(orders, signatures, feeOrders, feeSignatures, tokenAmount);
```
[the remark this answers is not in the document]
dekz: Consolidated this and added a require to check `assetAmount == orders.length`

packages/contracts/src/contracts/current/forwarder/MixinBuyExactTokens.sol
dekz: We could but we would also have to for loop and transfer the tokens back. Unless spender and receiver were separated.

packages/contracts/src/contracts/current/forwarder/MixinERC721.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity ^0.4.24;
+
+import "zeppelin-solidity/contracts/token/ERC721/ERC721Basic.sol";
+contract MixinERC721 {
```
fabioberger: Add new line between import and contract declaration.

packages/contracts/src/contracts/current/forwarder/MixinErrorMessages.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.4.24;
+
+contract MixinErrorMessages {
+    string constant VALUE_GT_ZERO = "VALUE_GT_ZERO";
```
fabioberger: Let's write out `GREATER_THAN`

packages/contracts/src/contracts/current/forwarder/MixinErrorMessages.sol
```diff
@@ -0,0 +1,13 @@
...
+    string constant FEE_PROPORTION_TOO_LARGE = "FEE_PROPORTION_TOO_LARGE";
+    string constant TAKER_ASSET_ZRX_REQUIRED = "TAKER_ASSET_ZRX_REQUIRED";
+    string constant TAKER_ASSET_WETH_REQUIRED = "TAKER_ASSET_WETH_REQUIRED";
+    string constant SAME_ASSET_TYPE_REQUIRED = "SAME_ASSET_TYPE_REQUIRED";
+    string constant NOT_ACCEPTABLE_THRESHOLD = "NOT_ACCEPTABLE_THRESHOLD";
```
fabioberger: `UNACCEPTABLE`

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,147 @@
...
+    /// @return remainingTakerTokenAmount the amount remaining after fees have been deducted and paid to feeRecipient.
+    function payAndDeductFee(
+        uint256 takerTokenAmount,
+        uint16 feeProportion,
+        address feeRecipient)
+        internal
+        returns (uint256 remainingTakerTokenAmount)
+    {
+        remainingTakerTokenAmount = takerTokenAmount;
+        if (feeProportion > 0 && feeRecipient != address(0x0)) {
```
fabioberger: Can't you just write: `address(0)`?

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,147 @@
...
+    /// @return bool of whether this is within an acceptable threshold
+    function isAcceptableThreshold(uint256 requestedTokenAmount, uint256 soldTokenAmount)
+        internal
+        pure
+        returns (bool)
+    {
+        uint256 acceptableProportion = safeDiv(safeMul(requestedTokenAmount, ALLOWABLE_EXCHANGE_PERCENTAGE), PERCENTAGE_DENOMINATOR);
+        return soldTokenAmount >= acceptableProportion;
+    }
+}
```
fabioberger: \n

packages/contracts/src/utils/artifacts.ts
```diff
@@ -26,6 +27,7 @@ export const artifacts = {
     ERC721Proxy: (ERC721Proxy as any) as ContractArtifact,
     Exchange: (Exchange as any) as ContractArtifact,
     EtherToken: (EtherToken as any) as ContractArtifact,
+    Forwarder: (Forwarder as any) as ContractArtifact,
```
fabioberger: Can you add the Forwarder to the `migrations` package compilation?

packages/contracts/test/forwarder/forwarder.ts
```diff
@@ -0,0 +1,429 @@
...
+            });
+            const newBalances = await erc20Wrapper.getBalancesAsync();
+            const takerBalanceBefore = balancesBefore[takerAddress][defaultMakerAssetAddress];
+            const takerBalanceAfter = newBalances[takerAddress][defaultMakerAssetAddress];
+            const afterEthBalance = await web3Wrapper.getBalanceInWeiAsync(takerAddress);
+            const expectedEthBalanceAfterGasCosts = initEthBalance.minus(fillAmount).minus(tx.gasUsed);
+            expect(takerBalanceAfter).to.be.bignumber.eq(takerBalanceBefore.plus(assetAmount));
+            expect(afterEthBalance).to.be.bignumber.eq(expectedEthBalanceAfterGasCosts);
+        });
+        it('should buy the exact amount of assets and with fee abstraction', async () => {
```
fabioberger: Typo: `and with`

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,48 @@
...
+        EtherToken _etherToken,
+        ZRXToken _zrxToken,
+        uint8 assetProxyId)
+        public
+        Ownable()
+    {
+        EXCHANGE = _exchange;
+        ETHER_TOKEN = _etherToken;
+        ZRX_TOKEN = _zrxToken;
+        setERC20ProxyApproval(assetProxyId);
```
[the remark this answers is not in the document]
dekz: No since this is for moving ZRX and WETH, we receive the ERC721 (so the maker needs the allowance) then we send the ERC721 directly back to msg.sender.

packages/contracts/src/contracts/current/forwarder/MixinBuyNFTTokens.sol
```diff
@@ -0,0 +1,84 @@
+pragma solidity ^0.4.22;
+pragma experimental ABIEncoderV2;
+
+import "./MixinForwarderCore.sol";
+import "./MixinForwarderQuote.sol";
+import "./MixinERC721.sol";
+
+contract MixinBuyNFTTokens is
```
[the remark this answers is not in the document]
dekz: This is gone, I'm not sure how you were able to review it since the file is been combined to MixinBuyExactAssets.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,132 @@
...
+        address feeRecipient)
+        payable
+        public
+        returns (Exchange.FillResults memory totalFillResults)
+    {
+        require(msg.value > 0, VALUE_GT_ZERO);
+        address takerAsset = readAddress(orders[0].takerAssetData, 0);
+        require(takerAsset == address(ETHER_TOKEN), TAKER_ASSET_WETH_REQUIRED);
+        uint8 proxyId = uint8(orders[0].makerAssetData[orders[0].makerAssetData.length-1]);
+        require(proxyId == 1 || proxyId == 2, UNSUPPORTED_TOKEN_PROXY);
```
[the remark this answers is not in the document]
dekz: yeah this is hopefully to be improved from greg's PR about decoding the asset data. But I'll move to constants for now.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(
+            isAcceptableThreshold(
+                remainingTakerAssetAmount,
+                totalFillResults.takerAssetFilledAmount),
+            UNACCEPTABLE_THRESHOLD);
+        withdrawPayAndDeductFee(remainingTakerAssetAmount, totalFillResults.takerAssetFilledAmount, feeProportion, feeRecipient);
+        return totalFillResults;
+    }
+
+    function buyExactERC20TokensInternal(
```
hysz: We've been adding function descriptions to the protocol contracts, and I think it could be useful in the forwarding contract as well. I realize we started doing this after this PR was created =\, but it would be super helpful.
Ex for this function:
```
/// @dev Buys an exact amount of an ERC20 token using WETH.
/// @param orders Orders to fill. The maker asset is the ERC20 token to buy. The taker asset is WETH.
/// @param signatures Proof that the orders were created by their respective makers.
/// @param feeOrders to fill. The maker asset is ZRX and the taker asset is WETH.
/// @param feeSignatures Proof that the feeOrders were created by their respective makers.
/// @param assetAmount Amount of the ERC20 token to buy.
/// @return totalFillResults Aggregated fill results of buying the ERC20 and ZRX tokens.
```

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,137 @@
...
+        Order[] memory orders,
+        bytes[] memory signatures,
+        Order[] memory feeOrders,
+        bytes[] memory feeSignatures,
+        uint256 assetAmount)
+        private
+        returns (Exchange.FillResults memory totalFillResults)
+    {
+        address makerTokenAddress = readAddress(orders[0].makerAssetData, 0);
```
hysz: Is it safe to assume that the takerTokenAddress is WETH?

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,137 @@
...
+        for (uint256 n = 0; n < orders.length; n++) {
+            // Fail if it wasn't fully filled otherwise we will keep WETH
+            Exchange.FillResults memory fillOrderResults = EXCHANGE.fillOrKillOrder(
+                orders[n],
+                orders[n].takerAssetAmount,
+                signatures[n]
+            );
+            addFillResults(totalFillResults, fillOrderResults);
+            uint256 tokenId = readUint256(orders[n].makerAssetData, 20);
+            transferNFTToken(makerTokenAddress, msg.sender, tokenId);
```
hysz: Can we inherit from ERC721Proxy and call transferFromInternal?

I realize this is slightly more efficient since we don't have to decode the `makerTokenAddress`, but it would reduce the complexity. Esp once we merge the PR with `receiverData`/`safeTransferFrom`.

packages/contracts/src/contracts/current/forwarder/MixinForwarderCore.sol
```diff
@@ -0,0 +1,147 @@
...
+    ///      5% of the total transaction, we return false.
+    /// @param requestedTokenAmount The amount the user requested, or sent in to a payable function
+    /// @param soldTokenAmount The amount of the token that was sold after fee abstraction
+    /// @return bool of whether this is within an acceptable threshold
+    function isAcceptableThreshold(uint256 requestedTokenAmount, uint256 soldTokenAmount)
+        internal
+        pure
+        returns (bool)
+    {
+        uint256 acceptableProportion = safeDiv(safeMul(requestedTokenAmount, ALLOWABLE_EXCHANGE_PERCENTAGE), PERCENTAGE_DENOMINATOR);
```
hysz: I've seen this in a few places, probably worth mentioning that we do have a `getPartialAmount` now in `LibMath`.

packages/contracts/src/contracts/current/forwarder/MixinForwarderExpectedResults.sol
```diff
@@ -0,0 +1,114 @@
...
+        returns (Exchange.FillResults memory totalFillResults)
+    {
+        for (uint256 i = 0; i < orders.length; i++) {
+            require(areBytesEqual(orders[i].makerAssetData, orders[0].makerAssetData), SAME_ASSET_TYPE_REQUIRED);
+            uint256 remainingTakerAssetFillAmount = safeSub(takerAssetFillAmount, totalFillResults.takerAssetFilledAmount);
+
+            Exchange.FillResults memory fillOrderExpectedResults = calculateFillOrderFillResults(orders[i], remainingTakerAssetFillAmount);
+
+            addFillResults(totalFillResults, fillOrderExpectedResults);
+            if (totalFillResults.takerAssetFilledAmount == takerAssetFillAmount) {
```
hysz: Is it possible to have a small margin of error?

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
dekz: This is checked in the public methods:
```
        address takerAsset = readAddress(orders[0].takerAssetData, 0);
        require(takerAsset == address(ETHER_TOKEN), TAKER_ASSET_WETH_REQUIRED);
```

packages/contracts/src/contracts/current/forwarder/MixinForwarderExpectedResults.sol
dekz: Can you go into more detail about the small margin of error you potentially see?

dekz:
Some changes from on-going discussions:

- [ ] `batchFillOrders` to return `fillResults` (even if this doesn't make sense for mixed erc20 tokens in a batch fill)
- [ ] Use batchFillOrders in Forwarder
- [ ] Look at storing `msg.sender` in the Forwarder and using during onTokensReceived.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,48 @@
...
+    {
+        require(msg.sender == address(ETHER_TOKEN), DEFAULT_FUNCTION_WETH_CONTRACT_ONLY);
+    }
+
+    /// @dev Sets the allowances to the proxy for this contract
+    function setERC20ProxyApproval(uint8 erc20AssetProxyId)
+        public
+        onlyOwner
+    {
+        address proxyAddress = EXCHANGE.getAssetProxy(erc20AssetProxyId);
```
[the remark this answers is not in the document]
dekz: This is a loose enforcement of state between the Exchange and Forwarder. A Forwarder could only have its allowances set to proxies that exist in the Exchange contract, rather then any arbitrary address. 

I expect this to almost never be called unless we were to upgrade proxies in the Exchange.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,155 @@
...
+            if (calculatedMarketBuyResults.takerFeePaid > 0) {
+                // Fees are required for these orders. Buy enough ZRX to cover the future market buy
+                Exchange.FillResults memory feeTokensResult = buyFeeTokensInternal(
+                    feeOrders, feeSignatures, calculatedMarketBuyResults.takerFeePaid);
+                totalFillResults.takerAssetFilledAmount = feeTokensResult.takerAssetFilledAmount;
+                totalFillResults.takerFeePaid = feeTokensResult.takerFeePaid;
+            }
+            // Make our market buy of the requested tokens with the remaining balance
+            requestedTokensResults = EXCHANGE.marketBuyOrders(orders, assetAmount, signatures);
+            require(requestedTokensResults.makerAssetFilledAmount == assetAmount, UNACCEPTABLE_THRESHOLD);
```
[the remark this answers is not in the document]
dekz: That check is only because we can overbuy ZRX due to rounding. That is, we always want to buy more rather than buy less. If we buy less we cannot fully satisfy the order, so we overbuy slightly.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,155 @@
...
+            requestedTokensResults = buyFeeTokensInternal(orders, signatures, assetAmount);
+            // When buying ZRX we round up which can result in a small margin excess
+            require(requestedTokensResults.makerAssetFilledAmount >= assetAmount, UNACCEPTABLE_THRESHOLD);
+        } else {
+            Exchange.FillResults memory calculatedMarketBuyResults = calculateMarketBuyFillResults(orders, assetAmount);
+            if (calculatedMarketBuyResults.takerFeePaid > 0) {
+                // Fees are required for these orders. Buy enough ZRX to cover the future market buy
+                Exchange.FillResults memory feeTokensResult = buyFeeTokensInternal(
+                    feeOrders, feeSignatures, calculatedMarketBuyResults.takerFeePaid);
+                totalFillResults.takerAssetFilledAmount = feeTokensResult.takerAssetFilledAmount;
```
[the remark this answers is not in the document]
dekz: We don't need one as if we don't have enough ZRX then it will fail to fill the assetAmount.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
```diff
@@ -0,0 +1,48 @@
...
+import "./MixinBuyExactAssets.sol";
+import "./MixinMarketBuyERC20Tokens.sol";
+import "../utils/Ownable/Ownable.sol";
+
+contract Forwarder is
+    MixinForwarderCore,
+    MixinMarketBuyERC20Tokens,
+    MixinBuyExactAssets
+{
+    constructor (
```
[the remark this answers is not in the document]
dekz: These values will change on testnets though.

packages/contracts/src/contracts/current/forwarder/MixinBuyExactAssets.sol
```diff
@@ -0,0 +1,155 @@
...
+        uint256 assetAmount,
+        uint16  feeProportion,
+        address feeRecipient
+    )
+        payable
+        public
+        returns (Exchange.FillResults memory totalFillResults)
+    {
+        require(msg.value > 0, VALUE_GREATER_THAN_ZERO);
+        address takerAsset = readAddress(orders[0].takerAssetData, 0);
```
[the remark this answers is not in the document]
dekz: This is doable, but I think I'm running into some popping issues. Where the length is modified. Since the string is not a compile time constant this feels like a problem.

packages/contracts/src/contracts/current/forwarder/Forwarder.sol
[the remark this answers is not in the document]
abandeali1: Yeah, just wanted to add a reminder.

packages/contracts/src/contracts/current/forwarder/MixinERC721.sol
```diff
@@ -0,0 +1,33 @@
...
+        bytes memory assetData,
+        address from,
+        address to,
+        uint256 amount
+    )
+        internal
+    {
+        // Pop off the proxy id as this needs to be done prior
+        popLastByte(assetData);
+        ERC721Proxy.transferFromInternal(assetData, from, to, amount);
```
abandeali1: I found this confusing, since we're not actually calling the proxy here. Maybe `import { ERC721Proxy as ERC721Transfer} from ...` would make this more clear.

packages/contracts/src/contracts/current/forwarder/MixinForwarderExpectedResults.sol
```diff
@@ -0,0 +1,130 @@
...
+    function calculateMarketSellFillResults(
+        Order[] memory orders,
+        uint256 takerAssetFillAmount
+    )
+        internal
+        view
+        returns (Exchange.FillResults memory totalFillResults)
+    {
+        for (uint256 i = 0; i < orders.length; i++) {
+            require(
```
abandeali1: Do we still need this check after optimizing the calldata?

packages/contracts/src/contracts/current/forwarder/MixinERC721.sol
dekz: This would be solved with #659 which separates the responsibilities of ERC721 Transfer and the AssetProxy. In the meantime I'll import it and alias it.

packages/contracts/src/contracts/current/forwarder/MixinForwarderExpectedResults.sol
dekz: Good catch I think we can completely drop this

packages/contract_templates/partials/call.handlebars
```diff
@@ -1,3 +1,3 @@
-public {{this.name}} = {
+public {{this.tsName}} = {
```
abandeali1: Why do we need this change?

packages/contracts/src/2.0.0/forwarder/MixinExpectedResults.sol
```diff
@@ -0,0 +1,158 @@
...
+        LibOrder.OrderInfo memory orderInfo = EXCHANGE.getOrderInfo(order);
+        if (orderInfo.orderStatus != uint8(LibOrder.OrderStatus.FILLABLE)) {
+            return fillResults;
+        }
+        uint256 remainingTakerAssetAmount = safeSub(order.takerAssetAmount, orderInfo.orderTakerAssetFilledAmount);
+        uint256 takerAssetFilledAmount = min256(takerAssetFillAmount, remainingTakerAssetAmount);
+
+        fillResults.takerAssetFilledAmount = takerAssetFilledAmount;
+        fillResults.makerAssetFilledAmount = getPartialAmount(
+            fillResults.takerAssetFilledAmount,
```
abandeali1: Small optimization, but we should just use `takerAssetFilledAmount` here (this is fixed in the core contracts in #826 )

packages/contract_templates/partials/call.handlebars
dekz: Just to be safe I added backwards compatibility for callbacks on older ERC721 and new ERC721 with operator by overloading the onERC721Received and returning the correct bytes4.

This is a view and needs our overloaded workaround  `onERC721Received` and `onERC721Received1`.

PR #524, opened 2018-04-13
Remove truffle-hdwallet from 0x.js tests

dekz:
## Description

Removes the need for truffle-hdwallet in the 0x.js artifacts test. Removes an old version of bip39 from our dependencies.

## Types of changes

## Checklist:


packages/0x.js/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.36.1",
+        "changes": [
+            {
+                "note": "Remove truffle-hdwallet from artifact tests",
```
fabioberger: CHANGELOG's should actually only include changes that are relevant to users of our libraries. Usually things impacting the public interface or behavior of the library. Think this one doesn't qualify. Thoughts?

fabioberger: Yeah exactly... "Updated dependencies" is not ideal or even truthful in this case... I could be open to changing our conventions here, thoughts @LogvinovLeon?

dekz: I'm happy to change it to signal that something happened internally that is not relevant publicly. Also dropping the PR number?

PR #534, opened 2018-04-18
Add price information to wallet

BMillman19:
## Description

Modifies the wallet component to consume price information from the `/prices` endpoint and display it alongside token balance.

## Motivation and Context

N/A

## How Has This Been Tested?

```bash
cd packages/website
yarn dev
```

confirm wallet works as expected when `/prices` returns error

modify `BACKEND_BASE_URL` in `configs.ts` to `http://localhost:3000`

confirm wallet works as expected when `/prices` returns ok

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/wallet/wallet.tsx
+++ b/packages/website/ts/components/wallet/wallet.tsx
@@ -28,6 +28,7 @@ import { Dispatcher } from 'ts/redux/dispatcher';
 import {
     BalanceErrs,
     BlockchainErrs,
+    ItemByAddress,
     ProviderType,
     Side,
     Token,
@@ -35,6 +36,7 @@ import {
     TokenState,
     TokenStateByAddress,
 } from 'ts/types';
+import { configs } from 'ts/utils/configs';
 import { constants } from 'ts/utils/constants';
 import { utils } from 'ts/utils/utils';
 import { styles as walletItemStyles } from 'ts/utils/wallet_item_styles';
@@ -68,6 +70,11 @@ interface AllowanceToggleConfig {
 interface AccessoryItemConfig {
     wrappedEtherDirection?: Side;
     allowanceToggleConfig?: AllowanceToggleConfig;
+}
+
+interface WebsiteBackendPriceInfo {
+    price: string;
+    address: string;
 }
 
 const styles: Styles = {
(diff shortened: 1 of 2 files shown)


packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -161,13 +170,8 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
...
             const trackedTokenStateByAddress = this.state.trackedTokenStateByAddress;
-            _.each(newTokenAddresses, (tokenAddress: string) => {
-                trackedTokenStateByAddress[tokenAddress] = {
-                    balance: new BigNumber(0),
-                    allowance: new BigNumber(0),
-                    isLoaded: false,
-                };
-            });
+            const initialTrackedTokenStateByAddress = this._getInitialTrackedTokenStateByAddress(newTokenAddresses);
+            _.assign(trackedTokenStateByAddress, initialTrackedTokenStateByAddress);
```
fabioberger: This seems a bit of a round-about way of doing this. Why not simply:

```
const trackedTokenStateByAddress = initialTrackedTokenStateByAddress;
```

?

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -374,6 +387,16 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
         const result = `${formattedAmount} ${symbol}`;
         return <div style={styles.amountLabel}>{result}</div>;
     }
+    private _renderValue(amount: BigNumber, decimals: number, price?: BigNumber) {
```
fabioberger: Let's rename this to `_renderPrice`

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -434,41 +457,56 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
...
+    }
+    private async _getPricesByAddressAsync(tokenAddresses: string[]): Promise<ItemByAddress<BigNumber>> {
+        if (_.isEmpty(tokenAddresses)) {
+            return {};
+        }
+        const tokenAddressesQueryString = tokenAddresses.join(',');
+        const endpoint = `${configs.BACKEND_BASE_URL}/prices?tokens=${tokenAddressesQueryString}`;
+        const response = await fetch(endpoint);
+        if (response.status !== 200) {
+            return {};
```
fabioberger: Let's perhaps log more information about what went wrong to the console? If it was an error, let's also log to Rollbar.

BMillman19: Ok, I looked a bit more and we can improve / unify our error handling when requesting the `/wiki` an `/eth_gas_station` endpoints as well. Let's merge this as is and I'll follow up today with a unified object for hitting our backend

packages/website/ts/components/wallet/wallet.tsx
fabioberger: I see. Value is fine then, can't think of something better...

PR #540, opened 2018-04-19
Support Tuples returned from functions

dekz:
## Description

Allow abi-gen (through base-contract and contract_templates) to return tuples on function calls.

This moves us from ethers-contracts to ethers.js.

## Types of changes

## Checklist:


dekz:
There is probably a lot more we could clean up w.r.t delegating more features to ethers.js. I may punt on that work for now and pick it up in a later refactor. One thing that comes to mind is going up a level of abstraction and constructing an `ethers.Contract` rather than calling at the `Interface` and `Function` level. One blocker on this is the way ethers resolves ENS names, its a required function on all providers. 

I'll at least ping an issue on Ethers.js to re-consider that function on the provider and perhaps implement it as a non-standard `eth_resolveName` call.

packages/contract_templates/contract.handlebars
```diff
@@ -8,7 +8,7 @@ import { BaseContract } from '@0xproject/base-contract';
 import { BlockParam, BlockParamLiteral, CallData, ContractAbi, DataItem, MethodAbi, Provider, TxData, TxDataPayable } from '@0xproject/types';
 import { BigNumber, classUtils, promisify } from '@0xproject/utils';
 import { Web3Wrapper } from '@0xproject/web3-wrapper';
-import * as ethersContracts from 'ethers-contracts';
+import * as ethers from 'ethers'
```
LogvinovLeon: Semicolon

packages/typescript-typings/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "0.3.0",
+        "changes": [
+            {
+                "note": "Add types for `ethers.js`",
```
LogvinovLeon: Remove types for ethers-contracts

PR #541, opened 2018-04-19
Improve Monorepo-scripts

fabioberger:
## Description

This PR:
- Adds a new script that removes all the git tags created by a failed `lerna publish`. Unfortunately this command is not idempotent. 
- Adds checks to make sure a dev running our publish script has the requisite local setup to do so successfully. 
- Adds descriptions about each available script to the README.

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:

### DIFF
--- a/packages/monorepo-scripts/README.md
+++ b/packages/monorepo-scripts/README.md
@@ -4,7 +4,7 @@ This repository contains a few helpful scripts for working with this mono repo.
 
 #### Scripts
 
-**`yarn deps_versions`**: Since we use Lerna + yarn workspaces, shared dependencies between packages in the monorepo get hoisted to a top-level `node_modules` directory. If two packages use different versions of the same dependency however, both get installed. To avoid having many versions of a dependency installed, we try to keep dependency versions the same across packages in the monorepo. This script will list any dependencies for which we have multiple versions installed. We can then go through them and try to consolidate to a single version where possible.
+**`yarn deps_versions`**: Since we use Lerna + Yarn workspaces, shared dependencies between packages in the monorepo get hoisted to a top-level `node_modules` directory. If two packages use different versions of the same dependency however, both get installed. To avoid having many versions of a dependency installed, we try to keep dependency versions the same across packages in the monorepo. This script will list any dependencies for which we have multiple versions installed. We can then go through them and try to consolidate to a single version where possible.
 
 **`yarn find_unused_deps`**: Sometimes we accidentally leave dependencies listed in `package.json` that are no longer being used. This script finds potential dependencies that might no longer be in use. Please verify that it is no longer in use before removing, the `depcheck` package we use under-the-hood doesn't handle some TS quirks perfectly.
 
--- a/packages/monorepo-scripts/README.md
+++ b/packages/monorepo-scripts/README.md
@@ -2,6 +2,16 @@
 
 This repository contains a few helpful scripts for working with this mono repo.
 
+#### Scripts
+
+**`yarn deps_versions`**: Since we use Lerna + yarn workspaces, shared dependencies between packages in the monorepo get hoisted to a top-level `node_modules` directory. If two packages use different versions of the same dependency however, both get installed. To avoid having many versions of a dependency installed, we try to keep dependency versions the same across packages in the monorepo. This script will list any dependencies for which we have multiple versions installed. We can then go through them and try to consolidate to a single version where possible.
+
+**`yarn find_unused_deps`**: Sometimes we accidentally leave dependencies listed in `package.json` that are no longer being used. This script finds potential dependencies that might no longer be in use. Please verify that it is no longer in use before removing, the `depcheck` package we use under-the-hood doesn't handle some TS quirks perfectly.
+
+**`yarn remove_tags`**: Our publishing script calls `lerna publish` under-the-hood. If this command fails, it might have created new versioned git tags for each package. Removing these manually is tedious, so you can also run this command instead. Before doing so, check to see if `lerna` already created the publish commit. If so, first revert that with `git reset --hard HEAD~1`, then run this command.
+
+**`yarn test:publish`**: Execute a test-run of the publish script. This dry run won't actually publish, nor will it commit/push anything to Github.
+
 ## Usage
 
 #### Dependency versions
--- a/packages/monorepo-scripts/src/constants.ts
+++ b/packages/monorepo-scripts/src/constants.ts
@@ -4,4 +4,5 @@ export const constants = {
     monorepoRootPath: path.join(__dirname, '../../..'),
     stagingWebsite: 'http://staging-0xproject.s3-website-us-east-1.amazonaws.com',
     lernaExecutable: './node_modules/lerna/bin/lerna.js',
+    githubPersonalAccessToken: process.env.GITHUB_PERSONAL_ACCESS_TOKEN_0X_JS,
 };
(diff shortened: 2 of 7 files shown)


packages/monorepo-scripts/src/constants.ts
```diff
@@ -3,4 +3,5 @@ import * as path from 'path';
 export const constants = {
     monorepoRootPath: path.join(__dirname, '../../..'),
     stagingWebsite: 'http://staging-0xproject.s3-website-us-east-1.amazonaws.com',
+    lernaExecutable: './node_modules/lerna/bin/lerna.js',
```
dekz: Can this be updated to use path module for future windows support :P

packages/monorepo-scripts/src/publish.ts
```diff
@@ -107,30 +113,65 @@ package.ts. Please add an entry for it and try again.`,
     }
 }
 
+async function checkPublishRequiredSetupAsync(): Promise<boolean> {
+    // check to see if logged into npm before publishing
+    try {
+        await execAsync(`sudo npm whoami`);
```
dekz: Why the `sudo` here?

packages/monorepo-scripts/src/publish.ts
```diff
@@ -107,30 +113,65 @@ package.ts. Please add an entry for it and try again.`,
...
+    } catch (err) {
+        utils.log('You must be logged into npm in the commandline to publish. Run `npm login` and try again.');
+        return false;
+    }
+
+    // Check to see if Git personal token setup
+    if (_.isUndefined(process.env.GITHUB_PERSONAL_ACCESS_TOKEN_0X_JS)) {
```
dekz: Can we consolidate this and the githubPersonalAccessToken in postpublish_utils so that the environment variable exists in one place.

packages/monorepo-scripts/src/publish.ts
dekz: Why `sudo`?

packages/monorepo-scripts/src/constants.ts
dekz: This can be platform agnostic by using `path` module

packages/monorepo-scripts/src/publish.ts
```diff
@@ -38,8 +38,14 @@ const packageNameToWebsitePath: { [name: string]: string } = {
 };
 
 (async () => {
+    const hasRequiredSetup = await checkPublishRequiredSetupAsync();
+    if (!hasRequiredSetup) {
+        return; // abort
```
LogvinovLeon: Wouldn't it be better to throw a helpfull error message then to just silently fail?

packages/monorepo-scripts/README.md
```diff
@@ -2,6 +2,16 @@
 
 This repository contains a few helpful scripts for working with this mono repo.
 
+#### Scripts
+
+**`yarn deps_versions`**: Since we use Lerna + yarn workspaces, shared dependencies between packages in the monorepo get hoisted to a top-level `node_modules` directory. If two packages use different versions of the same dependency however, both get installed. To avoid having many versions of a dependency installed, we try to keep dependency versions the same across packages in the monorepo. This script will list any dependencies for which we have multiple versions installed. We can then go through them and try to consolidate to a single version where possible.
```
BMillman19: nit: should we capitalize "yarn" as well? just looking at the yarn website looks like they capitalize it

packages/monorepo-scripts/README.md
```diff
@@ -2,6 +2,16 @@
...
+
+**`yarn find_unused_deps`**: Sometimes we accidentally leave dependencies listed in `package.json` that are no longer being used. This script finds potential dependencies that might no longer be in use. Please verify that it is no longer in use before removing, the `depcheck` package we use under-the-hood doesn't handle some TS quirks perfectly.
+
+**`yarn remove_tags`**: Our publishing script calls `lerna publish` under-the-hood. If this command fails, it might have created new versioned git tags for each package. Removing these manually is tedious, so you can also run this command instead. Before doing so, check to see if `lerna` already created the publish commit. If so, first revert that with `git reset --hard HEAD~1`, then run this command.
```
BMillman19: do we need to `git push origin development -f` after doing the hard reset? I don't remember if the publish commit automatically gets pushed

packages/monorepo-scripts/src/utils.ts
```diff
@@ -17,4 +22,44 @@ export const utils = {
             cwd,
         });
     },
+    async getUpdatedLernaPackagesAsync(shouldIncludePrivate: boolean): Promise<LernaPackage[]> {
+        const updatedPublicPackages = await this.getLernaUpdatedPackagesAsync(shouldIncludePrivate);
+        const updatedPackageNames = _.map(updatedPublicPackages, pkg => pkg.name);
+
```
BMillman19: how come this extra newline didn't get deleted by prettier?

packages/monorepo-scripts/src/utils.ts
```diff
@@ -17,4 +22,44 @@ export const utils = {
...
+        });
+        const updatedPackages = JSON.parse(result.stdout);
+        if (!shouldIncludePrivate) {
+            const updatedPublicPackages = _.filter(updatedPackages, updatedPackage => !updatedPackage.private);
+            return updatedPublicPackages;
+        }
+        return updatedPackages;
+    },
+    getChangelogJSONIfExists(changelogPath: string) {
+        let changelogJSON: string;
```
BMillman19: we don't need to declare `changelogJSON` outside the scope of the try block

packages/monorepo-scripts/src/utils.ts
```diff
@@ -17,4 +22,44 @@ export const utils = {
...
+            return changelogJSON;
+        } catch (err) {
+            return undefined;
+        }
+    },
+    getChangelogJSONOrCreateIfMissing(changelogPath: string): string {
+        const changelogIfExists = this.getChangelogJSONIfExists(changelogPath);
+        if (_.isUndefined(changelogIfExists)) {
+            // If none exists, create new, empty one.
+            const emptyChangelogJSON = JSON.stringify([], null, 4);
```
BMillman19: assigning 4 to a named variable will help with signaling the intent here

packages/monorepo-scripts/src/publish.ts
fabioberger: Without it, the `npm whoami` won't pick up the users credentials... I'm guessing running `npm login` runs in the sudo userspace? Not entirely sure though.

packages/monorepo-scripts/src/publish.ts
fabioberger: We print helpful messages for each conditions within `checkPublishRequiredSetupAsync`. Do you still think this is necessary?

packages/monorepo-scripts/README.md
fabioberger: Nope, it does not push to Github **until** the end of the publish command. If it breaks before then, it only created the commit locally.

packages/monorepo-scripts/src/publish.ts
dekz: Is that environment specific and dependent on the way npm was installed and logged in? We don't run the further publish tasks as `sudo` so it seems a bit strange. My `npm whoami` returns correctly.

packages/monorepo-scripts/src/utils.ts
fabioberger: Newlines within method bodies are allowed no?

packages/monorepo-scripts/src/utils.ts
fabioberger: Let's just remove that. It doesn't do anything for an empty array.

packages/monorepo-scripts/src/publish.ts
fabioberger: Not sure why it won't work without sudo. I checked and the script is running in the same user space as my command line. Will add a hack comment and punt on this for now.

PR #543, opened 2018-04-19
Integrate relayer index with website backend

BMillman19:
## Description

Modifies the relayer_index component to consume relayer information from the `/relayers` endpoint.

## Motivation and Context

N/A

## How Has This Been Tested?

in the `website` package, modify `BACKEND_BASE_URL` in `configs.ts` to `http://localhost:3000`
then build and run the website:

```bash
cd packages/website
yarn dev
```

in a separate terminal, build and run `website-backend`

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/relayer_index/relayer_index.tsx
+++ b/packages/website/ts/components/relayer_index/relayer_index.tsx
@@ -1,12 +1,19 @@
 import { colors, Styles } from '@0xproject/react-shared';
+import * as _ from 'lodash';
 import { GridList } from 'material-ui/GridList';
 import * as React from 'react';
 
 import { RelayerGridTile } from 'ts/components/relayer_index/relayer_grid_tile';
-import { RelayerInfo } from 'ts/types';
+import { WebsiteBackendRelayerInfo } from 'ts/types';
+import { backendClient } from 'ts/utils/backend_client';
 
 export interface RelayerIndexProps {
     networkId: number;
+}
+
+interface RelayerIndexState {
+    relayerInfos?: WebsiteBackendRelayerInfo[];
+    error?: Error;
 }
 
 const styles: Styles = {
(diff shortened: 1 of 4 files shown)


packages/website/ts/components/relayer_index/relayer_index.tsx
```diff
@@ -25,94 +32,65 @@ const styles: Styles = {
...
+                    >
+                        {this.state.relayerInfos.map((relayerInfo: WebsiteBackendRelayerInfo) => (
+                            <RelayerGridTile
+                                key={relayerInfo.id}
+                                relayerInfo={relayerInfo}
+                                networkId={this.props.networkId}
+                            />
+                        ))}
+                    </GridList>
+                </div>
```
fabioberger: We could add a spinner here for the loading state. For the error state we should also show the user a helpful error message. Perhaps to refresh the page?

Even if you plan to do this in a later PR before make this a two part conditional with a TODO for the `!readyToRender` case?

packages/website/ts/components/relayer_index/relayer_index.tsx
```diff
@@ -25,94 +32,65 @@ const styles: Styles = {
...
+            const relayerInfos = await backendClient.getRelayerInfosAsync();
+            if (!this._isUnmounted) {
+                this.setState({
+                    relayerInfos,
+                });
+            }
+        } catch (error) {
+            if (!this._isUnmounted) {
+                this.setState({
+                    error,
```
fabioberger: Do we want to set the entire error message as the state? Maybe just an `errMsg` that could be displayed? My reasoning is that it probably makes more sense to convert Error objects into error messages suitable for this component in one place, rather then in a bunch of places.

BMillman19: I think I like keeping the entire error as the state and keeping the display logic in the render method when its time to display to the user

PR #548, opened 2018-04-22
Cleanup tests

abandeali1:
## Description

- Adds `ERC20Wrapper` and `ERC721Wrapper` classes, which include utilities for deploying tokens, deploying their respective proxies, setting balances/allowances, and querying balances
- Removes unused dependencies and types

## Motivation and Context

- There was previously a lot of code reuse in the `before` hooks of each test - this quickly became very difficult to read and made it difficult to modify the tests.

## Types of changes

*   [ ] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

*   [ ] Change requires a change to the documentation.
*   [ ] Added tests to cover my changes.
*   [ ] Added new entries to the relevant CHANGELOG.jsons.
*   [ ] Labeled this PR with the 'WIP' label if it is a work in progress.
*   [ ] Labeled this PR with the labels corresponding to the changed package.

PR #549, opened 2018-04-23
Merge Exchange and AssetProxyDispatcher (depends on #548)

abandeali1:
## Description

- Merges the Exchange contract with the AssetProxyDispatcher contract

## Motivation and Context
- This saves ~10K gas per fill (need to double check on Geth/Parity)
- Most of the benefit of having a separate AssetProxyDispatcher contract is that only one transaction is required to add an authorized address for all proxies. However, this can also be accomplished with an `owner` contract that can make batch calls of `addAuthorizedAddress`. While this might not scale with a large amount of proxies, it seems well worth the trade-off in the short term. The AssetProxyDispatcher can always be added into the system later if necessary.

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/test/TestAssetProxyDispatcher/TestAssetProxyDispatcher.sol
```diff
@@ -0,0 +1,33 @@
...
+contract TestAssetProxyDispatcher is MixinAssetProxyDispatcher {
+    function publicDispatchTransferFrom(
+        bytes memory assetMetadata,
+        address from,
+        address to,
+        uint256 amount)
+        public
+    {
+        dispatchTransferFrom(assetMetadata, from, to, amount);
+    }
```
fabioberger: Why do we need this? Isn't `MixinAssetProxyDispatcher` going to be part of the Exchange contract?

packages/contracts/test/exchange/core.ts
```diff
@@ -467,35 +445,6 @@ describe('Exchange core', () => {
...
-            expect(signedOrder.feeRecipientAddress).to.be.equal(logArgs.feeRecipientAddress);
-            expect(signedOrder.makerAssetData).to.be.equal(logArgs.makerAssetData);
-            expect(signedOrder.takerAssetData).to.be.equal(logArgs.takerAssetData);
-            expect(expectedFilledMakerAssetAmount).to.be.bignumber.equal(logArgs.makerAssetFilledAmount);
-            expect(expectedFilledTakerAssetAmount).to.be.bignumber.equal(logArgs.takerAssetFilledAmount);
-            expect(expectedFeeMPaid).to.be.bignumber.equal(logArgs.makerFeePaid);
-            expect(expectedFeeTPaid).to.be.bignumber.equal(logArgs.takerFeePaid);
-            expect(orderUtils.getOrderHashHex(signedOrder)).to.be.equal(logArgs.orderHash);
-        });
-
```
fabioberger: What do we now expect if there is no feeRecipient?

packages/contracts/test/exchange/dispatcher.ts
```diff
@@ -258,37 +254,12 @@ describe('AssetProxyDispatcher', () => {
...
-                    encodedProxyMetadata,
-                    makerAddress,
-                    takerAddress,
-                    amount,
-                    { from: exchangeAddress },
-                ),
-            ).to.be.rejectedWith(constants.REVERT);
-        });
-
-        it('should throw on transfer if requesting address is not authorized', async () => {
```
fabioberger: Why was this test removed?

packages/contracts/package.json
```diff
@@ -30,9 +30,9 @@
     },
     "config": {
         "abis":
-            "../migrations/src/artifacts/@(DummyERC20Token|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken|AssetProxyDispatcher|ERC20Proxy|ERC721Proxy|DummyERC721Token|LibBytes).json",
+            "../migrations/src/artifacts/@(DummyERC20Token|Exchange|TokenRegistry|MultiSigWallet|MultiSigWalletWithTimeLock|MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress|TokenRegistry|ZRXToken|TestAssetProxyDispatcher|ERC20Proxy|ERC721Proxy|DummyERC721Token|LibBytes|Authorizable).json",
```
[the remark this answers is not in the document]
abandeali1: I added it so that we can test `dispatchTransferFrom`, which is an internal function. `TestAssetProxyDispatcher` just wraps `dispatchTransferFrom` with a public function.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlement.sol
```diff
@@ -66,38 +57,32 @@ contract MixinSettlementProxy is
...
-                    order.feeRecipientAddress,
-                    takerFeePaid
-                );
-            }
-        }
+        makerFeePaid = getPartialAmount(takerAssetFilledAmount, order.takerAssetAmount, order.makerFee);
+        dispatchTransferFrom(
+            ZRX_PROXY_DATA,
+            order.makerAddress,
+            order.feeRecipientAddress,
```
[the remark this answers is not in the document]
abandeali1: We discussed this a bit and thought it felt a little arbitrary to make this check. What if someone wants to burn ZRX by sending it to `address(0)`?

packages/contracts/src/contracts/current/test/TestAssetProxyDispatcher/TestAssetProxyDispatcher.sol
abandeali1: See comment above - we can't unit test this without wrapping it with a public function.

packages/contracts/test/exchange/dispatcher.ts
abandeali1: There is no longer such a thing as an address that is authorized to call `dispatchTransferFrom`, it is simply an internal function within the Exchange contract.

packages/contracts/test/exchange/core.ts
abandeali1: Just a transfer to `address(0)`.

PR #550, opened 2018-04-23
Grab wallet price information by token symbol

BMillman19:
## Description

Query our backend service by symbol instead of token address

## Motivation and Context

Coin market cap API does not have address information

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/wallet/wallet.tsx
+++ b/packages/website/ts/components/wallet/wallet.tsx
@@ -526,10 +526,9 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
             ),
         );
         const joinedTokenSymbols = _.keys(tokenAddressBySymbol).join(',');
-        const quoteCurrency = 'USD';
         const queryParams = {
             fsyms: joinedTokenSymbols,
-            tsyms: quoteCurrency,
+            tsyms: configs.FIAT_QUOTE_CURRENCY_SYMBOL,
         };
         try {
             this._cryptoCompareLastFetchTimestampMs = Date.now();
@@ -542,7 +541,7 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
                 _.get(tokenAddressBySymbol, symbol),
             );
             const result = _.mapValues(priceInfoByAddress, priceInfo => {
-                const price = _.get(priceInfo, quoteCurrency);
+                const price = _.get(priceInfo, configs.FIAT_QUOTE_CURRENCY_SYMBOL);
                 const priceBigNumber = new BigNumber(price);
                 return priceBigNumber;
             });
--- a/packages/website/ts/components/wallet/wallet.tsx
+++ b/packages/website/ts/components/wallet/wallet.tsx
@@ -6,7 +6,7 @@ import {
     Styles,
     utils as sharedUtils,
 } from '@0xproject/react-shared';
-import { BigNumber } from '@0xproject/utils';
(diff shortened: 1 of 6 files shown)


packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -128,9 +130,13 @@ const FOOTER_ITEM_KEY = 'FOOTER';
 const DISCONNECTED_ITEM_KEY = 'DISCONNECTED';
 const ETHER_ITEM_KEY = 'ETHER';
 const USD_DECIMAL_PLACES = 2;
+const CRYPTO_COMPARE_MULTI_ENDPOINT = '/pricemulti';
+// Crypto compare recommends requesting no more than once every 10s: https://www.cryptocompare.com/api/?javascript#requests
+const CRYPTO_COMPARE_UPDATE_INTERVAL_MS = 10 * 1000;
```
fabioberger: I feel like we should fetch prices once every 10sec on `website-backend` and cache it there. With the current approach, every user will be pinging CryptoCompare directly, which is a lot more load for them. 

We did something similar to this with Coinmarketcap previously, we might be able to revive some of that old code: https://github.com/0xProject/website-backend/commit/f594e086c8c219475b8cb8f304441986898242af
Looks like it's similar to how we cache gas prices from `ethGasStation`.

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -491,16 +508,56 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
...
+                        const symbol = tokenIfExists.symbol;
+                        // the crypto compare api doesn't understand 'WETH' so we need to replace it with 'ETH'
+                        const key = symbol === ETHER_TOKEN_SYMBOL ? ETHER_SYMBOL : symbol;
+                        return [key, address];
+                    } else {
+                        return undefined;
+                    }
+                }),
+            ),
+        );
```
fabioberger: I think it would be simpler to write this as:

```
const tokenAddressBySymbol: {[symbol: string]: string} = {};
_.each(tokenAddresses, address => {
    const tokenIfExists = _.get(this.props.tokenByAddress, address);
    if (!_.isUndefined(tokenIfExists)) {
        const symbol = tokenIfExists.symbol;
        // the crypto compare api doesn't understand 'WETH' so we need to replace it with 'ETH'
       const key = symbol === ETHER_TOKEN_SYMBOL ? ETHER_SYMBOL : symbol;
       tokenAddressBySymbol[key] = address;
    }
});
```

As much as possible, let's try and avoid using more non-standard Lodash methods (`fromPairs`, `compact`, etc...) as they can reduce code readability.

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -491,16 +508,56 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
...
+        const joinedTokenSymbols = _.keys(tokenAddressBySymbol).join(',');
+        const quoteCurrency = 'USD';
```
fabioberger: Let's move this to `config.ts`. I could see us making this configurable at a later date and it would be best to only store this setting in one place.

packages/website/ts/components/wallet/wallet.tsx
BMillman19: Its a lot more complicated than that because each user has different tokens they are tracking

PR #561, opened 2018-04-26
Add Validator signature type

abandeali1:
## Description

`Contract` type:
- A multisig that can create orders with X amount of signatures from its owners
- A contract that allows multiple addresses to create orders with the contract's balances

`Validator` type:
- An EOA wants to sign orders using RSA, BLS, Schnorr, etc.
- Originally, we had a TxOrigin signature type, but this can now be done in an external Validator contract instead

- Adds a basic whitelist filter implementation

## Motivation and Context

This signature type works very well with sender abstraction. You can now route fills through contracts and only require a single signature for the Ethereum transaction (rather than 1 for the Ethereum tx and 1 for the 0x tx)

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,58 @@
...
+import "../../utils/Ownable/Ownable.sol";
+
+contract Whitelist is Ownable {
+
+    mapping (address => bool) public isWhitelisted;
+    MTransactions EXCHANGE;
+
+    bytes txOriginSignatureType = new bytes(1);
+
+    function Whitelist(address _exchange)
```
willwarren89: Is `constructor()` syntax supported in this version of Solidity?

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    function updateWhitelistStatus(address target, bool isApproved)
+        external
+        onlyOwner
+    {
+        isWhitelisted[target] = isApproved;
+    }
+
+    function fillOrderIfWhitelisted(
```
hysz: Could be good to add function descriptions. We've been including them in the newer contracts.
An example for this function:
```
/// @dev Fills the input order using msg.sender as the order taker.
         The sender must be both whitelisted and the originating address of this transaction.
         This function throws if the sender does not meet these requirements or the order cannot be filled.
/// @param order Order struct containing order specifications.
/// @param takerAssetFillAmount Desired amount of takerToken to sell.
/// @param salt Arbitrary number to ensure uniqueness of transaction.
/// @param orderSignature Proof that order has been created by maker.
```

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,80 @@
...
+        LibOrder.Order memory order,
+        uint256 takerAssetFillAmount,
+        uint256 salt,
```
hysz: What happens if a salt is reused?

packages/contracts/test/exchange/transactions.ts
```diff
@@ -202,4 +205,88 @@ describe('Exchange transactions', () => {
...
+            };
+            whitelistOrderFactory = new OrderFactory(makerPrivateKey, defaultOrderParams);
+        });
+
+        beforeEach(async () => {
+            signedOrder = whitelistOrderFactory.newSignedOrder();
+            erc20Balances = await erc20Wrapper.getBalancesAsync();
+        });
+
+        it('should revert if taker has not been whitelisted', async () => {
```
hysz: Could be good to test revert if `msg.sender` is not the originating address.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,80 @@
...
+    Exchange EXCHANGE;
+
+    bytes txOriginSignature = new bytes(1);
+    bytes4 fillOrderFunctionSelector;
+
+    function Whitelist(address _exchange)
+        public
+    {
+        EXCHANGE = Exchange(_exchange);
+        txOriginSignature[0] = 0x04;
```
hysz: I'm not sure what `0x04` means. Could we add a comment?

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,80 @@
...
+import "../../protocol/Exchange/Exchange.sol";
+import "../../protocol/Exchange/LibOrder.sol";
+import "../../utils/Ownable/Ownable.sol";
+
+contract Whitelist is Ownable {
+
+    mapping (address => bool) public isWhitelisted;
+    Exchange EXCHANGE;
+
+    bytes txOriginSignature = new bytes(1);
```
hysz: Could we add brief descriptions for these?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -28,12 +28,13 @@ contract MixinSignatureValidator is MSignatureValidator {
...
-        Ecrecover,
         EIP712,
-        Trezor,
+        Ecrecover,
+        TxOrigin,
+        Caller,
         Contract,
-        PreSigned
+        PreSigned,
+        Trezor
```
fabioberger: Why were these re-ordered?

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
fabioberger: I'd prefer we make this a named constant if possible.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    function updateWhitelistStatus(address target, bool isApproved)
+        external
+        onlyOwner
+    {
+        isWhitelisted[target] = isApproved;
+    }
+
+    function fillOrderIfWhitelisted(
```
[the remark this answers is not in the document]
abandeali1: Will add some comments - seems like a lot of people will want to use this as a reference.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
abandeali1: It's just there to prevent collisions in the txHash, it can generally be set to 0 unless someone is submitting the exact same tx with the exact same orders.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
abandeali1: I ordered them by how frequently I think they will be used.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -131,20 +128,48 @@ contract MixinSignatureValidator is
...
-                signature.length == 66,
+                signature.length == 1,
                 INVALID_SIGNATURE_LENGTH
             );
-            v = uint8(signature[1]);
-            r = readBytes32(signature, 2);
-            s = readBytes32(signature, 34);
-            recovered = ecrecover(hash, v, r, s);
-            isValid = signer == recovered;
+            isValid = signer == tx.origin;
```
Recmo: `tx.origin` is very tempting for certain problems, but it is filled with security issues. Basically, the moment you use it for transaction verification, you open the door to phishing attacks by other contracts. Solidity devs agree you should not use it for authentication (like this PR aims to do).

In fact, it would have been removed if it wasn't for backwards compatibility:
https://github.com/ethereum/solidity/issues/683

It has also been added as a warning in static analysis:
https://github.com/ethereum/remix-ide/issues/173

IMHO the security risks (and public perception) are not worth the gain in user experience. I also think auditors would and should raise this as an issue.

Can we revisit the problem this is trying to address and come up with a different solution?
Perhaps the `SignatureType.Presigned` can be used. Or we can lobby for batch transactions in Ethereum. Or for batch signing in the UI clients.

PhABC: In my opinion, using `tx.origin` is fine when you use signature in the case of 0x. tx.origin is terrible when it's used as ownership for instance, since that's really easy to fish users for and then call the owned contract. In general,  Early alarms scared people away from this approach, but I think tx.origin can be powerful if all corner cases are taken into consideration.

In the present proposal, a malicious contract would need to 

1. Fool the maker to call their contract
2. Pass in a valid signed order from the maker that specified signature type as `SignatureType.TxOrigin`
3. Use this valid signature for a malicious purpose ...

For 3, the only thing I can see is that the maker order could be filled by the malicious contract, which doesn't seem to bad. What else could happen?

abandeali1: There are definitely some risks here (somewhat undecided on if this should make the cut). For example:

- You call some function on a contract that calls a chain of other contracts. Somewhere along the way, a contract calls fillOrder using the TxOrigin signature type, making you unintentionally fill a malicious signed order.
- In a similar vein, the contract could call fillOrder on a malicious _unsigned_ order but pass in a TxOrigin signature for that order. This pretty much has the same result as the previous scenario.

What makes this tricky is that this could be hidden in a long chain of contract calls, making it hard to audit beforehand. If signers like MetaMask showed you how a transaction would change the state, the risk of this happening could decrease a lot.

PhABC: Oh, I see. I just realized that there is no "0x" signature required from tx.origin, apologies. Yeah, this can definitely lead to bad stuff.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,121 @@
...
+        // Check if maker is on the whitelist
+        require(
+            isWhitelisted[order.makerAddress],
+            ADDRESS_NOT_WHITELISTED
+        );
+
+        // Check if taker is on the whitelist.
+        require(
+            isWhitelisted[takerAddress],
+            ADDRESS_NOT_WHITELISTED
```
fabioberger: Let's have an error message that specifies which address is not whitelisted. Currently one isn't sure if it's the maker or taker address that is not whitelisted.

packages/contracts/package.json
```diff
@@ -12,15 +12,16 @@
     },
     "scripts": {
         "watch": "tsc -w",
-        "prebuild": "run-s clean copy_artifacts generate_contract_wrappers",
+        "prebuild": "run-s clean compile copy_artifacts generate_contract_wrappers",
```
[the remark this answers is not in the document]
abandeali1: The artifacts we copy over usually end up out of date if we don't compile first.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -131,20 +147,55 @@ contract MixinSignatureValidator is
...
-            isValid = signer == recovered;
+            isValid = signer == msg.sender;
             return isValid;
 
-        // Signature from Trezor hardware wallet
+        // Signature verified by signer contract.
+        // If used with an order, the maker of the order is the signer contract.
+        } else if (signatureType == SignatureType.Signer) {
+            isValid = ISigner(signer).isValidSignature(hash, signature);
+            return isValid;
```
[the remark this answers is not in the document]
abandeali1: Pretty much any of our signature types could be implemented in a validator, but I think it's still worth keeping the most used ones as defaults for the sake of efficiency. 

One other question I'm thinking about is how to prevent people from removing their validator's approval as a griefing mechanism. For this reason, I think it's still preferable that we keep as many defaults as possible.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,128 @@
...
+    /// @param orderSignature Proof that order has been created by maker.
+    function fillOrderIfWhitelisted(
+        LibOrder.Order memory order,
+        uint256 takerAssetFillAmount,
+        uint256 salt,
+        bytes memory orderSignature
+    )
+        public
+    {
+        address takerAddress = msg.sender;
```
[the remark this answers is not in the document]
abandeali1: Check the next line xD

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
```diff
@@ -0,0 +1,128 @@
...
+    function isValidSignature(
+        bytes32 hash,
+        address signer,
+        bytes signature
+    )
+        external
+        view
+        returns (bool isValid)
+    {
+        return signer == tx.origin;
```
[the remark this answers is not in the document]
abandeali1: These are both interesting ideas - need to think about it!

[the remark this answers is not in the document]
abandeali1: One more idea - when approving a validator, we an add a `sender` parameter. So `approvedValidators` would store signer => sender => validator, and then we check if `approvedValidators[signer][msg.sender][validator] == true`. Thoughts @Recmo ?

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSignatureValidator.sol
```diff
@@ -25,14 +25,15 @@ contract MSignatureValidator is
...
-        EIP712,
-        Trezor,
-        Contract,
-        PreSigned
+        Illegal,    // 0x00, default value
+        Invalid,    // 0x01
+        EIP712,     // 0x02
+        Ecrecover,  // 0x03
+        Caller,     // 0x04
+        Signer,     // 0x05
```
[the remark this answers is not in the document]
abandeali1: What about `Wallet`?

packages/dev-utils/src/web3_factory.ts
```diff
@@ -41,6 +41,7 @@ export const web3Factory = {
             }
             provider.addProvider(
                 new GanacheSubprovider({
+                    gasLimit: constants.GAS_ESTIMATE,
```
fabioberger: Let's rename `GAS_ESTIMATE` to `GAS_LIMIT`. `estimate` can also refer to gasAmount.

packages/contracts/test/zrx_token.ts
```diff
@@ -21,12 +21,6 @@ describe('ZRXToken', () => {
     let MAX_UINT: BigNumber;
     let zrxToken: ZRXTokenContract;
 
-    before(async () => {
-        await blockchainLifecycle.startAsync();
-    });
-    after(async () => {
-        await blockchainLifecycle.revertAsync();
-    });
```
fabioberger: I think these are unsucessful merges. We want these in here. Ditto in `unlimited_allowance_token.ts` and `token_registry.ts`.

packages/contracts/test/libraries/lib_bytes.ts
```diff
@@ -32,12 +33,6 @@ describe('LibBytes', () => {
     const testBytes32 = '0x102030405060708090a0b0c0d0e0f0102030405060708090a0b0c0d0e0f01020';
     const testUint256 = new BigNumber(testBytes32, 16);
 
-    before(async () => {
-        await blockchainLifecycle.startAsync();
-    });
-    after(async () => {
-        await blockchainLifecycle.revertAsync();
-    });
```
fabioberger: Re-add these.

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -54,12 +54,6 @@ describe('Exchange wrappers', () => {
     let defaultMakerAssetAddress: string;
     let defaultTakerAssetAddress: string;
 
-    before(async () => {
-        await blockchainLifecycle.startAsync();
-    });
-    after(async () => {
-        await blockchainLifecycle.revertAsync();
-    });
```
fabioberger: Re-add these.

packages/contracts/test/exchange/transactions.ts
```diff
@@ -55,13 +57,9 @@ describe('Exchange transactions', () => {
...
     let defaultTakerTokenAddress: string;
+    let makerPrivateKey: Buffer;
+    let takerPrivateKey: Buffer;
 
-    before(async () => {
-        await blockchainLifecycle.startAsync();
-    });
-    after(async () => {
-        await blockchainLifecycle.revertAsync();
-    });
```
fabioberger: Re-add these.

fabioberger: And everywhere else these we removed. Many places.

packages/contracts/src/contracts/current/protocol/AssetProxy/mixins/MAuthorizable.sol
```diff
@@ -38,5 +38,5 @@ contract MAuthorizable is
     );
 
     /// @dev Only authorized addresses can invoke functions with this modifier.
-    modifier onlyAuthorized { _; }
+    modifier onlyAuthorized { revert(); _; }
```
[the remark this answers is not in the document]
abandeali1: This is meant to be overwritten, but we added a revert just in case it is not.

packages/contracts/src/contracts/current/test/Whitelist/Whitelist.sol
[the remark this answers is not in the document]
Recmo: That one is interesting too. But then you always need to explicitly approve a sender.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -49,6 +56,19 @@ contract MixinSignatureValidator is
         preSigned[hash][signer] = true;
     }
 
+    /// @dev Approves a Validator contract to verify signatures on signer's behalf.
+    /// @param validator Address of Validator contract.
+    /// @param approval Approval or disapproval of  Validator contract.
+    function approveSignatureValidator(
```
hysz: Should there be a way to "unapprove" a validator?

abandeali1: Naming isn't super clear here, but you can pass in `false` to disapprove a validator.

PR #562, opened 2018-04-26
Remove taker address, makerAssetData, and takerAssetData from events

abandeali1:
## Description

Removes `takerAddress`, `makerAssetData`, and `takerAssetData` from `Fill` and `Cancel` events. The only remaining fields are those that are indexed and are not available in the calldata of the functions.

## Motivation and Context

Removing these fields saves ~3000 gas per fill, and also frees up stack variables (allowing us to remove the `emitFillEvent` function.

This might be somewhat controversial. Users will definitely want to access this removed data in a lot of cases, and this adds an extra step. The new flow will look like:

1. See event of interest
2. Lookup tx + parse calldata

Why I think adding this extra step doesn't matter much:
- In most cases, users should be watching transaction as well as events. Since they will typically see the tx before the event occurs, this step doesn't actually add any extra work.
- If there are any other fields that the user is interested in that weren't already contained in the event logs, they would have to lookup the tx anyways.
- Speed isn't super important for historical data. 
- We can pretty easily abstract this in our dev tools.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
@@ -53,23 +53,18 @@ contract MixinExchangeCore is
 
     event Fill(
         address indexed makerAddress,
-        address takerAddress,
         address indexed feeRecipientAddress,
+        bytes32 indexed orderHash,
         uint256 makerAssetFilledAmount,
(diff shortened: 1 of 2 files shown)


fabioberger:
LGTM once tests pass. Do some of our tests check for these values in the logged events?

abandeali1:
We decided against this until we have an efficient State Watcher implementation. Closing this for now.

PR #563, opened 2018-04-26
Add hard coded proxyId into each AssetProxy

abandeali1:
## Description

- Hard codes a proxyId into each AssetProxy
- Check that the metadata being passed in is intended for the receiving proxy
- Check the that ids match when registering a new proxy, unless the proxy is being removed

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinAssetProxyDispatcher.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinAssetProxyDispatcher.sol
@@ -54,8 +54,7 @@ contract MixinAssetProxyDispatcher is
     }
 
     /// @dev Registers an asset proxy to an asset proxy id.
-    ///      An id can only be assigned to a single proxy at a given time,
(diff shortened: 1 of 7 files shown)


packages/contracts/src/contracts/current/protocol/Exchange/MixinAssetProxyDispatcher.sol
```diff
@@ -66,11 +66,19 @@ contract MixinAssetProxyDispatcher is
         external
         onlyOwner
     {
```
dekz: The doc comment `however, an asset proxy may be registered to multiple ids.` Is this still possible? Since we've hardcoded the assetProxyID in the assetProxy and only expose a single `getProxyId()` function.

abandeali1: You're right, will remove this.

abandeali1:
@hysz I think any new proxy deployment should have a new id, although it's still possible to use an old one (but users would still have to approve the new one in order to make use of it).

PR #566, opened 2018-04-30
Order Watcher validate order on addOrder

dekz:
## Description

Allow a user to specify whether they would like to receive an order state update when the order is added to order state watcher

## Motivation and Context

When pulling orders from SRA it's beneficial to get the order state before adding to the internal order book. 

It's a bit of wasted effort to do this outside of the order watcher through order state utils as you could have to create a couple of lazy stores out of band.

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/CHANGELOG.json
+++ b/packages/0x.js/CHANGELOG.json
@@ -9,6 +9,10 @@
             {
                 "note": "Update Web3 Provider Engine to 14.0.4",
                 "pr": 555
+            },
+            {
+                "note": "Order Watcher config to validate order when calling addOrder",
+                "pr": 566
             }
         ]
     },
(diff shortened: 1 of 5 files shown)


fabioberger:
@LogvinovLeon this is the same config that Avonian wanted... Should we re-consider it?

dekz:
I've also made some changes to perform the balance and allowance requests asynchronously. I've noticed that these can take a noticeable amount of time when it hits a cold cache. 

After experimenting, I think the added benefit of this (rather than requesting the state directly through `exchange.getOrderState` is that it also primes the cache for future updates. Rather than hitting a cold cache as soon as a partial fill happens.

fabioberger:
I think we should make these changes @dekz. Is it possible to wait with merging? I already started refactoring OrderWatcher out of 0x.js. Maybe we could re-base these changes once that gets merged?

dekz:
Yeah there's no rush at all

dekz:
Going to re-open this after v2.

Issue #567, opened 2018-04-30
validateOrderFillableOrThrowAsync throws INSUFFICIENT_MAKER_FEE_BALANCE on ZRX buy


lukeautry:
My understanding of the expected behavior is that ZRX fees can be met as long as the other side of the trade has sufficient ZRX to pay the fees. This is pretty important for an exchange that accepts ZRX for fees as it allows users to enter the market without actually having ZRX yet. This works well with filling orders.

We've noticed that new limit orders (WETH maker, ZRX taker) are flagged as INSUFFICIENT_MAKER_FEE_BALANCE by validateOrderFillableOrThrowAsync in one of our workers. This is an example of an order that was flagged:

```
{
   "exchangeContractAddress":"0x12459c951127e0c374ff9105dda097662a027093",
   "expirationUnixTimestampSec":"4102444800",
   "feeRecipient":"0x173a2467cece1f752eb8416e337d0f0b58cad795",
   "maker":"0xfaa53a74dd64e46b6e7a82893a4dec6904bdb606",
   "makerFee":"96353762307020101",
   "makerTokenAddress":"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
   "makerTokenAmount":"180100000000000000",
 "salt":"36174368916192981979559457829026432346101896339670862948314220586933861773564",
   "serializedEcSignature":"{\"v\":27,\"r\":\"0xb620ea5714e9d1052e40b537aa71fe65a45d490a01718ace8cc479e337dedc80\",\"s\":\"0x561b95bfeae5edcf0da93bb1ce358220699e79a6bf931e4622d6a12f7b545349\"}",
   "taker":"0x0000000000000000000000000000000000000000",
   "takerFee":"200000000000000000",
   "takerTokenAddress":"0xe41d2489571d322189246dafa5ebde1f4699f498",
   "takerTokenAmount":"100000000000000000000",
   "orderHash":"0x422a2c767e24a4d121cfee29157007efff0a543e9c91e8db9506b2f93ced4d98"
}
```

1) Is this actually a valid order, i.e. does the Exchange contract support 'just in time' purchases of ZRX?
2) If so, shouldn't the JS client handle this edge case?

dekz:
This should be fixed in #568

PR #568, opened 2018-04-30
Execute Taker side in trade simulation

dekz:
#567

## Description

Simulate both sides of the trade in the exchange simulator

## Motivation and Context

It's possible to just in time fill for fees when buying ZRX. Exchange simulator was only trading the maker side to the taker, as such it would fail the fee simulation as the maker did not receive any ZRX.

The current order state validation code looks intentional to not execute the Taker side (as its likely the order is open to any taker). The exchange simulator now allows for open orders to be simulated. It is assumed there will be further verification with a real taker using `validateFillOrderBalancesAllowancesThrowIfInvalidAsync` or that the taker will have sufficient balance and allowance.

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/src/utils/exchange_transfer_simulator.ts
+++ b/packages/0x.js/src/utils/exchange_transfer_simulator.ts
@@ -67,7 +67,9 @@ export class ExchangeTransferSimulator {
         tradeSide: TradeSide,
         transferType: TransferType,
     ): Promise<void> {
-        // If we are simulating an open order then 0x0 will have no balance or allowance
+        // HACK: When simulating an open order (e.g taker is NULL_ADDRESS), we don't want to adjust balances/
+        // allowances for the taker. We do however, want to increase the balance of the maker since the maker
+        // might be relying on those funds to fill subsequent orders or pay the order's fees.
         if (from === constants.NULL_ADDRESS && tradeSide === TradeSide.Taker) {
             await this._increaseBalanceAsync(tokenAddress, to, amountInBaseUnits);
             return;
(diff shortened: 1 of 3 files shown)


packages/0x.js/src/utils/exchange_transfer_simulator.ts
```diff
@@ -66,6 +67,11 @@ export class ExchangeTransferSimulator {
         tradeSide: TradeSide,
         transferType: TransferType,
     ): Promise<void> {
+        // If we are simulating an open order then 0x0 will have no balance or allowance
```
fabioberger:
```
// HACK: When simulating an open order (e.g taker is NULL_ADDRESS), we don't want to adjust balances/
// allowances for the taker. We do however, want to increase the balance of the maker since the maker
// might be relying on those funds to fill subsequent orders or pay the order's fees.
```

packages/0x.js/src/utils/order_validation_utils.ts
```diff
@@ -124,31 +124,12 @@ export class OrderValidationUtils {
         if (!_.isUndefined(expectedFillTakerTokenAmount)) {
             fillTakerTokenAmount = expectedFillTakerTokenAmount;
         }
-        const fillMakerTokenAmount = OrderValidationUtils._getPartialAmount(
+        await OrderValidationUtils.validateFillOrderBalancesAllowancesThrowIfInvalidAsync(
```
fabioberger: How does this work without throwing an error? `validateFillOrderBalancesAllowancesThrowIfInvalidAsync` also attempts a `Fee` transfer from taker `0x0`...

PR #569, opened 2018-04-30
Batch transfers in proxies

abandeali1:
## Description

Adds a `batchTransferFrom` function in each asset proxy. While this isn't actually used today, it opens doors to batch call `transferFrom` in the future, which could significantly decrease gas costs.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/AssetProxy/proxies/ERC20Proxy.sol
+++ b/packages/contracts/src/contracts/current/protocol/AssetProxy/proxies/ERC20Proxy.sol
@@ -17,6 +17,7 @@
 */
 
 pragma solidity ^0.4.21;
+pragma experimental ABIEncoderV2;
 
 import "../IAssetProxy.sol";
 import "../../../utils/LibBytes/LibBytes.sol";
@@ -33,9 +34,9 @@ contract ERC20Proxy is
 
     /// @dev Transfers ERC20 tokens. Either succeeds or throws.
     /// @param assetMetadata ERC20-encoded byte array.
-    /// @param from Address to transfer token from.
-    /// @param to Address to transfer token to.
-    /// @param amount Amount of token to transfer.
+    /// @param from Address to transfer asset from.
+    /// @param to Address to transfer asset to.
+    /// @param amount Amount of asset to transfer.
     function transferFrom(
         bytes assetMetadata,
         address from,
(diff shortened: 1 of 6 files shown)


packages/contracts/test/asset_proxy/proxies.ts
```diff
@@ -21,7 +22,7 @@ chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
-describe('Asset Transfer Proxies', () => {
+describe.only('Asset Transfer Proxies', () => {
```
dekz: cheeky

packages/contracts/src/contracts/current/protocol/AssetProxy/proxies/ERC20Proxy.sol
```diff
@@ -29,18 +30,61 @@ contract ERC20Proxy is
...
+        );
+    }
+
+    /// @dev Makes multiple transfers of assets. Either succeeds or throws.
+    /// @param assetMetadata Array of byte arrays encoded for the respective asset proxy.
+    /// @param from Array of addresses to transfer assets from.
+    /// @param to Array of addresses to transfer assets to.
+    /// @param amounts Array of amounts of assets to transfer.
+    function batchTransferFrom(
+        bytes[] memory assetMetadata,
```
dekz: Would an overloaded for ERC20 proxy with a single `assetMetadata` make sense in the future? In the example of the marketSell/Buy this data should always be the same or does this break our duck typing when using proxies?

abandeali1: Potentially. I don't think this will end up being too costly since we aren't repeating the `assetMetadata` a ton in any calldata, though.

packages/contracts/src/contracts/current/protocol/AssetProxy/proxies/ERC20Proxy.sol
```diff
@@ -33,16 +34,59 @@ contract ERC20Proxy is
...
+        address[] memory from,
+        address[] memory to,
+        uint256[] memory amounts)
+        public
```
dekz: We use external for `transferFrom` but public here, I expect it's the exchange contract who will be batchTransferring so should this also be external?

abandeali1: External visibility doesn't work with ABIv2 unfortunately.

abandeali1:
@hysz 
1. Good question. Might not be exactly the same in the future, but since they're the same for these cases might as well create a base class?
2. I think the logic for grouping by `proxyId` and deciding when to call `batchTransferFrom` is pretty complex, and might not make it into v2. Maybe we can just use it for fees though, that seems easy enough.

PR #571, opened 2018-05-04
Update MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress

abandeali1:
## Description

Rather than hard coding in the `tokenTransferProxy` as the only address that can call `removeAuthorizedAddress` with no timelock, the multisig now maintains a mapping of multiple proxy addresses that are approved to do so.

## Motivation and Context

This change needed to be made to support the new AssetProxy architecture. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
+++ b/packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
@@ -18,15 +18,19 @@
 
 pragma solidity ^0.4.10;
 
-import { MultiSigWalletWithTimeLock } from "../MultiSigWalletWithTimeLock/MultiSigWalletWithTimeLock.sol";
+import "../MultiSigWalletWithTimeLock/MultiSigWalletWithTimeLock.sol";
 
-contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWalletWithTimeLock {
+contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is
+    MultiSigWalletWithTimeLock
+{
+    event AssetProxyRegistration(address assetProxyContract, bool isRegistered);
 
-    address public TOKEN_TRANSFER_PROXY_CONTRACT;
+    // Mapping of AssetProxy contract address => approved to execute removeAuthorizedAddress without time lock.
+    mapping (address => bool) public isAssetProxyRegistered;
 
-    modifier validRemoveAuthorizedAddressTx(uint transactionId) {
+    modifier validRemoveAuthorizedAddressTx(uint256 transactionId) {
         Transaction storage tx = transactions[transactionId];
-        require(tx.destination == TOKEN_TRANSFER_PROXY_CONTRACT);
+        require(isAssetProxyRegistered[tx.destination]);
         require(isFunctionRemoveAuthorizedAddress(tx.data));
         _;
     }
@@ -35,21 +39,36 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
     /// @param _owners List of initial owners.
     /// @param _required Number of required confirmations.
     /// @param _secondsTimeLocked Duration needed after a transaction is confirmed and before it becomes executable, in seconds.
-    /// @param _tokenTransferProxy Address of TokenTransferProxy contract.
+    /// @param _assetProxyContracts Array of AssetProxy contract addresses.
     function MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress(
-        address[] _owners,
-        uint _required,
-        uint _secondsTimeLocked,
-        address _tokenTransferProxy)
+        address[] memory _owners,
+        uint256 _required,
+        uint256 _secondsTimeLocked,
+        address[] memory _assetProxyContracts)
         public
         MultiSigWalletWithTimeLock(_owners, _required, _secondsTimeLocked)
     {
-        TOKEN_TRANSFER_PROXY_CONTRACT = _tokenTransferProxy;
+        for (uint256 i = 0; i < _assetProxyContracts.length; i++) {
+            require(_assetProxyContracts[i] != address(0));
+            isAssetProxyRegistered[_assetProxyContracts[i]] = true;
+        }
+    }
+
+    /// @dev Sets approval for calling removeAuthorizedAddress on an AssetProxy contract without a timelock.
+    /// @param assetProxyContract Address of AssetProxy contract.
+    /// @param isRegistered Status of approval for AssetProxy contract.
+    function registerAssetProxy(address assetProxyContract, bool isRegistered)
+        public
+        onlyWallet
+        notNull(assetProxyContract)
+    {
+        isAssetProxyRegistered[assetProxyContract] = isRegistered;
+        AssetProxyRegistration(assetProxyContract, isRegistered);
     }
 
     /// @dev Allows execution of removeAuthorizedAddress without time lock.
     /// @param transactionId Transaction ID.
-    function executeRemoveAuthorizedAddress(uint transactionId)
+    function executeRemoveAuthorizedAddress(uint256 transactionId)
         public
         notExecuted(transactionId)
         fullyConfirmed(transactionId)
@@ -68,15 +87,26 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
     /// @dev Compares first 4 bytes of byte array to removeAuthorizedAddress function signature.
     /// @param data Transaction data.
     /// @return Successful if data is a call to removeAuthorizedAddress.
-    function isFunctionRemoveAuthorizedAddress(bytes data)
+    function isFunctionRemoveAuthorizedAddress(bytes memory data)
         public
-        constant
+        pure
         returns (bool)
     {
-        bytes4 removeAuthorizedAddressSignature = bytes4(sha3("removeAuthorizedAddress(address)"));
-        for (uint i = 0; i < 4; i++) {
-            require(data[i] == removeAuthorizedAddressSignature[i]);
-        }
+        bytes4 removeAuthorizedAddressSelector = bytes4(keccak256("removeAuthorizedAddress(address)"));
+        bytes4 first4Bytes = readFirst4(data);
+        require(removeAuthorizedAddressSelector == first4Bytes);
         return true;
+    }
+
+    function readFirst4(bytes memory data)
+        public
+        pure
+        returns (bytes4 result)
+    {
+        require(data.length >= 4);
+        assembly {
+            result := mload(add(data, 32))
+        }
+        return result;
     }
 }
--- a/packages/contracts/test/multi_sig_with_time_lock.ts
+++ b/packages/contracts/test/multi_sig_with_time_lock.ts
@@ -1,42 +1,40 @@
 import { LogWithDecodedArgs, ZeroEx } from '0x.js';
-import { BlockchainLifecycle, web3Factory } from '@0xproject/dev-utils';
-import { AbiDecoder, BigNumber } from '@0xproject/utils';
-import { Web3Wrapper } from '@0xproject/web3-wrapper';
+import { BlockchainLifecycle } from '@0xproject/dev-utils';
+import { BigNumber } from '@0xproject/utils';
 import * as chai from 'chai';
+import * as _ from 'lodash';
 import * as Web3 from 'web3';
 
-import * as multiSigWalletJSON from '../../build/contracts/MultiSigWalletWithTimeLock.json';
-import { MultiSigWalletContract } from '../src/contract_wrappers/generated/multi_sig_wallet';
-import { MultiSigWalletWithTimeLockContract } from '../src/contract_wrappers/generated/multi_sig_wallet_with_time_lock';
-import { artifacts } from '../src/utils/artifacts';
+import {
+    MultiSigWalletWithTimeLockContract,
+    SubmissionContractEventArgs,
+} from '../src/contract_wrappers/generated/multi_sig_wallet_with_time_lock';
 import { constants } from '../src/utils/constants';
 import { MultiSigWrapper } from '../src/utils/multi_sig_wrapper';
-import { ContractName, SubmissionContractEventArgs } from '../src/utils/types';
+import { ContractName } from '../src/utils/types';
 
 import { chaiSetup } from './utils/chai_setup';
 import { deployer } from './utils/deployer';
 import { provider, web3Wrapper } from './utils/web3_wrapper';
 
-const MULTI_SIG_ABI = artifacts.MultiSigWalletWithTimeLockArtifact.networks[constants.TESTRPC_NETWORK_ID].abi;
 chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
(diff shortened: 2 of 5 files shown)


packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -68,15 +87,26 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
...
         public
-        constant
+        pure
         returns (bool)
     {
-        bytes4 removeAuthorizedAddressSignature = bytes4(sha3("removeAuthorizedAddress(address)"));
-        for (uint i = 0; i < 4; i++) {
-            require(data[i] == removeAuthorizedAddressSignature[i]);
-        }
+        bytes4 removeAuthorizedAddressSelector = bytes4(keccak256("removeAuthorizedAddress(address)"));
```
dekz: should this be a constant?

packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -35,21 +39,36 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
     /// @param _owners List of initial owners.
```
dekz: line above:
`and asset proxy addresses`

packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -35,21 +39,36 @@ contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWall
...
         MultiSigWalletWithTimeLock(_owners, _required, _secondsTimeLocked)
     {
-        TOKEN_TRANSFER_PROXY_CONTRACT = _tokenTransferProxy;
+        for (uint256 i = 0; i < _assetProxyContracts.length; i++) {
+            require(_assetProxyContracts[i] != address(0));
+            isAssetProxyRegistered[_assetProxyContracts[i]] = true;
+        }
+    }
+
+    /// @dev Sets approval for calling removeAuthorizedAddress on an AssetProxy contract without a timelock.
```
dekz: `Registers or deregisters an AssetProxy for future removeAuthorizedAddress without a timelock`

packages/contracts/src/contracts/current/multisig/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -18,15 +18,19 @@
 
 pragma solidity ^0.4.10;
```
dekz: Since we'll have to redeploy this with the new functionality should we bump the solidity version up?

abandeali1: Hmm, the base MultiSig contracts have been heavily audited with 0.4.10, so it might not be worth bumping up. We might have to downgrade some of our syntax if we want to keep the old version though.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+    {
+        bytes4 first4Bytes = readFirst4(data);
+        require(REMOVE_AUTHORIZED_ADDRESS_SELECTOR == first4Bytes);
+        return true;
+    }
+
+    /// @dev Reads the first 4 bytes from a byte array of arbitrary length.
+    /// @param data Byte array to read first 4 bytes from.
+    /// @return First 4 bytes of data.
+    function readFirst4(bytes memory data)
```
hysz: Should we add `readBytes4` to `LibBytes`?

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    /// @dev Compares first 4 bytes of byte array to removeAuthorizedAddress function signature.
+    /// @param data Transaction data.
+    /// @return Successful if data is a call to removeAuthorizedAddress.
+    function isFunctionRemoveAuthorizedAddress(bytes memory data)
+        public
+        pure
+        returns (bool)
+    {
+        bytes4 first4Bytes = readFirst4(data);
```
hysz: Nit - Variable name.
More descriptive to use something like `bytes4 functionSelector = readFirst4(data);`.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+contract AssetProxyOwner is
+    MultiSigWalletWithTimeLock
+{
+
+    event AssetProxyRegistration(address assetProxyContract, bool isRegistered);
+
+    // Mapping of AssetProxy contract address => approved to execute removeAuthorizedAddress without time lock.
+    mapping (address => bool) public isAssetProxyRegistered;
+
+    bytes4 constant REMOVE_AUTHORIZED_ADDRESS_SELECTOR = bytes4(keccak256("removeAuthorizedAddress(address)"));
```
hysz: We use a few names for this: address selector, function selector, and function signature. Could be good to settle on one for consistency.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+
+    modifier validRemoveAuthorizedAddressTx(uint256 transactionId) {
```
hysz: Not sure if we generally include function comments for modifiers, but probably good practice.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+        _;
+    }
+
+    /// @dev Contract constructor sets initial owners, required number of confirmations,
+    ///      time lock, and list of AssetProxy addresses.
+    /// @param _owners List of initial owners.
+    /// @param _required Number of required confirmations.
+    /// @param _secondsTimeLocked Duration needed after a transaction is confirmed and before it becomes executable, in seconds.
+    /// @param _assetProxyContracts Array of AssetProxy contract addresses.
+    function AssetProxyOwner(
```
hysz: Should we group these parameters by type?

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+    ///      removeAuthorizedAddress without a timelock.
+    /// @param assetProxyContract Address of AssetProxy contract.
+    /// @param isRegistered Status of approval for AssetProxy contract.
+    function registerAssetProxy(address assetProxyContract, bool isRegistered)
+        public
+        onlyWallet
+        notNull(assetProxyContract)
+    {
+        isAssetProxyRegistered[assetProxyContract] = isRegistered;
+        AssetProxyRegistration(assetProxyContract, isRegistered);
```
hysz: I believe we need to include `emit` now.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+    function executeRemoveAuthorizedAddress(uint256 transactionId)
+        public
+        notExecuted(transactionId)
+        fullyConfirmed(transactionId)
+        validRemoveAuthorizedAddressTx(transactionId)
+    {
+        Transaction storage tx = transactions[transactionId];
+        tx.executed = true;
+        if (tx.destination.call.value(tx.value)(tx.data))
+            Execution(transactionId);
```
hysz: Assuming `Execution` and `ExecutionFailure` are events, we should include `emit`

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,119 @@
...
+    function AssetProxyOwner(
+        address[] memory _owners,
+        uint256 _required,
+        uint256 _secondsTimeLocked,
+        address[] memory _assetProxyContracts)
+        public
+        MultiSigWalletWithTimeLock(_owners, _required, _secondsTimeLocked)
+    {
+        for (uint256 i = 0; i < _assetProxyContracts.length; i++) {
+            require(_assetProxyContracts[i] != address(0));
```
hysz: Could save some gas to store `_assetProxyContracts[i]` in a variable. Saves on redundant `mload` instructions.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
abandeali1: Generally, function signature is the function name + parameter types, and the selector is the first 4 bytes of the hash of the function signature.We probably do have some inconsistencies though.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
abandeali1: I'm not updating the syntax because we'll probably end up deploying this with Solidity 0.4.10. The base contract have all been heavily audited and tested over time, so I'd rather not do anything to modify their bytecode unless necessary.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
abandeali1: Going to move this to `LibBytes` and keep the variable name generic.

packages/contracts/test/asset_proxy_owner.ts
```diff
@@ -0,0 +1,346 @@
...
+                SECONDS_TIME_LOCKED,
+            );
+            const isErc20ProxyRegistered = await newMultiSig.isAssetProxyRegistered.callAsync(erc20Proxy.address);
+            const isErc721ProxyRegistered = await newMultiSig.isAssetProxyRegistered.callAsync(erc721Proxy.address);
+            expect(isErc20ProxyRegistered).to.equal(true);
+            expect(isErc721ProxyRegistered).to.equal(true);
+        });
+        it('should throw if a null address is included in assetProxyContracts', async () => {
+            const assetProxyContractAddresses = [erc20Proxy.address, ZeroEx.NULL_ADDRESS];
+            expect(
```
hysz: I believe we have to do `return expect` to force a wait.

packages/contracts/test/asset_proxy_owner.ts
```diff
@@ -0,0 +1,346 @@
...
+                removeAuthorizedAddressData,
+            );
+            expect(isFunctionRemoveAuthorizedAddress).to.be.true();
+        });
+    });
+
+    describe('registerAssetProxy', () => {
+        it('should throw if not called by multisig', async () => {
+            const isRegistered = true;
+            expect(
```
hysz: Same here; I believe we need a `return expect`

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    /// @dev Allows execution of removeAuthorizedAddress without time lock.
+    /// @param transactionId Transaction ID.
+    function executeRemoveAuthorizedAddress(uint256 transactionId)
+        public
+        notExecuted(transactionId)
+        fullyConfirmed(transactionId)
+        validRemoveAuthorizedAddressTx(transactionId)
+    {
+        Transaction storage tx = transactions[transactionId];
```
dekz: Can this just call the MultiSigWallet `executeTransaction` after the additional modifier validation?

packages/contracts/test/asset_proxy_owner.ts
```diff
@@ -0,0 +1,346 @@
+import { LogWithDecodedArgs, ZeroEx } from '0x.js';
```
fabioberger: Ditto above. Use `web3-wrapper` for awaiting transaction mining.

packages/contracts/test/multi_sig_with_time_lock.ts
```diff
@@ -153,39 +119,29 @@ describe('MultiSigWalletWithTimeLock', () => {
...
                     artifacts.MultiSigWalletWithTimeLock,
                     provider,
                     txDefaults,
                     owners,
-                    SIGNATURES_REQUIRED,
+                    REQUIRED_APPROVALS,
                     SECONDS_TIME_LOCKED,
                 );
-                multiSigWrapper = new MultiSigWrapper((multiSig as any) as MultiSigWalletContract);
+                multiSigWrapper = new MultiSigWrapper(multiSig, zeroEx);
```
fabioberger: Pass in `provider` here instead of `zeroEx` and instantiate a `Web3Wrapper` inside it.

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -0,0 +1,110 @@
...
+        address[] memory _owners,
+        address[] memory _assetProxyContracts,
+        uint256 _required,
+        uint256 _secondsTimeLocked
+    )
+        public
+        MultiSigWalletWithTimeLock(_owners, _required, _secondsTimeLocked)
+    {
+        for (uint256 i = 0; i < _assetProxyContracts.length; i++) {
+            address assetProxy = _assetProxyContracts[i];
```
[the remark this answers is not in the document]
abandeali1: `registerAssetProxy` has the `onlyWallet` modifier, and `msg.sender` is just the account deploying the contract here.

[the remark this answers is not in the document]
dekz: ah yeah true I should've dug deeper onto those modifiers, it could be achieved with an internal only shared function?

packages/contracts/src/contracts/current/protocol/AssetProxyOwner/AssetProxyOwner.sol
abandeali1: We can't because we're trying to bypass the `pastTimeLock` modifier.

dekz: I guess we're adverse to refactoring to `executeTransactionInternal` and a public executeTransaction with the modifiers? Mostly due to the fact that these contracts have already been audited?

PR #521, opened 2018-04-11
[WIP] Match and Batch fill

Recmo:
## Types of changes

*   [ ] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

*   [ ] Change requires a change to the documentation.
*   [ ] Added tests to cover my changes.
*   [ ] Added new entries to the relevant CHANGELOGs.
*   [ ] Updated the new versions of the changed packages in the relevant CHANGELOGs.
*   [ ] Labeled this PR with the 'WIP' label if it is a work in progress.
*   [ ] Labeled this PR with the labels corresponding to the changed package.

PR #490, opened 2018-03-30
Split fillOrder into several functions

Recmo:
### Splits

This PR splits `fillOrder` into several steps with their own functions:

1. `orderStatus(order, signature)`
2. `getFillAmounts(order, filledAmount, takerTokenFillAmount, taker)`
3. `updateState(order, orderHash, [amounts])` 

The original `fillOrder` is then reconstructed using these.

### Status codes

* The throw on `takerTokenFillAmount == 0` is removed. The new behaviour is to do a zero fill.
* Throws on invalid orders and signatures are turned into errors.

## Motivation and Context

For batch optimizations we need to separate the token transfers from the rest of the `fillOrder` functionality. This PR allows this.

## How Has This Been Tested?

The first commit does all of the refactor, but does not change behavior. It passes all the existing tests. The second commit changes behavior (breaks tests) and the third updates the tests to pass again. Tests where adjusted, not removed.

The newly added public functions are not tested independently, but receive coverage through their use in `fillOrder`.

## Types of changes

*   [ ] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

*   [ ] Change requires a change to the documentation.
*   [x] Added tests to cover my changes.
*   [ ] Added new entries to the relevant CHANGELOGs.
*   [ ] Updated the new versions of the changed packages in the relevant CHANGELOGs.
*   [ ] Labeled this PR with the 'WIP' label if it is a work in progress.
*   [ ] Labeled this PR with the labels corresponding to the changed package.

PR #572, opened 2018-05-05
Atomic Order Matching

hysz:
## Description

Atomic Order Matching - Work in Progress.

Match two complementary orders that have a positive spread. Each order is filled at their respective price point. However, the calculations are carried out as though the orders are both being filled at the right order's price point. The profit made by the left order goes to the taker (who matched the two orders).

This PR is based on two earlier PR's by Remco:
- https://github.com/0xProject/0x-monorepo/pull/490/files
- https://github.com/0xProject/0x-monorepo/pull/521/files

In this PR:
- Refactoring of exchange core 
- Atomic order matching mixin
- Lots of test cases for atomic order matching: ERC20 / ERC721 
- Several detailed comments inline
- Rebased against Solidity v0.4.23

Todo:
- I think we have a good test set for the order matching. With the refactoring in this PR, we could write more granular test cases for the exchange core.

## Motivation and Context

The exchange core functions were getting really long. It was good to refactor/split these up. The refactor was also necessary to write a clean implementation of order matching, since there's a lot of overlap between functions.

Atomic order matching is useful for arbitrage and exchanges that take profit from the spread. 

## How Has This Been Tested?

Test cases were created for new functionality. We did not change any existing functionality. Tests still pass.

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -77,134 +77,332 @@ contract MixinExchangeCore is
...
+        bytes32 orderHash,
+        uint256 filledAmount,
+        bytes memory signature,
+        address takerAddress,
+        uint256 takerAssetFillAmount)
+        internal
+    {
+        // Ensure order status is not invalid
+        if (orderStatus == uint8(Status.INVALID)) {
+            emit ExchangeStatus(uint8(orderStatus), orderHash);
```
abandeali1: `revert` reverts all state, including events, so this patten won't work. Instead, let's just use a `require` with a reason.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -77,134 +77,332 @@ contract MixinExchangeCore is
...
-            emit ExchangeError(uint8(Errors.ORDER_FULLY_FILLED), orderHash);
-            return fillResults;
+    /// @dev Calculates amounts filled and fees paid by maker and taker.
+    /// @param order to be filled.
+    /// @param orderStatus Status of order to be filled.
+    /// @param filledAmount Amount of order already filled.
+    /// @param takerAssetFillAmount Desired amount of order to fill by taker.
+    /// @return status Return status of calculating fill amounts. Returns Status.SUCCESS on success.
+    /// @return fillResults Amounts filled and fees paid by maker and taker.
+    function calculateFillAmounts(
```
abandeali1: I'd rename this to `calculateFillResults`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -77,134 +77,332 @@ contract MixinExchangeCore is
...
+        uint256 filledAmount,
+        uint256 takerAssetFillAmount)
+        public
+        pure
+        returns (
+            uint8 status,
+            FillResults memory fillResults)
+    {
+        // Fill Amount must be greater than 0
+        if(takerAssetFillAmount <= 0) {
```
abandeali1: Missing a space after `if` here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -77,134 +77,332 @@ contract MixinExchangeCore is
...
-        if (isRoundingError(fillResults.takerAssetFilledAmount, order.takerAssetAmount, order.makerAssetAmount)) {
-            emit ExchangeError(uint8(Errors.ROUNDING_ERROR_TOO_LARGE), orderHash);
-            fillResults.takerAssetFilledAmount = 0;
-            return fillResults;
+        if (isRoundingError(
+            fillResults.takerAssetFilledAmount,
+            order.takerAssetAmount,
+            order.makerAssetAmount))
+        {
+            status = uint8(Status.ROUNDING_ERROR_TOO_LARGE);
```
abandeali1: We need to reset `fillResults.takerAssetFilledAmount` to 0 before returning here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+    /// @dev Validates context for matchOrders. Succeeds or throws.
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    function validateMatchOrdersContextOrRevert(
+        Order memory leftOrder,
+        Order memory rightOrder)
+        internal
+    {
+        // The leftOrder maker asset must be the same as the rightOrder taker asset.
+        require(areBytesEqual(leftOrder.makerAssetData, rightOrder.takerAssetData));
```
abandeali1: Rather than checking that these are equal, it would be more efficient to _assume_ that they are equal (less calldata to pass in). 

This can be done by passing in null byte arrays for `makerAssetData` and `takerAssetData` in `rightOrder` OR having `rightOrder` be a different struct that excludes these fields (slightly more efficient).

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+
+        // Make sure there is a positive spread.
+        // There is a positive spread iff the cost per unit bought (OrderA.MakerAmount/OrderA.TakerAmount) for each order is greater
+        // than the profit per unit sold of the matched order (OrderB.TakerAmount/OrderB.MakerAmount).
+        // This is satisfied by the equations below:
+        // <leftOrder.makerAssetAmount> / <leftOrder.takerAssetAmount> >= <rightOrder.takerAssetAmount> / <rightOrder.makerAssetAmount>
+        // AND
+        // <rightOrder.makerAssetAmount> / <rightOrder.takerAssetAmount> >= <leftOrder.takerAssetAmount> / <leftOrder.makerAssetAmount>
+        // These equations can be combined to get the following:
+        require(safeMul(leftOrder.makerAssetAmount, rightOrder.makerAssetAmount) >= safeMul(leftOrder.takerAssetAmount, rightOrder.takerAssetAmount));
```
abandeali1: Can we write this in multiple lines to make it more readable?

```
require(
    safeMul(leftOrder.makerAssetAmount, rightOrder.makerAssetAmount) >=
    safeMul(leftOrder.takerAssetAmount, rightOrder.takerAssetAmount)
);
```

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+        Order memory leftOrder,
+        Order memory rightOrder,
+        uint8 leftOrderStatus,
+        uint8 rightOrderStatus,
+        uint256 leftOrderFilledAmount,
+        uint256 rightOrderFilledAmount)
+        internal
+        returns (
+            uint8 status,
+            MatchedOrderFillAmounts memory matchedFillOrderAmounts)
```
abandeali1: This should be `matchedOrderFillAmounts`, right?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+        // We settle orders at the price point defined by the right order (profit goes to the order taker)
+        // The constraint can be either on the left or on the right.
+        // The constraint is on the left iff the amount required to fill the left order
+        // is less than or equal to the amount we can spend from the right order:
+        //    <leftTakerAssetAmountRemaining> <= <rightTakerAssetAmountRemaining> * <rightMakerToTakerRatio>
+        //    <leftTakerAssetAmountRemaining> <= <rightTakerAssetAmountRemaining> * <rightOrder.makerAssetAmount> / <rightOrder.takerAssetAmount>
+        //    <leftTakerAssetAmountRemaining> * <rightOrder.takerAssetAmount> <= <rightTakerAssetAmountRemaining> * <rightOrder.makerAssetAmount>
+        uint256 rightTakerAssetAmountRemaining = safeSub(rightOrder.takerAssetAmount, rightOrderFilledAmount);
+        uint256 leftTakerAssetAmountRemaining = safeSub(leftOrder.takerAssetAmount, leftOrderFilledAmount);
+        if(safeMul(leftTakerAssetAmountRemaining, rightOrder.takerAssetAmount) <= safeMul(rightTakerAssetAmountRemaining, rightOrder.makerAssetAmount))
```
abandeali1: Break this out into multiple lines for readability. Also missing a space after `if`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+        {
+            // Left order is the constraint: maximally fill left
+            (   status,
+                matchedFillOrderAmounts.left
+            ) = calculateFillAmounts(
+                leftOrder,
+                leftOrderStatus,
+                leftOrderFilledAmount,
+                leftTakerAssetAmountRemaining);
+            if(status != uint8(Status.SUCCESS)) {
```
abandeali1: Space after `if`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+                return (status, matchedFillOrderAmounts);
+            }
+
+            // The right order just spent <leftTakerAssetAmountRemaining> of their maker asset to fill the left order.
+            // The amount right gets in return is:
+            //    <leftOrderAmountBought> * <rightOrderProfitPerUnitSold>
+            // =  <matchedFillOrderAmounts.left.takerAssetFilledAmount> * <rightOrder.takerAssetAmount> / <rightOrder.makerAssetAmount>
+            // TODO: Ensure rounding error is in the right direction.
+            if(isRoundingError(rightOrder.takerAssetAmount, rightOrder.makerAssetAmount, matchedFillOrderAmounts.left.takerAssetFilledAmount)) {
```
abandeali1: Space after `if`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+
+            // Compute fill amounts for right order
+            (   status,
+                matchedFillOrderAmounts.right
+            ) = calculateFillAmounts(
+                rightOrder,
+                rightOrderStatus,
+                rightOrderFilledAmount,
+                rightOrderAmountToFill);
+            if(status != uint8(Status.SUCCESS)) {
```
abandeali1: Space after `if`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+                return (status, matchedFillOrderAmounts);
+            }
+
+            // The right order must spend at least as much as we're transferring to the left order's maker.
+            // If the amount transferred from the right order is greater than what is transferred, it is a rounding error amount.
+            // Ensure this difference is negligible by dividing the values with each other. The result should equal to ~1.
+            assert(matchedFillOrderAmounts.right.makerAssetFilledAmount >= matchedFillOrderAmounts.left.takerAssetFilledAmount);
```
abandeali1: Use `require` over `assert`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+            if(isRoundingError(matchedFillOrderAmounts.right.makerAssetFilledAmount, matchedFillOrderAmounts.left.takerAssetFilledAmount, 1)) {
+                status = uint8(Status.ROUNDING_ERROR_TOO_LARGE);
+                return (status, matchedFillOrderAmounts);
+            }
+        } else {
```
abandeali1: There's a lot of duplicate logic in this else block. Can we break this out into another function?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlement.sol
```diff
@@ -85,4 +95,87 @@ contract MixinSettlement is
...
+    /// @dev Settles matched order by transferring appropriate funds between order makers, taker, and fee recipient.
+    /// @param leftOrder First matched order.
+    /// @param rightOrder Second matched order.
+    /// @param matchedFillOrderAmounts Struct holding amounts to transfer between makers, taker, and fee recipients.
+    /// @param taker Address that matched the orders. The taker receives the spread between orders as profit.
+    function settleMatchedOrders(
+        Order memory leftOrder,
+        Order memory rightOrder,
+        MatchedOrderFillAmounts memory matchedFillOrderAmounts,
+        address taker)
```
abandeali1: `takerAddress`

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+    /// @return leftFillResults Amounts filled and fees paid by maker and taker of left order.
+    /// @return leftFillResults Amounts filled and fees paid by maker and taker of right order.
+    function matchOrders(
+        Order memory leftOrder,
+        Order memory rightOrder,
+        bytes leftSignature,
+        bytes rightSignature)
+        public
+        returns (
+            MatchedOrderFillAmounts memory matchedFillOrderAmounts)
```
abandeali1: Can we put this on the same line as `returns`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+    ///      Each order is filled at their respective price point. However, the calculations are
+    ///      carried out as though the orders are both being filled at the right order's price point.
+    ///      The profit made by the left order goes to the taker (who matched the two orders).
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    /// @param leftSignature Proof that order was created by the left maker.
+    /// @param rightSignature Proof that order was created by the right maker.
+    /// @return leftFillResults Amounts filled and fees paid by maker and taker of left order.
+    /// @return leftFillResults Amounts filled and fees paid by maker and taker of right order.
+    function matchOrders(
```
abandeali1: Nice! This function is very readable/easy to understand.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,285 @@
...
+        Order memory leftOrder,
+        Order memory rightOrder,
+        bytes leftSignature,
+        bytes rightSignature)
+        public
+        returns (
+            MatchedOrderFillAmounts memory matchedFillOrderAmounts)
+    {
+        // Get left status
+        OrderInfo memory leftOrderInfo;
```
abandeali1: Is this a struct because of the stack variable limitations?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: That assumption sounds like it could be dangerous, no? I'm also not positive it would make this call that much more efficient. This is an `internal` method that receives memory addresses as function parameters, opposed to entire structs.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: Changed this to `matchedFillResults` and `MatchedFillResults` to match the struct name (`FillResults`) in the `MixinExchangeCore`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: Yeah =\ ... I left a note on the struct definition.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: 👍  I'll replace with `require(..., [message])` after rebasing against `v0.4.23`

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
hysz: 👍  I'll replace with `require(..., [message])` after rebasing against `v0.4.23`

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: Good call 👍

hysz: Updated

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -77,134 +77,332 @@ contract MixinExchangeCore is
...
+    ////// Core exchange functions //////
+
+
+    /// @dev Gets information about an order.
+    /// @param order Order to gather information on.
+    /// @return status Status of order. Statuses are defined in the LibStatus.Status struct.
+    /// @return orderHash Keccak-256 EIP712 hash of the order.
+    /// @return filledAmount Amount of order that has been filled.
+    function getOrderInfo(Order memory order)
         public
```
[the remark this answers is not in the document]
hysz: ^ Could also be used by the forwarding contract. It should probably also be moved into `LibOrder`. Then we could use the `[ContractName]Test` pattern you used in the Solidity 0.4.23 PR.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,96 +172,242 @@ contract MixinExchangeCore is
...
+        // Compute proportional fill amounts
+        uint8 status;
+        (status, fillResults) = calculateFillResults(order, orderStatus, takerAssetFilledAmount, takerAssetFillAmount);
+        if (status != uint8(Status.SUCCESS)) {
+            emit ExchangeStatus(uint8(status), orderHash);
+            return fillResults;
+        }
+
+        // Settle order
+        (fillResults.makerAssetFilledAmount, fillResults.makerFeePaid, fillResults.takerFeePaid) =
```
abandeali1: Isn't `fillResults` already populated from `calculateFillResults`? We can just make `settleOrder` take the `fillResults` instance and not do the calculations again within `settleOrder`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,96 +172,242 @@ contract MixinExchangeCore is
...
+        // Update exchange internal state
+        updateFilledState(order, takerAddress, orderHash, fillResults);
+        return fillResults;
+    }
+
+    /// @dev Validates context for cancelOrder. Succeeds or throws.
+    /// @param order that was cancelled.
+    /// @param orderStatus Status of order that was cancelled.
+    /// @param orderHash Hash of order that was cancelled.
+    function validateCancelOrderContextOrRevert(
```
abandeali1: I prefer `validateCancelOrRevert`.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,96 +172,242 @@ contract MixinExchangeCore is
...
+    ///               unfillable state (see LibStatus.STATUS for order states)
+    function cancelOrder(Order memory order)
+        public
+        returns (bool)
+    {
+        // Fetch current order status
+        bytes32 orderHash;
+        uint8 orderStatus;
+        uint256 takerAssetFilledAmount;
+        (orderStatus, orderHash, takerAssetFilledAmount) = getOrderInfo(order);
```
abandeali1: Since we're not actually using `takerAssetFilledAmount`, we can write `(orderStatus, orderHash, ) = getOrderInfo(order);`

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+        require(
+            safeMul(leftOrder.makerAssetAmount, rightOrder.makerAssetAmount) >=
+            safeMul(leftOrder.takerAssetAmount, rightOrder.takerAssetAmount),
+            NEGATIVE_SPREAD
+        );
+    }
+
+    /// @dev Validates matched fill results. Succeeds or throws.
+    /// @param matchedFillResults Amounts to fill and fees to pay by maker and taker of matched orders.
+    function validateMatchedOrderFillResultsOrThrow(MatchedFillResults memory matchedFillResults)
```
abandeali1: `validateMatchOrThrow`

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+    LibBytes,
+    LibMath,
+    LibStatus,
+    LibOrder,
+    LibExchangeErrors,
+    MExchangeCore,
+    MMatchOrders,
+    MSettlement,
+    MTransactions
+    {
```
abandeali1: Extra indent here

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+        // We settle orders at the price point defined by the right order (profit goes to the order taker)
+        // The constraint can be either on the left or on the right.
+        // The constraint is on the left iff the amount required to fill the left order
+        // is less than or equal to the amount we can spend from the right order:
+        //    <leftTakerAssetAmountRemaining> <= <rightTakerAssetAmountRemaining> * <rightMakerToTakerRatio>
+        //    <leftTakerAssetAmountRemaining> <= <rightTakerAssetAmountRemaining> * <rightOrder.makerAssetAmount> / <rightOrder.takerAssetAmount>
+        //    <leftTakerAssetAmountRemaining> * <rightOrder.takerAssetAmount> <= <rightTakerAssetAmountRemaining> * <rightOrder.makerAssetAmount>
+        uint256 rightTakerAssetAmountRemaining = safeSub(rightOrder.takerAssetAmount, rightOrderFilledAmount);
+        uint256 leftTakerAssetAmountRemaining = safeSub(leftOrder.takerAssetAmount, leftOrderFilledAmount);
+        uint256 leftOrderAmountToFill = 0;
```
abandeali1: Let's not initialize these to 0 for consistency (they will automatically be initialized to 0).

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+        ) {
+            // Left order is the constraint: maximally fill left
+            leftOrderAmountToFill = leftTakerAssetAmountRemaining;
+
+            // The right order receives an amount proportional to how much was spent.
+            // TODO: Can we ensure rounding error is in the correct direction?
+            rightOrderAmountToFill = safeGetPartialAmount(
+                rightOrder.takerAssetAmount,
+                rightOrder.makerAssetAmount,
+                leftOrderAmountToFill);
```
abandeali1: Closing parenthesis should be on a newline.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+        );
+    }
+
+    /// @dev Calculates partial value given a numerator and denominator.
+    ///      Throws if there is a rounding error.
+    /// @param numerator Numerator.
+    /// @param denominator Denominator.
+    /// @param target Value to calculate partial of.
+    /// @return Partial value of target.
+    function safeGetPartialAmount(
```
abandeali1: Can we move this to `LibMath`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+            // The left order receives an amount proportional to how much was spent.
+            // TODO: Can we ensure rounding error is in the correct direction?
+            leftOrderAmountToFill = safeGetPartialAmount(
+                rightOrder.makerAssetAmount,
+                rightOrder.takerAssetAmount,
+                rightOrderAmountToFill);
+        }
+
+        // Calculate fill results for left order
+        (   status,
```
abandeali1:
```
(status, matchedFillResults.left) = calculateFillResults(
    leftOrder,
    leftOrderStatus,
    leftOrderFilledAmount,
    leftOrderAmountToFill
);
```

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+        // Calculate fill results for right order
+        (   status,
+            matchedFillResults.right
+        ) = calculateFillResults(
+            rightOrder,
+            rightOrderStatus,
+            rightOrderFilledAmount,
+            rightOrderAmountToFill);
+        if (status != uint8(Status.SUCCESS)) {
+            return (status, matchedFillResults);
```
abandeali1: Hmm, if the left status is successful but the right is not, `matchedFillResults` will be inaccurate here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+
+        // Get right status
+        OrderInfo memory rightOrderInfo;
+        (   rightOrderInfo.orderStatus,
+            rightOrderInfo.orderHash,
+            rightOrderInfo.orderFilledAmount
+        ) = getOrderInfo(rightOrder);
+        if (rightOrderInfo.orderStatus != uint8(Status.ORDER_FILLABLE)) {
+            emit ExchangeStatus(uint8(rightOrderInfo.orderStatus), rightOrderInfo.orderHash);
+            return matchedFillResults;
```
abandeali1: Same comment as above - `matchedFillResults` can be inaccurate here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,307 @@
...
+            leftOrder.makerAddress,
+            matchedFillResults.right.takerAssetFilledAmount);
+
+        // Settle matched orders. Succeeds or throws.
+        settleMatchedOrders(leftOrder, rightOrder, matchedFillResults, takerAddress);
+
+        // Update exchange state
+        updateFilledState(
+            leftOrder,
+            rightOrder.makerAddress,
```
abandeali1: We should just use `takerAddress` for both `updateFilledState` calls.

packages/contracts/src/contracts/current/protocol/Exchange/MixinSettlement.sol
```diff
@@ -96,4 +100,91 @@ contract MixinSettlement is
...
+        // Optimized for:
+        // * leftOrder.feeRecipient =?= rightOrder.feeRecipient
+
+        // Not optimized for:
+        // * {left, right}.{MakerAsset, TakerAsset} == ZRX
+        // * {left, right}.maker, takerAddress == {left, right}.feeRecipient
+
+        // leftOrder.MakerAsset == rightOrder.TakerAsset
+        // Taker should be left with a positive balance (the spread)
+        dispatchTransferFrom(
```
abandeali1: Currently, transfers look like:
leftMaker => taker
taker => rightMaker
rightMaker => leftMaker
The spread is always coming from the leftMaker.

However, we never want the taker to actually take custody of the tokens (other than the spread). So what would be more desirable would be:
leftMaker => rightMaker
rightMaker => leftMaker
leftMaker => taker

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MExchangeCore.sol
```diff
@@ -63,4 +64,113 @@ contract MExchangeCore is
...
+    /// @return status Status of order. Statuses are defined in the LibStatus.Status struct.
+    /// @return orderHash Keccak-256 EIP712 hash of the order.
+    /// @return takerAssetFilledAmount Amount of order that has been filled.
+    function getOrderInfo(LibOrder.Order memory order)
+        public
+        view
+        returns (
+            uint8 orderStatus,
+            bytes32 orderHash,
+            uint256 takerAssetFilledAmount);
```
abandeali1: Ending parenthesis should be on a newline when we have multiple line returns.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -48,64 +51,108 @@ contract MixinExchangeCore is
...
-    function fillOrder(
-        Order memory order,
-        uint256 takerAssetFillAmount,
-        bytes memory signature)
+    /// @dev Gets information about an order.
+    /// @param order Order to gather information on.
+    /// @return status Status of order. Statuses are defined in the LibStatus.Status struct.
+    /// @return orderHash Keccak-256 EIP712 hash of the order.
+    /// @return takerAssetFilledAmount Amount of order that has been filled.
+    function getOrderInfo(Order memory order)
```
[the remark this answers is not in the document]
dekz: Knowing the amount remaining to be filled is needed in the forwarding contract. Is there an alternative? In the mean time I have been keeping getUnavailable around in my PR.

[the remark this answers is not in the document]
hysz: @dekz  - I chatted offline with @abandeali1 about this & we decided to keep it public. Hoping it can be useful in the forwarding contract, for getting amount to fill & order state.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,315 @@
...
+        if (
+            safeMul(leftTakerAssetAmountRemaining, rightOrder.takerAssetAmount) <=
+            safeMul(rightTakerAssetAmountRemaining, rightOrder.makerAssetAmount)
+        ) {
+            // Left order is the constraint: maximally fill left
+            leftOrderAmountToFill = leftTakerAssetAmountRemaining;
+
+            // The right order receives an amount proportional to how much was spent.
+            // TODO: Can we ensure rounding error is in the correct direction?
+            rightOrderAmountToFill = safeGetPartialAmount(
```
dekz:
```
getPartialAmount(a,b,c) = (a*c)/d
```
```
rightOrderAmountToFill = (rightOrder.takerAssetAmount * leftTakerAssetAmountRemaining)/rightOrder.makerAssetAmount
```
We've already partially calculated this multiplication above, is it cheaper to store it and divide here? I believe it is the same in the else clause as well.

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSettlement.sol
```diff
@@ -19,23 +19,32 @@
...
+    /// @dev Settles matched order by transferring appropriate funds between order makers, taker, and fee recipient.
+    /// @param leftOrder First matched order.
+    /// @param rightOrder Second matched order.
+    /// @param matchedFillResults Struct holding amounts to transfer between makers, taker, and fee recipients.
+    /// @param takerAddress Address that matched the orders. The taker receives the spread between orders as profit.
+    function settleMatchedOrders(
+        LibOrder.Order memory leftOrder,
+        LibOrder.Order memory rightOrder,
+        LibFillResults.MatchedFillResults memory matchedFillResults,
+        address takerAddress
```
dekz: Can we refactor this to have a similar parameter order to settleOrder, so takerAddress before fillResults?

packages/contracts/src/utils/asset_proxy_utils.ts
```diff
@@ -8,25 +8,62 @@ export const assetProxyUtils = {
...
+        return value;
+    },
     encodeERC20ProxyData(tokenAddress: string): string {
         const encodedAssetProxyId = assetProxyUtils.encodeAssetProxyId(AssetProxyId.ERC20);
         const encodedAddress = assetProxyUtils.encodeAddress(tokenAddress);
         const encodedMetadata = Buffer.concat([encodedAssetProxyId, encodedAddress]);
         const encodedMetadataHex = ethUtil.bufferToHex(encodedMetadata);
         return encodedMetadataHex;
     },
+    decodeERC20ProxyData(proxyData: string): string /* tokenAddress */ {
```
dekz: I would like if this had a type which includes the asset proxy Id. And also include a `decodeProxyData` which returns id, token address (optional) and data (optional). Forwarding wrapper currently switches on the asset id and has to pull it out.

packages/contracts/src/contracts/current/protocol/Exchange/interfaces/IExchangeCore.sol
```diff
@@ -48,4 +48,18 @@ contract IExchangeCore {
...
+
+    /// @dev Gets information about an order: status, hash, and amount filled.
+    /// @param order Order to gather information on.
+    /// @return status Status of order. Statuses are defined in the LibStatus.Status struct.
+    /// @return orderHash Keccak-256 EIP712 hash of the order.
+    /// @return takerAssetFilledAmount Amount of order that has been filled.
+    function getOrderInfo(LibOrder.Order memory order)
+        public
+        view
+        returns (
```
dekz: Is there a reason this doesn't return an OrderInfo struct?

packages/contracts/src/utils/exchange_wrapper.ts
```diff
@@ -225,6 +225,30 @@ export class ExchangeWrapper {
         const filledAmount = new BigNumber(await this._exchange.filled.callAsync(orderHashHex));
         return filledAmount;
     }
+    public async getOrderInfoAsync(
+        signedOrder: SignedOrder,
+    ): Promise<[number /* orderStatus */, string /* orderHash */, BigNumber /* orderTakerAssetAmountFilled */]> {
```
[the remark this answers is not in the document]
dekz: Commented above about this contract function returning an OrderInfo struct, if it did it would also be typed in the generated contract wrappers and would bubble up here.

packages/contracts/src/utils/exchange_wrapper.ts
```diff
@@ -225,6 +225,30 @@ export class ExchangeWrapper {
...
+        const params = orderUtils.createMatchOrders(signedOrderLeft, signedOrderRight);
+        const txHash = await this._exchange.matchOrders.sendTransactionAsync(
+            params.left,
+            params.right,
+            params.leftSignature,
+            params.rightSignature,
+            { from },
+        );
+        const tx = await this._getTxWithDecodedExchangeLogsAsync(txHash);
+        tx.logs = _.filter(tx.logs, log => log.address === this._exchange.address);
```
dekz: this isn't needed as it is performed in `_getTxWithDecodedExchangeLogsAsync`

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,118 +207,247 @@ contract MixinExchangeCore is
...
+    /// @dev Updates state with results of a fill order.
+    /// @param order that was filled.
+    /// @param takerAddress Address of taker who filled the order.
+    /// @param takerAssetFilledAmount Amount of order already filled.
+    /// @return fillResults Amounts filled and fees paid by maker and taker.
+    function updateFilledState(
+        Order memory order,
+        address takerAddress,
+        bytes32 orderHash,
+        uint256 takerAssetFilledAmount,
```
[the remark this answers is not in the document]
hysz: Hmm yeah that is confusing. I've updated the wording here and elsewhere for consistency.

packages/contracts/src/contracts/current/protocol/Exchange/interfaces/IExchangeCore.sol
hysz: That's a good idea. This struct was originally created as a workaround for the stack limit. But yeah, it totally makes sense to just return it rather than reinterpret the output.

packages/contracts/test/utils/match_order_tester.ts
```diff
@@ -0,0 +1,352 @@
...
+    /// @param feeTokenAddress Address of ERC20 fee token.
+    /// @param takerAddress Address of taker (the address who matched the two orders)
+    /// @param erc20BalancesByOwner Current ERC20 balances.
+    /// @param erc721TokenIdsByOwner Current ERC721 token owners.
+    /// @param expectedTransferAmounts A struct containing the expected transfer amounts.
+    /// @return Expected ERC20 balances & ERC721 token owners after orders have been matched.
+    private static _calculateExpectedBalances(
+        signedOrderLeft: SignedOrder,
+        signedOrderRight: SignedOrder,
+        feeTokenAddress: string,
```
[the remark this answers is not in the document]
hysz: Negative; the fee token address is not currently included the order struct.

packages/contracts/test/utils/match_order_tester.ts
```diff
@@ -0,0 +1,352 @@
...
+    /// @param signedOrderLeft First matched order.
+    /// @param signedOrderRight Second matched order.
+    /// @param expectedLeftOrderFillAmoutBeforeMatch How much we expect the left order has been filled, prior to matching orders.
+    /// @param expectedRightOrderFillAmoutBeforeMatch How much we expect the right order has been filled, prior to matching orders.
+    /// @return TransferAmounts A struct containing the expected transfer amounts.
+    private async _calculateExpectedTransferAmountsAsync(
+        signedOrderLeft: SignedOrder,
+        signedOrderRight: SignedOrder,
+        expectedLeftOrderFillAmoutBeforeMatch: BigNumber,
+        expectedRightOrderFillAmoutBeforeMatch: BigNumber,
```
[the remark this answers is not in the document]
hysz: Yeah that is a confusing name... I've updated to simply `orderFilledAmountLeft` and `orderFilledAmountRight`. These seem more sensible and align with the naming scheme in the smart contracts.

packages/contracts/src/utils/asset_proxy_utils.ts
hysz: I've updated `decodeERC20ProxyData` & `decodeERC721ProxyData` accordingly. Also see the newly created function `decodeProxyData`, which uses the generalized form you requested.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,315 @@
...
+    /// @param leftSignature Proof that order was created by the left maker.
+    /// @param rightSignature Proof that order was created by the right maker.
+    /// @return matchedFillResults Amounts filled and fees paid by maker and taker of matched orders.
+    function matchOrders(
+        Order memory leftOrder,
+        Order memory rightOrder,
+        bytes leftSignature,
+        bytes rightSignature
+    )
+        public
```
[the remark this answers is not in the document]
dekz: ohh mannnnn

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,315 @@
...
+        Order memory rightOrder,
+        uint8 leftOrderStatus,
+        uint8 rightOrderStatus,
+        uint256 leftOrderFilledAmount,
+        uint256 rightOrderFilledAmount
+    )
+        internal
+        returns (MatchedFillResults memory matchedFillResults)
+    {
+        // We settle orders at the price point defined by the right order (profit goes to the order taker)
```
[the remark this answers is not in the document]
hysz: > we are finding the minimum remaining fillable amounts given two orders

We're comparing the fillable amount of the left order with the spendable amount of the right order.
Another way to phrase it would be: If right can sell more than left can buy, then right fills left.

I've updated this comment for clarity, but certainly let me know if the wording can still be improved.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: I'd prefer to keep it this way for readability.
Seems easier to reason about the calculation: AmountSold * ExchangeRate.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,118 +207,247 @@ contract MixinExchangeCore is
...
+        bytes32 orderHash
+    )
+        internal
     {
-        // Compute the order hash
-        bytes32 orderHash = getOrderHash(order);
-
-        // Validate the order
+        // Ensure order is valid
+        // An order can only be cancelled if it is Status.FILLABLE;
```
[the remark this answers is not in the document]
fabioberger: Ah gotcha, my bad. Makes sense!

packages/contracts/test/utils/match_order_tester.ts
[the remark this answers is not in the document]
fabioberger: Ah yep, you're right. Isn't it always ZRX address? Why not hard-code it?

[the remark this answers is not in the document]
hysz: Actually yeah that's a good point, we could just pass it into the constructor once rather than to each call. (Updated!)

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -70,42 +75,243 @@ contract MixinExchangeCore is
...
+            return orderInfo;
+        }
+        if (makerEpoch[order.makerAddress] > order.salt) {
+            orderInfo.orderStatus = uint8(Status.ORDER_CANCELLED);
+            return orderInfo;
+        }
+
+        // Fetch filled amount and validate order availability
+        orderInfo.orderFilledAmount = filled[orderInfo.orderHash];
+        if (orderInfo.orderFilledAmount >= order.takerAssetAmount) {
```
abandeali1: Is there ever a case where orderInfo.orderFilledAmount > order.takerAssetAmount?

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -70,42 +75,243 @@ contract MixinExchangeCore is
...
+        uint256 takerAssetFilledAmount = min256(takerAssetFillAmount, remainingTakerAssetAmount);
+
+        // Validate fill order rounding
+        if (isRoundingError(
+            takerAssetFilledAmount,
+            order.takerAssetAmount,
+            order.makerAssetAmount))
+        {
+            status = uint8(Status.ROUNDING_ERROR_TOO_LARGE);
+            return (status, fillResults);
```
abandeali1: Why do we `return (status, fillResults)` here but just `return` in the previous `if` blocks? (I actually prefer this, so would change the previous 2).

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    /// @param leftSignature Proof that order was created by the left maker.
+    /// @param rightSignature Proof that order was created by the right maker.
+    /// @return matchedFillResults Amounts filled and fees paid by maker and taker of matched orders.
+    /// TODO: Make this function external once supported by Solidity (See Solidity Issues #3199, #1603)
+    function matchOrders(
+        Order memory leftOrder,
+        Order memory rightOrder,
+        bytes leftSignature,
```
abandeali1: `bytes memory` :)

abandeali1: `bytes memory` :)

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    function assertValidMatch(
+        Order memory leftOrder,
+        Order memory rightOrder
+    )
+        internal
+    {
+        // The leftOrder maker asset must be the same as the rightOrder taker asset.
+        require(
```
abandeali1: I still think we should optimize here by just copying `leftOrder.makerAssetData` to `rightOrder.takerAssetData`. This is one of the bigger optimizations we can make throughout our codebase. Willing to punt on it for MVP and add a TODO though.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+            areBytesEqual(leftOrder.makerAssetData, rightOrder.takerAssetData),
+            ASSET_MISMATCH_MAKER_TAKER
+        );
+
+        // The leftOrder taker asset must be the same as the rightOrder maker asset.
+        require(
```
abandeali1: Same as above.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+            areBytesEqual(leftOrder.takerAssetData, rightOrder.makerAssetData),
+            ASSET_MISMATCH_TAKER_MAKER
+        );
+
+        // Make sure there is a positive spread.
```
abandeali1: I think this wording should be flipped - don't you want a negative spread?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+        require(
+            safeMul(leftOrder.makerAssetAmount, rightOrder.makerAssetAmount) >=
+            safeMul(leftOrder.takerAssetAmount, rightOrder.takerAssetAmount),
+            NEGATIVE_SPREAD
+        );
+    }
+
+    /// @dev Validates matched fill results. Succeeds or throws.
+    /// @param matchedFillResults Amounts to fill and fees to pay by maker and taker of matched orders.
+    function assertValidMatch(MatchedFillResults memory matchedFillResults)
```
abandeali1: I'd prefer not to overload this function. How about `assertValidMatchOrders` and `assertValidMatchResults`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+            // TODO: Can we ensure rounding error is in the correct direction?
+            leftOrderAmountToFill = safeGetPartialAmount(
+                rightOrder.makerAssetAmount,
+                rightOrder.takerAssetAmount,
+                rightOrderAmountToFill
+            );
+        }
+
+        // Calculate fill results for left order
+        uint8 status;
```
abandeali1: Combine these lines with Solidity 0.4.24

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
```diff
@@ -0,0 +1,308 @@
...
+    function assertValidMatch(MatchedFillResults memory matchedFillResults)
+        internal
+    {
+        // The left order must spend at least as much as we're sending to the combined
+        // amounts being sent to the right order and taker
+        uint256 amountSpentByLeft = safeAdd(
+            matchedFillResults.right.takerAssetFilledAmount,
+            matchedFillResults.takerFillAmount
+        );
+        require(
```
abandeali1: What miscalculation needs to happen in order for this to evaluate to false? Is this something we expect to ever hit, or are we just being extra safe?

packages/contracts/src/contracts/current/protocol/Exchange/interfaces/IMatchOrders.sol
```diff
@@ -0,0 +1,44 @@
...
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    /// @param leftSignature Proof that order was created by the left maker.
+    /// @param rightSignature Proof that order was created by the right maker.
+    /// @return matchedFillResults Amounts filled and fees paid by maker and taker of matched orders.
+    /// TODO: Make this function external once supported by Solidity (See Solidity Issues #3199, #1603)
+    function matchOrders(
+        LibOrder.Order memory leftOrder,
+        LibOrder.Order memory rightOrder,
+        bytes leftSignature,
```
abandeali1: `bytes memory` again

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -51,6 +51,15 @@ contract LibOrder {
         bytes takerAssetData;
     }
 
+    struct OrderInfo {
+        // See LibStatus for a complete description of order statuses
+        uint8 orderStatus;
+        // Keccak-256 EIP712 hash of the order
+        bytes32 orderHash;
+        // Amount of order that has been filled
+        uint256 orderFilledAmount;
```
abandeali1: I know it's verbose, but I prefer `orderTakerAssetFilledAmount` here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
abandeali1: If we expect this to sometimes be hit, we should add tests for that scenario.

hysz: Just a sanity check. And it's actually covered by the subsequent rounding error test. I'll remove this.

packages/contracts/src/contracts/current/protocol/Exchange/MixinMatchOrders.sol
hysz: I've added a todo to look into optimizations, although generally I do like being explicit and even redundant in security-related code. I'd be more comfortable doing this if we could explicitly test revert scenarios. For example: we expect this to fail with a bad signature".

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,118 +331,116 @@ contract MixinExchangeCore is
...
-        public
-        returns (bool)
+    /// @dev Validates context for cancelOrder. Succeeds or throws.
+    /// @param order that was cancelled.
+    /// @param orderStatus Status of order that was cancelled.
+    /// @param orderHash Hash of order that was cancelled.
+    function assertValidCancel(
+        Order memory order,
+        uint8 orderStatus,
+        bytes32 orderHash
```
dekz: assertValidFill takes in the taker address, thoughts on passing it via parameters for consistency in `assertValidCancel`?

Recmo: That would be cleaner. It would also make the function pure. `assertValidFill` is also pure, except for calling `isValidSignature`.

Issue #471, opened 2018-03-21
wrong declaration of FilterResult.get() in index.d.ts


thorstenhirsch:
## Expected Behavior

All callbacks are declared more or less like this:
`callback: (err: Error, result: string)`

Except for FilterResult.get(), which is declared like this:
`get(callback: () => void): void;`

But that's wrong as you can see in the web3 docs (https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethfilter):
```
// get all past logs again.
var myResults = filter.get(function(error, logs){ ... });
```

Please fix in index.d.tx

thorstenhirsch:
Okay, I think this is the correct fix:
`callback: (err: Error, logs: string[])`
But let me test it first. Maybe I need to add "| null" or "| undefined". I will come back to you next week.

fabioberger:
Any update on this @thorstenhirsch?

thorstenhirsch:
Sorry, I totally forgot that one.
Please stay tuned, I will send you a PR within the next days.

2018-04-17 12:00 GMT+02:00 Fabio Berger <notifications@github.com>:

PR #573, opened 2018-05-07
fix: typescript type of web3.eth.filter.get

thorstenhirsch:
## Description
filter.get() is the synchronous variant of filter.watch().

## Motivation and Context
While filter.watch() already has a full type declaration in your index.d.ts, filter.get() has not. So far I was using a timer in my typescript app that called filter.get(), so I needed the correct type declaration.
However I'm currently moving away from the timer and will use filter.watch() in the future... but anyway, maybe some other users will need this PR.

This PR solves #471 .

## How Has This Been Tested?
In my app... which is closed source, unfortunately. I changed index.d.ts in node_modules (after "yarn install") and tslint was happy with my change.

## Types of changes

## Checklist:

### DIFF
--- a/packages/typescript-typings/types/web3/index.d.ts
+++ b/packages/typescript-typings/types/web3/index.d.ts
@@ -73,7 +73,7 @@ declare module 'web3' {
         }
 
         interface FilterResult {
-            get(callback: () => void): void;
+            get(callback: (err: Error, result: LogEntryEvent[]) => void): void;
             watch(callback: (err: Error, result: LogEntryEvent) => void): void;
             stopWatching(callback?: () => void): void;
         }


packages/typescript-typings/types/web3/index.d.ts
```diff
@@ -73,7 +73,7 @@ declare module 'web3' {
         }
 
         interface FilterResult {
-            get(callback: () => void): void;
+            get(callback: (err: Error, result: LogEntryEvent[]) => void): void;
```
fabioberger: Let's rename `result` to `logs` to more closely match the [docs](https://github.com/ethereum/wiki/wiki/JavaScript-API#example-50).

PR #575, opened 2018-05-08
Update order hash to match latest eip712

dekz:
## Description

Update to the latest EIP712 schema for order hash.

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/test/TestLibs/TestLibs.sol
```diff
@@ -69,6 +69,14 @@ contract TestLibs is
         return orderHash;
     }
 
+    function publicGetOrderSchemaHash()
```
fabioberger: Why is this pre-fixed with `public`? It's obviously public no?

dekz: Its because we cannot override the visibility in an implementing contract. We could implement this as getOrderSchemaHash since there is no conflicting private method, but this is was for consistency.

abandeali1: I think we can drop the public from the name, all of the other methods prefixed with public are used for testing purposes. Also should change this to `external`.

PR #577, opened 2018-05-09
Prevent multiple connections in 0x Connect websocket

dekz:
## Description

Prevents multiple connections from being created (and their respective connection handlers)

## Motivation and Context

In practice, calling a large number of subscribes asynchronously can cause connect to open multiple websocket connections as there is no locking mechanism. Rather than introduce a lock and queue we can introduce internal state and an interval.

## Types of changes

## Checklist:

### DIFF
--- a/packages/connect/src/types.ts
+++ b/packages/connect/src/types.ts
@@ -11,7 +11,6 @@ export interface Client {
 }
 
 export interface OrderbookChannel {
-    connectAsync: () => Promise<void>;
     subscribe: (subscriptionOpts: OrderbookChannelSubscriptionOpts, handler: OrderbookChannelHandler) => void;
     close: () => void;
 }
--- a/packages/connect/src/types.ts
+++ b/packages/connect/src/types.ts
@@ -11,6 +11,7 @@ export interface Client {
 }
 
 export interface OrderbookChannel {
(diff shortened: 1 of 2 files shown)


fabioberger:
As per our offline conversation, let's instead use a lock around the connection creation logic so that subsequent attempts at establishing a connection will await the completion of previous calls.

dekz:
In this latest commit I changed it to use internal state to prevent multiple connections/multiple handlers being registered by tracking some internal state. 

@BMillman19 I know you're working on some websocket stuff currently, how do you feel about this approach versus introducing a locking mechanism. I think a locking mechanism would force async down to the caller to await on subscribes. This feels like it breaks event emitter conventions/interfaces a bit. Thoughts?

BMillman19:
@dekz I was thinking about this recently, in order to simplify this, we could have the subscribe method be only called once with an array of subscriptionOpts

dekz:
I think that could help, we would still need an additional `subscribe` or `add` for long running processes that sync tokenPairs and want to add new ones without a restart. Perhaps a second call to subscribe with more options would allow that but it kind of can put us in a similar situations. Hmmmm

BMillman19:
@dekz lemme know if you want to hop on a call and we can talk thru it

PR #578, opened 2018-05-09
Document Node.js version requirement and add it to package.json

albrow:
## Description

Most packages don't specify the Node version they need in __package.json__ so I added it to any that were missing. I also added a note to the README.

Some packages asked for version >= 6.0. After speaking with Fabio and Leonid, we decided to set our Node version to the latest LTS release, which is Node 8. I'm open to keeping it on Node 6 if we need to for some reason.

## Motivation and Context

Historically, I've experienced problems with different team members using different versions of Node. To pre-empt these problems, it is a good idea to specify the Node version we are using up front so that our team, as well as any third-parties who use the packages, are on the same page.

## How Has This Been Tested?

I ran all tests with `yarn lerna:test`. Since there are no code changes, there are no new tests to write. We just needed to make sure the tests pass with the new Node version.

## Types of changes

(This doesn't really fall under any of these categories. I would count is a documentation change).

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.11
+      - image: circleci/node:6.12
     environment:
       CONTRACTS_COMMIT_HASH: '9ed05f5'
     working_directory: ~/repo
@@ -29,7 +29,7 @@ jobs:
             - ~/repo
   test-installation:
     docker:
-      - image: circleci/node:8.11
+      - image: circleci/node:6.12
     working_directory: ~/repo
     steps:
       - restore_cache:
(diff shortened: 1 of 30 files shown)


LogvinovLeon:
I was suggesting 8.11 because I wasn't sure if it will work with 6.12, but now I see that that is the version we are using on CI. In that case - we have a strong evidence that it does work and I'd be in favor of lowering the version requirement to 6.12. Other than that - LGTM.

fabioberger:
Thanks for doing this @albrow! I agree with Leo, we might as well officially support v6.12 since it's more inclusive and we don't seem to be reliant on the newer features. Do you agree with this reasoning?

Really glad we will have it well documented in all packages!  🙏

albrow:
@fabioberger @LogvinovLeon Agreed. Just changed the minimum version to 6.12 and made sure all the tests still pass.

PR #581, opened 2018-05-09
Add Validator signature type

abandeali1:
## Description

Adds a new signature type `Validator` that allows an EOA to delegate signature verification to a contract.

## Motivation and Context

This is slightly different from the `Contract` signature type. With the `Contract` type, the `Signer` contract _is_ the maker of the order. The `Validator` type allows an address to delegate signature verification to a contract, but that contract does not have to be the make of the order.

Examples of where each type is useful.
`Contract` type:
- A multisig that can create orders with X amount of signatures from its owners
- A contract that allows multiple addresses to create orders with the contract's balances

`Validator` type:
- An EOA wants to sign orders using RSA, BLS, Schnorr, etc.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
@@ -20,6 +20,7 @@ pragma solidity ^0.4.23;
 
 import "./mixins/MSignatureValidator.sol";
 import "./interfaces/ISigner.sol";
+import "./interfaces/IValidator.sol";
 import "./libs/LibExchangeErrors.sol";
 import "../../utils/LibBytes/LibBytes.sol";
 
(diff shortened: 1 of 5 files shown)


packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSignatureValidator.sol
```diff
@@ -32,6 +32,7 @@ contract MSignatureValidator is
         EIP712,
         Trezor,
         Contract,
+        Validator,
```
fabioberger: Hm... Since Validator is most likely also a smart contract, the naming here is not ideal. What about `MakerContract` and `NonMakerContract`?

packages/contracts/src/contracts/current/test/TestLibBytes/TestLibBytes.sol
```diff
@@ -25,6 +25,23 @@ contract TestLibBytes is
     LibBytes
 {
 
+    /// @dev Performs a deep copy of a section of a byte array.
+    /// @param b Byte array that will be sliced.
+    /// @param startIndex Index of first byte to copy.
+    /// @param endIndex Index of last byte to copy + 1.
+    /// @return A deep copy of b from startIndex to endIndex.
+    function publicDeepCopyBytes(
```
fabioberger: Can we perhaps move all functions used solely in tests into a separate section with a comment stating that they are only used in tests?

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSignatureValidator.sol
fabioberger: I don't love these names either...

abandeali1: Maybe `ContractOwner` and `ContractDelegate`? Don 't love `maker` here, since you can use these sig types for other things too (sender abstraction)

packages/contracts/src/contracts/current/test/TestLibBytes/TestLibBytes.sol
abandeali1: Is the `test` directory not descriptive enough?

packages/contracts/src/contracts/current/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -169,9 +170,38 @@ contract MixinSignatureValidator is
...
+            return isValid;
+
+        // Signature verified by validator contract.
+        // If used with an order, the maker of the order can still be an EOA.
+        // A signature using this type should be encoded as:
+        // | Offset | Length | Contents                        |
+        // | 0x00   | 1      | Signature type is always "\x07" |
+        // | 0x01   | 20     | Address of validator contract   |
+        // | 0x15   | **     | Signature to validate           |
+        } else if (signatureType == SignatureType.Validator) {
```
dekz: Maybe I'm missing something obvious here, with the Signature Contract the makerAddress is the signer, so we call back into the contract of Signer (which is makerAddress) to validate the signature. This shows intent as both the makerAddress and the validator are the same.

In the Signature Validator case a separate Validator is called (not makerAddress). How do we know the makerAddress intended to use the Signature Validator? Couldn't I deploy a malicious Validator that accepts any signatures and create fake orders using that Validator?

Should there be an Authorized Validator check for signer?

abandeali1: Yep, you're right. There needs to be a way of approving a validator contract.

packages/contracts/src/contracts/current/protocol/Exchange/mixins/MSignatureValidator.sol
fabioberger: Yeah, perhaps `signer` and `validator`.

PR #582, opened 2018-05-10
[WIP] Create a private Ethereum network for testing purposes

albrow:
This is a WIP PR to create a private, single-node Ethereum network which we can use for testing purposes. The network is only intended to be available on localhost and has peer discovery disabled. I'm calling this network the "devnet".

## Progress:

* [x] Create a pre-configured, runnable Docker image for the devnet.
* [ ] Add the ability to manipulate time on the devnet node. This will likely require some changes to Geth to expose a new RPC endpoint.
* [x] Create a small utility library to emulate support for some of the features we need from Ganache. This includes things like taking snapshots and ~~starting a one-off process to mine a few blocks~~ (we no longer need to do this since the devnet now insta-mines whenever there are new blocks). We should be able to do all of these things via RPC calls to the devnet.
* [x] Use the utility library for the tests in the contracts package.
* [ ] Start and stop the devnet automatically for the tests that require it (if it is not already running).

## Motivation and Context

We currently use [Ganache](http://truffleframework.com/ganache/) whenever we need to run tests against an Ethereum network. While Ganache is a great tool in general, it has some subtle differences in behavior compared to a real Ethereum network and may not always have the latest features. For these reasons, running our tests against a more realistic Ethereum network is desirable.

## How Has This Been Tested?

(TODO: use the new devnet for tests in the contracts package and fill in this section)

## Types of changes

## Checklist:


albrow:
@fabioberger does it make sense to create a new package for this as I've done?

packages/devnet/genesis.json
```diff
@@ -0,0 +1,60 @@
+{
+  "config": {
+    "chainId": 22395,
```
LogvinovLeon: Why that number? If there is no reason - let's use 50. That's the network id we're using for testing everywhere else.

packages/devnet/node0/keystore/UTC--2018-05-08T19-46-01.673162553Z--35e52c375871faaa9923defbda5820dfa5aefc8f
```diff
@@ -0,0 +1 @@
+{"address":"35e52c375871faaa9923defbda5820dfa5aefc8f","crypto":{"cipher":"aes-128-ctr","ciphertext":"85e47d4528324af10d32b2adba85f5c0d34df23c26975e9a510617ca864b8f18","cipherparams":{"iv":"7d0b4b7602a95aa8c379f3028dee72d7"},"kdf":"scrypt","kdfparams":{"dklen":32,"n":262144,"p":1,"r":8,"salt":"65528ddb97676091aac1f7f9857e53ccb1ac848a436e248ccd3c9fe6fe9fff7b"},"mac":"76aaa9893739cf50f990b5678a27256220d3300e4682ebcba00ff58952f8b131"},"id":"39277553-7094-4239-8418-3dcd07499b60","version":3}
```
LogvinovLeon: newline

packages/devnet/genesis.json
albrow: No reason. This was randomly generated by puppeth. Just changed it to 50.

packages/devnet/node0/keystore/UTC--2018-05-08T19-46-01.673162553Z--35e52c375871faaa9923defbda5820dfa5aefc8f
albrow: Added.

packages/contracts/test/exchange/core.ts
```diff
@@ -628,14 +629,14 @@ describe('Exchange', () => {
             expect(newBalances).to.be.deep.equal(balances);
         });
 
-        it.skip('should throw if maker allowances are too low to fill order and \
+        it('should throw if maker allowances are too low to fill order and \
                 shouldThrowOnInsufficientBalanceOrAllowance = true', async () => {
             await rep.approve.sendTransactionAsync(tokenTransferProxy.address, new BigNumber(0), { from: maker });
-            expect(
+            await expectRevertOrAlwaysFailingTransaction(
```
albrow: Previously, there was a really subtle bug here. Without `await`, if the promise returned by `expect` rejects, it ***does not*** result in a test failure. Instead, it results in a warning being logged about an unhandled promise rejection (which whizzes by and is really easy to miss).

This is a really unfortunate footgun in JavaScript, but it is somewhat possible to guard against it by [making Node throw when it detects unhandled promise rejections](https://medium.com/@dtinth/making-unhandled-promise-rejections-crash-the-node-js-process-ffc27cfcc9dd). I hope to address this in a separate PR.

albrow: https://github.com/0xProject/0x-monorepo/pull/597 will help us catch cases like this.

albrow:
This has been superseded by #622, which is based on the `v2-prototype` branch instead of `development`.

PR #597, opened 2018-05-17
Use make-promises-safe to catch unhandled rejections in tests

albrow:
## Description

This PR pulls in the `make-promises-safe` package as a dependency and imports it in all relevant test files. The effect of this is that if there are ever any unhandled rejections in our tests, the process will crash and we will be made aware of them.

## Motivation and Context

JavaScript has a major footgun: Unhandled promise rejections are either silently ignored, or in Node version 6 or later, they are logged to stdout (but the process keeps going, so e.g., this means tests still pass). This [short blog post](https://medium.com/@dtinth/making-unhandled-promise-rejections-crash-the-node-js-process-ffc27cfcc9dd) goes into more detail.

As a specific example, as part of my work on #582, I noticed that there were some places in our existing tests where we were [not handling promise rejections correctly](https://github.com/0xProject/0x-monorepo/pull/582#discussion_r188796875). Mistakes like this are really easy to make and it is impractical to hope to catch every one in code review.

The way to fix this is to utilize the 'unhandledRejection' event:

```javascript
process.on('unhandledRejection', up => { throw up })
```

In theory, this will cause Node to crash instead of merely logging the error. However, even this has some subtleties and can be tricky to get right, especially in the context of asynchronous code. Luckily the `make-promises-safe` package does what we want and covers some of the edge cases. It works via side-effect so you just have to `import` it to get Node to crash instead of logging whenever there are uncaught exceptions.

This is a per-process global effect, so I specifically ***only*** added it to our test files. We don't want to make the decision to change the default Node behavior for anyone who imports our packages.

## How Has This Been Tested?

I ran all the tests and verified they still pass.

In addition, I constructed a test case which contains an unhandled promise rejection and confirmed:

1. Before importing the `make-promises-safe` package, the test passes.
2. After importing the `make-promises-safe` package, the test process crashes and we can see the error in the logs.

However, since such a test case always results in a crash now, we can't leave it in our codebase. I removed it after confirming the behavior.

## Types of changes

## Checklist:


packages/0x.js/package.json
```diff
@@ -100,11 +101,11 @@
         "@0xproject/assert": "^0.2.9",
         "@0xproject/base-contract": "^0.3.1",
         "@0xproject/contract-wrappers": "^0.0.1",
-        "@0xproject/order-watcher": "^0.0.1",
         "@0xproject/order-utils": "^0.0.4",
+        "@0xproject/order-watcher": "^0.0.1",
+        "@0xproject/sol-compiler": "^0.4.3",
         "@0xproject/types": "^0.6.3",
         "@0xproject/typescript-typings": "^0.3.1",
-        "@0xproject/sol-compiler": "^0.4.3",
```
fabioberger: It looks like you are running prettier on this file... We added `package.json` to `.prettierignore`, is it not adhering to that in your IDE?

fabioberger: @LogvinovLeon why did we add it to gitignore? It might be better to use prettier here.

albrow: I didn't edit any __package.json__ files directly. It was all done via `lerna add`.

PR #598, opened 2018-05-18
Mobile optimize the portal layout

BMillman19:
## Description

* Move app specific colors out of `react-shared` and into the website package
* Consolidate `shouldShowPortalV2` logic into utils
* Redesign the drawer menu for portal v2
* Modulate the number of columns displayed in the `RelayerIndex` component based on screen width
* Switch between different layouts in the Portal component based on screen width
* Fix a missing key error in `TokenLink` component

## Motivation and Context

If you're not mobile first, you're mobile last.

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/portal/drawer_menu.tsx
+++ b/packages/website/ts/components/portal/drawer_menu.tsx
@@ -1,9 +1,11 @@
-import { colors, Styles } from '@0xproject/react-shared';
+import { Styles } from '@0xproject/react-shared';
 import * as _ from 'lodash';
 import * as React from 'react';
 
-import { Menu } from 'ts/components/portal/menu';
+import { defaultMenuItemEntries, Menu } from 'ts/components/portal/menu';
 import { Identicon } from 'ts/components/ui/identicon';
+import { WebsitePaths } from 'ts/types';
+import { colors } from 'ts/utils/colors';
 import { utils } from 'ts/utils/utils';
 
 const IDENTICON_DIAMETER = 45;
@@ -11,7 +13,7 @@ const BORDER_RADIUS = '50%';
 
 const styles: Styles = {
     root: {
-        backgroundColor: '#4A4A4A',
+        backgroundColor: colors.drawerMenuBackground,
         width: '100%',
         height: '100%',
     },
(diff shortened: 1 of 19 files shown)


packages/website/ts/components/portal/menu.tsx
```diff
@@ -38,49 +49,66 @@ const menuItemEntries: MenuItemEntry[] = [
...
     iconName: string;
     selected: boolean;
+    theme: MenuTheme;
 }
 const MenuItemLabel: React.StatelessComponent<MenuItemLabelProps> = (props: MenuItemLabelProps) => {
     const styles: Styles = {
-        iconStyle: {
-            color: props.selected ? SELECTED_ICON_COLOR : DEFAULT_ICON_COLOR,
+        root: {
+            backgroundColor: props.selected ? props.theme.selectedBackgroundColor : undefined,
```
fragosti: So do we use `undefined` instead of `null` in our code? I can get onboard with that but I am more accustomed with using null. The convention I've always followed is that you use null to express "the value for none" and undefined to express "this was never defined". I think this is an age-old question in JS and as long as we're consistent I'm happy to go with undefined.

packages/website/ts/components/portal/portal.tsx
```diff
@@ -448,13 +465,17 @@ export class Portal extends React.Component<PortalProps, PortalState> {
         const newScreenWidth = utils.getScreenWidth();
         this.props.dispatcher.updateScreenWidth(newScreenWidth);
     }
+    private _isSmallScreen(): boolean {
+        const result = this.props.screenWidth === ScreenWidths.Sm;
+        return result;
```
fragosti: If this type of stuff is what you were thinking about making a Higher Order Component then I 100% agree (not now though)

packages/website/ts/components/portal/menu.tsx
BMillman19: we pretty much use `undefined` everywhere except for conditional rendering of react components: https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

packages/website/ts/components/portal/drawer_menu.tsx
```diff
@@ -0,0 +1,68 @@
...
+        height: '100%',
+    },
+    identicon: {
+        borderWidth: 3,
+        borderStyle: 'solid',
+        borderColor: colors.white,
+        borderRadius: BORDER_RADIUS,
+        MozBorderRadius: BORDER_RADIUS,
+        WebkitBorderRadius: BORDER_RADIUS,
+    },
```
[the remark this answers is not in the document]
BMillman19: I'll make a ticket for auditing places where we use inline styles that may need this kind of treatment

packages/website/ts/components/portal/portal.tsx
BMillman19: I was thinking specifically about generalizing / abstracting away the mechanism of alerting the dispatcher of screen width updates which is baked into the portal component now: https://github.com/0xProject/0x-monorepo/blob/a26cc7c14d4e59102d5f84d534f75c42b8683cad/packages/website/ts/components/portal/portal.tsx#L293

fragosti: Yea I _think_ we're on the same page? Basically it should be possible to dependency inject stuff like screen width and these helper methods into any arbitrary component. We can do this using redux and connect if we want (dispatching the new screen width in some root component or something), or just bypassing that and just do the listening in the HOC. The argument for not bypassing would be that we can claim that App = fn(local state, redux store) and is therefore more predictable / pure / etc...

PR #601, opened 2018-05-22
Contracts test fixes

albrow:
This PR includes a handful of fixes for the tests in the contracts package.

## Description

- Added `before` and `after` hooks which include calls to `blockchainLifecycle` where needed. This ensures that we always reset the state after the tests are done running.
- Added `awaitTransactionMinedAsync` calls for every time we call `sendTransactionAsync`. Usually, Ganache is fast enough that we can assume the transaction was more or less instantly mined, but every now and then that is not the case and we see random test failures. When I was testing against Geth, it was pretty much a necessity because Geth does more things and is a little bit slower. I set the poll interval to 100ms (and changed some existing calls from 1s to 100ms), so in practice it doesn't add much to the total time it takes to run the tests.
- Fixed some cases where we were not handling promise rejections. #597 mitigates the fallout of failing to handle rejected promises, but its still better to fix our test code. (It's the difference between mocha crashing vs. failing the test with a good error message).

## Motivation and Context

I encountered some of these small issues while working on testing our contracts against Geth and we thought it would be a good idea to split them into a separate PR. Smaller PRs are easier to review.

## How Has This Been Tested?

I simply ran all the existing contracts tests and ensured they still pass.

## Types of changes

## Checklist:


packages/contracts/test/exchange/core.ts
```diff
@@ -520,23 +532,39 @@ describe('Exchange core', () => {
...
-            await erc20TokenA.approve.sendTransactionAsync(erc20Proxy.address, constants.INITIAL_ERC20_ALLOWANCE, {
-                from: makerAddress,
-            });
+            await web3Wrapper.awaitTransactionMinedAsync(
+                await erc20TokenA.approve.sendTransactionAsync(erc20Proxy.address, new BigNumber(0), {
+                    from: makerAddress,
+                }),
+                constants.AWAIT_TRANSACTION_MINED_MS,
+            );
+            await expect(exchangeWrapper.fillOrderAsync(signedOrder, takerAddress)).to.be.rejectedWith(
```
fabioberger: I think the reason we forgot to await these calls to `except` is because the `expect` method does not following our convention of suffixing async method names with `async`. 

Let's fix this by importing `expect` as `expectAsync`.

packages/contracts/test/multi_sig_with_time_lock.ts
```diff
@@ -169,7 +183,8 @@ describe('MultiSigWalletWithTimeLock', () => {
                 txHash = await multiSig.confirmTransaction.sendTransactionAsync(txId, {
                     from: owners[1],
                 });
-                const confRes = await zeroEx.awaitTransactionMinedAsync(txHash);
+                await web3Wrapper.awaitTransactionMinedAsync(txHash, constants.AWAIT_TRANSACTION_MINED_MS);
+                const confRes = await zeroEx.awaitTransactionMinedAsync(txHash, constants.AWAIT_TRANSACTION_MINED_MS);
```
fabioberger: `confRes` is not a very descriptive variable name. What about using the types name `transactionReceiptWithDecodedLogs` or `transactionReceipt` here instead? I know verbosity of variable names is a stylistic thing, but it does have a real impact on readability. This is especially true during code reviews when you do not have access to "go-to-definitions" and other text editor aids.

packages/contracts/test/exchange/core.ts
albrow: Well, `expect` is not actually asynchronous. We often use it it synchronously in expressions like `expect(isAuthorized).to.be.false();`. It's the `to.be.rejectedWith` part that that's asynchronous and returns a promise.

I'm not sure if there's anything we can do here.

packages/contracts/test/multi_sig_with_time_lock.ts
albrow: I didn't actually come up with that name; it was already there before. But sure, I'll change it.

packages/contracts/test/multi_sig_with_time_lock.ts
```diff
@@ -139,8 +139,7 @@ describe('MultiSigWalletWithTimeLock', () => {
                 expect(initialSecondsTimeLocked).to.be.equal(0);
 
                 txHash = await multiSig.executeTransaction.sendTransactionAsync(txId, { from: owners[0] });
-                await web3Wrapper.awaitTransactionMinedAsync(txHash, constants.AWAIT_TRANSACTION_MINED_MS);
```
albrow: I think both web3Wrapper.awaitTransactionMinedAsync and zeroEx.awaitTransactionMinedAsync do the same thing, so I removed one of them. @fabioberger can you confirm that's correct?

PR #603, opened 2018-05-22
Improve wallet layout using flexbox

BMillman19:
## Description

* Refactor the `Wallet` component and `WrappedEtherItem` component to use flexbox for layout
* New `IconButton` component

## Types of changes

## Checklist:


packages/website/ts/components/wallet/wrap_ether_item.tsx
```diff
@@ -152,16 +158,18 @@ export class WrapEtherItem extends React.Component<WrapEtherItemProps, WrapEther
     }
     private _renderIsEthConversionHappeningSpinner(): React.ReactElement<{}> {
         return this.state.isEthConversionHappening ? (
-            <div className="pl1" style={{ paddingTop: 10 }}>
+            <div className="pl3 pr2" style={styles.conversionSpinner}>
                 <i className="zmdi zmdi-spinner zmdi-hc-spin" />
             </div>
-        ) : null;
+        ) : (
+            <div style={styles.conversionSpinnerPlaceholder} />
```
BMillman19: can we set the css property `visibility` to `hidden` here in order to get rid of the magic number?

packages/website/ts/components/ui/icon_button.tsx
```diff
@@ -0,0 +1,63 @@
...
+    public static defaultProps: Partial<IconButtonProps> = {
+        onClick: _.noop,
+        labelText: '',
+        color: colors.mediumBlue,
+    };
+    public constructor(props: IconButtonProps) {
+        super(props);
+        this.state = {
+            isHovering: false,
+        };
```
fragosti: I think you can make this component stateless and just user the css :hover selector for this hover effect. https://www.w3schools.com/cssref/sel_hover.asp

We do have this all.less file where we can write less. However this also works :)

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -444,16 +418,19 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
     private _renderAccessoryItems(config: AccessoryItemConfig): React.ReactElement<{}> {
         const shouldShowWrappedEtherAction = !_.isUndefined(config.wrappedEtherDirection);
         const shouldShowToggle = !_.isUndefined(config.allowanceToggleConfig);
+        // if we don't have a toggle, we still want some space to the right of the "wrap" button so that it aligns with
+        // the "unwrap" button in the row below
+        const toggle = shouldShowToggle ? (
+            this._renderAllowanceToggle(config.allowanceToggleConfig)
+        ) : (
+            <div style={{ width: NO_ALLOWANCE_TOGGLE_SPACE_WIDTH }} />
```
fragosti: Maybe you can achieve this layout by using justify-content: space-between (and removing the fixed width), which would push the elements far apart -- so the balance and toggle will be in the same place regardless of whether there is an element in between them. For the alignment of the wrap button, you may have to add some margin/spacing on the left to make it look like it does now with this approach.

packages/website/ts/components/wallet/wrap_ether_item.tsx
fragosti: 👍

packages/website/ts/components/ui/icon_button.tsx
fragosti: Ok I agree with that. I do think that this type of thing is better suited for CSS but I agree that having the styles co-located with the component code trumps that. 

*Thoughts*
However, I do think we will run into this more often, like, we need to be able to use CSS. Psuedo selectors and media queries are going to come up -- also it's just generally not great that we're using inline styles so heavily, although basscss and material UI is doing a lot of the heavy lifting. We could use css-modules, but we would need to change the structure of our project (one directory per component so that the stylesheets can live next to the component), or we can use CSS-IN-JS like styled-components. But yea not relevant now --- but in the future 😄

packages/website/ts/components/wallet/wallet.tsx
BMillman19: Alright I looked into this a bit more and I think it might be ok to keep this as is for now since it is a pretty special case and we will still need some magic numbers to get the "wrap" button to align with the "unwrap" button". `AllowanceToggle` is not set up to show undefined configurations so it will require a refactor of that component

fragosti: Just occurred to me that it may have made sense to use a column layout. This works though -- a consideration for the future.
http://basscss.com/#basscss-grid

PR #613, opened 2018-05-23
Check transaction receipt status

albrow:
## Description

This PR introduces an `awaitTransactionSuccessAsync` method in `Web3Wrapper`. It works similarly to `awaitTransactionMinedAsync` but checks the transaction receipt and throws if it has a status not equal to 1.

## Motivation and Context

In lots of different places in our tests, we call `awaitTransactionMinedAsync` to wait for a transaction to be added to the blockchain. However, in Ethereum, just because a transaction was mined doesn't mean it was successful. Failed transactions are still mined (though don't affect any state in the blockchain).

To check if a transaction was successful, you need to check the `status` field of the [transaction receipt](https://github.com/ethereum/wiki/wiki/JavaScript-API#web3ethgettransactionreceipt). The `awaitTransactionSuccessAsync` method does this and throws if it is not 1. Clients are inconsistent about how they report a failure; it's either 0, null, or undefined. However, they appear to be consistent in representing success with a status of 1.

When I was trying to debug some intermittent failures on my Geth branch, I noticed we were sometimes getting a transaction receipt status of `null`. This led to failures later on in the test which were difficult to track down. I still don't know why these intermittent transaction failures are happening but with this change we will at least be able to fail fast and see the error closer to where the problem actually occurred.

## How Has This Been Tested?

I confirmed the existing tests still pass (save for a known issue with the 0x.js package) after replacing all `awaitTransactionMinedAsync` calls with `awaitTransactionSuccessAsync`.

I'm not sure how to construct a test case for the `awaitTransactionSuccessAsync` method. Usually, Ganache throws at `sendTransactionAsync` if the transaction fails, before we even call `awaitTransactionSuccessAsync`. As mentioned above, I have seen transaction receipts with a status of `null`, but it didn't happen every time and I only saw it with Geth. Even with Geth, it usually fails at the gas estimation step, before we call `awaitTransactionSuccessAsync`. I can't reliably reproduce the `null` case.

If someone has a suggestion, I'd love to hear it.

## Types of changes

## Checklist:

### DIFF
--- a/packages/web3-wrapper/src/web3_wrapper.ts
+++ b/packages/web3-wrapper/src/web3_wrapper.ts
@@ -328,6 +328,10 @@ export class Web3Wrapper {
     }
     /**
      * Waits for a transaction to be mined and returns the transaction receipt.
+     * Note that just because a transaction was mined does not mean it was
+     * successful. You need to check the status code of the transaction receipt
+     * to find out if it was successful, or use the helper method
+     * awaitTransactionSuccessAsync.
      * @param   txHash            Transaction hash
      * @param   pollingIntervalMs How often (in ms) should we check if the transaction is mined.
      * @param   timeoutMs         How long (in ms) to poll for transaction mined until aborting.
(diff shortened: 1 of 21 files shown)


albrow:
Hang on, merge conflicts just showed up. Fixing them now...

albrow:
It looks like CI is failing to generate coverage reports, but besides that the tests are passing.

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -377,6 +381,28 @@ export class Web3Wrapper {
...
+     * @return  Transaction receipt with decoded log args.
+     */
+    public async awaitTransactionSuccessAsync(
+        txHash: string,
+        pollingIntervalMs: number = 1000,
+        timeoutMs?: number,
+    ): Promise<TransactionReceiptWithDecodedLogs> {
+        const receipt = await this.awaitTransactionMinedAsync(txHash, pollingIntervalMs, timeoutMs);
+        if (receipt.status !== 1) {
+            throw new Error('Transaction failed (transaction receipt did not have a status of 1)');
```
fabioberger: Let's add the `txHash` in this error message to help the user while debugging.

dekz: I'm not sure if the section in brackets is required?
`Transaction Failed: ${txHash}` feels like a good enough error message.

albrow:
Merging in spite of coverage failure after talking with Fabio and Leo. The tests are passing and I'd like to move forward with other work.

PaulRBerg:
@albrow do you happen to know off the top of your head how to fix this issue: https://github.com/0xProject/0x-monorepo/issues/2119?

Seems to be related to this PR.

PR #622, opened 2018-05-25
Run contract tests against private Geth network

albrow:
This supersedes #582 and is based on the `v2-prototype` branch instead of `development`.

This is a WIP PR to create a private, single-node Ethereum network which we can use for testing purposes. The network is only intended to be available on localhost and has peer discovery disabled. I'm calling this network the "devnet".

## Progress:

* [x] Create a pre-configured, runnable Docker image for the devnet.
* [x] Fix a bug in Geth that caused it to crash in certain conditions when `debug_setHead` was called. (See https://github.com/0xProject/go-ethereum/pull/1).
* [x] Add the ability to manipulate time on the devnet node by adding a new feature to Geth. (See https://github.com/0xProject/go-ethereum/pull/2).
* [x] Modify BlockChainLifecycle to use the `debug_setHead` method on Geth.  
* [x] Modify Web3Wrapper to work with new Geth RPC methods.
* [x] Temporarily hard-code the contracts tests to use Geth instead of Ganache.
* [x] Fix remaining test cases in the contracts package.
* [x] Make two different provider configurations: one for Geth and one for Ganache and run tests against both. (It's possible we'll decide to have Geth tests only run if a flag or environment variable is set). 
* [ ] ~~Automatically start/stop the Docker container when the contracts test are run. Or, if this turns out to be unreliable, provide instructions for manually starting/stopping the Docker container~~ (We decided to hold off on this for now).

## Motivation and Context

We currently use [Ganache](http://truffleframework.com/ganache/) whenever we need to run tests against an Ethereum network. While Ganache is a great tool in general, it has some subtle differences in behavior compared to a real Ethereum network and may not always have the latest features. For these reasons, running our tests against a more realistic Ethereum network is desirable.

## How Has This Been Tested?

All of the tests in the contracts package have been updated to test against Geth as well as Ganache.

## Types of changes

## Checklist:


albrow:
Removed WIP tag. This PR is now ready for review.

packages/contracts/src/utils/assertions.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Rejects if the given Promise does not reject with an error indicating
+ * insufficient funds.
+ * @param p the Promise which is expected to reject
+ * @returns a new Promise which will reject if the conditions are not met and
+ * otherwise resolve with no value.
+ */
+export function expectInsufficientFundsAsync<T>(p: Promise<T>): PromiseLike<void> {
+    return _expectEitherErrorAsync(p, 'insufficient funds', "sender doesn't have enough funds");
```
fabioberger: Let's consistently use single quotes. Surprised linter didn't catch this.

packages/contracts/src/utils/constants.ts
```diff
@@ -19,6 +19,13 @@ const TESTRPC_PRIVATE_KEYS_STRINGS = [
 export const constants = {
     INVALID_OPCODE: 'invalid opcode',
     REVERT: 'revert',
+    LIB_BYTES_GT_ZERO_LENGTH_REQUIRED: 'Length must be greater than 0.',
+    LIB_BYTES_GTE_4_LENGTH_REQUIRED: 'Length must be greater than or equal to 4.',
+    LIB_BYTES_GTE_20_LENGTH_REQUIRED: 'Length must be greater than or equal to 20.',
+    LIB_BYTES_GTE_32_LENGTH_REQUIRED: 'Length must be greater than or equal to 32.',
+    LIB_BYTES_INDEX_OUT_OF_BOUNDS: 'Specified array index is out of bounds.',
+    ERC20_INSUFFICIENT_BALANCE: 'Insufficient balance to complete transfer.',
+    ERC20_INSUFFICIENT_ALLOWANCE: 'Insufficient allowance to complete transfer.',
```
fabioberger: This error message seems generic enough to also apply to an ERC721 transfer... Does it not?

packages/contracts/src/utils/increase_time.ts
```diff
@@ -0,0 +1,31 @@
...
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        firstAccount = accounts[0];
+    }
+
+    const offset = await web3Wrapper.increaseTimeAsync(seconds);
+    // Note: we need to send a transaction after increasing time so
+    // that a block is actually mined. The contract looks at the
+    // last mined block for the timestamp.
+    await web3Wrapper.awaitTransactionSuccessAsync(
+        await web3Wrapper.sendTransactionAsync({ from: firstAccount, to: firstAccount, value: 0 }),
```
fabioberger: Lol, funny that this is a valid tx.

packages/contracts/src/utils/web3_wrapper.ts
```diff
@@ -5,15 +5,33 @@ import { Provider } from 'ethereum-types';
 
 import { coverage } from './coverage';
 
-export const txDefaults = {
+const testProvider = process.env.TEST_PROVIDER || 'ganache';
```
fabioberger: Can we please make the test providers into an enum. I'd prefer we don't repeat string literal values in order to ease future refactoring.

packages/contracts/test/exchange/core.ts
```diff
@@ -712,7 +684,9 @@ describe('Exchange core', () => {
                     salt: new BigNumber(3),
                 }),
             ];
-            await exchangeWrapper.batchFillOrdersNoThrowAsync(signedOrders, takerAddress);
+            await exchangeWrapper.batchFillOrdersNoThrowAsync(signedOrders, takerAddress, {
+                gas: 490000,
```
fabioberger: Please assign this gas amount to a well-named constant. Also, if this is a hack, please add a `// HACK:` comment explaining why this is hard-coded.

packages/contracts/test/exchange/wrapper.ts
```diff
@@ -197,12 +198,13 @@ describe('Exchange wrappers', () => {
                 takerAssetAmount: Web3Wrapper.toBaseUnitAmount(new BigNumber(200), 18),
             });
             const takerAssetFillAmount = signedOrder.takerAssetAmount.div(2);
+
             await exchangeWrapper.fillOrderNoThrowAsync(signedOrder, takerAddress, {
                 takerAssetFillAmount,
+                gas: 250000,
```
fabioberger: Ditto for this and all other hard-coded gas amounts.

packages/dev-utils/src/blockchain_lifecycle.ts
```diff
@@ -1,24 +1,60 @@
...
+
 export class BlockchainLifecycle {
     private _web3Wrapper: Web3Wrapper;
     private _snapshotIdsStack: number[];
     constructor(web3Wrapper: Web3Wrapper) {
         this._web3Wrapper = web3Wrapper;
         this._snapshotIdsStack = [];
     }
-    // TODO: In order to run these tests on an actual node, we should check if we are running against
-    // TestRPC, if so, use snapshots, otherwise re-deploy contracts before every test
```
fabioberger: 🍾 😆

packages/dev-utils/src/web3_factory.ts
```diff
@@ -19,16 +19,22 @@ export interface Web3Config {
...
     rpcUrl?: string; // default: localhost:8545
+    shouldUseFakeGasEstimate?: boolean; // default: true
 }
 
 export const web3Factory = {
     getRpcProvider(config: Web3Config = {}): ProviderEngine {
         const provider = new ProviderEngine();
         const hasAddresses = _.isUndefined(config.hasAddresses) || config.hasAddresses;
+        config.shouldUseFakeGasEstimate =
+            _.isUndefined(config.shouldUseFakeGasEstimate) || config.shouldUseFakeGasEstimate;
```
fabioberger: We should probably have `shouldUseFakeGasEstimate` default to `false`. Currently it defaults to `true`.

packages/devnet/README.md
```diff
@@ -0,0 +1,75 @@
+## 0x Devnet
```
fabioberger: Hm, are we sold on this name?

packages/devnet/README.md
```diff
@@ -0,0 +1,75 @@
...
+primary way you are expected to interact with the devnet. The following
+endpoints are supported: `personal,db,eth,net,web3,txpool,miner,debug`.
+
+You can stop the network with `docker stop` and it will automatically clean up
+after itself. (`docker stop` typically requires you to use `docker ps` to find
+the name of the currently running container).
+
+### Configuration
+
+The devnet network only has a single node and using PoA instead of PoW. That
```
fabioberger: nit: `and using` -> `and uses`

packages/devnet/run.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# Create log directory for Geth
+mkdir -p /var/log
+
+# Start Geth in background and redirect output to log file
+/geth --verbosity 5 --datadir node0/ --syncmode 'full' --nat none --nodiscover --port 30310 --txpool.journal '' \
+    --rpc --rpcaddr '0.0.0.0' --rpcport 8501 --rpcapi 'personal,db,eth,net,web3,txpool,miner,debug' \
+    --networkid 50 --gasprice '2000000000' --targetgaslimit '0x4c4b400000' --mine --etherbase '0xe8816898d851d5b61b7f950627d04d794c07ca37' \
+    --unlock '0xe8816898d851d5b61b7f950627d04d794c07ca37,0x5409ed021d9299bf6814279a6a1411a7e866a631,0x6ecbe1db9ef729cbe972c83fb886247691fb6beb,0xe36ea790bc9d7ab70c55260c66d52b1eca985f84,0xe834ec434daba538cd1b9fe1582052b880bd7e63,0x78dc5d2d739606d31509c31d654056a45185ecb6,0xa8dda8d7f5310e4a9e24f8eba77e091ac264f872,0x06cef8e666768cc40cc78cf93d9611019ddcb628,0x4404ac8bd8f9618d27ad2f1485aa1b2cfd82482d,0x7457d5e02197480db681d3fdf256c7aca21bdc12,0x91c987bf62d25945db517bdaa840a6c661374402' \
+    --password=node0/password.txt \
```
fabioberger: Can we put each flag on it's own line? It would help with readability quite a bit.

packages/devnet/run.sh
```diff
@@ -0,0 +1,28 @@
...
+    > /var/log/geth &
+
+# Wait for Geth to unlock the first account
+sleep 10
+
+# Send some transactions.
+# Note(albrow): 🐉 We have to do this so that debug.setHead works correctly.
+# (Geth does not seem to like debug.setHead(0), so by sending some transactions
+# we increase the current block number beyond 0). Additionally, some tests seem
+# to break when there are fewuer than 3 blocks in the chain. (We have no idea
```
fabioberger: Typo: `fewuer`

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -403,8 +421,18 @@ export class Web3Wrapper {
         }
         return receipt;
     }
+    /**
+     * Sets the current head of the local chain by block number. Note, this is a
+     * destructive action and may severely damage your chain. Use with extreme
+     * caution.
```
fabioberger: Let's mention to JSON RPC action name: `debug_setHead`. Does Parity also implement this? If not, we might want to mention that it's Geth specific, maybe even throw if backing node is not Geth?

packages/contracts/src/utils/constants.ts
LogvinovLeon: ERC721 doesn't have the notion of balance. It's either 1 or 0. It would be better to have smth like `You must own that NFT to complete transfer`

albrow: Intuitively it seems like it. But I pulled it directly from __ER20Token.sol__ and it looks like __ERC721Token.sol__ doesn't use it.

packages/contracts/src/utils/increase_time.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { constants } from './constants';
+import { web3Wrapper } from './web3_wrapper';
+
+let firstAccount: string | undefined;
+
+/**
+ * Increases time by the given number of seconds and then mines a block so that
+ * the current block timestamp has the offset applied.
+ * @param seconds the Promise which is expected to reject
```
LogvinovLeon: Isn't it a number?

packages/contracts/src/utils/increase_time.ts
albrow: @LogvinovLeon suggested it. It's actually good because it prevents any possible side effects (e.g. balances changing when you increase time).

packages/web3-wrapper/src/web3_wrapper.ts
LogvinovLeon: It will throw the error like: 'debug_setHead is not supported'

packages/contracts/src/utils/web3_wrapper.ts
albrow: 👍

packages/devnet/README.md
albrow: I like it because "testnet" has a generally accepted meaning that is slightly different. A testnet is typically open to the public, has some faucet mechanisms, etc. I'm open to changing it though.

packages/devnet/README.md
albrow: 👍

packages/web3-wrapper/src/web3_wrapper.ts
albrow: 👍 I added some clarity to the comment.

We already will throw if the backing provider doesn't support the `debug_setHead` endpoint. I think this is sufficient. Furthermore, what we do now is a form of feature detection (does this provider support this feature?) as opposed to specifically detecting the provider type (is this provider Geth?). The former is generally preferable if possible because it allows for adapting to future changes. If Parity or another node adds support in the future, we would incorrectly be throwing in the latter case.

packages/contracts/src/utils/increase_time.ts
albrow: 👍

packages/dev-utils/src/web3_factory.ts
albrow: I think we should actually keep it as defaulting to `true` until we investigate the effects of using the Ganache gas estimator in the near future. It might turn out that we don't need the fake gas estimator anymore, but historically it's always been on. Does that sound okay?

packages/contracts/test/exchange/core.ts
albrow: 👍 @fabioberger I added the `// HACK:` comments here.

However, I'm not sure a constant increases code clarity in this case. These numbers were all created on a case-by-case basis for each test (by slowly increasing the gas until I found something that works). Some of the values happen to be the same now, but that's somewhat a coincidence. There was variance between each test case in terms of the actual amount of gas used and the values might change again in the future if the tests ever change.

With the comment added, I think it's pretty clear what these numbers do and why they are needed. Does it really make sense to create constants named something like `GAS_LIMIT_FOR_FILL_ORDER_NO_THROW_SHOULD_TRANSFER_THE_CORRECT_AMOUNTS`?

packages/devnet/run.sh
albrow: 👍

packages/contracts/src/utils/assertions.ts
fabioberger: Yeah, I believe that's what we do in the rest of the codebase. Only for consistency's sake.

packages/dev-utils/src/web3_factory.ts
fabioberger: Yep, makes sense!

packages/contracts/src/utils/assertions.ts
albrow: I tried changing it and prettier changed it back to what it is currently. I think that means I should leave it that way.

packages/web3-wrapper/src/web3_wrapper.ts
fabioberger: Makes sense! Great!

PR #623, opened 2018-05-25
Implement the metamask portion of the onboarding flow

fragosti:
## Description

Complete the logic for the portion of the on-boarding flow that may or may not appear depending on app / user state (ie. is metamask installed and is it locked?). This was a bit tricky because you need to automatically change the state of the onboarding in response to a metamask log in / log out etc... but it works.

This branch also adds an abstraction to allow us to persist an arbitrary portion of our redux state to local storage and is using this abstraction to save whether the user has seen onboarding before. This powers the "auto-show" feature of onboarding, where a user will automatically be put through onboarding the first time they visit the site.

## How Has This Been Tested?
Manually.

## Types of changes

## Checklist:

*   [] Labeled this PR with the 'WIP' label if it is a work in progress.


packages/website/ts/containers/portal_onboarding_flow.ts
```diff
@@ -11,22 +11,39 @@ interface PortalOnboardingFlowProps {}
...
+    userAddress: string;
+    providerType: ProviderType;
+    injectedProviderName: string;
+    blockchainIsLoaded: boolean;
+    hasBeenSeen: boolean;
 }
 
 interface ConnectedDispatch {
-    onClose: () => void;
+    setIsRunning: (isRunning: boolean) => void;
```
BMillman19: should we prefix these with `update` so that we are consistent with the associated redux types?

packages/website/ts/index.tsx
```diff
@@ -75,7 +77,19 @@ const LazyOrderUtilsDocumentation = createLazyComponent('Documentation', async (
...
+    reducer,
+    stateStorage.getPersistedDefaultState(),
+    devToolsEnhancer({ name: '0x Website Redux Store' }),
+);
+store.subscribe(
+    _.throttle(() => {
+        stateStorage.saveState({
+            hasPortalOnboardingBeenSeen: store.getState().hasPortalOnboardingBeenSeen,
+        });
+    }, 1000),
```
BMillman19: lets put `1000` into a named constant

packages/website/ts/local_storage/local_storage.ts
```diff
@@ -26,6 +26,20 @@ export const localStorage = {
         }
         window.localStorage.removeItem(key);
     },
+    getObject(key: string): object {
```
BMillman19: return type should be `object | undefined`

fragosti:
Addressing feedback in #635. So closing.

packages/website/ts/local_storage/local_storage.ts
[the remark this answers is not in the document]
fragosti: Letting it throw now.

PR #626, opened 2018-05-25
Add support for browser websocket client and fix multiple subscriptions bug

BMillman19:
## Description

- Removes `WebSocketOrderbookChannel` from the public interface and replaces it with `orderbookChannelFactory`
- Separates the responsibility of initial connection to the websocket and subsequent message parsing / handling between `orderbookChannelFactory` and `WebSocketOrderbookChannel` respectively
- subscriptions to an `OrderbookChannel` will always be made after initial connection with this API
- Change the implementation of `WebSocketOrderbookChannel` to use `WebSocket.w3cwebsocket`

## Motivation and Context

We need real time dapps

## How Has This Been Tested?

Tested with branches of portal and the 0x starter project

## Types of changes

## Checklist:


packages/connect/src/orderbook_channel_factory.ts
```diff
@@ -0,0 +1,26 @@
+import * as WebSocket from 'websocket';
+
+import { OrderbookChannel, WebsocketClientEventType } from './types';
```
dekz: `WebsocketClientEventType` is unused

packages/connect/src/orderbook_channel_factory.ts
```diff
@@ -0,0 +1,26 @@
...
+import { assert } from './utils/assert';
+import { WebSocketOrderbookChannel } from './ws_orderbook_channel';
+
+export const orderbookChannelFactory = {
+    /**
+     * Instantiates a new WebSocketOrderbookChannel instance
+     * @param   url                 The relayer API base WS url you would like to interact with
```
dekz: bit of weird spacing here

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -55,112 +49,84 @@ export class WebSocketOrderbookChannel implements OrderbookChannel {
...
-            });
-            this._client.on(WebsocketClientEventType.ConnectFailed, error => {
-                callback(error, undefined);
-            });
-            this._client.connect(this._apiEndpointUrl);
+    /**
+     * Send a message to the client if it has been instantiated and it is open
+     */
+    private _sendMessage(message: any): void {
+        if (this._client.readyState === WebSocket.w3cwebsocket.OPEN) {
```
dekz: should this throw if the websocket some how closes? Like a connection drop out?

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -55,112 +49,84 @@ export class WebSocketOrderbookChannel implements OrderbookChannel {
...
+            this._alertAllHandlersToError(new Error(`Message does not contain utf8Data`));
+            return;
+        }
+        // try to parse the message data and route it to the correct handler
+        try {
+            const utf8Data = message.data;
+            const parserResult = orderbookChannelMessageParser.parse(utf8Data);
+            const subscription = this._subscriptions[parserResult.requestId];
+            if (_.isUndefined(subscription)) {
+                this._alertAllHandlersToError(new Error(`Message has unknown requestId: ${utf8Data}`));
```
dekz: I think we should just console log this and drop move on, rather than alert every subscription. There's not much a subscription could do about it.

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -55,112 +49,84 @@ export class WebSocketOrderbookChannel implements OrderbookChannel {
...
+    }
+    private _alertAllHandlersToClose(): void {
+        _.forEach(this._subscriptions, subscription => {
+            subscription.handler.onClose(this, subscription.subscriptionOpts);
+        });
+    }
+    private _handleWebSocketMessage(message: any): void {
+        // if we get a message with no data, alert all handlers and return
+        if (_.isUndefined(message.data)) {
+            this._alertAllHandlersToError(new Error(`Message does not contain utf8Data`));
```
dekz: I think console log this and drop it.

packages/connect/src/utils/assert.ts
```diff
@@ -0,0 +1,25 @@
+import { assert as sharedAssert } from '@0xproject/assert';
+// We need those two unused imports because they're actually used by sharedAssert which gets injected here
```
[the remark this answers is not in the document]
fabioberger: This does seems strange. `@0xproject/assert` does import them... Can we double check that this is needed? If so, please add more of a `// HACK: Explanation` above.

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -55,112 +49,84 @@ export class WebSocketOrderbookChannel implements OrderbookChannel {
...
-        assert.isFunction('handler.onClose', _.get(handler, 'onClose'));
-        this._subscriptionCounter += 1;
+            handler,
+        };
+        this._subscriptions.push(newSubscription);
         const subscribeMessage = {
             type: 'subscribe',
             channel: 'orderbook',
-            requestId: this._subscriptionCounter,
+            requestId: this._subscriptions.length - 1,
```
fabioberger: Should we increment here or use a random hash instead? I'm concerned about how unsubscribing, then re-subscribing would result in the same `requestId` and the backend server would think it was communicating with the same subscription...

packages/connect/src/ws_orderbook_channel.ts
fabioberger: Why @dekz?

packages/connect/src/ws_orderbook_channel.ts
fabioberger: But this is a developer error and should be fixed no? Throwing forces it to be fixed.

packages/connect/src/ws_orderbook_channel.ts
dekz: This is changing to a UUID in v2 of SRA.

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -1,166 +1,100 @@
...
-                }
-                callback(undefined, this._connectionIfExists);
-            });
-            this._client.on(WebsocketClientEventType.ConnectFailed, error => {
-                callback(error, undefined);
-            });
-            this._client.connect(this._apiEndpointUrl);
+    private _handleWebSocketMessage(message: any): void {
+        if (_.isUndefined(message.data)) {
+            this._handler.onError(this, new Error(`Message does not contain utf8Data`));
```
dekz: Any reason for  `utf8data` rather than `data`?

packages/connect/src/ws_orderbook_channel.ts
```diff
@@ -1,166 +1,100 @@
...
+                case OrderbookChannelMessageTypes.Snapshot: {
+                    this._handler.onSnapshot(this, subscriptionOpts, parserResult.payload);
+                    break;
+                }
+                case OrderbookChannelMessageTypes.Update: {
+                    this._handler.onUpdate(this, subscriptionOpts, parserResult.payload);
+                    break;
+                }
+                default: {
+                    this._handler.onError(this, new Error(`Message has unknown type parameter: ${utf8Data}`));
```
dekz: This feels nicer. One handler gets this error and maybe can do something about it. We do have an available subscriptionOpts at this point so we could be more explicit about the subscription which caused this message (which may allow a handler to take an action like re-subscribe). It would require an optional parameter in the function definition.

```
onError: (
         err: Error,
         subscriptionOpts?: OrderbookChannelSubscriptionOpts,
     ) => void;
```

Thoughts on this?

packages/connect/src/types.ts
```diff
@@ -11,17 +11,10 @@ export interface Client {
 }
 
 export interface OrderbookChannel {
-    subscribe: (subscriptionOpts: OrderbookChannelSubscriptionOpts, handler: OrderbookChannelHandler) => void;
+    subscribe: (subscriptionOpts: OrderbookChannelSubscriptionOpts) => void;
     close: () => void;
```
dekz: should the `uri` be discoverable here with a getter? It seems like it would be hard to track down which uri was spitting out errors if you were subscribed to multiple channels.

BMillman19: hmm i feel that it does break encapsulation somewhat, perhaps we can have it in the error message itself

PR #627, opened 2018-05-26
On-Chain AssetData Decoding Lib + safeTransferFrom for ERC721 + Memcpy

hysz:
## Description

Asset data decoding has been moved out of the proxies and into a library. Also implemented safeTransferFrom with additional receiver data for ERC721. Extended LibBytes to support nested byte arrays, which is how we decode ERC721 receiver data. Created LibMem, which includes a generic memcpy that is now used by LibBytes.

-- Updates--
An updated LibMem is now in this PR, which includes an optimized version of Memcpy that supports overlapping memory ranges (courtesy of @Recmo).

Also several different names were being used to describe asset data: asset metadata, proxy data, proxy metadata. All instances have been replaced with asset data.

## Motivation and Context

The decoder library is particularly useful for the forwarding contract. It also allows for more granular testing of asset data. 

Using ERC721's safeTransferFrom makes transferring tokens to contracts more flexible and foolproof. 
 
Memcpy is used for an update to LibBytes, which allows us to store nested byte arrays. This is used to store data for an ERC721 token receiver.

## How Has This Been Tested?

Included are tests for the decoder library, ERC721 receiver contracts, LibBytes.readBytes/writeBytes,  and Memcpy. All existing tests still pass.

## Types of changes

## Checklist:

*   [X] Added tests to cover my changes.
*   [X] Labeled this PR with the 'WIP' label if it is a work in progress.


packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+            // Compute number of complete words to copy + remaining bytes
+            let lenFullWords := div(add(length, 0x1F), 0x20)
+            let remainder := mod(length, 0x20)
+            if gt(remainder, 0) {
+                lenFullWords := sub(lenFullWords, 1)
+            }
+
+            // Copy full words from source to dest
+            let offset := 0
+            let maxOffset := mul(0x20, lenFullWords)
```
Recmo: We can safely use `length` here instead of `maxOffset`. The value will be slightly larger than `maxOffset` by `remainder`, but this is okay because every iteration we increase with `0x20` which is more than `remainder` so the same number of loop iterations will occur.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+            for {offset := 0} lt(offset, maxOffset) {offset := add(offset, 0x20)} {
```
Recmo: `{offset := 0}` is redundant. `offset` was already cleared above.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+                mstore(add(dest, offset),  mload(add(source, offset)))
```
Recmo: We can take one `add` out of the loop by having the loop run from `dest` to `dest + length`.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+            }
+
+            // Copy remaining bytes
+            if gt(remainder, 0) {
```
Recmo: If `length` is at least one full word, there is another trick we can do: Copy the last 32 bytes. There will be some overlap, but this is okay.

If `length` is less than a full word, and we can not overwrite the data after it, then we can either use `mstore8` or we need to read the trailing data and incorporate it in our `mstore`.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+                // We only want to keep the X bytes, zeroing out the remaining bytes.
+                // We accomplish this by a right shift followed by a left shift.
+                // Example:
+                //   Suppose a word of 8 bits has all 1's: [11111111]
+                //   Let X = 7 (we want to copy the first 7 bits)
+                //   Apply a right shift of 1: [01111111]
+                //   Apply a left shift of 1: [11111110]
+                let sourceShiftFactor := exp(2, mul(8, sub(0x20, remainder)))
+                let sourceWord := mload(add(source, offset))
+                let sourceBytes := mul(div(sourceWord, sourceShiftFactor), sourceShiftFactor)
```
Recmo: `mul`s and `div`s are more expensive than `and`s. It would be more efficient to use bitmasks instead of shifts.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+        uint256 dest,
+        uint256 source,
+        uint256 length
+    )
+        internal
+        pure
+    {
+        // Base cases
+        if(length == 0) return;
+        if(source == dest) return;
```
Recmo: There is an edge case when `[source, source+length)` and `[dest, dest+length)` overlap. If `dest < source` it should be fine, but currently the other case is not correctly handled.

packages/contracts/src/contracts/current/test/TestLibMem/TestLibMem.sol
```diff
@@ -0,0 +1,231 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.4.24;
+
+import "../../utils/LibMem/LibMem.sol";
+import "../../utils/LibBytes/LibBytes.sol";
+
+contract TestLibMem is
```
Recmo: This implements the test logic in Solidity, which is a break from how we usually test this. Instead we could write a contract that exposes a function `TestMemCopy(bytes memory, uint256 from, uint256 to, uint256 length) returns (bytes memory)`.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,104 @@
...
+            if gt(remainder, 0) {
+                // Read a full word from source, containing X bytes to copy to dest.
+                // We only want to keep the X bytes, zeroing out the remaining bytes.
+                // We accomplish this by a right shift followed by a left shift.
+                // Example:
+                //   Suppose a word of 8 bits has all 1's: [11111111]
+                //   Let X = 7 (we want to copy the first 7 bits)
+                //   Apply a right shift of 1: [01111111]
+                //   Apply a left shift of 1: [11111110]
+                let sourceShiftFactor := exp(2, mul(8, sub(0x20, remainder)))
```
Recmo: `2^(8 * x) == (2^8)^x` =>  `exp(256, sub(0x20, remainder))`

packages/contracts/src/contracts/current/utils/LibAssetProxyDecoder/LibAssetProxyDecoder.sol
```diff
@@ -0,0 +1,74 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.4.24;
+pragma experimental ABIEncoderV2;
+
+import "../LibBytes/LibBytes.sol";
+
+contract LibAssetProxyDecoder is
```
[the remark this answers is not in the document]
hysz: That's a good point. I've moved the logic back into the proxies, but kept it as separate functions. That way it can still be used by the forwarding contract.

packages/contracts/src/contracts/current/protocol/AssetProxy/ERC721Proxy.sol
```diff
@@ -35,24 +37,31 @@ contract ERC721Proxy is
...
 
     /// @dev Internal version of `transferFrom`.
-    /// @param assetMetadata Encoded byte array.
+    /// @param proxyData Encoded byte array.
     /// @param from Address to transfer asset from.
     /// @param to Address to transfer asset to.
     /// @param amount Amount of asset to transfer.
     function transferFromInternal(
-        bytes memory assetMetadata,
+        bytes memory proxyData,
```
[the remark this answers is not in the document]
hysz: Updated this across all contracts/tests.

packages/contracts/src/contracts/current/protocol/AssetProxy/ERC721Proxy.sol
```diff
@@ -62,19 +71,9 @@ contract ERC721Proxy is
...
-        );
-        address token = readAddress(assetMetadata, 1);
-        uint256 tokenId = readUint256(assetMetadata, 21);
-
         // Transfer token.
         // Either succeeds or throws.
-        // @TODO: Call safeTransferFrom if there is additional
-        //        data stored in `assetMetadata`.
-        ERC721Token(token).transferFrom(from, to, tokenId);
+        ERC721Token(token).safeTransferFrom(from, to, tokenId, data);
```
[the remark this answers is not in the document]
hysz: Seems slightly less foolproof, but after doing some tests it looks like we'll save several thousand gas. Even in the case where the receiving address belongs to an EOA, it costed an extra ~4000 gas to call `safeTransferFrom` 🤔...

hysz:
@Recmo -- Merged [#632](https://github.com/0xProject/0x-monorepo/pull/632) into this branch! Really nice work :). It's so solid I almost didn't want to touch anything, but I did address the [wraparound edge case](https://github.com/0xProject/0x-monorepo/pull/632/files#diff-5ee7377abf6921bebaf25d073e3177afR104) in [this commit](https://github.com/0xProject/0x-monorepo/pull/627/commits/fbafd8b4747d63bc0942dd7f8bfa2a2155c9002e). Feel free to give feedback or swap out my solution if you have a better way to handle this.

packages/contracts/test/exchange/match_orders.ts
```diff
@@ -161,16 +161,16 @@ describe('matchOrders', () => {
             // Create orders to match
             const signedOrderLeft = orderFactoryLeft.newSignedOrder({
                 makerAddress: makerAddressLeft,
-                makerAssetData: assetProxyUtils.encodeERC20ProxyData(defaultERC20MakerAssetAddress),
-                takerAssetData: assetProxyUtils.encodeERC20ProxyData(defaultERC20TakerAssetAddress),
+                makerAssetData: assetProxyUtils.encodeERC20AssetData(defaultERC20MakerAssetAddress),
```
dekz: does this have defaults so the assetData doesn't need to be specified every time, only when it needs to be changed from the default?

packages/contracts/src/contracts/current/protocol/AssetProxy/ERC721Proxy.sol
```diff
@@ -86,4 +85,44 @@ contract ERC721Proxy is
...
+            LENGTH_AT_LEAST_53_REQUIRED
+        );
+
+        // Decode asset data.
+        token = readAddress(assetData, 0);
+        tokenId = readUint256(assetData, 20);
+        if (length > 53) {
+            receiverData = readBytes(assetData, 52);
+        }
+        proxyId = uint8(assetData[length-1]);
```
dekz: `length - 1`

packages/contracts/test/asset_proxy/decoder.ts
```diff
@@ -0,0 +1,111 @@
+import { BlockchainLifecycle, devConstants, web3Factory } from '@0xproject/dev-utils';
+import { assetProxyUtils, generatePseudoRandomSalt } from '@0xproject/order-utils';
+import { AssetData, AssetProxyId, ERC20AssetData, ERC721AssetData } from '@0xproject/types';
+import { BigNumber } from '@0xproject/utils';
+import { Web3Wrapper } from '@0xproject/web3-wrapper';
+import BN = require('bn.js');
+import * as chai from 'chai';
+import { LogWithDecodedArgs, TransactionReceiptWithDecodedLogs } from 'ethereum-types';
```
dekz: I think there are some unnecessary imports here as we aren't testing LogWithDecodedArgs?

packages/contracts/test/asset_proxy/proxies.ts
```diff
@@ -1,26 +1,37 @@
...
-import { provider, web3Wrapper } from '../../src/utils/web3_wrapper';
+import { LogDecoder } from '../../src/utils/log_decoder';
+import { provider, txDefaults, web3Wrapper } from '../../src/utils/web3_wrapper';
 
 chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
 describe('Asset Transfer Proxies', () => {
+    const nullDataHex = '0x';
```
dekz: `constants.NULL_BYTES`

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,140 @@
...
+                    let dend := add(dest, length)
+
+                    // Remember the last 32 bytes of source
+                    // This needs to be done here and not after the loop
+                    // because we may have overwritten the last bytes in
+                    // source already due to overlap.
+                    let last := mload(send)
+
+                    // Copy whole words front to back
+                    for {let i := 0} lt(i, nwords) {i := add(i, 1)} {
```
Recmo: I don't think we need `i` here, we can loop as long as `source < send` or alternatively `dest < dend`. This saves one `ADD` per loop iteration, plus we don't need `i` and `nwords` anymore.

The explanation of why this works and why it requires `send` and `dend` to be `32` bytes less than `length` is a bit complex, but we can put that in a comment.

packages/contracts/test/asset_proxy/proxies.ts
```diff
@@ -264,21 +268,106 @@ describe('Asset Transfer Proxies', () => {
...
+                const amount = new BigNumber(1);
+                return expect(
+                    erc721Proxy.transferFrom.sendTransactionAsync(
+                        encodedAssetData,
+                        makerAddress,
+                        erc20Proxy.address, // the ERC20 proxy does not have an ERC721 receiver
+                        amount,
+                        { from: exchangeAddress },
+                    ),
+                ).to.be.rejectedWith(constants.REVERT);
```
dekz: This may now need to be changed with the latest Geth testing changes. [Details](https://github.com/0xProject/0x-monorepo/blob/785b9811f30869f01242ce9ff81c282bf7f5352f/packages/contracts/src/utils/assertions.ts#L39)

packages/contracts/test/libraries/lib_bytes.ts
```diff
@@ -1,4 +1,6 @@
 import { BlockchainLifecycle, devConstants, web3Factory } from '@0xproject/dev-utils';
+import { assetProxyUtils } from '@0xproject/order-utils';
+import { generatePseudoRandomSalt } from '@0xproject/order-utils';
```
dekz: combine these imports

packages/contracts/test/libraries/lib_bytes.ts
```diff
@@ -293,4 +324,174 @@ describe('LibBytes', () => {
...
+            // The length of the nested array is 32 bytes. By storing less than 32 bytes, a length cannot be read.
+            const offset = new BigNumber(0);
+            return expect(libBytes.publicReadBytes.callAsync(byteArrayShorterThan32Bytes, offset)).to.be.rejectedWith(
+                constants.REVERT,
+            );
+        });
+
+        it('should fail if we store a nested byte array length, without a nested byte array)', async () => {
+            const offset = new BigNumber(0);
+            return expect(libBytes.publicReadBytes.callAsync(testBytes32, offset)).to.be.rejectedWith(constants.REVERT);
```
dekz: this may also need to be changed with the new failure assertions.

packages/order-utils/src/asset_proxy_utils.ts
```diff
@@ -100,50 +114,67 @@ export const assetProxyUtils = {
...
-        const addressOffset = ERC20_PROXY_METADATA_BYTE_LENGTH - 1;
-        const encodedTokenAddress = encodedProxyMetadata.slice(0, addressOffset);
         const tokenAddress = assetProxyUtils.decodeAddress(encodedTokenAddress);
-        const tokenIdOffset = ERC721_PROXY_METADATA_BYTE_LENGTH - 1;
-        const encodedTokenId = encodedProxyMetadata.slice(addressOffset, tokenIdOffset);
+        const encodedTokenId = encodedAssetData.slice(tokenIdOffset, receiverDataLengthOffset);
         const tokenId = assetProxyUtils.decodeUint256(encodedTokenId);
-        const erc721ProxyData = {
+        const nullData = '0x';
+        let receiverData = nullData;
```
dekz: Can drop nullData here and just assign `constants.NULL_BYTE` to receiverData

packages/order-utils/src/asset_proxy_utils.ts
```diff
@@ -65,33 +66,46 @@ export const assetProxyUtils = {
...
             throw new Error(
-                `Could not decode ERC20 Proxy Data. Expected length of encoded data to be 53. Got ${
-                    encodedProxyMetadata.byteLength
+                `Could not decode ERC20 Proxy Data. Expected length of encoded data to be at least 53. Got ${
+                    encodedAssetData.byteLength
                 }`,
             );
         }
-        const encodedAssetProxyId = encodedProxyMetadata.slice(-1);
+        const addressOffset = 0;
```
dekz: I'd vote for turning these into constants at the top of the file like `ERC721_ASSET_DATA_BYTE_LENGTH`

packages/order-utils/src/asset_proxy_utils.ts
```diff
@@ -100,50 +114,67 @@ export const assetProxyUtils = {
...
             );
         }
-        const encodedAssetProxyId = encodedProxyMetadata.slice(-1);
+        const encodedAssetProxyId = encodedAssetData.slice(-1);
         const assetProxyId = assetProxyUtils.decodeAssetProxyId(encodedAssetProxyId);
         return assetProxyId;
     },
-    decodeProxyData(proxyData: string): ProxyData {
-        const assetProxyId = assetProxyUtils.decodeProxyDataId(proxyData);
+    decodeAssetData(assetData: string): AssetData {
```
dekz: I think this could be better with a [Union type](http://www.typescriptlang.org/docs/handbook/advanced-types.html) of ERC20AssetData or ERC721AssetData. This will parse the proxy id out and the token address out, but it doesn't tell me what it is.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
hysz: Yeah I do think that's cleaner. 

The only reason I changed it was to handle the case where `send` turns "negative" and wraps around. We hit this edge case when `0 <= source < 32 - length%32`.

Here's an example of the edge case that you can paste into Remix: https://pastebin.com/fjfh890q

packages/contracts/src/contracts/current/test/DummyERC721Receiver/DummyERC721Receiver.sol
```diff
@@ -0,0 +1,63 @@
...
+    function onERC721Received(
+        address _from,
+        uint256 _tokenId,
+        bytes _data
+    )
+        public
+        returns (bytes4)
+    {
+        emit TokenReceived(_from, _tokenId, _data);
+        return ERC721_RECEIVED;
```
fabioberger: Where is `ERC721_RECEIVED` declared?

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,140 @@
...
+            memoryAddress := input
+        }
+        return memoryAddress;
+    }
+
+    /// @dev Copies `length` bytes from memory location `source` to `dest`.
+    /// @param dest memory address to copy bytes to.
+    /// @param source memory address to copy bytes from.
+    /// @param length number of bytes to copy.
+    function memcpy(
```
fabioberger: Can we rename this to `memCopy`? This shortcut doesn't save anything but hurts readability.

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,140 @@
...
+            // We handle overlap in the source and destination range by
+            // changing the copying direction. This prevents us from
+            // overwriting parts of source that we still need to copy.
+            //
+            // This correctly handles source == dest
+            //
+            if (source > dest) {
+                assembly {
+                    // Record the total number of full words to copy
+                    let nwords := div(length, 32)
```
fabioberger: Camel-case: `nWords`

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
```diff
@@ -0,0 +1,140 @@
...
+
+                    // We subtract 32 from `send` and `dend` because it
```
fabioberger: Camel-case: `dEnd` and `sEnd`. Thought this was the english word `send` and a typo `dend`.

packages/migrations/src/migration.ts
```diff
@@ -0,0 +1,132 @@
+import { Provider, TxData } from '@0xproject/types';
```
fabioberger: This file shouldn't exist. Merge issue. The V2 migration file is now at `src/v2/migration.ts`

packages/order-utils/src/asset_proxy_utils.ts
fabioberger: Agreed.

packages/contracts/src/contracts/current/test/DummyERC721Receiver/DummyERC721Receiver.sol
hysz: In the ERC721 Receiver interface -- `tokens/ERC721Token/IERC721Receiver.sol`

hysz:
@dekz live for the rebase 💯

packages/contracts/src/contracts/current/utils/LibMem/LibMem.sol
Recmo: I did think of that in my original implementation. This won't happen because nothing is allocated at bytes [0..96) in Solidity. The first two memory slots are scratch space, the third contains the free memory pointer.

Starting from 0.5.0 a fourth will contain a permanent zero, making it 128 bytes. But this zero is also used for `bytes` and `string`s  of length zero, so we might see them in a memcpy.

Recmo: In the `dest > source` case we can also use `lt(dest, dend)`. This will make the edge case much less likely. (It will only happen if both dest and source 1) are longer than 32 bytes, 2) overlap in within 32 bytes and 3) are allocated in the first 32 bytes).

Recmo: We can also use a signed less than `slt` instead of `lt`. Valid addresses will always be less than 2^255, so they can be safely reinterpretted as a signed integer.

```
for {} slt(dest, dend) {} {
```

PR #633, opened 2018-05-29
Feature/contracts/errors

abandeali1:
## Description

- Reverts on all failures in `fillOrder` and `cancelOrder` (if any amount > 0 can be filled, that is considered a success).
- Removes a lot of logic from `matchOrders` that was made redundant with the first change.
- Encodes data in revert reasons using `string(abi.encodePacked(bytes32 err, ... ))`. Need some feedback on the format here.
  - Do we like using ABI encoding or tight packing of the arguments here? Tightly packing (which is what is currently being used) is more efficient but can only be used when all of the arguments are a fixed size (except the last one).
  - Errors are identified by the 32 byte keccak256 hash of the error string. This can probably be reduced to 4 bytes (though this doesn't matter if we use regular ABI encoding).
  - Which data should actually be encoded in the revert reasons? I've kept it pretty minimalistic and am typically not including any function args. I do think it makes sense to include identifying hashes and any state that is not immutable when the transaction is submitted.

## Types of changes

## Checklist:


Recmo:
> Errors are identified by the 32 byte keccak256 hash of the error string. This can probably be reduced to 4 bytes (though this doesn't matter if we use regular ABI encoding).

In terms of gas cost there is not much we would save by doing this, unlike the case with function calls, where bytes4 is used.

Recmo:
> Which data should actually be encoded in the revert reasons? I've kept it pretty minimalistic and am typically not including any function args. I do think it makes sense to include identifying hashes and any state that is not immutable when the transaction is submitted.

The information we provide is only interesting for smart contracts that call ours. Anyone else can use debugging to figure out what happened.

Given that, functions arguments are not useful. Mutable stuff like `fillAmounts` and cancels is. Then the calling contract can use that info to decide what to do next.

Maybe we can split the error messages into different catagories: invalid calls (caller screwed up) and operation failed (caller has bad luck). It would be useful to quickly distinguish between the two and not have to implement all our error messages.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -126,32 +124,28 @@ contract MixinExchangeCore is
...
     /// @dev Gets information about an order: status, hash, and amount filled.
     /// @param order Order to gather information on.
     /// @return OrderInfo Information about the order and its state.
-    ///                   See LibOrder.OrderInfo for a complete description.
+    ///         See LibOrder.OrderInfo for a complete description.
     function getOrderInfo(Order memory order)
         public
         view
-        returns (LibOrder.OrderInfo memory orderInfo)
+        returns (OrderInfo memory orderInfo)
```
dekz: We seem to hop between `Order memory order` and `LibOrder.Order memory order`, what's the scenario where we want to include it versus reference it?

dekz: We have a style meeting scheduled so I'll punt on the above till then.

packages/contracts/src/contracts/current/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -170,148 +164,124 @@ contract MixinExchangeCore is
...
-    /// @param takerAssetFillAmount Desired amount of order to fill by taker.
-    /// @return status Return status of calculating fill amounts. Returns Status.SUCCESS on success.
     /// @return fillResults Amounts filled and fees paid by maker and taker.
-    function calculateFillResults(
+    function updateFilledState(
         Order memory order,
-        uint8 orderStatus,
+        address takerAddress,
+        bytes32 orderHash,
         uint256 orderTakerAssetFilledAmount,
```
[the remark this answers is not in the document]
dekz: We have a style meeting scheduled so I'll punt on the above till then.

PR #637, opened 2018-05-30
Add missing EIP191 prefix for EIP712 and Execute Transaction

dekz:
## Description

Missed the EIP191 prefix while working on the previous PR.

```
encode(domainSeparator : 𝔹²⁵⁶, message : 𝕊) = "\x19\x01" ‖ domainSeparator ‖ hashStruct(message) where domainSeparator and hashStruct(message) are defined below.
```

## Types of changes

## Checklist:


packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -21,9 +21,16 @@ pragma solidity ^0.4.24;
...
-        "DomainSeparator(address contract)"
+        "DomainSeparator(",
+        "string name,",
+        "string version,",
+        "address contract",
+        ")"
     );
 
+    bytes32 constant DOMAIN_SEPARATOR_NAME_HASH = keccak256("0x Protocol");
+    bytes32 constant DOMAIN_SEPARATOR_VERSION_HASH = keccak256("1");
```
fabioberger: Shouldn't this be `2`?

dekz: This number represents our EIP712 schema, not the protocol version.

packages/contracts/test/exchange/transactions.ts
```diff
@@ -115,7 +115,7 @@ describe('Exchange transactions', () => {
         await blockchainLifecycle.revertAsync();
     });
 
-    describe('executeTransaction', () => {
+    describe.only('executeTransaction', () => {
```
BMillman19: remove dat

packages/contracts/test/exchange/core.ts
```diff
@@ -33,7 +33,7 @@ chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
-describe('Exchange core', () => {
+describe.only('Exchange core', () => {
```
BMillman19: remove dat

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,55 @@
...
+
+*/
+
+pragma solidity ^0.4.24;
+
+contract LibEIP712 {
+    string public constant EIP191_HEADER = "\x19\x01";
+    bytes32 public constant EIP712_DOMAIN_SEPARATOR_NAME_HASH = keccak256("0x Protocol");
+
+    bytes32 public constant EIP712_DOMAIN_SEPARATOR_VERSION_HASH = keccak256("1");
```
Recmo: Version 2 ?

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -73,17 +73,13 @@ contract LibOrder {
         view
         returns (bytes32 orderHash)
     {
-        // TODO: EIP712 is not finalized yet
-        // Source: https://github.com/ethereum/EIPs/pull/712
-        orderHash = keccak256(
-            DOMAIN_SEPARATOR_SCHEMA_HASH,
-            keccak256(address(this)),
-            ORDER_SCHEMA_HASH,
+        orderHash = createEIP712Message(
```
Recmo: The way EIP712 is intended is that we leave out `createEIP712Message` here and instead wrap the message in `SignatureValidator`.

Pro: getOrderHash will be simpler and cheaper.
Con: orderHash will not include the domain separator (but the signature still would).

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -34,6 +37,30 @@ contract MixinTransactions is
     // Address of current transaction signer
     address public currentContextAddress;
 
+    bytes32 constant EXECUTE_TRANSACTION_SCHEMA_HASH = keccak256(
+        "ExecuteTransaction(",
```
hysz: Should this be `executeTransaction`?

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -34,6 +37,30 @@ contract MixinTransactions is
...
+        "uint256 salt,",
+        "address signer,",
+        "bytes data",
+        ")"
+    );
+
+    function getExecuteTransactionHash(uint256 salt, address signer, bytes data)
```
hysz: Function descriptions here & below plz :)

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,62 @@
...
+
+contract LibEIP712 {
+    string public constant EIP191_HEADER = "\x19\x01";
+
+    bytes32 public constant EIP712_DOMAIN_SEPARATOR_NAME_HASH = keccak256("0x Protocol");
+
+    bytes32 public constant EIP712_DOMAIN_SEPARATOR_VERSION_HASH = keccak256("2");
+
+    bytes32 public constant EIP712_DOMAIN_SEPARATOR_SCHEMA_HASH = keccak256(
+        abi.encodePacked(
```
abandeali1: I prefer doing `keccak256(abi.encodePacked(...)) on the same line. This way we can quickly tell where it's missing, and we should be doing it everytime we call `keccak256` (recent compiler versions will warn about not doing it).

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,62 @@
...
+                        EIP712_DOMAIN_SEPARATOR_VERSION_HASH,
+                        bytes32(address(this))
+                    )
+                ),
+                hashStruct
+            )
+        );
+        return message;
+    }
+}
```
abandeali1: Newline

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -34,6 +37,30 @@ contract MixinTransactions is
     // Address of current transaction signer
     address public currentContextAddress;
 
+    bytes32 constant EXECUTE_TRANSACTION_SCHEMA_HASH = keccak256(
```
abandeali1: Add an `EIP712` prefix here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
abandeali1: I'd probably prefer just `Transaction`, but I could see how that might be confusing. Maybe `ZeroExTransaction`?

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,62 @@
...
+    );
+
+    function createEIP712Message(bytes32 hashStruct)
+        internal
+        view
+        returns (bytes32 message)
+    {
+        // TODO: EIP712 is not finalized yet
+        // Source: https://github.com/ethereum/EIPs/pull/712
+        // TODO: Cache the Domain Separator
```
abandeali1: Let's go ahead and cache this as a state variable in a constructor, will be much easier to read :)

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
dekz: I've split the two so we can decide what to do without much modification

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,55 @@
...
+        "string version,",
+        "address verifyingContract",
+        ")"
+    ));
+
+    // Hash of the EIP712 Domain Separator data
+    bytes32 public EIP712_DOMAIN_HASH;
+
+    constructor ()
+    public
```
abandeali1: Missing an indent here.

abandeali1:
Looks good after fixing linting errors.

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -34,6 +37,33 @@ contract MixinTransactions is
...
+    function hashZeroExTransaction(uint256 salt, address signer, bytes data)
+        internal
+        view
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked(
+            EIP712_ZEROEX_TRANSACTION_SCHEMA_HASH,
+            salt,
+            bytes32(signer),
+            keccak256(abi.encodePacked(data))
```
Recmo: `abi.encodedPacked` is redundant here. `keccak256(data)` is sufficient.

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -0,0 +1,55 @@
...
+    {
+        EIP712_DOMAIN_HASH = keccak256(abi.encodePacked(
+            EIP712_DOMAIN_SEPARATOR_SCHEMA_HASH,
+            keccak256(abi.encodePacked("0x Protocol")),
+            keccak256(abi.encodePacked("2")),
```
Recmo: `keccak256(bytes("0x Protocol"))`, `keccak256(bytes("2"))` and `bytes32(this)` are sufficient.

Recmo: Actually, if you use `abi.encode` instead of the packed one, you don't even need the `bytes32` around addresses. Like here: https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.sol#L47

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -85,27 +87,32 @@ contract LibOrder {
...
+            bytes32(order.senderAddress),
+            order.makerAssetAmount,
+            order.takerAssetAmount,
+            order.makerFee,
+            order.takerFee,
+            order.expirationTimeSeconds,
+            order.salt,
+            keccak256(abi.encodePacked(order.makerAssetData)),
+            keccak256(abi.encodePacked(order.takerAssetData))
+        ));
```
Recmo: Same here, we can drop all the `abi.encodePacked` and `bytes32` and use a `abi.encode`. It will likely safe us some gas, because typecastings are not very efficient in solidity.

packages/order-utils/src/eip712_utils.ts
```diff
@@ -0,0 +1,95 @@
...
+            } else if (parameter.type === EIP712Types.Uint256) {
+                encodedValues.push(value);
+            } else if (parameter.type === EIP712Types.Address) {
+                encodedValues.push(EIP712Utils.pad32Address(value));
+            } else {
+                throw new Error(`Unable to encode ${parameter.type}`);
+            }
+        }
+        return encodedValues;
+    },
```
Recmo: There is also `rawEncode` from `ethereumjs-abi` that can be used here, like in this example https://github.com/ethereum/EIPs/blob/master/assets/eip-712/Example.js#L82. Advantage of using `rawEncode` is that it automatically handles all the different number types from Solidity correctly.

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
dekz: Discussed on slack the issue from doing this

packages/order-utils/src/eip712_utils.ts
dekz: This doesn't seem to like our BigNumbers. Can this be a future TODO or is it blocking the merge?

Recmo: This can for sure be a TODO.

I noticed the BigNumber get's confused when there are multiple instances of the library. It is very hard / impossible to synchronize those.

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
fabioberger: If you could add a comment to that effect that would be great.

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -18,28 +18,30 @@
...
+            "uint256 makerAssetAmount,",
+            "uint256 takerAssetAmount,",
+            "uint256 makerFee,",
+            "uint256 takerFee,",
+            "uint256 expirationTimeSeconds,",
+            "uint256 salt,",
+            "bytes makerAssetData,",
+            "bytes takerAssetData",
+            ")"
+        )
```
abandeali1: End parens should be on the same line here.

PR #640, opened 2018-05-30
Add awaitTransactionSuccess where needed

albrow:
See #601 and #613 for some background. We had some code pushed in the last few days that didn't follow the practice of calling `awaitTransactionSuccess` when needed.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/utils/token_registry_wrapper.ts
+++ b/packages/contracts/src/utils/token_registry_wrapper.ts
@@ -4,13 +4,16 @@ import { TokenRegistryContract } from '../contract_wrappers/generated/token_regi
 
 import { Token } from './types';
 
+import { constants } from './constants';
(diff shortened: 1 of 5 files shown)


packages/contracts/src/utils/token_registry_wrapper.ts
```diff
@@ -4,13 +4,16 @@ import { TokenRegistryContract } from '../contract_wrappers/generated/token_regi
 
 import { Token } from './types';
 
+import { constants } from './constants';
+import { web3Wrapper } from './web3_wrapper';
```
abandeali1: Can we pass in a provider and instantiate a new web3Wrapper instance here like we do in the other wrappers?

albrow: 👍

PR #644, opened 2018-05-31
Move TokenState fetching logic up into Portal.tsx so it can be used by Wallet and PortalOnboardingFlow

fragosti:
## Description

I noticed that there is similar "TokenState" (balance and allowances for tokens) fetching logic in wallet.tsx, token_balances.tsx and that I would need the same information in portal_onboarding_flow.tsx.

I initially thought about doing a major refactor (https://app.asana.com/0/628666249318202/692511197180876) but decided that this was out of scope for now so just moved the logic into portal.tsx which has all 3 of the components mentioned above. I moved the same logic out of wallet.tsx but kept it in token_balances.tsx since that is used in legacy portal and won't be around for long.

I also went ahead and implemented what is required for the WETH onboarding step, which was easy once I had access to the users WETH balance.

## How Has This Been Tested?

😁

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
+++ b/packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
@@ -69,13 +69,13 @@ export class PortalOnboardingFlow extends React.Component<PortalOnboardingFlowPr
                 target: '.eth-row',
                 content: 'Before you begin you will need to send some ETH to your metamask wallet',
                 placement: 'right',
-                continueButtonDisplay: this._userHasEth() ? 'enabled' : 'disabled',
+                continueButtonDisplay: this._userHasVisibleEth() ? 'enabled' : 'disabled',
             },
             {
                 target: '.weth-row',
                 content: 'You need to convert some of your ETH into tradeable Wrapped ETH (WETH)',
                 placement: 'right',
-                continueButtonDisplay: this._userHasWeth() ? 'enabled' : 'disabled',
+                continueButtonDisplay: this._userHasVisibleWeth() ? 'enabled' : 'disabled',
             },
         ];
         return steps;
@@ -85,11 +85,11 @@ export class PortalOnboardingFlow extends React.Component<PortalOnboardingFlowPr
         return !_.isEmpty(this.props.userAddress);
     }
 
-    private _userHasEth(): boolean {
+    private _userHasVisibleEth(): boolean {
         return this.props.userEtherBalanceInWei > new BigNumber(0);
     }
 
-    private _userHasWeth(): boolean {
+    private _userHasVisibleWeth(): boolean {
         const ethToken = utils.getEthToken(this.props.tokenByAddress);
         if (!ethToken) {
             return false;
--- a/packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
+++ b/packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 
 import { BigNumber } from '@0xproject/utils';
 import { OnboardingFlow, Step } from 'ts/components/onboarding/onboarding_flow';
-import { ProviderType, TokenByAddress } from 'ts/types';
+import { ProviderType, TokenByAddress, TokenStateByAddress } from 'ts/types';
 import { utils } from 'ts/utils/utils';
 
 export interface PortalOnboardingFlowProps {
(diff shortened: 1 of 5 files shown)


packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
```diff
@@ -89,8 +90,12 @@ export class PortalOnboardingFlow extends React.Component<PortalOnboardingFlowPr
     }
 
     private _userHasWeth(): boolean {
-        // TODO: https://app.asana.com/0/681385331277907/690722374136933
-        return false;
+        const ethToken = utils.getEthToken(this.props.tokenByAddress);
+        if (!ethToken) {
+            return false;
+        }
+        const wethTokenState = this.props.trackedTokenStateByAddress[ethToken.address];
```
BMillman19: the `TokenState` has a `isLoaded` property that may be false, do we want to handle that case? The user may have WETH but the balance is not updated yet

packages/website/ts/components/portal/portal.tsx
```diff
@@ -487,6 +528,51 @@ export class Portal extends React.Component<PortalProps, PortalState> {
         const isSmallScreen = this.props.screenWidth === ScreenWidths.Sm;
         return isSmallScreen;
     }
+
+    private _getCurrentTrackedTokens(): Token[] {
+        return this._getTrackedTokens(this.props.tokenByAddress);
+    }
+
+    private _getTrackedTokens(tokenByAddress: TokenByAddress): Token[] {
```
BMillman19: it might be worth looking at some of these utility methods across our different components and identifying the ones that don't rely on the components state or props and moving those into a set of pure utility functions

packages/website/ts/components/portal/portal.tsx
```diff
@@ -487,6 +528,51 @@ export class Portal extends React.Component<PortalProps, PortalState> {
...
+        const allTokens = _.values(tokenByAddress);
+        const trackedTokens = _.filter(allTokens, t => t.isTracked);
+        return trackedTokens;
+    }
+
+    private _getInitialTrackedTokenStateByAddress(trackedTokens: Token[]): TokenStateByAddress {
```
BMillman19: same as above

packages/website/ts/components/portal/portal.tsx
```diff
@@ -487,6 +528,51 @@ export class Portal extends React.Component<PortalProps, PortalState> {
...
+            trackedTokenStateByAddress[token.address] = {
+                balance: new BigNumber(0),
+                allowance: new BigNumber(0),
+                isLoaded: false,
+            };
+        });
+        return trackedTokenStateByAddress;
+    }
+
+    private async _fetchBalancesAndAllowancesAsync(tokenAddresses: string[]): Promise<void> {
```
BMillman19: should we just go ahead and parallelize this?

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -65,13 +65,14 @@ export interface WalletProps {
...
     location: Location;
+    trackedTokenStateByAddress: TokenStateByAddress;
     onToggleLedgerDialog: () => void;
     onAddToken: () => void;
     onRemoveToken: () => void;
+    refetchTokenStateAsync: (tokenAddress: string) => Promise<void>;
 }
 
 interface WalletState {
-    trackedTokenStateByAddress: TokenStateByAddress;
```
BMillman19: can we get rid of the `trackedTokens` prop as well now?

BMillman19:
Also linter is failing

packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
fragosti: I just renamed the fn to be _hasVisibleWeth

packages/website/ts/components/wallet/wallet.tsx
fragosti: We cannot as we use that to render, although I agree it would be nice to consolidate trackedTokens and trackedTokenStateByAddress.

packages/website/ts/components/portal/portal.tsx
fragosti: Yea... that or we need a Tokens class. I started moving some of this stuff in utils.ts.

fragosti: Just checked -- only used in Portal and TokenBalances.

packages/website/ts/components/portal/portal.tsx
fragosti: Only used in Portal and TokenBalances

PR #648, opened 2018-05-31
Add shared order support

BMillman19:
## Description

Add support for passing orders as a query param .

## Motivation and Context

Features that are present in legacy portal should be supported in the new portal.

## How Has This Been Tested?

Build and try this link: https://localhost:3572/portal/fill?order={%22signedOrder%22:{%22maker%22:%220x14e2f1f157e7dd4057d02817436d628a37120fd1%22,%22taker%22:%220x0000000000000000000000000000000000000000%22,%22makerFee%22:%220%22,%22takerFee%22:%220%22,%22makerTokenAmount%22:%22100000000000000000%22,%22takerTokenAmount%22:%22100000000000000000%22,%22makerTokenAddress%22:%220xd0a1e359811322d97991e03f863a0c30c2cf029c%22,%22takerTokenAddress%22:%220x6ff6c0ff1d68b964901f986d4c9fa3ac68346570%22,%22expirationUnixTimestampSec%22:%222524636800%22,%22feeRecipient%22:%220x0000000000000000000000000000000000000000%22,%22salt%22:%223729388640488766414530793849304185181521424157202933320100841904548344703401%22,%22ecSignature%22:{%22v%22:27,%22r%22:%220xb1882bc79a536d08eeb27754c361f10c03d8272e343f0a6b944656e4f1fb3455%22,%22s%22:%220x2e9c575ee586c357f68429a2fdaa58a5b3a899a415758332a6979cc0f55ed227%22},%22exchangeContractAddress%22:%220x90fe2af704b34e0224bf2299c838e04d4dcf1364%22},%22metadata%22:{%22makerToken%22:{%22name%22:%22Wrapped%20Ether%22,%22symbol%22:%22WETH%22,%22decimals%22:18},%22takerToken%22:{%22name%22:%220x%20Protocol%20Token%22,%22symbol%22:%22ZRX%22,%22decimals%22:18}}}

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/legacy_portal/legacy_portal.tsx
+++ b/packages/website/ts/components/legacy_portal/legacy_portal.tsx
@@ -38,6 +38,7 @@ import {
 } from 'ts/types';
 import { configs } from 'ts/utils/configs';
 import { constants } from 'ts/utils/constants';
+import { orderParser } from 'ts/utils/order_parser';
 import { Translate } from 'ts/utils/translate';
 import { utils } from 'ts/utils/utils';
 
@@ -86,7 +87,7 @@ export class LegacyPortal extends React.Component<LegacyPortalProps, LegacyPorta
     }
     constructor(props: LegacyPortalProps) {
         super(props);
-        this._sharedOrderIfExists = this._getSharedOrderIfExists();
+        this._sharedOrderIfExists = orderParser.parse(window.location.search);
         this._throttledScreenWidthUpdate = _.throttle(this._updateScreenWidth.bind(this), THROTTLE_TIMEOUT);
 
         const isViewingBalances = _.includes(props.location.pathname, `${WebsitePaths.Portal}/balances`);
--- a/packages/website/ts/components/legacy_portal/legacy_portal.tsx
+++ b/packages/website/ts/components/legacy_portal/legacy_portal.tsx
@@ -363,32 +363,6 @@ export class LegacyPortal extends React.Component<LegacyPortalProps, LegacyPorta
             isWethNoticeDialogOpen: false,
         });
     }
-    private _getSharedOrderIfExists(): Order | undefined {
-        const queryString = window.location.search;
-        if (queryString.length === 0) {
-            return undefined;
-        }
-        const queryParams = queryString.substring(1).split('&');
-        const orderQueryParam = _.find(queryParams, queryParam => {
-            const queryPair = queryParam.split('=');
-            return queryPair[0] === 'order';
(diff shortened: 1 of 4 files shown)


packages/website/ts/components/legacy_portal/legacy_portal.tsx
```diff
@@ -86,7 +87,7 @@ export class LegacyPortal extends React.Component<LegacyPortalProps, LegacyPorta
     }
     constructor(props: LegacyPortalProps) {
         super(props);
-        this._sharedOrderIfExists = this._getSharedOrderIfExists();
+        this._sharedOrderIfExists = orderParser.parse(window.location.search);
```
fragosti: I don't think this will lead to bugs, but it may be better to use `withRouter` ? 
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md

The only way this could lead to a bug is if somehow that query param is pushed using the history API.

packages/website/ts/utils/order_parser.ts
```diff
@@ -0,0 +1,33 @@
+import { logUtils } from '@0xproject/utils';
+import * as _ from 'lodash';
+
+import { portalOrderSchema } from 'ts/schemas/portal_order_schema';
+import { validator } from 'ts/schemas/validator';
+import { Order } from 'ts/types';
+
+export const orderParser = {
+    parse(queryString: string): Order | undefined {
```
fragosti: Have we been adding `undefined` just to be explicit?

BMillman19: this function returns undefined so we should be as explicit as possible to help the type system catch errors

packages/website/ts/components/legacy_portal/legacy_portal.tsx
BMillman19: yea i think its worth revisiting how this works, i think we might just be able to to push this down to the `FillOrder` component but keeping it the same as before to avoid refactoring too much

packages/website/ts/utils/order_parser.ts
fragosti: I guess what I'm saying is that we have strict null type checking off, so I think you don't necessarily need to explicitly add `undefined` but in the event that we turn it back on, this is useful (less work to get everything compiling).

BMillman19: I see yes, we should turn that back on at some point, but in either case, the more explicit the better

PR #652, opened 2018-06-02
Implement subscribe form on landing page

fragosti:
screen shot 2018-06-04 at 5 49 49 pm
screen shot 2018-06-04 at 5 50 44 pm
screen shot 2018-06-04 at 5 50 59 pm

## Description

This implements the subscribe form on the landing page as seen here https://www.figma.com/file/RdbYgOanBjzzjk9fkqr1xEK6/landingpage-mock-v3-jan-18

This change also:
- Adds styled-components and polished as dependencies. 
- Implements non-material UI Input, Button, and Text components using styled-components.
- Refactors the CTAs on on the landing page (wanted to refactor the entire landing page but that was going to be a lot).
- Move some style stuff in its own folder. 
- New client code for hitting the new website-backend endpoint (which still has to be deployed)

TODO: Changelog

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

*   [] Labeled this PR with the 'WIP' label if it is a work in progress.


packages/website/ts/components/ui/text.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export type TranslatedTextProps = TextProps & TranslatedProps;
+
+export const TranslatedText: React.StatelessComponent<TranslatedTextProps> = ({
+    translate,
+    children,
+    deco,
+    ...textProps,
+}) => <Text {...textProps}>{translate.get(children, deco)}</Text>;
```
fragosti: This is not used yet... Maybe in a future refactor of the landing page

packages/website/ts/utils/utils.ts
```diff
@@ -306,7 +306,8 @@ export const utils = {
         return parsedProviderName;
     },
     getBackendBaseUrl(): string {
-        return isDogfood() ? configs.BACKEND_BASE_STAGING_URL : configs.BACKEND_BASE_PROD_URL;
+        return 'http://localhost:3000';
```
fragosti: TODO: Remove

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+                message = 'Sorry, something went wrong. Try again later.';
+                break;
+            case SubscribeFormStatus.Loading:
+                message = 'One second...';
+                break;
+            case SubscribeFormStatus.Success:
+                message = `Thanks! ${this.state.lastSubmittedEmail} is now on the mailing list.`;
+                break;
+            case SubscribeFormStatus.None:
+                break;
```
fabioberger: Let's add a default case that throws for early detection in case this gets refactored in the future.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+        }
+        return (
+            <Container isHidden={!message} marginTop="30px">
+                <Text center={true} fontFamily="Roboto Mono" fontColor={colors.grey}>
+                    {message || 'spacer text'}
+                </Text>
+            </Container>
+        );
+    }
+
```
fabioberger: Remove newline.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+            case SubscribeFormStatus.Success:
+                message = `Thanks! ${this.state.lastSubmittedEmail} is now on the mailing list.`;
+                break;
+            case SubscribeFormStatus.None:
+                break;
+        }
+        return (
+            <Container isHidden={!message} marginTop="30px">
+                <Text center={true} fontFamily="Roboto Mono" fontColor={colors.grey}>
+                    {message || 'spacer text'}
```
fabioberger: Is `spacer text` ever shown to the user?

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+            </Container>
+        );
+    }
+
+    private _handleEmailInputChange(event: React.ChangeEvent<HTMLInputElement>): void {
+        this.setState({ emailText: event.target.value });
+    }
+    private async _handleFormSubmitAsync(event: React.FormEvent<HTMLInputElement>): Promise<void> {
+        event.preventDefault();
+        if (!this.state.emailText) {
```
fabioberger: Let's use `_.isEmpty` or `_.isUndefined` here. Otherwise this conditional feels too loose.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+        });
+        try {
+            const response = await backendClient.subscribeToNewsletterAsync(this.state.emailText);
+            const status = response.status === 200 ? SubscribeFormStatus.Success : SubscribeFormStatus.Error;
+            this._setStatus(status);
+        } catch (error) {
+            logUtils.log(error);
+            this._setStatus(SubscribeFormStatus.Error);
+        } finally {
+            this.setState({ emailText: '' });
```
fabioberger: Do we want to clear if an error occurs? My vote would be no. This way, it's easier for the user to retry or edit the existing email.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+    private _setStatus(status: SubscribeFormStatus, then?: () => void): void {
```
fabioberger: `then` optional param isn't used anywhere.

packages/website/ts/pages/landing/landing.tsx
```diff
@@ -782,15 +774,7 @@ export class Landing extends React.Component<LandingProps, LandingState> {
...
-                                <RaisedButton
-                                    style={{ borderRadius: 6, minWidth: 150 }}
-                                    buttonStyle={lightButtonStyle}
-                                    labelColor={colors.white}
-                                    backgroundColor={colors.heroGrey}
-                                    labelStyle={buttonLabelStyle}
-                                    label={this.props.translate.get(Key.BuildCallToAction, Deco.Cap)}
-                                    onClick={_.noop}
-                                />
+                                <CTA fontSize="15px">{this.props.translate.get(Key.BuildCallToAction, Deco.Cap)}</CTA>
```
fabioberger: Honestly, it took me a good second to understand what a CTA was... Can we be more verbose and just name it `CallToAction`?

packages/website/ts/utils/fetch_utils.ts
```diff
@@ -20,6 +20,18 @@ export const fetchUtils = {
         const result = await response.json();
         return result;
     },
+
```
fabioberger: remove new line.

packages/website/ts/components/ui/input.tsx
```diff
@@ -0,0 +1,43 @@
...
+    font-size: ${props => props.fontSize};
+    width: ${props => props.width};
+    padding: 0.8em 1.2em;
+    border-radius: 3px;
+    font-family: 'Roboto Mono';
+    color: ${props => props.fontColor};
+    border: none;
+    background-color: ${props => props.backgroundColor};
+    &::placeholder {
+        color: ${props => props.placeholder};
```
fragosti: placeholderColor

packages/website/ts/components/ui/input.tsx
```diff
@@ -0,0 +1,43 @@
...
+    placeholder?: string;
+    backgroundColor?: string;
+    onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+const PlainInput: React.StatelessComponent<InputProps> = ({ value, className, placeholder, onChange }) => (
+    <input className={className} value={value} onChange={onChange} placeholder={placeholder} />
+);
+
+export const Input = styled(PlainInput)`
```
BMillman19: same as above

packages/website/ts/utils/fetch_utils.ts
```diff
@@ -20,6 +20,18 @@ export const fetchUtils = {
...
+    async postAsync(baseUrl: string, path: string, body: object): Promise<Response> {
```
BMillman19: we should have some error handling here like `requestAsync` above

packages/website/ts/components/forms/subscribe_form.tsx
fragosti: Oh typescript doesn't complain? eesh. Maybe a linting rule will help but doing this for now.

packages/website/ts/components/forms/subscribe_form.tsx
fragosti: I've noticed our convention of using these utils in the codebase, could you briefly walk me through the advantages? To me it seems like "loose" is a good quality to have in this case, as we are catching `null | undefined | ''` and the rest is caught by typescript. I see how it is more explicit, but don't see other advantages. I personally like relying on javascript falsyness because it is often exactly what I mean, but maybe I have bad habits :P

packages/react-shared/src/utils/colors.ts
```diff
@@ -22,7 +22,7 @@ const baseColors = {
     heroGrey: '#404040',
     projectsGrey: '#343333',
     darkestGrey: '#272727',
-    dharmaDarkGrey: '#252525',
+    darkButtonGrey: '#252525',
```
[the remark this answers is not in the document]
fragosti: What is the difference between "app" colors and "shared" colors. Where else are the other colors used? I feel like we should just have a color palette and enforce that across everything.

I do agree that having the text "button" is a bit specific, so will change the name in the meantime.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -0,0 +1,121 @@
...
+export class SubscribeForm extends React.Component<SubscribeFormProps, SubscribeFormState> {
+    public state = {
+        emailText: '',
+        lastSubmittedEmail: '',
+        status: SubscribeFormStatus.None,
+    };
+    public render(): React.ReactNode {
+        const formFontSize = '15px';
+        return (
+            <Container className="flex flex-column items-center justify-between md-mx2 sm-mx2">
```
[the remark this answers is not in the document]
fragosti: Mostly by accident. But I don't think it's a bad thing ... I do think this is a nice level of abstraction to work on (always be using components with a clear API instead of raw elements). 
```
<Container {...layout props...}>
    <Text {...text props...}>
        hey 
    <Text>
</Container>
```
Is easier to read / has clearer separation of concerns than
```
<div style={... styles for layout and for the text style...}>
    hey
</div>
```

packages/website/ts/components/ui/button.tsx
```diff
@@ -0,0 +1,79 @@
...
+    onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+}
+
+const PlainButton: React.StatelessComponent<ButtonProps> = ({ children, onClick, type, className }) => (
+    <button type={type} className={className} onClick={onClick}>
+        {children}
+    </button>
+);
+
+export const Button = styled(PlainButton)`
```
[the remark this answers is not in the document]
fragosti: I tried and I'm not sure it's worth the overhead here. The generics are working and visual studio code does give you a valid type. To explicitly type this you need to import our "theme interface" (currently empty) every time which doesn't seem worth it?

packages/website/ts/components/forms/subscribe_form.tsx
fabioberger: Adding that rule now.

packages/website/ts/components/forms/subscribe_form.tsx
fabioberger: Sure. First off, we actually avoid using `null` values in the codebase. Instead we always use `undefined`. This came out of a longer discussion with @LogvinovLeon where we reasoned that JS returns `undefined` in cases where a property or value does not exist. JS never returns `null`. 

The main question we had was: "Is there a good reason to distinguish between `undefined` and `null` values? Do they provide some semantic meaning to the developer?". The official answer is that `undefined` refers to a value that has been declared but not assigned. `null` must be intentionally assigned and signifies the value has been intentionally assigned to an empty or non-existent value. Although is difference could make sense in some niche cases, most of the time, we simply care if the value is non-existent or existent. 

To simplify the codebase and to enforce stricter type checking, we always use `undefined` and check with `_.isUndefined` (and `_.isEmpty` for string values that can be empty). Being stricter in the conditional checking means that we won't accidentally have a "falsy" value such as `0`, `false`, `NaN` or `''` trigger the conditional, then what we really wanted to check for was whether the value was defined or not. We've found that we find bugs much sooner by having this stricter policy.

Does this make sense? Like always, we are open to changing any code convention if an alternative proposition makes more rational sense. The only downside of our approach that we've encountered so far is having to wrap some public interfaces of dependencies that currently return `null` rather then `undefined`. We also make an exception for JSX returning methods since `null` is special-cased in React to avoid rendering elements to the DOM.

fragosti: That makes sense, although this specific example may not be the best show-case for it. If I use `_.isEmpty` instead of just checking for falsyness, I actually open myself up to bugs as I have to assume that `emailText` is always a non-null non-undefined string. Typescript can guarantee me that it is a string, but with the current compiler settings (in `website`) it cannot guarantee me that it is not `null` or `undefined`. You mention that we don't use `null`, so that narrows it down to `undefined`. So I guess our convention in this case would be to use both `_.isUndefined` and `_.isEmpty`. However, as you mentioned there are some cases in which `null` does sneak in (we have to allow it for JSX and 3rd party libraries tend to use it), so we are still making an assumption and hoping for the best.

So I guess I disagree :P. But this is code is a closed system and I think following convention (and shipping) trumps some argument about how it could theoretically be incorrect. Maybe we can revisit this in the future, going to use `isUndefined` and `isEmpty` for now.

PR #657, opened 2018-06-04
Speedup CI builds

LogvinovLeon:
## Description

Enables the `skipLibCheck` flag in TS.
https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#new---skipLibCheck
It speeds up the build from 247s to 207s for me.

Stops running prettier on generated contract wrappers as it takes time.

Also removes Lerna bootstrap step on CI. We use yarn workspaces now and it's not needed anymore. (1m 20s gain).

## Motivation and Context

Cause time is money.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,12 @@ jobs:
           key: yarn-packages-{{ .Branch }}-{{ checksum "yarn.lock" }}
           paths:
             - node_modules/
-      - run: yarn build
+      - run: >
+            if [ -z "$(git diff --name-only v2-prototype packages/website)" ]; then
+                yarn build --exclude website
+            else
+                yarn build
+            fi
       - save_cache:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
           paths:


albrow:
Can we update the PR description to mention the prettier changes? (Or if they were included here unintentionally, put them in a separate PR).

packages/abi-gen/src/utils.ts
```diff
@@ -56,7 +56,7 @@ export const utils = {
                     const componentType = `${component.name}: ${componentValueType}`;
                     return componentType;
                 });
-                const tsType = `{${componentsType}}`;
+                const tsType = `{${componentsType.join(';')}}`;
```
fabioberger: If this is a bug fix, mind adding a CHANGELOG entry?

LogvinovLeon: It's not a bug fix. Before it was using a comma and that's also OK. I changed it to ';' because the linter was complaining. Users should not be affected so I don't think it belongs to CHANGELOG

PR #658, opened 2018-06-04
Don't automatically rebuild when running yarn test

albrow:
## Description

Prior to this change, for each package, we were usually running `clean` and `build` automatically before `test`. This makes sense when `clean` and `build` are fast, but as our monorepo has grown, this slows down the process for running tests significantly. Another recent change that makes the old approach somewhat outdated is https://github.com/0xProject/0x-monorepo/pull/588. We now have the ability to build or watch a single package and all its dependencies, which has changed the way most of us work.

After speaking with a few other engineers, we believe that in most common workflows, we have already run `build` (sometimes implicitly through running `watch` in the background) before we run `test`. It is not usually necessary to run it again. If you really do want the old behavior of running `clean` and `build` before tests, I added a new `rebuild-and-test` command to each relevant package.

We also expect that this change will make CI faster. In CI, we build the entire monorepo once and then run the tests, so building again for each test is unnecessary.

## Types of changes

## Checklist:


packages/0x.js/package.json
```diff
@@ -21,7 +21,8 @@
         "generate_contract_wrappers": "abi-gen --abis 'src/compact_artifacts/@(Exchange|Token|TokenTransferProxy|EtherToken|TokenRegistry|DummyToken).json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers/generated --backend ethers && prettier --write 'src/contract_wrappers/generated/**.ts'",
         "lint": "tslint --project .",
         "test:circleci": "run-s test:coverage",
-        "test": "run-s clean test:commonjs",
+        "test": "run-s test:commonjs",
```
albrow: This __package.json__ file is a little different than the rest. Is it okay to skip the build step in `test:commonjs`?

LogvinovLeon: Yes

packages/0x.js/package.json
```diff
@@ -17,17 +17,15 @@
...
         "watch": "tsc -w",
         "prebuild": "run-s clean generate_contract_wrappers",
-        "build": "run-p build:umd:prod build:commonjs; exit 0;",
+        "build": "tsc && copyfiles -u 2 './src/compact_artifacts/**/*.json' ./lib/src/compact_artifacts && copyfiles -u 3 './lib/src/monorepo_scripts/**/*' ./scripts",
         "generate_contract_wrappers": "abi-gen --abis 'src/compact_artifacts/@(Exchange|Token|TokenTransferProxy|EtherToken|TokenRegistry|DummyToken).json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers/generated --backend ethers && prettier --write 'src/contract_wrappers/generated/**.ts'",
         "lint": "tslint --project .",
         "test:circleci": "run-s test:coverage",
-        "test": "run-s clean test:commonjs",
+        "test": "yarn run_mocha",
+        "rebuild-and-test": "run-s build test",
```
fabioberger: @albrow moving forward let's try and keep `yarn script` names consistent by always using underscores over dashes. I've gone ahead and fixed these in a PR I will open soon.

packages/0x.js/package.json
```diff
@@ -17,17 +17,15 @@
...
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
         "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
         "clean": "shx rm -rf _bundles lib test_temp scripts src/contract_wrappers/generated",
-        "build:umd:prod": "NODE_ENV=production webpack",
```
fabioberger: @LogvinovLeon are we no longer generating UMD bundles?

PR #659, opened 2018-06-04
Split transfer implementation and AssetProxyMixin

dekz:
## Description

Split the Transfer implementation and the AssetProxy-ness so the Transfer impl can be mixed in without the AssetProxy.

## Motivation and Context

This is useful for the Forwarding contract as we can get the validation and the and safeTransfer without the AssetProxy public methods.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/AssetProxy/MixinERC20Transfer.sol
+++ b/packages/contracts/src/contracts/current/protocol/AssetProxy/MixinERC20Transfer.sol
@@ -1,0 +1,71 @@
+/*
+
+  Copyright 2018 ZeroEx Intl.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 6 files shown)


packages/contracts/src/contracts/current/protocol/AssetProxy/MixinERC20Transfer.sol
```diff
@@ -0,0 +1,71 @@
...
+pragma experimental ABIEncoderV2;
+
+import "../../utils/LibBytes/LibBytes.sol";
+import "../../tokens/ERC20Token/IERC20Token.sol";
+
+contract MixinERC20Transfer is
+    LibBytes
+{
+    // Id of this proxy.
+    uint8 constant PROXY_ID = 1;
```
abandeali1: There have been some changes in the implementation of `transferFromInternal` for both proxies, so they no longer require `PROXY_ID`. We can move `PROXY_ID` back to the actual proxies after those changes have been merged into this branch.

dekz: awww yissssss

PR #661, opened 2018-06-05
Fix broken link to `contract_templates`

feuGeneA:
Note there is still a broken link here, a previously existing one which I would have fixed but for which I couldn't find the appropriate link.  It's the "wrapper files" one in the second paragraph.


fabioberger:
Thanks for the fix @feuGeneA! I'll get the last one!

PR #662, opened 2018-06-05
Tried enabling no-unused-variable...

fragosti:
## Description

So I tried enabling the "no-unused-variable" tslint rule in the mono-repo and this is what happened... and I'm not even done yet. I figured it would be good to post this PR as a discussion before progressing. I kind of want this rule enabled in `website` but maybe it's too much of a risk to enable it everywhere all at once?

Also noticed some weirdness with tslint in general. Is "yarn lint" from the root directory supposed to be passing?

## Types of changes

## Checklist:

### DIFF
--- a/packages/0x.js/test/0x.js_test.ts
+++ b/packages/0x.js/test/0x.js_test.ts
@@ -1,14 +1,12 @@
 import { ContractWrappers } from '@0xproject/contract-wrappers';
-import { BlockchainLifecycle, devConstants, web3Factory } from '@0xproject/dev-utils';
+import { BlockchainLifecycle } from '@0xproject/dev-utils';
 import { BigNumber } from '@0xproject/utils';
 import * as chai from 'chai';
 import * as _ from 'lodash';
 import 'make-promises-safe';
 import 'mocha';
-import * as path from 'path';
-import * as Sinon from 'sinon';
 
-import { ApprovalContractEventArgs, LogWithDecodedArgs, Order, TokenEvents, ZeroEx } from '../src';
+import { ApprovalContractEventArgs, LogWithDecodedArgs, TokenEvents, ZeroEx } from '../src';
 
 import { chaiSetup } from './utils/chai_setup';
 import { constants } from './utils/constants';
@@ -137,6 +135,7 @@ describe('ZeroEx library', () => {
             const proxyAddress = zeroEx.proxy.getContractAddress();
             const txHash = await zeroEx.token.setUnlimitedProxyAllowanceAsync(zrxTokenAddress, coinbase);
             const txReceiptWithDecodedLogs = await zeroEx.awaitTransactionMinedAsync(txHash);
+            // tslint:disable-next-line:no-unnecessary-type-assertion
             const log = txReceiptWithDecodedLogs.logs[0] as LogWithDecodedArgs<ApprovalContractEventArgs>;
             expect(log.event).to.be.equal(TokenEvents.Approval);
             expect(log.args._owner).to.be.equal(coinbase);
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -44,6 +44,7 @@ export class BaseContract {
         const constructorAbiIfExists = _.find(
             abi,
             (abiDefinition: AbiDefinition) => abiDefinition.type === AbiType.Constructor,
+            // tslint:disable-next-line:no-unnecessary-type-assertion
         ) as ConstructorAbi | undefined;
         if (!_.isUndefined(constructorAbiIfExists)) {
             return constructorAbiIfExists;
--- a/packages/contract-wrappers/src/contract_wrappers/token_wrapper.ts
+++ b/packages/contract-wrappers/src/contract_wrappers/token_wrapper.ts
@@ -1,6 +1,6 @@
 import { schemas } from '@0xproject/json-schemas';
 import { LogWithDecodedArgs } from '@0xproject/types';
-import { AbiDecoder, BigNumber } from '@0xproject/utils';
+import { BigNumber } from '@0xproject/utils';
 import { Web3Wrapper } from '@0xproject/web3-wrapper';
 import * as _ from 'lodash';
 
--- a/packages/fill-scenarios/tslint.json
+++ b/packages/fill-scenarios/tslint.json
@@ -1,3 +1,6 @@
 {
-    "extends": ["@0xproject/tslint-config"]
+    "extends": ["@0xproject/tslint-config"],
+    "rules": {
+        "no-unused-variable": false
+    }
 }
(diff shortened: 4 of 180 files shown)


packages/base-contract/src/index.ts
```diff
@@ -44,6 +44,7 @@ export class BaseContract {
         const constructorAbiIfExists = _.find(
             abi,
             (abiDefinition: AbiDefinition) => abiDefinition.type === AbiType.Constructor,
+            // tslint:disable-next-line:no-unnecessary-type-assertion
```
fabioberger: Why were these removable now?

packages/fill-scenarios/tslint.json
```diff
@@ -1,3 +1,6 @@
 {
-    "extends": ["@0xproject/tslint-config"]
+    "extends": ["@0xproject/tslint-config"],
+    "rules": {
+        "no-unused-variable": false
```
fabioberger: Why not add it as a base rule?

packages/monorepo-scripts/src/postpublish_utils.ts
```diff
@@ -93,8 +93,10 @@ export const postpublishUtils = {
         const notes = this.getReleaseNotes(packageName, version);
         const releaseName = this.getReleaseName(packageName, version);
         const tag = this.getTag(packageName, version);
+        // tslint:disable-next-line:no-unused-variable
```
fabioberger: Why are so many of these necessary?

fabioberger:
Wow. This rule is killer. We obviously desperately need it, there have been so many unused variables and imports... 

What "weirdness" did you notice? Can you be more specific?

Also, yes, the `yarn lint` run from the root directory is passing on `v2-prototype` and is part of CircleCI.

packages/monorepo-scripts/src/postpublish_utils.ts
fragosti: I can just wrap them in a `disable` comment (or remove the var and disable the no-dangling promises rule)

packages/base-contract/src/index.ts
fragosti: Not sure what you mean.

packages/fill-scenarios/tslint.json
fragosti: It is a base rule but I disabled it for this package because I noticed there were generated files. Maybe I should just ignore that directory?

For example `dummy_token.ts` 
/**
 * This file is auto-generated using abi-gen. Don't edit directly.
 * Templates can be found at https://github.com/0xProject/0x-monorepo/tree/development/packages/contract_templates.
 */

fragosti:
@fabioberger so we are ok with moving forward with this? The weirdness is that I don't understand why I started failing the "no-unnecessary-type-assertion" rule as well and had to sprinkle all the tslint disable calls.

packages/monorepo-scripts/src/postpublish_utils.ts
fragosti: So to clarify: I am being extremely conservative with removing. I'm only deleting the expression if it's a simple assignment, or if it's a read-only assignment. If the expression could have a side effect, or if it is a private method on a class (that could be used in tests or elsewhere) I just add the tslint comment. 

Not sure if your comment here refers to this file in particular (which I can clean up) or more generally.

Here is an example of a private method that we use pretty freely for some reason.

```js
src/contract_wrappers.ts:        (this.exchange as any)._setNetworkId(networkId);
src/contract_wrappers.ts:        (this.tokenRegistry as any)._setNetworkId(networkId);
src/contract_wrappers.ts:        (this.token as any)._setNetworkId(networkId);
src/contract_wrappers.ts:        (this.proxy as any)._setNetworkId(networkId);
src/contract_wrappers.ts:        (this.etherToken as any)._setNetworkId(networkId);
src/contract_wrappers/contract_wrapper.ts:    private _setNetworkId(networkId: number): void {
```

packages/base-contract/src/index.ts
fabioberger: Why were you able to remove these unnecessary type assertions, whereas before we disabled the tslint rule?

packages/fill-scenarios/tslint.json
fabioberger: Yeah, we should ignore generated files. I thought I saw @LogvinovLeon do this in a recent PR.

packages/0x.js/test/0x.js_test.ts
```diff
@@ -137,6 +135,7 @@ describe('ZeroEx library', () => {
             const proxyAddress = zeroEx.proxy.getContractAddress();
             const txHash = await zeroEx.token.setUnlimitedProxyAllowanceAsync(zrxTokenAddress, coinbase);
             const txReceiptWithDecodedLogs = await zeroEx.awaitTransactionMinedAsync(txHash);
+            // tslint:disable-next-line:no-unnecessary-type-assertion
```
fabioberger: I think this might be necessary... It looks like a TSLint bug where it isn't taking into account the typeVariable when doing type equality. It therefore thinks we are casting to it's existing type, but we are actually casting to the same type, with a different typeVariable. In this specific example: `LogWithDecodedArgs<DecodedLogArgs>` is being cast to `LogWithDecodedArgs<ApprovalContractEventArgs>`.

packages/base-contract/src/index.ts
fabioberger: nvm!

packages/fill-scenarios/tslint.json
fabioberger: I've gone ahead and excluded generated contract wrapper files from the linter.

packages/contract-wrappers/tslint.json
```diff
@@ -1,3 +1,6 @@
 {
-    "extends": ["@0xproject/tslint-config"]
+    "extends": ["@0xproject/tslint-config"],
+    "linterOptions": {
+        "exclude": ["src/contract_wrappers/generated/*"]
```
fabioberger: Did this work for you? It did not for me. I was only able to get it working by passing it in as a commandline option. I think we can remove this.

fragosti: It did seem to work

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -510,6 +496,8 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
             <IconButton iconName={buttonIconName} labelText={buttonLabel} onClick={onClick} color={colors.mediumBlue} />
         );
     }
+<<<<<<< HEAD
+=======
```
fragosti: wtf git

packages/contract-wrappers/src/contract_wrappers/token_wrapper.ts
```diff
@@ -414,6 +414,7 @@ export class TokenWrapper extends ContractWrapper {
         );
         return logs;
     }
+    // tslint:disable:no-unused-variable
```
fabioberger: @fragosti this will disable this rule for the rest of the file... I don't think that's what we want right?

fragosti: 🤦‍♂️

PR #665, opened 2018-06-05
Improve Build/Watch Commands

fabioberger:
## Description
The PR addresses the following issues:

- Currently the monorepo root script `watch` is identically named to the individual sub-package `watch` command. It is therefore easy to accidentally run the `watch` command from a sub-package and expect it to also watch dependencies. In order to avoid this, this PR renames `watch` to `watch_without_deps` in sub-packages. Running `watch` from a sub-package now throws a "Command not found" error, alerting the dev that they are running the command from the wrong directory.

- We currently use the special `prebuild` script name to run steps that must run before `tsc`. These steps should also run before a call to `watch`, but currently don't. We therefore rename them to the non-special name `pre_build` and run it before both the `watch` and `build` commands.

- Currently many `pre_build` steps call the `clean` command. This is many times duplicitous or unnecessary. We therefore remove it from `pre_build`. Dev's can still use the root `yarn rebuild` or `yarn clean` commands to achieve the same effect.

**Note:** Ideally running `watch` from a sub-package, would watch both it and it's dependencies. This is however not possible since the `wsrun` command requires it be called from the root package.json. I've tried using the `yarn` `--cwd` flag to point to the root package.json but it does not seem to work for `yarn run` commands.

## Motivation and Context

Reduce ambiguity and unnecessary overhead.

## How Has This Been Tested?

I ran each `build` and `watch` commands successfully.

## Types of changes

## Checklist:


LogvinovLeon:
LGTM after tests pass

BMillman19:
I think `watch_without_deps` is a bit confusing especially if you are coming from another singular-package code base. Perhaps we can change the top level `watch` command to `watch_with_deps` and leave everything else as `watch` ?

fabioberger:
@BMillman19 we could do that but I'm worried devs will then gravitate towards using `watch` and be annoyed when modified dep changes don't get picked up. Even internally we are already used to calling `watch`... which is part of the problem. The fact of the matter is that no one should be calling `watch_without_deps` directly, everyone should be using the root `watch` script.

PR #672, opened 2018-06-06
Account management polish

BMillman19:
## Description

- Makes a number of changes to the "manage your wallet" section of portal to tidy it up visiually
- Fix an error when navigating directly to `/portal/weth` and the user's ether balance has not loaded yet

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/fill_order.tsx
+++ b/packages/website/ts/components/fill_order.tsx
@@ -39,6 +39,8 @@ interface FillOrderProps {
     initialOrder: Order;
     dispatcher: Dispatcher;
     lastForceTokenStateRefetch: number;
+    isFullWidth?: boolean;
+    shouldRenderHeader?: boolean;
 }
 
 interface FillOrderState {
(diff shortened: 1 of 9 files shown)


packages/website/ts/components/fill_order.tsx
```diff
@@ -61,6 +63,10 @@ interface FillOrderState {
 }
 
 export class FillOrder extends React.Component<FillOrderProps, FillOrderState> {
+    public static defaultProps: Partial<FillOrderProps> = {
+        isFullWidth: false,
+        shouldRenderHeader: true,
```
fragosti: nit: can this be `shouldHideHeader` just so we can have the default value of optional booleans be false?

packages/website/ts/components/generate_order/generate_order_form.tsx
```diff
@@ -56,6 +58,10 @@ interface GenerateOrderFormState {
 }
 
 export class GenerateOrderForm extends React.Component<GenerateOrderFormProps, GenerateOrderFormState> {
+    public static defaultProps: Partial<GenerateOrderFormProps> = {
+        isFullWidth: false,
+        shouldRenderHeader: true,
```
fragosti: same nit

BMillman19:
@fragosti these props will be temporary until we can remove the legacy portal. So let's just keep it as is.

PR #674, opened 2018-06-06
Run contracts tests against Geth in CI

albrow:
## Description

This PR updates __.circleci/config.yaml__ to run tests against Geth. This also required redistributing some of our existing tests in that configuration file.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,28 @@ jobs:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
           paths:
             - ~/repo
-  test-1:
+  test-contracts-ganache:
+    docker:
+      - image: circleci/node:9
+    working_directory: ~/repo
+    steps:
+      - restore_cache:
+          keys:
+            - repo-{{ .Environment.CIRCLE_SHA1 }}
+      - run: yarn wsrun test:circleci contracts
+  test-contracts-geth:
+    docker:
+      - image: circleci/node:9
+      - image: albrow/0x-devnet
+    working_directory: ~/repo
+    steps:
+      - restore_cache:
+          keys:
+            - repo-{{ .Environment.CIRCLE_SHA1 }}
+      # HACK(albrow): we need to sleep 15 seconds to ensure the devnet is
+      # initialized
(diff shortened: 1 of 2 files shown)


packages/devnet/run.sh
```diff
@@ -35,7 +35,9 @@ sleep 10
 # to break when there are fewer than 3 blocks in the chain. (We have no idea
 # why, but it was consistently reproducible).
 /geth --datadir node0/ attach --exec 'eth.sendTransaction({"from": "0x5409ED021D9299bf6814279A6A1411A7e866A631", "to": "0x84bd1cfa409cb0bb9b23b8b1a33515b4ac00a0af", "value": "0x1"})'
+sleep 2
```
albrow: I had to add this extra sleep in order for the tests to work on CI. Without it, all 3 transactions ended up in the same block, making it so there is only 1 block in the chain (besides the genesis block). Who knows why, but our tests fail unless the blockchain has at least 3 blocks.

albrow:
Removed WIP tag. This PR is ready for review. However, we do need to merge https://github.com/0xProject/0x-monorepo/pull/679 first.

.circleci/config.yml
```diff
@@ -35,48 +35,52 @@ jobs:
...
-      - save_cache:
-          key: coverage-sol-compiler-{{ .Environment.CIRCLE_SHA1 }}
-          paths:
-            - ~/repo/packages/sol-compiler/coverage/lcov.info
-  test-contracts:
+      - run: yarn wsrun test:circleci contracts
+  test-contracts-geth:
     docker:
       - image: circleci/node:9
+      - image: albrow/0x-devnet
```
fabioberger: Can you add an Asana task to move this image to the 0xproject docker account?

fabioberger: So does this image contain both the Geth node and the node env. where the tests will run?

albrow: So it took me a while to figure this out because it's not the most well-documented feature. The first image listed here is the "primary image" which will run all of our tests. Any additional images listed will be spun up independently by CircleCI (but in the same network).

PR #676, opened 2018-06-06
Use source-map-support package to include correct line numbers in mocha

albrow:
## Description

This PR uses the [source-map-support](https://www.npmjs.com/package/source-map-support) package in order to get source-mapped stack traces when tests fail in mocha.

## Motivation and Context

Before:

```
 |   1) LibBytes
 |        popByte
 |          should pop the last byte from the input and return it:
 |      Error: This is a typescript error!
 |       at Object.<anonymous> packages/contracts/(lib/test/libraries/lib_bytes.js:144:31)
 |       at step packages/contracts/(lib/test/libraries/lib_bytes.js:32:23)
 |       at Object.next packages/contracts/(lib/test/libraries/lib_bytes.js:13:53)
 |       at fulfilled packages/contracts/(lib/test/libraries/lib_bytes.js:4:58)
 |       at <anonymous>
```

After:

```
 |   1) LibBytes
 |        popByte
 |          should pop the last byte from the input and return it:
 |      Error: This is a typescript error!
 |       at Object.<anonymous> packages/contracts/(test/libraries/lib_bytes.ts:79:19)
 |       at step packages/contracts/(lib/test/libraries/lib_bytes.js:32:23)
 |       at Object.next packages/contracts/(lib/test/libraries/lib_bytes.js:13:53)
 |       at fulfilled packages/contracts/(lib/test/libraries/lib_bytes.js:4:58)
 |       at <anonymous>
 |
```

## Types of changes

## Checklist:


albrow:
Just noticed that it looks like this only fixes the first line of the stack trace. There is an issue here for it: https://github.com/evanw/node-source-map-support/issues/183. I say we still move forward since this is pretty easy to add and better than nothing. We can track progress on that issue and maybe even take a stab at improving it ourselves in the future.

albrow:
Okay so I did a little more investigating and it's not *always* just the first line that gets correctly source-mapped. In the example above, the reason only the first line works are that the other positions in the stack trace are generated JavaScript code to emulate async/await.

Outside of async/await emulation and a few other niche cases, we should get correct source-mapping for the entire stack trace.

albrow:
@fabioberger there is a way to have sourcemap support outside of Mocha but it requires adding:

```
import * as sourceMapSupport from 'source-map-support';
sourceMapSupport.install();
```

to every file. Seems like too much of a burden to keep up.

Alternatively you can use the -r flag when running node:

```
node -r source-map-support/register compiled.js
```

PR #684, opened 2018-06-11
Move ExchangeTransferSimulator & OrderValidationUtils to Order-Utils

fabioberger:
## Description

Combinatorial testing can be done more efficiently for single and batch fills with the help of `ExchangeTransferSimulator` and `OrderValidationUtils`. To this end, this PR copies both over to `order-utils`, refactors them and their tests for V2 and removes their dependencies on `contract-wrappers`.

It **does not** remove the old modules from `contract-wrappers` since the package is not yet updated for V2. These modules **should be** removed during the refactor of `contract-wrappers` to V2. @LogvinovLeon 

Additional changes:
- Export ArtifactWriter from `migrations` package.
- Refactor ExchangeTransferSimulator public interface to accept an `AbstractBalanceAndProxyAllowanceLazyStore` so that this module could be re-used in different contexts.
- Export `parseECSignature` method from `order-utils`.

## Motivation and Context

We want to be able to generated expected balances/allowances after both single and batch fills in our combinatorial tests. These modules make this trivial to implement and let's us re-use code!

## How Has This Been Tested?

Tests pass.

## Types of changes

## Checklist:


packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -54,7 +55,7 @@ interface ExchangeContractErrCodesToMsgs {
  */
 export class ExchangeWrapper extends ContractWrapper {
     private _exchangeContractIfExists?: ExchangeContract;
-    private _orderValidationUtils: OrderValidationUtils;
+    private _orderValidationUtilsIfExists?: OrderValidationUtils;
```
LogvinovLeon: I don't understand the need for that change. Please elaborate. Makes the code bigger & more complex

packages/order-utils/test/global_hooks_test.ts
```diff
@@ -0,0 +1,45 @@
...
+        gas: devConstants.GAS_LIMIT,
+        from: devConstants.TESTRPC_FIRST_ADDRESS,
+    };
+
+    const networkId = constants.TESTRPC_NETWORK_ID;
+    const artifactsDir = `lib/src/artifacts`;
+    const artifactsWriter = new ArtifactWriter(artifactsDir, networkId);
+
+    const erc20proxy = await ERC20ProxyContract.deployFrom0xArtifactAsync(artifacts.ERC20Proxy, provider, txDefaults);
+    artifactsWriter.saveArtifact(erc20proxy);
```
LogvinovLeon: Why do we need to write those artifacts? Aren't those for tests?

packages/migrations/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export { runV1MigrationsAsync } from './v1/migration';
 export { runV2MigrationsAsync } from './v2/migration';
+export { ArtifactWriter } from './artifact_writer';
```
LogvinovLeon: Could you explain - why do we need to use ArtifactWriter outside of migrations?

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -0,0 +1,223 @@
...
+        signedOrder: SignedOrder,
+        fillTakerTokenAmount: BigNumber,
+        takerAddress: string,
+        zrxTokenAddress: string,
+    ): Promise<BigNumber> {
+        if (fillTakerTokenAmount.eq(0)) {
+            throw new Error(ExchangeContractErrs.OrderFillAmountZero);
+        }
+        const orderHash = orderHashUtils.getOrderHashHex(signedOrder);
+        // TODO: Verify all signature types! To do this, we need access to a Provider...
```
LogvinovLeon: Is this out of scope for this PR?

packages/order-utils/test/utils/simple_erc20_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ERC20TokenContract } from '../../src/generated_contract_wrappers/e_r_c20_token';
+
+export class SimpleERC20BalanceAndProxyAllowanceFetcher implements AbstractBalanceAndProxyAllowanceFetcher {
+    private _erc20TokenContract: ERC20TokenContract;
+    private _erc20ProxyAddress: string;
+    constructor(erc20TokenWrapper: ERC20TokenContract, erc20ProxyAddress: string) {
+        this._erc20TokenContract = erc20TokenWrapper;
+        this._erc20ProxyAddress = erc20ProxyAddress;
+    }
+    public async getBalanceAsync(assetData: string, userAddress: string): Promise<BigNumber> {
```
LogvinovLeon: Will that fail linter as an unused param? If it will not - let's add a rule and disable it here. Will be more clear that way

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
fabioberger: We need to get the ExchangeContract (an async call) so we cannot do this in the constructor unfortunately.

LogvinovLeon: Why were we able to do that before?

packages/order-utils/test/global_hooks_test.ts
fabioberger: That's fair. I can imagine people writing their own migrations wanting access to this method... Any hesitations about exposing it?

fabioberger: Good point was able to remove this by deploying within the test itself.

packages/migrations/src/index.ts
fabioberger: I can imagine people writing their own migrations wanting access to this method... Any hesitations about exposing it?

packages/order-utils/src/order_validation_utils.ts
fabioberger: Ah good catch, added it.

packages/migrations/src/index.ts
LogvinovLeon: No. Let's just make sure that it's ready (documented and has nice API).

packages/order-utils/test/utils/simple_erc20_balance_and_proxy_allowance_fetcher.ts
fabioberger: Just added the rule and fixed all the issues that arose.

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
fabioberger: We didn't instantiate the ExchangeContract (generated wrapper), but instead passed in `contract-wrappers` ExchangeWrapper.

packages/migrations/src/index.ts
fabioberger: Good point. Let's punt on this for now.

PR #688, opened 2018-06-12
Speedup awaitTransactionMinedAsync and reduce polling interval in contracts tests

albrow:
## Description

This PR fixes what is arguably a bug in `awaitTransactionMinedAsync`. Previously, we were waiting for the first interval (default 1s) to pass before checking if the transaction was mined. The new implementation immediately checks if the transaction was mined, and then if it was not mined, begins polling on every interval.

Along the same vein of increasing performance while waiting for transactions, I also changed `constants.AWAIT_TRANSACTION_MINED` in the contracts package from 100ms to 0ms. (In reality v8 limits the minimum interval to 10ms, but in any case the CPU doesn't overwork in practice because we are still blocking on network IO).

These two changes drastically increase the speed of the tests, especially in the contracts package. It's a difference of ~240s for both providers to ~98s for Ganache and ~60s for Geth!

## Types of changes

## Checklist:


fabioberger:
WHOA! Tests are soooo much faster now 👍

PR #697, opened 2018-06-13
[Discussion] userData in Order structure

dekz:
## Description
Add an opaque field named `bytes userData` to the Order Format.

## Motivation and Context

0x architecture follows a pattern of off-chain orders on-chain settlement. Makers sign an order off-chain and after setting appropriate allowances is able to continuously trade without any additional blockchain interaction. 

The emergence of filter contracts and forwarding contracts enable 0x to be a building block in a particular dApp. Currently there is no way for more additional "data" to be signed by the maker as proof of intent. This leads to misuse of `salt` or `assetData` fields. These fields are 0x specific and have 0x specific meaning and can change in 0x exchange upgrades. 

This PR introduces a `bytes userData` field that is completely opaque to the 0x contracts. Though it is used and verified in order hash and signature calculation. This allows additional data encoding for use in filtering and forwarding contracts. For example, in a Dutch Auction contract parameters like the `auctionStartTime`. As the data remains opaque to 0x Exchange contracts unlike the salt or asset data this will continue to work in future protocol upgrades.

### Concerns
This adds an additional field to the Order format, which even when null adds a slight over head to all fill operations.

A basic gas cost estimate
```
fillOrder (without userData)
199682
fillOrder (userData: 0x0)
200832
fillOrder (userData: 0x1B61a3ed31b43c8780e905a260a35faefcc527be7516aa11c0256729b5b351bc3340349190569279751135161d22529dc25add4f6069af05be04cacbda2ace225403)
205748
```

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
@@ -38,6 +38,7 @@ contract LibOrder {
         "uint256 salt,",
         "bytes makerAssetData,",
         "bytes takerAssetData,",
+        "bytes userData,",
         ")"
     ));
 
@@ -66,6 +67,7 @@ contract LibOrder {
         uint256 salt;
         bytes makerAssetData;
         bytes takerAssetData;
+        bytes userData;
     }
 
     struct OrderInfo {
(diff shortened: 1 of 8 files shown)


Recmo:
Would it be sufficient to add a `bytes32 userDataHash`? The actual user-data can then be passed to the forwarding contract in a separate parameter. It would still be tied to the signed order. The userdata can then be any kind of type, instead of having to be an opaque byte array.

Pro: less overhead for the exchange contract, more flexibility in included data.

Con: you now need to store/query/submit two pieces of information instead of one.

Neither this or the original proposal will look good in a signing UI. EIP712 does not help a lot here.

dekz:
limiting to a bytes32 data is probably ok (it could be a hash or a total 32 byte encoded field). It likely doesn't reduce the callData at the top of the funnel as it would be provided to the Forwarding contract at call time.

dekz:
Closing as we will gather usage on salt encoding for changes in future protocol versions.

PR #700, opened 2018-06-14
[discussion] Handle tokens that do not return bool

Recmo:
## Description

See title.

Still needs test coverage.

## Types of changes

## Checklist:


Recmo:
This should be similar. If not, we can always replicate whatever 0.4.22 is doing in assembler. But I'd rather do something more explicit, because 0.4.22 is broken and the reason it works is because of pure luck: It does a call setting the return data to overwrite the call data. Then it reads the first 32 bytes of calldata. If nothing was returned this still contains the call data. This starts with the `bytes4` function selector. As it is non-zero it evaluates to true.

Eventhough this should never fail, I'd rather not depend on hacks like that. The way I'm solving it now should be similar in gas cost. Any difference should be less than 100 Gas, and I estimate it can be optimized away entirely.

PR #702, opened 2018-06-14
[sol-cov] Only collect coverage for provided sources

perissology:
## Motivation and Context

When solidity generates source maps during contract compilation, the
contracts are referred to by an id, which corresponds to an array index.

We may not want to cover all sources that were included in a compilation,
but because we use array indexes (vs. the id that is provided by solidity
compiler) to map the contract to the sourceMap, the provided sourceCodes
array must include the code at the correct index. This can result in
empty slots in the sourceCodes array.

This commit allows the coverage to only be collected for the contracts
with provided sourceCode.

## How Has This Been Tested?

Tested with our repos. I was not able to run `yarn test` as it failed before I made any changes

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-cov/src/source_maps.ts
+++ b/packages/sol-cov/src/source_maps.ts
@@ -12,6 +12,9 @@ export interface SourceLocation {
 }
 
 export function getLocationByOffset(str: string): LocationByOffset {
+    if (_.isUndefined(str)) {
+        return {};
+    }
     const locationByOffset: LocationByOffset = { 0: { line: 1, column: 0 } };
     let currentOffset = 0;
     for (const char of str.split('')) {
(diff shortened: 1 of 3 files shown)


packages/sol-cov/src/source_maps.ts
```diff
@@ -12,6 +12,9 @@ export interface SourceLocation {
 }
 
 export function getLocationByOffset(str: string): LocationByOffset {
+    if (_.isUndefined(str)) {
```
LogvinovLeon: Can we check that above and just not call this function?
I'd prefer that solution because now we're checking for undefined the variable that's guaranteed to be undefined by the type system. Let's either change the types or check before.

LogvinovLeon:
Prettier fails on CI. Please run `yarn prettier` to fix formatting

Issue #681, opened 2018-06-10
Cannot fill order on Portal DApp


fuJiin:
## Expected Behavior

Tried filling an order generated from Portal DApp's "Generate Order" page. I've verified that I have sufficient balance, and the order should go through.

## Current Behavior

I'm getting a "Failed to fill order, please refresh and try again" error message. I've tried several times and get the same message. See screenshot below.

![0x-fill-order](https://user-images.githubusercontent.com/203277/41207336-185270a0-6cc9-11e8-849a-8c97604fe791.png)

## Possible Reason

Seems like both Infura and `https://mainnet.0xproject.com` are missing `Access-Control-Allow-Origin` headers, resulting in 504s.

## Steps to Reproduce (for bugs)

1. Generate an order from the "Generate order" page (https://0xproject.com/portal).
2. Click on the generated Bit.ly link using a different wallet address.
3. Click "FILL ORDER" on the "Fill order" page (https://0xproject.com/portal/fill). Make sure there is sufficient balance for the token you are filling with.

## Context

We're testing behavior of a modified ERC20 token on mainnet before launching. Wanted to make sure the modifications don't break compatibility with 0x DEXs.

## Your Environment

| Network |
| ------- |
| mainnet    |

fragosti:
@fuJiin thanks for the bug! I am seeing similar behavior but not because of the CORS issues you mention. Can you confirm that the order fill is actually not going through for you, because it is for me although the UI errors for another reason.

@fabioberger @BMillman19 there is some code after the transaction is mined that attempts to extract `filledTakerTokenAmount` from the logs and is failing / crashing. Know anything about this?
![screen shot 2018-06-12 at 11 27 47 am](https://user-images.githubusercontent.com/3066135/41309576-fbe0cc8e-6e33-11e8-8536-9be6a2aef652.png)

fragosti:
The fix has been deployed.

fuJiin:
Hi there - still running into issues here. 

Doing this via API directly shows that the estimated gas provided is insufficient, and that increasing `gasLimit` to 500,000 fixed the issue. The UI may have a similar issue, so you folks can either increase the default `gasLimit` or let users set it themselves.

fragosti:
@fuJiin the portal dApp is fixed as far as I can tell. If you are seeing issues with the API, can you open a different issue and offer steps to reproduce?

fuJiin:
[...]

PR #704, opened 2018-06-14
Fix filling orders on Portal

fragosti:
## Description

Filling an order on portal was broken (UI said transaction didn't go through, but it did). This was because we had some logic that was depending on the log from the transaction being correctly decoded and it was not being decoded correctly. Apparently Web3Wrapper came bundled with the ABIs at one point, but then that changed, and now you need to add the ABIs manually.

Thanks @LogvinovLeon for help on this. He will follow up on this issue more generally in https://app.asana.com/0/628666249318202/711799808770003

This will close https://github.com/0xProject/0x-monorepo/issues/681

## Motivation and Context

Filling an order is useful.

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-wrappers/src/contract_wrappers/contract_wrapper.ts
+++ b/packages/contract-wrappers/src/contract_wrappers/contract_wrapper.ts
@@ -35,7 +35,8 @@ const CONTRACT_NAME_TO_NOT_FOUND_ERROR: {
     Exchange: ContractWrappersError.ExchangeContractDoesNotExist,
 };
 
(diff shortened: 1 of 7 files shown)


packages/contract-wrappers/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "0.0.3",
```
BMillman19: I believe this should be a minor bump to 0.1.0 according to our conventions

PR #705, opened 2018-06-14
Introduce subprovider for printing revert stack traces

albrow:
## Description

This PR introduces a new subprovider, `RevertTraceSubprovider`, which can be enabled in the contracts tests by setting `SOLIDITY_REVERT_TRACE=true` (or running  `yarn wsrun test:trace contracts`).

Much like the coverage and profiler subproviders, the `RevertTraceSubprovider` will inject itself into calls to `eth_sendTransaction`, `eth_call`, and `eth_estimateGas`. It gets the trace data for each transaction, keeps track of the call stack, and looks for the `REVERT` opcode. If it sees the `REVERT` opcode, it will simply print out the stack trace which led to the revert.

In effect this means that if our tests cause a revert, we will be able to see the origin of that revert as a line number in a Solidity file 🎉 

screen shot 2018-06-13 at 7 10 36 pm
screen shot 2018-06-13 at 7 10 28 pm

It's not perfect, but it's a good starting point. One thing we would like to add in the future is including function calls in the stack trace. Right now, entries only appear in the stack trace when we call to another contract (e.g., via the `CALL` opcode) and we don't track function calls within the same contract.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/utils/web3_wrapper.ts
+++ b/packages/contracts/src/utils/web3_wrapper.ts
@@ -7,6 +7,8 @@ import { coverage } from './coverage';
 import { profiler } from './profiler';
 import { revertTrace } from './revert_trace';
 
+import * as _ from 'lodash';
+
 enum ProviderType {
     Ganache = 'ganache',
     Geth = 'geth',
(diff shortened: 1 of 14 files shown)


packages/contracts/src/utils/web3_wrapper.ts
```diff
@@ -48,10 +50,11 @@ const providerConfigs = testProvider === ProviderType.Ganache ? ganacheConfigs :
...
 const isProfilerEnabled = env.parseBoolean(EnvVars.SolidityProfiler);
-if (isCoverageEnabled && isProfilerEnabled) {
-    throw new Error(
-        `Unfortunately for now you can't enable both coverage and profiler at the same time. They both use coverage.json file and there is no way to configure that.`,
-    );
+const isRevertTraceEnabled = env.parseBoolean(EnvVars.SolidityRevertTrace);
+const enabledSubproviderCount = _.filter([isCoverageEnabled, isProfilerEnabled, isRevertTraceEnabled], _.identity)
+    .length;
+if (enabledSubproviderCount > 1) {
+    throw new Error(`Only one of coverage, profiler, and revert trace subproviders can be enabled at a time`);
```
LogvinovLeon: Maybe use or instead of and?

jdkanani:
I just added revert trace for @trufflesuite. Have to manage `sources` and `sourceCodes` fields globally instead of contract-wise.

Thanks for inspiration :) @dekz

jdkanani:
I created `sol-trace` https://github.com/maticnetwork/sol-trace package for @trufflesuite.

If you have the plan to support it in `sol-cov` out of the box, let me know I will make `sol-trace` deprecated.

PR #712, opened 2018-06-16
Change relayer grid tile to use logos and primary colors

BMillman19:
screen shot 2018-06-16 at 1 15 41 pm

## Description

Update relayer grid tile design and added max width to the portal layout

## Motivation and Context

Previous screenshot header images were not really useful, margins increase readability

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/relayer_index/relayer_grid_tile.tsx
+++ b/packages/website/ts/components/relayer_index/relayer_grid_tile.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 
 import { TopTokens } from 'ts/components/relayer_index/relayer_top_tokens';
 import { Container } from 'ts/components/ui/container';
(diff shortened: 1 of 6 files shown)


packages/website/ts/components/relayer_index/relayer_grid_tile.tsx
```diff
@@ -125,11 +128,16 @@ class ImgWithFallback extends React.Component<ImgWithFallbackProps, ImgWithFallb
         };
     }
     public render(): React.ReactNode {
-        if (this.state.imageLoadFailed || _.isUndefined(this.props.src)) {
```
fragosti: I think this logic is useful and re-usable so probably should be in its own file. We can even call it `Image` and just have the fallback thing be part of the API.

PR #717, opened 2018-06-18
ERC721Proxy Always call safeTransferFrom

dekz:
## Motivation and Context

In the cases where it is unknown who the taker is the maker should not dictate the behaviour of transfer. It the taker is a contract we should still check whether this contract is able to receive the ERC721 token. This will prevent accidental lockups in filter/forwarding contracts and be aligned with the intention of the standard.

Some contracts may rely on the behaviour of the callback being called, a maker could disrupt this when providing null receiver data. 

## How Has This Been Tested?

Modified tests

## Types of changes

## Checklist:


dekz:
From initial tests on GETH (0x-devnet) this is an increase in 3000 gas cost between safeTransferFrom and transferFrom. This is an unexpected increase and is not immediately obvious as to why this is the case. Removing any additional callback code (simply a transferFrom wrapper) also increases gas by around 2000. 

This perplexes me.

abandeali1:
So it looks like the `canTransfer` modifier is redundant on both versions of `safeTransferFrom` (since they just call `transferFrom`, which already has it). If we remove those, looks like we save about 3000 gas?

hysz:
And on `safeTransferFrom` with no additional data `canTransfer` would be called 3x...

PR #718, opened 2018-06-18
Improve Pre-publishing flow

fabioberger:
## Description

In trying to publish, we discovered several scenarios that were breaking the flow:
- Trying to publish a version that has been unpublished on NPM. 
- Trying to push create git tags (locally and remotely) that already existed.
- Package.json versions that didn't match NPM and CHANGELOG versions that did not match package.json.

This PR adds pre-publish checks to make sure these cases are not present before running the publishing flow. As a consequence, we were also able to remove the `remove_tags` script.

## Motivation and Context

Make the publishing flow less likely to fail.

## How Has This Been Tested?

Manually.

## Types of changes

## Checklist:


packages/monorepo-scripts/src/prepublish_checks.ts
```diff
@@ -1,9 +1,129 @@
...
+    const localTagVersionsByPackageName = await utils.getGitTagsByPackageNameAsync(packageNames, localGitTags);
+
+    const remoteGitTags = await utils.getRemoteGitTagsAsync();
+    const remoteTagVersionsByPackageName = await utils.getGitTagsByPackageNameAsync(packageNames, remoteGitTags);
+
+    for (const lernaPackage of updatedPublicLernaPackages) {
+        const currentVersion = lernaPackage.package.version;
+        const packageName = lernaPackage.package.name;
+        const packageLocation = lernaPackage.location;
+        const nextVersion = await utils.getNextPackageVersionAsync(currentVersion, packageName, packageLocation);
```
LogvinovLeon: I think - we should remove any tag that's bigger. Not just the next version tag.

packages/monorepo-scripts/src/prepublish_checks.ts
```diff
@@ -1,9 +1,129 @@
...
+        throw new Error(`Please fix the above package.json/NPM inconsistencies.`);
+    }
+}
+
+async function checkChangelogFormatAsync(updatedPublicLernaPackages: LernaPackage[]): Promise<void> {
+    utils.log('Check CHANGELOGs for inconsistencies...');
+    const changeLogInconsistencies = [];
+    for (const lernaPackage of updatedPublicLernaPackages) {
+        const packageName = lernaPackage.package.name;
+        const changelog = changelogUtils.getChangelogOrCreateIfMissing(packageName, lernaPackage.location);
```
LogvinovLeon: Should we create a changelog in prepublish checks? Maybe just skip if we don't have one?

packages/monorepo-scripts/src/utils/changelog_utils.ts
```diff
@@ -44,12 +51,58 @@ export const changelogUtils = {
...
+        fs.writeFileSync(changelogJSONPath, JSON.stringify(changelog, null, '\t'));
+        await this.prettifyAsync(changelogJSONPath, constants.monorepoRootPath);
+    },
+    async writeChangelogMdFileAsync(packageLocation: string, changelog: Changelog): Promise<void> {
+        const changelogMarkdownPath = path.join(packageLocation, 'CHANGELOG.md');
+        fs.writeFileSync(changelogMarkdownPath, JSON.stringify(changelog, null, '\t'));
+        await this.prettifyAsync(changelogMarkdownPath, constants.monorepoRootPath);
+    },
+    async prettifyAsync(filePath: string, cwd: string): Promise<void> {
+        await execAsync(`prettier --write ${filePath} --config .prettierrc`, {
```
LogvinovLeon: prettier has an api. But I guess it's OK.

packages/monorepo-scripts/src/utils/npm_utils.ts
```diff
@@ -0,0 +1,28 @@
+import 'isomorphic-fetch';
+import * as _ from 'lodash';
+
+import { PackageRegistryJson } from '../types';
+
+const NPM_REGISTRY_BASE_URL = 'https://registry.npmjs.org';
+const SUCCESS_STATUS = 200;
```
LogvinovLeon: https://www.npmjs.com/package/http-status-codes
Let's stop hard-coding it in every place

LogvinovLeon: Will be better for grepability

packages/monorepo-scripts/src/utils/semver_utils.ts
```diff
@@ -0,0 +1,56 @@
+import * as _ from 'lodash';
+import semverSort = require('semver-sort');
+
+// Regex that matches semantic versions only including digits and dots.
+const SEM_VER_REGEX = /^(\d+\.){1}(\d+\.){1}(\d+){1}$/gm;
```
LogvinovLeon: Why all those `{1}`? Won't it be the same without them?

packages/monorepo-scripts/src/utils/semver_utils.ts
```diff
@@ -0,0 +1,56 @@
...
+     * @param b version to compare a against
+     * @return Whether version a is lessThan version b
+     */
+    lessThan(a: string, b: string): boolean {
+        this.assertValidSemVer('a', a);
+        this.assertValidSemVer('b', b);
+        if (a === b) {
+            return false;
+        }
+        const sortedVersions = semverSort.desc([a, b]);
```
LogvinovLeon: That is a weird way to do it. Why not use that package? https://www.npmjs.com/package/semver It has exactly the function

packages/monorepo-scripts/src/utils/semver_utils.ts
```diff
@@ -0,0 +1,56 @@
...
+        const isAGreaterThanB = sortedVersions[0] === a;
+        return isAGreaterThanB;
+    },
+    assertValidSemVer(variableName: string, version: string): void {
```
LogvinovLeon: https://www.npmjs.com/package/semver `semver.valid`

LogvinovLeon: Read through this package API. It has a lot of goodies

packages/monorepo-scripts/src/utils/utils.ts
```diff
@@ -43,22 +39,82 @@ export const utils = {
...
-            return undefined;
+    async getNextPackageVersionAsync(
+        currentVersion: string,
+        packageName: string,
+        packageLocation: string,
+    ): Promise<string> {
+        let nextVersion;
+        const changelog = changelogUtils.getChangelogOrCreateIfMissing(packageName, packageLocation);
+        if (_.isEmpty(changelog)) {
+            nextVersion = this.getNextPatchVersion(currentVersion);
```
LogvinovLeon: semver.inc(currentVersion, 'patch')

packages/monorepo-scripts/src/utils/utils.ts
```diff
@@ -43,22 +39,82 @@ export const utils = {
...
         }
+        const lastEntry = changelog[0];
+        nextVersion =
+            lastEntry.version === currentVersion ? this.getNextPatchVersion(currentVersion) : lastEntry.version;
```
LogvinovLeon: same

packages/monorepo-scripts/src/prepublish_checks.ts
fabioberger: Only the next version will break the publishing flow. I also can't think of any scenarios where there would be multiple greater versions remaining. I think we can punt on this for now.

LogvinovLeon: I just think it's more assertive. Why not handle the more general case if we can?

packages/monorepo-scripts/src/prepublish_checks.ts
fabioberger: It just creates a file with an empty `[]`. I think that is fine, since it will need to exist eventually.

packages/monorepo-scripts/src/utils/npm_utils.ts
fabioberger: To do this refactor requires editing many files, across packages. I'd rather we do that on `v2-prototype`. I agree with you but will be leaving this for now.

PR #723, opened 2018-06-19
Consolidate account state display message logic

BMillman19:
## Description

* Abstract out logic that renders strings such as "account loading" and "no wallet detected"
* Remove use of empty identicon when no wallet is loaded and replace with a grey circle.
* Use `Container` and `Text` components instead of inline styling

## Motivation and Context

DRY

## How Has This Been Tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/top_bar/provider_display.tsx
+++ b/packages/website/ts/components/top_bar/provider_display.tsx
@@ -9,6 +9,7 @@ import { ProviderPicker } from 'ts/components/top_bar/provider_picker';
 import { Container } from 'ts/components/ui/container';
 import { DropDown } from 'ts/components/ui/drop_down';
 import { Identicon } from 'ts/components/ui/identicon';
+import { Image } from 'ts/components/ui/image';
 import { Text } from 'ts/components/ui/text';
 import { Dispatcher } from 'ts/redux/dispatcher';
 import { colors } from 'ts/style/colors';
@@ -75,7 +76,7 @@ export class ProviderDisplay extends React.Component<ProviderDisplayProps, Provi
                     </Text>
                 </Container>
                 {isProviderMetamask && (
-                    <img src="/images/metamask_icon.png" style={{ width: ROOT_HEIGHT, height: ROOT_HEIGHT }} />
+                    <Image src="/images/metamask_icon.png" height={ROOT_HEIGHT} width={ROOT_HEIGHT} />
                 )}
             </div>
         );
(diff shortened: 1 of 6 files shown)


packages/website/ts/components/top_bar/provider_display.tsx
```diff
@@ -66,21 +61,21 @@ export class ProviderDisplay extends React.Component<ProviderDisplayProps, Provi
...
                     {this._isBlockchainReady() ? (
                         <Identicon address={this.props.userAddress} diameter={ROOT_HEIGHT} />
                     ) : (
                         <CircularProgress size={ROOT_HEIGHT} thickness={2} />
                     )}
                 </div>
-                <div style={{ marginLeft: 12, paddingTop: 3 }}>
-                    <div style={{ fontSize: 16, color: colors.darkGrey }}>{displayMessage}</div>
-                </div>
+                <Container marginLeft="12px" marginRight="12px">
```
fragosti: fwiw I'm a bit conflicted about `Container` in the context of us having bass.css with all the padding / margin utility classes. I think the argument for `Container` is that it makes components more portable, and removes our dependency (although we still use the "flex" class heavily) on a global stylesheet. Also having our styles strictly typed is nice. 
OTH I hate having two ways of doing the same thing, and basscss is the prior art here and works quite well.

Nothing actionable here: just wondering what your opinion is...
We could make `Container` support an API like this https://github.com/jxnblk/bass-component but depend on the global stylesheet (ie. props become classNames). Then at least one day we are better set-up for a migration if we were so inclined. I think it would be pretty easy to make Container extend this component (both are made with styled-components)

packages/website/ts/components/top_bar/provider_display.tsx
```diff
@@ -66,21 +61,21 @@ export class ProviderDisplay extends React.Component<ProviderDisplayProps, Provi
...
+                    <Text fontSize="14px" fontColor={colors.darkGrey}>
+                        {displayMessage}
+                    </Text>
+                </Container>
                 {isProviderMetamask && (
-                    <div style={{ marginLeft: 16 }}>
-                        <img src="/images/metamask_icon.png" style={{ width: ROOT_HEIGHT, height: ROOT_HEIGHT }} />
-                    </div>
+                    <img src="/images/metamask_icon.png" style={{ width: ROOT_HEIGHT, height: ROOT_HEIGHT }} />
```
fragosti: Would be cool if `Image` supported `height` and `width` props in it's API.

packages/website/ts/components/ui/identicon.tsx
```diff
@@ -34,14 +30,22 @@ export class Identicon extends React.Component<IdenticonProps, IdenticonState> {
...
-                <img
-                    src={icon.toDataURL()}
-                    style={{
-                        width: diameter,
-                        height: diameter,
-                        imageRendering: 'pixelated',
-                    }}
-                />
+                {!_.isEmpty(address) ? (
+                    <img
```
fragosti: Same `Image` comment.

PR #725, opened 2018-06-19
Include source code snippets in revert stack traces

albrow:
## Description

This PR improves the `RevertTraceSubprovider` by adding relevant source code snippets to the stack trace. This makes it much easier to follow the stack trace at a glance. Here's an example:

screen shot 2018-06-18 at 5 24 33 pm

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-cov/src/get_source_range_snippet.ts
+++ b/packages/sol-cov/src/get_source_range_snippet.ts
@@ -1,0 +1,179 @@
+import * as ethUtil from 'ethereumjs-util';
+import * as _ from 'lodash';
+import * as Parser from 'solidity-parser-antlr';
+
+import { SingleFileSourceRange, SourceRange, SourceSnippet } from './types';
+import { utils } from './utils';
+
+interface ASTInfo {
+    type: string;
+    node: Parser.ASTNode;
+    name: string | null;
+    range: SingleFileSourceRange;
+}
+
+// Parsing source code for each transaction/code is slow and therefore we cache it
+const parsedSourceByHash: { [sourceHash: string]: Parser.ASTNode } = {};
+
+export function getSourceRangeSnippet(sourceRange: SourceRange, sourceCode: string): SourceSnippet | null {
+    const sourceHash = ethUtil.sha3(sourceCode).toString('hex');
+    if (_.isUndefined(parsedSourceByHash[sourceHash])) {
+        parsedSourceByHash[sourceHash] = Parser.parse(sourceCode, { loc: true });
+    }
+    const astNode = parsedSourceByHash[sourceHash];
+    const visitor = new ASTInfoVisitor();
+    Parser.visit(astNode, visitor);
+    const astInfo = visitor.getASTInfoForRange(sourceRange);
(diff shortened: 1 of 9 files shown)


albrow:
@LogvinovLeon we need to update the type definitions for solidity-parser-antlr to include `loc` in `ASTNode` and a few other changes. What's the best way to do that?

LogvinovLeon:
@albrow Create a PR to DT

LogvinovLeon:
You can convert that stuff to any for the 5 days it takes for the PR to get accepted. Or you can override it in typescript-typings and remove it later.

LogvinovLeon:
It can also take as little as 30 min for the PR that simple to get accepted

packages/sol-cov/src/get_source_range_snippet.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+    const sourceCodeInRange = utils.getRange(sourceCode, sourceRange.location);
+    return {
+        ...astInfo,
+        source: sourceCodeInRange,
+        fileName: sourceRange.fileName,
+    };
+}
+
+// A visitor which collects ASTInfo for contract definitions, function
```
LogvinovLeon: This is outdated

packages/sol-cov/src/revert_trace_subprovider.ts
```diff
@@ -97,18 +98,62 @@ export class RevertTraceSubprovider extends TraceCollectionSubprovider {
...
+    if (snippetString !== '') {
+        result += `:\n        ${snippetString}`;
+    }
+    return result;
+}
+
+function getSourceSnippetString(sourceSnippet: SourceSnippet): string {
+    switch (sourceSnippet.type) {
+        case 'ContractDefinition':
+            return `contract ${sourceSnippet.name}`;
```
LogvinovLeon: Is this the only reason why we copied the ASTVisitor? If so - let;s just parse the name with a regex and delete the second ASTVisitor

albrow:
DefinitelyTyped PR is here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26676

packages/sol-cov/src/revert_trace_subprovider.ts
albrow: There are a lot of differences between this AST visitor and the existing one:

- We use different parser options in order to get the location information as a line and column number (`{ loc: true }`).
- The information we collect as we walk through the AST is different. We build up a list of `ASTInfo` instead of `CoverageEntriesDescription`. For stack traces we're ultimately looking for units of information about a specific range of code (including the relevant section of the source code, the type of AST node, etc).
- We do walk over contract definitions. The existing AST visitor skips them. 
- We don't walk over binary branches. We may be able to skip other types of AST nodes as well, but I left them there for now.

It may be possible to combine the common functionality between these two AST visitors, but it's not exactly low-hanging fruit. IMO we should wait until after we implement tracing function calls.

LogvinovLeon:
DT PR got merged. Let's get tests to pass and merge it!

PR #730, opened 2018-06-19
Assembler orderHash function

Recmo:
## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
@@ -97,24 +97,28 @@ contract LibOrder is
     function hashOrder(Order memory order)
         internal
         pure
-        returns (bytes32)
+        returns (bytes32 result)
     {
-        // Note: changing this to abi.encode results in an error on 0.4.24
-        // UnimplementedFeatureError: Variable inaccessible, too deep inside stack (17)
-        return keccak256(abi.encodePacked(
-            EIP712_ORDER_SCHEMA_HASH,
-            bytes32(order.makerAddress),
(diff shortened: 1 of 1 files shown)


LogvinovLeon:
How much gas per order do we safe with that?

Recmo:
@LogvinovLeon I was hoping you could tell me that :) It shouldn't be more then about 200 gas difference, but there's a chance that `abi.encodePacked` is particularly inefficient, then it can be much more.

LogvinovLeon:
Solidity gas profiler is already merged. You can also just look into ganache logs and compare

Recmo:
I'm using the fillOrder test in `core.ts` line 134.  `ganache.log` has too much noise for me to figure out which tx to look at, so I just added a `console.log` for the fillOrder receipt gas. I then repeatedly ran the same test.

Gas used without this PR: 
```
196115
196179
196179
196179
```

Gas used with: 
```
193541
193605
193541
193605
```

Aside: I'm a bit worried by our tests not being deterministic. We should look into this.

The difference of the averages is 2590 gas saved with this PR. Significantly more than I expected, `abi.encode` is not very efficient.

It saves 1.3% gas. Not amazing, but worth it IMHO.

LogvinovLeon:
Agree. Let's add those results as a justifying comment in code maybe?
Also - could you look into this non-determinism? Shouldn't happen with geth.

abandeali1:
Shocking that this saves so much gas. It's probably worth doing the same for calculating transaction hashes.

abandeali1:
@LogvinovLeon @Recmo our tests aren't deterministic because we use random salts for the orders, resulting in slight differences in calldata costs (I think a few tests use random addresses as well). Agreed we should probably change this.

PR #734, opened 2018-06-19
Update Island shadow styling and use Island in more places

BMillman19:
## Description

* Updated the dropshadow styling we use across portal
* Used the `Island` component in come places where we should be

## Testing instructions

* Go to /portal
* Verify that the shadow looks as designed

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/top_bar/provider_display.tsx
+++ b/packages/website/ts/components/top_bar/provider_display.tsx
@@ -10,6 +10,7 @@ import { Container } from 'ts/components/ui/container';
 import { DropDown } from 'ts/components/ui/drop_down';
 import { Identicon } from 'ts/components/ui/identicon';
 import { Image } from 'ts/components/ui/image';
+import { Island } from 'ts/components/ui/island';
 import { Text } from 'ts/components/ui/text';
 import { Dispatcher } from 'ts/redux/dispatcher';
 import { colors } from 'ts/style/colors';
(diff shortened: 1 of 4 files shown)


packages/website/ts/components/top_bar/provider_display.tsx
```diff
@@ -62,7 +61,7 @@ export class ProviderDisplay extends React.Component<ProviderDisplayProps, Provi
             this.props.providerType === ProviderType.Injected ? injectedProviderName : 'Ledger Nano S';
         const isProviderMetamask = providerTitle === constants.PROVIDER_NAME_METAMASK;
         const hoverActiveNode = (
-            <div className="flex items-center p1" style={styles.root}>
+            <Island className="flex items-center p1" style={styles.root}>
```
fragosti: I forget why I added the `style` prop to `Island`, but I think it would be best if we tried not to use it. I guess I'm not entirely sure if these components are `Island`s but, assuming they are: can we get rid of the style injection by adding a `borderRadius` prop to Island, and achieving all the other things by adding a `Container` with padding wrapping the children (and wrapping the `Island` for paddingTop)

BMillman19: Ok I played around with removing the style prop from Island but it was messing up some layout down the line. I'll add a tech debt ticket for removing style

PR #744, opened 2018-06-21
Remove unreachable PreSigned check

AusIV:
This code was unreachable, as it had the exact same condition as line 206.

## Description

Line 206 checks for signatureType == SignatureType.PreSigned. This was repeated on line 236, but that condition could never be met after the condition on line 206.


AusIV:
I'm not sure why the coverage check failed. It seems to be unable to find something it's looking for, but I can't imagine it's related to my change. In any case, this change should increase code coverage by removing unreachable code.

abandeali1:
Nice catch, thank you!

PR #745, opened 2018-06-21
Emit an event when validator authorizations change

AusIV:
Without an event to tell us when users add and remove validators, orders that use validators must be verified on a recurring basis. By adding an event when a user changes a validator authorization, we can remove orders proactively if a user de-authorizes a validator.

I feel like there should be a test to go with this, but the tests are written in solidity, and I'm not sure how (or if it's even possible) to detect events emitted in solidity.


AusIV:
I'm in no way tied to this particular implementation, but without some kind of event to track when validator authorizations change, OpenRelay won't be able to support orders that use validators.

AusIV:
I see this is already covered by #740 - withdrawing.

PR #747, opened 2018-06-21
Fix some bugs in sol-cov

albrow:
## Description

This PR fixes two bugs in sol-cov:

1. There was sometimes an infinite loop being triggered due to a bug in a `while` loop in __revert_trace_subprovider.ts__.
2. The wrong contract data was sometimes being returned by `getContractDataIfExists`, especially when dealing with interfaces.

## Testing instructions

Run `SOLIDITY_REVERT_TRACE=true yarn wsrun test contracts` and verify output.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-cov/src/revert_trace_subprovider.ts
+++ b/packages/sol-cov/src/revert_trace_subprovider.ts
@@ -75,6 +75,7 @@ export class RevertTraceSubprovider extends TraceCollectionSubprovider {
             }
             const bytecodeHex = stripHexPrefix(bytecode);
             const sourceMap = isContractCreation ? contractData.sourceMap : contractData.sourceMapRuntime;
+
             const pcToSourceRange = parseSourceMap(
                 contractData.sourceCodes,
                 sourceMap,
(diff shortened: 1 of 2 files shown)


packages/sol-cov/src/revert_trace_subprovider.ts
```diff
@@ -88,16 +89,19 @@ export class RevertTraceSubprovider extends TraceCollectionSubprovider {
...
             // mapped source range.
-            while (_.isUndefined(sourceRange)) {
+            while (_.isUndefined(sourceRange) && pc > 0) {
                 sourceRange = pcToSourceRange[pc];
                 pc -= 1;
-                if (pc <= 0) {
-                    this._logger.warn(
-                        `could not find matching sourceRange for structLog: ${evmCallStackEntry.structLog}`,
-                    );
-                    continue;
```
albrow: What I meant to do here was continue to the next iteration of the `for` loop, but since we're inside a `while` loop it was just continuing to the next iteration of the `while` loop which doesn't really do anything. If the `sourceRange` was undefined, this caused an infinite loop.

PR #749, opened 2018-06-21
Converted `hashZeroExTransaction` to assembly. Saves 1k gas

hysz:
## Description

Rewrote `hashZeroExTransaction` in assembly. Similar to what @Recmo did in [this RP](https://github.com/0xProject/0x-monorepo/pull/730/files).

Saved ~1k gas. Benchmarked using test on line 134 of `transactions.ts` (`should transfer the correct amounts when signed by taker and called by sender`).

## Testing instructions

No new tests were created. All existing tests pass.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
@@ -66,7 +66,7 @@ contract MixinTransactions is
             let memPtr := mload(64)
             mstore(memPtr, schemaHash)
             mstore(add(memPtr, 32), salt)
-            mstore(add(memPtr, 64), signerAddress)
+            mstore(add(memPtr, 64), and(signerAddress, 0xffffffffffffffffffffffffffffffffffffffff))
             mstore(add(memPtr, 96), dataHash)
             result := keccak256(memPtr, 128)
         }
--- a/packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
@@ -51,17 +51,27 @@ contract MixinTransactions is
     /// @param signerAddress Address of transaction signer.
     /// @param data AbiV2 encoded calldata.
     /// @return EIP712 hash of the Transaction.
-    function hashZeroExTransaction(uint256 salt, address signerAddress, bytes data)
+    function hashZeroExTransaction(
+        uint256 salt,
+        address signerAddress,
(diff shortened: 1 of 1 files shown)


packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -51,17 +51,25 @@ contract MixinTransactions is
     /// @param signer Address of transaction signer.
     /// @param data AbiV2 encoded calldata.
     /// @return EIP712 hash of the Transaction.
-    function hashZeroExTransaction(uint256 salt, address signer, bytes data)
+    function hashZeroExTransaction(
+        uint256 salt,
+        address signer,
+        bytes data
```
abandeali1: `memory`

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -51,17 +51,25 @@ contract MixinTransactions is
...
+        bytes32 dataHash = keccak256(data);
+        assembly {
+            let memPtr := mload(64)
+            mstore(memPtr, schemaHash)
+            mstore(add(memPtr, 32), salt)
+            mstore(add(memPtr, 64), signer)
+            mstore(add(memPtr, 96), dataHash)
+            result := keccak256(memPtr, 128)
+        }
     }
```
abandeali1: Let's add an explicit `return result;` here.

packages/contracts/src/contracts/current/protocol/Exchange/MixinTransactions.sol
```diff
@@ -51,17 +51,27 @@ contract MixinTransactions is
...
-            signerAddress,
-            keccak256(data)
-        ));
+        bytes32 schemaHash = EIP712_ZEROEX_TRANSACTION_SCHEMA_HASH;
+        bytes32 dataHash = keccak256(data);
+        assembly {
+            let memPtr := mload(64)
+            mstore(memPtr, schemaHash)
+            mstore(add(memPtr, 32), salt)
+            mstore(add(memPtr, 64), signerAddress)
```
Recmo: Solidity afaik does not guarantee that unused bits of an address are zero. To be safe we should do a `signerAddress := and(signerAddress, 0x0xffffffffffffffffffffffffffffffffffffffff)`.

hysz: Ooh yeah I don't think that's undefined. Good catch 👍

PR #760, opened 2018-06-22
Check Revert Reasons in Contract Tests

fabioberger:
## Description

Before this PR, our contract tests only tested **if** a revert happened, but not **which** revert happened (except in SignatureValidator & LibBytes tests). This PR checks the revert reason emitted by the contract, so that we are sure the test failed for the expected reason.

In doing so, I uncovered a few places where the tests weren't testing what it was intended to test. Furthermore, I also found a spot that was missing an explicit `require` statement (See: #748). Lastly, I also changed some revert reasons to follow the codebase convention of being all uppercase, snake-case.

This process also made me uncover several revert reasons that aren't currently tested for:

- INVALID_SENDER
- INVALID_SIGNATURE
- REENTRANCY_ILLEGAL
- INVALID_TX_SIGNATURE
- LENGTH_65_REQUIRED
- INDEX_OUT_OF_BOUNDS
- AUTHORIZED_ADDRESS_MISMATCH

We should add tests for these cases.

Miscellaneous:
- Added a `remove_node_modules` yarn command that removes all node_modules folders, nested and hoisted.

## Testing instructions

Run `PKG=contracts yarn test`

## Types of changes

## Checklist:


packages/contracts/test/asset_proxy/proxies.ts
```diff
@@ -160,14 +164,15 @@ describe('Asset Transfer Proxies', () => {
...
                 // Perform a transfer; expect this to fail.
-                return expectRevertOrAlwaysFailingTransactionAsync(
+                return expectRevertReasonOrAlwaysFailingTransactionAsync(
                     erc20Proxy.transferFrom.sendTransactionAsync(
                         encodedAssetData,
                         makerAddress,
                         takerAddress,
                         transferAmount,
-                        { from: notAuthorized },
+                        { from: exchangeAddress },
```
fabioberger: This is an example of an incorrect test that was reverting because `from` was unauthorized, not before allowances were too low.

fabioberger:
We are going to wait with merging this until a couple other PR's get merged in.

packages/contracts/src/utils/assertions.ts
```diff
@@ -51,6 +51,18 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
 
+/**
+ * Rejects if the given Promise does not reject with the given revert reason or "always
+ * failing transaction" error.
+ * @param p the Promise which is expected to reject
+ * @param reason a specific revert reason
+ * @returns a new Promise which will reject if the conditions are not met and
+ * otherwise resolve with no value.
+ */
+export function expectRevertReasonOrAlwaysFailingTransactionAsync<T>(p: Promise<T>, reason: string): PromiseLike<void> {
```
albrow: Could we change this to `reason: RevertReasons` to encourage people to add any new reasons to the enum instead of passing a raw string?

Maybe we should also change the name of the enum to the singular `RevertReason` too since a plural name for a type reads a little awkwardly. [Singular names for enum types](https://www.typescriptlang.org/docs/handbook/enums.html) seem more common.

packages/contracts/test/exchange/transactions.ts
```diff
@@ -59,6 +59,12 @@ describe('Exchange transactions', () => {
     after(async () => {
         await blockchainLifecycle.revertAsync();
     });
+    beforeEach(async () => {
```
albrow: 🙏

PR #762, opened 2018-06-22
Sort wallet tokens by timestamp added and scroll to newly added token

BMillman19:
## Description

* Add `trackedTimestamp` property to `Token` interface
* Sort tokens by `trackedTimestamp` in wallet
* Add ids to token rows

## Testing Instructions

* Keep adding tokens to your wallet

## Types of changes

## Checklist:


packages/website/ts/components/generate_order/asset_picker.tsx
```diff
@@ -257,6 +257,7 @@ export class AssetPicker extends React.Component<AssetPickerProps, AssetPickerSt
         };
 
         newTokenEntry.isTracked = true;
+        newTokenEntry.trackedTimestamp = Date.now();
```
fragosti: I've seen us use `moment().unix()` in these instances instead of Date.

packages/website/ts/types.ts
```diff
@@ -15,6 +15,7 @@ export interface Token {
     decimals: number;
     isTracked: boolean;
     isRegistered: boolean;
+    trackedTimestamp?: number;
```
fragosti: Miiiiiiiight want to consider consolidating `isTracked` and `trackedTimestamp` just because having both opens us up to bugs where they are inconsistent (ex: isTracked  = true, trackedTimestamp: undefined }. So this would be mean just having `trackedTimeStamp?`. Also blockchain.ts does some work with `isTracked` -- is that still valid?

packages/website/ts/components/generate_order/asset_picker.tsx
fabioberger: Yeah, let's try and always use `moment`. It's a much more robust date/time library then `Date`.

packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -149,6 +149,20 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
...
+    public componentDidUpdate(prevProps: WalletProps): void {
+        const currentTrackedTokens = this.props.trackedTokens;
+        const differentTrackedTokens = _.difference(currentTrackedTokens, prevProps.trackedTokens);
+        const firstDifferentTrackedToken = _.head(differentTrackedTokens);
+        // check if there is only one different token, and if that token is a member of the current tracked tokens
+        // this means that the token was added, not removed
+        if (
+            !_.isUndefined(firstDifferentTrackedToken) &&
+            _.size(differentTrackedTokens) === 1 &&
+            currentTrackedTokens.includes(firstDifferentTrackedToken)
```
fabioberger: Use `_.includes`. Our convention is to always use Lodash for consistencies sake.

fragosti: Oh. Ok. Will do.

packages/website/ts/types.ts
BMillman19: My only worry about dropping the isTracked field is for users who already have tracked tokens in storage, we may be dropping them

fragosti: Oh good point. The backwards compat argument is good. How will those users experience adding new tokens? Will we be prepending to Wallet or appending? If it's pre-pending, then it might be worth writing a little migration script anyway.

BMillman19: We do have some utilities for clearing tracked tokens local storage for backwards incompatible changes

fragosti: Up to you! I do think the chances of us running into this inconsistency in an unexpected way are relatively high down the line. Maybe this falls under https://app.asana.com/0/search/720658791610697/690670364591515

Where we would redux-ify all this logic and call moment().unix() in the reducer (and maybeee get rid of isTracked)

PR #765, opened 2018-06-22
[sol-cov] show when modifier _; is covered

perissology:
## Description

Adds support for marking `_;` statements within a modifier as covered.

I'm not too familiar w/ ts, so I'm not sure if it's possible to check the type of `ast.expression` such as:
```
if (ast.expression instanceof Parser.Identity) {...}
```

lmk if there is a better way to make that check in the if statement.

Also, the types for `solidity-parser-antlr` need to be updated to include `name` in the `Identifier` node

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-cov/src/ast_visitor.ts
+++ b/packages/sol-cov/src/ast_visitor.ts
@@ -79,7 +79,8 @@ export class ASTVisitor {
         this._visitStatement(ast);
     }
     public ExpressionStatement(ast: Parser.ExpressionStatement): void {
-        if (ast.expression.type === 'Identifier' && ast.expression.name === '_') {
+        // TODO update types
+        if (ast.expression.type === 'Identifier' && (ast as any).expression.name === '_') {
             // this is "_;" in a modifier
             // add to the modifiersStatementdIds so we can collect coverage info
             this._modifiersStatementIds.push(this._entryId);
--- a/packages/sol-cov/src/ast_visitor.ts
+++ b/packages/sol-cov/src/ast_visitor.ts
@@ -79,6 +79,11 @@ export class ASTVisitor {
         this._visitStatement(ast);
     }
     public ExpressionStatement(ast: Parser.ExpressionStatement): void {
+        if (ast.expression.type === 'Identifier' && ast.expression.name === '_') {
+            // this is "_;" in a modifier
+            // add to the modifiersStatementdIds so we can collect coverage info
+            this._modifiersStatementIds.push(this._entryId);
+        }
         this._visitStatement(ast.expression);
     }
     public InlineAssemblyStatement(ast: Parser.InlineAssemblyStatement): void {


LogvinovLeon:
You'd need to submit a PR to DefinitelyTyped to update the typings for that package.
It might be confusing for a ts newcomer - so I'll do that for you. In the meantime, you can either cast that thing to `any` before use or override the types in typescript-typings in the repo.

```typescript
stuff.field // Field not found
(stuff as any).field // OK
```

LogvinovLeon:
I'm not sure I understand the intent of that PR.
`modifierStatementIds` mapping is kinda a hack because no opcode maps back to the invocation of the modifier with no parameters. So when we have:
```solidity
function f() 
onlyOwner
hasAParam(a)
{
}
```
we detect hasParam invocation line as covered because code writes the args into calldata, but onlyOwner is uncovered because no code was needed there. That's why we check if there is a covered modifier after and if it is - we assume that this one is also covered.
`_modifiersStatementIds` collect the ids of the modifier invocation statements. Not the code inside the modifiers. The code inside should be marked as covered either way. If it isn't - we should fix that. I just don't feel like this would solve it. Also - could you write a test for it?

perissology:
right, with modifiers, the compiler will replace the `_;` with a jump to the function that invoked the modifier, thus there is no opcode that maps back to that line.

coverage report w/o this PR:
screen shot 2018-06-25 at 7 44 48 am

coverage report w/ this PR:
screen shot 2018-06-25 at 7 46 15 am

LogvinovLeon:
Ok. Makes sense. But why doesn't compiler map that `JUMP` opcode to that modifier? Where does it map it to? Seems like a compiler bug to me. If it is - let's please report and mark that solution as a HACK with a link to an issue in Solidity. I'm OK merging it, just want to dig into that a little bit more and see where does that opcode map.

LogvinovLeon:
Also - if I do get it right this code:
```solidity
modifier mod() {
    if (true) {
        _;
    }
    if (false) {
        _;
    }
}
```

Will now show both modifiers as covered even though second one could not possibly be.
Look at [this logic](https://github.com/0xProject/0x-monorepo/blob/v2-prototype/packages/sol-cov/src/coverage_subprovider.ts#L111). It will find the enclosing function (mod in our case) and mark the modifier as covered whenever there is any statement within that function that is covered. This makes sense when modifierStatementIds contain only modifier invocations, but this PR is trying to reuse it for smth else and it's not quite working. I understant that we all want higher coverage and getting 85% from the tool when we actually have 99% is annoying and damaging for the brand. I'd be down to go deeper down the road of sol-cov hints.
/* solcov covered statement */

/* solcov covered branch */

/* solcov covered function */

etc...
as well as proposed [here](https://github.com/0xProject/0x-monorepo/pull/766):
/* solcov ignore next */

Actually - marking some code as covered will raise your coverage results more than ignoring uncovered code ;)

perissology:
So it appears that it is optimized away. Here's some screen shots viewing the opcode & corresponding source:

This shows 4 trace steps surrounding the jumpdest.

screen shot 2018-06-27 at 8 39 52 am
screen shot 2018-06-27 at 8 39 36 am
screen shot 2018-06-27 at 8 39 55 am
screen shot 2018-06-27 at 8 40 01 am


Your right in the fact that both of the following will be marked as covered:
```
modifier mod() {
    if (true) {
        _;
    }
    if (false) {
        _;
    }
}
```

I opened [this issue](https://github.com/ethereum/solidity/issues/4364) to see if this is a bug in the compiler or not

perissology:
yes, no rush. Using `/* solcov ignore next */` works fine for now

LogvinovLeon:
Looks like the solidity issue is a wontfix and is closed. That means that the solution to that issue is `/* solcov ignore next */`

PR #774, opened 2018-06-25
Added some comments for ASM hashing

hysz:
## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
+++ b/packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
@@ -99,11 +99,7 @@ contract LibOrder is
         bytes32 makerAssetDataHash = keccak256(order.makerAssetData);
         bytes32 takerAssetDataHash = keccak256(order.takerAssetData);
         // Assembly for more efficiently computing:
-        // keccak256(abi.encodePacked(
-        //   DOMAIN_SEPARATOR_SCHEMA_HASH,
-        //   keccak256(abi.encodePacked(address(this))),
-        //   ORDER_SCHEMA_HASH,
-        //   keccak256(abi.encodePacked(
+        //   keccak256(abi.encode(
         //       order.makerAddress,
         //       order.takerAddress,
         //       order.feeRecipientAddress,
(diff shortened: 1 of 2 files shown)


packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -98,6 +98,26 @@ contract LibOrder is
         bytes32 schemaHash = EIP712_ORDER_SCHEMA_HASH;
         bytes32 makerAssetDataHash = keccak256(order.makerAssetData);
         bytes32 takerAssetDataHash = keccak256(order.takerAssetData);
+        // Assembly for more efficiently computing:
+        // keccak256(abi.encodePacked(
+        //   DOMAIN_SEPARATOR_SCHEMA_HASH,
+        //   keccak256(abi.encodePacked(address(this))),
```
abandeali1: Let's drop the `encodePacked` from the single argument hashes.

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -98,6 +98,26 @@ contract LibOrder is
...
+        //   ORDER_SCHEMA_HASH,
+        //   keccak256(abi.encodePacked(
```
Recmo: `abi.encode` not `abi.encodePacked` (which would produce a different hash value).

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -98,6 +98,26 @@ contract LibOrder is
         bytes32 schemaHash = EIP712_ORDER_SCHEMA_HASH;
         bytes32 makerAssetDataHash = keccak256(order.makerAssetData);
         bytes32 takerAssetDataHash = keccak256(order.takerAssetData);
+        // Assembly for more efficiently computing:
+        // keccak256(abi.encodePacked(
```
Recmo: This function does not compute the outer hash, only the inner one (lines 106-119).

packages/contracts/src/contracts/current/protocol/Exchange/libs/LibOrder.sol
Recmo: `abi.encode` not `abi.encodePacked` (which would produce a different hash value).

PR #660, opened 2018-06-04
ETH LightWallet Subprovider

cavanmflynn:
## Description

Adds an [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet) subprovider.

## Motivation and Context

Allows users to hook eth-lightwallet into the web3-provider-engine.

## How Has This Been Tested?

Unit tests, which call the subprovider methods directly as well as through a provider.

## Types of changes

*   [ ] Bug fix (non-breaking change which fixes an issue)
*   [x] New feature (non-breaking change which adds functionality)
*   [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

*   [ ] Change requires a change to the documentation.
*   [x] Added tests to cover my changes.
*   [x] Added new entries to the relevant CHANGELOG.jsons.
*   [x] Labeled this PR with the 'WIP' label if it is a work in progress.
*   [x] Labeled this PR with the labels corresponding to the changed package.

PR #775, opened 2018-06-25
Eth lightwallet subprovider (replaces #660)

fabioberger:
## Description

Adds an [eth-lightwallet](https://github.com/ConsenSys/eth-lightwallet) subprovider.

Also fixes several TSLint errors that weren't caught on the v2-prototype branch for some reason.

## Testing instructions

Unit tests, which call the subprovider methods directly as well as through a provider.

## Types of changes

## Checklist:


packages/subproviders/test/unit/eth_lightwallet_subprovider_test.ts
```diff
@@ -0,0 +1,185 @@
...
+    });
+    describe('direct method calls', () => {
+        describe('success cases', () => {
+            it('returns a list of accounts', async () => {
+                const accounts = await ethLightwalletSubprovider.getAccountsAsync();
+                expect(accounts[0]).to.be.equal(FAKE_ADDRESS);
+                expect(accounts.length).to.be.equal(NUM_GENERATED_ADDRESSES);
+            });
+            it('signs a personal message hash', async () => {
+                const signingAccount = (await ethLightwalletSubprovider.getAccountsAsync())[0];
```
dekz:
```typescript
const accounts = await ethLightwalletSubprovider.getAccountsAsync();
const signingAccount = accounts[0];
```
for readability

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
```diff
@@ -0,0 +1,83 @@
...
+     * @param address Address of the account to sign with
+     * @return Signature hex string (order: rsv)
+     */
+    public async signPersonalMessageAsync(data: string, address: string): Promise<string> {
+        if (_.isUndefined(data)) {
+            throw new Error(WalletSubproviderErrors.DataMissingForSignPersonalMessage);
+        }
+        assert.isHexString('data', data);
+        assert.isETHAddressHex('address', address);
+        const result: ECSignatureBuffer = lightwallet.signing.signMsgHash(
```
dekz: Is this type not inferred from the type definition?

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
```diff
@@ -0,0 +1,83 @@
...
+     * @return Signed transaction hex string
+     */
+    public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
+        if (_.isUndefined(txParams.from) || !addressUtils.isAddress(txParams.from)) {
+            throw new Error(WalletSubproviderErrors.FromAddressMissingOrInvalid);
+        }
+
+        const tx = new EthereumTx(txParams);
+        const txHex = tx.serialize().toString('hex');
+        let signedTxHex: string = lightwallet.signing.signTx(this._keystore, this._pwDerivedKey, txHex, txParams.from);
```
dekz: Is this type not inferred from the type definition?

packages/subproviders/test/unit/eth_lightwallet_subprovider_test.ts
```diff
@@ -0,0 +1,185 @@
...
+
+                // Keccak-256 hash of 'hello world'
+                const messageHash = '0x47173285a8d7341e5e972fc677286384f802f8ef42a5ec5f03bbfa254cb01fad';
```
dekz: Does this need to be a hex encoded string? Our other tests just use the string 'hello world' and tests against known fixture data.

dekz:
Had to make a few slight changes. Eth lightwallet doesn't seem to [encode the r,s,v](https://github.com/ConsenSys/eth-lightwallet/blob/master/lib/signing.js#L61) values the same way as ethereumjs-util does in [toRpcSig](https://github.com/ethereumjs/ethereumjs-util/blob/8abcba5f88835f97d0b5da7808aaa99b4eaddba5/index.js#L412).

And Eth lightwallet does not add the personal prefix to the front of the message before hashing. It is usually the norm for `eth_sign` to receive a non prefixed value and the [signer adds the prefix](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_sign). 

I've updated the tests to use our mnemonic and hd path, so it matches fixture data.

dekz:
Looks like there's some loss of information when passing the hex encoded transaction data to eth-lightwallet. chainId is dropped and defaulted to 0. So testnet signature's create a different signature to our test data. 

We can export the private key in the same way that eth-lighwallet does internally and construct a PrivateKeyWalletSubprovider to perform the work. This also shares the validation logic.

packages/subproviders/src/globals.d.ts
```diff
@@ -3,6 +3,59 @@
 // tslint:disable:async-suffix
 // tslint:disable:completed-docs
 
+// Ethereumjs-tx declarations
```
fabioberger: Remove if there aren't any.

packages/subproviders/src/globals.d.ts
```diff
@@ -3,6 +3,59 @@
...
+        public close(): Promise<void>;
+    }
+}
+
+declare module '@ledgerhq/hw-transport-node-hid' {
+    export default class TransportNodeHid implements LedgerTransport {
+        public static create(): Promise<LedgerTransport>;
+        public close(): Promise<void>;
+    }
+}
```
fabioberger: Can we please move these declarations to `typescript-typings` package?

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
```diff
@@ -0,0 +1,73 @@
...
+     * If you are not using this via a ProviderEngine instance, you can call it directly.
+     * @param txParams Parameters of the transaction to sign
+     * @return Signed transaction hex string
+     */
+    public async signTransactionAsync(txParams: PartialTxParams): Promise<string> {
+        // Lightwallet loses the chain id information when hex encoding the transaction
+        // this results in a different signature on certain networks. PrivateKeyWallet
+        // respects this as it uses the parameters passed in
+        let privKey = this._keystore.exportPrivateKey(txParams.from, this._pwDerivedKey);
+        const privKeyWallet = new PrivateKeyWalletSubprovider(privKey);
```
fabioberger: We should instantiate `PrivateKeyWalletSubprovider` in the constructor so we do this once rather then for each signing call.

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
```diff
@@ -0,0 +1,73 @@
...
+        const privKeySignature = await privKeyWallet.signTransactionAsync(txParams);
+        privKey = '';
```
fabioberger: Move this up a line.

packages/subproviders/src/globals.d.ts
```diff
@@ -3,6 +3,59 @@
...
+interface ECSignature {
+    v: number;
+    r: string;
+    s: string;
+}
+interface ECSignatureBuffer {
+    v: number;
+    r: Buffer;
+    s: Buffer;
+}
```
fabioberger: We should remove these types from `global.d.ts`? `ECSignatureBuffer` is already added to `@0xproject/types`, `ECSignature` as well. Let's use them from there.

packages/typescript-typings/CHANGELOG.json
```diff
@@ -3,6 +3,9 @@
         "version": "0.4.2",
         "changes": [
             {
+                "note": "Add types for `eth-lightwallet`",
+                "pr": 775
+            },
                 "note": "Improve 'web3-provider-engine' typings",
```
fabioberger: Bracket missing here.

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
dekz: I have greater concerns with keeping the private keys in memory which seems counter intuitive to eth-lightwallets key vault design.

packages/subproviders/src/globals.d.ts
LogvinovLeon: Or even better - to DT

packages/subproviders/src/globals.d.ts
```diff
@@ -3,6 +3,59 @@
...
+
+declare module '@ledgerhq/hw-transport-u2f' {
+    export default class TransportU2F implements LedgerTransport {
+        public static create(): Promise<LedgerTransport>;
+        public close(): Promise<void>;
+    }
+}
+
+declare module '@ledgerhq/hw-transport-node-hid' {
```
LogvinovLeon: Move all of those declarations to DefinitelyTypes

packages/subproviders/src/globals.d.ts
dekz: err looks like they've already been moved over to `typescript-typings`. I'll drop them from here.

packages/subproviders/src/globals.d.ts
dekz: These are now removed, have they already been submitted to DT? What's the process?

fabioberger: @LogvinovLeon are they complete enough to move to DT?

packages/subproviders/src/subproviders/eth_lightwallet_subprovider.ts
fabioberger: Ah ok, gotcha. I guess we can leave it then.

packages/subproviders/src/globals.d.ts
LogvinovLeon: There are no restrictions on completeness. We can move partial definitions there.

packages/order-watcher/src/order_watcher/order_watcher.ts
```diff
@@ -301,6 +298,7 @@ export class OrderWatcher {
             }
             case EtherTokenEvents.Withdrawal: {
                 // Invalidate cache
+                // tslint:disable-next-line
```
fabioberger: Please specify the specific TSLint rule that you are disabling.

PR #780, opened 2018-06-26
Refactor Blockchain.ts to allow arbitrary provider state changes

fragosti:
## Description
This started off as a task to support the new metamask (which will stop refreshing on network change https://github.com/MetaMask/metamask-extension/pull/4558/files#diff-d0dadea64ae305b53b5c625db0e0ef1fL28) but then I realized that the architecture doesn't support changing network ID (although there was some code that implied it could, what was that trying to achieve, anyone know?) so I refactored things a bit.

Supporting the new metamask involves using a somewhat unofficial (?) api where you can subscribe to changes in metamask.

The refactor involves just having one function when we want to init providers / web3wrappers or when we want to change them (such as on network change, or using the ledger).

## Testing instructions
You need to download the "beta" version of metamask.
https://twitter.com/fragosti/status/1008875873917808645

And then access Portal and Legacy Portal and try going between networks and providers to make sure everything works as expected. 
Also you should try both versions of metamask.

I've done all this already, including testing / fixing ledger support.
## Types of changes

## Checklist:


packages/website/ts/blockchain.ts
```diff
@@ -92,16 +94,65 @@ export class Blockchain {
...
         return providerNameIfExists;
     }
-    private static async _getProviderAsync(injectedWeb3: Web3, networkIdIfExists: number): Promise<Provider> {
+    private static _getInjectedWeb3(): any {
+        return (window as any).web3;
+    }
+    private static _getFallbackNetworkId(): number {
+        return configs.IS_MAINNET_ENABLED ? constants.NETWORK_ID_MAINNET : constants.NETWORK_ID_KOVAN;
+    }
+    private static async _getInjectedWeb3ProviderNetworkIdIfExistsAsync(): Promise<number> {
```
fabioberger: Return type should be `Promise<number | undefined>` no?

packages/website/ts/blockchain.ts
```diff
@@ -92,16 +94,65 @@ export class Blockchain {
...
+            } catch (err) {
+                // Ignore error and proceed with networkId undefined
+            }
+        }
+        return networkIdIfExists;
+    }
+    private static async _getProviderAsync(
+        injectedWeb3: Web3,
+        networkIdIfExists: number,
+        userLedgerProvider: boolean = false,
```
fabioberger: As boolean variable, should follow naming convention: https://github.com/0xProject/0x-monorepo/blob/v2-prototype/packages/tslint-config/rules/booleanNamingRule.ts#L5

packages/website/ts/blockchain.ts
```diff
@@ -92,16 +94,65 @@ export class Blockchain {
...
+    ): Promise<[Provider, LedgerSubprovider]> {
         const doesInjectedWeb3Exist = !_.isUndefined(injectedWeb3);
+        const isNetworkIdDefined = !_.isUndefined(networkIdIfExists);
```
fabioberger: Nit: `isNetworkIdAvailable`

packages/website/ts/blockchain.ts
```diff
@@ -92,16 +94,65 @@ export class Blockchain {
...
+                // Ignore error and proceed with networkId undefined
+            }
+        }
+        return networkIdIfExists;
+    }
+    private static async _getProviderAsync(
+        injectedWeb3: Web3,
+        networkIdIfExists: number,
+        userLedgerProvider: boolean = false,
+    ): Promise<[Provider, LedgerSubprovider]> {
```
fabioberger: Should be `LedgerSubproviderIfExists` since it's value might be `undefined`.

packages/website/ts/blockchain.ts
```diff
@@ -185,84 +238,17 @@ export class Blockchain {
...
-        this._blockchainWatcher = new BlockchainWatcher(
-            this._dispatcher,
-            this._web3Wrapper,
-            this.networkId,
-            shouldPollUserAddress,
-        );
-        this._contractWrappers.setProvider(provider, this.networkId);
-        await this._blockchainWatcher.startEmittingNetworkConnectionAndUserBalanceStateAsync();
-        this._dispatcher.updateProviderType(ProviderType.Ledger);
+        const useLedgerProvider = true;
```
fabioberger: `shouldUseLedgerProvider`

packages/website/ts/blockchain.ts
```diff
@@ -185,84 +238,17 @@ export class Blockchain {
...
-        this._userAddressIfExists = userAddresses[0];
-
-        this._contractWrappers.setProvider(provider, this.networkId);
-
-        await this.fetchTokenInformationAsync();
-        await this._blockchainWatcher.startEmittingNetworkConnectionAndUserBalanceStateAsync();
-        this._dispatcher.updateProviderType(ProviderType.Injected);
-        delete this._ledgerSubprovider;
-        delete this._cachedProvider;
+        const userLedgerProvider = false;
```
fabioberger: Same as above.

packages/website/ts/blockchain.ts
```diff
@@ -185,84 +238,17 @@ export class Blockchain {
...
+        this._dispatcher.updateBlockchainIsLoaded(false);
+        // We don't want to be out of sync with the network metamask declares.
```
fabioberger: Nit: Not just metamask, any extension that's injecting a web3 instance no?

I guess this brings up a larger question, will all this logic work for a non-Metamask injected web3 instance? Is there a way for us to gracefully handle other extensions? I'm thinking of Mist, Toshi, etc...

packages/website/ts/blockchain.ts
```diff
@@ -774,49 +773,62 @@ export class Blockchain {
...
-        if (!_.isUndefined(injectedWeb3)) {
-            try {
-                networkIdIfExists = _.parseInt(await promisify<string>(injectedWeb3.version.getNetwork)());
-            } catch (err) {
-                // Ignore error and proceed with networkId undefined
+        const networkIdIfExists = await Blockchain._getInjectedWeb3ProviderNetworkIdIfExistsAsync();
+        this.networkId = !_.isUndefined(networkIdIfExists) ? networkIdIfExists : Blockchain._getFallbackNetworkId();
+        const injectedWeb3 = Blockchain._getInjectedWeb3();
+        if (injectedWeb3) {
+            const injectedProviderObservable = injectedWeb3.currentProvider.publicConfigStore;
```
fabioberger: To be more defensive (since people might be injecting all kinds of `web3` objects), let's make sure `injectedWeb3.currentProvider` is not `undefined` before indexing into `publicConfigStore`.

packages/website/ts/blockchain.ts
```diff
@@ -774,49 +773,62 @@ export class Blockchain {
...
+            if (injectedProviderObservable && !this._injectedProviderObservable) {
```
fabioberger: Explicit `undefined` checks:

`if (!_.isUndefined(injectedProviderObservable) && _.isUndefined(this._injectedProviderObservable))`

packages/website/ts/components/dialogs/ledger_config_dialog.tsx
```diff
@@ -282,7 +282,10 @@ export class LedgerConfigDialog extends React.Component<LedgerConfigDialogProps,
         if (didSucceed) {
             this.setState({
                 stepIndex: LedgerSteps.SELECT_ADDRESS,
+                connectionErrMsg: '',
             });
+        } else {
+            this.props.dispatcher.updateBlockchainIsLoaded(true);
```
fabioberger: Hm, this seems like an odd place for this. Mind explaining or adding a `// HACK` comment if this is one? If we fail to fetch the addresses and balances, we set blockchainIsLoaded to `true`?

packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
```diff
@@ -159,9 +159,11 @@ class PlainPortalOnboardingFlow extends React.Component<PortalOnboardingFlowProp
         const ethToken = utils.getEthToken(this.props.tokenByAddress);
         const zrxToken = utils.getZrxToken(this.props.tokenByAddress);
         if (ethToken && zrxToken) {
-            const ethTokenAllowance = this.props.trackedTokenStateByAddress[ethToken.address].allowance;
-            const zrxTokenAllowance = this.props.trackedTokenStateByAddress[zrxToken.address].allowance;
-            return ethTokenAllowance > new BigNumber(0) && zrxTokenAllowance > new BigNumber(0);
+            const ethTokenState = this.props.trackedTokenStateByAddress[ethToken.address];
+            const zrxTokenState = this.props.trackedTokenStateByAddress[zrxToken.address];
+            if (ethTokenState && zrxTokenState) {
+                return ethTokenState.allowance > new BigNumber(0) && zrxTokenState.allowance > new BigNumber(0);
```
fabioberger: Use `ethTokenState.allowance.gt(0)` and `zrxTokenState.allowance.gt(0)`

packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
```diff
@@ -222,6 +224,9 @@ class PlainPortalOnboardingFlow extends React.Component<PortalOnboardingFlowProp
             return null;
         }
         const tokenState = this.props.trackedTokenStateByAddress[token.address];
```
fabioberger: `tokenStateIfExists`

packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
```diff
@@ -222,6 +224,9 @@ class PlainPortalOnboardingFlow extends React.Component<PortalOnboardingFlowProp
...
+        if (!tokenState) {
```
fabioberger: `_.isUndefined(tokenState)`

packages/website/ts/blockchain.ts
```diff
@@ -92,16 +94,65 @@ export class Blockchain {
         }
         return providerNameIfExists;
     }
-    private static async _getProviderAsync(injectedWeb3: Web3, networkIdIfExists: number): Promise<Provider> {
+    private static _getInjectedWeb3(): any {
+        return (window as any).web3;
+    }
+    private static _getFallbackNetworkId(): number {
+        return configs.IS_MAINNET_ENABLED ? constants.NETWORK_ID_MAINNET : constants.NETWORK_ID_KOVAN;
```
BMillman19: We can probably remove `IS_MAINNET_ENABLED`, I think it was just a feature flag for before 0x was deployed on mainnet

packages/website/ts/components/dialogs/ledger_config_dialog.tsx
fragosti: Yea now that you mention it the behavior isn't as it should be. It starts loading immediately when you click to switch to ledger (which is why I opted to remove the loading state for failures, or there is no way to get out of it)

fragosti:
@BMillman19 Toshi doesn't seem to let you dynamically change networks. Cipher does but I just tried it and it behaves like dogfood right now, which is that you have to refresh. I could re-write some polling logic that polls the injected provider but would rather do that in a separate branch.

packages/website/ts/blockchain.ts
fabioberger: Ah yep, that works!

packages/website/ts/components/dialogs/ledger_config_dialog.tsx
fabioberger: ok, either let's refactor or leave a comment explaining.

PR #786, opened 2018-06-27
Add and apply solidity linter

focuzz:
## Adds Security and [Style Guide](http://solidity.readthedocs.io/en/develop/style-guide.html) checking and validation on solidity contracts.

This PR adds [Solhint](https://github.com/protofire/solhint) to achieve this goal.
I have only updated the code on those contracts where it had no impact on the functionality, mostly indentation issues. This should make code review fairly straightforward.

I have only configured the linter to output a warning instead of throwing errors as a starting point. Solhint can add more value. If you like this approach, more time can be invested analyzing and fixing those issues. Or you can simply update the linter configuration to ignore them without a warning.

Run:
`npm run lint-contracts`


packages/contracts/package.json
```diff
@@ -31,7 +31,8 @@
         "coverage:report:html": "istanbul report html && open coverage/index.html",
         "profiler:report:html": "istanbul report html && open coverage/index.html",
         "coverage:report:lcov": "istanbul report lcov",
-        "test:circleci": "yarn test"
+        "test:circleci": "yarn test",
+        "lint-contracts": "solhint src/contracts/**/*.sol"
     },
```
feuGeneA: consider also appending ` && yarn lint-contracts` to the `lint` script command.

fabioberger:
@focuzz, thanks for the PR! As you can see, we've just underwent a major directory restructuring of our contracts. Would you mind dropping all your changes to .sol files and re-running `solhint --fix` on the contracts?

LogvinovLeon:
@focuzz Sorry it's taking so long. @abandeali1 and @hysz will look into that later, but first and foremost - rebase this PR of the latest v2-prototype. We've just changed the directory structure.
Then - remove linting of external smart contracts (EtherDelta, AccountLevels, interfaces etc).
I don't think it makes sense for us to rewrite EtherDelta. Then remove linting of v1 contracts. They're already deployed and immutable.

abandeali1:
Agree with @LogvinovLeon , we should be ignoring all of the v1 contracts as well as `MultiSigWallet`. We should not be making any changes that modify the bytecode of already deployed contracts.

I'd also be interested to hear your reasoning for using solhint over Solium. Solium seems to be much more widely used, but I'm open to using whichever works best for our needs. Thoughts?

packages/contracts/src/contracts/previous/MultiSigWalletWithTImeLockExceptRemoveAuthorizedAddress/MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress.sol
```diff
@@ -20,8 +20,10 @@ pragma solidity ^0.4.10;
 
 import "../../current/multisig/MultiSigWalletWithTimeLock.sol";
 
+
 contract MultiSigWalletWithTimeLockExceptRemoveAuthorizedAddress is MultiSigWalletWithTimeLock {
 
+// solhint-disable-next-line var-name-mixedcase
```
Recmo: Needs four space indentation.

focuzz:
@LogvinovLeon Sorry for the slow reply, there were state holidays here.

You've asked why Solhint is better than Solium.
Yesterday I asked developer @idrabenia about the comparison, and he provided me with the next link https://github.com/OpenZeppelin/openzeppelin-solidity/issues/508#issuecomment-346569323

+ @idrabenia is also the member of @protofire team like me )

focuzz:
I did rebase, but did small mistake during it. So, now Github proposes to re-merge entire changes since PR created. I will create a new branch and new PR.

Issue #785, opened 2018-06-27
abi-gen shouldn't re-gen when it doesn't have to


feuGeneA:
## Expected Behavior
When an ABI .json hasn't changed, abi-gen shouldn't re-generate its contract wrapper.

## Current Behavior
abi-gen always generates contract wrappers, whether the target wrapper is up to date or not.

## Possible Solution
To check to see whether the generation is needed or not, abi-gen could either look at the .json timestamp, or it could use whatever magic sol-compiler is using to check whether a .sol file has changed (it isn't timestamps!).

## Steps to Reproduce (for bugs)

```
gene@precision5510:~/0x-monorepo/packages/metacoin$ yarn generate_contract_wrappers
yarn run v1.7.0
$ abi-gen --abis 'artifacts/Metacoin.json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers --backend ethers && prettier --write 'src/contract_wrappers/**.ts'
Found 7 partial templates
Found 1 ABI files
Processing: Metacoin...
Created: src/contract_wrappers/metacoin.ts
src/contract_wrappers/metacoin.ts 415ms
Done in 1.05s.
gene@precision5510:~/0x-monorepo/packages/metacoin$ ls -l src/contract_wrappers/metacoin.ts 
-rw-rw-r-- 1 gene gene 20004 Jun 27 14:31 src/contract_wrappers/metacoin.ts
gene@precision5510:~/0x-monorepo/packages/metacoin$ yarn generate_contract_wrappers
yarn run v1.7.0
$ abi-gen --abis 'artifacts/Metacoin.json' --template ../contract_templates/contract.handlebars --partials '../contract_templates/partials/**/*.handlebars' --output src/contract_wrappers --backend ethers && prettier --write 'src/contract_wrappers/**.ts'
Found 7 partial templates
Found 1 ABI files
Processing: Metacoin...
Created: src/contract_wrappers/metacoin.ts
src/contract_wrappers/metacoin.ts 401ms
Done in 1.04s.
gene@precision5510:~/0x-monorepo/packages/metacoin$ ls -l src/contract_wrappers/metacoin.ts 
-rw-rw-r-- 1 gene gene 20004 Jun 27 14:32 src/contract_wrappers/metacoin.ts
```

## Context
[...]

PR #788, opened 2018-06-28
Using timestamps, skip generation of already-up-to-date contract wrappers

feuGeneA:
## Description

Fix for #785 

## Testing instructions

Prepare: checkout 0x-monorepo, `cd packages/metacoin`, `yarn build`.
Then `yarn generate_contract_wrappers`, and ensure that the output says "Already up to date: ..." and not "Created: ...".

Also run new tests of abi-gen via `yarn test` in abi-gen folder.

## Types of changes

*   [X] New feature (non-breaking change which adds functionality)

## Checklist:

*   [X] Labeled this PR with the 'WIP' label if it is a work in progress.


feuGeneA:
Needs testing! (both manual and automated)

packages/abi-gen/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "timestamp": 1530150506,
```
LogvinovLeon: Remove timestamp. Build system adds it automatically

packages/abi-gen/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
...
+        "version": "0.3.3",
+        "changes": [
+            {
+                "note": "skip generation of wrappers that are already up to date"
```
LogvinovLeon: Missing comma at the end

packages/abi-gen/src/index.ts
```diff
@@ -70,16 +70,26 @@ function registerPartials(partialsGlob: string): void {
...
+}
+
+function writeOutputFile(name: string, renderedTsCode: string): void {
+    const filePath = makeOutputFilePath(name);
     fs.writeFileSync(filePath, renderedTsCode);
     logUtils.log(`Created: ${chalk.bold(filePath)}`);
 }
 
+function isOutFileUpToDate(abiFile: string, outFile: string): boolean {
+    const abiTime = fs.statSync(abiFile).mtimeMs;
```
LogvinovLeon: I like that. Maybe we can go even further and hash the ABI? That way if someone removes all the artifacts and recompiles the same contracts we still do no work.

feuGeneA: I do like that idea, and I saw that [that's how sol-compiler works](https://github.com/0xProject/0x-monorepo/blob/988a3347582cd812aadeaa0cdaef67389c8d833d/packages/sol-compiler/src/compiler.ts#L131).

But then the question is: where would we store the hash of the ABI?

In sol-compiler's case, the hash of the source (and it's dependencies! interestingly) is stored in the ABI output (as field `sourceTreeHashHex`).

In abi-gen's case, there is no existing, natural place to keep that hash.  Perhaps we could store it in a comment in the contract wrapper?  Or perhaps in some side file?  Those seemed convoluted / overly complicated to me, so I opted for the timestamp method.  And, personally, if I wipe out the ABI files and generate them anew, that means that I want to force regeneration of the wrappers too, though admittedly I'd only do that when I'm questioning my own sanity. :)

The more I think about it, the more I like the idea of putting the ABI hash in a comment in the generated wrapper.  But that would be more work, and if that's the route you'd prefer then I would advise we push through the timestamp changes and then have a separate piece of work to replace the timestamp checking with hashing.  In any case, you tell me what you want and I'll make it happen. :)

feuGeneA:
All tested!

@LogvinovLeon your ABI hash change request got lost in the shuffle as I refactored to prepare for unit tests.  Please review [my previous comment](https://github.com/0xProject/0x-monorepo/pull/788#discussion_r198871037) and let me know your disposition with regard to timestamp vs. hashing.  If you're okay with timestamps (at least for the time being) then I'm calling this PR done.

LogvinovLeon:
Timestamps vs hashes. I was thinking about putting it in a comment, but I get your arguments and I think timestamps are also a good solution. Maybe even better.

One more thing before merging. All our package that have tests report test coverage to coveralls. 
Please look how it's done in other places and add it here. After it LGTM and ready to merge.
You'll need to add some entries here: https://github.com/0xProject/0x-monorepo/blob/v2-prototype/.circleci/config.yml#L82
And add a `test:coverage` command. Should be easy.
https://github.com/0xProject/0x-monorepo/blob/v2-prototype/packages/assert/package.json#L21

packages/abi-gen/package.json
```diff
@@ -29,12 +31,19 @@
     "dependencies": {
         "@0xproject/typescript-typings": "^0.4.1",
         "@0xproject/utils": "^0.7.1",
-        "ethereum-types": "^0.0.1",
+        "@types/sleep": "^0.0.7",
+        "@types/tmp": "^0.0.33",
```
fabioberger: Types that aren't part of the exported, public interface, they should be `devDependencies`. Please move them there.

packages/abi-gen/package.json
```diff
@@ -29,12 +31,19 @@
...
+        "chai": "^4.1.2",
         "chalk": "^2.3.0",
+        "dirty-chai": "^2.0.1",
```
fabioberger: Ditto with `chai` and `dirty-chai`, testing deps should be listed under `devDependencies`. This is because end-users of this library do not need these deps to use the package. Listing under `devDependencies` means these deps don't get installed for end-users.

packages/abi-gen/package.json
```diff
@@ -29,12 +31,19 @@
...
+        "ethereum-types": "^0.0.1",
         "glob": "^7.1.2",
         "handlebars": "^4.0.11",
         "lodash": "^4.17.4",
         "mkdirp": "^0.5.1",
+        "mocha": "^5.2.0",
```
fabioberger: Ditto with `mocha`

packages/abi-gen/src/index.ts
```diff
@@ -118,6 +105,13 @@ for (const abiFileName of abiFileNames) {
         process.exit(1);
     }
 
+    const outFilePath = `${args.output}/${utils.makeOutputFileName(namedContent.name)}.ts`;
```
fabioberger: Let's break this into two lines for readability: 

```
const fileName = utils.makeOutputFileName(namedContent.name);
const outFilePath = `${args.output}/${fileName}.ts`;
```

packages/abi-gen/src/utils.ts
```diff
@@ -92,4 +93,24 @@ export const utils = {
...
+        if (fileName === 'z_r_x_token') {
+            fileName = 'zrx_token';
+        }
+        return fileName;
+    },
+    writeOutputFile(filePath: string, renderedTsCode: string): void {
+        fs.writeFileSync(filePath, renderedTsCode);
+    },
+    isOutputFileUpToDate(abiFile: string, outputFile: string): boolean {
+        const abiTime = fs.statSync(abiFile).mtimeMs;
```
fabioberger: The variable name should elucidate the units of time used. Sec? Ms? Hours? It is clear on this line, but not further down where the variable is reused. E.g `abiFileModifiedTimeMs`

packages/abi-gen/src/utils.ts
```diff
@@ -92,4 +93,24 @@ export const utils = {
...
+        try {
+            const outputFileTime = fs.statSync(outputFile).mtimeMs;
```
fabioberger: `outputFileModifiedTimeMs`

packages/abi-gen/src/utils.ts
```diff
@@ -92,4 +93,24 @@ export const utils = {
...
+            return outputFileTime > abiTime;
```
fabioberger: Should be `greater or equal` no? It's not going to impact the code, since these will never be equal, but logically it is more correct to say `>=`.

packages/abi-gen/src/utils.ts
```diff
@@ -92,4 +93,24 @@ export const utils = {
...
+        } catch (e) {
```
fabioberger: Our convention is to use variable name `err` for errors. `catch (err) {`

fabioberger: Let's explicitly check for the `ENOENT` error here and return `false` if it is thrown. For all other errors, we should re-throw it. E.g `Permission denied` errors (see all: https://nodejs.org/api/errors.html#errors_common_system_errors)

packages/abi-gen/src/utils.ts
feuGeneA: I modeled this after my knowledge of how GNU make works.  I was just going off of instinct, but I looked it up just now, and make works the way I wrote it here.  From the GNU make manual: "The make utility examines time relationships and shall update those derived files (called targets) that have modified times *earlier than* the modified times of the files (called prerequisites) from which they are derived."  (emphasis mine)

Does that change your mind?  (I'm not advocating for one way or the other, just explaining my rationale; happy to defer to your judgement.)

packages/abi-gen/package.json
feuGeneA: Really appreciate your explanation around this.

packages/abi-gen/src/index.ts
feuGeneA: Thanks for this suggestion.  The length of this line irked me too.  I originally wrote it as
```
const outFilePath =
    `${args.output}/${utils.makeOutputFileName(namedContent.name)}.ts`;
```
but `prettier:ci` wouldn't let me keep it that way.

feuGeneA:
Regarding that conflict, `ethereum-types` got moved by my `yarn add`s because it wasn't previously in alphabetical order.

feuGeneA:
@fabioberger I believe I've made all of the changes you requested.

feuGeneA:
@LogvinovLeon I believe I've got all the coverage stuff set up right (works for me locally), but the check is failing with `Couldn't find a repository matching this job`, which I think is a result of the PR being sourced from my fork rather than from a branch in the repo.  I was just complaining to @dekz about this in the rocket chat; pretty annoying, would love to figure out how to make that work...

packages/abi-gen/src/utils.ts
fabioberger: Thanks for the explanation! Now that I think about it again, I actually think your solution is more correct. If they both have the same modified timestamp, there would have been no time for the output to be adjusted given the new ABI file (hinting at them being modified independently). Thus, the output would be out of date and would need to be re-generated given the new ABI. Let's keep it how it is.

feuGeneA:
@LogvinovLeon I don't think there's much hope of getting the coverage check working here any time soon, per #823 .  However, I got it to [run successfully in my own CircleCI](https://circleci.com/gh/feuGeneA/0x-monorepo/18) instance*, so in theory everything is just fine. :)

\* You'll see I ran the job against a `dummy` branch, trying to reduce the noise in this PR, but [that branch is identical to the one on which this PR is based.](https://github.com/feuGeneA/0x-monorepo/compare/abi-gen-ignore-unchanged...feuGeneA:dummy?expand=1)

feuGeneA:
I see some conflicts crept in, let me sort that out.

feuGeneA:
All conflicts resolved and all checks passed!!

@LogvinovLeon coverage working as intended! :)

packages/abi-gen/src/index.ts
```diff
@@ -120,6 +107,14 @@ for (const abiFileName of abiFileNames) {
         process.exit(1);
     }
 
+    const outFileName = utils.makeOutputFileName(namedContent.name);
+    const outFilePath = `${args.output}/${outFileName}.ts`;
+
+    if (utils.isOutputFileUpToDate(abiFileName, outFilePath)) {
+        logUtils.log('Already up to date: ' + outFilePath);
```
fabioberger: Nit: Please use `template strings` over concatenation everywhere. This would be:

```
logUtils.log(`Already up to date: ${outFilePath}`);
```

fabioberger: Once this get's fixed, we will merge this :)

feuGeneA:
@LogvinovLeon I was just doing some additional testing, and I discovered that [your recent commit titled `Remove generated contract directory before writing new contracts...`](https://github.com/0xProject/0x-monorepo/commit/eafdd4d42d1cab2e579ca8ed1423d285acf78850) is precluding the utility of this PR.   Why was that change made?  Can I revert that aspect of your commit?  (There was another aspect too, about converting `e_r_c` filenames to `erc`, which of course I would leave in place.)

feuGeneA:
@LogvinovLeon I went ahead and got rid of the `rimraf` stuff, since that's the only way this PR still makes any sense.  If you object, let me know how you want to proceed.

I also added a test case to cover the `e_r_c`->`erc` conversion :)

packages/abi-gen/src/index.ts
feuGeneA: All fixed!  But before merging do check with @LogvinovLeon on his opinion on the other change I made at the same time (explained in comments below).

LogvinovLeon:
I've done that because I didn't want the old files to be left in the directory.
Imagine I have contracts A and B. And then remove B. Generated A wrapper will never be removed.

LogvinovLeon:
I know it's not a proper stage to ask that question, but why are you not OK woth re-generating contract_wrappers? Perf reasons? For us it's pretty quick.

feuGeneA:
@LogvinovLeon I understand your logic, but I respectfully disagree with the approach.

Consider the TypeScript scenario.  If you remove a `.ts` file from your project directory, `tsc` won't go ahead and clean up the corresponding `.js` output files.  (And it certainly won't delete the whole output folder.)

C++ is another scenario that I know well.  `.cpp` files are compiled into `.o` files, but if you remove a `.cpp` file then the corresponding `.o` file will indeed never be removed, until you do it manually.

It's a developer's responsibility to understand the outputs of their tool chain and to manually remove those outputs when their corresponding sources have been removed.

In short, yes I want this for performance reasons.  I gave a subjective rationale in the Context section of #785 , which I think all on its own is enough to justify these changes.  But I'm glad you asked this question because it drove me to take some objective measurements of the performance impact.

I gauged performance based on the re-generation of wrappers in `0x-metarepo/packages/contracts`, since that has more ABI inputs than any of the other packages.  For each of \[the v2-prototype branch, my PR branch], I did a successful `yarn build` of the package, followed by many subsequent runs of `time yarn generate_contract_wrappers`.  Under v2-prototype, the re-generation varied between 0.44 and 0.46 seconds.  Under my PR branch, the re-generation consistently took 0.33 seconds.

I know that may not sound like a lot, but really I think it is.  On a percentage basis it's 27%.  And imagine a project that has hundreds of ABI's instead of the 25 that the contracts package has.  That time could really add up.

PR #791, opened 2018-06-28
Fix NotFound component and render 0x.js docs for /docs

BMillman19:
## Description:

* Fix buggy `NotFound` component
* Create a new route `/docs` that goes to the 0x.js docs
* How come the type checking didn't catch this? Does it not work for jsx?

## Testing Instructions:

* Try going to `/docs` verify that 0x.js docs pops up
* Try going to `/hello` or some other unregistered route, verify a 404 page shows up

## Checklist:

### DIFF
--- a/packages/website/ts/pages/not_found.tsx
+++ b/packages/website/ts/pages/not_found.tsx
@@ -11,15 +11,15 @@ export interface NotFoundProps {
     dispatcher: Dispatcher;
 }
 
-export const NotFound = (_props: NotFoundProps) => {
+export const NotFound = (props: NotFoundProps) => {
     return (
         <div>
-            <TopBar blockchainIsLoaded={false} location={this.props.location} translate={this.props.translate} />
+            <TopBar blockchainIsLoaded={false} location={props.location} translate={props.translate} />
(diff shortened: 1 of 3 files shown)


packages/website/ts/pages/not_found.tsx
```diff
@@ -11,15 +11,15 @@ export interface NotFoundProps {
     dispatcher: Dispatcher;
 }
 
-export const NotFound = (_props: NotFoundProps) => {
+export const NotFound = (props: NotFoundProps) => {
```
fragosti: What type does Visual Studio Code give this? I feel like it's just typed as (props: NotFoundProps) => React.ReactNode instead of React.StatelessComponent<NotFoundProps>.

Also (props: NotFoundProps) is technically incorrect because of the implied children props that is always present.

I still find it weird that it didn't catch this. Wonder if the explicit typing will catch it (I feel like it won't, I've run into runtime errors where I've done this by accident).

fragosti: Did you try React.ReactStatelessComponent?

PR #794, opened 2018-06-28
Improve reliability of the minimum block number hack for devnet/geth

albrow:
## Description

There's a weird bug we've known about for a long time having to do with the devnet/Geth. If you try to deploy a contract, it will fail if there are less than 3 blocks on the chain. To address this, we introduced a hack in the __run.sh__ file, which is run when the devnet is initialized. The hack was to simply send 3 transactions with a value of 0, effectively increasing the length of the chain by 3 without really changing anything else.

However, @fabioberger reported a bug where Geth was sometimes resetting the block number all the way back to 0 (and resetting any corresponding state). This meant that the next time you tried to deploy a contract after this happened, it would fail.

This PR improves the reliability of the minimum block hack by moving it out of the __run.sh__ file and into __blockchain_lifecycle.ts__. The advantage is twofold:

1. It's faster because we have the ability to use `awaitTransactionMinedAsync` and wait exactly the right amount of time before sending the next transaction. In __run.sh__ we had to resort for sleeping for 3 seconds between each transaction.
2. It's more reliable because we can check the block number every time we call `startAsync()`. This means that even if Geth resets back to the genesis block, we can re-mine the minimum number of blocks and our tests will still work.

## Testing instructions

The bug reported by @fabioberger seems to be fairly non-deterministic and (may be due to a race condition in Geth). You can *usually* reproduce it by deploying extra dummy ERC20 tokens. For example, you can make this change in __contracts/test/asset_proxy/proxies.ts__:

```diff
[zrxToken] = await erc20Wrapper.deployDummyTokensAsync(
-    constants.NUM_DUMMY_ERC20_TOKEN_TO_DEPLOY,
+    constants.NUM_DUMMY_ERC20_TOKEN_TO_DEPLOY + 10,
    constants.DUMMY_TOKEN_DECIMALS,
);
```

Without the changes in this PR, the tests will fail at some point when trying to deploy a contract. With the changes in this PR applied, the tests should work (but you might see a warning about the minimum number of blocks not being met).

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,9 @@ jobs:
       - restore_cache:
           keys:
             - repo-{{ .Environment.CIRCLE_SHA1 }}
-      # HACK(albrow): we need to sleep 15 seconds to ensure the devnet is
+      # HACK(albrow): we need to sleep 10 seconds to ensure the devnet is
       # initialized
-      - run: sleep 15 && TEST_PROVIDER=geth yarn wsrun test contracts
+      - run: sleep 10 && TEST_PROVIDER=geth yarn wsrun test contracts
   test-rest:
     docker:
       - image: circleci/node:9
--- a/packages/dev-utils/package.json
+++ b/packages/dev-utils/package.json
@@ -47,6 +47,7 @@
     "dependencies": {
         "@0xproject/subproviders": "^0.10.4",
         "@0xproject/types": "^0.8.1",
+        "@0xproject/utils": "^0.7.1",
         "ethereum-types": "^0.0.1",
         "@0xproject/typescript-typings": "^0.4.1",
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -53,9 +53,9 @@ jobs:
       - restore_cache:
           keys:
             - repo-{{ .Environment.CIRCLE_SHA1 }}
-      # HACK(albrow): we need to sleep 15 seconds to ensure the devnet is
+      # HACK(albrow): we need to sleep 10 seconds to ensure the devnet is
```
albrow: We can reduce this wait time now because we are no longer waiting for transactions to be mined in __run.sh__.

PR #796, opened 2018-06-29
Update expectRevertReasonOrAlwaysFailingTransactionAsync to check status codes

albrow:
## Description

This PR fixes the assertions to work for an edge case on Geth: `sendTrasnaction` succeeds but the txReceipt has a status code of 0, indicating a failure.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/utils/assertions.ts
+++ b/packages/contracts/src/utils/assertions.ts
@@ -1,8 +1,10 @@
 import { RevertReason } from '@0xproject/types';
 import * as chai from 'chai';
+import { TransactionReceipt, TransactionReceiptWithDecodedLogs } from 'ethereum-types';
 import * as _ from 'lodash';
 
 import { constants } from './constants';
+import { web3Wrapper } from './web3_wrapper';
 
 const expect = chai.expect;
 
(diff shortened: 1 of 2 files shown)


packages/contracts/test/exchange/match_orders.ts
```diff
@@ -651,13 +651,7 @@ describe('matchOrders', () => {
...
             // Match orders
             return expectRevertReasonOrAlwaysFailingTransactionAsync(
-                matchOrderTester.matchOrdersAndVerifyBalancesAsync(
-                    signedOrderLeft,
-                    signedOrderRight,
-                    takerAddress,
-                    erc20BalancesByOwner,
-                    erc721TokenIdsByOwner,
-                ),
+                exchangeWrapper.matchOrdersAsync(signedOrderLeft, signedOrderRight, takerAddress),
```
albrow: I had to update this because the new version of `expectRevertReasonOrAlwaysFailingTransactionAsync` is more strict about the types of promises it accepts.

packages/contracts/src/utils/assertions.ts
```diff
@@ -53,18 +55,45 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
+ * 3) The given Promise fulfills with a txReceipt that has a status of 0 or '0', indicating the transaction failed.
+ * 4) The given Promise fulfills with a txHash and corresponding txReceipt has a status of 0 or '0'.
  * @param p the Promise which is expected to reject
  * @param reason a specific revert reason
  * @returns a new Promise which will reject if the conditions are not met and
  * otherwise resolve with no value.
  */
-export function expectRevertReasonOrAlwaysFailingTransactionAsync<T>(
-    p: Promise<T>,
+export async function expectRevertReasonOrAlwaysFailingTransactionAsync(
```
albrow: This function is getting pretty crazy in terms of the types of inputs it accepts and this is not how I would prefer to write it in a vacuum. However, writing it this way requires the least number of changes in our test code. I may improve this in a future PR by making the inputs more strict and updating our test code.

packages/contracts/src/utils/assertions.ts
```diff
@@ -53,18 +55,45 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
+    p: Promise<string | TransactionReceiptWithDecodedLogs | TransactionReceipt>,
     reason: RevertReason,
-): PromiseLike<void> {
-    return _expectEitherErrorAsync(p, 'always failing transaction', reason);
+): Promise<void> {
+    return p
+        .then(async result => {
```
abandeali1: Can we maybe split this out into another function? This change makes it so that tests can pass and we _think_ that the revert reason is being checked when it actually isn't.

abandeali1: Discussed offline - this is a stopgap for now, but we will add different checks for geth and ganache in a separate PR.

packages/contracts/src/utils/assertions.ts
```diff
@@ -53,18 +55,45 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
+            let txReceiptStatus: string | 0 | 1 | null;
+            if (typeof result === 'string') {
```
fabioberger: Please use `_.isString` instead of `typeOf`

packages/contracts/src/utils/assertions.ts
```diff
@@ -53,18 +55,45 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
-    p: Promise<T>,
+export async function expectRevertReasonOrAlwaysFailingTransactionAsync(
+    p: Promise<string | TransactionReceiptWithDecodedLogs | TransactionReceipt>,
     reason: RevertReason,
-): PromiseLike<void> {
-    return _expectEitherErrorAsync(p, 'always failing transaction', reason);
+): Promise<void> {
+    return p
+        .then(async result => {
+            let txReceiptStatus: string | 0 | 1 | null;
```
fabioberger: Do we not have a type for this?

packages/contracts/src/utils/assertions.ts
```diff
@@ -53,18 +55,45 @@ export function expectRevertOrAlwaysFailingTransactionAsync<T>(p: Promise<T>): P
...
+            if (typeof result === 'string') {
+                // Result is a txHash. We need to make a web3 call to get the receipt.
+                const txReceipt = await web3Wrapper.getTransactionReceiptAsync(result);
+                txReceiptStatus = txReceipt.status;
+            } else if ('status' in result) {
+                // Result is a TransactionReceiptWithDecodedLogs or TransactionReceipt
+                // and status is a field of result.
+                txReceiptStatus = result.status;
+            } else {
+                throw new Error('Unexpected result type');
```
fabioberger: Add the result type to the error string for easier debuggability.

packages/contracts/test/exchange/match_orders.ts
fabioberger: Does it still verify the balances?

PR #802, opened 2018-07-02
FillOrder Combinatorial Testing Fixes

fabioberger:
## Description

This PR fixes some issues found while running 20,000 orderFill scenarios.

- Added ability to permute maker/taker allowance/balance scenarios
- Fixed order of validation checks in TS implementation
- Transfer away ERC721 instead of burning
- Fix bug where we were using takerAddress from order (which could be nullAddress) instead of actual takerAddress.

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/src/utils/asset_wrapper.ts
+++ b/packages/contracts/src/utils/asset_wrapper.ts
@@ -85,8 +85,7 @@ export class AssetWrapper {
                         userAddress,
                     );
                 } else if (tokenOwner === userAddress && desiredBalance.eq(0)) {
-                    // Burn token
-                    // await erc721Wrapper.burnAsync(assetProxyData.tokenAddress, assetProxyData.tokenId, userAddress);
+                    // Transfer token to someone else
                     const userAddresses = await (erc721Wrapper as any)._web3Wrapper.getAvailableAddressesAsync();
                     const nonOwner = _.find(userAddresses, a => a !== userAddress);
                     await erc721Wrapper.transferFromAsync(
--- a/packages/contracts/src/utils/asset_wrapper.ts
+++ b/packages/contracts/src/utils/asset_wrapper.ts
@@ -86,7 +86,15 @@ export class AssetWrapper {
                     );
                 } else if (tokenOwner === userAddress && desiredBalance.eq(0)) {
                     // Burn token
-                    await erc721Wrapper.burnAsync(assetProxyData.tokenAddress, assetProxyData.tokenId, userAddress);
+                    // await erc721Wrapper.burnAsync(assetProxyData.tokenAddress, assetProxyData.tokenId, userAddress);
+                    const userAddresses = await (erc721Wrapper as any)._web3Wrapper.getAvailableAddressesAsync();
+                    const nonOwner = _.find(userAddresses, a => a !== userAddress);
+                    await erc721Wrapper.transferFromAsync(
+                        assetProxyData.tokenAddress,
+                        assetProxyData.tokenId,
+                        tokenOwner,
+                        nonOwner,
+                    );
                     return;
                 } else if (
                     (userAddress !== tokenOwner && desiredBalance.eq(0)) ||
--- a/packages/contracts/src/utils/core_combinatorial_utils.ts
+++ b/packages/contracts/src/utils/core_combinatorial_utils.ts
@@ -783,25 +783,17 @@ export class CoreCombinatorialUtils {
 
             case AllowanceAmountScenario.TooLow:
                 const tooLowAllowance = takerFee.minus(1);
-                await this.assetWrapper.setProxyAllowanceAsync(
-                    signedOrder.takerAddress,
-                    this.zrxAssetData,
-                    tooLowAllowance,
-                );
+                await this.assetWrapper.setProxyAllowanceAsync(this.takerAddress, this.zrxAssetData, tooLowAllowance);
                 break;
 
             case AllowanceAmountScenario.Exact:
                 const exactAllowance = takerFee;
-                await this.assetWrapper.setProxyAllowanceAsync(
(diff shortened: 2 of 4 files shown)


packages/contracts/src/utils/asset_wrapper.ts
```diff
@@ -85,8 +85,15 @@ export class AssetWrapper {
                         userAddress,
                     );
                 } else if (tokenOwner === userAddress && desiredBalance.eq(0)) {
-                    // Burn token
-                    await erc721Wrapper.burnAsync(assetProxyData.tokenAddress, assetProxyData.tokenId, userAddress);
+                    // Transfer token to someone else
+                    const userAddresses = await (erc721Wrapper as any)._web3Wrapper.getAvailableAddressesAsync();
+                    const nonOwner = _.find(userAddresses, a => a !== userAddress);
+                    await erc721Wrapper.transferFromAsync(
```
LogvinovLeon: What's wrong with burning? Seems more clear to me.

packages/contracts/src/utils/core_combinatorial_utils.ts
```diff
@@ -145,13 +145,39 @@ export class CoreCombinatorialUtils {
...
     public assetWrapper: AssetWrapper;
     public static generateFillOrderCombinations(): FillScenario[] {
-        const takerScenarios = [TakerScenario.Unspecified];
-        const feeRecipientScenarios = [FeeRecipientAddressScenario.EthUserAddress];
-        const makerAssetAmountScenario = [OrderAssetAmountScenario.Large];
-        const takerAssetAmountScenario = [OrderAssetAmountScenario.Large];
-        const makerFeeScenario = [OrderAssetAmountScenario.Large];
-        const takerFeeScenario = [OrderAssetAmountScenario.Large];
-        const expirationTimeSecondsScenario = [ExpirationTimeSecondsScenario.InFuture];
+        const takerScenarios = [
```
LogvinovLeon: Maybe we should introduce a config variable instead of commenting out the code?

packages/contracts/src/utils/core_combinatorial_utils.ts
```diff
@@ -193,16 +275,16 @@ export class CoreCombinatorialUtils {
                 },
                 takerAssetFillAmountScenario: fillScenarioArray[9] as TakerAssetFillAmountScenario,
                 makerStateScenario: {
-                    traderAssetBalance: BalanceAmountScenario.Higher,
-                    traderAssetAllowance: AllowanceAmountScenario.Higher,
-                    zrxFeeBalance: BalanceAmountScenario.Higher,
-                    zrxFeeAllowance: AllowanceAmountScenario.Higher,
+                    traderAssetBalance: fillScenarioArray[10] as BalanceAmountScenario,
```
LogvinovLeon: What does `no-magic-numbers` have to say about it?

packages/contracts/test/exchange/fill_order.ts
```diff
@@ -67,13 +67,7 @@ describe('FillOrder Tests', () => {
...
             _.forEach(fillScenarios, fillScenario => {
                 const orderScenario = fillScenario.orderScenario;
-                const description = `Combinatorial OrderFill: ${orderScenario.feeRecipientScenario} ${
-                    orderScenario.makerAssetAmountScenario
-                } ${orderScenario.takerAssetAmountScenario} ${orderScenario.makerFeeScenario} ${
-                    orderScenario.takerFeeScenario
-                } ${orderScenario.expirationTimeSecondsScenario} ${orderScenario.makerAssetDataScenario} ${
-                    orderScenario.takerAssetDataScenario
-                }`;
+                const description = `Combinatorial OrderFill: ${JSON.stringify(fillScenario)}`;
```
LogvinovLeon: JSON.stringify(obj, null, 2)

packages/contracts/src/utils/asset_wrapper.ts
fabioberger: One cannot set an allowance for a non-existent token. One can however set an allowance to a token owned by someone else. In the combinatorial tests, we permute through all kinds of maker/taker allowance/balance scenarios and in some, the maker has 0 balance, but allowance is set. This change is necessary to support these permutations.

packages/contracts/src/utils/core_combinatorial_utils.ts
fabioberger: The point of the commented out code here is to elucidate the options that are not being tested, and to make it very easy to test them locally at will. I do see your point that this could become out-of-date, but it is still greppable, so I think we should take the risk.

packages/contracts/src/utils/core_combinatorial_utils.ts
fabioberger: Ah, it seems that `no-magic-numbers` is disables for `contracts` package... We should enable it in a separate PR.

packages/contracts/test/exchange/fill_order.ts
fabioberger: Since this is a test description, I purposefully left it more compact so that it doesn't take up tons of space. I'm willing to go back on this decision if you feel strongly.

packages/contracts/src/utils/core_combinatorial_utils.ts
LogvinovLeon: Agree. But I still think that we shouldn't have magic numbers in this PR. Why do we have that as an array? Why not object with properties?

fabioberger: Agreed, will add a TSLint-disable. The `_getAllCombinations` method is used to be agnostic to the scenarios passed into it. This makes it more easily reusable (Alex is going to re-use it for combinatorially testing inner-methods).

PR #707, opened 2018-06-14
OrderFill Declarative Testing & Combinatorial Tests [Merge after #760]

fabioberger:
## Description

This PR introduces a more declarative approach to writing fillOrder tests. Instead of manually:

- Setting up a particular scenario
- Calling fillOrder
- verifying that the outcomes of the fill match expectations (filledAmount, balances/allowances, event logs, etc...)

The test writer must simply defines a fillOrder scenario he/she would like to test. That is it. The testing utils will then setup the scenario, predict the outcome of the fill, call fillOrder and verify the the fill succeeded/failed as expected. The fill prediction code is entirely separate from the smart contract logic, providing a helpful secondary implementation of the 0x protocol logic that further sanity checks the behavior of the smart contracts.

Example fillScenario declaration:

```
const fillScenario = {
    orderScenario: {
        takerScenario: TakerScenario.Unspecified,
        feeRecipientScenario: FeeRecipientAddressScenario.EthUserAddress,
        makerAssetAmountScenario: OrderAssetAmountScenario.Large,
        takerAssetAmountScenario: OrderAssetAmountScenario.Large,
        makerFeeScenario: OrderAssetAmountScenario.Small,
        takerFeeScenario: OrderAssetAmountScenario.Zero,
        expirationTimeSecondsScenario: ExpirationTimeSecondsScenario.InFuture,
        makerAssetDataScenario: AssetDataScenario.ERC20NonZRXEighteenDecimals,
        takerAssetDataScenario: AssetDataScenario.ERC721,
    },
    takerAssetFillAmountScenario: TakerAssetFillAmountScenario.ExactRemainingFillableTakerAssetAmount,
    makerStateScenario: {
        traderAssetBalance: TokenAmountScenario.Higher,
        traderAssetAllowance: TokenAmountScenario.TooLow,
        zrxFeeBalance: TokenAmountScenario.Exact,
        zrxFeeAllowance: TokenAmountScenario.Higher,
    },
    takerStateScenario: {
        traderAssetBalance: TokenAmountScenario.Higher,
        traderAssetAllowance: TokenAmountScenario.Exact,
        zrxFeeBalance: TokenAmountScenario.Higher,
        zrxFeeAllowance: TokenAmountScenario.Exact,
    },
};
```

[...]

PR #807, opened 2018-07-02
Add combinatorial tests for internal Exchange functions

albrow:
## Description

This PR lays the foundation for combinatorial tests for the internal Exchange functions. I would like to discuss whether we think these kinds of tests are worthwhile before continuing.

The set of values I'm using for combinatorial testing are based on @Recmo's [comment](https://github.com/0xProject/0x-monorepo/pull/707#pullrequestreview-129357978) on #707. However, I had to considerably reduce the set of values we use because the number of test cases ended up being far too large (and caused Node to run out of memory).

Right away I found that one of the generated test cases fails. Here's the output:

```
 |   1) Exchange core internal
 |        updateFilledState
 |          generated test case 28:
 |      Error: VM Exception while processing transaction: invalid opcode
 | 	Test case: ["0","115792089237316195423570985008687907853269984665640564039457584007913129639935","0","0"]
 |       at Object.<anonymous> packages/contracts/(test/exchange/core_internal.ts:129:27)
 |       at step packages/contracts/(lib/test/exchange/core_internal.js:40:23)
 |       at Object.throw packages/contracts/(lib/test/exchange/core_internal.js:21:53)
 |       at rejected packages/contracts/(lib/test/exchange/core_internal.js:13:65)
 |       at <anonymous>
 |       at process._tickCallback packages/contracts/(internal/process/next_tick.js:188:7)
```

`invalid opcode` is a pretty scary-sounding error since all we're doing is changing the numbers passed in as arguments to a function. Is this a bug in our contract code or an anticipated limitation of using big numbers in general?

For context, there are currently 2,401 test cases for the `updateFilledState` function and each test case takes 25-50ms to run. Speaking conservatively, it would take about 10 minutes to run through all the test cases for this one function.

If we can sort out what's going on with the current test failure, here are some other open questions for discussion. In general, I'm wondering whether these kinds of combinatorial tests are worthwhile and whether we should just write more traditional unit tests for internal functions instead.

1. I estimate that we if we extrapolate this testing strategy to all internal functions, the tests would take up to an hour or more. This is far beyond what is reasonable for CI. Is it even worth running these tests manually after big changes or every month or so? Alternatively, is it worth spending time optimizing the tests to run faster?
2. The current testing strategy I'm using resets the state after each test case. This means we will miss edge cases that have to do with leftover state. Is there a good, general-purpose strategy we can use for writing combinatorial tests that retain state? How will it affect our test run time?
3. I'm currently only generating test cases for combinations of numeric values. Is it worth testing other types of inputs (e.g. addresses?). How will it affect our test run time?
4. I'm currently only testing that there are no reverts and the transaction is successful. The tests do not check the return value or verify any updates to the state. Do we need to do those things? Is it possible to do it in a general-purpose way? How will it affect our test run time?

## Testing instructions

Just run `PKG=contracts yarn test`.

## Types of changes

## Checklist:


packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
+import { BigNumber } from '@0xproject/utils';
+import * as _ from 'lodash';
+
+export const BigNumbers = [];
+
+const MAX_256_BIT_NUMBER = new BigNumber(2).pow(256).minus(1);
+
+export const positiveNaturalBigNumbers = [
```
albrow: These values are the basis for generating the inputs that make up the different test cases.

packages/contracts/test/exchange/core_internal.ts
```diff
@@ -0,0 +1,134 @@
...
+    describe.only('updateFilledState', async () => {
+        const testCases = combinatorics.cartesianProduct(
+            positiveNaturalBigNumbers,
+            positiveNaturalBigNumbers,
+            positiveNaturalBigNumbers,
+            positiveNaturalBigNumbers,
+        );
+        console.log(`Generated ${testCases.length} combinatoric test cases.`);
+        let counter = -1;
+        testCases.forEach(async testCase => {
```
albrow: Each test case consists of four numeric values made up of different combinations of the numbers in `positiveNaturalBigNumbers`. For each case, we simply pass the values through to `sendTransactionAsync` and make sure the transaction is successful.

packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
...
+    new BigNumber(0),
+    new BigNumber(1),
+    new BigNumber(2),
+    new BigNumber(2).pow(64),
+    MAX_256_BIT_NUMBER,
+    MAX_256_BIT_NUMBER.div(2).floor(),
+    MAX_256_BIT_NUMBER.sqrt().floor(),
```
Recmo: This evaluated to `2**128 - 1`, which does *not* cause overflow when multiplied by itself.

I would avoid using the `div` and `sqrt` function and instead specify the numbers as directly as possible (either as `new BigNumber(2).pow(255).sub(1)` or as `0x7FFFFFFFFFFFFFF...`.

As it stands it is not clear if you intended to be exactly on or just below the threshold of overflow. A comment can also clarify this.

packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const MAX_256_BIT_NUMBER = new BigNumber(2).pow(256).minus(1);
+
+export const positiveNaturalBigNumbers = [
+    new BigNumber(0),
+    new BigNumber(1),
+    new BigNumber(2),
+    new BigNumber(2).pow(64),
+    MAX_256_BIT_NUMBER,
+    MAX_256_BIT_NUMBER.div(2).floor(),
```
Recmo: This evaluates to `2**255-1`. This is the largest number that does *not* cause overflow when added to itself.

Is this what you intended?

packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
...
+    MAX_256_BIT_NUMBER.sqrt().floor(),
+];
+
+export const negativeNaturalNumbers = _.map(positiveNaturalBigNumbers, n => n.neg());
```
Recmo: Negative numbers in Solidity use twos-complement, which means their values range from `-2**128 ... 2**128-1`. Note that in twos complement there is always one more negative number than positive (not counting zero).

Values like `MAX_256_BIT_NUMBER` and `-MAX_256_BIT_NUMBER` do not exist in singed integer.

It's best not to think of them as "natural numbers". Such a thing does not exist in Solidity. In Solidity there are several kinds of numbers: `uint256`, `uint248` .. `uint8`, `sint256`, ... `sint8`. All of them behave according to their own set of rules and should have their own specific tests sets. Similarly other data types such as `bool`, `address`,`byte1` ... `byte32`, `bytes`, `string` should also have their own test sets. Of course we only need to implement the ones that we actually use.

packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
...
+export const naturalBigNumbers = _.concat(positiveNaturalBigNumbers, negativeNaturalNumbers);
+
+function bigNumbersToNumbers(bigNumbers: BigNumber[]): number[] {
+    return _.map(bigNumbers, (x: BigNumber) => x.toNumber());
```
Recmo: This function it looses precision and we need it to be exact. The `toNumber()` should only be used for `console.log` purposes, and that should not be done in this file.

packages/contracts/test/utils/combinatorial_sets.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+export const positiveNaturalNumbers = bigNumbersToNumbers(positiveNaturalBigNumbers);
+export const naturalNumbers = bigNumbersToNumbers(naturalBigNumbers);
```
Recmo: These can be confused with the `BigNumber` versions, but they are subtly incorrect. It's best to leave them out of this file.

packages/contracts/test/utils/combinatorial_sets.ts
Recmo: If we want to be able to nicely display the test sets, maybe we can turn them into objects:

```
export const uin256 {
    '2²⁵⁶ - 1': new BigNumber(2).pow(256).sub(1),
    '2²⁵⁵':     new BigNumber(2).pow(255),
    ...
}
```

packages/contracts/test/utils/combinatorial_sets.ts
albrow: 👍I'm going to change our approach here to use a set of inputs for each type.

packages/contracts/test/utils/combinatorial_sets.ts
albrow: 👍I'm going to update this to use both the *maximum* value that *should not* overflow and the *minimum* value that *should* overflow under certain conditions. We can then use a reference function to differentiate between cases where we expect or do not expect an overflow.

packages/utils/src/configured_bignumber.ts
```diff
@@ -4,6 +4,7 @@ import { BigNumber } from 'bignumber.js';
 // more then 20 digits. We want to avoid this behavior, so we set EXPONENTIAL_AT to a high number
 BigNumber.config({
     EXPONENTIAL_AT: 1000,
+    DECIMAL_PLACES: 100,
```
albrow: I was surprised to find that `bignumber.js` *does not* support infinite precision division. (I'm used to BigNumber libraries that will treat results of division as rational numbers consisting of a numerator and denominator when possible, thereby allowing theoretically infinite precision). Some tests were failing because `(reallyBigNumber.sub(1)) / (reallyBigNumber) === new BigNumber(1)` for sufficiently large values of `reallyBigNumber`. This change helps, but it is worth noting that it is *still* possible for rounding issues when dividing really big numbers.

albrow:
Still working on this. I'll ping some people when it's ready for review again.

packages/utils/src/configured_bignumber.ts
Recmo: BigNumber.js is a float-point library with arbitrary precision. We should stay in the domain of integers, which means not using `div` but always using integer division with `divToInt`. Sometimes it requires a bit of algebra to create the correct verification logic in tests. (i.e. don't test A/B = C, but test A = B * C).

Recmo: For 512 bits of precision (product of two 256-bit numbers) we would need at least 155 decimals. Depending on how we build our tests, we might need more.

But like mentioned, I would rather stay in the exact domain and avoid rounding everywhere. According to the BigNumber.js docs only division and related operations such as square root, base conversion and negative powers involve rounding. It seems`divToInt` does not use `DECIMAL_PLACES` and is always precise.

Perhaps we can set `DECIMAL_PLACES` to a perversely low value to make tests fail that use imprecise operations.

albrow: I discovered `divToInt` after making this change and I am using now in the one place in our tests where we need to do division. So the immediate issue is solved, but it still kind of scares me that it's so easy to produce precision-related errors.

@fabioberger any thoughts here regarding how it affects our users?

packages/contracts/src/2.0.0/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -205,7 +205,6 @@ contract MixinExchangeCore is
     /// @param order that was filled.
     /// @param takerAddress Address of taker who filled the order.
     /// @param orderTakerAssetFilledAmount Amount of order already filled.
-    /// @return fillResults Amounts filled and fees paid by maker and taker.
```
albrow: The function doesn't actually return anything, so I updated this comment.

packages/contracts/test/exchange/fill_order.ts
```diff
@@ -2,7 +2,10 @@ import { BlockchainLifecycle } from '@0xproject/dev-utils';
 import * as _ from 'lodash';
 
 import { chaiSetup } from '../utils/chai_setup';
-import { CoreCombinatorialUtils, coreCombinatorialUtilsFactoryAsync } from '../utils/core_combinatorial_utils';
+import {
```
albrow: @fabioberger I renamed this file and some of the identifiers in it because they seem more focused on `fillOrder`. Does this sound right to you?

packages/contracts/test/exchange/libs.ts
```diff
@@ -77,96 +77,4 @@ describe('Exchange libs', () => {
             });
         });
     });
-
-    describe('LibMath', () => {
-        describe('isRoundingError', () => {
```
albrow: I *think* the the new test cases in the combinatorial tests for `isRounding` error are sufficient. However, let me know if I've missed something and we want to keep some of these tests around.

packages/contracts/test/exchange/libs.ts
```diff
@@ -77,96 +77,4 @@ describe('Exchange libs', () => {
...
-                const denominator = new BigNumber(676373677);
-                const target = new BigNumber(105762562);
-                // rounding error = ((76564*105762562/676373677) - floor(76564*105762562/676373677)) /
-                // (76564*105762562/676373677) = 0.0007%
-                const isRoundingError = await libs.publicIsRoundingError.callAsync(numerator, denominator, target);
-                expect(isRoundingError).to.be.false();
-            });
-        });
-
-        describe('getPartialAmount', () => {
```
albrow: I'm a lot more confident that all of these cases are covered in the combinatorial test for `getPartialAmount`. I think it is safe to remove these.

packages/contracts/test/utils/combinatorial_utils.ts
```diff
@@ -0,0 +1,119 @@
...
+ * @param testFuncAsync a test function which, e.g., makes a call or sends a
+ * transaction to a contract. It accepts the same N arguments returns the
+ * "actual result" or "actual error" for a given test case.
+ * @param values an array of N arrays. Each inner array is a set of possible
+ * values which are each passed into both the reference function and the test
+ * function.
+ * @return A Promise that resolves if the test passes and rejects if the test
+ * fails, according to the rules described above.
+ */
+export async function testCombinatoriallyWithReferenceFuncAsync(
```
albrow: It took a significant amount of type-wrangling but the resulting generic function is very powerful. In the near future I would like to adapt more of our existing tests to use this.

@fabioberger and @Recmo does this look like something you could utilize in existing/future combinatorial tests? Is anything unclear about how it works or is there some new behavior you would like to add?

packages/migrations/artifacts/2.0.0/IAssetData.json
```diff
@@ -61,9 +61,9 @@
...
+		"2.0.0/protocol/AssetProxy/interfaces/IAssetData.sol": "/*\n\n  Copyright 2018 ZeroEx Intl.\n\n  Licensed under the Apache License, Version 2.0 (the \"License\");\n  you may not use this file except in compliance with the License.\n  You may obtain a copy of the License at\n\n    http://www.apache.org/licenses/LICENSE-2.0\n\n  Unless required by applicable law or agreed to in writing, software\n  distributed under the License is distributed on an \"AS IS\" BASIS,\n  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n  See the License for the specific language governing permissions and\n  limitations under the License.\n\n*/\n\n// solhint-disable\npragma solidity 0.4.24;\n\n\n// @dev Interface of the asset proxy's assetData.\n// The asset proxies take an ABI encoded `bytes assetData` as argument.\n// This argument is ABI encoded as one of the methods of this interface.\ninterface IAssetData {\n\n    function ERC20Token(address tokenContract)\n        external\n        pure;\n    \n    function ERC721Token(\n        address tokenContract,\n        uint256 tokenId,\n        bytes receiverData\n    )\n        external\n        pure;\n    \n}\n"
```
albrow: I didn't want to modify all these artifacts but it was the only way I could figure out how to fix the merge conflicts and get tests passing on CI.

albrow:
Removing WIP tag. This PR is ready for review.

I was able to get around most performance concerns by reducing the size of the input set. For example in `addFillResults` there are a total of 8 uint256 parameters (4 each for 2 `FillResults`). However, most of the values are independent in the sense that they are not used together in any mathematical operation. I reason that we do not need to test independent values combinatorially because changing one of them does not have any effect on the other. Therefore instead of testing all combinations of 8 uint256 parameters, I reduced the input set to 2 uint256 parameters. One other optimization that I did was skipping `startAsync` and `revertAsync` for tests which do not modify the state at all (e.g. because the tests only make contract calls and do not send any transactions).

After all of this, the total runtime for the combinatorial tests I added in this PR is ~60s for Geth and ~80s for Ganache. I think this is much more palatable and I am happy running these tests frequently both locally and in CI.

Of the original internal functions that @Recmo suggested we test, here are the ones that are covered in this PR:

- [x] `addFillResults`
- [x] `getPartialAmount`
- [x] `isRoundingError`
- [ ] `assertValidFill`
- [x] `calculateFillResults`
- [x] `updateFilledState`
- [ ] `settleOrder`

`assertValidFill` is much more complicated and IMO it would be too difficult to write a reference function. In particular, we would need to re-implement `isValidSignature` in JavaScript/TypeScript. I don't think it is a good fit for this kind of testing strategy.

`settleOrder` is not a good fit IMO for the opposite reason. It doesn't really do anything but call `dispatchTransferFrom`. There is no reason to believe that combinations of different inputs would have any effect on it (it depends on entirely on `dispatchTransferFrom`).

If there are any other internal functions that we think would be a good fit for this testing strategy and should be added in this PR, I would be happy to discuss.

packages/contracts/test/exchange/fill_order.ts
fabioberger: Yep, makes sense. Initially I thought I'd get to more then fillOrder, but that didn't happen.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,326 @@
...
+        }
+        async function referenceAddFillResultsAsync(
+            totalValue: BigNumber,
+            singleValue: BigNumber,
+        ): Promise<FillResults> {
+            // Note(albrow): Here, each of totalFillResults and
+            // singleFillResults will consist of fields with the same values.
+            // This should be safe because none of the fields in a given
+            // FillResults are ever used together in a mathemetical operation.
+            // They are only used with the corresponding feild from *the other*
```
fabioberger: Typo: `feild`

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,326 @@
...
+            // FillResults, which are different.
+            const totalFillResults = makeFillResults(totalValue);
+            const singleFillResults = makeFillResults(singleValue);
+            // Note(albrow): _.mergeWith mutates the first argument! To
```
fabioberger: Let's stick to the convention of prefixing comments about hacks/workarounds with `// HACK`.

packages/contracts/test/utils/combinatorial_utils.ts
fabioberger: Well done! My only concern is whether I'd be able to immediately know the test case details that caused the test to fail. Could we perhaps include the `testCase` in the test description?

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,98 @@
...
+export async function testWithReferenceFuncAsync<P0, P1, P2, P3, R>(
+    referenceFunc: (p0: P0, p1: P1, p2: P2, p3: P3) => Promise<R>,
+    testFunc: (p0: P0, p1: P1, p2: P2, p3: P3) => Promise<R>,
+    values: [P0, P1, P2, P3],
+): Promise<void>;
+export async function testWithReferenceFuncAsync<P0, P1, P2, P3, P4, R>(
+    referenceFunc: (p0: P0, p1: P1, p2: P2, p3: P3, p4: P4) => Promise<R>,
+    testFunc: (p0: P0, p1: P1, p2: P2, p3: P3, p4: P4) => Promise<R>,
+    values: [P0, P1, P2, P3, P4],
+): Promise<void>;
```
fabioberger: Why are these types defined in two places? Let's declare them in one place and re-export them in the other.

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,98 @@
...
+        expectedErr = e.message;
+    }
+    try {
+        const actualResult = await testFuncAsync(...values);
+        if (!_.isUndefined(expectedErr)) {
+            throw new Error(
+                `Expected error containing ${expectedErr} but got no error\n\tTest case: ${testCaseString}`,
+            );
+        }
+        expect(JSON.stringify(actualResult)).to.equal(JSON.stringify(expectedResult));
```
fabioberger: If this fails, it won't print out the `testCaseString`. Can we make that happen?

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,98 @@
...
+function _getParameterNames(func: (...args: any[]) => any): string[] {
+    return _.toString(func)
+        .replace(/[/][/].*$/gm, '') // strip single-line comments
+        .replace(/\s+/g, '') // strip white space
+        .replace(/[/][*][^/*]*[*][/]/g, '') // strip multi-line comments
+        .split('){', 1)[0]
+        .replace(/^[^(]*[(]/, '') // extract the parameters
+        .replace(/=[^,]+/g, '') // strip any ES6 defaults
+        .split(',')
+        .filter(Boolean); // split & filter [""]
```
fabioberger: Did you write this or did you get it from somewhere online? If online, please add a `// Source: ` comment above it.

LogvinovLeon:
Looks like you've generated an extra large artifact for ZRX token. With intermediate asm and ast. It is over 9000 lines. Do we need it?

packages/contracts/test/utils/combinatorial_utils.ts
albrow: So the way it works right now is that if any test fails, it prints out the input values that caused it to fail (and pulls the names for these arguments from the reference function to help you understand what they represent). It look like this:

```
1) Exchange core internal functions
        getPartialAmount
            getPartialAmount 10/1000:

    Error: expected '"0"' to equal '"100"'
    Test case: {"numerator":"0","denominator":"1","target":"0"}
```

I think we really only care about the input values if a test fails and including the inputs in each test description can clutter the terminal output. Does that sound good? Is there enough information printed on test failure?

packages/contracts/test/utils/test_with_reference.ts
albrow: I'm not sure what you mean exactly. There is a subtle differences between the type definitions for `testWithReferenceFuncAsync` and `testCombinatoriallyWithReferenceFuncAsync`. Namely, the combinatorial one expects an array of arrays of values instead of just an array of values.

The only thing I can think of to reduce code duplication here in the type definitions is to define types like this:

```typescript
type asyncFunc2<P0, P1, R> = (p0: P0, p1: P1) => Promise<R>;
type asyncFunc3<P0, P1, P2, R> = (p0: P0, p1: P1, p2: P2) => Promise<R>;
// ...
```

And then use them like this:

```typescript
export async function testWithReferenceFuncAsync<P0, P1, R>(
    referenceFunc: asyncFunc2<P0, P1, R>,
    testFunc: asyncFunc2<P0, P1, R>,
    values: [P0, P1],
): Promise<void>;
export async function testWithReferenceFuncAsync<P0, P1, P2, R>(
    referenceFunc: asyncFunc3<P0, P1, P2, R>,
    testFunc: asyncFunc3<P0, P1, P2, R>,
    values: [P0, P1, P2],
): Promise<void>;
// ...
```

But IMO this is less readable because it requires readers to then look at the definition for `asyncFunc2`, `asyncFunc3`, etc. And it doesn't really make our code any shorter either.

I did go ahead rename the `values` argument in `testCombinatoriallyWithReferenceFuncAsync` to `allValues` to make the difference more obvious. Do you have any other ideas? Maybe there's something I'm missing.

packages/contracts/test/exchange/libs.ts
albrow: After looking at this in more depth I decided to re-add some test cases here for `isRoundingError` that are not covered by the combinatorial tests.

packages/contracts/test/utils/combinatorial_utils.ts
fabioberger: Yep, that works great.

packages/contracts/test/utils/test_with_reference.ts
fabioberger: Makes sense. Nope, think this works.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+    feeRecipientAddress: '0x0000000000000000000000000000000000000000',
+    expirationTimeSeconds: new BigNumber(0),
+};
+
+const emptySignedOrder: SignedOrder = {
+    ...emptyOrder,
+    signature: '',
+};
+
+interface FillResults {
```
abandeali1: Can we add this to `types.ts`? I'm guessing we'll use it elsewhere at some point.

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
```diff
@@ -0,0 +1,109 @@
...
+    ///      Modifies the first FillResults instance specified.
+    ///      Note that this function has been modified from the original
+    //       internal version to return the FillResults.
+    /// @param totalFillResults Fill results instance that will be added onto.
+    /// @param singleFillResults Fill results instance that will be added to totalFillResults.
+    /// @return newTotalFillResults The result of adding singleFillResults to totalFilResults.
+    function publicAddFillResults(FillResults memory totalFillResults, FillResults memory singleFillResults)
+        public
+        pure
+        returns (FillResults)
```
abandeali1: `returns (FillResults memory)`

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
```diff
@@ -0,0 +1,109 @@
...
+pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
+
+import "../../protocol/Exchange/Exchange.sol";
+
+
+contract TestExchangeInternals is
+    Exchange
+{
+    constructor() public Exchange("") {}
```
abandeali1:
```
constructor ()
    public
    Exchange("")
{}
```

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @dev Calculates partial value given a numerator and denominator.
+    /// @param numerator Numerator.
+    /// @param denominator Denominator.
+    /// @param target Value to calculate partial of.
+    /// @return Partial value of target.
+    function publicGetPartialAmount(
+        uint256 numerator,
+        uint256 denominator,
+        uint256 target)
```
abandeali1: Ending paren should be on a newline (like the above function).

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @dev Checks if rounding error > 0.1%.
+    /// @param numerator Numerator.
+    /// @param denominator Denominator.
+    /// @param target Value to multiply with numerator/denominator.
+    /// @return Rounding error is present.
+    function publicIsRoundingError(
+        uint256 numerator,
+        uint256 denominator,
+        uint256 target)
```
abandeali1: Ditto.

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
```diff
@@ -0,0 +1,109 @@
...
+    function publicUpdateFilledState(
+        Order memory order,
+        address takerAddress,
+        bytes32 orderHash,
+        uint256 orderTakerAssetFilledAmount,
+        FillResults memory fillResults
+    )
+        public
+    {
+        updateFilledState(order, takerAddress, orderHash, orderTakerAssetFilledAmount, fillResults);
```
abandeali1: Let's split this function call into multiple lines.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+            const fillResults = {
+                makerAssetFilledAmount: new BigNumber(0),
+                takerAssetFilledAmount,
+                makerFeePaid: new BigNumber(0),
+                takerFeePaid: new BigNumber(0),
+            };
+            await web3Wrapper.awaitTransactionSuccessAsync(
+                await testExchange.publicUpdateFilledState.sendTransactionAsync(
+                    emptySignedOrder,
+                    '0x0000000000000000000000000000000000000000',
```
abandeali1: `constants.NULL_ADDRESS`

packages/contracts/test/utils/combinatorial_utils.ts
```diff
@@ -0,0 +1,119 @@
...
+    // Min that does overflow.
+    new BigNumber(2).pow(256),
+];
+
+// A set of values corresponding to the bytes32 type in Solidity. This set
+// contains some notable edge cases, including some values which will overflow
+// the bytes32 type.
+export const bytes32Values = [
+    // Min
+    '0x00000000000000000000000000000000',
```
abandeali1: These are all only 16 bytes.

packages/contracts/test/exchange/internal.ts
albrow: 👍 It was actually already there so I just imported it.

packages/contracts/test/utils/combinatorial_utils.ts
albrow: Nice catch. When I updated the values here it exposed an interesting edge case in `publicUpdateFilledState`. It appears to always return `0` if `orderHash` overflows the `bytes32` type. (Previously, I *though* I was overflowing the `bytes32` type but the input value didn't actually have enough bytes).

Effectively this means that when you call the internal function `updateFilledState` it does not work right if `orderHash` overflows `bytes32`. This might be fine since it's an internal function and we would never expect this to happen. What do you think?

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,110 @@
...
+): Promise<void> {
+    const paramNames = _getParameterNames(referenceFunc);
+    const testCaseString = JSON.stringify(_.zipObject(paramNames, values));
+    let expectedResult: any;
+    let expectedErr: string | undefined;
+    try {
+        expectedResult = await referenceFunc(...values);
+    } catch (e) {
+        expectedErr = e.message;
+    }
```
Recmo: I think this would be a bit cleaner if the actual tests are done inside the try/catch clauses.

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,110 @@
...
+    } catch (e) {
+        if (_.isUndefined(expectedErr)) {
+            throw new Error(`${e.message}\n\tTest case: ${testCaseString}`);
+        } else {
+            expect(e.message).to.contain(expectedErr, `${e.message}\n\tTest case: ${testCaseString}`);
+        }
+    }
+    if (!_.isUndefined(actualResult) && !_.isUndefined(expectedResult)) {
+        expect(JSON.stringify(actualResult)).to.equal(
+            JSON.stringify(expectedResult),
```
Recmo: `JSON.stringify` is not a safe comparison. Order of properties can differ even though the objects are the same. Why not `expect(actualResult).to.deep.equal(expectedResult)`?

packages/contracts/test/utils/test_with_reference.ts
Recmo: This looks like quite a hack. I would be okay with outputting an tuple/array of values without names and removing this hack.

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -0,0 +1,110 @@
...
+ * @return A Promise that resolves if the test passes and rejects if the test
+ * fails, according to the rules described above.
+ */
+export async function testWithReferenceFuncAsync(
+    referenceFunc: (...args: any[]) => Promise<any>,
+    testFuncAsync: (...args: any[]) => Promise<any>,
+    values: any[],
+): Promise<void> {
+    const paramNames = _getParameterNames(referenceFunc);
+    const testCaseString = JSON.stringify(_.zipObject(paramNames, values));
```
Recmo: This is only needed on test failure, but is now a nontrivial computation for every test.

packages/utils/src/configured_bignumber.ts
```diff
@@ -1,9 +1,14 @@
...
-// more then 20 digits. We want to avoid this behavior, so we set EXPONENTIAL_AT to a high number
 BigNumber.config({
+    // By default BigNumber's `toString` method converts to exponential notation if the value has
+    // more then 20 digits. We want to avoid this behavior, so we set EXPONENTIAL_AT to a high number
     EXPONENTIAL_AT: 1000,
+    // Note(albrow): This is the lowest value for which
+    // `x.div(y).floor() === x.divToInt(y)`
+    // for all values of x and y <= MAX_UINT256, where MAX_UINT256 is the
+    // maximum number represented by the uint256 type in Solidity (2^256-1).
+    DECIMAL_PLACES: 78,
```
Recmo: We should never use `div(..)` or `floor()`. Always use `divToInt(..)`. But that is out of scope for this PR.

packages/contracts/test/utils/combinatorial_utils.ts
```diff
@@ -0,0 +1,119 @@
...
+    new BigNumber(2).pow(255),
+    // Max that does not overflow.
+    new BigNumber(2).pow(256).minus(1),
+    // Min that does overflow.
+    new BigNumber(2).pow(256),
+];
+
+// A set of values corresponding to the bytes32 type in Solidity. This set
+// contains some notable edge cases, including some values which will overflow
+// the bytes32 type.
```
Recmo: `bytes32` can't overflow. There are no math operations defined on it.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+            orderTakerAssetFilledAmount: BigNumber,
+            // tslint:disable-next-line:no-unused-variable
+            orderHash: string,
+        ): Promise<BigNumber> {
+            const totalFilledAmount = takerAssetFilledAmount.add(orderTakerAssetFilledAmount);
+            if (totalFilledAmount.greaterThan(MAX_UINT256)) {
+                throw overflowErrorForSendTransaction;
+            }
+            // Note(albrow): In the Solidity implementation, if orderHash
+            // overflows bytes32, the return value is always 0.
```
Recmo: Solidity has no overflows for `bytes32`. Maybe you are testing the input encoders of the Javascript wrappers? But these don't need testing.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+        ): Promise<boolean> {
+            if (numerator.greaterThan(MAX_UINT256)) {
+                throw invalidOpcodeErrorForCall;
+            } else if (denominator.greaterThan(MAX_UINT256)) {
+                throw invalidOpcodeErrorForCall;
+            } else if (denominator.eq(new BigNumber(0))) {
+                throw invalidOpcodeErrorForCall;
+            } else if (target.greaterThan(MAX_UINT256)) {
+                throw invalidOpcodeErrorForCall;
+            }
```
Recmo: No need to check for `MAX_UINT256`.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+            if (remainder.eq(new BigNumber(0))) {
+                return false;
+            }
+            if (product.greaterThan(MAX_UINT256)) {
+                throw overflowErrorForCall;
+            }
+            if (product.eq(new BigNumber(0))) {
+                throw invalidOpcodeErrorForCall;
+            }
+            const remainderTimes1000000 = remainder.mul(new BigNumber('1000000'));
```
Recmo: `mul(1000000)`?

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+
+async function _getInvalidOpcodeErrorForCall(): Promise<Error> {
+    const errMsg = await getGanacheOrGethError('invalid opcode', 'Contract call failed');
+    return new Error(errMsg);
+}
+
+async function _getOverflowErrorForSendTransaction(): Promise<Error> {
+    const errMsg = await getGanacheOrGethError('UINT256_OVERFLOW', 'always failing transaction');
+    return new Error(errMsg);
+}
```
Recmo: I think these should be part of the utils. We don't want to change every test file when we add Parity or some other client.

But synchronizing the error objects between the different clients is a challenge. Fortunately there are only a couple of errors we should be concerned with (OutOfGas, InvalidOpcode, InvalidJump, Revert(message), etc.)

packages/contracts/test/exchange/fill_order.ts
```diff
@@ -2,7 +2,10 @@ import { BlockchainLifecycle } from '@0xproject/dev-utils';
...
+    FillOrderCombinatorialUtils,
+    fillOrderCombinatorialUtilsFactoryAsync,
+} from '../utils/fill_order_combinatorial_utils';
```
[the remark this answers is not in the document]
albrow: As soon as this is merged, I'm planning to work with Fabio to update `fill_order.ts` to use the generic functions in this PR. Mind if we wait to do further refactoring until then?

packages/contracts/test/utils/test_with_reference.ts
albrow: @Recmo I actually found it really hard to understand a test failure when all I could see was an array of 3-4 numbers without any additional context. This is a hack, but it has real utility, is only for testing, and nothing bad is really going to happen if it breaks. It's pretty well isolated too. Okay if we keep it?

packages/utils/src/configured_bignumber.ts
albrow: AFAIK, we don't. This is for the benefit of our users just in case they do it.

packages/contracts/test/utils/test_with_reference.ts
albrow: `expect(actualResult).to.deep.equal(expectedResult)` results in output that is a lot harder to read when a test fails, especially for the `BigNumber` type, which is frequently used.

```
 |       AssertionError: Test case: {"numerator":"0","denominator":"1","target":"0"}: expected { Object (s, e, ...) } to deeply equal { Object (s, e, ...) }
 |       + expected - actual
 |
 |        {
 |          "c": [
 |       -    0
 |       +    10
 |          ]
 |       -  "e": 0
 |       +  "e": 1
 |          "s": 1
 |        }
 |
```

That said, `testWithReferenceFuncAsync` is supposed to work for any type, so it's important to use a reliable method of comparison. I went ahead and made this change.

If anyone else knows of a way to have both reliable comparison and readable output (particularly for `BigNumber` types) I'd be happy to hear suggestions.

packages/contracts/test/utils/test_with_reference.ts
albrow: I'm not sure what you mean. Can you provide an example?

packages/contracts/test/utils/test_with_reference.ts
albrow: 👍I changed the code so it is only computed on test failure.

packages/contracts/test/exchange/internal.ts
```diff
@@ -0,0 +1,329 @@
...
+    numerator: BigNumber,
+    denominator: BigNumber,
+    target: BigNumber,
+): Promise<BigNumber> {
+    const invalidOpcodeErrorForCall = await _getInvalidOpcodeErrorForCall();
+    if (numerator.greaterThan(MAX_UINT256)) {
+        throw invalidOpcodeErrorForCall;
+    } else if (target.greaterThan(MAX_UINT256)) {
+        throw invalidOpcodeErrorForCall;
+    }
```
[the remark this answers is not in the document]
albrow: @Recmo while testing, I actually did find a bug in the encoding process when you provide values too big to fit in the corresponding Solidity types. It took me a while to figure out what was going on, but it is now clear that this doesn't have anything to with contracts and should be fixed separately.

I removed the too-big-to-fit types from the set of values in __combinatorial_utils.ts__ and removed this and related checks from the reference functions.

packages/contracts/src/2.0.0/test/TestExchangeInternals/TestExchangeInternals.sol
fabioberger: We really need a working linter for this kind of stuff...

packages/contracts/test/utils/test_with_reference.ts
Recmo: Seems like something we should patch in Chai.

PR #810, opened 2018-07-03
Remove increase balance link from balance_bounded_input component

BMillman19:
## Description

* This fixes a bug where when entering an amount of ether to wrap / unwrap. The input error when you have an insufficient balance links you to part of the app that doesn't exist anymore
* This PR just removes that link

## Testing instructions

* Click on "wrap" in the wallet
* Type in an amount of ETH to wrap that is greater than your current balance
* Observe that you can see an "Insufficient balance" error without a link

## Types of changes

## Checklist:


fabioberger:
@BMillman19 what about other places where the `balance_bounded_input` field is used? E.g the `Fill order` form? Do we not care to link people to somewhere where they can request test ZRX and ETH?

PR #812, opened 2018-07-03
Improve robustness of revert reason assertions

albrow:
## Description

This PR addresses some of the issues with the band-aid introduced in https://github.com/0xProject/0x-monorepo/pull/796. The biggest change is that we now switch on the node type and check for certain things depending on whether we are running against Geth or Ganache.

There are now four main assertions:

- `expectTransactionFailedAsync`
- `expectTransactionFailedWithoutReasonAsync`
- `expectContractCallFailed`
- `expectContractCallFailedWithoutReasonAsync`

This naming convention more accurately describes what is happening in the assertion and also abstracts away some of the details about what specific error messages are. The old naming convention (e.g. `expectRevertOrAlwaysFailingTransaction`) provides unnecessary details and would not scale well if we added support for another node type.

The one really special case is that for `expectTransactionFailedAsync` we fall back to `expectTransactionFailedWithoutReasonAsync` when running tests against Geth because `eth_sendTranaction` does not return revert reasons on Geth. I also added a warning in this case so that whoever is writing the test will be more aware of this restriction.

## Testing instructions

Just run `PKG=contracts yarn test`.

## Types of changes

## Checklist:


packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -489,6 +495,21 @@ export class Web3Wrapper {
     public async setHeadAsync(blockNumber: number): Promise<void> {
         await this._sendRawPayloadAsync<void>({ method: 'debug_setHead', params: [this._web3.toHex(blockNumber)] });
     }
+    /**
+     * Returns either NodeType.Geth or NodeType.Ganache depending on the type of
+     * the backing Ethereum node. Throws for any other type of node. This
+     * function caches the result and so subsequent calls are fast.
+     */
+    public async getNodeTypeAsync(): Promise<NodeType> {
```
albrow: Since both the assertions and __blockchain_lifecycle.ts__ switch on the `NodeType`, I decided to add the common functionality here.

packages/contracts/test/exchange/match_orders.ts
```diff
@@ -602,7 +602,7 @@ describe('matchOrders', () => {
             // Cancel left order
             await exchangeWrapper.cancelOrderAsync(signedOrderLeft, signedOrderLeft.makerAddress);
             // Match orders
-            return expectRevertReasonOrAlwaysFailingTransactionAsync(
+            return expectTransactionFailedAsync(
```
albrow: On #796, @fabioberger asked:

> Does it still verify the balances?

(For context, I changed this from `matchOrderTester.matchOrdersAndVerifyBalancesAsync` because the new assertions are more strict about the type of Promises they accept).

No, but neither did the old version. If you look closely at the [implementation of `matchOrderTester.matchOrdersAndVerifyBalancesAsync`](https://github.com/0xProject/0x-monorepo/blob/5b64b3ea937326978b5742ec1b3692ebe5c41991/packages/contracts/test/utils/match_order_tester.ts#L111) it calls `await this._exchangeWrapper.matchOrdersAsync(signedOrderLeft, signedOrderRight, takerAddress);` prior to checking the balances. If `matchOrdersAsync` rejects, it does not check the balances. Maybe this was an oversight?

packages/contracts/test/utils/assertions.ts
```diff
@@ -1,108 +1,153 @@
...
-    p: Promise<string | TransactionReceiptWithDecodedLogs | TransactionReceipt>,
-    reason: RevertReason,
-): Promise<void> {
+export async function expectTransactionFailedWithoutReasonAsync(p: sendTransactionResult): Promise<void> {
     return p
         .then(async result => {
-            let txReceiptStatus: string | 0 | 1 | null;
-            if (typeof result === 'string') {
-                // Result is a txHash. We need to make a web3 call to get the receipt.
+            let txReceiptStatus: null | string | 0 | 1;
```
albrow: On #796, @fabioberger asked:

> Do we not have a type for this?

No, we don't have a type for this. Seems like something that should be added to `ethereum-types` in a separate PR?

albrow:
On #796, @abandeali1 asked:

> Can we maybe split this out into another function? This change makes it so that tests can pass and we think that the revert reason is being checked when it actually isn't.

All cases where we were doing this in the past should be addressed by the fact that we switch on the `NodeType`. To be extra careful, I also added a log message whenever the revert reason is not supported (currently this is only for `eth_sendTransaction` on Geth).

albrow:
@fabioberger any other feedback you left on #796 should be fixed here. Please double-check 😄

packages/contracts/test/exchange/match_orders.ts
fabioberger: @hysz was it an oversight?

packages/contracts/test/utils/assertions.ts
fabioberger: Yep, could be in this PR actually, since it's also in the monorepo and we already have `ethereum-types` as a dep of `contracts` package.

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -489,6 +495,21 @@ export class Web3Wrapper {
     public async setHeadAsync(blockNumber: number): Promise<void> {
         await this._sendRawPayloadAsync<void>({ method: 'debug_setHead', params: [this._web3.toHex(blockNumber)] });
     }
+    /**
+     * Returns either NodeType.Geth or NodeType.Ganache depending on the type of
+     * the backing Ethereum node. Throws for any other type of node. This
+     * function caches the result and so subsequent calls are fast.
```
fabioberger: I don't see any caching going on here, and there shouldn't be any. Please remove this sentence.

packages/contracts/test/utils/assertions.ts
albrow: 👍 I'll update it. I guess I misunderstood how ethereum-types fit in with the rest of our packages.

packages/web3-wrapper/src/web3_wrapper.ts
albrow: I meant to add caching here (just did so in the most recent commit). I think we should cache this. This is a function that gets called more than a thousand times and the `NodeType` never changes for a given instance of `Web3Wrapper`. I just checked and it does improve performance by a few seconds (not a huge difference but it's 5 lines of code so probably worth it).

albrow: For context: Running tests against Geth went from ~60s to ~53s and Ganache went from ~81s to ~78s.

packages/contracts/test/exchange/match_orders.ts
hysz: IIRC the idea was that if `matchOrdersAsync` throws, then the transaction reverted and balances need not be checked. Now that we enforce _revert reasons_ the tests are far more robust 👍.

PR #816, opened 2018-07-03
Make onboarding and wallet copy dynamic based on OS

fragosti:
## Description

I added some utilities to read the operating system and then I'm switching on that to dynamically prompt the user to install toshi or install metamask.
screen shot 2018-07-03 at 1 37 52 pm
screen shot 2018-07-03 at 2 29 10 pm
## Types of changes

*   [ ] Bug fix (non-breaking change which fixes an issue)
*   [ ] New feature (non-breaking change which adds functionality)
*   [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

*   [ ] Change requires a change to the documentation.
*   [ ] Added tests to cover my changes.
*   [ ] Added new entries to the relevant CHANGELOG.jsons.
*   [ ] Labeled this PR with the 'WIP' label if it is a work in progress.

] Labeled this PR with the labels corresponding to the changed package.

PR #820, opened 2018-07-04
Various mobile portal fixes and top bar layout refactor

BMillman19:
## Description

* Removed "0x Relayers" and "Your" account headers in mobile
* Decreased margins on mobile
* Reworked the top bar to use flex box for centering instead of hard coded padding
* Removed a couple of options from the wallet menu that aren't relevant to mobile

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/components/portal/portal.tsx
+++ b/packages/website/ts/components/portal/portal.tsx
@@ -91,7 +91,7 @@ interface PortalState {
 
 interface AccountManagementItem {
     pathName: string;
-    headerText: string;
+    headerText?: string;
     render: () => React.ReactNode;
 }
 
@@ -106,7 +106,8 @@ const TOP_BAR_HEIGHT = TopBar.heightForDisplayType(TopBarDisplayType.Expanded);
 const LEFT_COLUMN_WIDTH = 346;
 const MENU_PADDING_LEFT = 185;
 const LARGE_LAYOUT_MAX_WIDTH = 1200;
-const LARGE_LAYOUT_MARGIN = 30;
+const LARGE_LAYOUT_SIDE_PADDING = 30;
+const SMALL_LAYOUT_SIDE_PADDING = 20;
 
 export class Portal extends React.Component<PortalProps, PortalState> {
     private _blockchain: Blockchain;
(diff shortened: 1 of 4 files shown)


packages/website/ts/components/portal/portal.tsx
```diff
@@ -243,7 +244,9 @@ export class Portal extends React.Component<PortalProps, PortalState> {
                         position: 'fixed',
                         zIndex: zIndex.topBar,
                     }}
-                    maxWidth={LARGE_LAYOUT_MAX_WIDTH}
+                    maxWidth={!this._isSmallScreen() ? LARGE_LAYOUT_MAX_WIDTH : undefined}
+                    paddingLeft={!this._isSmallScreen() ? LARGE_LAYOUT_SIDE_PADDING : SMALL_LAYOUT_SIDE_PADDING}
+                    paddingRight={!this._isSmallScreen() ? LARGE_LAYOUT_SIDE_PADDING : SMALL_LAYOUT_SIDE_PADDING}
```
fragosti: Should we use css for this?

packages/website/ts/components/top_bar/top_bar.tsx
```diff
@@ -211,15 +216,21 @@ export class TopBar extends React.Component<TopBarProps, TopBarState> {
...
-                    <div className="col col-2 sm-pl1 md-pl2 lg-pl0" style={{ paddingTop: 15 }}>
-                        <Link to={`${WebsitePaths.Home}`} className="text-decoration-none">
-                            <img src={logoUrl} height="30" />
-                        </Link>
-                    </div>
-                    <div className={`col col-${isExpandedDisplayType ? '8' : '9'} lg-hide md-hide`} />
-                    <div className={`col col-${isExpandedDisplayType ? '6' : '5'} sm-hide xs-hide`} />
+            <div
+                style={{ ...styles.topBar, ...bottomBorderStyle, ...this.props.style, ...{ height } }}
+                className="pb1 flex items-center"
```
fragosti: Yea I think it's probably better to do this responsiveness stuff by using styled components and media queries? I added a way to do this that is currently broken but fixed here https://github.com/0xProject/0x-monorepo/pull/816/files#diff-00f964471edfccc5064ec54c2d1299c1R5

I think it makes sense to do it in JS when you can't do it in CSS, like when the layout is completely different on mobile and desktop.

BMillman19: How do you suggest we update the `TopBarProps` interface to best support this. I want the max-width and padding to reflect how we are laying out `LargeLayout` and `SmallLayout` in `Portal`

PR #821, opened 2018-07-04
Remove Web3.js From 0x-monorepo 🍾 

fabioberger:
## Description

This PR converts the remaining calls to `Web3.js` to use our `_sendRawPayloadAsync` method. Additionally, it also removes our reliance on `Web3`'s formatter for marshalling and unmarshalling JSON RPC payloads. We now do this ourselves.

This PR should reduce 0x.js bundle size by at least 8MB, and also reduce the installation time of the package drastically.

## Testing instructions

Run all the tests (includes additional test coverage for the web-wrapper package).

## Types of changes

## Checklist:


packages/assert/src/index.ts
```diff
@@ -41,8 +41,8 @@ export const assert = {
         value: string,
         stringEnum: any /* There is no base type for every string enum */,
     ): void {
-        const doesBelongToStringEnum = !_.isUndefined(stringEnum[value]);
-        const enumValues = _.keys(stringEnum);
+        const enumValues = _.values(stringEnum);
+        const doesBelongToStringEnum = _.includes(enumValues, value);
```
fabioberger: Our assertion was previously testing the enum keys, not values (which is fine when they are the same... but not when they are not).

packages/base-contract/src/index.ts
```diff
@@ -80,7 +80,7 @@ export class BaseContract {
             // Awaiting https://github.com/Microsoft/TypeScript/pull/13288 to be merged
         } as any;
         if (_.isUndefined(txDataWithDefaults.gas) && !_.isUndefined(estimateGasAsync)) {
-            txDataWithDefaults.gas = await estimateGasAsync(txData);
+            txDataWithDefaults.gas = await estimateGasAsync(txDataWithDefaults as any);
```
fabioberger: We need to be passing in `from` to `estimateGasAsync`. @LogvinovLeon, do you see any issues with this change?

LogvinovLeon:
8 MB? What's the current bundle size? Can you attach webpack snapshots? I have hard time believing it.

packages/base-contract/src/index.ts
fabioberger: This is actually not true, we only need `from` when sending transactions, not when estimating gas. Reverted this change.

LogvinovLeon:
LGTM after the web3-wrapper CHANGELOG is updated

PR #824, opened 2018-07-05
Add and apply solidity linter (V2, updated)

focuzz:
## Adds Security and [Style Guide](http://solidity.readthedocs.io/en/develop/style-guide.html) checking and validation on solidity contracts.

This PR adds [Solhint](https://github.com/protofire/solhint) to achieve this goal.
I have only updated the code on those contracts where it had no impact on the functionality, mostly indentation issues. This should make code review fairly straightforward.

I have only configured the linter to output a warning instead of throwing errors as a starting point. Solhint can add more value. If you like this approach, more time can be invested analyzing and fixing those issues. Or you can simply update the linter configuration to ignore them without a warning.

Run:
`npm run lint-contracts`

### DIFF
--- a/packages/contracts/.solhint.json
+++ b/packages/contracts/.solhint.json
@@ -1,0 +1,17 @@
+{
+    "extends": "default",
+    "rules": {
+        "bracket-align": "warn",
+        "code-complexity": "warn",
+        "const-name-snakecase": "warn",
+        "expression-indent": "warn",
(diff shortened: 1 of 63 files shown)


abandeali1:
This should be good to merge after one small change. Can you add `&& yarn lint-contracts` to the `lint` command in package.json?

PR #825, opened 2018-07-05
OrderWatcher Fixes

fabioberger:
## Description

This PR:
- Makes OrderWatcher resilient against block re-orgs, and other blockchain quirks (e.g if transactions skip the nodes mempool and go straight to being confirmed). All these scenarios now lead to proper event emissions.
- Does not close the OrderWatcher subscription if it's underlying Blockstream instance encounters inconsistent responses from the backing Ethereum node.

## Testing instructions

- Run the tests.
- Start running an OrderWatcher, fetch all the orders on ERCDEX, add them to the OrderWatcher, visually verify that the OrderWatcher emits events, does not go down if blockstream errors are encountered.

## Types of changes

## Checklist:


packages/order-watcher/test/event_watcher_test.ts
```diff
@@ -1,126 +0,0 @@
-import { callbackErrorReporter } from '@0xproject/dev-utils';
```
fabioberger: These tests no longer make sense not that we just use blockstream under-the-hood.

LogvinovLeon:
Do you have more info on that test failure?

fabioberger:
Essentially Infura sometimes returns a response to `getBlockAsync` including a block other then the one requested. This naturally trips up Blockstream. 

Since Blockstream was refactored to handle all these edge cases by recovering to the last known checkpoint, we can now safely log (if verbose mode configured), and otherwise ignore these error messages.

PR #828, opened 2018-07-05
Fix some portal bugs, and poll for best gas price

fragosti:
## Description

This change addresses some small issues with Portal, as well as change our gas price polling set-up. Right now we just use the "average" gas price found on page-load, but it would be more ideal to poll that value and to use "fast" since it is hard to ask your users to go through onboarding steps with 5 minute waiting times.

## Testing instructions

1. Make sure no more annoying onboarding closes happen when you use Firefox.
2. Make sure that ledger dialog always shows a default network.
3. Make sure that the gas price being used allows for transactions to go through in a timely manner (and that it is being updated).

## Types of changes

## Checklist:

### DIFF
--- a/packages/website/ts/blockchain.ts
+++ b/packages/website/ts/blockchain.ts
@@ -89,6 +89,7 @@ export class Blockchain {
     private _userAddressIfExists: string;
     private _ledgerSubprovider: LedgerSubprovider;
     private _defaultGasPrice: BigNumber;
+    private _watchGasPriceIntervalId: NodeJS.Timer;
     private static _getNameGivenProvider(provider: Provider): string {
         const providerType = utils.getProviderType(provider);
         const providerNameIfExists = providerToName[providerType];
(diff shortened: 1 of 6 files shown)


packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
```diff
@@ -77,6 +77,8 @@ class PlainPortalOnboardingFlow extends React.Component<PortalOnboardingFlowProp
                 updateOnboardingStep={this._updateOnboardingStep.bind(this)}
                 disableOverlay={this.props.screenWidth === ScreenWidths.Sm}
                 isMobile={this.props.screenWidth === ScreenWidths.Sm}
+                // This is necessary to ensure onboarding stays open once the user unlocks metamask and clicks away
+                disableCloseOnClickOutside={utils.getBrowserType() === BrowserType.Firefox}
```
BMillman19: Can we try opera and brave as well?

packages/website/ts/blockchain.ts
```diff
@@ -798,8 +798,30 @@ export class Blockchain {
...
+    private _startWatchingGasPrice(): void {
+        if (!_.isUndefined(this._watchGasPriceIntervalId)) {
+            return; // we are already watching
+        }
+        // tslint:disable-next-line:no-floating-promises
+        this._updateDefaultGasPriceAsync();
+        this._watchGasPriceIntervalId = intervalUtils.setAsyncExcludingInterval(
+            this._updateDefaultGasPriceAsync.bind(this),
+            // 1 minute
+            60000,
```
BMillman19: Lets pull this out into a constant

packages/website/ts/components/onboarding/portal_onboarding_flow.tsx
fragosti: Added Opera (had same issue). Brave also seemed to have this issue but it's not as easy to detect so doesn't seem worth it.

PR #829, opened 2018-07-06
Fix invalid transaction format errors when going through `SingerSubprovider`

fragosti:
## Description
https://github.com/0xProject/0x-monorepo/pull/817 introduced a bug where most transactions going through `SignerSubprovider` fail. The problem is that we get raw JSON for the RPC call in `handleRequest` in `SingerSubprovider` but web3wrapper expects a `TxData`. The main difference is that some fields were strings / hex when they should have been number.

Questions:
When do I use number vs BigNumber?

## Testing instructions

Try toggling an allowance toggle on the website. It should not crash!

## Types of changes

## Checklist:


packages/utils/src/format_utils.ts
```diff
@@ -1,7 +1,9 @@
-import { BigNumber } from '@0xproject/utils';
```
fabioberger: Let's please not move this to the `utils` package. I am very unsatisfied with these methods that were largely adapted from Web3.js. Let's keep these internal to Web3Wrapper for now.

PR #836, opened 2018-07-06
Overflow free getPartialAmount and isRoundingError

Recmo:
## Description

Current implementation of `getPartialAmount` and `isRoundingError` can overflow, causing a failure of the transaction. These overflows do not return revert reasons and we never clearly specified when they could occur. This PR implements math that gives the correct result in all cases, without overflow.

This PR also adds the requirement that `numerator / denominator <= 1`. This guarantees that a result always exists (without it the result could be more than 2^256). In several places the order of the arguments had `target` and `numerator` swapped, this is corrected.

In `marketBuyOrder` and `marketBuyOrderNoThrow` the value acting as `numerator` could become larger than `denominator`. This creates a potential for an unsolvable overflow. It is fixed by taking the minimum of the outstanding amount and the order amount.

The math in `getPartialAmount` uses a whole collection of theorems I developed some time ago. They are derived and explained in a series of Medium posts:

1. https://medium.com/wicketh/mathemagic-chinese-remainder-theorem-a5604dfd96b0
2. https://medium.com/wicketh/mathemagic-full-multiply-27650fec525d
3. https://medium.com/wicketh/mathemagic-512-bit-division-in-solidity-afa55870a65
4. https://medium.com/wicketh/mathemagic-multiplicative-inverses-b093e2a0a6d3

### Error handling

This PR introduces a new revert reason `NUMERATOR_GT_DENOMINATOR` when `getPartialAmount` is called with a fraction larger than one. This should never happen under correct operation of the exchange. (It would indicate an order that is more than 100% filled).

### Gas cost and optimization

Impact on gas consumption has not been measured yet. It should be good for `getPartialAmount` since it only adds significant work when the current implementation would have failed. `isRoundingError` became substantially more complex, but it is still a rather insignificant function gas wise.

The `isRoundingError` would benefit significantly from being merged with `getPartialAmount` as they share a lot of intermediate values. We could have a `getPartialAmount` and `getPartialAmountWithBoundedError` where the latter would revert with `EXCESSIVE_ROUNDING_ERROR` when the error exceeds the treshold.

## Testing instructions

* Check the Math.
* Check the Solidity implementation.
* Check the JS reference implementation (requires only math, not Math).
* Uncomment and/or add more test values.
* Run the combinatorial tests to check equivalence of Solidity and reference.
* Hand compute the result on select values.
* Check changes to the Exchange contracts.
* Especially check changes to `marketBuyOrder` and `marketBuyOrderNoThrow`.

## Types of changes

## Checklist:


packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
```diff
@@ -37,13 +39,104 @@ contract LibMath is
...
+        assembly {
+            let mm := mulmod(target, numerator, not(0))
+            prod0 := mul(target, numerator)
+            prod1 := sub(sub(mm, prod0), lt(mm, prod0))
+        }
+        
+        // Handle non-overflow cases, 256 by 256 division
+        if (prod1 == 0) {
+            assembly {
+                // If denominator is zero then by the precodition
```
Recmo: precondition

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
```diff
@@ -37,13 +39,104 @@ contract LibMath is
...
+        inv *= 2 - denominator * inv; // inverse mod 2**16
+        inv *= 2 - denominator * inv; // inverse mod 2**32
+        inv *= 2 - denominator * inv; // inverse mod 2**64
+        inv *= 2 - denominator * inv; // inverse mod 2**128
+        inv *= 2 - denominator * inv; // inverse mod 2**256
+        // If denominator is zero, inv is now 128
+        
+        // Because the division is now exact we can divide by multiplying
+        // with the modular inverse of denominator. This will give us the
+        // correct result modulo 2**256. Since the precoditions guarantee
```
Recmo: preconditions

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
```diff
@@ -57,16 +150,89 @@ contract LibMath is
         pure
         returns (bool isError)
     {
-        uint256 remainder = mulmod(target, numerator, denominator);
-        if (remainder == 0) {
-            return false; // No rounding error.
+        // The relative error is defined as the difference between
+        // the actual value and ideal value divided by the ideal value.
+        // If the error must be less than or equal to 0.1% of the ideal
```
Recmo: - if

packages/contracts/test/exchange/lib_math.ts
```diff
@@ -0,0 +1,237 @@
...
+    // Wrapper arround mocha because Mocha doesn't know BigNumber
+    function expectEqual(result: any, expected: any) {
+        if (isBigNumber(expected)) {
+            expect(result).to.be.bignumber.equal(expected);
+        } else {
+            expect(result).to.equal(expected);
+        }
+    }
+
+    // We need to use funGen instead of fund directly because of:
```
Recmo: `funcGen`, `func`

Recmo:
TODO: Split testing code off from the new implementation.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
```diff
@@ -37,13 +39,104 @@ contract LibMath is
...
+        
+        // Make division exact by subtracting the remainder from [prod1 prod0]
+        // Compute remainder using mulmod
+        // mulmod(_, _, 0) == 0
+        uint256 remainder;
+        assembly {
+            remainder := mulmod(target, numerator, denominator)
+        }
+        // Subtract 256 bit number from 512 bit number
+        assembly {
```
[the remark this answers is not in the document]
Recmo: I wanted to minimize the number of assembly statements. (Back when this was written we did not use a lot of assembly yet)

fabioberger:
After speaking offline with @Recmo, his thoughts on this PR are: "I’m neutral to slightly positive towards it. It’s cool and it solves an edge case but it also adds significant complexity and the edge case is nearly nonexistent".

Since there isn't enough consensus that this is definitively the way forward at this time, I'm going to close this PR for now. Feel free to revisit, reopen in the future if we desire this refactor in the future.

PR #837, opened 2018-07-06
Use 0.4.10 in AssetProxyOwner

abandeali1:
## Description

- Hard codes Solidity 0.4.10 in `AssetProxyOwner`
- Removes use of `LibBytes` and adds `readBytes4` function (`LibBytes` does not compile with 0.4.10)

## Types of changes

## Checklist:


packages/contracts/src/2.0.0/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -16,16 +16,14 @@
 
 */
 
-pragma solidity ^0.4.10;
+pragma solidity 0.4.10;
```
Recmo: Why the downgrade to `0.4.10`?

packages/contracts/src/2.0.0/protocol/AssetProxyOwner/AssetProxyOwner.sol
```diff
@@ -40,7 +38,7 @@ contract AssetProxyOwner is
     modifier validRemoveAuthorizedAddressAtIndexTx(uint256 transactionId) {
         Transaction storage tx = transactions[transactionId];
         require(isAssetProxyRegistered[tx.destination]);
-        require(tx.data.readBytes4(0) == REMOVE_AUTHORIZED_ADDRESS_AT_INDEX_SELECTOR);
+        require(readBytes4(tx.data, 0) == REMOVE_AUTHORIZED_ADDRESS_AT_INDEX_SELECTOR);
```
Recmo: The using syntax exists in 0.4.10 but it would require a library for just the `readBytes4` function. Doesn't seem worth it. (no action)

packages/contracts/src/2.0.0/protocol/AssetProxyOwner/AssetProxyOwner.sol
abandeali1: The MultiSigWallet contracts have all been thoroughly audited with 0.4.10, so I'd rather not make any changes to their bytecode. Not really concerned about the efficiency of this contract since it will be used very infrequently.

PR #857, opened 2018-07-11
Add getOrdersInfo function

abandeali1:
## Description

This doesn't add a ton of functionality, but allows us to easily validate a batch of orders with a single RPC call.

TODO: Add tests

## Types of changes

## Checklist:

### DIFF
--- a/packages/contracts/test/exchange/core.ts
+++ b/packages/contracts/test/exchange/core.ts
@@ -15,14 +15,14 @@ import { ERC721ProxyContract } from '../../generated_contract_wrappers/erc721_pr
 import { ExchangeCancelEventArgs, ExchangeContract } from '../../generated_contract_wrappers/exchange';
 import { artifacts } from '../utils/artifacts';
 import { expectTransactionFailedAsync } from '../utils/assertions';
-import { getLatestBlockTimestampAsync } from '../utils/block_timestamp';
+import { getLatestBlockTimestampAsync, increaseTimeAndMineBlockAsync } from '../utils/block_timestamp';
 import { chaiSetup } from '../utils/chai_setup';
 import { constants } from '../utils/constants';
 import { ERC20Wrapper } from '../utils/erc20_wrapper';
 import { ERC721Wrapper } from '../utils/erc721_wrapper';
 import { ExchangeWrapper } from '../utils/exchange_wrapper';
 import { OrderFactory } from '../utils/order_factory';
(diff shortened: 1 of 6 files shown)


packages/contracts/test/exchange/core.ts
```diff
@@ -481,6 +482,118 @@ describe('Exchange core', () => {
...
+            const orderInfo = await exchangeWrapper.getOrderInfoAsync(signedOrder);
+            const expectedOrderHash = orderHashUtils.getOrderHashHex(signedOrder);
+            const expectedTakerAssetFilledAmount = takerAssetFillAmount;
+            const expectedOrderStatus = OrderStatus.CANCELLED;
+            expect(orderInfo.orderHash).to.be.equal(expectedOrderHash);
+            expect(orderInfo.orderTakerAssetFilledAmount).to.be.bignumber.equal(expectedTakerAssetFilledAmount);
+            expect(orderInfo.orderStatus).to.equal(expectedOrderStatus);
+        });
+        it('should return the correct orderInfo for an expired and unfilled order', async () => {
+            await increaseTimeAndMineBlockAsync(signedOrder.expirationTimeSeconds.toNumber());
```
albrow: `expirationTimeSeconds` is a timestamp, denoted as milliseconds since the unix epoch, whereas `increaseTimeAndMineBlockAsync` expects a duration in seconds. I think what you meant to do was this:

```typescript
const timeOffset = signedOrder.expirationTimeSeconds
    .sub(Date.now())
    .divToInt(1000)
    .toNumber();
await increaseTimeAndMineBlockAsync(timeOffset);
```

abandeali1: It was somewhat lazy, but increasing the time by the actual expiration timestamp does guarantee that the order will be expired and also shouldn't affect anything else.

albrow: There is no way to reverse a time offset and there is a limit to the total time offset that can be applied, so it actually can matter if the offset is way too big. (We're talking about ~1.5 trillion seconds when it should be ~600 seconds, so it really is a huge difference).

In addition, if we're trying to test that order expiration is working correctly, we should be testing that as accurately as possible. The expiration time goes through a lot of conversions and is not hard to imagine that there could be a bug somewhere (e.g. we forgot to convert seconds to milliseconds). As written, the test would pass even if there was a bug causing the expiration time to be off by 10 orders of magnitude.

I'm okay with adding a small buffer here, but not one that is this big.

PR #866, opened 2018-07-11
Integrate Heap analytics into the website

fragosti:
## Description

I basically removed react-ga, included Heap (which is not available as a UMD module) and replaced all the calls to GA with calls to Heap. 

I increased the amount of data we are logging and improved the way we are logging the data since Heap provides a better / more expansive API.
https://docs.heapanalytics.com/reference#addeventproperties. 

I also fixed `onPageLoadAsync` which wasn't actually working.

## Testing instructions

Play around with the website and go to our heap dashboard (I can send invite) and see how things are going! The main reason we want this is for the SQL feature though.

## Types of changes

## Checklist:


packages/website/ts/utils/analytics.ts
```diff
@@ -1,27 +1,88 @@
...
+            return heap;
+        } else {
+            throw new Error('Could not find the Heap SDK on the page.');
+        }
+    }
+    constructor(heap: HeapAnalytics) {
+        this._heap = heap;
+    }
+    // HeapAnalytics Wrappers
+    public async indentifyAsync(id: string, idType: string): Promise<void> {
```
BMillman19: Similar to my comment in the sourcemaps PR, is there a clean way to isolate the no floating promises rule to this file and change the api to be non async? It will be less confusing that way, so it is explicit that the code does not need to be awaited

packages/website/ts/utils/utils.ts
```diff
@@ -313,14 +313,13 @@ export const utils = {
         const baseUrl = `https://${window.location.hostname}${hasPort ? `:${port}` : ''}`;
         return baseUrl;
     },
-    async onPageLoadAsync(): Promise<void> {
+    onPageLoadPromise: new Promise((resolve, _reject) => {
```
BMillman19: Hmm this is a bit inconsistent with how we normally handle promises, is this functionally different than creating a new promise each time? Was the main issue that resolve wasn't being called in the if statement?

fragosti: The issue is that you can't set `window.onload = resolve` multiple times (ie. for multiple promises, a new one created for each function call) because window.onload can only be one function and is only called once. So with the previous method here is what would happen:
```
window.onload = A
window.onload = B
window.onload = C
window.onload() // C is called
```
The Promises made for A and B are never resolves and await forever.  Imho this is the best way to do it.

BMillman19: I see, are we at risk of never resolving if we set `window.onload` too late?

fragosti: Previously, the only `onPageLoadAsync()` that worked was the last call, all previous Promises that were returned will never resolve because `window.onload` points to the latest resolve function. Calling `onPageLoadAsync()` after `window.onload` worked fine because we checked for that and returned.

packages/website/ts/components/forms/subscribe_form.tsx
```diff
@@ -112,6 +113,9 @@ export class SubscribeForm extends React.Component<SubscribeFormProps, Subscribe
         try {
             const response = await backendClient.subscribeToNewsletterAsync(this.state.emailText);
             const status = response.status === 200 ? SubscribeFormStatus.Success : SubscribeFormStatus.Error;
+            if (status === SubscribeFormStatus.Success) {
+                analytics.indentify(this.state.emailText, 'email');
```
BMillman19: typo? "identify"

fragosti: wow. Great catch

PR #877, opened 2018-07-14
Removed receiverData and `onReceive` callback from ERC721 proxy.

hysz:
## Description

In the current proxy implementation, an ERC721 taker has the opportunity to execute code on invalid state.

A transfer looks like this:
```
TraderA holds TokenA
TraderB holds TokenB

1. TraderA creates an order to trade TokenA for TokenB.
2. TraderB fills this order.

The settlement workflow looks like this:

#1
TokenA is transferred to TraderB.
TraderB holds **both** TokenA and TokenB.
TraderB.onReceived is called.

#2
TokenB is transferred to TraderA.
TraderA holds TokenB; TraderB holds TokenA.
TraderA.onReceived is called.
```

This could pose a problem if, for example, `TokenA` and `TokenB` were NFT’s that when combined allow you to perform some restricted operation. 

A more robust strategy could be to provide generalized callbacks that are executed after settlement, and which can be used with any proxy. Perhaps this is something we can look into for V3.

## Testing instructions

All existing tests pass. A new test was created to verify that the ERC721 `onReceived` callback is not getting called.

## Types of changes

## Checklist:


packages/contracts/src/2.0.0/protocol/AssetProxy/ERC721Proxy.sol
```diff
@@ -83,34 +83,26 @@ contract ERC721Proxy is
                 // WARNING: The ABIv2 specification allows additional padding between
                 //          the Params and Data section. This will result in a larger
                 //          offset to assetData.
-
```
hysz: Question: Do you think we should change the ID of this proxy?

It is currently:
```
bytes4 constant internal PROXY_ID = bytes4(keccak256("ERC721Token(address,uint256,bytes)"));
```

We could change this to:
```
bytes4 constant internal PROXY_ID = bytes4(keccak256("ERC721Token(address,uint256)"));
```

I've left it unchanged for now so that forwarding contracts can include the additional bytes, if they'd like to implement callbacks.

packages/contracts/test/asset_proxy/proxies.ts
```diff
@@ -252,7 +252,7 @@ describe('Asset Transfer Proxies', () => {
                 expect(newOwnerMakerAsset).to.be.bignumber.equal(takerAddress);
             });
 
-            it('should call onERC721Received when transferring to a smart contract without receiver data', async () => {
+            it('should not call onERC721Received when transferring to a smart contract', async () => {
```
hysz: Left one callback sanity check in place. Removed the others.

packages/contracts/src/2.0.0/protocol/AssetProxy/ERC721Proxy.sol
hysz: Yeah, I suppose in both cases the extra data would just be displayed as an unstructured byte array. I'll make the update.

hysz: * Updated

PR #892, opened 2018-07-18
Created LibAbiEncoder with `fillOrderNoThrow`

hysz:
## Description

Moved this code into a separate library so it can be used by the forwarding contract.

## Testing instructions

All existing tests pass.

## Checklist:

### DIFF
--- a/packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
@@ -23,6 +23,10 @@ import "./LibOrder.sol";
 
 
 contract LibAbiEncoder {
+
+    // solhint-disable max-line-length
+    bytes4 constant internal FILL_ORDER_SELECTOR = bytes4(keccak256("fillOrder((address,address,address,address,uint256,uint256,uint256,uint256,uint256,uint256,bytes,bytes),uint256,bytes)"));
+    // solhint-enable max-line-length
    
     /// @dev ABI encodes calldata for `fillOrder` in memory and returns the address range.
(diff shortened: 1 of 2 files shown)


packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
```diff
@@ -0,0 +1,215 @@
...
+    /// @return calldataBegin Memory address of ABI encoded calldata.
+    /// @return calldataLength Lenfgth of ABI encoded calldata.
+    function abiEncodeFillOrder(
+        LibOrder.Order memory order,
+        uint256 takerAssetFillAmount,
+        bytes memory signature,
+        bytes4 fillOrderSelector
+    )
+        public
+        returns (
```
abandeali1: I think this is probably more reusable if it just returns a byte array (then you don't necessarily have to use assembly to make the delegatecall).

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
```diff
@@ -0,0 +1,215 @@
...
+    /// @param order Order struct containing order specifications.
+    /// @param takerAssetFillAmount Desired amount of takerAsset to sell.
+    /// @param signature Proof that order has been created by maker.
+    /// @return calldataBegin Memory address of ABI encoded calldata.
+    /// @return calldataLength Lenfgth of ABI encoded calldata.
+    function abiEncodeFillOrder(
+        LibOrder.Order memory order,
+        uint256 takerAssetFillAmount,
+        bytes memory signature,
+        bytes4 fillOrderSelector
```
abandeali1: Let's just hard code the selector into the function, like we do in the other areas we ABI encode data in assembly.

PR #910, opened 2018-07-24
[website] Use new designs with tooltips for allowance toggles

fragosti:
## Description

screen shot 2018-07-24 at 3 55 42 pm

https://0x.wake.com/96

The circular progress spinners are pretty annoying as they wobble. I made a new `Spinner` component that animates more or less like the material UI one but isn't on meth.

## Testing instructions

https://dogfood.0xproject.com/ and play around!

## Types of changes

## Checklist:


packages/website/ts/components/inputs/allowance_state_toggle.tsx
```diff
@@ -0,0 +1,158 @@
...
+    }
+    constructor(props: AllowanceStateToggleProps) {
+        super(props);
+        const tokenState = props.tokenState;
+        this.state = {
+            allowanceState: AllowanceStateToggle._getAllowanceState(tokenState),
+            prevTokenState: tokenState,
+        };
+    }
+
```
BMillman19: nit: extra newline here

packages/website/ts/components/inputs/allowance_state_toggle.tsx
```diff
@@ -0,0 +1,158 @@
...
+            });
+        }
+    }
+    private _getTooltipContent(): React.ReactNode {
+        const symbol = this.props.token.symbol;
+        const isLockingToken = this.props.tokenState.allowance.gt(0);
+        switch (this.state.allowanceState) {
+            case AllowanceState.Loading:
+                return (
+                    <Text noWrap={true} fontColor="white">
```
BMillman19: nit: `colors.white` for consistency

packages/website/ts/components/inputs/allowance_state_toggle.tsx
```diff
@@ -0,0 +1,158 @@
...
+            const tokenState = nextProps.tokenState;
+            this.setState({
+                prevTokenState: tokenState,
+                allowanceState: AllowanceStateToggle._getAllowanceState(nextTokenState),
+            });
+        }
+    }
+    private _getTooltipContent(): React.ReactNode {
+        const symbol = this.props.token.symbol;
+        const isLockingToken = this.props.tokenState.allowance.gt(0);
```
BMillman19: can we put this in the Loading case or AND it with `this.state.allowanceState === AllowanceState.Loading`

BMillman19: I think we need to handle the case when the tokenState is loading its initial state, when we are doing the initial load the tooltip incorrectly displays "locking/unlocking X" when really we just are waiting to read the state from the blockchain

packages/website/ts/components/ui/allowance_state_view.tsx
```diff
@@ -0,0 +1,50 @@
...
+                <Container position="relative" top="3px" left="5px">
+                    <Spinner size={18} strokeSize={2} />
+                </Container>
+            );
+        default:
+            return null;
+    }
+};
+
+const renderCheck = (color: string = '#37D400') => (
```
BMillman19: can we put this color in the shared colors file?

packages/website/ts/components/inputs/allowance_state_toggle.tsx
fragosti: this is a tslint rule :P

PR #911, opened 2018-07-25
Update/Fix V2-beta-testnet migration & Use it in all RC packages

fabioberger:
## Description

This PR:
- Updates the v2-beta-testnet migration script given additional changes to the smart contracts
- Currently, the RC packages were still referencing artifacts without Kovan addresses, this PR makes it such that they all copy the artifacts from the migrations package, and more specifically from `v2-beta-testnet` if possible.
- Modifies the WETH9 contract so that it's event arg names match up with the suggested names of the ERC20 spec. This allows us to present our library users with a cohesive interface for event args, and hides the Hungarian notation names from them.
- Adds additional migration instructions to migrations README.

## Checklist:


packages/contracts/src/2.0.0/tokens/EtherToken/WETH9.sol
```diff
@@ -22,10 +22,10 @@ contract WETH9 {
     string public symbol   = "WETH";
     uint8  public decimals = 18;
 
-    event  Approval(address indexed src, address indexed guy, uint wad);
-    event  Transfer(address indexed src, address indexed dst, uint wad);
-    event  Deposit(address indexed dst, uint wad);
-    event  Withdrawal(address indexed src, uint wad);
+    event  Approval(address indexed _owner, address indexed _spender, uint _value);
```
abandeali1: Rather than modifying these events, can we just use `IEtherToken.sol`? (looks like we need to add the `Deposit` and `Withdrawal` events to that, which should be there anyways)

fabioberger: Discussed offline, will be addressed in the future.

fabioberger: Discussed offline, will be addressed in a future PR.

PR #914, opened 2018-07-26
0x.js Return signature as a hex encoded string

dekz:
## Description

Return signature as a hex encoded string with the signature type appended.

[Compiled a list of signers](https://docs.google.com/spreadsheets/d/1SDLSd9s0pbl5dXwjv2BjtAtnwY_kmgVuFj0mOd2ZaAk/edit#gid=0), their prefixes and their hex encoding of v,r,s values. The `eth_sign` json rpc example returns v,r,s and this is dominant in the space (geth, metamask, ethereumjs-utils) so we now check for this first when serialising for 0x contract use (v,r,s,signatureType).

[Started a discussion on the Ethereum Magicians forum](https://ethereum-magicians.org/t/scheme-for-off-chain-signatures-and-encoding-for-on-chain-verification/999) to define this in a more explicit way.

## Checklist:


packages/0x.js/src/0x.ts
```diff
@@ -238,19 +244,19 @@ export class ZeroEx {
      * @param   orderHash       Hex encoded orderHash to sign.
      * @param   signerAddress   The hex encoded Ethereum address you wish to sign it with. This address
      *          must be available via the Provider supplied to 0x.js.
-     * @param   MessagePrefixOpts  Options regarding the desired prefix and whether to add it before calling `eth_sign`
-     * @return  An object containing the Elliptic curve signature parameters generated by signing the orderHash.
+     * @param   SignerProviderType  The type of Signer Provider which implements `eth_sign`. E.g Metamask, Ledger, Trezor or EthSign.
```
fabioberger: This is a bit confusing, how can `EthSign` be a signer provider type? Maybe we just call it `Other`? There are definitively other signers out there. We can also add a note to this `@param` that `Other` assumes the prefix is added by the signer, and uses the prefix specified in the Ethereum wiki docs.

fabioberger: I'm also not a fan of the naming `SignerProviderType`, why not just `BackingSigner`? Or `SignerType`?

packages/order-utils/CHANGELOG.json
```diff
@@ -13,6 +13,9 @@
         "changes": [
             {
                 "note": "Dependencies updated"
+            },
+            {
+                "note": "Update ecSignOrderHashAsync to return signature string with signature type byte"
```
fabioberger: PR # and also move up the a new changelog entry.

packages/order-utils/src/signature_utils.ts
```diff
@@ -191,22 +191,22 @@ export function isValidECSignature(data: string, signature: ECSignature, signerA
...
 
 /**
- * Signs an orderHash and returns it's elliptic curve signature.
+ * Signs an orderHash and returns it's elliptic curve signature and signature type.
  * This method currently supports TestRPC, Geth and Parity above and below V1.6.6
  * @param   orderHash       Hex encoded orderHash to sign.
  * @param   signerAddress   The hex encoded Ethereum address you wish to sign it with. This address
  *          must be available via the Provider supplied to 0x.js.
- * @param   hashPrefixOpts Different signers add/require different prefixes be appended to the message being signed.
+ * @param   messagePrefixOpts Different signers add/require different prefixes be prepended to the message being signed.
```
fabioberger: Rename `messagePrefixOpts` here.

packages/order-utils/src/signature_utils.ts
```diff
@@ -231,51 +232,80 @@ export async function ecSignOrderHashAsync(
...
         }
     }
 
     throw new Error(OrderError.InvalidSignature);
 }
-
+/**
+ * Combines ECSignature with V,R,S and the relevant signature type for use in 0x protocol
+ * @param ecSignature The ECSignature of the signed data
+ * @param messagePrefixType The MessagePrefixType of the signed data
```
fabioberger: Update this comment to new param.

packages/order-utils/src/signature_utils.ts
```diff
@@ -231,51 +232,80 @@ export async function ecSignOrderHashAsync(
...
+ */
+export function convertToSignatureWithType(signature: string, type: SignatureType): string {
+    const signatureBuffer = Buffer.concat([ethUtil.toBuffer(signature), ethUtil.toBuffer(type)]);
+    const signatureHex = `0x${signatureBuffer.toString('hex')}`;
+    return signatureHex;
+}
 /**
  * Adds the relevant prefix to the message being signed.
  * @param message Message to sign
  * @param messagePrefixType The type of message prefix to add. Different signers expect
```
fabioberger: Update comment.

packages/order-utils/test/signature_utils_test.ts
```diff
@@ -188,12 +182,57 @@ describe('Signature utils', () => {
...
-                shouldAddPrefixBeforeCallingEthSign: false,
+            const ecSignature = await ecSignOrderHashAsync(
+                fakeProvider,
+                orderHash,
+                makerAddress,
+                SignerProviderType.EthSign,
+            );
+            expect(ecSignature).to.equal(expectedSignature);
+        });
+        // Note this is due to a bug in Metamask where it does not prefix before signing, this is a known issue and is to be fixed in the future
```
fabioberger: Can we add a `// Source: URL` here to the issue or PR, so future dev's can check on it's status?

packages/types/src/index.ts
```diff
@@ -142,6 +142,16 @@ export enum SignatureType {
...
 }
 
+/**
+ * The Signer Provider Type. Some Signer implementations use different message prefixes or implement different
+ * eth_sign behaviour. Note EthSign is compatible with the Ledger device.
+ */
+export enum SignerProviderType {
+    EthSign = 'ETH_SIGN',
+    Metamask = 'METAMASK',
+    Trezor = 'TREZOR',
```
fabioberger: I think we should still include `Ledger` despite it being identical under-the-hood. Might give developers more certainty it'll work as expected when using Ledger?

Also, let's rename `EthSign` to `Other`, since it's not a signer name, and there are many more Signers out there but we should assume they work to-spec until someone raises an issue otherwise.

dekz: Yeah I had Ledger here as well originally to signal the developer.  I'm not a huge fan of other since it is not that intuitive. EthSign is aligned with our SignatureTypes in contracts. Once Metamask fixes their bug the only availble options would be `Trezor` and `Other`

packages/0x.js/src/0x.ts
dekz: Any implementation which implements the EthSign as per the spec is an EthSign provider (regardless of where the key lives and signing process happens). Anything else is out of spec to EthSign and should be separately called out. Eventually we would only have two options, Spec Compliant (EthSign, as in eth_sign) and Trezor. 

There are definitely other signers out there, but they would either be spec compliant (EthSign) or we would flag them as their own SignerProviderType to implement the workarounds.

`BackingSigner` sounds like it should have available options for `PrivateKey` and `Mnemonic` which doesn't make much sense here.

fabioberger: I think we are on the same page but speaking past each other a bit. I'm proposing we do:

```
* @param   SignerType  The type of signer that will execute `eth_sign`. E.g Metamask, Trezor or Other. (note: `Other` will act in a `eth_sign` spec-compliant manner)
```

This makes it clear that `Metamask` and `Trezor` are the anomalies here and everything else can just pass in `Other` here.

Re: Knowing which are r,s,v/v,r,s, at one point Parity & Geth differed here. This might have been fixed at some point, but we kept it to remain backward-compatible. We could probably remove this at some point too.

packages/order-utils/src/signature_utils.ts
```diff
@@ -225,57 +225,96 @@ export async function ecSignOrderHashAsync(
...
+        case SignerType.Ledger:
+        case SignerType.Default: {
+            const signatureType = SignatureType.EthSign;
+            const signatureWithType = convertToSignatureWithType(signatureHex, signatureType);
+            return signatureWithType;
+        }
+        case SignerType.Trezor: {
+            const signatureType = SignatureType.Trezor;
+            const signatureWithType = convertToSignatureWithType(signatureHex, signatureType);
+            return signatureWithType;
```
fabioberger: Rather then return, lets move the body of `convertToSignatureWithType` below the switch statement.

packages/0x.js/src/0x.ts
```diff
@@ -238,19 +237,21 @@ export class ZeroEx {
      * @param   orderHash       Hex encoded orderHash to sign.
      * @param   signerAddress   The hex encoded Ethereum address you wish to sign it with. This address
      *          must be available via the Provider supplied to 0x.js.
-     * @param   MessagePrefixOpts  Options regarding the desired prefix and whether to add it before calling `eth_sign`
-     * @return  An object containing the Elliptic curve signature parameters generated by signing the orderHash.
+     * @param   signerType the type signer that will perform the `eth_sign` operation. E.g Default, Metamask, Ledger or Trezor.
```
fabioberger: `the type signer that` -> `the signer type that`

PR #915, opened 2018-07-26
Add strictArgumentEncodingCheck to BaseContract and use it in contract templates

albrow:
## Description

This PR adds a new static method `strictArgumentEncodingCheck` to `BaseContract`. This method is called in `callAsync` and `sendTransactionAsync` in all contract templates, and its purpose is to protect against encoding-related bugs. It does this by performing a simple (on the surface at least) check:

1. First we encode the arguments we are about to send.
2. Then we decode them.
3. We compare the original arguments to the decoded ones to make sure nothing has changed.

After I added this strict check, I also updated some of our tests as needed. There were some cases where we were relying on implicit padding or type coercion by the encoder. In such cases, we are now more exact about what arguments we pass through to smart contracts.

## Testing instructions

Just run `yarn test`.

## Types of changes

## Checklist:

### DIFF
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -95,7 +95,11 @@ export class BaseContract {
             const original = args[i];
             const decoded = rawDecoded[i];
             if (!abiUtils.isAbiDataEqual(params.names[i], params.types[i], original, decoded)) {
-                throw new Error(`Cannot safely encode argument: ${params.names[i]} (${original}) of type ${params.types[i]}. (Possible type overflow or other encoding error)`);
+                throw new Error(
+                    `Cannot safely encode argument: ${params.names[i]} (${original}) of type ${
+                        params.types[i]
+                    }. (Possible type overflow or other encoding error)`,
+                );
             }
         }
     }
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -82,6 +82,23 @@ export class BaseContract {
         }
         return txDataWithDefaults;
     }
+    // Throws if the given arguments cannot be safely/correctly encoded based on
+    // the given inputAbi. An argument may not be considered safely encodeable
+    // if it overflows the corresponding Solidity type, there is a bug in the
+    // encoder, or the encoder performs unsafe type coercion.
+    public static strictArgumentEncodingCheck(inputAbi: DataItem[], args: any[]): void {
+        const coder = (ethers as any).utils.AbiCoder.defaultCoder;
+        const params = abiUtils.parseEthersParams(inputAbi);
+        const rawEncoded = coder.encode(params.names, params.types, args);
+        const rawDecoded = coder.decode(params.names, params.types, rawEncoded);
+        for (let i = 0; i < rawDecoded.length; i++) {
+            const original = args[i];
+            const decoded = rawDecoded[i];
+            if (!abiUtils.isAbiDataEqual(params.names[i], params.types[i], original, decoded)) {
+                throw new Error(`Cannot safely encode argument: ${params.names[i]} (${original}) of type ${params.types[i]}. (Possible type overflow or other encoding error)`);
+            }
+        }
+    }
     protected _lookupEthersInterface(functionSignature: string): ethers.Interface {
         const ethersInterface = this._ethersInterfacesByFunctionSignature[functionSignature];
         if (_.isUndefined(ethersInterface)) {
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -87,7 +87,7 @@ export class BaseContract {
     // if it overflows the corresponding Solidity type, there is a bug in the
     // encoder, or the encoder performs unsafe type coercion.
     public static strictArgumentEncodingCheck(inputAbi: DataItem[], args: any[]): void {
-        const coder = (ethers as any).utils.AbiCoder.defaultCoder;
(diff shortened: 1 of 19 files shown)


packages/utils/src/abi_utils.ts
```diff
@@ -1,7 +1,166 @@
...
+        for (let i = 0; i < x.length; i++) {
+            if (!isAbiDataEqual(name, newType, x[i], y[i])) {
+                return false;
+            }
+        }
+        return true;
+    }
+    if (_.startsWith(type, 'tuple(')) {
+        if (_.isString(name)) {
+            throw new Error('Internal error: type was tuple but names was a string');
```
albrow: There are a handful of these sanity checks in this PR. In general, I like to fail fast when something unexpected happens instead of letting things fail later on. Let me know if there is a better/more conventional way to do this.

packages/utils/src/abi_utils.ts
```diff
@@ -1,7 +1,166 @@
...
+        } else if (_.isNull(name)) {
+            throw new Error('Internal error: type was tuple but names was a null');
```
fabioberger: When would we expect this to happen? Can we print out some identifying input that'll make it easier for them to spot the culprit?

packages/utils/src/abi_utils.ts
```diff
@@ -1,7 +1,166 @@
...
+        }
+        // For tuples, we iterate through the underlying values and check each
+        // one individually.
+        const types = splitTupleTypes(type);
+        if (types.length !== name.names.length) {
+            throw new Error(
+                `Internal error: parameter types/names length mismatch (${types.length} != ${name.names.length})`,
```
fabioberger: What other error prefixes should we use? And when exactly is coming considered an `Internal error`?

packages/utils/src/abi_utils.ts
```diff
@@ -1,7 +1,166 @@
...
+                ? (name.names[i] as string)
+                : ((name.names[i] as EthersNestedParamName).name as string);
+            if (!isAbiDataEqual(name.names[i], types[i], x[nestedName], y[nestedName])) {
+                return false;
+            }
+        }
+        return true;
+    } else if (type === 'address' || type === 'bytes') {
+        // HACK(albrow): ethers.js sometimes changes the case of addresses/bytes
+        // when decoding/encoding. To account for that, we convert to lowercase
```
fabioberger: Do you mean that ethers.js is modifying by reference? If so, could you submit an issue and add the source link here? That way, once it's fixed, we could remove this logic.

packages/base-contract/src/index.ts
```diff
@@ -82,6 +82,27 @@ export class BaseContract {
         }
         return txDataWithDefaults;
     }
+    // Throws if the given arguments cannot be safely/correctly encoded based on
+    // the given inputAbi. An argument may not be considered safely encodeable
+    // if it overflows the corresponding Solidity type, there is a bug in the
+    // encoder, or the encoder performs unsafe type coercion.
+    public static strictArgumentEncodingCheck(inputAbi: DataItem[], args: any[]): void {
+        const coder = (ethers as any).utils.AbiCoder.defaultCoder;
```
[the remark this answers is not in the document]
albrow: Yes we should fix/update the type definitions. I think we should do that separately instead of waiting for the new type defs to get released.

packages/utils/src/abi_utils.ts
albrow: This is what I was asking when I mentioned error-handling conventions. To me, an "internal error" is due to a bug in the implementation itself as opposed to invalid input or network/database errors.

I think we should standardize the kinds of error messages we return in our libraries, possibly by including prefixes, but that lies outside of the scope of this PR. Is it okay to leave this for now and we can circle back on it in the future? Or should I change this error message right now to make it more clear?

packages/base-contract/src/index.ts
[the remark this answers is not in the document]
albrow: Actually after looking into this more, we're still using @0xproject/typescript-typings for the type definitions for ethers. I added the type definitions there and removed the use of `any` here. I will also create an Asana task for either moving @0xproject/typescript-typings/ethers to DefinitelyTyped or upgrading ethers to 4.0 which is the TypeScript version.

packages/utils/src/abi_utils.ts
fabioberger: Yep, that's fine. I'm down to standardize them though, would be great if you could take the lead on that.

packages/utils/src/abi_utils.ts
fabioberger: I see. Yep, I'd just mention that Ethers always returns checksummed addresses, even when initially supplied a non-checksummed address.

packages/utils/package.json
```diff
@@ -5,12 +5,13 @@
...
     "description": "0x TS utils",
-    "main": "lib/index.js",
-    "types": "lib/index.d.ts",
+    "main": "lib/src/index.js",
+    "types": "lib/src/index.d.ts",
     "scripts": {
         "watch_without_deps": "tsc -w",
         "build": "tsc && copyfiles -u 2 './lib/monorepo_scripts/**/*' ./scripts",
         "clean": "shx rm -rf lib scripts",
+        "test": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --bail --exit",
```
LogvinovLeon: Add a command to run tests with coverage, run tests on CI. Also add those tests to be runned on CI. Look at other packages for an example.

PR #924, opened 2018-07-29
Improved Docs & Removed monorepo-scripts from packages

fabioberger:
## Description

Currently `monorepo-scripts` is a dependency of every public package. The reason? We are using Lerna's `postpublish.js` script hook to generate/upload docs and publish release notes to Github. This PR changes that.

As the number of packages published in the repo has grown, it has become untenable to have an individual release note for each package. This PR therefore aggregates them into a single release note that looks like:

```
### 0x.js

- Fixed a tiny, inconsequential bug (#789)
- Add a `findALover` method to `ZeroEx` (#793)

### web3-wrapper

- Removed unused param from `getBlockAsync` (#794)

...and so on for each package
```

Additionally the doc generation/upload step has been moved to the `publish.ts` script and happens after the packages are published. Importantly, the way we generate the doc JSON representations has changed. It now does the following:

- Find all exports in packages `index.ts`
- Find the files for each export (if it's internal or within another monorepo package)
- Call Typedoc with list of files containing the public interface of the package
- Go through the resulting Typedoc JSON, and remove all exported elements not exported in the main package's `index.ts`.
- Add entries for any external packages re-exported by our package (previously the docs simply omitted these). 

These steps have resulted in our doc JSON files being 3X smaller then they used to be! 🍾 🎆 

Additionally, we've gotten rid of a lot of package specific configurations using the above procedure. This PR also adds checks to make sure that any types that are present in the package's public interface **are exported** by it's `index.ts` file. It will even complain if the package exports superfluous type declarations. These checks now happen in every CircleCI run.

In order to make all package's pass these new constraints, I had to modify their `index.ts`. Some larger changes were also needed, like moving types from `sol-compiler` to `types` that are used by many packages now. 

Miscelaneous fixes:
- Run `yarn install` twice in CircleCI since I was encountering the same issue as locally.
- Modified the public interface of 0x.js so that it simply re-exports it's sub-packages.
- Added OrderWatcher back into 0x.js
- Fixed a bunch of TSLint errors from across many unchanged packages that suddenly popped up.
- Fixed scroll offset issue

## Testing instructions

Run the tests. Render the doc pages.

## Checklist:


.circleci/config.yml
```diff
@@ -17,6 +17,9 @@ jobs:
             - yarn-packages-{{ .Branch }}
             - yarn-packages-master
             - yarn-packages-
+      - run:
+          name: yarn
+          command: yarn --frozen-lockfile install || true
```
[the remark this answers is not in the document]
fabioberger: So, I ran into the same issue on CI as we've been having locally with `yarn`, where the first `yarn install` fails, and the second succeeds. Running `yarn install` twice on CI fixed the issue. We ignore any errors generated from the first call, because it is likely to fail, and is then called again. If the first failed for another reason then the quirk  we are experiencing, we can safely assume that the second time we call `yarn install`, it will fail again and the build will fail.

packages/react-shared/src/utils/utils.ts
```diff
@@ -33,6 +33,27 @@ export const utils = {
...
+            if (isNumber && (i === 0 || isPrevNumber)) {
+                return char;
+            }
+            if (char === char.toUpperCase() && i !== 0) {
+                return ` ${char}`;
+            }
+            return char;
+        });
+        let finalText = charArray.join('');
+        const exceptions = { 'EIP ': 'E I P', 'ZRX ': 'Z R X', 'ERC ': 'E R C', RPC: 'R P C' };
```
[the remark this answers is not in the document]
fabioberger: I replaced my overly complex code with:

```
import changeCase = require('change-case');

changeCase.camelCase(changeCase.snake('ERC20Wrapper'))
```

Which perfectly deals with lowercasing any capitalised acronym as well as properly handles numbers in variable names.

fabioberger:
@albrow when I created the `react-docs-example` we had just separated out the docs rendering code from the website and were under the impression that others would want to use it. This didn't pan out. Additionally, I just don't feel confident given this most recent refactor that this package is ready for prime-time. We can always add it back once it is, but right now, I'd rather not give the impression that it's ready.

albrow:
@fabioberger 👍 sounds good to me. I'll take any chance we can get to reduce maintenance burden right now.

packages/monorepo-scripts/src/utils/doc_generate_and_upload_utils.ts
```diff
@@ -0,0 +1,485 @@
...
+            if (_.isArray(nodeValue)) {
+                _.each(nodeValue, aNode => {
+                    updatedReferenceNames = DocGenerateAndUploadUtils._getAllReferenceNames(
+                        innerPropertyName,
+                        aNode,
+                        updatedReferenceNames,
+                    );
+                });
+            } else if (_.isObject(nodeValue)) {
+                updatedReferenceNames = updatedReferenceNames = DocGenerateAndUploadUtils._getAllReferenceNames(
```
fragosti: Woah this compiles?

packages/monorepo-scripts/src/utils/doc_generate_and_upload_utils.ts
```diff
@@ -0,0 +1,485 @@
...
+            const linkIfExists = docGenConfigs.EXTERNAL_EXPORT_TO_LINK[externalExport];
+            if (_.isUndefined(linkIfExists)) {
+                externalExportsWithoutLinks.push(externalExport);
+                return;
+            }
+            externalExportToLink[externalExport] = linkIfExists;
+        });
+        if (!_.isEmpty(externalExportsWithoutLinks)) {
+            throw new Error(
+                `Found the following external exports in ${
```
fragosti: Nice.

packages/monorepo-scripts/src/utils/github_release_utils.ts
```diff
@@ -0,0 +1,104 @@
...
+        constants.monorepoRootPath,
+        'packages',
+        packageNameWithoutNamespace,
+        'CHANGELOG.json',
+    );
+    const changelogJSON = readFileSync(changelogJSONPath, 'utf-8');
+    const changelogs = JSON.parse(changelogJSON);
+    const latestLog = changelogs[0];
+    // If only has a `Dependencies updated` changelog, we don't include it in release notes
+    if (latestLog.changes.length === 1 && latestLog.changes[0].note === constants.dependenciesUpdatedMessage) {
```
fragosti: Should we make this more explicit and put a `shouldOmitFromReleaseNotes` boolean in `CHANGELOG.json`?

packages/monorepo-scripts/src/utils/utils.ts
```diff
@@ -48,7 +48,7 @@ export const utils = {
                     };
                     packages.push(pkg);
                 } catch (err) {
-                    utils.log(`Couldn't find a 'package.json' for ${subpackageName}. Skipping.`);
+                    // Couldn't find a 'package.json' for package. Skipping.
```
fragosti: Too noisy?

packages/react-docs-example/public/css/basscss_responsive_type_scale.css
```diff
@@ -1,84 +0,0 @@
-/* Basscss Responsive Type Scale */
-/* Modified by Fabio Berger to include xs prefix */
```
fragosti: Why delete all these files?

packages/website/translations/chinese.json
```diff
@@ -13,7 +13,7 @@
     "TRADITIONAL_ASSETS": "传统资产",
     "DIGITAL_GOODS": "数字商品",
     "OFFCHAIN_ORDER_RELAY": "链下订单中继",
-    "OONCHAIN_SETTLEMENT": "链上最终结算",
```
fragosti: Was wondering about this lol

packages/monorepo-scripts/src/utils/utils.ts
fabioberger: Yeah, too noisy and very unlikely to be an issue.

packages/react-docs-example/public/css/basscss_responsive_type_scale.css
fabioberger: when I created the `react-docs-example` package, we had just separated out the docs rendering code from the website and were under the impression that others would want to use it. This didn't pan out. Additionally, I just don't feel confident given this most recent refactor that this package is ready for prime-time. We can always add it back once it is, but right now, I'd rather not give the impression that it's ready.

packages/website/translations/chinese.json
fabioberger: Lol, silly typo.

PR #925, opened 2018-07-30
[contracts] Fix comments, styling and small optimization for hashOrder

abandeali1:
## Description

- Fixes some misleading and unclear comments/styling
- Adds small optimization + readability improvement to `hashOrder`

## Checklist:

### DIFF
--- a/packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
@@ -123,10 +123,10 @@ contract MixinTransactions is
         bytes32 dataHash = keccak256(data);
 
         // Assembly for more efficiently computing:
-        // keccak256(abi.encode(
+        // keccak256(abi.encodePacked(
         //     EIP712_ZEROEX_TRANSACTION_SCHEMA_HASH,
         //     salt,
-        //     signerAddress,
+        //     bytes32(signerAddress),
         //     keccak256(data)
         // ));
 
--- a/packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
@@ -131,11 +131,15 @@ contract MixinTransactions is
         // ));
 
         assembly {
+            // Load free memory pointer
             let memPtr := mload(64)
-            mstore(memPtr, schemaHash)
-            mstore(add(memPtr, 32), salt)
-            mstore(add(memPtr, 64), and(signerAddress, 0xffffffffffffffffffffffffffffffffffffffff))
-            mstore(add(memPtr, 96), dataHash)
+
+            mstore(memPtr, schemaHash)                                                               // hash of schema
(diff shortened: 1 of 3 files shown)


packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
```diff
@@ -123,20 +123,27 @@ contract MixinTransactions is
         bytes32 dataHash = keccak256(data);
 
         // Assembly for more efficiently computing:
-        // keccak256(abi.encode(
+        // keccak256(abi.encodePacked(
         //     EIP712_ZEROEX_TRANSACTION_SCHEMA_HASH,
         //     salt,
-        //     signerAddress,
+        //     bytes32(signerAddress),
```
Recmo: Is this a matter of style or is there a performance difference? Both should work fine.

packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
```diff
@@ -123,20 +123,27 @@ contract MixinTransactions is
...
+            mstore(memPtr, schemaHash)                                                               // hash of schema
+            mstore(add(memPtr, 32), salt)                                                            // salt
+            mstore(add(memPtr, 64), and(signerAddress, 0xffffffffffffffffffffffffffffffffffffffff))  // signerAddress
+            mstore(add(memPtr, 96), dataHash)                                                        // hash of data
+
+            // Compute hash
             result := keccak256(memPtr, 128)
+
+            // Increment free memory pointer
+            mstore(64, add(memPtr, 128))
```
Recmo: Wasteful. The allocated memory does not leave the scope, it should be discarded after the `keccak256` line. Increasing the `memptr` leaves it in memory, increasing max memory usages and thus gas cost.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibEIP712.sol
```diff
@@ -59,8 +59,31 @@ contract LibEIP712 {
...
+
+            mstore(memPtr, 0x1901000000000000000000000000000000000000000000000000000000000000)  // EIP191 header
+            mstore(add(memPtr, 2), eip712DomainHash)                                            // EIP712 domain hash
+            mstore(add(memPtr, 34), hashStruct)                                                 // Hash of struct
+
+            // Compute hash
+            result := keccak256(memPtr, 66)
+
+            // Increment free memory pointer
+            mstore(64, add(memPtr, 66))
```
Recmo: Same.

packages/contracts/src/2.0.0/protocol/Exchange/MixinTransactions.sol
abandeali1: Mostly style. I just think there's less cognitive overhead in "pad everything to 32 bytes" than "ABI encode". If you already know what ABI encoding is, it's pretty obvious that you can use it to accomplish this. If you don't know what ABI encoding is, it's not obvious why we are doing it at all.

PR #933, opened 2018-07-31
[contracts] Refactor token implementations

abandeali1:
## Description

- Refactors and optimizes token implementations (mostly ERC721)
- Adds tests for ERC20 tokens that don't return a value
- Adds tests for ERC721 tokens

## Checklist:


packages/contracts/src/2.0.0/tokens/ERC721Token/MintableERC721Token.sol
```diff
@@ -0,0 +1,83 @@
...
+
+contract MintableERC721Token is
+    ERC721Token
+{
+
+    /// @dev Function to mint a new token
+    ///      Reverts if the given token ID already exists
+    /// @param _to Address of the beneficiary that will own the minted token
+    /// @param _tokenId ID of the token to be minted by the msg.sender    
+    function _mint(address _to, uint256 _tokenId)
```
dekz: can we expose a public mintable function for testnets?

abandeali1: We have `DummyERC721Token`. Maybe I should just drop the `onlyOwner` modifiers?

abandeali1: Going to remove `onlyOwner` from `mint` but leave it on `burn`.

PR #934, opened 2018-07-31
Initial forwarder contract wrapper

BMillman19:
## Description

* Adds a new `ForwarderWrapper` class to the `contract-wrappers` package
* Only implements the `marketSellOrdersWithEthAsync` and `marketBuyOrdersWithEthAsync` transactions
* Provides many assertions for type safety and correct orders
* Adds the `getOrdersInfoAsync` method to `ExchangeWrapper` with accompanying test

## ToDo

* Convenience / utility methods for calculating how much ETH to send with the `marketBuyOrdersWithEthAsync` tx in order to cover the required `makerAssetFillAmount`
* Possible make the `ethAmount` parameter optional and use the convenience methods to estimate by default
* More assertions (do orders provided cover the specified makerAssetFillAmount, do the fee orders cover the required fees, are the order funded?)
* make sure fee percentage is the correct format
* order optimizations (NULL out certain order fields to save on gas, this is safe because the contract makes some assumptions, i.e. assume that all takerAssetData for the input orders is for ether token)

## Testing instructions

```bash
yarn install && PKG=@0xproject/contract-wrappers yarn build
cd packages/contract-wrappers
yarn test
```

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
+++ b/packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
@@ -22,9 +22,16 @@ export class ForwarderWrapper extends ContractWrapper {
     public abi: ContractAbi = artifacts.Forwarder.compilerOutput.abi;
     private _forwarderContractIfExists?: ForwarderContract;
     private _contractAddressIfExists?: string;
-    constructor(web3Wrapper: Web3Wrapper, networkId: number, contractAddressIfExists?: string) {
+    private _zrxContractAddressIfExists?: string;
+    constructor(
+        web3Wrapper: Web3Wrapper,
+        networkId: number,
+        contractAddressIfExists?: string,
+        zrxContractAddressIfExists?: string,
+    ) {
         super(web3Wrapper, networkId);
         this._contractAddressIfExists = contractAddressIfExists;
+        this._zrxContractAddressIfExists = zrxContractAddressIfExists;
     }
     /**
      * Purchases as much of orders' makerAssets as possible by selling up to 95% of transaction's ETH value.
@@ -53,6 +60,7 @@ export class ForwarderWrapper extends ContractWrapper {
         feeRecipientAddress: string = constants.NULL_ADDRESS,
         txOpts: TransactionOpts = {},
     ): Promise<string> {
+        // type assertions
         assert.doesConformToSchema('signedOrders', signedOrders, schemas.signedOrdersSchema);
         await assert.isSenderAddressAsync('takerAddress', takerAddress, this._web3Wrapper);
         assert.isBigNumber('ethAmount', ethAmount);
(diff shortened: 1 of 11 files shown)


packages/contract-wrappers/src/types.ts
```diff
@@ -172,13 +174,13 @@ export enum TransferType {
...
 export interface OrderInfo {
-    orderStatus: number;
+    orderStatus: OrderStatus;
     orderHash: string;
     orderTakerAssetFilledAmount: BigNumber;
 }
 
 export enum OrderStatus {
-    INVALID,
+    INVALID = 0,
```
fragosti: Why this change?

packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
```diff
@@ -0,0 +1,209 @@
...
+        networkId: number,
+        contractAddressIfExists?: string,
+        zrxContractAddressIfExists?: string,
+    ) {
+        super(web3Wrapper, networkId);
+        this._contractAddressIfExists = contractAddressIfExists;
+        this._zrxContractAddressIfExists = zrxContractAddressIfExists;
+    }
+    /**
+     * Purchases as much of orders' makerAssets as possible by selling up to 95% of transaction's ETH value.
```
fragosti: nit: found the "Purchases" phrasing here a bit confusing. I like how the spec defines it which is "marketSellOrders calls fillOrder sequentially for each provided order and signature until the total takerAssetAmount has been sold by the taker. "

packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
```diff
@@ -0,0 +1,209 @@
...
+        assert.feeOrdersCanBeUsedForForwarderContract(
+            signedFeeOrders,
+            this.getZRXTokenAddress(),
+            this.getEtherTokenAddress(),
+        );
+        const normalizedTakerAddress = takerAddress.toLowerCase();
+        const normalizedFeeRecipientAddress = feeRecipientAddress.toLowerCase();
+        const forwarderContractInstance = await this._getForwarderContractAsync();
+        const txHash = await forwarderContractInstance.marketSellOrdersWithEth.sendTransactionAsync(
+            signedOrders,
```
abandeali1: We should optimize this calldata before sending the transaction. The forwarding contract makes a lot of assumptions about the `assetData` fields that allow us to pass in empty byte arrays and save on gas.

For `orders`, all of the `takerAssetData` fields and all but the first order's `makerAssetData` fields can be empty. For all of the `feeOrders`, all of the `makerAssetData` and `takerAssetData` fields can be empty.

packages/contract-wrappers/src/types.ts
BMillman19: I just like it being more explicit

BMillman19:
@dekz I see you bro: https://github.com/0xProject/0x-monorepo/pull/936

PR #936, opened 2018-08-02
ForwarderWrapper order optimizations

BMillman19:
## Description

* Because the forwarder contract makes assumptions about `makerAssetData` and `takerAssetData` of the input orders, we can convert them to null ('0x') values to save on gas.
* This adds a `marketOrderOptimizationUtils` object that contains a set of pure functions for making order optimizations w/o mutation or side effects.

## Testing instructions

```bash
yarn install && PKG=@0xproject/contract-wrappers yarn build
cd packages/contract-wrappers && yarn test
```

## Types of changes

## Checklist:


BMillman19:
@abandeali1 perhaps I can include these optimizations in the exchange_wrapper as well?

packages/contract-wrappers/CHANGELOG.json
```diff
@@ -5,6 +5,10 @@
             {
                 "note": "Add ForwarderWrapper",
                 "pr": 934
+            },
+            {
+                "note": "Optimize orders in ForwarderWrapper",
+                "pr": 935
```
BMillman19: wrong PR number

packages/order-utils/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "1.1.0-rc.2",
+        "changes": [
+            {
+                "note": "Added a synchronous `createOrder` method in `orderFactory`"
+            }
```
BMillman19: add PR number

packages/contract-wrappers/src/utils/market_orders_optimization_utils.ts
```diff
@@ -0,0 +1,44 @@
+import { SignedOrder } from '@0xproject/types';
+import * as _ from 'lodash';
+
+import { constants } from './constants';
+
+export const marketOrdersOptimizationUtils = {
```
abandeali1: I think this should be renamed to something like `calldataOptimizationUtils`, since there are functions that aren't market buys or sells that we will also eventually optimize (like `matchOrders`).

packages/contract-wrappers/src/utils/market_orders_optimization_utils.ts
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * Takes an array of orders and outputs an array of equivalent orders where all takerAssetData are '0x' and
+     * all makerAssetData are '0x' except for that of the first order, which retains its original value
+     * @param   orders         An array of SignedOrder objects
+     * @returns optimized orders
+     */
+    optimizeMarketOrders(orders: SignedOrder[]): SignedOrder[] {
```
abandeali1: This should be `optimizeForwarderOrders`. The actual market buy and sell functions within the exchange contract will be optimized in a slightly different way.

packages/contract-wrappers/src/utils/market_orders_optimization_utils.ts
```diff
@@ -0,0 +1,44 @@
...
+        );
+        return optimizedOrders;
+    },
+    /**
+     * Takes an array of orders and outputs an array of equivalent orders where all takerAssetData are '0x' and
+     * all makerAssetData are '0x'
+     * @param   orders         An array of SignedOrder objects
+     * @returns optimized orders
+     */
+    optimizeFeeOrders(orders: SignedOrder[]): SignedOrder[] {
```
abandeali1: `optimizeForwarderFeeOrders`

packages/order-utils/src/order_factory.ts
```diff
@@ -1,17 +1,17 @@
...
+import { constants } from './constants';
 import { orderHashUtils } from './order_hash';
 import { generatePseudoRandomSalt } from './salt';
 import { ecSignOrderHashAsync } from './signature_utils';
 import { MessagePrefixType } from './types';
 
 export const orderFactory = {
-    async createSignedOrderAsync(
-        provider: Provider,
+    createOrder(
```
abandeali1: This function doesn't feel like it's doing much to me. Why is `expirationTimeSecondsIfExists` the only optional parameter? You can definitely make a case for default `takerAddress`, `senderAddress`, `feeRecipientAddress`, and `salt` values, so only making `expirationTimeSeconds` optional feels random to me. IMO this function would be more useful if it took a partial order as input and returned a full order with sensible defaults.

BMillman19: I can make more of them optional, this PR just refactored the existing one slightly so that we can create a normal order w/o the signature

packages/contract-wrappers/CHANGELOG.json
```diff
@@ -1,10 +1,14 @@
 [
     {
-        "version": "1.1.0-rc.2",
+        "version": "1.0.1-rc.3",
```
BMillman19: @fabioberger not sure if I'm doing this rc version bumping correctly. If we're on an rc package do we just bump the rc version for any type change and keep the left side the same?

BMillman19: I accidentally bumped it to 1.1.0-rc.2 in a previous PR but since this hasn't been published yet its safe to bump down

BMillman19: Double checked with fabio offline, we're supposed to just bump the right side for any change while it is RC

PR #937, opened 2018-08-02
Add marketUtils object for assisting with market buy calculations

BMillman19:
## Description

* We need a way to take a set of orders and an amount of maker asset to fill as input and output a subset of orders that cover the fill amount
* We need a way to take a set of orders and find the total amount of fees required to fill them

## Testing instructions

```bash
yarn install && PKG=@0xproject/order-utils yarn build
cd packages/order-utils && yarn test
```

## Checklist:

### DIFF
--- a/packages/json-schemas/schemas/basic_type_schemas.ts
+++ b/packages/json-schemas/schemas/basic_type_schemas.ts
@@ -7,7 +7,7 @@ export const addressSchema = {
 export const hexSchema = {
     id: '/Hex',
     type: 'string',
-    pattern: '^0x([0-9a-f][0-9a-f])+$',
+    pattern: '^0x(([0-9a-f][0-9a-f])+)?$',
 };
 
 export const numberSchema = {
--- a/packages/order-utils/src/market_utils.ts
+++ b/packages/order-utils/src/market_utils.ts
@@ -1,0 +1,109 @@
+import { schemas } from '@0xproject/json-schemas';
+import { OrderRelevantState, SignedOrder } from '@0xproject/types';
+import { BigNumber } from '@0xproject/utils';
+import * as _ from 'lodash';
+
+import { assert } from './assert';
+import { constants } from './constants';
+
+export const marketUtils = {
+    /**
+     * Takes an array of orders and returns a subset of those orders that has enough makerAssetAmount (taking into account on-chain balances,
+     * allowances, and partial fills) in order to fill the input makerAssetFillAmount plus slippageBufferAmount. Iterates from first order to last.
+     * Sort the input by ascending rate in order to get the subset of orders that will cost the least ETH.
+     * @param   signedOrders         An array of objects that conform to the SignedOrder interface. All orders should specify the same makerAsset.
+     *                               All orders should specify WETH as the takerAsset.
+     * @param   orderStates          An array of objects corresponding to the signedOrders parameter that each contain on-chain state
+     *                               relevant to that order.
+     * @param   makerAssetFillAmount The amount of makerAsset desired to be filled.
+     * @param   slippageBufferAmount An additional amount makerAsset to be covered by the result in case of trade collisions or partial fills.
+     * @return  Resulting orders and remaining fill amount that could not be covered by the input.
+     */
+    findOrdersThatCoverMakerAssetFillAmount(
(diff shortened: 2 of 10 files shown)


packages/order-utils/src/market_utils.ts
```diff
@@ -0,0 +1,109 @@
...
+        return marketUtils.findOrdersThatCoverMakerAssetFillAmount(
+            signedFeeOrders,
+            feeOrderStates,
+            totalFeeAmount,
+            slippageBufferAmount,
+        );
+    },
+};
+
+const getMakerAssetAmountAvailable = (orderState: OrderRelevantState) => {
```
BMillman19: There is a case missing here where the maker does not have enough fees / fee allowance to cover the makerFees

BMillman19: @fabioberger @LogvinovLeon does `getOpenOrderRelevantStateAsync` in `OrderStateUtils` already  cover all these edge cases when calculating `remainingFillableMakerAssetAmount` it looks like it might, just want to double check

packages/json-schemas/schemas/basic_type_schemas.ts
```diff
@@ -7,7 +7,7 @@ export const addressSchema = {
 export const hexSchema = {
     id: '/Hex',
     type: 'string',
-    pattern: '^0x([0-9a-f][0-9a-f])+$',
+    pattern: '^0x(([0-9a-f][0-9a-f])+)?$',
```
BMillman19: This will now match `0x`

packages/order-utils/src/market_utils.ts
```diff
@@ -0,0 +1,108 @@
...
+     *                               All orders should specify WETH as the takerAsset.
+     * @param   orderStates          An array of objects corresponding to the signedOrders parameter that each contain on-chain state
+     *                               relevant to that order.
+     * @param   makerAssetFillAmount The amount of makerAsset desired to be filled.
+     * @param   slippageBufferAmount An additional amount makerAsset to be covered by the result in case of trade collisions or partial fills.
+     * @return  Resulting orders and remaining fill amount that could not be covered by the input.
+     */
+    findOrdersThatCoverMakerAssetFillAmount(
+        signedOrders: SignedOrder[],
+        orderStates: OrderRelevantState[],
```
BMillman19: We can possible change this param from `orderState: OrderRelevantState[]` to `remainingMakerAssetFillAmounts: BigNumber` but will require the developer to reach into OrderRelevantState that they are grabbing from `OrderStateUtils` instead of just dumping it all into here

packages/order-utils/src/market_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+     * @param   slippageBufferAmount                An additional amount of makerAsset to be covered by the result in case of trade collisions or partial fills.
+     * @return  Resulting orders and remaining fill amount that could not be covered by the input.
+     */
+    findOrdersThatCoverMakerAssetFillAmount(
+        signedOrders: SignedOrder[],
+        remainingFillableMakerAssetAmounts: BigNumber[],
+        makerAssetFillAmount: BigNumber,
+        slippageBufferAmount: BigNumber = constants.ZERO_AMOUNT,
+    ): { resultOrders: SignedOrder[]; remainingFillAmount: BigNumber } {
+        // type assertions
```
fabioberger: Don't feel this comment and the next are necessary. Can we remove?

packages/order-utils/src/market_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+            remainingFillableFeeAmounts,
+            totalFeeAmount,
+            slippageBufferAmount,
+        );
+        return {
+            resultOrders,
+            remainingFeeAmount: remainingFillAmount,
+        };
+        // TODO: add more orders here to cover rounding
+        // https://github.com/0xProject/0x-protocol-specification/blob/master/v2/forwarding-contract-specification.md#over-buying-zrx
```
fabioberger: Is this going to be addressed in a subsequent PR?

BMillman19: Yes! Ideally we would be able to embed asana links here...

PR #941, opened 2018-08-07
[contracts] Test fillOrder return values and abiEncodeFillOrder

abandeali1:
## Description

- Test `FillResults` and `abiEncodeFillOrder` within the combinatorial tests for `fillOrder`
- Increment the free memory pointer at the end of the assembly blocks in `dispatchTransferFrom` and `abiEncodeFillOrder`. It _shouldn't_ be an issue to overwrite this memory (which was happening previously), but feels like a good safety addition.

## Checklist:

### DIFF
--- a/packages/contracts/src/2.0.0/protocol/Exchange/MixinAssetProxyDispatcher.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/MixinAssetProxyDispatcher.sol
@@ -159,22 +159,18 @@ contract MixinAssetProxyDispatcher is
                 }
 
                 /////// Call `assetProxy.transferFrom` using the constructed calldata ///////
-                let inputLen := sub(cdEnd, cdStart)
                 let success := call(
                     gas,                    // forward all gas
                     assetProxy,             // call address of asset proxy
                     0,                      // don't send any ETH
                     cdStart,                // pointer to start of input
-                    inputLen,               // length of input  
+                    sub(cdEnd, cdStart),    // length of input  
                     cdStart,                // write output over input
                     512                     // reserve 512 bytes for output
                 )
                 if iszero(success) {
                     revert(cdStart, returndatasize())
                 }
-
-                // Increment free memory pointer
-                mstore(64, add(cdStart, inputLen))
             }
         }
     }
--- a/packages/contracts/src/2.0.0/protocol/Exchange/MixinAssetProxyDispatcher.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/MixinAssetProxyDispatcher.sol
@@ -159,18 +159,22 @@ contract MixinAssetProxyDispatcher is
                 }
 
                 /////// Call `assetProxy.transferFrom` using the constructed calldata ///////
+                let inputLen := sub(cdEnd, cdStart)
                 let success := call(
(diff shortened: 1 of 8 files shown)


abandeali1:
Looks like incrementing the free memory pointer adds ~100 gas per fill due to memory expansion. @Recmo @hysz thoughts on if this is a good tradeoff?

packages/contracts/src/2.0.0/protocol/Exchange/MixinAssetProxyDispatcher.sol
```diff
@@ -159,18 +159,22 @@ contract MixinAssetProxyDispatcher is
...
+                    inputLen,               // length of input  
                     cdStart,                // write output over input
                     512                     // reserve 512 bytes for output
                 )
                 if iszero(success) {
                     revert(cdStart, returndatasize())
                 }
+
+                // Increment free memory pointer
+                mstore(64, add(cdStart, inputLen))
```
Recmo: The memory is discarded within the asm block, so we don't need to increment the free memory pointer. (And if we do it should have been the max of `inputLen` and `returndatasize()` :) )

Note that Solidity does the same thing when it generates assembly for a call.

Gas cost will get *quadratically* worse for large batch orders.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
```diff
@@ -207,10 +204,11 @@ contract LibAbiEncoder {
...
             // Set length of calldata
-            mstore(
-                fillOrderCalldata,
-                sub(dataAreaEnd, add(fillOrderCalldata, 0x20))
-            )
+            let calldataLen := sub(dataAreaEnd, add(fillOrderCalldata, 0x20))
+            mstore(fillOrderCalldata, calldataLen)
+
+            // Increment free memory pointer
+            mstore(0x40, add(fillOrderCalldata, add(calldataLen, 0x20)))
```
Recmo: This looks fine, but `mstore(0x40, dataAreaEnd)` should also work.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibAbiEncoder.sol
```diff
@@ -207,10 +204,10 @@ contract LibAbiEncoder {
...
 
             // Set length of calldata
-            mstore(
-                fillOrderCalldata,
-                sub(dataAreaEnd, add(fillOrderCalldata, 0x20))
-            )
+            mstore(fillOrderCalldata, sub(dataAreaEnd, add(fillOrderCalldata, 0x20)))
+
+            // Increment free memory pointer
+            mstore(0x40, add(fillOrderCalldata, dataAreaEnd))
```
Recmo: This looks off. I think it should be `mstore(0x40, dataAreaEnd)`. After the signature loop, `dataAreaEnd` will point to the first byte beyond the last one written. Adding `fillOrderCalldata` would add two pointers, which is weird.

abandeali1: Whoops, that is most definitely a mistake

PR #943, opened 2018-08-07
[WIP] Update TypeScript to version ^2.8.0

albrow:
## Description

This PR changes the TypeScript version we use in the monorepo from 2.7 to 2.8. This upgrade was somewhat forced upon us because of a change made in a dependency. (The type definitions for react and prop-types in some part of the dependency tree are not compatible with version 2.7).

I discovered this by running __packages/monorepo-scripts/src/test_installation.ts__. It will fail on TypeScript version 2.7 but succeed if we use version 2.8.

## Testing instructions

Start verdaccio. Run `yarn run:publish:local` and then `yarn test:installation:local`.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
         "npm-run-all": "^4.1.2",
         "prettier": "^1.11.1",
         "source-map-support": "^0.5.6",
-        "wsrun": "^2.2.0"
+        "wsrun": "^2.2.0",
+        "typescript": "^2.8.0"
     },
     "resolutions": {
         "ethers": "0xproject/ethers.js#eip-838-reasons"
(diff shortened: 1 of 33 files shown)


albrow:
Looks like TSLint requires us to include typescript as a dependency for each package. I guess I'll update this PR to do this.

fabioberger:
Is this ready for a review @albrow? Saw your comment above but looks like you haven't added back in the Typescript dep.

ritave:
Just stumbled upon this PR - keep in mind that Typescript 2.9 has breaking changes concerning JSON importing and is a bit bigger change, might consider moving straight to 2.9?

albrow:
@fabioberger sorry I got pulled away for something else last night. Marked this PR as WIP and will finish it now.

albrow:
Superseded by https://github.com/0xProject/0x-monorepo/pull/945. I didn't realize that version `^2.8.0` gets resolved to `2.9.2`. So #945 specifies exactly version `2.9.2` and updates the branch name and PR description for clarity.

PR #944, opened 2018-08-08
[WIP] [sol-compiler] Support parallel compilation

feuGeneA:
## Description

Enable parallel compilation via Node.js Child Processes.  A call to `Compiler.compileAsync()` will `child_process.spawn()` a separate process for each source module indicated during `Compiler` construction, and then it will wait to return until all compilations have completed.

## Types of changes

## Checklist:

*   [x] Prefix the title of this PR with `[WIP]` if it is a work in progress.
*   [x] Prefix the title of this PR with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [x] Refactor to move solcjs invocation to behind a command-line interface, so that it can be spawned as a parallel process via Node's Child Process API.
*   [x] Change `Compiler` to actually spawn multiple parallel processes, and to wait for them all to complete before finishing a call to `compileAync()`.
*   [x] Fix bug with truncated JSON stdout from command-line utility.  [Check.](https://github.com/0xProject/0x-monorepo/pull/944#issuecomment-411964262)
*   [x] Consider other IPC methods, besides parsing stdin/stdout to and from JSON.  Some may be more performant or appropriate for our use case, and perhaps we could simplify our code.  For example, we may be able to eliminate much of our parsing and glue code by using Node's plain-old-js-object-driven message events API, or maybe there's an RPC technique that will allow the `Resolver` to stay in the parent process.  [Check.](https://github.com/0xProject/0x-monorepo/pull/944#issuecomment-412229684)
*   [x] Allay race condition when downloading `solcjs` executable.  Two parallel processes may try to download it to the same location at the same time.  Perhaps move `compilerBinFilename` initialization block from `cli_resolver_solc.ts` to just before its invocation in the Compiler class, so that downloading the executable happens in the single-threaded parent process rather than in the parallel child processes.
*   [x] Implement a configuration/option for degree of parallelism.  As it stands now, if your project has 50 contracts, then you'll get 50 parallel processes.  Perhaps a command line flag, akin to [GNU Make's -j](https://www.gnu.org/software/make/manual/html_node/Parallel.html).  [Check](https://github.com/0xProject/0x-monorepo/pull/944/commits/eb70b76ee7462ada8e71d00d7bb3d4c56dca6d9e).
*   [ ] Clean up `yargs` usage in cli_resolver_solc.ts, marking things required, etc.
*   [ ] Add tests to cover my changes, or decided that tests would be too impractical.
*   [ ] Updat documentation as needed.


feuGeneA:
Metacoin build failing in CI but succeeding in my local environment.  Work to do...

packages/sol-compiler/src/compiler.ts
```diff
@@ -1,26 +1,19 @@
...
-    NPMResolver,
-    RelativeFSResolver,
-    Resolver,
-    URLResolver,
-} from '@0xproject/sol-resolver';
-import { fetchAsync, logUtils } from '@0xproject/utils';
+import { NameResolver, Resolver } from '@0xproject/sol-resolver';
+import { logUtils } from '@0xproject/utils';
 import chalk from 'chalk';
+import * as child_process from 'child_process';
```
fabioberger: Let's import it as `childProcess` since that suites our naming conventions and is inconsequential given that we are already importing it using `* as`.

packages/sol-compiler/src/utils/solc_wrapper.ts
```diff
@@ -0,0 +1,63 @@
...
+        solcjs = await response.text();
+        fs.writeFileSync(compilerBinFilename, solcjs);
+    }
+    const solcInstance = solc.setupMethods(requireFromString(solcjs, compilerBinFilename));
+
+    const stdinBuffer = fs.readFileSync(0);
+    const resolver: Resolver = constructResolver(argv.contractsDir);
+    process.stdout.write(
+        solcInstance.compileStandardWrapper(stdinBuffer.toString(), importPath => {
+            const sourceCodeIfExists = resolver.resolve(importPath);
```
fabioberger: We should do an `_.isUndefined` check here. If a variable name ends with `IfExists`, it means it might be undefined. The subsequent indexing into `.source` could fail if it is indeed `undefined`.

feuGeneA:
Metacoin build passing now, but now 0x.js build failing in both CI and my local environment (at least it's consistent...).  Work to do...

feuGeneA:
There are actually a handful of different failures among different packages.  Right now I'm chasing down the one with the most glaring clue, and hoping that resolving that one will lead to resolution of some or all of the others.

The one I'm chasing down is in the build of the `contracts` package.  It appears that the stdout buffer coming from the new command-line utility is being maxed out.  (The JSON being given to compiler.ts is truncated, and its length exactly matches the length of the stdout buffer.)  I'm in the process of modifying compiler.ts to repeatedly drain that stdout buffer, until the stream is marked as closed.

I'm also exploring parallelism at the same time, as I'm having to change to a different node.js `child_process` API, for both the stream handling and the parallelism, so exploring both changes at once right now.

feuGeneA:
I've got an initial implementation of parallel compilation in place here, and it seems to be working well. 
 You can see it in action in the latest CircleCI build for this PR

However, there is still an outstanding issue in the build of the contracts package: It appears that the stdout buffer coming from the new command-line utility is being maxed out. (The JSON data coming into compiler.ts is truncated, and its length exactly matches the length of the stdout buffer, and running the `solc-wrapper` script directly, from the command line, yields the full output, roughly double the length of what compiler.ts receives.)

I had hoped that the problem would be solved when switching from `child_process.execFileAsync` to `child_process.spawn`, since with `spawn` one must provide a callback to receive batches of data, whereas with `execFileAsync` the data is given by the return value from the `execFileAsync()` call itself.  Unfortunately that hope didn't come to fruition.

Investigation continues...

feuGeneA:
Added two new checklist items to the original comment above:
* Clean up yargs usage in solc_wrapper.ts, marking things required, etc.
* Allay race condition when downloading solcjs executable.  Two parallel processes may try to download it to the same location at the same time.

feuGeneA:
Added a checklist item to the original comment above:
* Implement a configuration/option for degree of parallelism. As it stands now, if your project has 50 contracts, then you'll get 50 parallel processes. Perhaps a command line flag, akin to [GNU Make's -j](https://www.gnu.org/software/make/manual/html_node/Parallel.html).

feuGeneA:
I had clean runs of the important code here.  The latest failures are just over a silly module rename that I just had to do.  I'll fix them up shortly.

feuGeneA:
Added checklist item:

*  Consider whether solcjs should be modified to support accepting a callback for resolving dependencies.

feuGeneA:
I considered using other IPC methods besides stdin/stdout, and decided it wasn't worth the trouble.  Specifically, I looked at using child_process.fork() instead of .spawn(), enabling us to pass a plain old JS object between processes, and that seemed to make sense, and would relieve us from needing to have parsing logic, but a first attempt at an implementation was taking long to debug, and I abandoned it since we already have something working and since there's no reason to believe it would improve performance.

feuGeneA:
At this point, in terms of functionality, this is done: it is functional.

However, more testing needs to be done, both manual and automated.

feuGeneA:
When working on this I noticed that there's a lot of repetitive compiling happening: when a file has a warning, you see the warnings listed multiple times, implying that they're being parsed more than once, wasting resources.

I dug deeper and discovered that we can pass all of the contract file names as arguments to one process invocation, rather than invoking a separate process for each file, which appears to be a much better approach, which I've implemented in #965.

PR #953, opened 2018-08-09
[order-utils] Add rate and sorting utilities

BMillman19:
## Description

* Add utilities for calculating fee-adjusted rates
* Add utilities for sorting order by best fee-adjusted rate

When calculating rates for orders we need to take into account how much fees the taker needs to pay to fill the order. In the case of fee orders (ZRX/WETH) the fee is taken out of the makerAmount. In the case of non ZRX orders (ex. REP/WETH) a feeRate can be provided to estimate how much takerAsset is actually required to fill that order when performing fee abstraction

## Testing instructions

```bash
yarn install && PKG=@0xproject/order-utils yarn build
cd packages/order-utils && yarn test
```

## Types of changes

## Checklist:

*   [x] Prefixed the title of this PR with `[WIP]` if it is a work in progress.
*   [x] Updated documentation, or decided that no doc change is needed.

### DIFF
--- a/packages/order-utils/src/rate_utils.ts
+++ b/packages/order-utils/src/rate_utils.ts
@@ -1,6 +1,7 @@
 import { schemas } from '@0xproject/json-schemas';
 import { SignedOrder } from '@0xproject/types';
 import { BigNumber } from '@0xproject/utils';
+import * as _ from 'lodash';
 
 import { assert } from './assert';
 import { constants } from './constants';
@@ -12,12 +13,16 @@ export const rateUtils = {
      * @param   signedOrder     An object that conforms to the signedOrder interface
      * @param   feeRate         The market rate of ZRX denominated in takerAssetAmount
      *                          (ex. feeRate is 0.1 takerAsset/ZRX if it takes 1 unit of takerAsset to buy 10 ZRX)
+     *                          Defaults to 0
      * @return  The rate (takerAsset/makerAsset) of the order adjusted for fees
      */
-    getFeeAdjustedRateOfOrder(signedOrder: SignedOrder, feeRate: BigNumber): BigNumber {
+    getFeeAdjustedRateOfOrder(signedOrder: SignedOrder, feeRate: BigNumber = constants.ZERO_AMOUNT): BigNumber {
         assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
         assert.isBigNumber('feeRate', feeRate);
-        assert.assert(feeRate.greaterThan(constants.ZERO_AMOUNT), `Expected feeRate: ${feeRate} to be greater than 0`);
+        assert.assert(
+            feeRate.gte(constants.ZERO_AMOUNT),
+            `Expected feeRate: ${feeRate} to be greater than or equal to 0`,
+        );
         const takerAssetAmountNeededToPayForFees = signedOrder.takerFee.mul(feeRate);
         const totalTakerAssetAmount = takerAssetAmountNeededToPayForFees.plus(signedOrder.takerAssetAmount);
         const rate = totalTakerAssetAmount.div(signedOrder.makerAssetAmount);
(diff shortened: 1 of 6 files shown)


packages/order-utils/src/rate_utils.ts
```diff
@@ -0,0 +1,48 @@
...
+        const rate = totalTakerAssetAmount.div(signedOrder.makerAssetAmount);
+        return rate;
+    },
+    /**
+     * Takes a signed fee order (makerAssetData corresponds to ZRX and takerAssetData corresponds to WETH) and calculates
+     * the fee adjusted rate (WETH/ZRX) by dividing the takerAssetAmount by the makerAmount minus the takerFee
+     * @param   signedFeeOrder    An object that conforms to the signedOrder interface
+     * @return  The rate (WETH/ZRX) of the fee order adjusted for fees
+     */
+    getFeeAdjustedRateOfFeeOrder(signedFeeOrder: SignedOrder): BigNumber {
```
fragosti: Should we assert that the pair is indeed WETH/ZRX? It could be another schema ?

BMillman19: It's a bit hard to assert w/o having to pass in the network context, i think its fine, the calculations don't necessarily rely on the order being WETH/ZRX, the results just don't make sense if a different type of order is passed in

packages/order-utils/src/rate_utils.ts
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Takes a signed order and calculates the fee adjusted rate (takerAsset/makerAsset) by calculating how much takerAsset
+     * is required to cover the fees (feeRate * takerFee), adding the takerAssetAmount and dividing by makerAssetAmount
+     * @param   signedOrder     An object that conforms to the signedOrder interface
+     * @param   feeRate         The market rate of ZRX denominated in takerAssetAmount
+     *                          (ex. feeRate is 0.1 takerAsset/ZRX if it takes 1 unit of takerAsset to buy 10 ZRX)
+     *                          Defaults to 0
+     * @return  The rate (takerAsset/makerAsset) of the order adjusted for fees
+     */
+    getFeeAdjustedRateOfOrder(signedOrder: SignedOrder, feeRate: BigNumber = constants.ZERO_AMOUNT): BigNumber {
```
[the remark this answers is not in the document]
BMillman19: hmm yea good call, I had some issues with some other functions that took in orders and had to return orders, in those cases you actually have to use generics so that the type system knows a SignedOrder should be spit out if a SignedOrder is passed in, let me see if I can finagle it

packages/order-utils/src/sorting_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+    },
+};
+
+type RateCalculator = (signedOrder: SignedOrder) => BigNumber;
+
+// takes an array of orders, copies them, and sorts the copy based on the rate definition provided by rateCalculator
+const sortOrders = (signedOrders: SignedOrder[], rateCalculator: RateCalculator): SignedOrder[] => {
+    const copiedOrders = _.cloneDeep(signedOrders);
+    const feeOrderComparator = getOrderComparator(rateCalculator);
+    copiedOrders.sort(feeOrderComparator);
```
abandeali1: Javascript sort isn't stable (it can sort in a different order depending on browser/environment, see: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). We should use `_.sortBy` here just in case.

packages/order-utils/src/sorting_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+type Comparator<T> = (first: T, second: T) => number;
+
+// takes a function that calculates rate for a signed order and returns a comparator that sorts based on rate
+const getOrderComparator = (rateCalculator: RateCalculator): Comparator<SignedOrder> => (
+    firstSignedOrder,
+    secondSignedOrder,
+) => {
+    const firstOrderRate = rateCalculator(firstSignedOrder);
+    const secondOrderRate = rateCalculator(secondSignedOrder);
+    if (firstOrderRate.lt(secondOrderRate)) {
```
abandeali1: Nit: `.comparedTo` does this for you (http://mikemcl.github.io/bignumber.js/#cmp)

packages/order-utils/src/sorting_utils.ts
BMillman19: You actually can't support via a comparator using _.sortBy

Issue #922, opened 2018-07-27
ORDER_FILL_ROUNDING_ERROR


rhrahul:
## Expected Behavior

When I call function,

`zeroEx.exchange.getOrderStateAsync(signedOrder)`

I got error,
`ORDER_FILL_ROUNDING_ERROR`

and order is not valid.

Any Suggestions ?

## Your Environment

|             Package | Version |
| ------------------: | :------ |
|             `0x.js` | 0.38.6  |
| `Exchange Contract` | v1      |

| Network |
| ------- |
| mainnet |

ikramansari77:
Ok sure,

I've 2 example,
First one working, second isn't.

```
{
    "maker": "#MAKERADDRESS",
    "taker": "0x0000000000000000000000000000000000000000",
    "exchangeContractAddress": "0x12459c951127e0c374ff9105dda097662a027093",
    "expirationUnixTimestampSec": "1534459419",
    "salt":
      "78749679624920336452693503448551535852550093042368521716715030697767324451372",
    "makerTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
    "takerTokenAddress": "0x5554e04e76533e1d14c52f05beef6c9d329e1e30",
    "makerTokenAmount": "500000000000000",
    "takerTokenAmount": "5",
    "makerFee": "0",
    "takerFee": "0",
    "feeRecipient": "0x0000000000000000000000000000000000000000",
    "orderHash":
      "0xbf79eb316e21427b86e0ab9d4d1412d5ae80ea5d77fa9e9262ff33796e6285a8",
    "ecSignature": {
      SIGNATURE
    }
  }
```

Here is second one, (NOT WORKING)

```
{
    "maker": "#MAKERADDRESS",
    "taker": "0x0000000000000000000000000000000000000000",
    "exchangeContractAddress": "0x12459c951127e0c374ff9105dda097662a027093",
    "expirationUnixTimestampSec": "1534459246",
    "salt":
      "78815684001532766840760980688541997753359462156134547970156133834539808922535",
    "makerTokenAddress": "0x5554e04e76533e1d14c52f05beef6c9d329e1e30",
    "takerTokenAddress": "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
    "makerTokenAmount": "6",
    "takerTokenAmount": "600000000000000000",
    "makerFee": "0",
    "takerFee": "0",
[...]

PR #959, opened 2018-08-13
[Order-utils] Order is valid when maker amount is very small

dekz:
#922 

Previously our min fillable calculation would throw a rounding error
when encountering a valid order (with a small maker amount). This was
inconsistent with the on-chain logic which allowed this order to be
filled.

e.g:
```
        makerAssetAmount: new BigNumber('10'),
        takerAssetAmount: new BigNumber('10000000000000000'),
```

Old logic would perform:
```
            minFillableTakerAssetAmountWithinNoRoundingErrorRange = signedOrder.takerAssetAmount
                .dividedBy(ACCEPTABLE_RELATIVE_ROUNDING_ERROR)
                .dividedBy(signedOrder.makerAssetAmount);
```
```
minFillableTakerAssetAmountWithinNoRoundingErrorRange 
=> 10000000000000000/0.0001/10
# a very big number
sidedOrderRelevantState.remainingFillableAssetAmount.lessThan(
                minFillableTakerAssetAmountWithinNoRoundingErrorRange)
=> false -> RoundingError
```

## Checklist:

### DIFF
--- a/packages/order-utils/test/order_state_utils_test.ts
+++ b/packages/order-utils/test/order_state_utils_test.ts
@@ -2,7 +2,9 @@ import { BigNumber } from '@0xproject/utils';
 import * as chai from 'chai';
 import 'mocha';
 
-import { AbstractBalanceAndProxyAllowanceFetcher, AbstractOrderFilledCancelledFetcher, OrderStateUtils } from '../src';
+import { AbstractBalanceAndProxyAllowanceFetcher } from '../src/abstract/abstract_balance_and_proxy_allowance_fetcher';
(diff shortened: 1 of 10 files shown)


packages/order-utils/test/order_state_utils_test.ts
```diff
@@ -0,0 +1,122 @@
+import { BigNumber } from '@0xproject/utils';
+import * as chai from 'chai';
+import 'mocha';
+
+import { AbstractBalanceAndProxyAllowanceFetcher, AbstractOrderFilledCancelledFetcher, OrderStateUtils } from '../src';
```
fabioberger: Let's not import from `./src` anymore but rather from the exact file, it makes it harder to grep for things.

PR #963, opened 2018-08-14
Refactor 0x.js exported interface

fabioberger:
## Description

Until now, we've been maintaining a legacy public interface for `0x.js`. Now that we've broken up the package into several smaller packages that can be imported separately:
- `@0xproject/contract-wrappers`
- `@0xproject/order-utils`
- `@0xproject/order-watcher`

It makes sense to remove the legacy interface and instead, re-export the same exports from these individual packages as the exports available via `0x.js`. 

### Benefits 

- This will make it trivial for someone to migrate from `0x.js` to only importing the sub-packages they actually use. 
- It should still be a useable and intuitive interface since it's good enough for the sub-packages.
- We don't need to maintain the duplicitous code in `0x.ts`, nor any tests for them. 

## Checklist:


fabioberger:
Will make this part of my other PR since I used the missing/superfluous export detection feature in there to perfect the index.ts file.

PR #965, opened 2018-08-14
[sol-compiler] Compile in batches rather than one at a time

feuGeneA:
## Description

Changed compiler invocation so that, instead of compiling one contract at a time, they're compiled in batches.  All contracts that require the same compiler version are compiled within the same batch.  With this change, compile times for `packages/contracts` were cut from 120 seconds to 60 (on my machine).

I took baby steps with the commits, to make reviewing easier, because there was a lot of code moving around but little actual change.  Note that in the commit that actually introduces the batching, there is some intentionally erroneous whitespace (a lack of indenting), which I did in order to make the diff easier to read.  Also, as a consequence of the baby steps, there may be a need for some significant style/organization changes; feedback from that perspective would be very welcome.

## Testing instructions

Check out the branch and run the tests for `sol-compiler`

## Types of changes

## Checklist:


packages/sol-compiler/src/compiler.ts
```diff
@@ -53,6 +53,30 @@ const DEFAULT_COMPILER_SETTINGS: solc.CompilerSettings = {
 };
 const CONFIG_FILE = 'compiler.json';
 
+async function getSolcAsync(
```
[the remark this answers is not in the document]
feuGeneA: Thanks, I didn't even realize static methods were a thing in TypeScript.

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +131,117 @@ export class Compiler {
...
+                const didSourceChange =
+                    currentArtifact.sourceTreeHashHex !== contractData[contractSource.path].sourceTreeHashHex;
+                shouldCompile = !isUserOnLatestVersion || didCompilerSettingsChange || didSourceChange;
+            }
+            if (!shouldCompile) {
+                continue;
             }
-            solcjs = await response.text();
-            fs.writeFileSync(compilerBinFilename, solcjs);
+            let solcVersion = this._solcVersionIfExists;
```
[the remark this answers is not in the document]
feuGeneA: I addressed this concern by taking a slightly different approach, making it so that `solcVersion` *can't* be undefined.  That's the intent of this variable anyways (note that right after the orignal assignment there was a check to set it if it was undefined).  Let me know what you think.

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +131,117 @@ export class Compiler {
...
+                solcVersion = semver.maxSatisfying(availableCompilerVersions, solcVersionRange);
+            }
+            if (_.isUndefined(inputsByVersion[solcVersion])) {
+                inputsByVersion[solcVersion] = {
+                    standardInput: {
+                        language: 'Solidity',
+                        sources: {},
+                        settings: this._compilerSettings,
+                    },
+                    contractsToCompile: [],
```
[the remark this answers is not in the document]
feuGeneA: No, you've misunderstood.  I've added a comment to try to clarify.  Let me know if I should explain more.

packages/sol-compiler/src/compiler.ts
```diff
@@ -53,6 +53,30 @@ const DEFAULT_COMPILER_SETTINGS: solc.CompilerSettings = {
...
+        const SUCCESS_STATUS = 200;
+        if (response.status !== SUCCESS_STATUS) {
+            throw new Error(`Failed to load ${fullSolcVersion}`);
+        }
+        solcjs = await response.text();
+        fs.writeFileSync(compilerBinFilename, solcjs);
+    }
+    const solcInstance = solc.setupMethods(requireFromString(solcjs, compilerBinFilename));
+    return { solcInstance, fullSolcVersion };
+}
```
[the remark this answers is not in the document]
feuGeneA: I think this is more complicated than a simple method extraction.  The reason is that the `compilerBinFilename` variable would still need to be used after the call to `isCompilerAvailableLocally`.

However, I did make a change to fix the correctness issue you raised.  Take a look at that and tell me what you think.

"Legacy" is all relative!  I'm definitely all for leaving things better than you found them.  Good to know that's the 0x philosophy too.

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +131,117 @@ export class Compiler {
...
             }
         }
-        const compiledData = compiled.contracts[contractSource.path][contractName];
+    }
+    private async _verifyAndPersistCompiledContractAsync(
+        contractPath: string,
+        contractMetadata: {
+            currentArtifactIfExists: ContractArtifact | void;
+            sourceTreeHashHex: string;
+            contractName: string;
```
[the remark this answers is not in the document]
feuGeneA: I think I addressed this comment in the "extract interfaces..." commit.

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -9,29 +11,28 @@ import { CompilerOptions, ContractArtifact } from '../src/utils/types';
 import { exchange_binary } from './fixtures/exchange_bin';
 import { constants } from './util/constants';
 
+chai.use(chaiAsPromised);
```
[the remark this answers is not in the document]
feuGeneA: I did this, but... I also noticed that there are now 10 different packages that have identical `chai_setup.ts` files.  I took a stab at consolidating them into `dev-utils`, but then got an error message `ERROR: Dependency cycle detected: @0xproject/web3-wrapper <- @0xproject/dev-utils <- @0xproject/subproviders <- @0xproject/web3-wrapper` while trying to `yarn clean`.  So, I decided to defer doing that cleanup, and raised an [Asana task for it](https://app.asana.com/0/684263176955174/781760362193367).

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -47,4 +48,65 @@ describe('#Compiler', function(): void {
         const exchangeBinaryWithoutMetadata = exchange_binary.slice(0, -metadataHexLength);
         expect(unlinkedBinaryWithoutMetadata).to.equal(exchangeBinaryWithoutMetadata);
     });
+    it("should throw when Whatever.sol doesn't contain a Whatever contract", async () => {
+        const contract = 'BadContractName';
+
+        const exchangeArtifactPath = `${artifactsDir}/${contract}.json`;
+        if (fsWrapper.doesPathExistSync(exchangeArtifactPath)) {
+            await fsWrapper.removeFileAsync(exchangeArtifactPath);
+        }
```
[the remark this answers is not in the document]
feuGeneA: Your call.  In my opinion it's too small to matter much, but I don't have a very strong opinion about it.

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -47,4 +48,65 @@ describe('#Compiler', function(): void {
...
+            compilerOpts.contracts = [contract];
+
+            artifactPath = `${artifactsDir}/${contract}.json`;
+            if (fsWrapper.doesPathExistSync(artifactPath)) {
+                await fsWrapper.removeFileAsync(artifactPath);
+            }
+
+            await new Compiler(compilerOpts).compileAsync();
+
+            timeCompiled = (await fsWrapper.statAsync(artifactPath)).mtimeMs;
```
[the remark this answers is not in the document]
feuGeneA: I wrote these variable names in an ambiguous way that caused you to misunderstand what's happening in these tests.  I've changed some variable names, which I think will clarify things; let me know.

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -47,4 +48,65 @@ describe('#Compiler', function(): void {
...
+        });
+        it('recompilation should update artifact when source has changed', async () => {
+            fsWrapper.appendFileAsync(join(contractsDir, `${contract}.sol`), ' ');
+
+            await new Compiler(compilerOpts).compileAsync();
+
+            const timeRecompiled = (await fsWrapper.statAsync(artifactPath)).mtimeMs;
+
+            expect(timeRecompiled).to.not.equal(timeCompiled);
```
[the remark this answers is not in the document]
feuGeneA: I wrote these variable names in an ambiguous way that caused you to misunderstand what's happening in these tests. I've changed some variable names, which I think will clarify things; let me know.

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -47,4 +48,65 @@ describe('#Compiler', function(): void {
...
+        });
+        it("recompilation should NOT update artifact when source hasn't changed", async () => {
+            await new Compiler(compilerOpts).compileAsync();
+
+            const timeRecompiled = (await fsWrapper.statAsync(artifactPath)).mtimeMs;
+
+            expect(timeRecompiled).to.equal(timeCompiled);
```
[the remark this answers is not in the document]
feuGeneA: I wrote these variable names in an ambiguous way that caused you to misunderstand what's happening in these tests. I've changed some variable names, which I think will clarify things; let me know.

packages/sol-compiler/src/compiler.ts
```diff
@@ -65,6 +82,32 @@ export class Compiler {
     private readonly _artifactsDir: string;
     private readonly _solcVersionIfExists: string | undefined;
     private readonly _specifiedContracts: string[] | TYPE_ALL_FILES_IDENTIFIER;
+    private static async _getSolcAsync(
+        solcVersion: string,
+    ): Promise<{ solcInstance: solc.SolcInstance; fullSolcVersion: string }> {
+        if (_.isUndefined(binPaths[solcVersion])) {
+            throw new Error(`${solcVersion} is not a known compiler version`);
+        }
+        const fullSolcVersion = binPaths[solcVersion];
```
fabioberger: Rather then use `binPaths[solcVersion]` twice, lets declare `fullSolcVersion` higher up and reference it in two places.

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +131,117 @@ export class Compiler {
         } else {
             contractNamesToCompile = this._specifiedContracts;
         }
-        for (const contractNameToCompile of contractNamesToCompile) {
-            await this._compileContractAsync(contractNameToCompile);
-        }
+        return this._compileContractsAsync(contractNamesToCompile);
```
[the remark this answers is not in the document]
fabioberger: Ah I see. Instead we should do:

```
await this._compileContractsAsync(contractNamesToCompile);
```

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +150,100 @@ export class Compiler {
...
+        for (const contractName of contractNames) {
+            const contractSource = this._resolver.resolve(contractName);
+            contractPathToData[contractSource.path] = {
+                contractName,
+                currentArtifactIfExists: await getContractArtifactIfExistsAsync(this._artifactsDir, contractName),
+                sourceTreeHashHex: `0x${this._getSourceTreeHash(
+                    path.join(this._contractsDir, contractSource.path),
+                ).toString('hex')}`,
+            };
+            if (!this._shouldCompile(contractPathToData[contractSource.path])) {
```
fabioberger: I'd prefer we do:

```
const contractData = {
                contractName,
                currentArtifactIfExists: await getContractArtifactIfExistsAsync(this._artifactsDir, contractName),
                sourceTreeHashHex: `0x${this._getSourceTreeHash(
                    path.join(this._contractsDir, contractSource.path),
                ).toString('hex')}`,
            };
contractPathToData[contractSource.path] = contractData;
...

if (!this._shouldCompile(contractData)) {
...
```

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +150,100 @@ export class Compiler {
...
+                continue;
             }
-            solcjs = await response.text();
-            fs.writeFileSync(compilerBinFilename, solcjs);
+            const solcVersion = _.isUndefined(this._solcVersionIfExists)
+                ? semver.maxSatisfying(_.keys(binPaths), parseSolidityVersionRange(contractSource.source))
+                : this._solcVersionIfExists;
+            if (_.isUndefined(versionToInputs[solcVersion])) {
+                // no inputs batched yet for this version.
+                // prepare object to hold this batch.
```
fabioberger: Rather then keep this comment here, let's do the following:

```
const isFirstContractWithThisVersion = _.isUndefined(versionToInputs[solcVersion]);
if (isFirstContractWithThisVersion) {
...
```

This eliminates the need for these comments.

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +150,100 @@ export class Compiler {
...
-            },
-            settings: this._compilerSettings,
-        };
-        const compiled: solc.StandardOutput = JSON.parse(
-            solcInstance.compileStandardWrapper(JSON.stringify(standardInput), importPath => {
-                const sourceCodeIfExists = this._resolver.resolve(importPath);
-                return { contents: sourceCodeIfExists.source };
-            }),
-        );
+        for (const solcVersion of _.keys(versionToInputs)) {
```
fabioberger: I know this is my suggestion. Feel this would be slightly more readable:

```
const solcVersions = _.keys(versionToInputs);
for (const solcVersion of solcVersions) {
```

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +150,100 @@ export class Compiler {
...
+    }
+    private async _verifyAndPersistCompiledContractAsync(
+        contractPath: string,
+        contractMetadata: ContractData,
+        fullSolcVersion: string,
+        compilerOutput: solc.StandardOutput,
+    ): Promise<void> {
+        const contractName = contractMetadata.contractName;
+        const sourceTreeHashHex = contractMetadata.sourceTreeHashHex;
+        const currentArtifactIfExists = contractMetadata.currentArtifactIfExists;
```
fabioberger: I'm not a huge fan of this, it feels deceptive of what all that this method takes in as parameters. Instead let's make this methods signature:

```
    private async _verifyAndPersistCompiledContractAsync(
        contractPath: string,
        contractName: string,
        sourceTreeHashHex: string,
        artifact: TYPE,
        fullSolcVersion: string,
        compilerOutput: solc.StandardOutput,
    ): Promise<void> {
```

packages/sol-compiler/src/compiler.ts
```diff
@@ -107,97 +150,100 @@ export class Compiler {
...
+
+        const compiledData = compilerOutput.contracts[contractPath][contractName];
```
fabioberger: Let's also pass in `compiledData`, not `compilerOutput`. Doesn't make sense to pass in the entirety of `compilerOutput` if that is all we actually need.

packages/sol-compiler/test/compiler_test.ts
```diff
@@ -47,4 +48,67 @@ describe('#Compiler', function(): void {
...
+            compilerOpts.contracts = [contract];
+
+            artifactPath = `${artifactsDir}/${contract}.json`;
+            if (fsWrapper.doesPathExistSync(artifactPath)) {
+                await fsWrapper.removeFileAsync(artifactPath);
+            }
+
+            await new Compiler(compilerOpts).compileAsync();
+
+            compilationTimestamp = (await fsWrapper.statAsync(artifactPath)).mtimeMs;
```
fabioberger: Is this the initial compilation?

packages/sol-compiler/src/compiler.ts
feuGeneA: I like it!

packages/sol-compiler/src/compiler.ts
feuGeneA: Love it!  Self documentation for the win!

packages/sol-compiler/src/compiler.ts
feuGeneA: I agree, it is deceptive, especially since it's just a one-off bundle of data that doesn't get re-used anywhere else.

packages/sol-compiler/src/compiler.ts
feuGeneA: But, we *do* use other properties on `compilerOutput`, specifically `sources` property, later in that method.

packages/sol-compiler/test/compiler_test.ts
feuGeneA: Yes.

packages/sol-compiler/src/compiler.ts
fabioberger: I see, in that case, let's leave it how it is. Thanks for clarifying!

packages/sol-compiler/src/compiler.ts
```diff
@@ -65,6 +82,32 @@ export class Compiler {
...
+    private static async _getSolcAsync(
+        solcVersion: string,
+    ): Promise<{ solcInstance: solc.SolcInstance; fullSolcVersion: string }> {
+        const fullSolcVersion = binPaths[solcVersion];
+        if (_.isUndefined(fullSolcVersion)) {
+            throw new Error(`${solcVersion} is not a known compiler version`);
+        }
+        const compilerBinFilename = path.join(SOLC_BIN_DIR, fullSolcVersion);
+        let solcjs: string;
+        const isCompilerAvailableLocally = fs.existsSync(compilerBinFilename);
```
albrow: If this is meant to run in parallel, we should change all these `Sync` function calls to the async equivalent. See https://github.com/0xProject/0x-monorepo/blob/development/packages/monorepo-scripts/src/test_installation.ts#L22 for an example.

PR #975, opened 2018-08-16
[WIP] [Contract-wrappers] Exchange execute transaction encoder

dekz:
## Description

Provide a pleasant, validated and type safe interface to the 0x Exchange Execute Transaction functionality. 

## Checklist:


packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1097,6 +1098,11 @@ export class ExchangeWrapper extends ContractWrapper {
         const zrxAssetData = assetDataUtils.encodeERC20AssetData(zrxTokenAddress);
         return zrxAssetData;
     }
+    public async executeTransactionEncoderAsync(): Promise<ExecuteTransactionEncoder> {
```
fabioberger: The name `executeTransactionEncoderAsync` was confusing to me since the first word is an verb that could also allude to the action this method does.

Perhaps `transactionEncoderAsync` is sufficient. This method should also have a comment explaining that this does not refer to an Ethereum tx.

Then the class would be `TransactionEncoder` and each method would be: `encodeFillOrderTx`.

Thoughts on this naming?

packages/contract-wrappers/src/utils/transaction_encoder.ts
```diff
@@ -0,0 +1,293 @@
...
+        const messageHex = `0x${eip721MessageBuffer.toString('hex')}`;
+        return messageHex;
+    }
+    /**
+     * Encodes a fillOrder transaction.
+     * @param  signedOrder           An object that conforms to the SignedOrder interface.
+     * @param  takerAssetFillAmount  The amount of the order (in taker asset baseUnits) that you wish to fill.
+     * @return Hex encoded abi of the function call.
+     */
+    public fillOrder(signedOrder: SignedOrder, takerAssetFillAmount: BigNumber): string {
```
fabioberger: Looks like these haven't been renamed to `getFillOrderTx`, etc... Please rename.

PR #977, opened 2018-08-16
Set curstom inspect printer in BigNumber

Recmo:
## Description

```
console.log(result.logs);
```

Before:

```
...
makerAssetFilledAmount: BigNumber { s: 1, e: 19, c: [Array] },
takerAssetFilledAmount: BigNumber { s: 1, e: 18, c: [Array] },
makerFeePaid: BigNumber { s: 1, e: 18, c: [Array] },
takerFeePaid: BigNumber { s: 1, e: 18, c: [Array] },
...
```

After:

```
...
makerAssetFilledAmount: 5000000000000000000,
takerAssetFilledAmount: 10000000000000000000,
makerFeePaid: 1000000000000000000,
takerFeePaid: 1000000000000000000,
...
```

We should probably push this to upstream `BigNumber.js`.

## Testing instructions

Please check:

* that I'm using the correct method of detecting a NodeJS runtime, and
* for a more elegant typescript solution than the current cast to `any`.

## Types of changes

It sets a pretty printer on the `BigNumber` prototype so NodeJS will print readable numbers in `console.log`.

## Checklist:

### DIFF
--- a/packages/utils/src/configured_bignumber.ts
+++ b/packages/utils/src/configured_bignumber.ts
@@ -11,4 +11,15 @@ BigNumber.config({
     DECIMAL_PLACES: 78,
 });
 
+// Set a debug print function for NodeJS
+if (typeof window === 'undefined') { // Check if running in NodeJS
+    const util = require('util');
+
+    // Set a custom util.inspect function
+    (BigNumber.prototype as any)[util.inspect.custom] = function() {
+        // Return the readable string representation
+        return this.toString();
+    };
+}
+
 export { BigNumber };


LogvinovLeon:
Do you know why does prettier fail to prettify those changes? Is there an issue for that?

Recmo:
@LogvinovLeon  Do you mean the code prettier or some prettifying plugin we have for Chai output?

Recmo:
Chai has it's own pretty printer. It would be nice if we hook that one to. This apparently involves monkey patching chai:

https://stackoverflow.com/questions/26980779/how-can-i-get-chai-to-show-actual-and-expected-values-using-tostring

packages/utils/src/configured_bignumber.ts
```diff
@@ -11,4 +11,22 @@ BigNumber.config({
     DECIMAL_PLACES: 78,
 });
 
+// Set a debug print function for NodeJS
+// Upstream issue: https://github.com/MikeMcl/bignumber.js/issues/188
+import isNode = require('detect-node');
+if (!isNode) {
```
fabioberger: Ack, this should be `if (isNode) {` right?

Recmo: Yes :)

PR #985, opened 2018-08-17
[contracts] Add contract for batch validating orders

abandeali1:
## Checklist:


packages/contracts/src/2.0.0/extensions/OrderValidator/OrderValidator.sol
```diff
@@ -0,0 +1,190 @@
...
+
+
+contract OrderValidator {
+
+    bytes4 constant internal ERC20_DATA_ID = bytes4(keccak256("ERC20Token(address)"));
+    bytes4 constant internal ERC721_DATA_ID = bytes4(keccak256("ERC721Token(address,uint256)"));
+
+    using LibBytes for bytes;
+
+    struct TraderInfo {
```
BMillman19: can we get balance and allowance of ZRX for each of these as well?

abandeali1: Whoops, forgot about that.

packages/migrations/src/2.0.0-beta-testnet/migration.ts
```diff
@@ -75,6 +76,16 @@ export const runV2TestnetMigrationsAsync = async (
     );
     artifactsWriter.saveArtifact(assetProxyOwner);
 
+    // Deploy OrderValidator
```
BMillman19: moving this to the end of the migration script fixes the tests

packages/contracts/src/2.0.0/extensions/OrderValidator/OrderValidator.sol
```diff
@@ -0,0 +1,218 @@
...
+        ordersInfo = EXCHANGE.getOrdersInfo(orders);
+        tradersInfo = getTradersInfo(orders, takerAddresses);
+        return (ordersInfo, tradersInfo);
+    }
+
+    /// @dev Fetches balance and allowances for maker and taker of order.
+    /// @param order The order structure.
+    /// @param takerAddress Address that will be filling the order.
+    /// @return Balances and allowances of maker and taker of order.
+    function getTraderInfo(LibOrder.Order memory order, address takerAddress)
```
dekz: Might be some unexpected behaviour if order has specified a taker address which doesn't match the provided address.

abandeali1: We could special case that here, but it's probably easy enough to do off-chain.

PR #991, opened 2018-08-21
Add support for TypeScript project references

albrow:
## Description

[Project references](https://www.typescriptlang.org/docs/handbook/project-references.html) are a new feature in TypeScript 3.0 which can enable faster, more efficient builds for projects which are composed of many packages (e.g. a monorepo like our own).

Getting this to work is a lot harder than it may seem on the surface. I needed to update a lot of __tsconfig.json__ and __package.json__ files or else certain packages would break in subtle and surprising ways (for example, in some packages I had to add `"rootDir": "./src"`).

There are two new commands in __package.json__ at the monorepo root.

1. `yarn build:ts` will build build all the necessary TypeScript files for all packages in the correct order based on their dependencies. It is much smarter about which files need to be recompiled and is pretty much instantaneous if no TypeScript source files have changed. However, it does not run any `pre_build` steps. This means that typically you will need to run `yarn build` the first time you clone the monorepo or e.g., if you made any changes to contracts.
2. `yarn watch:ts` is like `yarn build:ts` but actively and continuously watches all TypeScript files across all packages, recompiling every package and those that depend on it as needed.

Everything else works pretty much the same as before but will be a little bit more efficient in some cases.

As an added bonus: with this change "Go to Definition" in VS Code will take you to the TypeScript source code with the definition instead of the old behavior of taking you to the __.d.ts__ file 🎉 

## Testing instructions

Run `yarn build` and `yarn test`. Test out `yarn build:ts` and `yarn watch:ts` to make sure they work as expected.

## Types of changes

## Checklist:


albrow:
In spite of some of the issues I listed above, preliminary results look good.

The first time you run `yarn build` has been reduced from ~70s to ~60s on my machine. Subsequent runs have been reduced to ~30s. A majority of that time is the `yarn pre_build` step which runs even if nothing has changed. If you know you haven't changed any contracts or code which would be generated with `yarn pre_build`, you can skip that step by running `tsc -b` or `yarn tsc_build` directly. `tsc -b` is really intelligent about only recompiling when needed so it is nearly instantaneous if there are no code changes.

packages/0x.js/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
     "extends": "../../tsconfig",
     "compilerOptions": {
-        "outDir": "lib"
+        "outDir": "lib",
+        "rootDir": "."
```
albrow: `rootDir` is required here because tslint breaks if it is missing and tsconfig.json contains `composite: true`. `composite: true` is required in order to enable project references.

packages/0x.js/webpack.config.js
```diff
@@ -49,6 +49,8 @@ module.exports = {
                         loader: 'awesome-typescript-loader',
                         query: {
                             declaration: false,
+                            declarationMap: false,
```
albrow: These new `query` fields are overrides for the new default fields for the `tsconfig.json` at the root of the monorepo. We need to override these fields because they are incompatible with awesome-typescript-loader/webpack.

packages/contracts/test/utils/web3_wrapper.ts
```diff
@@ -1,5 +1,5 @@
 import { devConstants, env, EnvVars, web3Factory } from '@0xproject/dev-utils';
-import { prependSubprovider } from '@0xproject/subproviders';
+import { prependSubprovider, Web3ProviderEngine } from '@0xproject/subproviders';
```
albrow: With our new tsconfig settings, the TypeScript compiler requires this type to be explicitly imported.

packages/sra-api/package.json
```diff
@@ -52,7 +52,7 @@
         "swagger-cli": "^2.1.1",
         "ts-node": "^7.0.0",
         "tslint": "5.11.0",
-        "typescript": "2.7.1"
+        "typescript": "3.0.1"
```
albrow: This was an oversight and all packages should be using TypeScript 3.0.1 now.

packages/tslint-config/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
     "extends": "../../tsconfig",
     "compilerOptions": {
-        "typeRoots": ["../../node_modules/@types", "node_modules/@types"],
-        "outDir": "lib"
+        "outDir": "lib",
```
albrow: `typeRoots` is already specified in the base __tsconfig.json__ file. We don't need to specify it again.

packages/website/tsconfig.json
```diff
@@ -9,6 +9,8 @@
         "strictNullChecks": false,
         "noImplicitThis": false,
         "declaration": false,
+        "declarationMap": false,
```
albrow: These are overrides for the new settings in the base __tsconfig.json__ file. Website is the only package that does not support project references right now. Project references requires `allowJs: false` and website requires `allowJs: true`.

tsconfig.json
```diff
@@ -3,14 +3,55 @@
...
+        "composite": true,
+        "declaration": true,
+        "declarationMap": true,
+        "sourceMap": true
+    },
+    // The root of the project is just a list of references and does not contain
+    // any top-level TypeScript code.
+    "include": [],
+    "references": [
+        { "path": "./packages/monorepo-scripts" },
```
albrow: The order here does not matter, but we will need to remember to add any new packages in the future.

packages/0x.js/package.json
```diff
@@ -15,7 +15,6 @@
     "main": "lib/src/index.js",
     "types": "lib/src/index.d.ts",
     "scripts": {
-        "watch_without_deps": "tsc -w",
```
albrow: We don't need `watch_without_deps` any more. Running `yarn watch:ts` in the monorepo root will automatically watch all packages and their dependencies and only recompile them as needed. It's much faster than the old approach.

packages/0x.js/webpack.config.js
fabioberger: Can you add this comment to the file itself? Feel like other devs will wonder about these in the future.

packages/website/tsconfig.json
fabioberger: @albrow can you add this commend in the JSON file with a hack?

tsconfig.json
fabioberger: Mind starting a Quip doc called "How to add new packages to 0x monorepo"? Feel like there are now some must-do steps:

- Add to `tsconfig.json`
- Add to top-level README after it's published

etc...

tsconfig.json
```diff
@@ -3,14 +3,55 @@
...
+        { "path": "./packages/sol-compiler" },
+        { "path": "./packages/order-utils" },
+        { "path": "./packages/react-docs" },
+        { "path": "./packages/sol-cov" },
+        { "path": "./packages/contracts" },
+        { "path": "./packages/fill-scenarios" },
+        { "path": "./packages/metacoin" },
+        { "path": "./packages/migrations" },
+        // Skipping website because it requires allowJs: false and this is
+        // incompatible with project references.
```
fabioberger: @fragosti & @BMillman19 any chance we could try and remove the `allowJs: true` config for website?

tsconfig.json
```diff
@@ -3,14 +3,55 @@
...
+        // { "path": "./packages/website" }
```
fabioberger: @albrow does this mean that if website depends on another monorepo package, they must be manually re-built? Or since say `react-shared` **is** part of the project references, running `yarn watch:ts` will take care of re-compiling it?

fabioberger: Just tested it out, and it works!!! 💯  That is super clutch! Iteration cycles are going to be sooooo much tighter 💃

tsconfig.json
fabioberger: Given that the below works, I guess it doesn't necessarily add much, so nvm!

fabioberger:
Related to the CI test failure: https://github.com/Microsoft/TypeScript/issues/26554

tsconfig.json
albrow: I added a task to Asana for this: https://app.asana.com/0/684263176955174/798329537896649.

albrow:
All tests passed. Removing WIP tag. This PR is ready for final review!

@fragosti can you review changes to sra-spec?

fabioberger:
LGTM

PR #1000, opened 2018-08-22
[WIP][contracts] Fix testWithReferenceFunc (PR #1000 🎉)

Recmo:
## Description

Currently when using `testWithReferenceFuncAsync` or testing infrastructure depending on it like combinatorial tests, the tests incorrectly pass if the testee should throw but doesn't.

This is because:

* [`testWithReferenceFuncAsync:78`](https://github.com/0xProject/0x-monorepo/blob/554d5f97df55779beed35ef73214e80b386d7927/packages/contracts/test/utils/test_with_reference.ts#L78) throws an error to fail the test.
* This is caught (incorrectly) in line 85.
* It branches to line 88 because an error was expected.
* The test passes because the thrown error contains the expected message.

It is fixed by encapsulating the behaviour in a single data structure and doing a deep comparison.

### TODO

*  [ ] Solve handling of Geth Errors. This might need a big refactor / redesign of the strategy.
*  [ ] Complete error messages: add insufficient funds, out of gas, etc.
*  [ ] Add typesafety, refactor and cleanup (see todos)

## Checklist:

### DIFF
--- a/packages/contracts/test/utils/test_with_reference.ts
+++ b/packages/contracts/test/utils/test_with_reference.ts
@@ -1,27 +1,22 @@
-import * as chai from 'chai';
 import * as _ from 'lodash';
-
-import { chaiSetup } from './chai_setup';
-
-chaiSetup.configure();
-const expect = chai.expect;
 
 class Value<T> {
     public value: T;
     constructor(value: T) {
         this.value = value;
     }
-    public toString() {
+    public toString(): string {
         return `[Value: ${this.value}]`;
     }
 }
 
+// tslint:disable-next-line: max-classes-per-file
 class ErrorMessage {
     public error: string;
     constructor(message: string) {
         this.error = message;
     }
-    public toString() {
+    public toString(): string {
         return `[Error: ${this.error}]`;
     }
 }
@@ -31,7 +26,7 @@ type PromiseResult<T> = Value<T> | ErrorMessage;
 // TODO: This seems like a generic utility function that could exist in lodash.
 //       We should replace it by a library implementation, or move it to our
 //       own.
-export async function evaluatePromise<T>(
+async function evaluatePromise<T>(
     promise: Promise<T>,
 ): Promise<PromiseResult<T>> {
     try {
--- a/packages/contracts/test/utils/test_with_reference.ts
+++ b/packages/contracts/test/utils/test_with_reference.ts
@@ -1,12 +1,14 @@
+import * as chai from 'chai';
 import * as _ from 'lodash';
+
+import { chaiSetup } from '../utils/chai_setup';
+
+chaiSetup.configure();
 
 class Value<T> {
     public value: T;
     constructor(value: T) {
         this.value = value;
-    }
-    public toString(): string {
-        return `[Value: ${this.value}]`;
     }
 }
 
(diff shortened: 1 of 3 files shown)


packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -64,39 +102,17 @@ export async function testWithReferenceFuncAsync(
...
+    const expected = await evaluatePromise(referenceFuncAsync(...values));
+
+    // Measure actual behaviour
+    const actual = await evaluatePromise(testFuncAsync(...values));
+
+    // Compare behaviour
+    // Note: We could also use `expect(actual).to.deep.equal(expected)` but
+    //       this results in a less readable error message.
+    if (!_.isEqual(actual, expected)) {
+        throw new Error(`Test case ${_getTestCaseString(referenceFuncAsync, values)} expected ${expected} to equal ${actual}`);
```
albrow: This doesn't work great for objects.

```
> const obj = {a: 'foo', b: 'bar'};
> console.log(`[Value: ${obj}]`);
[Value: [object Object]]
```

If two objects are not equal I wouldn't have any idea how they differ. Can we consider going back to `expect(actual).to.deep.equal(expected)` or maybe using a different technique here?

packages/contracts/test/utils/test_with_reference.ts
```diff
@@ -64,39 +102,17 @@ export async function testWithReferenceFuncAsync(
...
+    // Measure correct behaviour
+    const expected = await evaluatePromise(referenceFuncAsync(...values));
+
+    // Measure actual behaviour
+    const actual = await evaluatePromise(testFuncAsync(...values));
+
+    // Compare behaviour
+    // Note: We could also use `expect(actual).to.deep.equal(expected)` but
+    //       this results in a less readable error message.
+    if (!_.isEqual(actual, expected)) {
```
albrow: This also will result in test failures when error messages do not equal exactly what we expect. The old version only required that the actual error message *contains* the expected error message. The reason for this is differences between the error messages from Ganache and Geth that are somewhat outside of our control.

packages/contracts/test/utils/test_with_reference.ts
Recmo: I was try to reuse Chai's `objDisp` function, but it seems that is not publicly accessible. I'll see how expect-equal fares.

packages/contracts/test/utils/test_with_reference.ts
Recmo: Fixed that for Ganache, which now has accurate error messages. I'll look into Geth.

Recmo: Ok, now looks like this:

```
      AssertionError: test case {"numerator":"1","denominator":"2","target":"1"}: expected { value: true } to deeply equal { Object (error) }
      + expected - actual

       {
      -  "value": true
      +  "error": "VM Exception while processing transaction: revert UINT256_OVERFLOW"
       }
```

packages/contracts/test/utils/test_with_reference.ts
Recmo: Ok, now looks like this:

```
      AssertionError: test case {"numerator":"1","denominator":"2","target":"1"}: expected { value: true } to deeply equal { Object (error) }
      + expected - actual

       {
      -  "value": true
      +  "error": "VM Exception while processing transaction: revert UINT256_OVERFLOW"
       }
```

Recmo:
Main issue has been fixed by #1032. Refactoring efforts have become stale. Closing

PR #1002, opened 2018-08-23
[contracts] Add mutexes and reentrancy guards

abandeali1:
## Description

- Adds a modifier `nonReentrant` that protects against reentrancy
- `nonReentrant` is added to all functions that use `getCurrentContextAddress` or make external calls
- Adds tests for all functions that make external calls (SignatureValidator will be addressed in another PR)

## Checklist:


packages/contracts/src/2.0.0/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -86,43 +89,14 @@ contract MixinExchangeCore is
         bytes memory signature
     )
         public
```
Recmo: Can we promote it to `external` now?

packages/contracts/src/2.0.0/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -43,17 +47,14 @@ contract MixinWrapperFunctions is
         bytes memory signature
     )
         public
```
Recmo: `external`?

packages/contracts/src/2.0.0/protocol/Exchange/MixinWrapperFunctions.sol
```diff
@@ -117,11 +111,12 @@ contract MixinWrapperFunctions is
         bytes[] memory signatures
     )
         public
```
Recmo: `external`? (..and for further functions)

packages/contracts/src/2.0.0/protocol/Exchange/MixinExchangeCore.sol
abandeali1: `external` doesn't work functions that have structs as arguments yet.

PR #1004, opened 2018-08-23
[ethereum-types, react-docs, sol-compiler, sol-doc, types, typescript-typings, website] Add new Solidity documentation generation utility

feuGeneA:
## Description

Add new utility to generate documentation from Solidity smart contracts.

The way it was done for v1 contracts was using a utility called Doxity, but that project has languished, so we've decided to roll our own.

## Testing instructions

Run the unit tests in the new `sol-doc` component.  And/or, run the `generate*docs` and `deploy*docs` scripts in `sol-doc/package.json`.

## Types of changes

## Checklist:

*   [x] Modify `typescript-typings/types/solc/index.d.ts`... it doesn't currently support devdoc!
*   [x] Fix (hack) problem with yarn not properly symlinking `0x-monorepo/node_modules/@0xproject/types`.
*   [x] Move doc types (`DocAgnosticFormat`, etc) from `react-docs` into `packages/types`
*   [x] Transform solidity compiler's ABI output into doc format.
*   [x] Transform solidity compiler's `devdoc` output into doc format.
*   [x] Exercise & automate test cases: pass in `contractsToCompile`; `.sol` file without `devdoc` content; a solidity function that returns more than one value; constructor w/&w/out devdoc; method w/&w/out devdoc.
*   [x] Determine whether a Solidity fallback function should be considered `constant` or not.
*   [x] Run generated doc objects through whatever reads them.
*   [x] Determine whether the `@0xproject/types.SolidityMethod.callPath` property has any relevance to our usage here.
*   [x] Determine how we should be setting the `@0xproject/types.EventArg.isIndexed` property.
*   [x] Create test case passing in a comma-delimited list of contracts (I suspect it might be broken).


feuGeneA:
CI is failing the `test-publish` check with: `http://localhost:4873/@0xproject%2freact-shared: this package is already present`.  Full output is [here](https://circleci.com/gh/feuGeneA/0x-monorepo/454).  It's not clear to me yet whether this has anything to do with my changes or not.

feuGeneA:
The failures in `circleci: test-publish` are weird, and neither I nor @albrow suspect they're related to my changes.

Out of the last 4 runs, various different failure patterns have occurred.

In [the most recent run](https://circleci.com/workflow-run/1634c9a5-de08-41d3-9bd7-42b62ba26d34), `test-publish` failed with:
* `ERROR in package @0xproject/react-shared: Command failed: /home/circleci/repo/node_modules/typescript/bin/tsc`
* `ERROR in package @0xproject/sol-compiler: Command failed: ``npm install --save @0xproject/sol-compiler@1.1.2 --registry=http://localhost:4873/`` npm ERR! 404 Not Found: ipaddr.js@1.8.0`
* `ERROR in package @0xproject/sol-cov: Command failed: ``npm install --save @0xproject/sol-cov@2.1.2 --registry=http://localhost:4873/`` npm ERR! ENOMEM: not enough memory, write`

Note the explicit out of memory error, and the coincident HTTP 404.

Also in that most recent run, the `test-rest` test for `@0xproject/sra-report` failed, with some operations timing out.  I'm guessing that this was just a side effect of the system running out of memory.

In [the 2nd-to-most-recent run](https://circleci.com/gh/feuGeneA/0x-monorepo/490), the failure had the same `tsc` error from `react-shared`, and also `ERROR in package @0xproject/migrations: Command failed: npm install --save @0xproject/migrations@1.0.7 --registry=http://localhost:4873/`.

In [the 3rd-to-most-recent run](https://circleci.com/gh/feuGeneA/0x-monorepo/482), the test complained of hitting HTTP 404's in trying to pull a package from Verdaccio, and the Verdaccio log also seems to show an HTTP 500 regarding that same package.  (I see 404's and 500's like this virtually every time I try to run this test locally, but this is the first time I've spotted them in CI.)

In [the 4th-to-most-recent run](https://circleci.com/gh/feuGeneA/0x-monorepo/474)), the failure again  showed the error in `tsc` for `react-shared` but this time it did not mention `@0xproject/migrations`.

This IS a resource-intensive test (lots of work to do, and a high degree of parallelism), so maybe that has something to do with it.  I watched my machine try to run it, and it maxes out both my CPU (quad-core i7-6820HQ @ 2.7 GHz) and my RAM.  I have 32 GB installed, but at the time I ran only 4 GB free (same as Circle CI, coincidentally), and I watched it dwindle away from 4 GB down to about 250 MB at one point (could have dipped lower between sampling), and it _seemed_ like the 404's showed up right at the time of max memory usage, which did take some time to get to.  In a second local run, after freeing up more memory (~13 GB), I watched it eat up about 3 GB before running into 404's, proving that, on my machine at least, the 404's aren't from running out of memory.

In any case, I was able to get through a local run of `test-publish` for _just_ the packages that failed in CI, and they both passed just fine, so I'm calling the code clean at this point.

packages/sol-doc/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "scripts": {
+        "build": "tsc",
+        "test": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 5000 --exit",
+        "lint": "tslint --project . --format stylish",
+        "clean": "shx rm -rf lib"
+    },
+    "repository": "https://github.com/0xProject/0x-monorepo.git",
+    "author": "F. Eugene Aumson",
+    "license": "Apache-2.0",
+    "private": false,
```
fabioberger: Omit this.

packages/sol-doc/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "chai": "^4.1.2",
+        "chai-as-promised": "^7.1.0",
+        "chai-bignumber": "^2.0.2",
+        "dirty-chai": "^2.0.1",
+        "make-promises-safe": "^1.1.0",
+        "mocha": "^5.2.0",
+        "shx": "^0.2.2",
+        "source-map-support": "^0.5.0",
+        "tslint": "5.11.0"
+    }
```
fabioberger: Add: 

```
"publishConfig": {
        "access": "public"
    }
```

packages/sol-doc/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "test": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --timeout 5000 --exit",
+        "lint": "tslint --project . --format stylish",
+        "clean": "shx rm -rf lib"
+    },
+    "repository": "https://github.com/0xProject/0x-monorepo.git",
+    "author": "F. Eugene Aumson",
+    "license": "Apache-2.0",
+    "private": false,
+    "dependencies": {
+        "@0xproject/react-docs": "^1.0.5",
```
fabioberger: Remove this dep.

packages/sol-doc/package.json
```diff
@@ -0,0 +1,35 @@
...
+        "@0xproject/sol-compiler": "^1.0.5",
+        "@0xproject/utils": "^1.0.5",
+        "ethereum-types": "^1.0.4",
+        "lodash": "^4.17.10"
+    },
+    "devDependencies": {
+        "chai": "^4.1.2",
```
fabioberger: Add: `@0xproject/tslint-config`

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+                outputSelection: {
+                    ['*']: {
+                        ['*']: ['abi', 'devdoc'],
+                    },
+                },
+            },
+        };
+    }
+    /// run `contractsToCompile` through compiler, gathering output
+    public async generateAsync(contractsToCompile: string[]): Promise<SolidityDocFormat> {
```
fabioberger: Add proper method doc comments since this method is part of the public interface.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+import { MethodAbi } from 'ethereum-types';
+
+import { Compiler, CompilerOptions } from '@0xproject/sol-compiler';
+import { logUtils } from '@0xproject/utils';
+
+import { SolidityDocFormat } from './solidity_doc_format';
+
+export class SolidityDocGenerator {
+    private readonly _compilerOptions: CompilerOptions;
+    constructor(contractsDir: string, artifactsDir: string) {
```
fabioberger: Add proper method doc comments since this constructor is part of the public interface.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
+import * as _ from 'lodash';
+
+import { MethodAbi } from 'ethereum-types';
+
+import { Compiler, CompilerOptions } from '@0xproject/sol-compiler';
+import { logUtils } from '@0xproject/utils';
+
+import { SolidityDocFormat } from './solidity_doc_format';
+
+export class SolidityDocGenerator {
```
fabioberger: Add a doc comment describing what this class does.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+                    ['*']: {
+                        ['*']: ['abi', 'devdoc'],
+                    },
+                },
+            },
+        };
+    }
+    /// run `contractsToCompile` through compiler, gathering output
+    public async generateAsync(contractsToCompile: string[]): Promise<SolidityDocFormat> {
+        if (!_.isUndefined(contractsToCompile)) {
```
fabioberger: Let's assign `!_.isUndefined(contractsToCompile)` to a variable called `shouldOverrideCatchAllContractsConfig`, otherwise it's not immediately clear what happens if **no** contracts are explicitly supplied.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+            this._compilerOptions.contracts = contractsToCompile;
+        }
+
+        const compiler = new Compiler(this._compilerOptions);
+
+        const doc = new SolidityDocFormat();
```
fabioberger: `SolidityDocFormat` shouldn't be a class, but rather an interface and `doc` an empty array: 

```
const doc: SolidityDocFormat = [];
```

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        const compilerOutputs = await compiler.getCompilerOutputsAsync();
```
fabioberger: Lets put `const compiler = new Compiler(this._compilerOptions);` right above this line. It's considered good practice to have close proximity between where something is instantiated and used.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+        for (const compilerOutput of compilerOutputs) {
+            const solidityModules = _.keys(compilerOutput.contracts);
```
fabioberger: Why not call this `contractFileName`? That's what it is  according to the `StandardOutput` type definition.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+            for (const solidityModule of solidityModules) {
+                const compiledSolidityModule = compilerOutput.contracts[solidityModule];
```
fabioberger: Rename this to `contractNameToOutput`, since that is what it is.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+
+                const contracts = _.keys(compiledSolidityModule);
```
fabioberger: `contractNames`

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,66 @@
...
+                for (const contract of contracts) {
+                    const compiledContract = compiledSolidityModule[contract];
+                    if (_.isUndefined(compiledContract.abi)) {
+                        throw new Error('compiled contract did not contain ABI output.');
+                    }
+                    if (_.isUndefined(compiledContract.devdoc)) {
+                        throw new Error('compiled contract did not contain devdoc output.');
```
fabioberger: Are you sure we can make this a "hard" requirement? Maybe we should also handle contracts without comments, no?

packages/typescript-typings/types/solc/index.d.ts
```diff
@@ -87,62 +88,7 @@ declare module 'solc' {
...
-                legacyAST?: object;
-            };
-        };
-        contracts: {
-            [fileName: string]: {
-                [contractName: string]: StandardContractOutput;
-            };
-        };
-    }
+    export { ErrorType, ErrorSeverity, SolcError, StandardContractOutput, StandardOutput } from 'ethereum-types';
```
fabioberger: Let's only have one export from `ethereum-types` at the top of the file, and exported everything we need from it.

packages/ethereum-types/src/index.ts
```diff
@@ -415,6 +476,7 @@ export interface Source {
     id: number;
 }
 
+// TODO: move the following into the sol-compiler package.
```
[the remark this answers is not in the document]
feuGeneA: Right, I think I wrote that before I understood that, removed.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: I like it.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: Thanks, I do like that better.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: You're right, we can't make this a hard requirement, and we should indeed handle contracts without comments.  I had corrected this in other (yet unpushed) commits, but I went ahead and cherry picked that change into this branch.

feuGeneA:
@albrow thanks for the feedback.  I've made the `postinstall` script be cross platform, via `shx`, in [this commit](https://github.com/0xProject/0x-monorepo/pull/1004/commits/75664b66512fb32539626a42973d293ecad17584).

feuGeneA:
@BMillman19 & @fragosti , you were only roped in because I [tweaked the website in my satisfy linter commit](https://github.com/0xProject/0x-monorepo/pull/1004/commits/327b4ba5546907f3c2b6fddd2a4d3d767d1f6767#diff-1681afacf1538a7f8b2cbb9b2e2ce9cd).  As you'll see, all I did was remove a few things that the linter told me were unused.

Why it was an issue for me, I have to admit I don't understand.  The linter started complaining after I merged `development` into my branch, so I just made those changes to shut it up.  If I should have done something different, please let me know

albrow:
@feuGeneA the new linter warnings are probably because https://github.com/0xProject/0x-monorepo/pull/1047 was merged into development.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+            for (const contractName of contractNames) {
+                const compiledContract = contractNameToOutput[contractName];
+                if (_.isUndefined(compiledContract.abi)) {
+                    throw new Error('compiled contract did not contain ABI output');
+                }
+                doc[contractName] = _genDocSection(compiledContract);
+            }
+        }
+    }
+
```
fabioberger: Nit: unnecessary new line. We try to use new lines to delineate separate chunks of logic that are related.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+        events: [],
+    };
+
+    for (const abiDefinition of compiledContract.abi) {
+        switch (abiDefinition.type) {
+            case 'constructor':
+                docSection.constructors.push(_genConstructorDoc(abiDefinition, compiledContract.devdoc));
+                break;
+            case 'event':
+                (docSection.events as Event[]).push(_genEventDoc(abiDefinition));
```
fabioberger: Why is casting to `Event[]` necessary here? Doesn't it know that from the `DocSection` type?

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+                throw new Error(`unknown and unsupported AbiDefinition type '${abiDefinition.type}'`);
+        }
+    }
+
+    return docSection;
+}
+
+function _genConstructorDoc(abiDefinition: ConstructorAbi, devdocIfExists: DevdocOutput | undefined): SolidityMethod {
+    const { parameters, methodSignature } = _genMethodParamsDoc(
+        '', // TODO: update depending on how constructors are keyed in devdoc
```
fabioberger: Is this outstanding?

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+        devdocIfExists,
+    );
+
+    const comment = _devdocMethodDetailsIfExist(methodSignature, devdocIfExists);
+    // TODO: figure out why devdoc'd constructors don't get output by solc
+
+    const constructorDoc: SolidityMethod = {
+        isConstructor: true,
+        name: '', // sad we have to specify this
+        callPath: '', // TODO: wtf is this?
```
fabioberger: `callPath` is used to display methods at `Exchange.fillOrder(` rather then `fillOrder(`, making the doc signatures more copy-pastable. There is debate as to whether we should do this, and I'm open to arguments either way.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+        parameters,
+        returnType: { name: '', typeDocType: TypeDocTypes.Intrinsic }, // sad we have to specify this
```
fabioberger: Hm, the return type of the constructor should be an instance of the contract no?

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+        isConstant: false, // constructors are non-const by their nature, right?
```
fabioberger: ✅

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,284 @@
...
+            type: { name: abiParam.type, typeDocType: TypeDocTypes.Intrinsic },
+        };
+        parameters.push(parameter);
+        methodSignature = `${methodSignature}${abiParam.type},`;
+    }
+
+    if (methodSignature.slice(-1) === ',') {
+        methodSignature = methodSignature.slice(0, -1);
+    }
+    methodSignature += ')';
```
fabioberger: We try to avoid string concatenation. I could have written this as:

```
// Aggregate param types into `paramTypes` array.
const methodSignature = `${name}(${paramTypes.join(',')})`;
```

feuGeneA:
It looks like we will not be able to support natspec/devdoc comments on constructors any time soon :(

The solidity compiler's `devdoc` output is not including any information for constructors.  [A fix for this problem was merged into `github.com/ethereum/solidity`'s `ethereum:develop` branch a month ago](https://github.com/ethereum/solidity/pull/4542), but for some reason it was not incorporated into the release that came out just today.  I pinged the merger/releaser (same person), on that PR, to ask when we might see it in a release.

Also, given the ancient laundry list of unchecked items in [this issue](https://github.com/ethereum/solidity/issues/1141), and a few other outstanding issues around devdoc, it's looking like this work might be incomplete for some time :/

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: I use newlines like that too, and I used it in that way here.  This method has 3 newline-separated "sections", one is declaration of the variable to be returned, the next is initialization of that variable, and the last is the return of that variable.

Note that I followed this same pattern in several other methods (`_makeCompilerOptions()`, `_genDocSection()`, and `_genConstructorDoc`, to name a few), but looking back I can see that I wasn't consistent about it!

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: It's because `DocSection.events` is declared as optional (with `?`).  And, apparently it's not smart enough to know that the case statement context implies that it's no longer optional.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: I don't really understand... Certainly not enough to be making arguments one way or the other...

I just removed the TODO and left the empty string.  If you'd like to see something else here, please let me know.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: I made it be a `TypeDocTypes.Reference`, as I thought it was the best fit from among those `enum` choices.  Does that make sense to you?

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: Love it, thank you.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,286 @@
...
+    for (const arg of args) {
+        const name = arg.name;
+
+        const type: Type = {
+            name: arg.type,
+            typeDocType: TypeDocTypes.Intrinsic,
+        };
+
+        const eventArgDoc: EventArg = {
+            isIndexed: false, // TODO: wtf is this?
```
fabioberger: This is whether or not that parameter is indexed. See: https://solidity.readthedocs.io/en/v0.4.24/contracts.html#events

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,286 @@
...
+    const { parameters, methodSignature } = _genMethodParamsDoc('', abiDefinition.inputs, devdocIfExists);
+
+    const comment = _devdocMethodDetailsIfExist(methodSignature, devdocIfExists);
+
+    const constructorDoc: SolidityMethod = {
+        isConstructor: true,
+        name: '', // sad we have to specify this
+        callPath: '',
+        parameters,
+        returnType: { name: '', typeDocType: TypeDocTypes.Reference }, // sad we have to specify this
```
fabioberger: Let's return:

```
{ name: contractName, typeDocType: TypeDocTypes.Reference }
```

where `contractName` is the contract's name.

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,306 @@
...
+            type,
+        };
+
+        eventArgsDoc.push(eventArgDoc);
+    }
+    return eventArgsDoc;
+}
+
+/**
+ * Extract documentation for each method paramater from @param params.
```
fabioberger: Typo:  `paramater` -> `parameter`

packages/sol-doc/src/solidity_doc_generator.ts
```diff
@@ -0,0 +1,306 @@
...
+function _genMethodParamsDoc(
+    name: string,
+    abiParams: DataItem[],
+    devdocIfExists: DevdocOutput | undefined,
+): { parameters: Parameter[]; methodSignature: string } {
+    const parameters: Parameter[] = [];
+    for (const abiParam of abiParams) {
+        const parameter: Parameter = {
+            name: abiParam.name,
+            comment: '',
```
fabioberger: No parameter comments?

packages/website/md/docs/smart_contracts/2.0.0/introduction.md
```diff
@@ -0,0 +1,8 @@
+Welcome to the [0x smart contracts](https://github.com/0xProject/contracts) documentation! This documentation is intended for dApp developers who want to integrate 0x exchange functionality directly into their own smart contracts.
+
+### Helpful wiki articles:
+
+*   [Overview of 0x protocol architecture](https://0xproject.com/wiki#Architecture)
+*   [0x smart contract interactions](https://0xproject.com/wiki#Contract-Interactions)
+*   [Deployed smart contract addresses](https://0xproject.com/wiki#Deployed-Addresses)
+*   [0x protocol message format](https://0xproject.com/wiki#Message-Format)
```
fabioberger: So for V2, only the `Deployed smart contract addresses` wiki article remains. Let's redirect them to: https://github.com/0xProject/0x-protocol-specification/blob/master/v2/v2-specification.md

fabioberger: We also need to remove all these links from the V1 introduction.md

packages/website/ts/pages/documentation/doc_page.tsx
```diff
@@ -128,7 +135,22 @@ export class DocPage extends React.Component<DocPageProps, DocPageState> {
...
+        let docAgnosticFormat;
+        if (this.props.docsInfo.type === SupportedDocJson.TypeDoc) {
+            docAgnosticFormat = new TypeDocUtils(
+                versionDocObj as GeneratedDocJson,
+                this.props.docsInfo,
+            ).convertToDocAgnosticFormat();
+        } else if (this.props.docsInfo.type === SupportedDocJson.Solidity) {
+            // documenting solidity.
+            docAgnosticFormat = versionDocObj as DocAgnosticFormat;
+            // need to modify docsInfo like convertToDocAgnosticFormat() would do
```
fabioberger: Lets prefix this comment with `HACK:`. This needs to be addressed in a future refactor.

packages/sol-doc/src/solidity_doc_generator.ts
feuGeneA: It's a 2-stage buildup in this method here.  First stage is data from ABI, second is from devdoc.  In the ABI stage, we don't have the information for the parameter comment, but we must declare the field here because it's required on the `Parameter` type.  Later in this method, we set this field to something meaningful, iff we find relevant text in the devdoc.

feuGeneA:
CI runs are looking good, except for one thing:  `yarn lint` in `packages/types` is failing:  `ERROR: /home/gene/0x-monorepo/packages/types/src/index.ts[598, 2]: This file has 599 lines, which exceeds the maximum of 500 lines allowed. Consider breaking this file up into smaller parts`

([I did add 223 lines to that index.ts...](https://github.com/0xProject/0x-monorepo/pull/1004/files#diff-104aaed728820f983dd962a1e332e447))

Should we really break it up, or just disable it for now, or what?

fabioberger:
@feuGeneA I would just disable the linter for now. This file is definitively an exception to that rule.

fabioberger:
There has been another publish, please double check the versions in `sol-docs`'s package.json before merging.

fabioberger:
If you rebase off `development`, `test-publish` tests should be passing. @albrow fixed them.

PR #1003, opened 2018-08-23
[contracts] Add getPartialAmountCeil and isRoundingErrorCeil

Recmo:
## Description

**Includes:** #1005.

* Adds `getPartialAmountCeil` and `isRoundingErrorCeil` that round up instead of down.
* Add `DIVISION_BY_ZERO` error to new and existing functions.
* Update/add test coverage.

## Testing instructions

Added test coverages. Added tests have merge conflicts with PR #1000.

## Types of changes

* New feature (non-breaking change which adds functionality)

## Checklist:

*   [x] Prefix PR title with `[WIP]` if necessary.
*   [x] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [x] Add tests to cover changes as needed.
*   [x] Update documentation as needed.
*   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #1005, opened 2018-08-23
[contracts] Fix isRoundingError error

Recmo:
## Description

Currently `isRoundingError` is documented to check for errors >0.1% but it actually checks for >0.1001%.

This patch changes the behaviour and docs to test for >=0.1%. This way all acceptable rounding errors are strictly less than 0.1%, which is more intuitive.

It also avoids an unnecessary division.

## Checklist:

### DIFF
--- a/packages/contracts/test/exchange/internal.ts
+++ b/packages/contracts/test/exchange/internal.ts
@@ -63,6 +63,7 @@ describe('Exchange core internal functions', () => {
     let testExchange: TestExchangeInternalsContract;
     let invalidOpcodeErrorForCall: Error | undefined;
     let overflowErrorForSendTransaction: Error | undefined;
+    let divisionByZeroErrorForCall: Error | undefined;
 
     before(async () => {
         await blockchainLifecycle.startAsync();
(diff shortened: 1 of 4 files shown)


packages/contracts/test/exchange/internal.ts
```diff
@@ -79,6 +80,9 @@ describe('Exchange core internal functions', () => {
         overflowErrorForSendTransaction = new Error(
             await getRevertReasonOrErrorMessageForSendTransactionAsync(RevertReason.Uint256Overflow),
         );
+        divisionByZeroErrorForCall = new Error(
+            await getRevertReasonOrErrorMessageForSendTransactionAsync(RevertReason.DivisionByZero),
```
Recmo: This should be `getRevertReasonOrErrorMessageForCall` instead, which doesn't exists.

Recmo: Should be fixed with #1000. That should also fix the unit test currently failing in the CI.

Recmo:
Closing in favour of PR #1003

PR #1012, opened 2018-08-24
Use staticcall for external function calls in MixinSignatureValidator

abandeali1:
## Checklist:

### DIFF
--- a/packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
+++ b/packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
@@ -293,8 +293,20 @@ contract MixinSignatureValidator is
                 cdStart,          // write input over output
                 32                // output size is 32 bytes
             )
-            // Signature is valid if call did not revert and returned true
-            isValid := and(success, mload(cdStart))
+
+            switch success
+            case 0 {
+                // Revert with `Error("WALLET_ERROR")`
+                mstore(0, 0x08c379a000000000000000000000000000000000000000000000000000000000)
+                mstore(32, 0x0000002000000000000000000000000000000000000000000000000000000000)
+                mstore(64, 0x0000000c57414c4c45545f4552524f5200000000000000000000000000000000)
+                mstore(96, 0)
+                revert(0, 100)
+            }
+            case 1 {
+                // Signature is valid if call did not revert and returned true
+                isValid := mload(cdStart)
+            }
         }
         return isValid;
     }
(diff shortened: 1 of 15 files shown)


packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,78 @@ contract MixinSignatureValidator is
...
+            let success := staticcall(
+                gas,              // forward all gas
+                walletAddress,    // address of Wallet contract
+                cdStart,          // pointer to start of input
+                mload(calldata),  // length of input
+                cdStart,          // write input over output
+                32                // output size is 32 bytes
+            )
+            // Signature is valid if call did not revert and returned true
+            isValid := and(success, mload(cdStart))
```
Recmo: This only works if the called contract returns a non-zero odd number for `true`, like `1`. I'm not sure if the ABI guarantees that boolean true is always encoded as `1` (instead of anything non-zero like in C).

To fix we could do `and(success, iszero(iszero(mload(cdStart))))`.

packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,78 @@ contract MixinSignatureValidator is
...
+            let success := staticcall(
+                gas,               // forward all gas
+                validatorAddress,  // address of Validator contract
+                cdStart,           // pointer to start of input
+                mload(calldata),   // length of input
+                cdStart,           // write input over output
+                32                 // output size is 32 bytes
+            )
+            // Signature is valid if call did not revert and returned true
+            isValid := and(success, mload(cdStart))
```
Recmo: Same.

packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,78 @@ contract MixinSignatureValidator is
...
+    function isValidWalletSignature(
+        bytes32 hash,
+        address walletAddress,
+        bytes signature
+    )
+        internal
+        view
+        returns (bool isValid)
+    {
+        bytes memory calldata = abi.encodeWithSelector(
```
Recmo: I think this will bumb the memory pointer unnecessarily, leading to extra gas cost. (not an issue)

packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,78 @@ contract MixinSignatureValidator is
...
+        address validatorAddress,
+        bytes32 hash,
+        address signerAddress,
+        bytes signature
+    )
+        internal
+        view
+        returns (bool isValid)
+    {
+        bytes memory calldata = abi.encodeWithSelector(
```
Recmo: Same.

packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,102 @@ contract MixinSignatureValidator is
...
+            signature
+        );
+        assembly {
+            let cdStart := add(calldata, 32)
+            let success := staticcall(
+                gas,              // forward all gas
+                walletAddress,    // address of Wallet contract
+                cdStart,          // pointer to start of input
+                mload(calldata),  // length of input
+                cdStart,          // write input over output
```
maurelian: Should be
// write output over input

packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -257,4 +262,102 @@ contract MixinSignatureValidator is
...
+            signature
+        );
+        assembly {
+            let cdStart := add(calldata, 32)
+            let success := staticcall(
+                gas,               // forward all gas
+                validatorAddress,  // address of Validator contract
+                cdStart,           // pointer to start of input
+                mload(calldata),   // length of input
+                cdStart,           // write input over output
```
maurelian: Should be
// write output over input

PR #1015, opened 2018-08-24
Remove SignatureType.Caller

abandeali1:
## Checklist:


packages/contracts/src/2.0.0/protocol/Exchange/MixinSignatureValidator.sol
```diff
@@ -48,14 +48,16 @@ contract MixinSignatureValidator is
...
     {
-        require(
-            isValidSignature(
-                hash,
-                signerAddress,
-                signature
-            ),
-            "INVALID_SIGNATURE"
-        );
+        if (signerAddress != msg.sender) {
```
Recmo: How will this interact with `executeTransaction`? Seems like currently the exchange contract can become a valid signer for anything.

Recmo: I think we need to add a `require(msg.sender != this)` here.

Recmo: Not an issue since we are using `DELEGATECALL` and are not otherwise making arbitrary self-calls.

thec00n: Would this mean that if I call `executeTransaction` and then call `preSign` that I could presign orders for the `Exchange` contract? 

I recommend to not allow adding pre signed signatures if `signerAddress == msg.sender`.

abandeali1: `executeTransaction` uses `delegatecall`, so the Exchange contract should never be `msg.sender`.

We've basically determined that usage of `msg.sender` here should be safe (this is a pretty common pattern). If we were using `getCurrentContextAddress`, there would definitely be extra protections.

thec00n: There is no direct vulnerability in the 0x protocol relating to this change. A scenario where an entity creates a smart contract that has a token balance and has approved the AssetProxies could still be vulnerable if they allow forward calls to the exchange contract and the function `preSign` is callable. It's acknowledged that this is outside of the control of 0x and for that reason the issue is considered closed. 

I still recommend caution around the assumption that no signature validation needs to be performed when `msg.sender == signerAddress`.

PR #1023, opened 2018-08-25
[website] Upgrade all the things to React 16

fragosti:
## Description

We've been meaning to upgrade to React 16 and I've recently run into a situation where I actually need it (to use the ReDoc component to render the sra-spec json). 

The method mostly consisted of updating all the dependencies that depend on React, including our fork of react-highlight. I also had to remove [react-tap-event-plugin](https://www.npmjs.com/package/react-tap-event-plugin) but you can read their documentation for why that's ok.

I later discovered that a lot of material-ui things were not working (dropdowns, closing-outside-to-click for modals etc...). While we are on an old version of material-ui, I discovered there is a 0.x version that supports React 16... so upgraded to that. Despite doing that, I noticed the versioning dropdown was still mishehaving so I changed that to use material UI v3.

Should I change the other components in react-shared to use material-ui v3?

## Checklist:


packages/website/ts/components/ui/drop_down.tsx
```diff
@@ -35,11 +36,21 @@ export class DropDown extends React.Component<DropDownProps, DropDownState> {
         activateEvent: DropdownMouseEvent.Hover,
         closeEvent: DropdownMouseEvent.Hover,
     };
-    private _isHovering: boolean;
     private _popoverCloseCheckIntervalId: number;
+    public static getDerivedStateFromProps(props: DropDownProps, state: DropDownState): Partial<DropDownState> {
```
BMillman19: if we are adding this public method we should probably export `DropDownState` as well to remain consistent

packages/website/ts/components/ui/drop_down.tsx
```diff
@@ -35,11 +36,21 @@ export class DropDown extends React.Component<DropDownProps, DropDownState> {
...
+        switch (props.activateEvent) {
+            case DropdownMouseEvent.Click:
+                return { isDropDownOpen: state.isDropDownOpen };
+            case DropdownMouseEvent.Hover:
+                return { isDropDownOpen: state.isHovering };
+            default:
+                return {};
```
BMillman19: do we want to return the same state as in the constructor here?

fabioberger:
@fragosti would be amazing if we could migrate to material-ui V3

fragosti:
@fabioberger yea I tried. It's a LOT of work. React 16 is currently blocking other things I'd like to do so want to get this in :) 

I can make a ticket to get us onto v3 for react-shared though.

fabioberger:
No worries, makes sense.

packages/react-shared/package.json
```diff
@@ -44,13 +45,12 @@
...
         "is-mobile": "^0.2.2",
         "lodash": "^4.17.5",
-        "material-ui": "^0.17.1",
-        "react": "15.6.1",
-        "react-dom": "15.6.1",
-        "react-highlight": "0xproject/react-highlight",
+        "material-ui": "^0.20.0",
+        "react": "^16.4.2",
+        "react-dom": "^16.4.2",
+        "react-highlight": "0xproject/react-highlight#2f40a42e0a3f0ad126f9f42d505b97b603fc7162",
```
fabioberger: 🙏

PR #1027, opened 2018-08-27
New Topbar Developers Dropdown

fabioberger:
## Description

This PR implements the new developers dropdown designed by @burnsben and @chriskalani.

screen shot 2018-08-27 at 12 23 30

All PR's that are part of the developers section re-design will be merged into `dev-section-redesign` branch, and once it's complete, it will be merged into `development`.

## Checklist:


packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+import * as React from 'react';
+import { Link } from 'react-router-dom';
+import { DropDown } from 'ts/components/ui/drop_down';
+import { Deco, Key, WebsitePaths } from 'ts/types';
+import { constants } from 'ts/utils/constants';
+import { Translate } from 'ts/utils/translate';
+
+interface KeyToLinkInfo {
+    [key: string]: LinkInfo;
+}
```
fragosti: FYI added an `ObjectMap` type to `types.ts` since this comes up a lot. Not sure why typescript doesn't have a native `Dict` type or something.

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+    menuItemStyles: React.CSSProperties;
+    menuIconStyle: React.CSSProperties;
+}
+
+interface DevelopersDropDownState {}
+
+export class DevelopersDropDown extends React.Component<DevelopersDropDownProps, DevelopersDropDownState> {
+    public render(): React.ReactNode {
+        const activeNode = (
+            <div className="flex relative" style={{ color: this.props.menuIconStyle.color }}>
```
fragosti: We've started using a `Container` component for this kind of stuff. Trying to move away from in-line styles. Not a huge change... Just using this `Container` component with some common sense layout / size props and then a `Text` component to wrap text with some common sense text props.

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+                targetOrigin={{ horizontal: 'left', vertical: 'top' }}
+                style={this.props.menuItemStyles}
+                popoverStyle={{ borderRadius: 4, width: 427, height: 373, marginTop: 10 }}
+            />
+        );
+    }
+    private _renderDropdownMenu(): React.ReactNode {
+        const dropdownMenu = (
+            <div>
+                <div style={{ padding: '1.75rem' }}>
```
fragosti: Same `Container`

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+                    </div>
+                </div>
+                <div
+                    style={{
+                        padding: '1.125rem',
+                        textAlign: 'center',
+                        backgroundColor: colors.lightBgGrey,
+                        borderBottomLeftRadius: 4,
+                        borderBottomRightRadius: 4,
+                    }}
```
fragosti: `Container` would be good here as well.

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+                    <Link
+                        to={WebsitePaths.ZeroExJs/* TODO: Update once we have overview page */}
+                        className="text-decoration-none"
+                        style={{
+                            color: colors.lightBlueA700,
+                            fontWeight: 'bold',
+                            fontSize: 14,
+                        }}
+                    >
+                        VIEW ALL DOCUMENTATION
```
fragosti: TODO: Translate?

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -0,0 +1,188 @@
...
+    private _renderLinkSection(keyToLinkInfo: KeyToLinkInfo): React.ReactNode {
+        const linkStyle: React.CSSProperties = {
+            color: colors.lightBlueA700,
+            fontFamily: 'Roboto, Roboto Mono',
+        };
+        const numLinks = _.size(keyToLinkInfo);
+        let i = 0;
+        const links = _.map(keyToLinkInfo, (linkInfo: LinkInfo, key: string) => {
+            i++;
+            const isLast = i === numLinks;
```
fragosti: You could get rid of the counter and the `isLast` boolean if you created a styled-component and used the :last-child selector. Not an issue though.

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fabioberger: Why are we trying to move away from in-line styles? Is it because it's easier to read props?

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fabioberger: Nice! We  should adopt this convention everywhere. @LogvinovLeon thoughts?

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fabioberger: Also, don't see `flex` as an option in the Container component. Should I add it? 

Looking at the `.flex` class from [BassCss](http://basscss.com/), it applies:

```
display: -ms-flexbox;
display: flex;
```

So it's doing lots of browser compatibility under-the-hood. Are we sure we want to forgo that?

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fabioberger: Good catch!

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fabioberger: Can we show me an example of this?

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fragosti: Looking at the `Container` component, I realize it's still using inline styles and should probably be changed to be using styled-components, which does automatically add browser prefixes. I can do this.

You can still add classNames to `Container` component which is what we've been doing to get `flex` layouts. Agree that the separation of concerns there is a bit confusing (ie. when to use props, vs when to use basscsss classnames). As mentioned, we've been trying to stay away from inline styles and using basscss where convenient. In an ideal world, I think Container would have the basscss features as typed props (https://jxnblk.com/rebass/). The reasoning behind this is: it's nicer and more "Reacty", our dependencies on styles are more explicit and specific (no dependency on global stylesheet) so if we ever want to move some components to a re-usable package for example, the user of that package wouldn't also have to include rebass. 

Anyway, we're probably not going to do this package thing, but I think trying to remove inline styles is a no brainer :P

packages/website/ts/components/dropdowns/developers_drop_down.tsx
fragosti: Again, not an issue but I just like to use css instead of JS whenever possible :D. Also you technically can't add a class name conditionally in css so this may be wrong. Example would be.

```js
import { styled } from 'ts/style/theme';

const PaddingForLastChildComponent  = styled.div`
    &:last-child {
        padding-bottom: 5px;
    }
`;

<PaddingForLastChildComponent>
    {list.map(()=>...)
</PaddingForLastChildComponent>
```

PR #1038, opened 2018-08-29
[WIP] Mainnet migrations

abandeali1:
## Description

- Removes the constructor from the Exchange contract
- `yarn migrate:v2-mainnet` will compile the new contracts, generate wrappers, compile the TypeScript, and run the mainnet migration.

## Checklist:


fabioberger:
@abandeali1 we do want to merge all of this minus there ZRX hard-coding. How can we do that?

dekz:
We also need to line up a ZRX deposit transaction into the Forwarder

packages/migrations/src/migrate.ts
```diff
@@ -15,6 +16,7 @@ enum ContractVersions {
     V1 = '1.0.0',
     V2 = '2.0.0',
```
albrow: Should we just remove `V2` here? It seems like we should be using either `V2Testnet` or `V2Mainnet` everywhere.

PR #1047, opened 2018-08-30
Re-add no-unused-variable tslint rule

fragosti:
## Description

I'm not sure why we removed this rule. My guess is because it currently spews out deprecation warnings and recommends to use the compiler options instead. However, turning on the compiler options doesn't really work for us because we need to be able to [selectively not apply this rule to certain files](https://github.com/Microsoft/TypeScript/issues/12913).

It seems that a [lot of people feel this way, so it has been undeprecated](https://github.com/palantir/tslint/pull/2256) so we should go back to using it. We already had a few unused variables lying around.

## Checklist:


fragosti:
@albrow my confidence is pretty high that this didn't break the publishing flow so I'm going to merge.

PR #1050, opened 2018-09-02
Update MultiSig and add more tests

abandeali1:
## Description

- Updates the Gnosis MultiSigWallet to the most [recent version](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol) (with the only difference being that `external_call` is internal instead of private)
- Removes `revokeConfirmation` from `MultiSigWalletWithTimeLock`. This change allows confirmations to be revoked after the confirmation time has been set. Before, you could implicitly delay the execution of a transaction by increasing the number of required owners or increasing the timelock - now there is an explicit way to do so.
- Note that _anyone_ can call `executeTransaction` in `MultiSigWalletWithTimeLock`. This differs from the new behavior of the Gnosis MultiSigWallet.

## Checklist:

### DIFF
--- a/packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
+++ b/packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
@@ -1,13 +1,14 @@
 // solhint-disable
-pragma solidity ^0.4.10;
+pragma solidity ^0.4.15;
 
 
 /// @title Multisignature wallet - Allows multiple parties to agree on transactions before execution.
 /// @author Stefan George - <stefan.george@consensys.net>
 contract MultiSigWallet {
 
-    uint constant public MAX_OWNER_COUNT = 50;
-
+    /*
+     *  Events
+     */
     event Confirmation(address indexed sender, uint indexed transactionId);
     event Revocation(address indexed sender, uint indexed transactionId);
     event Submission(uint indexed transactionId);
@@ -18,6 +19,14 @@ contract MultiSigWallet {
     event OwnerRemoval(address indexed owner);
     event RequirementChange(uint required);
 
+    /*
+     *  Constants
+     */
+    uint constant public MAX_OWNER_COUNT = 50;
+
+    /*
+     *  Storage
+     */
     mapping (uint => Transaction) public transactions;
     mapping (uint => mapping (address => bool)) public confirmations;
     mapping (address => bool) public isOwner;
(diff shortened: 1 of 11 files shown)


packages/contracts/test/multisig/asset_proxy_owner.ts
```diff
@@ -422,6 +430,33 @@ describe('AssetProxyOwner', () => {
                 expect(isAuthorized).to.equal(false);
             });
 
+            it('should execute removeAuthorizedAddressAtIndex for registered address if fully confirmed and called by non-owner', async () => {
+                const removeAuthorizedAddressAtIndexData = erc20Proxy.removeAuthorizedAddressAtIndex.getABIEncodedTransactionData(
+                    authorized,
+                    erc20Index,
+                );
+                const submitRes = await multiSigWrapper.submitTransactionAsync(
```
hysz: Nit - add sanity check that `authorized` is an authorized address before removing it.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
```diff
@@ -222,20 +224,44 @@ contract MultiSigWallet {
...
                 ExecutionFailure(transactionId);
-                tx.executed = false;
+                txn.executed = false;
             }
         }
     }
 
+    // call has been separated into its own function in order to take advantage
+    // of the Solidity's code generator to produce a loop that copies tx.data into memory.
+    function external_call(address destination, uint value, uint dataLength, bytes data) internal returns (bool) {
```
hysz: Param description + `function` style

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
```diff
@@ -1,13 +1,14 @@
 // solhint-disable
-pragma solidity ^0.4.10;
+pragma solidity ^0.4.15;
```
hysz: Why only `0.4.15`?

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
```diff
@@ -222,20 +224,44 @@ contract MultiSigWallet {
...
 
+    // call has been separated into its own function in order to take advantage
+    // of the Solidity's code generator to produce a loop that copies tx.data into memory.
+    function external_call(address destination, uint value, uint dataLength, bytes data) internal returns (bool) {
+        bool result;
+        assembly {
+            let x := mload(0x40)   // "Allocate" memory for output (0x40 is where "free memory" pointer is stored by convention)
+            let d := add(data, 32) // First 32 bytes are the padded length of data, so exclude that
+            result := call(
+                sub(gas, 34710),   // 34710 is the value that solidity is currently emitting
```
hysz: I'm guessing the unaccounted gas is spent on stack management. Wonder if we should apply some padding. Cost shouldn't vary by much. But, for example, optimization runs could affect this value.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
```diff
@@ -222,20 +224,44 @@ contract MultiSigWallet {
...
+        confirmed(transactionId, msg.sender)
         notExecuted(transactionId)
     {
         if (isConfirmed(transactionId)) {
-            Transaction tx = transactions[transactionId];
-            tx.executed = true;
-            if (tx.destination.call.value(tx.value)(tx.data))
+            Transaction storage txn = transactions[transactionId];
+            txn.executed = true;
+            if (external_call(txn.destination, txn.value, txn.data.length, txn.data))
```
hysz: Style nit - include `if` braces when followed by multiline `else`.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
```diff
@@ -239,7 +239,6 @@ contract LibMath is
         }
         // Compute remainder as before
         uint256 remainder = mulmod(target, numerator, denominator);
-        // TODO: safeMod
```
hysz: Do you know what `safeMod` was going to check for?
Presumably just that the LHS argument is >= 0. In which case I guess we should be safe with only `safeSub`.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
abandeali1: This is taken from: https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol

Lot's of it doesn't match our conventions, but I'd rather not change it if we don't have to.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
abandeali1: See above.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
abandeali1: See above.

packages/contracts/src/2.0.0/multisig/MultiSigWallet.sol
abandeali1: See above.

packages/contracts/src/2.0.0/protocol/Exchange/libs/LibMath.sol
abandeali1: I'm actually not sure, but it should be safe with the `denominator > 0` check.

PR #1052, opened 2018-09-04
[typescript-typings] Remove types that were moved to DefinitelyTyped

LogvinovLeon:
## Checklist:


packages/subproviders/package.json
```diff
@@ -52,13 +52,16 @@
...
+        "@types/eth-lightwallet": "^3.0.0",
         "@types/bn.js": "^4.11.0",
         "@types/ethereumjs-tx": "^1.0.0",
         "@types/hdkey": "^0.7.0",
+        "@types/web3-provider-engine": "^14.0.0",
         "@types/lodash": "4.14.104",
         "@types/mocha": "^2.2.42",
         "@types/node": "^8.0.53",
         "@types/sinon": "^2.2.2",
+        "@types/ganache-core": "^2.1.0",
```
fabioberger: Think this needs to be a `dependency` since we expose some of it's types in the public interface.

packages/subproviders/package.json
```diff
@@ -52,13 +52,16 @@
     "devDependencies": {
         "@0xproject/tslint-config": "^1.0.6",
         "@types/bip39": "^2.4.0",
+        "@types/eth-lightwallet": "^3.0.0",
```
fabioberger: Same  as below.

PR #1059, opened 2018-09-05
[WIP] [sol-compiler] Output only relevant sources

dekz:
## Description

TODO is the ID under sources meant to represent the index into sourceCodes? From simple tests it feels redundant if so?

## Checklist:

### DIFF
--- a/packages/sol-compiler/src/compiler.ts
+++ b/packages/sol-compiler/src/compiler.ts
@@ -264,14 +264,28 @@ export class Compiler {
             }
         }
 
-        const sourceCodes = _.mapValues(
-            compilerOutput.sources,
-            (_1, sourceFilePath) => this._resolver.resolve(sourceFilePath).source,
+        // Output only the relevant source code based off the dependency tree
+        const dependencies = this._getDependencyList(contractName);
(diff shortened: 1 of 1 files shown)


feuGeneA:
I like the idea of the slimmed down output that this change provides, but I do have one reservation:

With this change in place, won't there be two redundant calls to `parseDependencies()`?

(Once via the call to `_getSourceTreeHash()` from `_compileContractsAsync()` (the existing path), and once via the call to your new `_getDependencyList()`, which is called from `_verifyAndPersistCompiledContractAsync()`, which in turn is called from `_compileContractsAsync()`, a bit after the aforementioned call to `_getSourceTreeHash()`.)

How significant do you think the performance impact of this redundancy is?  I don't see any file I/O or anything like that happening in that function; it's just string parsing.  But I do see there are regex's used there; not sure how big of a deal that is.

I'd like to see either a change to eliminate this redundancy, or else an explanation of why you think it's insignificant.

albrow:
@feuGeneA @dekz I think this is superseded by https://github.com/0xProject/0x-monorepo/pull/1108?

PR #1064, opened 2018-09-07
[website] Fix minting for new tokens on testnets

fragosti:
## Description

We deployed new versions of tokens for v2 that are mintable. This is a patch until we update the token registry

## Checklist:


fragosti:
@BMillman19 approved these changes

Issue #1076, opened 2018-09-19
awaitTransactionMinedAsync returns transaction receipt while still pending (Parity Only)


cavanmflynn:
`getTransactionReceiptAsync`, and therefore `awaitTransactionMinedAsync`, return a pending transaction receipt when using a Parity node. It may be purposeful for `getTransactionReceiptAsync` to return the pending transaction receipt as that is the behavior of the client, but it causes `awaitTransactionMinedAsync` to exit prematurely.

### Call

`const receipt = await this._web3Wrapper.awaitTransactionMinedAsync('0xb0d3179b95093f9de689f03b340a985d5b413b37c194d65412d72fb53ee008df');`

#### Node
`Geth/v1.8.13-patched-infura-omnibus-b59d4428/linux-amd64/go1.9.2`:

#### Immediate Output
`None - Tx is pending`

#### Node
`Parity//v1.10.6-stable-bc0d134-20180605/x86_64-linux-gnu/rustc1.26.1`

#### Immediate Output
```
{  
   "blockHash":null,
   "blockNumber":null,
   "contractAddress":null,
   "cumulativeGasUsed":"0xb808",
   "gasUsed":"0xb808",
   "logs":[  
      {  
         "address":"0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2",
         "blockHash":null,
         "blockNumber":null,
         "data":"0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
         "logIndex":null,
         "topics":[  
            "0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925",
            "0x00000000000000000000000015fdb20d7629cc87fa719234b26bbf580bdb8958",
            "0x0000000000000000000000002240dab907db71e64d3e0dba4800c83b5c502d4e"
         ],
         "transactionHash":null,
         "transactionIndex":null,
         "transactionLogIndex":null,
         "type":"pending"
      }
   ],
   "logsBloom":"0x00000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000042002000000080000000000000000200000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000000008000000000000000000400000000000000000010000000000000000000010200000000000000000000000000000000000000000000000000000000000",
[...]

PR #1079, opened 2018-09-21
Fix awaitTransactionMinedAsync for new Parity version

fabioberger:
## Description

Addressess: https://github.com/0xProject/0x-monorepo/issues/1076 by also checking that the `blockNumber` in the returned `transactionReceipt` is not `null`.

## Checklist:

### DIFF
--- a/packages/web3-wrapper/CHANGELOG.json
+++ b/packages/web3-wrapper/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "2.0.3",
+        "changes": [
+            {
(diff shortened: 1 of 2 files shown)


packages/web3-wrapper/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.0.3",
+        "changes": [
+            {
+                "note":
```
LogvinovLeon: Add PR number.

PR #1082, opened 2018-09-24
Fix block fetch error if block not found

fabioberger:
## Description

Currently `web3Wrapper.getBlockAsync` throws:

```
TypeError: Cannot read property 'difficulty' of null
```

If the request for a particular block fails. A block request could fail if a node isn't fully synced, or if it's become uncle'd, among other reasons. The Ethereum JSON RPC spec specified returning `null` for requests where the block cannot be found.

This PR therefore renames `getBlockAsync` to  `getBlockIfExistsAsync` and returns `undefined` (in-line with our codebase conventions) if the block was not found. 

## Checklist:

### DIFF
--- a/packages/web3-wrapper/CHANGELOG.json
+++ b/packages/web3-wrapper/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "3.0.0",
+        "changes": [
+            {
+                "note":
+                    "Rename `getBlockAsync` to `getBlockIfExistsAsync` and rather then throw if the requested block wasn't found, return undefined."
+            }
+        ]
+    },
+    {
         "version": "2.0.3",
         "changes": [
             {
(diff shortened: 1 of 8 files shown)


packages/web3-wrapper/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "3.0.0",
+        "changes": [
+            {
+                "note":
+                    "Rename `getBlockAsync` to `getBlockIfExistsAsync` and rather then throw if the requested block wasn't found, return undefined."
```
LogvinovLeon: PR number

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -329,23 +329,28 @@ export class Web3Wrapper {
     /**
      * Fetch a specific Ethereum block without transaction data
      * @param blockParam The block you wish to fetch (blockHash, blockNumber or blockLiteral)
-     * @returns The requested block without transaction data
+     * @returns The requested block without transaction data, or undefined if block was not found
```
LogvinovLeon: Can we add a comment why the block could be not found?

PR #1083, opened 2018-09-24
Feature/custom verdaccio dockerfile

albrow:
## Description

This PR creates a new "package" called verdaccio which contains a __Dockerfile__ and __conf.yaml__ Like the devnet package, it is just configuration files. There is no real code and it won't be published.

Here are the changes I made to __conf.yaml__:

- Disabled the website frontend.
- Disabled the audit middleware.
- Increased the number of retries for requests to npmjs.org.
- Increased `max_body_size` to 100MB.

This should slightly improve performance, reduce memory usage, and fix some issues we've been seeing in CI having to do with uplink request failures.

We can probably make further optimizations in the future.

## Testing instructions

To run the new docker image locally, you can do:

`sudo docker run --rm -i -p 4873:4873 albrow/0x-verdaccio`

Then run `yarn test:publish:circleci` in a separate process.

## Types of changes

## Checklist:

### DIFF
--- a/packages/verdaccio/conf.yaml
+++ b/packages/verdaccio/conf.yaml
@@ -1,0 +1,71 @@
+#
+# Copied from https://github.com/verdaccio/verdaccio/blob/90fc216649ecd8df31d3706b5c17555fa0a35e4a/conf/docker.yaml
+# Only the uplinks section has been modified.
+#
+# This is the config file used for the docker images.
+# It allows all users to do anything, so don't use it on production systems.
+#
+# Do not configure host and port under `listen` in this file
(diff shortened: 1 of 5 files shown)


packages/verdaccio/conf.yaml
```diff
@@ -0,0 +1,71 @@
+#
+# Copied from https://github.com/verdaccio/verdaccio/blob/90fc216649ecd8df31d3706b5c17555fa0a35e4a/conf/docker.yaml
+# Only the uplinks section has been modified.
```
albrow: Reminding myself to change this comment because other sections have been modified now.

PR #1085, opened 2018-09-25
[types] Move SRA types from @0xproject/connect to @0xproject/types

LogvinovLeon:
## Description

We want to reuse that types in other places incuding 0x-relayer-template and we need them exported.
Those types and SRA related, not connect related and that's why we created a new package.

## Testing instructions

It builds

## Checklist:

### DIFF
--- a/packages/connect/CHANGELOG.json
+++ b/packages/connect/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "2.0.3",
+        "changes": [
+            {
+                "note": "Import SRA-related types from @0xproject/types",
+                "pr": 1085
+            }
+        ]
+    },
+    {
         "timestamp": 1537541580,
         "version": "2.0.2",
         "changes": [
--- a/packages/connect/src/index.ts
+++ b/packages/connect/src/index.ts
@@ -1,23 +1,20 @@
 export { HttpClient } from './http_client';
 export { ordersChannelFactory } from './orders_channel_factory';
+export { Client, OrdersChannel, OrdersChannelHandler } from './types';
 export {
-    Client,
-    OrderConfigRequest,
-    OrderConfigResponse,
-    OrdersChannel,
-    OrdersChannelHandler,
-    OrdersChannelSubscriptionOpts,
-    OrderbookRequest,
-    OrderbookResponse,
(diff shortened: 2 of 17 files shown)


packages/connect/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "2.0.3",
+        "changes": [
+            {
+                "note": "Import SRA-related types from @0xproject/sra-types"
+            }
```
fabioberger: PR nr.

packages/sra-types/README.md
```diff
@@ -0,0 +1,69 @@
+## @0xproject/types
```
fragosti: Should be `sra-types`, but also why did you decide not to just put these in types? Or just depend on connect in your package?

packages/sra-types/.npmignore
```diff
@@ -0,0 +1,6 @@
+.*
+yarn-error.log
+/scripts/
+/src/
+tsconfig.json
+/lib/monorepo_scripts/
```
fabioberger: Remove this one, we don't have `monorepo_scripts` anymore in package `lib` dirs.

packages/sra-types/README.md
fabioberger: `sra-types`

packages/connect/src/index.ts
```diff
@@ -1,23 +1,21 @@
 export { HttpClient } from './http_client';
 export { ordersChannelFactory } from './orders_channel_factory';
-export {
-    Client,
+export { Client, OrdersChannel, OrdersChannelHandler } from './types';
+import {
```
fragosti: So if you want to use any of these types you need to add sra-types now too? Same question as above.

fragosti: Ok I see you answered this in the description. I guess the only question is why new package instead of just `types`

fragosti: And should connect re-export these types for convenience?

packages/sra-types/README.md
LogvinovLeon: You're right. Moved everything into types

LogvinovLeon:
@fragosti Could you take a look one more time?

PR #1086, opened 2018-09-25
[asset-buyer] Improve API and README

fragosti:
## Description

Just wanted to make some improvements to the API. I think it's getting there... maybe some naming improvements around `AssetBuyer` vs `AssetBuyerManager`, maybe `AssetBuyerManager` should be `AssetBuyer` and the main export and we can name the current `AssetBuyer` something else.

## Testing instructions
It builds!

## Checklist:


BMillman19:
@fragosti some comments on the min / max eth calculation approach: Firstly `feeAdjustedMaxRate` needs to be based on `maxEthAmount`, this is just a typo. Second, I think we need to tweak the logic. The max amount of eth is not equivalent the the sum of takerAsset amount for all orders, just that of the high end of the orders. Also the the fee orders and target orders cannot be combined into one list. Let's speak more about it in person.

packages/asset-buyer/src/asset_buyer_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+    public getAssetBuyerFromERC20TokenAddress(tokenAddress: string): AssetBuyer {
+        const assetData = assetDataUtils.encodeERC20AssetData(tokenAddress);
+        return this.getAssetBuyerFromAssetData(assetData);
+    }
+    /**
+     * Get a list of all the assetDatas that the instance supports
+     *
+     * @return  An array of assetData strings
+     */
+    public getAssetDatas(): string[] {
```
BMillman19: perhaps we want another public method like this: `doesSupportAssetData(assetData: string): bool`

BMillman19:
What do you think about combining AssetBuyer and AssetBuyerManager into one class?

PR #1069, opened 2018-09-12
Upgrade to TS version of ethers

LogvinovLeon:
## Description

Ethers.js releases the new version with typescript support. This PR upgrades us to that version.
This version is still in beta, but a couple of people at ETHBerlin convinced me that it's solid already.
We also stop using our custom fork for revert reasons as it's now merged upstream.

## Testing instructions

CI build passes

## Types of changes

Dependency upgrade

## Checklist:

*   [x] Prefix PR title with `[WIP]` if necessary.
*   [x] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [x] Add tests to cover changes as needed.
*   [x] Update documentation as needed.
*   [x] Add new entries to the relevant CHANGELOG.jsons.

PR #1090, opened 2018-09-26
[wip] [contract-wrappers] Handle revert with reason when decoding call result

dekz:
We use in-process ganache which throws on an RPC error. This means all of our tests get a nice revert error thrown when testing against ganache. This is not possible with RPC providers and a revert with reason result is returned. Our callAsync doesn't handle this and attempts to decode the revert with reason error log as a successful log, which results in an error while decoding.

This only works with our fork of ethers https://github.com/ethers-io/ethers.js/pull/188 and will need to be re-worked when updating to Ethers.js 4

## Checklist:


LogvinovLeon:
Please rebase off development. https://github.com/0xProject/0x-monorepo/pull/1069 just got merged.

fabioberger:
Ideally we would use a different provider for non-contracts tests that has `vmErrorsOnRPCResponse` set to `false`. @dekz is looking into this approach.

PR #1092, opened 2018-09-26
Add Flag to publish releases command-line tool

fabioberger:
## Description

If the publish flow fails mid-way, it requires publishing the release notes manually. The command-line currently runs `lerna updated` to figure out which packages to generate release notes for. This doesn't work during a manual publish, instead the caller needs to manually specify which packages to publish. This PR adds this flag to the command-line tool.

## Checklist:


LogvinovLeon:
Could you put all the version changes in a separate PR?

PR #1094, opened 2018-09-26
[website] Use coinbase wallet instead of Toshi, fix portal layout, update Clay's title

fragosti:
## Description

Main change is that Toshi has rebranded to Coinbase Wallet, so replace strings, verbiage in the code and logo.

## Checklist:


packages/website/ts/pages/about/about.tsx
```diff
@@ -194,7 +194,7 @@ const teamRow6: ProfileInfo[] = [
 const teamRow7: ProfileInfo[] = [
     {
         name: 'Clay Robbins',
-        title: 'Business Development Lead',
+        title: 'Ecosystem Development Lead',
```
fabioberger: 🙏

PR #1097, opened 2018-09-27
Export ExpirationWatcher from OrderWatcher

LogvinovLeon:
## Description

I think expiration watcher is usefull enough to be used separately. It has a nice clean configurable API.

## Checklist:

### DIFF
--- a/packages/order-watcher/CHANGELOG.json
+++ b/packages/order-watcher/CHANGELOG.json
@@ -1,5 +1,13 @@
 [
     {
+        "version": "2.1.0",
+        "changes": [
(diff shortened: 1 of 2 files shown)


packages/order-watcher/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "2.1.0",
+        "changes": [
+            {
+                "note": "Export ExpirationWatcher"
+            }
```
fabioberger: PR nr

PR #1098, opened 2018-09-27
Remove yarn postinstall hack

albrow:
## Description

This PR fixes a bug in the `postinstall` step introduced in [#1004](https://github.com/0xProject/0x-monorepo/pull/1004/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R40).

It was causing a build failure if you do a full re-install with `yarn remove_node_modules && yarn install && yarn build`:

```
@0xproject/sra-report
 | packages/sra-report/src/postman_environment_factory.ts(98,64): error TS2339: Property 'maker' does not exist on type 'SignedOrder'.
 | packages/sra-report/src/postman_environment_factory.ts(99,64): error TS2339: Property 'taker' does not exist on type 'SignedOrder'.
 | packages/sra-report/src/postman_environment_factory.ts(100,71): error TS2339: Property 'feeRecipient' does not exist on type 'SignedOrder'.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
`yarn build` failed with exit code 1
Stopping 27 active children
Aborted execution due to previous error
error Command failed with exit code 1.
```

This happens because the `sra-report` package calls for ` "@0xproject/types": "^0.8.2"` but we were overriding that with the latest version of `@0xproject/types`, which is incompatible.

For the purposes of this postinstall hack, I think we just need to set up the symlinks. We don't need to copy `@0xproject/types` into any of the sub-packages.

## Testing instructions

Run `yarn remove_node_modules && yarn install && yarn build`.

## Types of changes

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,9 +35,7 @@
         "test": "wsrun test $PKG --fast-exit --serial --exclude-missing",
         "generate_doc": "node ./packages/monorepo-scripts/lib/doc_generate_and_upload.js",
         "test:generate_docs:circleci": "for i in ${npm_package_config_packagesWithDocPages}; do yarn generate_doc --package $i --shouldUpload false --isStaging true || break -1; done;",
-        "lint": "wsrun lint $PKG --fast-exit --parallel --exclude-missing",
-        "comment:postinstall": "HACK: For some reason `yarn` is not setting up symlink properly for types. We temporarily make it manually. Remove this ASAP!",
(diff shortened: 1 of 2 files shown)


albrow:
Update: After talking with @feuGeneA and doing some experimentation, it seems like the postinstall hack is not necessary with the current state of our monorepo. We suspect the changes to the `website` package (especially the version of `@0xproject/types` that it depends on) made the yarn work again.

I've updated this PR to remove the hack altogether. Let's keep an eye on this and revisit if needed. There's also a new version of yarn (1.10) to consider.

PR #1101, opened 2018-09-30
[order-utils] Add signature verification and cancellation check to validateFillOrderThrowIfInvalidAsync

dekz:
## Description

OrderValidationUtils `validateOrderFillableOrThrowAsync ` does not check if an order has been cancelled on chain.

Update this function to also be consistent with other functions in this utility by also checking the signature of the order.

## Checklist:

### DIFF
--- a/packages/order-utils/src/order_validation_utils.ts
+++ b/packages/order-utils/src/order_validation_utils.ts
@@ -109,6 +109,31 @@ export class OrderValidationUtils {
             throw new Error(RevertReason.TransferFailed);
         }
     }
+    private static async _validateSignatureIsValidOrThrowAsync(
+        provider: Provider,
+        orderHash: string,
+        signedOrder: SignedOrder,
+    ): Promise<void> {
+        // Note: isValidSignatureAsync can throw if the format is incorrect
+        try {
+            const isValidSignature = await signatureUtils.isValidSignatureAsync(
+                provider,
+                orderHash,
+                signedOrder.signature,
+                signedOrder.makerAddress,
+            );
+            if (!isValidSignature) {
+                throw new Error(OrderError.InvalidSignature);
+            }
+        } catch (e) {
+            throw new Error(OrderError.InvalidSignature);
+        }
+    }
+    private static _validateOrderNotCancelledOrThrow(isCancelled: boolean): void {
+        if (isCancelled) {
+            throw new Error(RevertReason.OrderCancelled);
+        }
+    }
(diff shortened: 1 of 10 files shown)


packages/order-utils/src/order_validation_utils.ts
```diff
@@ -109,6 +109,31 @@ export class OrderValidationUtils {
...
+                orderHash,
+                signedOrder.signature,
+                signedOrder.makerAddress,
+            );
+            if (!isValidSignature) {
+                throw new Error(OrderError.InvalidSignature);
+            }
+        } catch (e) {
+            throw new Error(OrderError.InvalidSignature);
+        }
```
fabioberger: This feels like something that we should fix about `isValidSignatureAsync`. Can you catch the `format is incorrect` error within it's implementation and return `false` when it happens? Otherwise all callers need to wrap it in a try/catch.

fabioberger:
@dekz I remember @LogvinovLeon and I explicitly deciding not to verify the signature in this method because signature validation is something that happens once per-order, whereas verifying it's fill-ability is something that gets verified many times thereafter. Don't know if we still feel strongly about this however. Thoughts @LogvinovLeon?

dekz:
@fabioberger this has caught a few people out in RocketChat. They're expecting it to completely validate an order is valid. I agree its also a but misleading to not check the signature since that is part of its fillability. 

With v2 its also trickier since the result of a Wallet or SignatureValidator can change (up until the point it has been partially filled).

If we agree its important enough to have a helper which skips signatures but verifies everything else then we should also call this out in the docs so the developer knows to do this themselves. Having an all in one helper is exposed is definitely useful. `validateFillOrderThrowIfInvalidAsync` requires taker details which doesn't make it useful in some scenarios, namely pulling via SRA and ensuring the orders are *entirely* fillable.

dekz:
@fabioberger we also have a orderStateUtils which is useful for finding the state of an order. So perhaps for all OrderState checks we use that, and all OrderValidation we check everything (and make use of OrderState). I have a feeling there is a bit of duplication between this utils which could also be cleaned up.

LogvinovLeon:
I think we can reconsider this for v2 because the signature validation result can now change.

fabioberger:
Ok, I agree we should add signature verification since it's no longer necessarily a simply ECSignature, but could involve checking on-chain state.

fabioberger:
Closing in favor of https://github.com/0xProject/0x-monorepo/pull/1235

PR #1103, opened 2018-10-01
Implement Developer Home

fabioberger:
## Description

Implement the new developer home page. 

- New developer page top-bar
- developer home content, including packages & description copy
- Mobile optimized page & menu

screen shot 2018-10-01 at 13 05 35

screen shot 2018-10-01 at 13 41 19

dev_home

## Checklist:


BMillman19:
Just some random observations as I play with dogfood.0xproject.com:

- Bug: Hover over "developers" and click on "build a relayer", hover over "developers" again and click on "0x order basics", nothing happens
- It would be nice to have the different clickable elements slightly darken to let the user know it is clickable. We have an example in the Text component for how to easily achieve this using styled components the the `polished` package
- Alignment / padding is slightly off for the "Libraries & Tools" list in the docs page (see attached screenshot)
screen shot 2018-10-03 at 3 26 42 pm

packages/react-shared/src/components/link.tsx
```diff
@@ -0,0 +1,101 @@
...
+    to: string;
+    type?: LinkType;
+    shouldOpenInNewTab?: boolean;
+    style?: React.CSSProperties;
+    className?: string;
+    onMouseOver?: (event: React.MouseEvent<HTMLElement>) => void;
+    onMouseLeave?: (event: React.MouseEvent<HTMLElement>) => void;
+    onMouseEnter?: (event: React.MouseEvent<HTMLElement>) => void;
+    containerId?: string;
+}
```
fragosti: nit: Some props like `className` or `containerId` are ignored if a certain `type` prop is used. 

You could do something like (I think? This works in Flow):
```
interface CommonProps {}
interface ScrollLinkProps {
  type: LinkType.ReactRoute;
  containerId?: string
}
interface ExternalLinkProps {
  type: LinkType.External;
  className?: string;
}

export type LinkProps = CommonProps & (ExternalLinkProps | ScrollLinkProps);
```
Then when you set the type prop to something the type system can infer what other props are usable.

Also is the `style` prop necessary?

packages/website/ts/components/documentation/docs_logo.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { WebsitePaths } from 'ts/types';
+
+export interface DocsLogoProps {
+    height: number;
+    containerStyle?: React.CSSProperties;
+}
+
+export const DocsLogo: React.StatelessComponent<DocsLogoProps> = props => {
+    return (
+        <div style={props.containerStyle}>
```
fragosti: can we remove this wrapping div and just use `Container` when we want to alter the position of the `DocsLogo` which is what I assume is going on here?

packages/website/ts/components/documentation/docs_top_bar.tsx
```diff
@@ -0,0 +1,155 @@
...
+        const menuItems = _.map(menuItemInfos, menuItemInfo => {
+            return (
+                <a
+                    key={`menu-item-${menuItemInfo.title}`}
+                    href={menuItemInfo.url}
+                    target="_blank"
+                    className="text-decoration-none"
+                    style={{
+                        fontSize: 16,
+                    }}
```
fragosti: nit: I would remove this inline style and just wrap `menuItemInfo.title` in a `Text`

packages/website/ts/components/documentation/tutorial_button.tsx
```diff
@@ -0,0 +1,79 @@
...
+            >
+                <div
+                    className="flex relative"
+                    style={{
+                        borderRadius: 4,
+                        border: `1px solid ${this.state.isHovering ? colors.lightLinkBlue : colors.grey325}`,
+                        padding: 20,
+                        marginBottom: 15,
+                        backgroundColor: this.state.isHovering ? colors.lightestBlue : colors.white,
+                    }}
```
fragosti: You could have this be a stateless component and remove all the `onMouseLeave` `onMouseEnter` `onMouseOver` logic if it's only being used to modify `isHovering` and `isHovering` is only being used to change styles. With `styled-components` you can use pseudo selectors. See the `<Button/>` component for example.

packages/website/ts/components/dropdowns/developers_drop_down.tsx
```diff
@@ -118,27 +118,26 @@ export class DevelopersDropDown extends React.Component<DevelopersDropDownProps,
...
-                        borderBottomRightRadius: 4,
+                        color: colors.lightBlueA700,
+                        fontWeight: 'bold',
+                        fontSize: 14,
                     }}
                 >
-                    <Link
-                        to={WebsitePaths.ZeroExJs}
-                        className="text-decoration-none"
+                    <div
```
fragosti: nit: could be `Container`

fabioberger:
@BMillman19 addressed all your feedback, minus the onHover darkening of links. Adding that to V2 of dev docs.

fabioberger:
@fragosti re: scrollbar positioning, ideally not. I haven't found a fool-proof way of both having the 1024px width of the page centered at all times, and also have the scrollbar always snug with the right-side of the window, through window re-sizing and all. Happy to talk with you about potential ways to achieve that.

packages/react-shared/src/components/link.tsx
fabioberger: Hm, tried your approach but it was giving me issues with the `defaultProps` instance object. I ended up figuring out a way to get all props to apply to all link types, **except** for `shouldOpenInNewTab` & `linkType.ReactScroll`, which should never be passed in together. For that one, I added an assert that throws. This type of mistake will now be caught during development.

fabioberger: I went ahead and removed the `style` prop.

PR #1105, opened 2018-10-02
[WIP] Reorganize and modularize generated contract wrappers and artifacts

albrow:
## Description

This PR is a huge reorganization of how we generate, store, and copy generated files related to contracts (specifically artifacts and contract wrappers generated by abi-gen). 

## Checklist:


packages/contracts/package.json
```diff
@@ -1,38 +1,36 @@
 {
-    "private": true,
-    "name": "contracts",
-    "version": "2.1.48",
+    "name": "@0xproject/contracts",
+    "version": "2.0.0",
```
fabioberger: Hm, what's the rationale here? Lerna is going to publish `patch` updates of this package whenever it's deps update. Don't think it's possible to have this version track the "contract" version. Rather we should maybe add a second key called `"0xVersion: "2.0.0"`

packages/contracts/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './artifacts';
+import * as wrappers from './wrappers';
+export { wrappers };
```
fabioberger: Also, in `contracts` let's add a `.npmignore` file and make sure the bundle published to `npmjs.com` repository only includes these files and not the `tests` and `Solidity` source dirs.

albrow: 👍 I was planning to sort what to include in the published package out at the end. Want to make sure everything works inside the monorepo first.

albrow:
@LogvinovLeon the goal of this PR is to have all artifacts and generated contract wrappers in a single canonical location. Other packages that need those things will import the canonical package instead of copying files or calling `abi-gen` multiple times.

I'm planning to have the `contracts` package be that canonical location and will remove the actual artifacts from the `migrations` package. The reasoning for putting them in `contracts` instead of `migrations` is that it keeps the boundaries between packages more clear. If the artifacts were stored in `migrations`, we would need to have `build` and/or `clean` steps that either modify files in other packages or call scripts or reference files in other packages, which can lead to some unintuitive results and confusing build-order dependencies.

packages/contracts/package.json
albrow: I was thinking it would be cleaner to start at `2.0.0` and it shouldn't matter since this package has never been published before.

That said, you're right that in the long-term this version will probably not track the "contract" version exactly. I changed my mind and I think it's probably better to keep the same version number we had before. I'll change it back to `2.1.48`.

fabioberger:
Reminder: make sure to add `CHANGELOG.json` and `CHANGELOG.md` files to the `contracts` package.

albrow:
I added a summary of major changes to this PR. In terms of progress:

1. The `@0xproject/contracts` package has been created with the structure described above.
2. `@0xproejct/migrations` has been updated as described above. I removed the artifacts and added the new `getContractsAddress` function. Migrations also operate on only one version of the contracts at a time so I removed the v1 artifacts and related migrations.
3. `@0xproejct/contract-wrappers` has been updated to use the new locations for artifacts and abi-gen'ed contract wrappers.

The biggest things I still need to do are:

1. Update any other packages that rely on old locations for artifacts or abi-gen'ed contract wrappers.
2. Clean up `@0xproject/contracts` and get it ready to publish. This means only exporting what we need to export, adding things to __.npmignore__ and other config files, creating __CHANGELOG.json__, etc. I also want to double check bundle sizes and make sure they're manageable.
3. Make sure all the documentation is up to date.

albrow:
Updated high-level summary of changes:

1. All artifacts have been moved out of `@0xproject/migrations`. Bleeding edge development artifacts have been moved to the `contracts` package, where they are kept private and are not imported by any other packages. On the other hand, permanent, versioned artifacts are available in the new `@0xproject/contract-artifacts` package. They are encapsulated in an npm package and have a version number indicated by a __package.json__ file. You can import artifacts with `import * as artifacts from '@0xproject/contract-artifacts'`. (You can also import individual artifacts directly). This approach allows us to stop manually copying around artifact files to the packages that need them.
2. Abi-gen'ed contract wrappers have been moved into the new`@0xproject/abi-gen-wrappers` package. Like artifacts, they are encapsulated in an npm package. You can import them with `import * as wrappers from '@0xproject/abi-gen-wrappers'`. (You can also import individual wrappers directly). This approach allows us to avoid re-generating the same wrappers across many different packages.
3. Addresses of deployed contracts are no longer stored in artifact files. For tests, the `runMigrationsAsync` function in `@0xproject/migrations` simply returns the contract addresses. For contracts that have been deployed to the mainnet or various testnets, the addresses will be exposed as an object in one of our public packages (exact details TBD). This approach leads to a better separation between things that are temporary (addresses of contracts that were deployed for testing purposes) and permanent (addresses, abis, and source code of contracts that have already been deployed to mainnet or a testnet).

albrow:
Okay I think I have an approach that works now. Most of what's left to do is bookkeeping:

- [ ] Resolve merge conflicts.
- [ ] Add __README.md__ and __CHANGELOG.json__ to all three new packages.
- [ ] Update all __CHANGELOG.json__ for any changed packages.
- [ ] Update relevant documentation.
- [ ] Move v1 contracts back into contracts package (required for sol-doc).
- [ ] Update __.prettierignore__, __.gitignore__, and other ignore files.

packages/contract-artifacts/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import * as Exchange from '../artifacts/Exchange.json';
+import * as Forwarder from '../artifacts/Forwarder.json';
+import * as IValidator from '../artifacts/IValidator.json';
+import * as IWallet from '../artifacts/IWallet.json';
+import * as OrderValidator from '../artifacts/OrderValidator.json';
+import * as WETH9 from '../artifacts/WETH9.json';
+import * as ZRXToken from '../artifacts/ZRXToken.json';
+
+export {
+    AssetProxyOwner,
```
albrow: Note that we don't have to do the `(as any) as ContractArtifact` hack anymore! With `"resolveJsonModule": true` in __tsconfig.json__ TypeScript supports typed JSON imports. It also works with Project References, so when you watch, it will recompile when the JSON file(s) change.

packages/contract-wrappers/src/contract_wrappers/erc20_token_wrapper.ts
```diff
@@ -435,13 +435,15 @@ export class ERC20TokenWrapper extends ContractWrapper {
         if (!_.isUndefined(tokenContract)) {
             return tokenContract;
         }
-        const [abi, address] = await this._getContractAbiAndAddressFromArtifactsAsync(
-            artifacts.ERC20Token,
-            normalizedTokenAddress,
-        );
+        // TODO(albrow): Do we really still need this check? The default error
```
albrow: @LogvinovLeon or @fabioberger thoughts on this TODO? (Also appears a few other places).

packages/contract-wrappers/test/utils/token_utils.ts
```diff
@@ -1,14 +1,13 @@
...
 
-import { artifacts } from '../../src/artifacts';
-import { DummyERC721TokenContract } from '../../src/contract_wrappers/generated/dummy_erc721_token';
-
-import { constants } from './constants';
 import { provider, txDefaults, web3Wrapper } from './web3_wrapper';
 
 // Those addresses come from migrations. They're deterministic so it's relatively safe to hard-code them here.
 // Before we were fetching them from the TokenRegistry but now we can't as it's deprecated and removed.
+// TODO(albrow): Import these from the migrations package instead of hard-coding them.
```
albrow: I'm going to leave this for now. We're no worse off than we were before this PR and can always fix it later. This PR is already arguably too big as it is.

packages/contract-wrappers/test/revert_validation_test.ts
```diff
@@ -112,7 +107,7 @@ describe('Revert Validation ExchangeWrapper', () => {
                 makerTokenBalance,
             );
             await web3Wrapper.awaitTransactionSuccessAsync(txHash, constants.AWAIT_TRANSACTION_MINED_MS);
-            expect(
+            return expect(
```
albrow: This is subtle, but there was a race condition here before adding the `return`.

packages/contract-wrappers/test/exchange_wrapper_test.ts
```diff
@@ -454,13 +458,6 @@ describe('ExchangeWrapper', () => {
             })().catch(done);
         });
     });
-    describe('#getZRXTokenAddressAsync', () => {
```
albrow: Since we removed the token registry, this test isn't really doing anything.

packages/contract-wrappers/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -124,39 +112,8 @@ export abstract class ContractWrapper {
         const logWithDecodedArgs = abiDecoder.tryToDecodeLogOrNoop(log);
         return logWithDecodedArgs;
     }
-    protected async _getContractAbiAndAddressFromArtifactsAsync(
```
albrow: Since we aren't reading addresses from a file anymore, I was able to remove this function and others like it. For `ContractWrappers` addresses are now known at instantiation time and are synchronous, which simplifies a lot of our code.

packages/contract-addresses/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    orderValidator: string;
+}
+
+export enum NetworkId {
+    Mainnet = 1,
+    Ropsten = 3,
+    Kovan = 42,
+}
+
+const networkToAddresses: { [networkId: number]: ContractAddresses } = {
```
fabioberger: Rather then `{ [networkId: number]: ContractAddresses }`, let's use `ObjectMap<ContractAddresses>`

packages/contract-addresses/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+        assetProxyOwner: '0x2c824d2882baa668e0d5202b1e7f2922278703f8',
+        forwarder: '0xd85e2fa7e7e252b27b01bf0d65c946959d2f45b8',
+        orderValidator: '0xb389da3d204b412df2f75c6afb3d0a7ce0bc283d',
+    },
+};
+
+/**
+ * Used to get addresses of contracts that have been deployed to either the
+ * Ethereum mainnet or a supported testnet.
+ * @returns The set of addresses for contracts which have been deployed on the
```
fabioberger: `@param networkId` missing.

packages/contract-artifacts/package.json
```diff
@@ -0,0 +1,35 @@
+{
+    "name": "@0xproject/contract-artifacts",
+    "version": "1.0.0",
+    "engines": {
+        "node": ">=6.12"
+    },
+    "description": "Smart contract components of 0x protocol",
```
fabioberger: "0x smart contract compilation artifacts"

packages/contract-artifacts/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "bugs": {
+        "url": "https://github.com/0xProject/0x-monorepo/issues"
+    },
+    "homepage": "https://github.com/0xProject/0x-monorepo/packages/contract-artifacts/README.md",
+    "devDependencies": {
+        "typescript": "3.0.1",
+        "shx": "^0.2.2"
+    },
+    "dependencies": {
+    },
```
fabioberger: Omit if empty.

packages/contract-artifacts/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+    "extends": "../../tsconfig",
+    "compilerOptions": {
+        "outDir": "lib",
+        "rootDir": ".",
+        "resolveJsonModule": true
+    },
+    "include": ["./globals.d.ts", "./src/**/*"],
```
fabioberger: Where is the `globals.d.ts`?

packages/contract-wrappers/package.json
```diff
@@ -72,8 +65,11 @@
         "web3-provider-engine": "14.0.6"
     },
     "dependencies": {
+        "@0xproject/abi-gen-wrappers": "^1.0.0",
         "@0xproject/assert": "^1.0.13",
         "@0xproject/base-contract": "^3.0.1",
```
fabioberger: Do we still need `base-contract`? Think we can remove this too.

packages/contract-wrappers/src/contract_wrappers.ts
```diff
@@ -113,34 +129,39 @@ export class ContractWrappers {
...
+        );
+        this.orderValidator = new OrderValidatorWrapper(
+            this._web3Wrapper,
+            config.networkId,
+            contractAddresses.orderValidator,
         );
-        this.orderValidator = new OrderValidatorWrapper(this._web3Wrapper, config.networkId);
     }
     /**
      * Sets a new web3 provider for 0x.js. Updating the provider will stop all
```
fabioberger: Not your issue, but mind swapping `0x.js` with `contract-wrappers`

packages/contract-wrappers/src/contract_wrappers.ts
```diff
@@ -113,34 +129,39 @@ export class ContractWrappers {
...
      * subscriptions so you will need to re-subscribe to all events relevant to your app after this call.
      * @param   provider    The Web3Provider you would like the 0x.js library to use from now on.
```
fabioberger: `contract-wrappers`

packages/contract-wrappers/src/contract_wrappers/erc20_proxy_wrapper.ts
```diff
@@ -1,38 +1,39 @@
...
+    public address: string;
     private _erc20ProxyContractIfExists?: ERC20ProxyContract;
-    private _contractAddressIfExists?: string;
     /**
      * Instantiate ERC20ProxyWrapper
      * @param web3Wrapper Web3Wrapper instance to use
      * @param networkId Desired networkId
-     * @param contractAddressIfExists The contract address to use. This is usually pulled from
-     * the artifacts but needs to be specified when using with your own custom testnet.
+     * @param address (Optional) The address of the ERC20Proxy contract. If
```
fabioberger: Let's omit `(Optional)` since the `?` in TS denotes that it is optional. Worried we'd update one and not the other in the future.

packages/contract-wrappers/src/contract_wrappers/contract_wrapper.ts
```diff
@@ -124,39 +112,8 @@ export abstract class ContractWrapper {
...
-            const contractAddress = artifact.networks[this._networkId].address;
-            if (_.isUndefined(contractAddress)) {
-                throw new Error(CONTRACT_NAME_TO_NOT_FOUND_ERROR[artifact.contractName]);
-            }
-            return contractAddress;
-        } else {
-            return addressIfExists;
-        }
+    protected _getDefaultContractAddresses(): ContractAddresses {
+        return getContractAddressesForNetwork(this._networkId);
```
fabioberger: What is `_networkId` is a networkId we do not support? E.g someone of their own testnet 999?

packages/contract-addresses/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * given networkId.
+ */
+export function getContractAddressesForNetwork(networkId: NetworkId): ContractAddresses {
```
fabioberger: Can we rename to `getContractAddressesForNetworkOrThrow`? It makes it more explicit that if the `networkId` isn't recognized, this method throws.

packages/contract-wrappers/src/contract_wrappers/erc20_proxy_wrapper.ts
```diff
@@ -1,38 +1,39 @@
...
+     * undefined, will default to the known address corresponding to the
+     * networkId.
      */
-    constructor(web3Wrapper: Web3Wrapper, networkId: number, contractAddressIfExists?: string) {
+    constructor(web3Wrapper: Web3Wrapper, networkId: number, address?: string) {
         super(web3Wrapper, networkId);
-        this._contractAddressIfExists = contractAddressIfExists;
+        this.address = _.isUndefined(address) ? this._getDefaultContractAddresses().erc20Proxy : address;
```
fabioberger: I think we should wrap the call to `_getDefaultContractAddresses` in a try/catch and re-throw a more situation specific error message here. Perhaps: 

```
throw new Error(`'address' must be supplied to ERC20ProxyWrapper when instantiated with an unrecognized networkId (${networks.join(',')})`);
```

I find just allowing the base error to bubble up can be a source of a lot of confusion since it doesn't directly relate to what the user is currently doing.

packages/contract-wrappers/src/contract_wrappers/erc20_token_wrapper.ts
fabioberger: What is the default error? `doesContractExistAtAddressAsync` returns a boolean.

packages/contract-wrappers/src/contract_wrappers/ether_token_wrapper.ts
```diff
@@ -215,13 +202,16 @@ export class EtherTokenWrapper extends ContractWrapper {
         if (!_.isUndefined(etherTokenContract)) {
             return etherTokenContract;
         }
-        const [abi, address] = await this._getContractAbiAndAddressFromArtifactsAsync(
-            artifacts.EtherToken,
-            etherTokenAddress,
-        );
+        // TODO(albrow): Do we really still need this check? The default error
+        // looks okay to me.
+        // TODO(albrow): Should we normalize the token address here?
```
fabioberger: What do you mean by this?

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -35,43 +35,48 @@ import { TransactionEncoder } from '../utils/transaction_encoder';
...
+        address?: string,
+        zrxTokenAddress?: string,
         blockPollingIntervalMs?: number,
     ) {
         super(web3Wrapper, networkId, blockPollingIntervalMs);
         this._erc20TokenWrapper = erc20TokenWrapper;
         this._erc721TokenWrapper = erc721TokenWrapper;
-        this._contractAddressIfExists = contractAddressIfExists;
-        this._zrxContractAddressIfExists = zrxContractAddressIfExists;
+        this.address = _.isUndefined(address) ? this._getDefaultContractAddresses().exchange : address;
```
fabioberger: We need a more intelligible error message here too. Thinking about this again, what we should really do is **assert** that if `networkId` isn't one of our supported addresses, that `address` **is** supplied.

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1159,30 +1162,12 @@ export class ExchangeWrapper extends ContractWrapper {
...
-        return contractAddress;
-    }
-    /**
-     * Returns the ZRX token address used by the exchange contract.
-     * @return Address of ZRX token
-     */
-    public getZRXTokenAddress(): string {
-        const contractAddress = this._getContractAddress(artifacts.ZRXToken, this._zrxContractAddressIfExists);
-        return contractAddress;
-    }
```
fabioberger: Hm... any reason to remove these? We could just return it no? It's a big breaking change to remove these... I know relayers rely on that. Do we want to force them to move to using `contract-addresses` package?

packages/contract-wrappers/src/contract_wrappers/order_validator_wrapper.ts
```diff
@@ -1,30 +1,35 @@
...
+    public abi: ContractAbi = OrderValidator.compilerOutput.abi;
+    public address: string;
     private _orderValidatorContractIfExists?: OrderValidatorContract;
     /**
      * Instantiate OrderValidatorWrapper
-     * @param web3Wrapper Web3Wrapper instance to use
-     * @param networkId Desired networkId
+     * @param web3Wrapper Web3Wrapper instance to use.
+     * @param networkId Desired networkId.
+     * @param address (Optional) The address of the OrderValidator contract. If
```
fabioberger: remove `(optional)`

packages/contract-wrappers/test/revert_validation_test.ts
fabioberger: Can we add a TSLint rule for this? If not now, let's add a task to the backlog.

packages/website/ts/blockchain.ts
```diff
@@ -867,9 +868,13 @@ export class Blockchain {
             shouldUserLedgerProvider,
         );
         if (!_.isUndefined(this._contractWrappers)) {
-            this._contractWrappers.setProvider(provider, networkId);
+            this._contractWrappers.setProvider(provider);
         } else {
-            this._contractWrappers = new ContractWrappers(provider, { networkId });
+            const contractWrappersConfig = {
+                networkId,
+                contractAddresses: getContractAddressesForNetwork(networkId),
```
fabioberger: Omit this. It defaults to doing this internally.

tsconfig.json
```diff
@@ -22,10 +22,12 @@
     "references": [
         { "path": "./packages/0x.js" },
         { "path": "./packages/abi-gen" },
+        { "path": "./packages/abi-gen-wrappers" },
         { "path": "./packages/assert" },
         { "path": "./packages/asset-buyer" },
         { "path": "./packages/base-contract" },
         { "path": "./packages/connect" },
+        { "path": "./packages/contract-artifacts" },
```
fabioberger: Do we need an entry here for `contract-addressses`?

packages/contract-artifacts/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+    "extends": "../../tsconfig",
+    "compilerOptions": {
+        "outDir": "lib",
+        "rootDir": ".",
+        "resolveJsonModule": true
```
LogvinovLeon: When I tried to do that I've hit a problem that TSC auto-generates types for JSON files from their content and we had some partial artifacts in some modules, but I guess now that you standardized it  - it's not a problem any more. Happy about that.

.gitignore
```diff
@@ -78,22 +78,21 @@ packages/react-docs/example/public/bundle*
...
-# generated contract watcher
+# generated contract wrappers
+packages/abi-gen-wrappers/wrappers
 packages/0x.js/src/generated_contract_wrappers/
-packages/contracts/generated_contract_wrappers/
-packages/contract-wrappers/src/contract_wrappers/generated/
+packages/contracts/generated-wrappers/
 packages/metacoin/src/contract_wrappers
 packages/fill-scenarios/src/generated_contract_wrappers/
 packages/order-watcher/src/generated_contract_wrappers/
```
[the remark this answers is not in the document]
albrow: 👍  I have a TODO item on this PR to fix __.gitignore__

packages/abi-gen-wrappers/src/index.ts
```diff
@@ -0,0 +1,14 @@
+export * from '../wrappers/asset_proxy_owner';
```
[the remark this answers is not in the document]
albrow: IMO `src` should be for user-editable source code. These wrappers are generated code. What do you think?

packages/contract-addresses/src/index.ts
albrow: I don't think we need to add the `OrThrow` suffix to every function that could throw. Is there a specific reason why we should add it here?

packages/contract-wrappers/src/contract_wrappers/erc20_token_wrapper.ts
albrow: What I'm asking is whether we really need the `ContractWrappersError.ERC20TokenContractDoesNotExist` error. If we remove this check, `Web3Wrapper` throws an error along the lines of `"Tried to send a transaction to 0xabcde... but it is not a contract address."`. Together with a stack trace, I think this makes it pretty clear what went wrong. What do you think?

packages/contract-wrappers/test/revert_validation_test.ts
albrow: ➕ Good idea. I believe we already have a check for unhandled promise rejections. The problem is that `expect().to.be.rejectedWith` doesn't return a real `Promise`. Instead it returns a `Chai.PromisedAssertion`. There must be some way to check for it.

packages/contract-addresses/src/index.ts
```diff
@@ -0,0 +1,64 @@
+import * as _ from 'lodash';
+
+export interface ContractAddresses {
```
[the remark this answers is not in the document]
albrow: Hmm.. I'm not sure I agree with renaming it. In this case, I think the kind of information we are passing around (the addresses of deployed contracts) is more important to call out than the underlying data structure (a mapping of contract name to contract address). For similar reasons, it seems awkward to rename `getContractAddressesForNetwork` to `getContractToAddressForNetwork` or rename the `contractAddresses` parameter in the constructor for `ContractWrappers` to `contractToAddress`.

To look at it from another perspective, here are a few examples of how we use the `XToY` naming convention in `@0xproject/types`: 

```
export interface ExportNameToTypedocNames {
    [exportName: string]: string[];
}

export interface ExternalTypeToLink {
    [externalTypeName: string]: string;
}

export interface ExternalExportToLink {
    [externalExport: string]: string;
}
```

All of these cases represent a mapping of *arbitrary*  X to Y, and that's where the naming convention makes the most sense to me. Since we don't know ahead of time exactly what keys the type will hold (and whether other uses of the type will have the same keys), the underlying data structure is more important to call out. In other words, the data structure itself is what defines the type, not the data it holds.

All that said, if you feel strongly about it, I can be convinced.

packages/contract-addresses/src/index.ts
albrow: Here, `networkId` is a `number` so it's not the same thing as `ObjectMap<ContractAddresses>`. I think we should keep it as a `number` because that is how it appears in the constructors in `@0xproject/contract-wrappers` and in a few other places.

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
albrow: With the changes in this PR, `zrxTokenAddress` is simply a property of `ExchangeWrapper`, so you don't have to use the `contract-addresses` package if you don't want to. Removing this method simplifies our API surface and will make `contract-wrappers` easier to use/understand. Leaving it in would mean that there are now two ways to get the address and just creates a little bit of clutter.

Given that AFAIK we don't have any mechanism for marking functions/methods as deprecated, I think we should just remove outdated code like this.

If this was the only breaking change, I would leave it in. However, we already have some breaking changes to the `contract-wrappers` package in this PR and this one is definitely on the easy side in terms of updating.

packages/asset-buyer/src/asset_buyer.ts
```diff
@@ -109,6 +110,7 @@ export class AssetBuyer {
         this.expiryBufferSeconds = expiryBufferSeconds;
         this._contractWrappers = new ContractWrappers(this.provider, {
             networkId,
+            contractAddresses: getContractAddressesForNetwork(networkId),
```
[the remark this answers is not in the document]
albrow: Omitted for now. We can add it back in if we think users want to configure it.

packages/contract-artifacts/tsconfig.json
albrow: Just removed this. `global.d.ts` was required before I realized we could use `resolveJsonModules` here.

packages/contract-wrappers/src/contract_wrappers/erc20_proxy_wrapper.ts
albrow: @fabioberger Good idea. I gathered your feedback here and from other comments and added some additional logic. See if you like the way I handle this error in [the latest commit](https://github.com/0xProject/0x-monorepo/pull/1105/commits/ae94f8cdb3b15209f35517c32cf03c278b68db80).

packages/contract-artifacts/tsconfig.json
albrow: Yes, I still had to include some type assertions in the `contracts` package for the reasons you mentioned. I think it only works for `contract-artifacts` right now because we are using the trimmed down contract artifacts there.

packages/abi-gen-wrappers/src/index.ts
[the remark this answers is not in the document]
fabioberger: Hm... I thought of `src` and where all the TS code lived that needs to be compiled into `lib`. Those generated contracts have disclaimers in their headers, and we could also name the directory to `generated_wrappers`.

packages/contract-addresses/src/index.ts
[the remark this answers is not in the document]
fabioberger: Yeah, I see your point in this case. Thanks for the thoughtful explanation, let's leave it the way it is.

packages/contract-addresses/src/index.ts
fabioberger: Agreed, but the reason why I think it makes sense to add it here is because it will throw if the `contractAddresses` do not **exist** for the supplied `networkId`.

Common alternatives to throwing in this case are to return `undefined`, `null` or perhaps given this functions call signature, an object that conforms to the `ContractAddresses` interface but has all it's object values set to `undefined`. 

I think it makes sense to add the `OrThrow` prefix to `getter` methods that communicate the failure to retrieve something by throwing. 

Thoughts?

albrow: Fair enough. In general I would prefer less verbose naming conventions, but I understand your point in this case. I'll change it.

(As an aside, I really wish there was a better conventional way to communicate synchronous failure modes in the TypeScript type system. E.g. in Go you would communicate this by having a second return value of type `error`, or in Haskell you would use an `Either` monad).

packages/monorepo-scripts/src/utils/doc_generate_and_upload_utils.ts
```diff
@@ -103,6 +103,9 @@ export class DocGenerateAndUploadUtils {
             switch (node.kind) {
                 case ts.SyntaxKind.ExportDeclaration: {
                     const exportClause = (node as any).exportClause;
+                    if (_.isUndefined(exportClause)) {
```
albrow: @fabioberger commenting here to make sure you get a chance to review the changes I made to this file. It's easy to miss in a PR of this size.

packages/contract-wrappers/src/contract_wrappers/erc20_token_wrapper.ts
fabioberger: Yep, let's get rid of the mapping.

packages/abi-gen-wrappers/src/index.ts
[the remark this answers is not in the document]
albrow: 👍 Okay I'll change it.

albrow:
Removed WIP label. Ready for final review.

PR #1108, opened 2018-10-02
[WIP] [sol-compiler] persist artifacts with only relevant sources

feuGeneA:
## Description
[Fix bug where all source codes are included in each artifact.](https://app.asana.com/0/684263176955174/842516551768097/f)

> @albrow  [8:07 PM]
> Okay as far as I know, the issue first appeared after your PR for compiling contract code in batches: https://github.com/0xProject/0x-monorepo/pull/965
> What appears to be happening is that the “sources” and “sourceCodes” field for each artifact includes *all* the source code for all other contracts. This is making artifact files much much bigger than they need to be.

I modified the artifact persistence logic to include only the source codes for the target artifact (and its dependencies (recursively)), out of the larger set of compilation outputs, which includes all of the other contracts compiled together in that run of the compiler.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-compiler/src/compiler.ts
+++ b/packages/sol-compiler/src/compiler.ts
@@ -296,13 +296,12 @@ export class Compiler {
         compilerOutput: solc.StandardOutput,
     ): Promise<void> {
         const compiledContract = compilerOutput.contracts[contractPath][contractName];
-        const sourceCodes = _.mapValues(
-            compilerOutput.sources,
-            (_1, sourceFilePath) => this._resolver.resolve(sourceFilePath).source,
-        );
+
+        const { sourceCodes, sources } = this._getSourcesWithDependencies(contractPath);
+
         const contractVersion: ContractVersionData = {
             compilerOutput: compiledContract,
-            sources: compilerOutput.sources,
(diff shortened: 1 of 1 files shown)


feuGeneA:
Tests failing, my fault, WiP...

packages/sol-compiler/src/compiler.ts
```diff
@@ -333,6 +332,69 @@ export class Compiler {
         await fsWrapper.writeFileAsync(currentArtifactPath, artifactString);
         logUtils.warn(`${contractName} artifact saved!`);
     }
+    private _getSourcesWithDependencies(
+        contractPath: string,
+    ): { sourceCodes: { [sourceName: string]: string }; sources: { [sourceName: string]: { id: number } } } {
+        const sources = { [contractPath]: { id: 0 } };
```
LogvinovLeon: Could you please verify that the IDs generated that way don't break the source maps? Just check if the coverage still works with those artifacts. Source maps rely on ids geenrated by solidity compiler so you can't just make them auto-increment. Or at least if you do - you'll also need to rewrite the source maps with those new IDs

PR #1112, opened 2018-10-03
Add 0x Instant to bundle analysis

fragosti:
## Description

OK this PR turned out simpler than expected. I also wanted to create a CDN and upload assets to that CDN post-publish but we can just use unpkg for now. The UMD bundles will be available in the `public/` directory (along with the index.html file which will provide a live demo of the bundle per version which is pretty cool).

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,9 @@
     "bundlesize": [
         {
             "path": "packages/0x.js/_bundles/index.min.js"
+        },
+        {
+            "path": "packages/instant/public/main.bundle.js"
         }
     ],
     "devDependencies": {


BMillman19:
@fragosti So this only adds it to the bundle size CI job, correct? Publishing and showing up on unpkg is taken care of the `config.postpublish.assets` field in the instant package.json?

fragosti:
@BMillman19 everything that is not in .npmignore ends up being part of the package zip. This includes the `public` dir. The `config.postpublish.assets` field is misleading and is very github release notes specific.

PR #1113, opened 2018-10-03
[asset-buyer] Rework min / max rate calc and add tests

BMillman19:
## Description

* Updated logic in the `buy_quote_calculator` to more accurately calculate min and max rate
* Added tests for `buy_quote_calculator`

## Testing instructions

`yarn test` in the `asset-buyer` package

## Types of changes

## Checklist:


BMillman19:
@fragosti yup it just changes the math a bit, but arrives at the same rate calculation, which is expected

PR #1114, opened 2018-10-04
[instant] Add styles and redux to instant

fragosti:
## Description

I just added more structure to the 0x instant package. 
The button and text components were copied from website but are more restrictive with their props as they only allow a certain enumeration of colors.

I'm using the `ThemeProvider` from `styled-components` as a way to make sure we can:
1. Support a certain level of customizability. 
2. Can change the `primaryColor` of the UI for different tokens (which I think we are still planning on doing)?

There are some open questions regarding how much to share between instant and website. Moving things to `react-shared` isn't super straightforward because of the different `defaultProps`. Basically if we were to move things to `react-shared` we would have to write wrapper components around everything we use. Thoughts here appreciated. 

screen shot 2018-10-04 at 7 30 14 pm

## Checklist:


BMillman19:
@fragosti I agree, doesn't make too much sense to maintain it in `react-shared`, especially if we don't have a set design consistency between the website and instant (even less so with the design agency changing up the home page)

packages/instant/src/components/amount_input.tsx
```diff
@@ -0,0 +1,42 @@
+import { BigNumber } from '@0xproject/utils';
+import * as _ from 'lodash';
```
steveklebanoff: Per convo, added task here to look into if we're shipping all of lodash with instant: https://app.asana.com/0/710162581425378/860502394999228/f

packages/instant/src/components/amount_input.tsx
```diff
@@ -0,0 +1,42 @@
...
+            </Container>
+        );
+    }
+    private readonly _handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+        const value = event.target.value;
+        let bigNumberValue;
+        if (!_.isEmpty(value)) {
+            bigNumberValue = new BigNumber(event.target.value);
+        }
+        if (!_.isUndefined(this.props.onChange)) {
```
steveklebanoff: Based on the code here, I think if you change the AmountInput's value to be "empty", i.e. changing from "300" to "", we will not trigger the `this.props.onChange` event.   Seems OK, but just wanted to confirm that this is desired behavior.  

Something that's a little funky, is that `_.isEmpty` returns `true` for `''` but `false` for  `'       '`

packages/instant/src/components/order_details.tsx
```diff
@@ -0,0 +1,62 @@
...
+                letterSpacing="1px"
+                fontColor={ColorOption.primaryColor}
+                fontWeight={600}
+                textTransform="uppercase"
+                fontSize="14px"
+            >
+                Order Details
+            </Text>
+        </Container>
+        <OrderDetailsRow name="Token Price" primaryValue=".013 ETH" secondaryValue="$24.32" />
```
steveklebanoff: Nice generic component! 💯

packages/instant/src/components/ui/container.tsx
```diff
@@ -0,0 +1,64 @@
...
+    hasBoxShadow?: boolean;
+}
+
+const PlainContainer: React.StatelessComponent<ContainerProps> = ({ children, className }) => (
+    <div className={className}>{children}</div>
+);
+
+export const Container = styled(PlainContainer)`
+    box-sizing: border-box;
+    ${props => cssRuleIfExists(props, 'display')}
```
steveklebanoff: nice helper function here 💯

packages/instant/src/components/ui/flex.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+
+import { ColorOption, styled } from '../../style/theme';
+import { cssRuleIfExists } from '../../style/util';
+
+export interface FlexProps {
+    direction?: 'row' | 'column';
```
steveklebanoff: Just a thought:  instead of explicitly defining the possible values for these properties, we could get them from https://github.com/frenic/csstype which `React.CSSProperties` uses internally:

image

I'm not convinced this is worth doing though

packages/instant/src/components/ui/input.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const PlainInput: React.StatelessComponent<InputProps> = ({ value, className, placeholder, onChange }) => (
+    <input className={className} value={value} onChange={onChange} placeholder={placeholder} />
+);
+
+export const Input = styled(PlainInput)`
+    font-size: ${props => props.fontSize};
+    width: ${props => props.width};
+    padding: 0.1em 0em;
+    font-family: 'Inter UI';
```
steveklebanoff: Would it make sense for this font to be part of the theme?

packages/instant/src/types.ts
```diff
@@ -0,0 +1,9 @@
+export enum ActionTypes {
+    UPDATE_ETH_USD_PRICE,
+    UPDATE_SELECTED_ASSET_AMOUNT,
+}
+
+export interface Action {
+    type: ActionTypes;
+    data?: any;
```
steveklebanoff: I'm not that familiar with redux, so excuse the ignorance here, but I'm wondering: but does this mean that you can trigger an ActionType with any type of `data`?  If so, is there any way to be more specific, to only allow you to only send trigger certain actions with certain data types?

steveklebanoff:
Overall looks awesome 👍  Added a few comments and thoughts

packages/instant/src/components/amount_input.tsx
fragosti: Are we not? Seems like onChange is called for as long as it's defined. It will just be called with undefined as the value since there is no `undefined` or `empty` value for `BigNumber`

packages/instant/src/style/util.ts
```diff
@@ -0,0 +1,10 @@
+import * as _ from 'lodash';
+
+export const cssRuleIfExists = (props: any, rule: string): string => {
```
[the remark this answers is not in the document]
fragosti: Hmmm ok I think there is something to using the React.CSS properties for props, however we can't use it for the rule as we are defining that in `dash-case`. I guess we could start defining it in camelCase and then converting it to dash-case for the actual rule.. but then we'd have to import / define a method to do that etc... etc..

packages/instant/src/components/ui/input.tsx
fragosti: I think not for now... we actually only have 1 font in the entire widget I think

packages/instant/src/types.ts
fragosti: Good point -- I mean, in non-typed environments yes data is kind of anything and then the reducer  is the one that knows that for a certain Action.type the data is going to be of a certain form.

I think it's possible with union types, so you would end up with something like:
```
type Action = BooleanAction | NumberAction ;

interface BooleanAction {
 type: ActionTypes.UpdateBooleanValueFoo | ActionTypes.UpdateBooleanValueBar;
 data: boolean 
}
```
But doesn't seem worth it?

packages/instant/src/components/ui/flex.tsx
fragosti: This is cool!

packages/instant/src/types.ts
BMillman19: There might be some way to do it with generics that doesn't drastically increase the boilerplate.

I skimmed thru some of this: https://medium.com/@martin_hotell/improved-redux-type-safety-with-typescript-2-8-2c11a8062575

I'm fine with merging as is, especially because we already use this convention in website and it hasn't caused any problems, even with all the different action types there

packages/instant/src/components/amount_input.tsx
steveklebanoff: Whoops, read the code incorrectly -- you are correct.

Per convo, did discover an issue while testing this though: if you type in an invalid number (i.e. letters or a bunch of blank spaces), a JS error is thrown

packages/instant/src/components/amount_input.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export class AmountInput extends React.Component<AmountInputProps> {
+    public render(): React.ReactNode {
+        const { fontColor, fontSize, value } = this.props;
+        return (
+            <Container borderBottom="1px solid rgba(255,255,255,0.3)" display="inline-block">
+                <Input
+                    fontColor={fontColor}
+                    fontSize={fontSize}
+                    onChange={this._handleChange}
```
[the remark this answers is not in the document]
fragosti: nope! because we're using an arrow function.. this was driving me crazy in `website`

packages/instant/src/types.ts
steveklebanoff: I love typesafety, so to mean it seems worth it, but I'm cool with rolling as-is for now if that's what ya'll would like to do: we could potentially revisit later once we have more actions

packages/instant/src/components/amount_input.tsx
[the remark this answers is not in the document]
BMillman19: oh coolio

packages/instant/src/components/ui/input.tsx
fragosti: Yea for now I'm thinking that the `theme` represent the configurable aspect of our UI... and for now that's just the colors... but open to changing that. The input prop could also just take a `fontFamily` prop -- but we can add that when we need it :D

packages/instant/package.json
```diff
@@ -43,15 +43,19 @@
     },
     "homepage": "https://github.com/0xProject/0x-monorepo/packages/instant/README.md",
     "dependencies": {
-        "@0xproject/connect": "^2.0.4",
-        "@0xproject/types": "^1.1.4",
+        "@0xproject/asset-buyer": "^2.0.0",
+        "@0xproject/types": "^1.1.1",
```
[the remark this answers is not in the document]
fragosti: because mistakes

packages/instant/src/types.ts
fragosti: Lets look into this later - I agree that having types here would be great, but would like to avoid boilerplate if possible. Seems like a bit of rabbit hole and I think this branch is blocking a few things so going to go ahead and merge it once CI passes.

PR #1115, opened 2018-10-04
Upgrade Webpack

fabioberger:
## Description

This PR upgrades Webpack to v4.x along with all webpack-dependent packages.

Webpack 4 came with some [breaking changes](https://webpack.js.org/migrate/4/) which were addressed. We were hoping to use the built-in uglify support in v4 but unfortunately it still doesn't support minifying ES6 (some of our deps, or deps of deps are in ES6 -- e.g `rlp`). Instead we are using (terser-webpack-plugin)[https://github.com/webpack-contrib/terser-webpack-plugin] which is a fork of uglify-webpack-plugin with ES6 support.

New compilation times:
- Website dev: 35sec
- Website prod: 2min17sec
- 0x.js prod: 1min14sec

Since these times are not huge improvements, this PR also changes CI, such that website is built on every CI run BUT it is built in parallel with the test runs, after the build phase. No other test has the website as a prerequisite, allowing us to structure it this way. This will shave **2min17sec off every CI build** 🍾 .

## Checklist:


packages/website/package.json
```diff
@@ -7,11 +7,11 @@
     "private": true,
     "description": "Website and 0x portal dapp",
     "scripts": {
-        "build": "NODE_ENV=production node --max_old_space_size=8192 ../../node_modules/.bin/webpack",
-        "build:ci": "yarn build",
+        "build": "node --max_old_space_size=8192 ../../node_modules/.bin/webpack --mode production",
```
fragosti: does just using `webpack` work now?

fabioberger: Not sure. Gonna merge then check.

PR #1118, opened 2018-10-06
[order-watcher] Public function getWatchCounts

amaurer:
## Description

Added public function to get order watcher counts so I can effectively scale order watcher instances based on watch loads.

## Testing instructions

Function added to test suite, no real changes as it reads in internal property that is assumed properly maintained.

## Types of changes

Added function for getWatchCounts in order watcher and one test.

## Checklist:

*   [X ] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [X ] Add tests to cover changes as needed.
*   [X] Update documentation as needed.
*   [X ] Add new entries to the relevant CHANGELOG.jsons.

### DIFF
--- a/packages/order-watcher/CHANGELOG.json
+++ b/packages/order-watcher/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "2.1.2",
+        "changes": [
+            {
+                "note": "Added getWatchCounts function"
+            }
+        ],
+        "timestamp": 1538836030
+    },
+    {
         "version": "2.1.1",
         "changes": [
             {
--- a/packages/order-watcher/src/order_watcher/order_watcher.ts
+++ b/packages/order-watcher/src/order_watcher/order_watcher.ts
@@ -213,6 +213,10 @@ export class OrderWatcher {
         this._expirationWatcher.unsubscribe();
         intervalUtils.clearAsyncExcludingInterval(this._cleanupJobIntervalIdIfExists);
     }
+    /**
+     * Gets number of orderHashes currently being watched by the order watcher instance.
+     * @returns {number}
+     */
     public getWatchCount(): number {
         return _.size(this._orderByOrderHash);
     }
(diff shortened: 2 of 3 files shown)


packages/order-watcher/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.1.2",
+        "changes": [
+            {
+                "note": "Added getStats function and returns a Stats object"
+            }
+        ],
+        "timestamp": 1538836030
```
fabioberger: Remove the timestamp. This get's auto-generated when the version is published.

packages/order-watcher/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.1.2",
+        "changes": [
+            {
+                "note": "Added getStats function and returns a Stats object"
+            }
```
fabioberger: Add `PR` number here:

```
{
         "note": "Added getStats function and returns a Stats object",
         "pr": 1118
 }
```

packages/order-watcher/src/order_watcher/order_watcher.ts
```diff
@@ -57,6 +57,10 @@ interface OrderStateByOrderHash {
     [orderHash: string]: OrderState;
 }
 
+export interface Stats {
+    orderCount: number;
+}
```
fabioberger: Please move this to `order-watcher/src/types.ts` since we export it.

Then import it in this file, and export from `index.ts`.

fabioberger:
Thanks for changing that @amaurer! Looks good, just minor fixes then we can merge and publish.

packages/order-watcher/src/order_watcher/order_watcher.ts
fabioberger: Let's make sure to export this type from `index.ts` since it's part of the public interface of this package.

fabioberger:
Also, could you run `yarn prettier` from the monorepo root, as well as `yarn lint`?

fabioberger:
@amaurer thanks for addressing my comments. Can you also export `Stats` in the `index.ts` file?

Once this is done and the conflicts are resolved, I'll approve & merge.

PR #1121, opened 2018-10-08
Clarifies use of schemas outside of Javascript/TypeScript.

nebkor:
## Description

The documentation previously claimed that the schema files were ready to be used in non-Javascript-based systems, and this is not correct.

## Checklist:


fabioberger:
We should really move these into JSON files. Will add that to the backlog.

nebkor:
> We should really move these into JSON files. Will add that to the backlog.

I got about halfway through writing a Python program to resolve all the '$ref' keys in the serialized schema I was able to get a TypeScript repl to emit yesterday, when I realized I was writing a hand-rolled recursive descent parser, and decided to put that on hold until later :)

PR #1126, opened 2018-10-10
[python/order_utils] Project stub

feuGeneA:
## Description
An empty project, with respect to functionality, but one configured with
a test framework, linters, CI integration, etc.

https://app.asana.com/0/836857747873847/839549782781239/f

## Testing instructions

Watch CircleCI!  To test locally, have python 3.7 installed, and see commands available in `package.json`.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,6 +162,90 @@ jobs:
           key: coverage-web3-wrapper-{{ .Environment.CIRCLE_SHA1 }}
           paths:
             - ~/repo/packages/web3-wrapper/coverage/lcov.info
+  test-python:
+    working_directory: ~/repo
+    docker:
+      - image: circleci/python
+    steps:
+      - checkout
+      - run: sudo chown -R circleci:circleci /usr/local/bin
+      - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
+      - restore_cache:
+          key: deps9-{{ .Branch }}-{{ .Environment.CIRCLE_SHA1 }}
+      - run:
+          command: |
+            cd python-packages/order_utils
+            python -m ensurepip
+            python -m pip install -e .[dev]
+      - save_cache:
+          key: deps9-{{ .Branch }}-{{ .Environment.CIRCLE_SHA1 }}
+          paths:
+            - "/usr/local/bin"
+            - "/usr/local/lib/python3.7/site-packages"
+            - ".eggs"
+            - ".mypy_cache"
+            - ".pytest_cache"
+            - ".tox"
+      - run:
+          command: |
+            cd python-packages/order_utils
+            coverage run setup.py test
+      - save_cache:
+          key: coverage-python-order-utils-{{ .Environment.CIRCLE_SHA1 }}
+          paths:
+            - ~/repo/python-packages/order_utils/.coverage
+  test-rest-python:
+    working_directory: ~/repo
+    docker:
+      - image: circleci/python
+    steps:
+      - checkout
+      - run: sudo chown -R circleci:circleci /usr/local/bin
+      - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
+      - restore_cache:
+          key: deps9-{{ .Branch }}-{{ .Environment.CIRCLE_SHA1 }}
+      - run:
+          command: |
+            cd python-packages/order_utils
+            python -m ensurepip
+            python -m pip install -e .[dev]
+      - save_cache:
+          key: deps9-{{ .Branch }}-{{ .Environment.CIRCLE_SHA1 }}
+          paths:
+            - "/usr/local/bin"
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -169,6 +169,94 @@ jobs:
           key: coverage-web3-wrapper-{{ .Environment.CIRCLE_SHA1 }}
           paths:
             - ~/repo/packages/web3-wrapper/coverage/lcov.info
+  test-python:
```
fabioberger: What is the difference between `test-python` and `test-rest-python`? We have a single package, why can't we have a single CI process?

python-packages/order_utils/package.json
```diff
@@ -0,0 +1,17 @@
+{
+    "comment": "this file exists as an entry point to building this project, specifically for humans that are familiar with yarn and already have it installed.  this file is not used in any automation or CI.",
```
fabioberger: lol

python-packages/order_utils/setup.py
```diff
@@ -0,0 +1,119 @@
+"""setuptools module for order_utils package."""
+
+import subprocess  # nosec
```
fabioberger: What does `nosec` mean?

python-packages/order_utils/setup.py
```diff
@@ -0,0 +1,119 @@
...
+from shutil import rmtree
+from os import path, remove, walk
+
+from distutils.command.clean import clean  # type: ignore
+from setuptools import setup  # type: ignore
+import setuptools.command.build_py  # type: ignore
+from setuptools.command.test import test as TestCommand  # type: ignore
```
fabioberger: Are you ignoring the type checks because these packages don't have any?

python-packages/order_utils/setup.py
```diff
@@ -0,0 +1,119 @@
...
+            "pycodestyle",
+            "pydocstyle",
+            "pylint",
+            "pytest",
+            "sphinx",
+            "tox",
+        ]
+    },
+    python_requires=">=3.6, <4",
+    package_data={"zero_x.order_utils": ["py.typed"]},
```
fabioberger: `zero_ex`

python-packages/order_utils/setup.py
```diff
@@ -0,0 +1,119 @@
...
+    package_dir={"": "src"},
+    license="Apache 2.0",
+    keywords="ethereum",
```
fabioberger: `ethereum 0x decentralized blockchain dex exchange`

python-packages/order_utils/setup.py
```diff
@@ -0,0 +1,119 @@
...
+    packages=["zero_x.order_utils"],
```
fabioberger: Ditto above.

fabioberger:
Looking good!

.circleci/config.yml
feuGeneA: The reason I set it up as a two-stage thing is to enable fast failing.

Both tasks run the same set of tests.

`test-python` just calls `python` to run the tests, so uses the default python executable (via `PATH` resolution), and runs quite fast.

`test-rest-python` runs the tests via `tox`, which automatically sets up "virtual environments" to run the tests in.  Virtual environments are used in Python to run your code in a controlled environment, testing with a specific interpreter and a specific set of package dependencies.  Typically `tox` is used to ensure compatibility with multiple versions of python (and the versions of package dependencies that are available for each of those python versions), but possibilities abound:  For example, [web3.py uses it](https://github.com/ethereum/web3.py/blob/master/tox.ini) to run their tests against multiple different ethereum node implementations.

Having said all that, _at this point_, using `tox` is not adding any value for us.  (It made sense in my local environment, because I have both Python 3.6 and 3.7 installed, and we can support both.  However, CircleCI's Python container only contains 3.5 and 3.7, and we won't support 3.5 due to lack of type annotations, so right now `tox` is only using a single virtual environment, that for 3.7.)  However, I think that we will get a lot of value out of it as our codebase grows.

So, the reason I divided the two runs from each other is so that that failures would be raised quickly by `test-python`, so that we don't have to wait for all of the `tox` environment permutations to run before finding out what's broken.  But if you feel strongly that they should be combined into one run, I'm good with that too.

python-packages/order_utils/setup.py
feuGeneA: it's an ignore directive for the `bandit` linter, which is focused on security issues, so "nosec" means "no security checks".

python-packages/order_utils/setup.py
feuGeneA: yes.  or, maybe it's just that i don't have them installed/available in my environment?  this will evolve when i start actually coding with types and understand how the distribution/packaging of types works.

.circleci/config.yml
fabioberger: Let's keep them separate but omit the `tox` run from CI for now. We can always add it back later. Thoughts?

PR #1127, opened 2018-10-10
[sol-meta] Solidity meta compiler

Recmo:
## Description

Compiles Solidity to Solidity allowing you to transform the code in the process. Included transforms will be:

* Generate wrappers exposing non-public functions as public.
* Generate mock implementations of interfaces.

It includes a standalone CLI. Still looking for the best way to integrate into the test process.

## Checklist:


packages/sol-meta/src/cli.ts
```diff
@@ -0,0 +1,142 @@
...
+        if (config.contracts) {
+            contracts = { ...contracts, ...config.contracts };
+        }
+    }
+
+    // Handle command line arguments
+    if (argv.remapping) {
+        options.remapping = (argv.remapping as string[])
+            .map((str: string) => str.split('=', 2))
+            .reduce((a, [from, to]) => ({ ...a, [from]: to }), {});
```
LogvinovLeon:
```suggestion
            .reduce((remappingsAccumulator, [from, to]) => ({ ...remappingsAccumulator, [from]: to }), {});
```

packages/sol-meta/src/cli.ts
```diff
@@ -0,0 +1,142 @@
...
+            };
+
+            // Find path
+            const path = Object.keys(sources).find(path => contractName in sources[path].contracts);
+            if (path === undefined) {
+                throw new Error(`Could not find contract ${contractName}.`);
+            }
+
+            // Create mock contract
+            console.time('Generating mocks');
```
LogvinovLeon: What about a no-console linter rule?

packages/sol-meta/src/constructor.ts
```diff
@@ -0,0 +1,80 @@
...
+            statements: [],
+        },
+    };
+}
+
+// Solidity by itself does not give an error when a contract is unintentionally
+// abstract. We can force Solidity to produce an error and point us to the
+// abstract function by trying to runtime instantiate the contract. This function
+// produces a small contract that tries to instantiate the given contract.
+export function nonAbstractForcer(contractName: string): S.ContractDefinition {
```
LogvinovLeon: It might be too late for that comment, but wouldn't it be more readable if we use a completely different approach?
Instead of constructing the AST we construct solidity using the template strings and then parse it to AST.
I just saw that the commment text might be as well the code itself. ASTs are hard to read by humans and we already have a parser to convert.

packages/sol-meta/src/contractMocker.ts
```diff
@@ -0,0 +1,123 @@
...
+    options: ContractMockOptions,
+): S.SourceUnit {
+    const sourceInfo = sources[path];
+
+    // Flatten target contract, this collapses the inheritane hierarchy
+    if (!(contractName in sourceInfo.contracts)) {
+        throw new Error(`Could not find contract ${contractName} in "${path}".`);
+    }
+    const [flat, parents] = flattenContract(sourceInfo.contracts[contractName], name => {
+        if (!(name in sourceInfo.scope)) {
```
LogvinovLeon: I will not comment on any built-in function vs lodash for now. Let's just talk with Fabio about it as he've mentioned above.

packages/sol-meta/src/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+export const litString = (value: string): S.StringLiteral => ({
+    type: S.NodeType.StringLiteral,
+    value,
+});
+
+export type Litteral = string | number;
+
+export const litteral = (value: Litteral): S.Expression => {
+    if (isNumber(value)) {
+        return litNumber(value);
```
LogvinovLeon: That number is lit 👍

packages/sol-meta/test/index.ts
```diff
@@ -0,0 +1,46 @@
...
+import * as glob from 'glob';
+import 'mocha';
+import * as path from 'path';
+
+import { parse } from '../src/parser';
+import { unparse } from '../src/unparser';
+
+const expect = chai.expect;
+
+const promisify = <T>(func: ((...args: any[]) => void)) => (...args: any[]) =>
```
LogvinovLeon: We have that in @0x/utils

packages/sol-meta/test/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    contracts.forEach(({ name, source }) =>
+        it(`should parse ${name}`, () => {
+            parse(source);
+        }),
+    );
+});
+
+describe.only('Unparser', () => {
```
LogvinovLeon:
```suggestion
describe('Unparser', () => {
```

packages/sol-meta/test/index.ts
```diff
@@ -0,0 +1,46 @@
...
+    contracts.forEach(({ name, source }) =>
+        it(`should unparse ${name}`, () => {
+            const ast = parse(source);
+            const src = unparse(ast);
+            const ast2 = parse(src);
+            //expect(ast2).to.deep.equal(ast);
```
LogvinovLeon: ?

packages/sol-meta/test/interface.expected.ast.js
```diff
@@ -0,0 +1,289 @@
+x={ type: 'SourceUnit',
```
LogvinovLeon:
```suggestion
{ type: 'SourceUnit',
```

packages/sol-meta/test/interface.expected.ast
```diff
@@ -0,0 +1,339 @@
...
+                          { type: 'MemberAccess',
+                            expression: { type: 'Identifier', name: 'returnValues' },
+                            memberName: '_arg_1' } ],
+                       isArray: false } } ] },
+            visibility: 'internal',
+            modifiers: [],
+            isConstructor: false,
+            stateMutability: null } ],
+       kind: 'contract' } ] }
+{ type: 'SourceUnit',
```
LogvinovLeon: This AST file has 2 ASTs and a raw contract. Is that right?

packages/sol-meta/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "compilerOptions": {
+        "outDir": "lib",
+        "rootDir": "."
+    },
+    "include": [
+        "./src/**/*",
+        "./test/**/*",
+        "../typescript-typings/types/solidity-parser-antlr/*"
+    ]
+}
```
LogvinovLeon: Newline

packages/sol-meta/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+    "extends": "../../tsconfig",
+    "compilerOptions": {
+        "outDir": "lib",
+        "rootDir": "."
+    },
+    "include": [
+        "./src/**/*",
+        "./test/**/*",
+        "../typescript-typings/types/solidity-parser-antlr/*"
```
LogvinovLeon: Why not use typeRoots as in other places?

packages/typescript-typings/types/solidity-parser-antlr/index.d.ts
```diff
@@ -0,0 +1,808 @@
+// Type definitions for solidity-parser-antlr 0.3.2
+// Project: https://github.com/federicobond/solidity-parser-antlr
+// Definitions by: Leonid Logvinov <https://github.com/LogvinovLeon>
+//                 Alex Browne <https://github.com/albrow>
+//                 Remco Bloemen <https://github.com/recmo>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+
```
LogvinovLeon: Add a comment why you copied it back to the repo. What is changed here. Also add a link to a DT PR.

LogvinovLeon:
Can we also fix the tests before merging?

packages/contracts/contracts/protocol/Exchange/MixinTransactions.sol
```diff
@@ -16,6 +16,7 @@
 
 */
 pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
```
abandeali1: I believe this will actually change the bytecode of this contract and we purposely removed `pragma experimental ABIEncoderV2;` when not necessary because it reduced gas costs. What is the reason for adding this back in?

Edit: I see why we need this to generate public wrappers for all of these internal functions. Is there a better way to do this that doesn't require us altering the underlying contract?

packages/contracts/contracts/protocol/Exchange/libs/LibFillResults.sol
```diff
@@ -17,6 +17,7 @@
 */
 
 pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
```
abandeali1: Same as above.

packages/contracts/contracts/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -17,6 +17,7 @@
 */
 
 pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
```
abandeali1: Same as above.

packages/sol-meta/src/solcWrapper.ts
```diff
@@ -0,0 +1,60 @@
...
+            .map(({ value }) => value)[0] || 'latest';
+
+    // Get Solidity compiler
+    console.time('Loading solc-js');
+    const compiler = await Solc.getSolcAsync(version);
+    console.timeEnd('Loading solc-js');
+
+    // Solidity standard JSON input
+    // TODO: Typescript typings
+    // See: https://solidity.readthedocs.io/en/develop/using-the-compiler.html#compiler-input-and-output-json-description
```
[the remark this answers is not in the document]
Recmo: Only `StandardOutput` is there, `StandardInput` is missing.

packages/contracts/contracts/protocol/Exchange/libs/LibOrder.sol
```diff
@@ -45,11 +46,11 @@ contract LibOrder is
...
-        INVALID_MAKER_ASSET_AMOUNT,  // Order does not have a valid maker asset amount
-        INVALID_TAKER_ASSET_AMOUNT,  // Order does not have a valid taker asset amount
-        FILLABLE,                    // Order is fillable
-        EXPIRED,                     // Order has already expired
+        INVALID,                     // 0 Default value
+        INVALID_MAKER_ASSET_AMOUNT,  // 1 Order does not have a valid maker asset amount
+        INVALID_TAKER_ASSET_AMOUNT,  // 2 Order does not have a valid taker asset amount
+        FILLABLE,                    // 3 Order is fillable
+        EXPIRED,                     // 4 Order has already expired
         FULLY_FILLED,                // Order is fully filled
```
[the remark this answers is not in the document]
Recmo: Merge conflict

packages/sol-cov/package.json
```diff
@@ -48,7 +48,6 @@
         "@0x/typescript-typings": "^3.0.4",
         "@0x/utils": "^2.0.4",
         "@0x/web3-wrapper": "^3.1.1",
-        "@types/solidity-parser-antlr": "^0.2.0",
```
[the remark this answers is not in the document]
Recmo: They are resolved automatic, this line was causing automatic resolving to fail

packages/sol-cov/src/ast_visitor.ts
```diff
@@ -79,9 +79,6 @@ export class ASTVisitor {
     public WhileStatement(ast: Parser.WhileStatement): void {
         this._visitStatement(ast);
     }
-    public SimpleStatement(ast: Parser.SimpleStatement): void {
```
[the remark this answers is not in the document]
Recmo: AFAIK, there's no such node. The `solidity-parser-antlr` typing where missing a some of node types, and also had some types that did not correspond to nodes. I think the package was updated.

packages/sol-meta/README.md
```diff
@@ -0,0 +1,386 @@
...
+            }
+        ]
+    },
+    "contracts": {
+        "LibEIP712": {},
+        "MixinAssetProxyDispatcher": {},
+        "MixinExchangeCore": {},
+        "MixinMatchOrders": {
+            "constructors": {
+                "LibConstants": ["\"Different string\""]
```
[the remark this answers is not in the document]
Recmo: It's further down, I'll add a bit more.

packages/sol-meta/src/constructor.ts
Recmo: That's a reasonable approach. While writing I realized that it was probably better.

I did not go this route because it requires patching `soldity-parser-antlr` to parse fragments instead of whole contracts.

Another slight advantage is that the type system prevents you from accidentally producing incorrect code. Like the template `${a} * 3` will fail when `a = 4 + 4`.

packages/sol-meta/src/utils.ts
Recmo: 🔥

packages/sol-meta/tsconfig.json
Recmo: IDK. @fabioberger What is the correct way to get typings to work?

packages/contracts/contracts/protocol/Exchange/MixinTransactions.sol
Recmo: If the final `Exchange` contract has ABIv2 enabled, doesn't that automatically enable it for all parent contracts?

There are ways around it, I just thought it was incorrectly missing here since everything will be compiled with ABIv2 anyway.

packages/contracts/contracts/protocol/Exchange/libs/LibFillResults.sol
Recmo: Same as above :)

packages/contracts/contracts/protocol/Exchange/libs/LibOrder.sol
Recmo: ditto

packages/sol-meta/src/cli.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+    options.sources = [...options.sources, ...argv._];
+
+    // Parse all sources
+    console.time('Parsing sources');
+    const sources = await readSources(options.sources, options);
+    console.timeEnd('Parsing sources');
+
+    for (const contractName in contracts) {
+        if (contracts.hasOwnProperty(contractName)) {
```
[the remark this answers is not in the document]
Recmo:
```
 | packages/sol-meta/src/cli.ts:91:32 - error TS2549: Type '{ [contractName: string]: ContractMockOptions; }' is not an array type or a string type or does not have a '[Symbol.iterator]()' method that returns an iterator.
 |
 | 91     for (const contractName of contracts) {
 |                                   ~~~~~~~~~
 |
```

packages/sol-meta/src/constructor.ts
LogvinovLeon: Good point.

packages/sol-cov/package.json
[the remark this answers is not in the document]
fabioberger: I don't know more. Can you elaborate?

LogvinovLeon:
@Recmo Do we still want to merge it?

PR #1129, opened 2018-10-11
Use medium+ resource class for build-website

albrow:
## Testing instructions

CI appears to be failing here because we ran out of memory for the `build-website` job. This PR will hopefully fix that by bumping up the resource class.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,7 @@ jobs:
           paths:
             - ~/repo
   build-website:
+    resource_class: medium+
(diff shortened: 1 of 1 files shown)


fabioberger:
I'm ok with accepting this as a short-term fix, but we definitively need a longer term solution that doesn't depend on adding more resources to CI. The reason is that external developers won't have access to these additional CI resources, so their PR's will fail our CI tests. This is very external developer unfriendly.

PR #1131, opened 2018-10-11
[instant][types][order-utils][asset-buyer] Move over and clean up features from zrx-buyer

fragosti:
## Description

Straight port of features:
- Abstraction for fetching async data, and fetching ETH-USD price from coinbase.
- Connecting InstantHeading and BuyButton to the redux store.
- Instantiate provider, asset buyer, web3wrapper etc...
- Display the buy quote and allow the buy button to execute the buy quote.
- Add notion of dynamic `assetData` and `assetMetaData`.
- Add some utils for formatting eth price and eth-usd price. 
- Populate the order details section. 
- Add debouncing to fetches that result from input changes.
- Make redux actions type-safe.

I also changed some types but then ended up not needing the change... but still seems like a good change? Basically we should control all unions of different asset data types, such that the build fails if we add a type to that union, assuming we are using a switch statement in the implementation of the function that is meant to handle all asset data types.

screen shot 2018-10-12 at 4 11 58 pm

## Checklist:


packages/instant/src/components/instant_heading.tsx
```diff
@@ -26,18 +57,18 @@ export const InstantHeading: React.StatelessComponent<InstantHeadingProps> = pro
                 <SelectedAssetAmountInput fontSize="45px" />
                 <Container display="inline-block" marginLeft="10px">
                     <Text fontSize="45px" fontColor={ColorOption.white} textTransform="uppercase">
-                        rep
+                        zrx
```
fragosti: Remove hardcoded zrx values

packages/instant/src/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const sraApiUrl = 'https://api.radarrelay.com/0x/v2/';
+export const zrxContractAddress = '0xe41d2489571d322189246dafa5ebde1f4699f498';
+export const zrxDecimals = 18;
```
fragosti: remove hard-coded zrx values

packages/instant/src/containers/selected_asset_amount_input.ts
```diff
@@ -0,0 +1,57 @@
...
+});
+
+const mapDispatchToProps = (dispatch: Dispatch<Action>): ConnectedDispatch => ({
+    onChange: async value => {
+        // Update the input
+        dispatch({ type: ActionTypes.UPDATE_SELECTED_ASSET_AMOUNT, data: value });
+        // invalidate the last buy quote.
+        dispatch({ type: ActionTypes.UPDATE_LATEST_BUY_QUOTE, data: undefined });
+        // reset our buy state
+        dispatch({ type: ActionTypes.UPDATE_SELECTED_ASSET_BUY_STATE, data: AsyncProcessState.NONE });
```
fragosti: create action creators? would allow us to have typed actions...

packages/instant/src/redux/async_data.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ActionTypes } from '../types';
+import { coinbaseApi } from '../util/coinbase_api';
+
+import { store } from './store';
+
+export const asyncData = {
+    fetchAndDispatchToStore: async () => {
+        let ethUsdPriceStr = '0';
+        try {
+            ethUsdPriceStr = await coinbaseApi.getEthUsdPrice();
```
fragosti: coinbase API call should return `BigNumber`

packages/instant/src/redux/reducer.ts
```diff
@@ -1,16 +1,21 @@
+import { BuyQuote } from '@0xproject/asset-buyer';
 import { BigNumber } from '@0xproject/utils';
 import * as _ from 'lodash';
 
-import { Action, ActionTypes } from '../types';
+import { Action, ActionTypes, AsyncProcessState } from '../types';
 
 export interface State {
```
fragosti: add state for `Provider`, and other things required to remove hard-coded values.

fragosti: Not going to do this for now. Should be done when we implement different wallets.

packages/instant/src/components/order_details.tsx
```diff
@@ -1,53 +1,85 @@
...
+        );
+    }
+}
 
 export interface OrderDetailsRowProps {
     name: string;
-    primaryValue: string;
-    secondaryValue: string;
+    ethAmount?: BigNumber;
+    shouldConvertEthToUnitAmount?: boolean;
```
BMillman19: what is this mean? comments for this props definition would be good

packages/instant/src/components/order_details.tsx
```diff
@@ -1,53 +1,85 @@
...
         <Container padding="10px 0px" borderTop="1px dashed" borderColor={ColorOption.feintGrey}>
             <Flex justify="space-between">
                 <Text fontWeight={fontWeight} fontColor={ColorOption.grey}>
-                    {props.name}
+                    {name}
                 </Text>
                 <Container>
                     <Container marginRight="3px" display="inline-block">
-                        <Text fontColor={ColorOption.lightGrey}>({props.secondaryValue}) </Text>
+                        <Text fontColor={ColorOption.lightGrey}>({usdFormatter(ethAmount, ethUsdPrice)})</Text>
```
BMillman19: If we are missing the ETH/USD conversion we probably shouldn't render the parentheses

packages/instant/src/components/order_details.tsx
```diff
@@ -1,53 +1,85 @@
...
+export const OrderDetailsRow: React.StatelessComponent<OrderDetailsRowProps> = ({
+    name,
+    ethAmount,
+    shouldConvertEthToUnitAmount,
+    ethUsdPrice,
+    shouldEmphasize,
+}) => {
+    const fontWeight = shouldEmphasize ? 700 : 400;
+    const usdFormatter = shouldConvertEthToUnitAmount ? format.ethBaseAmountInUsd : format.ethUnitAmountInUsd;
+    const ethFormatter = shouldConvertEthToUnitAmount ? format.ethBaseAmount : format.ethUnitAmount;
```
BMillman19: do we need to bind format here?

packages/instant/src/components/buy_button.tsx
```diff
@@ -1,19 +1,56 @@
...
+    private readonly _handleClick = async () => {
+        // The button is disabled when there is no buy quote anyway.
+        if (_.isUndefined(this.props.buyQuote)) {
+            return;
+        }
+        this.props.onClick(this.props.buyQuote);
+        try {
+            const txnHash = await assetBuyer.executeBuyQuoteAsync(this.props.buyQuote, {
+                // HACK: There is a calculation issue in asset-buyer. ETH is refunded anyway so just over-estimate.
+                ethAmount: this.props.buyQuote.worstCaseQuoteInfo.totalEthAmount.mul(2),
```
BMillman19: i think this ethAmount should come from the buyQuoteInfo stored in redux so there is less chance we end up executing a buy thats different from what we are presenting in the order details

packages/instant/src/containers/selected_asset_buy_button.ts
```diff
@@ -0,0 +1,55 @@
...
+import { State } from '../redux/reducer';
+import { AsyncProcessState } from '../types';
+
+import { BuyButton } from '../components/buy_button';
+
+export interface SelectedAssetBuyButtonProps {}
+
+interface ConnectedState {
+    text: string;
+    buyQuote?: BuyQuote;
```
BMillman19: same comment about passing the same quoteInfo here then we are using in OrderDetails

packages/instant/src/util/asset_buyer.ts
```diff
@@ -0,0 +1,11 @@
+import { AssetBuyer } from '@0xproject/asset-buyer';
+
+import { sraApiUrl } from '../constants';
+
+import { getProvider } from './provider';
+
+const provider = getProvider();
+
+export const assetBuyer = AssetBuyer.getAssetBuyerForStandardRelayerAPIUrl(provider, sraApiUrl, {
+    expiryBufferSeconds: 300,
```
BMillman19: this should already be the default

packages/instant/src/util/coinbase_api.ts
```diff
@@ -0,0 +1,10 @@
+import { BigNumber } from '@0xproject/utils';
+
+const baseEndpoint = 'https://api.coinbase.com/v2';
+export const coinbaseApi = {
```
BMillman19: is coinbase the best choice for this?

BMillman19: I'm not sure, jw what thought process you had when choosing the coinbase API to supply the price. Some things to consider, support for other currencies for localization, rate limiting, downtime, etc.

packages/instant/src/components/amount_input.tsx
```diff
@@ -10,10 +10,13 @@ export interface AmountInputProps {
     fontColor?: ColorOption;
     fontSize?: string;
     value?: BigNumber;
-    onChange?: (value?: BigNumber) => void;
+    onChange: (value?: BigNumber) => void;
```
steveklebanoff: just curious -- why move this from being optional to being required but having a default no-op? is this a pattern we want to follow for callbacks?

packages/instant/src/redux/actions.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export enum ActionTypes {
+    UPDATE_ETH_USD_PRICE = 'UPDATE_ETH_USD_PRICE',
+    UPDATE_SELECTED_ASSET_AMOUNT = 'UPDATE_SELECTED_ASSET_AMOUNT',
+    UPDATE_SELECTED_ASSET_BUY_STATE = 'UPDATE_SELECTED_ASSET_BUY_STATE',
+    UPDATE_LATEST_BUY_QUOTE = 'UPDATE_LATEST_BUY_QUOTE',
+}
+
+export const actions = {
```
steveklebanoff: typed actions! 🎉 🙌 💯

packages/instant/src/containers/latest_buy_quote_order_details.ts
```diff
@@ -0,0 +1,26 @@
...
+export interface LatestBuyQuoteOrderDetailsProps {}
+
+interface ConnectedState {
+    buyQuoteInfo?: BuyQuoteInfo;
+    ethUsdPrice?: BigNumber;
+}
+
+const mapStateToProps = (state: State, _ownProps: LatestBuyQuoteOrderDetailsProps): ConnectedState => ({
+    // use the worst case quote info
+    buyQuoteInfo: _.get(state, 'latestBuyQuote.worstCaseQuoteInfo'),
```
steveklebanoff: minor: I would prefer to grab the worstCaseInfo without using `_.get` to keep the typesafety (i.e. ensuring `worstCaseQuoteInfo` is actually a property of `latestBuyQuote`

packages/instant/src/constants.ts
```diff
@@ -0,0 +1,6 @@
+import { BigNumber } from '@0xproject/utils';
+export const BIG_NUMBER_ZERO = new BigNumber(0);
+export const sraApiUrl = 'https://api.radarrelay.com/0x/v2/';
```
steveklebanoff: will we make this more configurable in a separate PR?

packages/instant/src/components/amount_input.tsx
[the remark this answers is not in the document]
fragosti: yep! This is what made most sense to me.

fragosti: So yea this is a bit confusing. @types/react and ts > 3.0.0 added understanding of `defaultProps`. So the way you think about the `Props` interface is that it's the internal interface (in this case, callback is not optional). However, if you have a default value the compiler knows that you actually don't have to pass the prop in (ie. the external interface is a function of internal interface + defaultProps). 

So in practice it's not required, has default no-ops, so internally you can assume it's not undefined.

And yea I like this pattern :P

packages/instant/src/components/asset_amount_input.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { assetMetaData } from '../data/asset_meta_data';
+import { ColorOption } from '../style/theme';
+
+import { AmountInput, AmountInputProps } from './amount_input';
+import { Container, Text } from './ui';
+
+export interface AssetAmountInputProps extends AmountInputProps {
+    assetData?: string;
+    onChange: (value?: BigNumber, assetData?: string) => void;
```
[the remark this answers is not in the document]
fragosti: In practice no, but those two things can be `undefined` independently and that case needs to be handled somewhere. The other option is to not render something, but in the event that we are rendering, we need to handle it here.

packages/instant/src/components/asset_amount_input.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export class AssetAmountInput extends React.Component<AssetAmountInputProps> {
+    public static defaultProps = {
+        onChange: _.noop.bind(_),
+    };
+    public render(): React.ReactNode {
+        const { assetData, onChange, ...rest } = this.props;
+        return (
+            <Container>
+                <AmountInput {...rest} onChange={this._handleChange} />
```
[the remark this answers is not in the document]
fragosti: Will do in a future PR (added to my list)

packages/instant/src/components/instant_heading.tsx
```diff
@@ -1,11 +1,32 @@
...
 
-export interface InstantHeadingProps {}
+export interface InstantHeadingProps {
+    selectedAssetAmount?: BigNumber;
+    totalEthBaseAmount?: BigNumber;
+    ethUsdPrice?: BigNumber;
+}
+
+const displaytotalEthBaseAmount = ({ selectedAssetAmount, totalEthBaseAmount }: InstantHeadingProps): string => {
+    if (_.isUndefined(selectedAssetAmount)) {
```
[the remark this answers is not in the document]
fragosti: I see format as just formatting ethereum amounts into strings (including ETH dollar amounts). This case is handling the input not having any value, which seems to specific.

packages/instant/src/containers/selected_asset_buy_button.ts
fragosti: I'm not sure what you mean. In terms of the API they could be different, but technically in this implementation they are not.

packages/instant/src/util/coinbase_api.ts
fragosti: I looked into a few options but did not do a crazy thorough comparison. Coinbase has 3 req / sec, 10k req / hour per IP, which seems reasonable. They also support other (fiat) currencies. Not much info available on downtime. Do you know of other APIs?

packages/instant/src/components/buy_button.tsx
```diff
@@ -1,19 +1,56 @@
...
-export const BuyButton: React.StatelessComponent<BuyButtonProps> = props => (
-    <Container padding="20px" width="100%">
-        <Button width="100%">
-            <Text fontColor={ColorOption.white} fontWeight={600} fontSize="20px">
-                Buy
-            </Text>
-        </Button>
-    </Container>
-);
+const boundNoop = _.noop.bind(_);
```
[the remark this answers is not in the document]
fragosti: The linter complains yes.

packages/instant/src/components/buy_button.tsx
```diff
@@ -1,19 +1,56 @@
...
+        }
+        this.props.onClick(this.props.buyQuote);
+        try {
+            const txnHash = await assetBuyer.executeBuyQuoteAsync(this.props.buyQuote, {
+                // HACK: There is a calculation issue in asset-buyer. ETH is refunded anyway so just over-estimate.
+                ethAmount: this.props.buyQuote.worstCaseQuoteInfo.totalEthAmount.mul(2),
+            });
+            await web3Wrapper.awaitTransactionSuccessAsync(txnHash);
+        } catch {
+            this.props.onBuyFailure(this.props.buyQuote);
```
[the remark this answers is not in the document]
fragosti: It throws.

Sweet!

packages/instant/src/components/order_details.tsx
fragosti: I just renamed everything hopefully this is clearer.

packages/instant/src/components/order_details.tsx
fragosti: We don't use `this` in format, and even if we did we would be fine because we are using arrow functions. In general I think it's a bit awkward to use the `function` keyword in es6 since we have classes and arrow functions. In es5, all we had was `function` and that was for both classes and functions.

packages/instant/src/components/buy_button.tsx
fragosti: It's not technically possible in this implementation, but yes the API allows that... I feel like that's ok?

packages/instant/src/containers/latest_buy_quote_order_details.ts
fragosti: I added a package https://github.com/rimeto/ts-optchain. The API is a bit ugly but it gives us the best of both worlds. Open to discussing, but didn't take long to include.

packages/instant/src/util/format.ts
```diff
@@ -0,0 +1,45 @@
...
+    ethUnitAmountInUsd: (
+        ethUnitAmount?: BigNumber,
+        ethUsdPrice?: BigNumber,
+        decimalPlaces: number = 2,
+        defaultText: string = '$0.00',
+    ): string => {
+        if (_.isUndefined(ethUnitAmount) || _.isUndefined(ethUsdPrice)) {
+            return defaultText;
+        }
+        return `$${ethUnitAmount.mul(ethUsdPrice).round(decimalPlaces)}`;
```
BMillman19: sometimes this results in strings with 1 decimal place like: "$3.7"

PR #1134, opened 2018-10-12
[contract-wrappers] Add optional validation to the forwarder wrapper

BMillman19:
## Description

Similar to the approach taken in exchange wrapper, make a call to an rpc
node in order to simulate the transaction before actually sending the
transaction. The decorator will parse revert reasons and other types of
errors into canonical errors that a consumer of the library expects when
interacting with a contract wrapper.

## Testing instructions

- Navigate to `packages/contract-wrappers`
- Run `yarn test`

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
+++ b/packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
@@ -8,10 +8,11 @@ import * as _ from 'lodash';
 import { artifacts } from '../artifacts';
 import { orderTxOptsSchema } from '../schemas/order_tx_opts_schema';
 import { txOptsSchema } from '../schemas/tx_opts_schema';
-import { TransactionOpts } from '../types';
+import { OrderTransactionOpts } from '../types';
 import { assert } from '../utils/assert';
 import { calldataOptimizationUtils } from '../utils/calldata_optimization_utils';
 import { constants } from '../utils/constants';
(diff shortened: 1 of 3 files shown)


packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
```diff
@@ -40,27 +41,28 @@ export class ForwarderWrapper extends ContractWrapper {
...
+    @decorators.asyncZeroExErrorHandler
     public async marketSellOrdersWithEthAsync(
         signedOrders: SignedOrder[],
         takerAddress: string,
         ethAmount: BigNumber,
         signedFeeOrders: SignedOrder[] = [],
         feePercentage: number = 0,
         feeRecipientAddress: string = constants.NULL_ADDRESS,
-        txOpts: TransactionOpts = {},
+        orderTransactionOpts: OrderTransactionOpts = {},
```
BMillman19: We default this to { shouldValidate: true } in exchange_wrapper and probably should do the same here

packages/contract-wrappers/src/contract_wrappers/forwarder_wrapper.ts
```diff
@@ -129,7 +153,7 @@ export class ForwarderWrapper extends ContractWrapper {
...
```
BMillman19: same comment about defaults here

PR #1135, opened 2018-10-12
[instant] Generic sliding error component

steveklebanoff:
## Description

Generic element for sliding an error message that slides up, shows for a few seconds, and then slides down in instant

Example usage:

```diff
--- a/packages/instant/src/components/zero_ex_instant_container.tsx
+++ b/packages/instant/src/components/zero_ex_instant_container.tsx
@@ -12,6 +12,9 @@ export interface ZeroExInstantContainerProps {}

 export const ZeroExInstantContainer: React.StatelessComponent<ZeroExInstantContainerProps> = props => (
     <Container width="350px">
+        <Container zIndex={1} position="relative">
+            <SlidingError icon="😢" message="You dont have enough ETH..." />
+        </Container>
         <Container
             zIndex={2}
             position="relative"
(
```

## Testing instructions

Checkout branch, apply diff above, see error message

## Types of changes

## Checklist:

*   [x] Add tests to cover changes as needed. - per convo, not needed


packages/instant/src/components/sliding_error.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
+
+import { ColorOption } from '../style/theme';
+
+import { SlideUpAndDownAnimationComponent } from './animations/slide_up_and_down_animation';
+
+import { Container, Text } from './ui';
+
+export interface ErrorProps {
+    icon: string;
```
steveklebanoff: For right now the only errors we have in the UI are emoji based, so using string here instead of needing to support JSX.Elements

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,92 @@
...
+        top: ${fromY}
+    }
+
+    to {
+        position: relative;
+        top: ${toY}
+    }
+`;
+
+export const SlideAnimation = styled.div<{ keyframes: string; animationType: string; animationDirection?: string }>`
```
steveklebanoff: Was on the fence about how customizable to make this, but for now just customized the things I needed to change, could add more customizations later

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,92 @@
...
+    animation-name: ${props => props.keyframes};
+    animation-duration: 0.3s;
+    animation-timing-function: ${props => props.animationType};
+    animation-delay: 0s;
+    animation-iteration-count: 1;
+    animation-fill-mode: ${props => props.animationDirection || 'none'};
+    position: relative;
+    z-index: -1;
```
steveklebanoff: Felt hacky putting this here, but I needed it for the slide to come up from behind the instant widget.  I could make this a prop but felt a little messy having to pass it around.

packages/instant/src/style/theme.ts
```diff
@@ -12,6 +12,8 @@ export enum ColorOption {
     feintGrey = 'feintGrey',
     darkGrey = 'darkGrey',
     white = 'white',
+    lightOrange = 'lightOrange',
```
steveklebanoff: Per convo, at some point we should separate out the colors which will be customizable by the widget versus the colors which are not.  I think at this point the only customizable color is `primaryColor`

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -10,11 +10,13 @@ import { Container, Flex } from './ui';
 export interface ZeroExInstantContainerProps {}
 
 export const ZeroExInstantContainer: React.StatelessComponent<ZeroExInstantContainerProps> = props => (
-    <Container hasBoxShadow={true} width="350px" backgroundColor={ColorOption.white} borderRadius="3px">
-        <Flex direction="column" justify="flex-start">
-            <InstantHeading />
-            <OrderDetails />
-            <BuyButton />
-        </Flex>
+    <Container width="350px">
```
steveklebanoff: Added this top level container so the error can go as its first child

packages/website/ts/pages/about/about.tsx
```diff
@@ -240,6 +240,16 @@ const teamRow8: ProfileInfo[] = [
     },
 ];
 
+const teamRow9: ProfileInfo[] = [
```
fragosti: This should not be in the diff right?

packages/instant/src/components/ui/text.tsx
```diff
@@ -47,6 +49,8 @@ export const Text = styled(PlainText)`
     ${props => (props.display ? `display: ${props.display}` : '')};
     ${props => (props.letterSpacing ? `letter-spacing: ${props.letterSpacing}` : '')};
     ${props => (props.textTransform ? `text-transform: ${props.textTransform}` : '')};
+    ${props => (props.marginRight ? `margin-right: ${props.marginRight}` : '')};
+    ${props => (props.marginLeft ? `margin-right: ${props.marginLeft}` : '')};
```
fragosti: You seem to have added these but not used them? Also, we try to keep positional and layout props in the `Container` component (yes, it does make it a bit of a catch-all), and keep text-related props to `Text`. So the way I would do this here, if you did need it, would be to wrap the `Text` in a `Container` with margin.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export interface SlideAnimationProps {
+    keyframes: string;
+    animationType: string;
+    animationDirection?: string;
+}
+export const SlideAnimation =
+    styled.div <
+    SlideAnimationProps >
+    `
```
fragosti: What is this syntax?? 🆒

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+> {
+    private _timeoutNumber: number | undefined;
+
+    constructor(props: SlideUpAndDownAnimationProps) {
+        super(props);
+        this._timeoutNumber = undefined;
+        this.state = {
+            slideState: 'up',
+        };
+    }
```
fragosti: this can just be `state` without the constructor. The constructor is only really necessary if you're accessing props. 

Also to type `props` as `SlideUpAndDownAnimation` is not technically correct but should work and I'm not sure what to think about it. Internally the react types take your `SlideUpAndDownAnimationProps` and adds the children prop such that `props` is actually `SlideUpAndDownAnimationProps & { children?: React.ReactNode }`

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fragosti: I think we should create a new stacking context for the container anyway since we do need to guarantee that it is above the page of the integrator.
ie. I think this z-index should be positive, and the container should simply have a higher index. 
We should probably add `position: relative` to the top level container and give it a z-index of 2, and the error one of 1. Also we should be keeping track of z-indexes in a central place.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+    z-index: -1;
+`;
+
+export const SlideUpAnimationComponent: React.StatelessComponent<{ downY: string }> = props => (
+    <SlideAnimation animationType="ease-in" keyframes={slideKeyframeGenerator(props.downY, '0px')}>
+        {props.children}
+    </SlideAnimation>
+);
+
+export const SlideDownAnimationComponent: React.StatelessComponent<{ downY: string }> = props => (
```
fragosti: Since there are two of `{ downY: string }` we should just create an interface? I've seen us being pretty consistent in not using this syntax and always creating an interface (another soft convention). Don't feel too strongly about it but just an FYI.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+interface SlideUpAndDownState {
+    slideState: 'up' | 'down';
+}
+
+export class SlideUpAndDownAnimationComponent extends React.Component<
+    SlideUpAndDownAnimationProps,
+    SlideUpAndDownState
+> {
+    private _timeoutNumber: number | undefined;
```
fragosti: does `_timeoutNumber?: number` work?

packages/instant/src/style/theme.ts
fragosti: Yea I ran into some color issues in my branch too. Seems there are different requirements now. I would rather we tackle this when my other branch (https://github.com/0xProject/0x-monorepo/pull/1131/files) is merged because it adds the concept of getting color from `assetMetaData` and that needs to fit into whatever scheme we come up with for dealing with changing colors.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from 'react';
+
+import { keyframes, styled } from '../../style/theme';
+
+const slideKeyframeGenerator = (fromY: string, toY: string) => keyframes`
+    from {
+        position: relative;
+        top: ${fromY}
```
fabioberger: No semi-colon required? Think we should at least agree on convention here. Think we've always been putting them, same as with trailing commas in objects, arrays, etc...

fabioberger: This is actually my biggest issue with interpolated string literals, they don't work with `prettier`, aren't type-safe and don't throw intelligible errors when you mistype something (e.g `backgroundColor` instead of `background-color`). Thoughts?

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fabioberger: Also why not just: `export const SlideAnimation = styled.div<SlideAnimationProps>`? Are you using `prettier`?

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+interface SlideUpAndDownState {
+    slideState: 'up' | 'down';
+}
+
+export class SlideUpAndDownAnimationComponent extends React.Component<
+    SlideUpAndDownAnimationProps,
+    SlideUpAndDownState
+> {
+    private _timeoutNumber: number | undefined;
+
```
fabioberger: Our (soft) convention is to use newlines sparingly. Usually to delineate chunks of logic. Not really between method declarations.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+    constructor(props: SlideUpAndDownAnimationProps) {
+        super(props);
+        this._timeoutNumber = undefined;
+        this.state = {
+            slideState: 'up',
+        };
+    }
+
+    public render(): JSX.Element {
```
fabioberger: Curious why you decided to return `JSX.Element`. We usually return `React.ReactNode`, but don't know what the functional difference is. Is `JSX.Element` a stricter sub-set of `React.ReactNode`? If so, I like it 😃

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+        return this._renderSlide();
+    }
+
+    public componentDidMount(): void {
+        this._timeoutNumber = window.setTimeout(() => {
```
fabioberger: I'd rename `_timeoutNumber` to `_timeoutId`(seems somewhat standard: http://mdn.beonex.com/en/DOM/window.setTimeout.html & https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout). We also follow this naming convention elsewhere.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,100 @@
...
+        return;
+    }
+
+    public componentWillUnmount(): void {
+        if (this._timeoutNumber) {
+            window.clearTimeout(this._timeoutNumber);
+        }
+    }
+
+    private readonly _renderSlide = (): JSX.Element => {
```
fabioberger: Why the syntax change here? Why not `private readonly _renderSlide(): JSX.Element {` ?

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: VSCode prettier integration seems to be autoformatting it like this:

prettier

packages/website/ts/pages/about/about.tsx
steveklebanoff: Correct -- not sure why it's showing. It doesn't show up in a local `git diff` -- should be OK to merge with this showing in the PR diff

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: I'm cool w/ creating a shared exported interface.  If someone was importing and using just _this_ shared component, would be useful for them to be able to use the exported interface.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: Don't feel strongly about this, my logic was: when I inspected a rendered StyledComponent using TS, it renders a `JSX.Element`, so that's what I used to determine the type definition:

image

I'm not sure about the functional difference, and a quick pass at googling didn't reveal too much.  Not sure if it's worth more time right now -- in the mean time, I'm down to follow the standard of `React.ReactNode`

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: Good call!

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fabioberger: hm, strange 🤷‍♂️

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fabioberger: If you use VSCode, this helps a bit: https://marketplace.visualstudio.com/items?itemName=jpoissonnier.vscode-styled-components

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: yes, and I like it better! 💯

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fabioberger: Thanks for the explanation, let's standardize for now.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: Good call on not needing the constructor to set `state`, I didn't know this.  Although as per slack convo, we had to change `slideState` to be an `Enum` to get typesafety working.

I think you're right that it's good to be explicit about `children` being a prop, will add

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
steveklebanoff: I was following the pattern of AmountInput's _handleChange here: https://github.com/0xProject/0x-monorepo/blob/development/packages/instant/src/components/amount_input.tsx#L32

But, I think for the function I was referencing, it was only defined this way to get the correct `this` context from the `Input`'s `onChange` event.  I shouldn't need the same thing here -- the syntax style you suggested reads more naturally -- I will change it.  thanks!

packages/instant/src/components/ui/text.tsx
steveklebanoff: Yes, I was originally using just `Text` for the error icon, and not `Container`, so I needed this change.  I ended up using a `Container` but forgot to remove this.  Glad this was brought up though -- now I better understand the pattern of making `Text` props only related to `text` specific properties. 👍

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fragosti: OH I didn't know you could do this. Cool.

steveklebanoff:
@fragosti Going to introduce the error boundary in a separate PR where I deal with the `AssetBuyer` errors

steveklebanoff:
@fragosti @fabioberger thanks for the thoughtful review.  Made a bunch of updates based on feedback, ready for more review when you have time

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
fragosti: Oh my point was that you're unnecessarily restricting the `props` object to something without children. But was not really an actionable point :) I think the action there would be to leave `props` untyped (if possible?) and let ts infer the type.

packages/instant/src/components/animations/slide_up_and_down_animation.tsx
```diff
@@ -0,0 +1,97 @@
...
+    animation-timing-function: ${props => props.animationType};
+    animation-delay: 0s;
+    animation-iteration-count: 1;
+    animation-fill-mode: ${props => props.animationDirection || 'none'};
+    position: relative;
+`;
+
+export interface SlideAnimationComponentProps {
+    downY: string;
+    children?: React.ReactNode;
```
fragosti: Yea sorry we don't need this :D I should have done a better job at explaining in the other comment. What you had before worked fine :)

PR #1137, opened 2018-10-15
[order-utils] fix: add missing devDependency

feuGeneA:
## Description

Added missing devDependency on abi-gen to order-utils.

## Testing instructions
`yarn clean ; PKG=@0xproject/order-utils yarn build`

## Checklist:

### DIFF
--- a/packages/order-utils/package.json
+++ b/packages/order-utils/package.json
@@ -39,6 +39,7 @@
     },
     "homepage": "https://github.com/0xProject/0x-monorepo/packages/order-utils/README.md",
     "devDependencies": {
+        "@0xproject/abi-gen": "^1.0.13",
         "@0xproject/dev-utils": "^1.0.12",
         "@0xproject/tslint-config": "^1.0.8",
         "@types/bn.js": "^4.11.0",


albrow:
@feuGeneA this is going to be made obsolete by https://github.com/0xProject/0x-monorepo/pull/1105. Maybe we should just wait until that is merged? Otherwise this would create merge conflicts for that PR.

feuGeneA:
@albrow I see, makes sense, it's just something I noticed in my on working copy.  Doesn't seem to have broken CI or anything though (cached yarn dependencies restoring symlinks?) so I'm happy to let it simmer.

albrow:
@feuGeneA yeah you're on the right track. It wasn't causing any build failures because of "dependency hoisting". Because we're using Yarn workspaces, Yarn moves shared `devDependencies` to the top-level __node_modules__ directory. Since Node recursively looks through parent directories when resolving dependencies, it finds `abi-gen` in the top-level __node_modules__  there and there aren't any issues.

You're also correct that even thought it works right now it is something we should address. I just think at the moment it's better to address it via #1105.

PR #1141, opened 2018-10-16
[migrations, dev-utils] Deployment config

kardasis:
## Description
Provided the option of using environment variables to configure certain aspects of deployment.

## Testing instructions
Run standard deployments with environment variables set.

## Types of changes

## Checklist:

*   [x ] Prefix PR title with `[WIP]` if necessary.
*   [x ] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [ x] Add tests to cover changes as needed.
*   [x ] Update documentation as needed.
*   [x ] Add new entries to the relevant CHANGELOG.jsons.


fabioberger:
Thanks for the contribution @kardasis! We love the idea of making these things more easily configurable, but have some asks in terms of "how" we do it.

1. Move all configs we want configurable into a `config.ts` module like so:

```
export const config = {
    RPC_URL: process.env.RPC_URL || DEFAULT_RPC_URL,
    ...
}
```

This was it's easy to find all configs in one place rather then have them scattered throughout the codebase.

2. Add some validation of env variables. These are supplied by the devUser, so they could be incorrect. Asserting them before assignment would provide them with better error messages.

Any other thoughts @LogvinovLeon @albrow @feuGeneA?

feuGeneA:
A thought: how about adding a `0X_` prefix to the environment variable names?  So, `0X_RPC_PORT`, `0X_GAS_LIMIT`, etc. I could see it being helpful in examining your configuration (`set | grep ^0X`), and it's not hard to imagine a conflict with another application over something as generically named as `RPC_PORT`.

Also, I think maybe there's some miscommunication.  @kardasis said, "...in every package that *uses* any of those constants??" to which @fabioberger replied, "Yes," but I think the idea is that you should add a config.ts file in the package that *defines* those variables, not every package that *uses* them.

So, specifically, add a `dev-utils/src/config.ts`, and make users of the old `constants.ts` now refer to `config.ts`.

Also, add a `migrations/src/config.ts`, move your references to `process.env` from the `migration.ts`s to there, and have the `migrations.ts` locations refer to the variables imported from the config module.  (Fabio, am I getting this right?)

fabioberger:
Thanks for clarifying @feuGeneA! I think a `0X_` prefix is a great idea. Let's do that too.

albrow:
As a general rule, I don't think we should use environment variables for configuring libraries.

When I say "libraries", I mean code that can be imported by other projects and cannot be executed directly. This is opposed to "executables" which *are* executed directly and cannot be imported.

The problem with environment variables in *libraries* is that they make configuration more implicit as opposed to explicit. It becomes much harder to keep track of which libraries require which environment variables, especially when you consider environment variables in transitive dependencies. I'm speaking from experience here and I have worked on large projects where keeping track of environment variables was a huge pain point.

If we want to make libraries more configurable, IMO the best way to do it is to use configuration objects (as we currently do in many places in the monorepo). There are a few benefits to defining configuration options this way:

1. They can be enforced by the type system at compile time. Environment variables can only be checked at runtime.
2. They are better supported by tooling (e.g. you can use the "Go to Definition" and "Show uses" features of your editor).
3. They are more flexible. You can create different instances of a class with different configuration options in a single process. You can run tests against multiple different configuration options in a single process (we actually do this kind of thing today in [one of the tests for contract-wrappers](https://github.com/0xProject/0x-monorepo/blob/development/packages/contract-wrappers/test/revert_validation_test.ts#L54)). Environment variables are global at the process-level and so don't offer this flexibility.

We can still use environment variables specifically for running migrations *as an executable*. In this case, the approach I think we should use is to separate the library code in the migrations package from the executable code. We can create an executable script which takes environment variables and simply passes them through to the library code. This means that other packages in the monorepo which use migrations as a library don't have to worry about environment variables, and people who want to run migrations as a one-off script have a way to configure them. This can probably apply to anything else in the monorepo which can be considered an executable.

The reason it is okay to define environment variables in executables but not in libraries is that executables will never be depended on by other code (at least not without spawning a new process) and will not be tested in the same way as libraries. They also are by definition run in the context of a single process, so a proccess-global scope is more appropriate. And of course configuration objects are not really possible in the context of running an executable (at least not at compile time), so we need another way to communicate configuration options.

PR #1142, opened 2018-10-17
[instant] Asset buyer errors

steveklebanoff:
## Description

Shows error messages for AssetBuyer `getBuyQuoteAsync` errors

Also:
- adds quoteState to keep track of the status of the current quote
- shows "--" if we get an error when fetching in top right
- renamed selectedAssetBuyState to buyOrderState

- [x] handling of additional AssetBuyer errors
- [x] tests for `bestNameForAsset`
- [x] polish

## Types of changes

* New feature

## Checklist:


packages/instant/src/util/error_flasher.ts
```diff
@@ -0,0 +1,27 @@
+import { Dispatch } from 'redux';
```
steveklebanoff: Wasn't sure the right place for this file, open to suggestions

packages/instant/src/containers/latest_error.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+import { connect } from 'react-redux';
+
+import { SlidingError } from '../components/sliding_error';
+import { State } from '../redux/reducer';
+import { errorDescription } from '../util/error_description';
+
+export interface LatestErrorComponentProps {
+    assetData?: string;
+    latestError?: any;
```
steveklebanoff: Per convo w/ @fragosti , we could had two options:

1) When error is "cleared" (after timeout), set `currentError` to be undefined. This would have been nice because it's the most straightforward representation of the current state, but animating the unmounting of the component would either be pretty hairy or involve introducing a new dependency.

2) Keep the error around in `latestError` but set `latestErrorDismissed` to `true`, and then render the sliding down error message.  I don't love keeping the last error around but treating it as "dismissed", but it makes the animation logic simple.

After discussing, I went with #2 for the drawbacks listed in #1.  Open to feedback though.

packages/instant/src/redux/store.ts
```diff
@@ -3,4 +3,5 @@ import { createStore, Store as ReduxStore } from 'redux';
 
 import { reducer, State } from './reducer';
 
-export const store: ReduxStore<State> = createStore(reducer);
+const reduxDevTools = (window as any).__REDUX_DEVTOOLS_EXTENSION__;
```
steveklebanoff: For using redux dev tools in chrome.  I couldn't find a pattern for this in the website code, so went with this `(window as any)` hack

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -12,6 +13,9 @@ export interface ZeroExInstantContainerProps {}
 
 export const ZeroExInstantContainer: React.StatelessComponent<ZeroExInstantContainerProps> = props => (
     <Container width="350px">
+        <Container zIndex={1} position="relative">
+            <LatestError />
+        </Container>
```
fragosti: 🔥

packages/instant/src/containers/latest_error.tsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export const LatestErrorComponent: React.StatelessComponent<LatestErrorComponentProps> = props => {
+    if (!props.latestError) {
+        return <div />;
+    }
+    const slidingDirection = props.latestErrorDismissed ? 'down' : 'up';
+    const { icon, message } = errorDescription(props.latestError, props.assetData);
+    return <SlidingError direction={slidingDirection} icon={icon} message={message} />;
+};
```
fragosti: A convention I'm trying to follow is to not have any Component code in these container files, they simply connect a pure presentational component to the redux store and do any state -> prop mapping. Seems like you don't need this component to me?

- The sliding direction logjc can just be done in mapStateToProps.
- icon is a pure function of state.
- message is a pure function of state.

The only thing is not rendering when `latestError` is not defined, but do we need that? The error hides behind container.

packages/instant/src/redux/reducer.ts
```diff
@@ -7,13 +7,22 @@ import { AsyncProcessState } from '../types';
...
+interface BaseState {
     selectedAssetData?: string;
     selectedAssetAmount?: BigNumber;
     selectedAssetBuyState: AsyncProcessState;
     ethUsdPrice?: BigNumber;
     latestBuyQuote?: BuyQuote;
 }
+interface StateWithError extends BaseState {
+    latestError: any;
+    latestErrorDismissed: boolean;
```
fragosti: what if we call this `latestErrorState` and it can be either `Present` or `Hidden` or something. And lets not allow it to be `undefined`.

packages/instant/src/util/error_description.ts
```diff
@@ -0,0 +1,23 @@
...
+export const errorDescription = (error?: any, assetData?: string): { icon: string; message: string } => {
+    let bestMessage: string | undefined;
+    if (error instanceof Error) {
+        bestMessage = humanReadableMessageForError(error, assetData);
+    }
+    return {
+        icon: '😢',
+        message: bestMessage || 'Something went wrong...',
+    };
+};
```
fragosti: I would perhaps create one `util/error.ts` file that has one `error` or `errorUtil` object export and also provides the `flashNewError` utility?

packages/instant/src/util/asset_data.ts
```diff
@@ -0,0 +1,18 @@
+import { AssetProxyId } from '@0xproject/types';
+
+import { assetMetaData } from '../data/asset_meta_data';
+
+// TODO: tests for this
+export const bestNameForAsset = (assetData: string | undefined, defaultString: string) => {
```
fragosti: I like the way this is structured but I'd like to follow our somewhat soft convention of exporting one thing from file.
So the api would be `assetDataUtils.bestName()`

packages/instant/src/redux/reducer.ts
```diff
@@ -7,13 +7,22 @@ import { AsyncProcessState } from '../types';
...
+}
+interface StateWithoutError extends BaseState {
+    latestError: undefined;
+    latestErrorDismissed: undefined;
+}
+export type State = StateWithError | StateWithoutError;
```
BMillman19: What if we had two types, `BaseState` that looks like above and `ErrorState` that looks like this:
```typescript
interface ErrorState {
    latestError?: any;
    latestErrorDismissed?: boolean;
}
```

packages/instant/src/util/asset_data.ts
```diff
@@ -0,0 +1,18 @@
...
+    if (assetData === undefined) {
```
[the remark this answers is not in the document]
steveklebanoff: ah I noticed that, but I try to avoid importing a library function if it's simple, easy and readable to perfrom the same thing without it, a la https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore

packages/instant/src/util/asset_data.ts
```diff
@@ -0,0 +1,18 @@
...
+        return defaultString;
+    }
+    const metaData = assetMetaData[assetData];
+    if (metaData === undefined) {
+        return defaultString;
+    }
+    if (metaData.assetProxyId === AssetProxyId.ERC20) {
+        return metaData.symbol.toUpperCase();
```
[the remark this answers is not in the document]
steveklebanoff: technically yes, I thought about doing this: I generally like using textTransform for displaying things in all uppercase, But, this function would have to return a JSX/HTML element instead of just a string.  I prefer the simplicity of the way it is now, but I could be convinced if you feel strongly

packages/instant/src/redux/store.ts
steveklebanoff: I like using the package and having it be dev only, fixed in 3f0d7c70bd156edc4039d08737e74e99c0b3a04e

packages/instant/src/redux/reducer.ts
steveklebanoff: I like this -- feels more clear than undefined or a boolean 💯.  Will update!

packages/instant/src/util/error_description.ts
steveklebanoff: I'll try this and see how it feels once implemented. `ErrorFlasher` is kinda big though, not sure if I love it being grouped together with other things

packages/instant/src/util/asset_data.ts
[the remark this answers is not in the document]
fragosti: Yea I kind of agree? Don't feel too strongly. However, as mentioned previously, I am a big fan of consistency and reducing decisions. It's nice to not have to think about "how am I going to check for undefined". And it's also nice for the entire codebase to do it the same way. Idk, I think since we already have lodash in the bundle it's worth doing that, but don't feel strongly about it.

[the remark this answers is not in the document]
steveklebanoff: >  I think since we already have lodash in the bundle it's worth doing that

If we are able to get our bundle optimized, the bundle should only include the lodash functions that we use, and it'd be nice to be pulling in one less function if we consistently used this approach for checking for `undefined`

That being said, I hear you RE: consistency and reducing decisions.  For the sake of those two points, I'm down to use the lodash function for now. We can revisit our approach to lodash in general as a team later, and perhaps establish a new consistent pattern moving forward at that point.

packages/instant/src/containers/latest_error.tsx
fragosti: You could change it so that `latestError` sticks around and you simply toggle it up and down in the flasher / redux. This kind of seems better to me because then it becomes impossible to make the error just disappear?

packages/instant/src/redux/reducer.ts
steveklebanoff: I like this idea -- but I think with the change to `Present | Hidden` type which is always present might not be as important

packages/instant/src/util/asset_data.ts
steveklebanoff: I wonder if this convention could be bad for optimizing builds.  Not sure.  Cool w/ keeping convention for now though

packages/instant/src/util/error_description.ts
steveklebanoff: Done in d45ffb33513057789a2e1cf03a9f01ee05846e17

packages/instant/src/components/instant_heading.tsx
```diff
@@ -47,11 +73,11 @@ export const InstantHeading: React.StatelessComponent<InstantHeadingProps> = pro
             <Flex direction="column" justify="space-between">
                 <Container marginBottom="5px">
                     <Text fontSize="16px" fontColor={ColorOption.white} fontWeight={500}>
-                        {displaytotalEthBaseAmount(props)}
+                        {loadingOrAmount(props.quoteState, displaytotalEthBaseAmount(props))}
```
fragosti: Feels kind of weird to have two nested `Text` components. Can't `displayTotalEthUsdAmount` take care of all of this?

packages/instant/src/containers/selected_asset_amount_input.ts
```diff
@@ -60,9 +72,14 @@ const mapDispatchToProps = (
...
         dispatch(actions.updateLatestBuyQuote(undefined));
         // reset our buy state
-        dispatch(actions.updateSelectedAssetBuyState(AsyncProcessState.NONE));
-        // tslint:disable-next-line:no-floating-promises
-        debouncedUpdateBuyQuoteAsync(dispatch, assetData, value);
+        dispatch(actions.updatebuyOrderState(AsyncProcessState.NONE));
+
+        if (!_.isUndefined(value) && !_.isUndefined(assetData)) {
+            // even if it's debounced, give them the illusion it's loading
+            dispatch(actions.updateBuyQuoteStatePending());
```
fragosti: ah good idea

packages/instant/src/redux/actions.ts
```diff
@@ -25,12 +25,22 @@ export enum ActionTypes {
...
+    SET_ERROR = 'SET_ERROR',
+    HIDE_ERROR = 'HIDE_ERROR',
+    CLEAR_ERROR = 'CLEAR_ERROR',
 }
 
 export const actions = {
     updateEthUsdPrice: (price?: BigNumber) => createAction(ActionTypes.UPDATE_ETH_USD_PRICE, price),
     updateSelectedAssetAmount: (amount?: BigNumber) => createAction(ActionTypes.UPDATE_SELECTED_ASSET_AMOUNT, amount),
-    updateSelectedAssetBuyState: (buyState: AsyncProcessState) =>
+    updatebuyOrderState: (buyState: AsyncProcessState) =>
```
fragosti: nit: this is not `camelCase`

packages/instant/src/redux/actions.ts
```diff
@@ -25,12 +25,22 @@ export enum ActionTypes {
     UPDATE_SELECTED_ASSET_AMOUNT = 'UPDATE_SELECTED_ASSET_AMOUNT',
     UPDATE_SELECTED_ASSET_BUY_STATE = 'UPDATE_SELECTED_ASSET_BUY_STATE',
     UPDATE_LATEST_BUY_QUOTE = 'UPDATE_LATEST_BUY_QUOTE',
+    UPDATE_BUY_QUOTE_STATE_PENDING = 'UPDATE_BUY_QUOTE_STATE_PENDING',
+    UPDATE_BUY_QUOTE_STATE_FAILURE = 'UPDATE_BUY_QUOTE_STATE_FAILURE',
```
fragosti: `SET_BUY_QUOTE_STATE_PENDING` & `SET_BUY_QUOTE_STATE_FAILURE` ? Update means it takes an argument.

packages/instant/src/redux/reducer.ts
```diff
@@ -7,21 +7,31 @@ import { AsyncProcessState } from '../types';
 
 import { Action, ActionTypes } from './actions';
 
+export enum LatestErrorDisplay {
```
fragosti: This should go in `types.ts` and can probably be more general.

packages/instant/src/util/error.ts
```diff
@@ -0,0 +1,62 @@
...
+        const assetName = assetDataUtil.bestNameForAsset(assetData, 'This asset');
+        return `${assetName} is currently unavailable`;
+    }
+
+    return undefined;
+};
+
+export const errorUtil = {
+    errorFlasher: new ErrorFlasher(),
+    errorDescription: (error?: any, assetData?: string): { icon: string; message: string } => {
```
fragosti: Thanks for doing this!

packages/instant/src/components/instant_heading.tsx
steveklebanoff: Good call -- yes, will fix

packages/instant/src/redux/actions.ts
steveklebanoff: Ah, wasn't hip to this convention.  Will change.

PR #1144, opened 2018-10-17
[order_utils.py] ERC20 asset data encoding and decoding

feuGeneA:
## Description

image

The examples in the doc pic are rendered from comments in the code, and they're also extracted as tests during unit test runs, so what you see are verified examples.  The hex data was taken from real outputs from the TypeScript implementation, as shown in the asset_data_utils.ts changes.

In addition to the ERC20 codec, also:
* Stopped ignoring type errors on 3rd party imports, by including interface stubs for them;
* Removed the unimplemented `signature-utils` module, which was just a stand-in when the python project support was first put in place.

## Types of changes

## Checklist:

*   [x] For type assertions, make variable name be in the exception message
*   [x] Use word "assert" in the names of the type assertion routines.
*   [x] Figure out how Python type stubs work and try to get rid of the `# type: ignore` that's written all over the place.


python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    return Web3.sha3(text=signature).hex()
+
+
+def method_id(name: str, types: List[str]) -> str:
+    """Return the 4-byte method identifier.
+
+    >>> method_id("ERC20Token", ["address"])
+    '0xf47261b0'
+    """
+    return event_id(name, types)[0:10]
```
fabioberger: Is `event_id` guaranteed to always be longer then 10 characters?

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    >>> parse_signature("ERC20Token(address)")
+    {'method': 'ERC20Token', 'args': ['address']}
+    """
+    matches = re.match(r"^(\w+)\((.+)\)$", signature)
+    if matches is None:
+        raise Exception("Invalid method signature")
+    return {"method": matches[1], "args": matches[2].split(",")}
+
+
+def elementary_name(name: str) -> str:
```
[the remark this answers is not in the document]
feuGeneA: Camel-case isn't "pythonic."  I chose the names per [standard coding conventions](https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments), and, [to satisfy the linters](http://pylint-messages.wikidot.com/messages:c0103).

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    """Convert from short to canonical names; barely implemented.
```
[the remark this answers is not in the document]
feuGeneA: I was on the fence about whether to keep this method or not.  I left it in place as a clue for the next person who works on this, as a clue that they need to do something to support other data types.  Some examples are `uint` -> `uint256`, and `fixed` -> `fixed128x128`; see more in [the ethereumjs-abi implementation](https://github.com/ethereumjs/ethereumjs-abi/blob/00ba8463a7f7a67fcad737ff9c2ebd95643427f7/lib/index.js#L9).

python-packages/order_utils/src/zero_ex/abi_utils.py
feuGeneA: Yes, because it's a Keccak-256 checksum, as returned from `Web3.sha3()`, which is always 64 hex digits, as suggested by some research on the algorithm, and as verified by anecdotally checking digest lengths for a few different message inputs (an empty string, and a 64-character string).

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,83 @@
...
+    args: List[str]
+
+
+def parse_signature(signature: str) -> MethodSignature:
+    """Parse a method signature into its constituent parts.
+
+    >>> parse_signature("ERC20Token(address)")
+    {'method': 'ERC20Token', 'args': ['address']}
+    """
+    matches = re.match(r"^(\w+)\((.+)\)$", signature)
```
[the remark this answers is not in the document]
feuGeneA: Done; see https://github.com/0xProject/0x-monorepo/pull/1144/commits/8c21d665d9654e90d7d7d835855e4714f47ebcb8 .

python-packages/order_utils/src/zero_ex/abi_utils.py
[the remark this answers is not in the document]
feuGeneA: Done; see https://github.com/0xProject/0x-monorepo/pull/1144/commits/eb03916c0a6800921b054533c46c880466dec3b5 .

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,83 @@
...
+
+    >>> elementary_name("address")
+    'address'
+    """
+    if name != "address":
+        raise Exception("Unsupported name " + name)
+    return name
+
+
+def event_id(name: str, types: List[str]) -> str:
```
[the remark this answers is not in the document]
feuGeneA: Done; see https://github.com/0xProject/0x-monorepo/pull/1144/commits/eb03916c0a6800921b054533c46c880466dec3b5 .

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,113 @@
...
+def parse_signature(signature: str) -> MethodSignature:
+    """Parse a method signature into its constituent parts.
+
+    >>> parse_signature("ERC20Token(address)")
+    {'method': 'ERC20Token', 'args': ['address']}
+    """
+    if not isinstance(signature, str):
+        raise TypeError(
+            "signature should be str, not " + type(signature).__name__
+        )
```
fabioberger: Can we move this to an `assert` module? And just call `assert.isString('signature', signature)` here? We will want to do this exact same thing elsewhere. This way we also don't copy-paste this boilerplate error message in 100 places.

python-packages/order_utils/src/zero_ex/abi_utils.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    if not isinstance(name, str):
+        raise TypeError("name should be str, not " + type(name).__name__)
+
+    return {
+        "int": "int256",
+        "uint": "uint256",
+        "fixed": "fixed128x128",
+        "ufixed": "ufixed128x128",
+    }.get(name, name)
```
fabioberger: Will this return an intelligible error message if it's not found? From the docs: 

```
Return the value for key if key is in the dictionary, else default. If default is not given, it defaults to None, so that this method never raises a KeyError.
```
https://docs.python.org/2/library/stdtypes.html#dict.get

I think we want to throw if returned `None`. 

General feedback: Let's prefer robust code with clear errors for library users over succinctness.

feuGeneA: I agree this looks weird, but I'm also not terribly familiar with all the pythonic idioms, so I'm not sure how many pythonistas would read it clearly.  There is no switch statement in python, and this was the alternative I liked best.  The second parameter to the `get()` method is the `default`, so here if the value of `name` isn't found in the dict, then `get()` will just return that value.  A naive approach might be to use `}[name]`, and then you'd have to worry about `KeyError`s, but [it seems that the whole purpose of `.get()` is for this specific case](https://stackoverflow.com/a/11041421/406289).

fabioberger: Hm... I don't think falling back to the `name` passed in makes sense here. This way the dev will expect it to map and it won't but also doesn't throw, so they will only find out later on in their code.

Let's check if `get` returned `None`, and if so, throw loudly.

feuGeneA: This method is essentially a port of [`ethereumjs-util`s `elementaryName`](https://github.com/ethereumjs/ethereumjs-abi/blob/00ba8463a7f7a67fcad737ff9c2ebd95643427f7/lib/index.js#L9), which returns what was given if it doesn't know of a mapping.  And we depend on this behavior in our use case, where we pass in `"address"`, which doesn't have a mapping.  My understanding is that some types (like `address`) have the same representation in both `short name` and `canonical name` forms, so no mapping/translation needs to be done for those types.  And of course you could pass in types that area already in `canonical name` form, such as `uint256`, which would need no translation.  In short, I think it would be excessive to explicitly code for all of the types that should be supported, and that a "pass-through" functionality, at this relatively low level, is perfectly appropriate.  That's my opinion, and why I have it, but I'm open to having my mind changed.

python-packages/order_utils/src/zero_ex/abi_utils.py
feuGeneA: [Done](https://github.com/0xProject/0x-monorepo/pull/1144/commits/08e8da00467b1bfa88d9ccbb905aaa31fc9e3b8c)

python-packages/order_utils/setup.py
```diff
@@ -34,14 +37,29 @@ def run(self):
...
             # pylint takes relatively long to run, so it runs last, to enable
             # fast failures.
         ]
+
+        # tell mypy where to find interface stubs for 3rd party libs
+        environ["MYPYPATH"] = path.join(
+            path.dirname(path.realpath(argv[0])), "stubs"
+        )
+
+        # HACK: until eth_abi releases
```
fabioberger: `HACK(gene):`

PR #1147, opened 2018-10-17
[asset-buyer][contract-wrappers] Throw SignatureRequestDenied and TransactionValueTooLow errors from AssetBuyer

BMillman19:
## Description

* Expanded the `AssetBuyerError` enum to include  `AssetBuyerError.SignatureRequestDenied` and `AssetBuyerError.TransactionValueTooLow`
* `SignatureRequestDenied` is thrown by the `executeBuyQuoteAsync()` method when the end-user denies the signing request
* The above is achieved buy extending the errorTransformers in the contract-wrappers decorators to handle errors thrown by MetaMask
* `TransactionValueTooLow` is thrown by the`executeBuyQuoteAsync()` method when forwarder market buy validation comes back with an error. This means that the input we formed to send to the forwarder will most likely result in a transaction failure. We should handle this error at the 0x Instant level and reset the current buyQuote

Todo:

- [ ] Test denying signing from ledger
- [ ] Add tests for AssetBuyer

## Testing instructions

`yarn test`

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-wrappers/src/utils/constants.ts
+++ b/packages/contract-wrappers/src/utils/constants.ts
@@ -14,4 +14,5 @@ export const constants = {
     ZERO_AMOUNT: new BigNumber(0),
     ONE_AMOUNT: new BigNumber(1),
     ETHER_TOKEN_DECIMALS: 18,
+    METAMASK_DENIED_SIGNATURE_PATTERN: 'MetaMask Tx Signature: User denied transaction signature',
 };
--- a/packages/contract-wrappers/src/utils/constants.ts
+++ b/packages/contract-wrappers/src/utils/constants.ts
@@ -14,5 +14,5 @@ export const constants = {
     ZERO_AMOUNT: new BigNumber(0),
     ONE_AMOUNT: new BigNumber(1),
     ETHER_TOKEN_DECIMALS: 18,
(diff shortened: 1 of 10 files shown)


packages/monorepo-scripts/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "1.0.6",
+        "changes": [
+            {
+                "note": "Add ForwarderError to IGNORED_EXCESSIVE_TYPES in docGenConfigs"
```
fragosti: pr

packages/contract-wrappers/src/utils/constants.ts
```diff
@@ -14,4 +14,5 @@ export const constants = {
     ZERO_AMOUNT: new BigNumber(0),
     ONE_AMOUNT: new BigNumber(1),
     ETHER_TOKEN_DECIMALS: 18,
+    METAMASK_DENIED_SIGNATURE_PATTERN: 'MetaMask Tx Signature: User denied transaction signature',
```
dekz: If you relax this a little you may also be able to catch Coinbase Wallet and some other wallet apps.

Looks like Coinbase Wallet does throw `User denied transaction signature` when rejecting a tx, might be worth relaxing to catch both?

PR #1159, opened 2018-10-18
[instant] Pass in liquiditySource, assetData and other settings from render

fragosti:
## Description

The branch mostly implements passing various settings from the top-level `render` method instead of hard coding them and then dealing with the implications.

- Added the concept of top-level props being added to the redux store state. 
- Added an `Asset` concept, which contains both assetData and the metaData.
- Added what is basically a join table for assetDatas across networks. Bad idea?
- Added an `AssetBuyer` to state - the plan is for us to re-instantiate this whenever some salient part of state changes (the provider?).
- Added network to state and use that information where relevant (such as when initializing AssetBuyer).
- The value of `primaryColor` is automatically updated when the `selectedAsset` is changed.

It's worth noting that we don't support changing `ZeroExInstant` props dynamically (ie. in the case where `ZeroExInstant` is being used in a React project) because the props are used to initialize the redux store on init and then ignored afterwards. For now this is probably ok, and reduces the complexity of the code.

Another thing worth noting is that the `metaData` field in the `Asset` object is not optional. If we don't find `metaData` regarding an assetData, then we throw an error. The rest of the app depends on us having this data. The plus side is that the integrator can (only) append to the default metaData we provide. This made more sense to me than the alternative because: 
1. We need **decimals** which is part of the `metaData` to even get a `BuyQuote`.
2. We need the rest of the data to deliver a good UX.
3. It reduces the complexity of the code.

## Questions
1. Do we want to support dynamically changing `ZeroExInstant` props/configs? If not, do we want to take it even further and not even pass that data into the redux `State` and instead write it to some global object we can always reference it?

## Checklist:


packages/instant/src/data/asset_data_network_mapping.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const sameAssetDatas: AssetDataByNetwork[] = [
+    {
+        [Network.Mainnet]: '0xf47261b0000000000000000000000000e41d2489571d322189246dafa5ebde1f4699f498',
+        [Network.Kovan]: '0xf47261b00000000000000000000000002002d3812f58e35f0ea1ffbf80a75a38c32175fa',
+    },
+];
+
+export const assetDataNetworkMapping = {
```
fragosti: remove this from data.. should be in utils

packages/instant/src/util/asset.ts
```diff
@@ -0,0 +1,34 @@
...
+        let mainnetAssetData: string | undefined = assetData;
+        if (network !== Network.Mainnet) {
+            mainnetAssetData = assetDataNetworkMapping.getAssociatedAssetDataIfExists(assetData, network);
+        }
+        if (_.isUndefined(mainnetAssetData)) {
+            throw new Error(ZeroExInstantError.AssetMetaDataNotAvailable);
+        }
+        const metaData = metaDataMap[mainnetAssetData];
+        if (_.isUndefined(metaData)) {
+            throw new Error();
```
fragosti: woops!

packages/instant/src/types.ts
```diff
@@ -22,12 +22,31 @@ export interface ERC20AssetMetaData {
...
+export interface ERC20Asset {
+    assetData: string;
+    metaData: ERC20AssetMetaData;
+}
+
+export interface ERC721Asset {
+    assetData: string;
+    metaData: ERC721AssetMetaData;
+}
+export interface Asset {
```
steveklebanoff: minor: would `Asset` being `ERC20Asset | ERC721Asset` be equivalent/easier to read?

packages/instant/test/util/asset_data.test.ts
```diff
@@ -1,17 +0,0 @@
-import { assetDataUtil } from '../../src/util/asset_data';
```
steveklebanoff: Think it'd be worth it to convert these tests to the new signature instead of removing them?

packages/instant/src/components/zero_ex_instant.tsx
```diff
@@ -1,23 +1,75 @@
...
-            <ZeroExInstantContainer />
-        </ThemeProvider>
-    </Provider>
-);
+
+export type ZeroExInstantProps = ZeroExInstantRequiredProps & Partial<ZeroExInstantOptionalProps>;
+
+export interface ZeroExInstantRequiredProps {
+    // TODO: Change API when we allow the selection of different assetDatas
+    assetData: string;
```
steveklebanoff: Would it make sense to call this something like `initialAssetData`?

packages/instant/src/util/asset.ts
```diff
@@ -0,0 +1,53 @@
...
+        assetMetaDataMap: ObjectMap<AssetMetaData>,
+        network: Network,
+    ): Asset => {
+        return {
+            assetData,
+            metaData: assetUtils.getMetaDataOrThrow(assetData, assetMetaDataMap, network),
+        };
+    },
+    getMetaDataOrThrow: (assetData: string, metaDataMap: ObjectMap<AssetMetaData>, network: Network): AssetMetaData => {
+        let mainnetAssetData: string | undefined = assetData;
```
steveklebanoff: it looks like this could actually contain data about another network (i.e. kovan) - is this correct?  if so, we should give a different name

packages/instant/src/components/zero_ex_instant.tsx
```diff
@@ -1,23 +1,75 @@
...
+    // TODO: Allow for a function that returns orders
+    liquiditySource: string;
```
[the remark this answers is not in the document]
fragosti: We spec'ed this out and it has to be called `liquiditySource` because it can also be a function. Open to changing this though. 
https://0xproject.quip.com/hBmlA07jXbEz/0x-Instant-Public-Spec

packages/instant/src/components/buy_button.tsx
```diff
@@ -1,16 +1,16 @@
...
 import { ColorOption } from '../style/theme';
-import { assetBuyer } from '../util/asset_buyer';
 import { util } from '../util/util';
 import { web3Wrapper } from '../util/web3_wrapper';
 
 import { Button, Container, Text } from './ui';
 
 export interface BuyButtonProps {
     buyQuote?: BuyQuote;
+    assetBuyer?: AssetBuyer;
```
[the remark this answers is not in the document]
fragosti: I agree -- I thought about how to do this for `assetBuyer` and for other parts of the state. The main issue is the provider... I guess we could point to a default RPC endpoint on init?

packages/instant/src/types.ts
fragosti: For reasons that are not entirely clear to me, they are not equivalent and `type Asset = ERC20Asset | ERC721Asset` does not compile when you are trying to create an `Asset` that has a `metaData` field of type `AssetMetaData`

packages/instant/src/redux/actions.ts
```diff
@@ -38,6 +39,7 @@ export const actions = {
     updatebuyOrderState: (buyState: AsyncProcessState) =>
         createAction(ActionTypes.UPDATE_SELECTED_ASSET_BUY_STATE, buyState),
     updateLatestBuyQuote: (buyQuote?: BuyQuote) => createAction(ActionTypes.UPDATE_LATEST_BUY_QUOTE, buyQuote),
+    updateSelectedAsset: (assetData?: string) => createAction(ActionTypes.UPDATE_SELECTED_ASSET, assetData),
```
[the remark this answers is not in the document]
fragosti: It is not used. I think I was just prototyping to see if the structure of things made sense to support this feature. Generally agree that there shouldn't be dead code in a PR. I can remove it, but I do think this will be it's implementation... so probably fine to leave?

packages/instant/src/components/zero_ex_instant.tsx
fragosti: As the comment says, I think this will change to be a list of assetDatas or nothing at all.

packages/instant/src/util/asset.ts
fragosti: To address this I simply change the order of the assignments such that it's never a lie essentially.

packages/instant/src/redux/actions.ts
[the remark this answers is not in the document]
steveklebanoff: cool w/ leaving this

packages/instant/src/components/zero_ex_instant.tsx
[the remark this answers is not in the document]
steveklebanoff: Thanks for the link -- wasn't hip to this.  

The spec says `liquiditySource` can be "a function that returns a list of signed 0x orders".  If we are actually going to support that in the future, `liquiditySource`is a good name

packages/instant/src/types.ts
steveklebanoff: Cool w/ leaving as is, not a big deal

packages/instant/src/util/asset.ts
steveklebanoff: I feel like I'm missing something, so feel free to correct here: to me, the following code looks like the variable name `mainnnetAssetData` could still actually be assigned to testnet data:

```ts
        if (network !== Network.Mainnet) {
            mainnetAssetData = assetUtils.getAssociatedAssetDataIfExists(assetData, network);
        }
```

If this is the case, could we just rename to `networkAssetData`?

packages/instant/test/util/asset.test.ts
```diff
@@ -0,0 +1,47 @@
+import { AssetProxyId, ObjectMap } from '@0x/types';
```
steveklebanoff: thanks for updating these tests 🙏 !

packages/instant/src/components/buy_button.tsx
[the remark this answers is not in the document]
steveklebanoff: Per slack convo, there will be times when `assetBuyer` is undefined in the future, i.e. when a user is selecting their wallet.  But, we will never render this `BuyButton` when `assetBuyer` is undefined.  To make things a bit cleaner, we can update `assetBuyer` to be required for `BuyButton`, and not render `BuyButton` if there is no `assetBuyer` on the state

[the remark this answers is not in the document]
fragosti: @steveklebanoff mind if I do this as a follow up? I think there are a few branches that depend on this one now.

packages/instant/src/types.ts
fragosti: @BMillman19 we need `ERC20Asset` for the amount input yea

packages/instant/src/components/buy_button.tsx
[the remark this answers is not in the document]
steveklebanoff: yes, i'm cool w/ this being done in a follow up PR 👍

packages/instant/src/util/asset.ts
fragosti: Ah. So in my view metaData never changes (maybe this is not the case?), so I didn't feel like duplicating it. So instead I want to be able to take an assetData string that is specific to Kovan, find the corresponding mainnet assetData string, and then use the mainnet assetData string to access our mapping of assetDatas to metaData

packages/instant/src/components/asset_amount_input.tsx
```diff
@@ -2,37 +2,38 @@ import { BigNumber } from '@0x/utils';
...
-
 import { ColorOption } from '../style/theme';
+import { ERC20Asset } from '../types';
+import { assetUtils } from '../util/asset';
 import { util } from '../util/util';
 
 import { AmountInput, AmountInputProps } from './amount_input';
 import { Container, Text } from './ui';
 
+// Asset amounts only apply to ERC20 assets
```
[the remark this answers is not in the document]
fragosti: I agree -- I renamed this in my other branch so going to do this as a followup.

packages/instant/src/components/buy_button.tsx
```diff
@@ -37,13 +37,13 @@ export class BuyButton extends React.Component<BuyButtonProps> {
     }
     private readonly _handleClick = async () => {
         // The button is disabled when there is no buy quote anyway.
-        if (_.isUndefined(this.props.buyQuote)) {
+        if (_.isUndefined(this.props.buyQuote) || _.isUndefined(this.props.assetBuyer)) {
```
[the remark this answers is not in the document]
fragosti: Going to add it for now. Will do the optionality stuff in a followup

packages/instant/src/data/asset_data_network_mapping.ts
```diff
@@ -0,0 +1,15 @@
+import * as _ from 'lodash';
+
+import { Network } from '../types';
+
+interface AssetDataByNetwork {
```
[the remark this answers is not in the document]
fragosti: not sure what this would get us.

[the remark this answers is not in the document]
fragosti: apparently not.

packages/instant/src/data/asset_data_network_mapping.ts
```diff
@@ -0,0 +1,15 @@
...
+    [Network.Kovan]?: string;
+    [Network.Mainnet]?: string;
+}
+
+export const assetDataNetworkMapping: AssetDataByNetwork[] = [
```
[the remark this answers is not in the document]
fragosti: I'm assuming that metaData is always the same across networks, is that not the case?

packages/instant/src/util/asset.ts
steveklebanoff: Ah, I understand now.  Cool as-is.

PR #1161, opened 2018-10-18
[instant] Order details loading state

steveklebanoff:
## Description

- Adds concept of pulsing "-" when order details are loading
- Resolves additional feedback left on https://github.com/0xProject/0x-monorepo/pull/1142 

## Checklist:


packages/instant/tslint.json
```diff
@@ -2,6 +2,7 @@
     "extends": ["@0x/tslint-config"],
     "rules": {
         "custom-no-magic-numbers": false,
-        "semicolon": [true, "always", "ignore-bound-class-methods"]
+        "semicolon": [true, "always", "ignore-bound-class-methods"],
+        "max-classes-per-file": false
```
steveklebanoff: I refactored the `EthAmountRow`component to be a class instead of a function, and tslint started complaining.  This felt like a silly distinction, so thought we should turn off the rule -- open to thoughts on this though

packages/instant/src/components/order_details.tsx
```diff
@@ -58,43 +66,51 @@ export interface EthAmountRowProps {
...
-                    {rowLabel}
-                </Text>
-                <Container>
-                    {usdPriceSection}
+export class EthAmountRow extends React.Component<EthAmountRowProps> {
+    public static defaultProps = {
+        shouldEmphasize: false,
+        isEthAmountInBaseUnits: true,
+    };
+    public static displayName = 'EthAmountRow';
```
fragosti: I don't think classes need this

packages/instant/src/components/order_details.tsx
```diff
@@ -58,43 +66,51 @@ export interface EthAmountRowProps {
...
+                                    <AmountPlaceholder color={ColorOption.lightGrey} isPulsating={isLoading} />
+                                </Container>,
+                            )}
+                        </Text>
+                    </Container>
+                </Flex>
+            </Container>
+        );
+    }
+    private _usdSection(): React.ReactNode {
```
fragosti: nit: `_renderUsdSection`

packages/instant/src/components/instant_heading.tsx
```diff
@@ -7,79 +7,82 @@ import { ColorOption } from '../style/theme';
...
-};
 
-const displayUsdAmount = ({
-    totalEthBaseAmount,
-    selectedAssetAmount,
-    ethUsdPrice,
-}: InstantHeadingProps): React.ReactNode => {
-    if (_.isUndefined(selectedAssetAmount)) {
-        return '$0.00';
+    private _placeholderAmountToAlwaysShow(): React.ReactNode | null {
```
fragosti: nit: `_placeholderAmountToAlwaysShow` naming is a bit confusing, we aren't always showing it right?

Also I think `React.ReactNode` has `null` in it I think

packages/instant/src/components/order_details.tsx
steveklebanoff: got it -- i like the `render` prefix convention 🎉

packages/instant/src/components/instant_heading.tsx
steveklebanoff: Good point -- looking at this again, this is confusing. We aren't always showing it. 

This started as "shouldAlwaysShowPlaceHolder: boolean`, but then turned into the function returning the placeholder because it specifies the pulsating prop differently based on a condition.

packages/instant/src/components/amount_placeholder.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { ColorOption } from '../style/theme';
+
+import { Pulse } from './animations/pulse';
+
+import { Text } from './ui';
+
+export interface PlainPlaceholder {
+    color: ColorOption;
+}
+export const PlainPlaceholder: React.StatelessComponent<PlainPlaceholder> = props => (
```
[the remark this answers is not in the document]
steveklebanoff: I'm down.   TBH I was feeling a little unsure as to what the 0x pattern/convention was for deciding what to export.

packages/instant/src/components/instant_heading.tsx
steveklebanoff: Fixed with 2f4e498

steveklebanoff:
@fragosti updated!

packages/instant/src/components/instant_heading.tsx
```diff
@@ -43,22 +42,24 @@ export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
...
+                        </Container>
+                        <Container opacity={0.7}>{this._placeholderOrAmount(this._dollarAmount.bind(this))}</Container>
                     </Flex>
                 </Flex>
             </Container>
         );
     }
 
-    private _placeholderAmountToAlwaysShow(): React.ReactNode | null {
+    private _placeholderOrAmount(amountFunction: () => React.ReactNode): React.ReactNode {
```
fragosti: why pass in the function when you can just pass in the value?

Also you don't have to bind if you use arrow functions as seen elsewhere in the code.

steveklebanoff: passed in a function instead of a value for a small optimization: didn't want the function to evaluate if it didn't have to

PR #1170, opened 2018-10-19
[instant] Failure state

steveklebanoff:
## Description

- Adds failure icon and "Try again" button to instant when failure happens

## Types of changes

* New Feature

## Checklist:


packages/instant/src/containers/selected_asset_button.tsx
```diff
@@ -0,0 +1,36 @@
...
+    buyOrderState: state.buyOrderState,
+});
+
+const SelectedAssetButtonPresentationComponent: React.StatelessComponent<{
+    buyOrderState: AsyncProcessState;
+}> = props => {
+    if (props.buyOrderState === AsyncProcessState.FAILURE) {
+        return <SelectedAssetRetryButton />;
+    } else if (props.buyOrderState === AsyncProcessState.SUCCESS) {
+        return <SecondaryButton text="Success" isDisabled={true} />;
```
steveklebanoff: Will make this and processing buttons nicer in separate PR

packages/instant/src/components/retry_button.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+import { SecondaryButton } from './secondary_button';
+
+export interface RetryButtonProps {
+    onClick: () => void;
+}
+
+export const RetryButton: React.StatelessComponent<RetryButtonProps> = props => {
```
fragosti: Do you foresee adding functionality to this in the future? If not, seems a bit overkill to me (you could just inline the JSX).

packages/instant/src/components/secondary_button.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { Button, ButtonProps } from './ui/button';
+import { Text } from './ui/text';
+
+export interface SecondaryButtonProps extends ButtonProps {
+    text: string;
+}
+
+export const SecondaryButton: React.StatelessComponent<SecondaryButtonProps> = props => {
+    const buttonProps = _.omit(props, 'text');
```
fragosti: I usually do `{ text, ...rest }` but this works too. Just a heads up.

packages/instant/src/components/ui/icon.tsx
```diff
@@ -0,0 +1,58 @@
...
+    },
+};
+
+export interface IconProps {
+    width: number;
+    height: number;
+    color: ColorOption;
+    icon: keyof IconInfoMapping;
+}
+export const Icon: React.SFC<IconProps> = props => {
```
fragosti: sweet

packages/instant/src/containers/selected_asset_button.tsx
fragosti: This is great -- I would just expect the presentational component to be in `components` and just have the wiring in `containers`.

Also `BuyOrderStateButton` and `SelectedAssetBuyOrderStateButton`?

packages/instant/src/containers/selected_asset_retry_button.tsx
```diff
@@ -0,0 +1,26 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+import { connect } from 'react-redux';
+import { Dispatch } from 'redux';
+
+import { Action, actions } from '../redux/actions';
+
+import { RetryButton } from '../components/retry_button';
+
+export interface SelectedAssetRetryButtonProps {}
```
fragosti: I would remove this file and have all this logic be handled by `BuyOrderStateButton` and the connector.

packages/instant/src/redux/actions.ts
```diff
@@ -45,4 +46,5 @@ export const actions = {
     setError: (error?: any) => createAction(ActionTypes.SET_ERROR, error),
     hideError: () => createAction(ActionTypes.HIDE_ERROR),
     clearError: () => createAction(ActionTypes.CLEAR_ERROR),
+    clearBuyQuoteAndSelectedAssetAmount: () => createAction(ActionTypes.CLEAR_BUY_QUOTE_AND_SELECTED_ASSET_AMOUNT),
```
fragosti: I think it's ok to name these actions & action creators a bit more like an API... so this would be `resetSomethingSomething?` Just a thought.

packages/instant/src/components/instant_heading.tsx
```diff
@@ -9,17 +9,20 @@ import { format } from '../util/format';
...
     totalEthBaseAmount?: BigNumber;
     ethUsdPrice?: BigNumber;
     quoteRequestState: AsyncProcessState;
+    buyOrderState: AsyncProcessState;
 }
 
 const placeholderColor = ColorOption.white;
 export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
     public render(): React.ReactNode {
+        const iconOrAmounts = this._renderIcon() || this._renderAmountsSection();
```
BMillman19: I believe we need to bind `this` here OR make `_renderIcon` and `_renderAmountsSection` arrow functions, e.g.

```
private readonly _renderAmountsSection = () => {
        return (
            <Container>
                <Container marginBottom="5px">{this._placeholderOrAmount(this._ethAmount)}</Container>
                <Container opacity={0.7}>{this._placeholderOrAmount(this._dollarAmount)}</Container>
            </Container>
        );
    }
```
See `buy_button.tsx`

I think we prefer the arrow function approach when in a render method for performance reasons, a new function does not need to be created for each render cycle

packages/instant/src/components/instant_heading.tsx
```diff
@@ -36,20 +39,43 @@ export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
                         textTransform="uppercase"
                         fontSize="12px"
                     >
-                        I want to buy
+                        {this._renderTopText()}
```
BMillman19: same here

packages/instant/src/components/secondary_button.tsx
steveklebanoff: Actually, in a future branch I am going to have this to just render `children` instead of a `text` prop, so this will be changing shortly in the future, going to leave as-is in this PR

packages/instant/src/components/ui/icon.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+type svgRule = 'evenodd' | 'nonzero' | 'inherit';
+interface IconInfo {
+    viewBox: string;
+    fillRule?: svgRule;
+    clipRule?: svgRule;
+    path: string;
+}
+interface IconInfoMapping {
+    failed: IconInfo;
```
BMillman19: perhaps we can write it like this?

```
enum IconType {
    Failed,
    Success,
}

interface IconInfoByType {
    [key: IconType]: IconInfo
}
```

packages/instant/src/components/ui/icon.tsx
```diff
@@ -0,0 +1,58 @@
...
+        path:
+            'M17 34C26.3887 34 34 26.3888 34 17C34 7.61121 26.3887 0 17 0C7.61133 0 0 7.61121 0 17C0 26.3888 7.61133 34 17 34ZM25.7539 13.0977C26.2969 12.4718 26.2295 11.5244 25.6035 10.9817C24.9775 10.439 24.0303 10.5063 23.4878 11.1323L15.731 20.0771L12.3936 16.7438C11.8071 16.1583 10.8574 16.1589 10.272 16.7451C9.68652 17.3313 9.6875 18.281 10.2734 18.8665L14.75 23.3373L15.8887 24.4746L16.9434 23.2587L25.7539 13.0977Z',
+    },
+};
+
+export interface IconProps {
+    width: number;
+    height: number;
+    color: ColorOption;
+    icon: keyof IconInfoMapping;
```
BMillman19: With the above comment this can be `iconType: IconType`

packages/instant/src/containers/selected_asset_retry_button.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const mapDispatchToProps = (
+    dispatch: Dispatch<Action>,
+    _ownProps: SelectedAssetRetryButtonProps,
+): ConnectedDispatch => ({
+    onClick: () => dispatch(actions.clearBuyQuoteAndSelectedAssetAmount()),
+});
+
+export const SelectedAssetRetryButton: React.ComponentClass<SelectedAssetRetryButtonProps> = connect(
+    null,
```
BMillman19: we usually prefer undefined, unless we are writing tsx in a render method

packages/instant/src/containers/selected_asset_button.tsx
steveklebanoff: Thanks!  Putting this into a separate file feels like overkill to me: it's a very simple component and will not be reused anywhere else.  Having this defined inline like it is now feels easier to understand, and reduces having to jump into another file.  If this component were to grow larger, be more complex, or be more reusable, I'd be more up for putting it into a separate file.  I'm open to discuss more if you feel strongly though.

Also the `SelectedAsset[component name]` pattern is feeling a bit verbose to me.  It feels more natural to me to follow a pattern like `Connected[component name]`: wondering if you think that is too vague though?

packages/instant/src/components/ui/icon.tsx
steveklebanoff: This suggestion would make it easier to add new icons, I like it

packages/instant/src/containers/selected_asset_retry_button.tsx
steveklebanoff: I originally had it designed like that, but it felt to me like the connector was getting a bit overloaded, it had to connect a lot of things:

**To determine which button to render:**
 `buyOrderState`

**For buy button:**
`assetBuyer `, `buyQuote `, `onBuyClick`, `onBuySuccess`, `onBuyFailure`

**For retry button:**
`onRetryClick`

So even if we are just rendering the retry button, we are still connecting all of these buy button functions.  Felt like too much going on, should be separated out. But, I do see that the current implementation has a bit of boilerplate and and adds a little indirection.  Thoughts?

packages/instant/src/components/retry_button.tsx
steveklebanoff: This came out of the design decision documented here: https://github.com/0xProject/0x-monorepo/pull/1170/files#r227176413   I do agree it feels a little bit like overkill though

packages/instant/src/components/instant_heading.tsx
steveklebanoff: Per convo, since this doesn't need to be bound, current implementation is fone

packages/instant/src/components/ui/icon.tsx
steveklebanoff: Per convo, we would lose some typesafety here, going to leave as-is

packages/instant/src/containers/selected_asset_retry_button.tsx
steveklebanoff: Yes, Option 2 is what we have right now.  I"m down to  move the component definition, but would like to keep the architecture of Option 2 as-is for now

packages/instant/src/containers/selected_asset_button.tsx
steveklebanoff: Cool, will consider a name and update if I can think of something better than `Connected...`. 

I will move the component to it's own file to follow the convention

steveklebanoff: For the naming, what do you think about this and buttons like the retry button having the prefix of `BuyOrder..` or `CurrentBuyOrder` ?  I feel like these buttons are more closely related to the buy order than they are the selected asset

packages/instant/src/components/instant_heading.tsx
steveklebanoff: private instance methods like this feel more natural to me, would prefer to leave

steveklebanoff: although I could see the appeal of not having to think about this

fragosti: I personally always do arrow functions just so I don't have to think about it but don't feel strongly about it.

packages/instant/src/components/buy_order_state_button.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+
+import { SelectedAssetBuyButton } from '../containers/selected_asset_buy_button';
+import { SelectedAssetRetryButton } from '../containers/selected_asset_retry_button';
+
+import { AsyncProcessState } from '../types';
+
+import { SecondaryButton } from './secondary_button';
+
+export interface BuyOrderStateButtonProps {
```
fragosti: thank you!

PR #1175, opened 2018-10-23
[instant] Create a ScalingInput component and use it in the amount input and upgrade to styled-components v4

fragosti:
## Description

The upgrade to styled-components was pretty straightforward. I also discovered you don't have to use the `PlainComponent` convention we were using before which is cool.

The scaling input was more complicated than expected, and while I'm not 100% happy with it I think it works well enough for now. I capped the maxLength of the input to 9, so the maximum amount of token you can buy is ~1bil.

It would be nice if the input always took up the remaining space available -- it looks a little off for really high numbers. But for numbers of low enough magnitude it looks good enough.

Example:
screen shot 2018-10-22 at 5 50 09 pm

Reason why the max width is only 4 characters:
screen shot 2018-10-22 at 5 05 56 pm

Demo: http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com/

## Checklist:


packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+    public static getPhase(startWidthCh: number, endWidthCh: number, value?: string): ScalingInputPhase {
+        if (_.isUndefined(value) || value.length <= startWidthCh) {
+            return ScalingInputPhase.Start;
+        }
+        if (value.length > startWidthCh && value.length <= endWidthCh) {
+            return ScalingInputPhase.Scaling;
+        }
+        return ScalingInputPhase.End;
+    }
+    public static getPhaseFromProps(props: ScalingInputProps): ScalingInputPhase {
```
steveklebanoff: not a big deal, but: on first pass I would think to make this a private instance method that has access to `this.props`, as opposed to a static public function that takes in `props` as a param.  are there benefits to the static public function approach?

steveklebanoff: ah nevermind, I see that the static functions are necessary for the `componentDidUpdate` function, where you are using this function for the previous props and current props

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+    };
+    private readonly _calculateFontSize = (phase: ScalingInputPhase): number => {
+        return ScalingInput.calculateFontSizeFromProps(this.props, phase);
+    };
+    private readonly _getInputWidthInPx = (): number => {
+        const ref = this._inputRef.current;
+        if (!ref) {
+            return 0;
+        }
+        return (ref as any).getBoundingClientRect().width;
```
steveklebanoff: we could do `(ref as HTMLInputElement)` but not sure if this is much better, as `Input` could change and this could not get updated

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+        if (value.length > startWidthCh && value.length <= endWidthCh) {
+            return ScalingInputPhase.Scaling;
+        }
+        return ScalingInputPhase.End;
+    }
+    public static getPhaseFromProps(props: ScalingInputProps): ScalingInputPhase {
+        const { value, startWidthCh, endWidthCh } = props;
+        return ScalingInput.getPhase(startWidthCh, endWidthCh, value);
+    }
+    public static calculateFontSize(
```
steveklebanoff: nice 👍

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+        onChange: util.boundNoop,
+        onFontSizeChange: util.boundNoop,
+        maxLength: 9,
+    };
+    public state = {
+        fixedWidthInPxIfExists: undefined,
+    };
+    private readonly _inputRef = React.createRef();
+    public static getPhase(startWidthCh: number, endWidthCh: number, value?: string): ScalingInputPhase {
+        if (_.isUndefined(value) || value.length <= startWidthCh) {
```
steveklebanoff: In my mind I thought of these phases as "None", "Scaling", and "Max", but I'm cool w/ Start, Scaling and End

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+        const charactersOverMax = value.length - endWidthCh;
+        const scalingFactor = (1 - percentageToReduceByPerCharacter) ** charactersOverMax;
+        const fontSize = scalingFactor * maxFontSizePx;
+        return fontSize;
+    }
+    public static calculateFontSizeFromProps(props: ScalingInputProps, phase: ScalingInputPhase): number {
+        const { endWidthCh, value, maxFontSizePx } = props;
+        return ScalingInput.calculateFontSize(endWidthCh, maxFontSizePx, phase, value);
+    }
+    public getSnapshotBeforeUpdate(): ScalingInputSnapshot {
```
steveklebanoff: Never knew about this part of React 🆒

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
...
+        };
+    }
+    public componentDidUpdate(
+        prevProps: ScalingInputProps,
+        prevState: ScalingInputState,
+        snapshot: ScalingInputSnapshot,
+    ): void {
+        const prevPhase = ScalingInput.getPhaseFromProps(prevProps);
+        const curPhase = ScalingInput.getPhaseFromProps(this.props);
+        const prevFontSize = ScalingInput.calculateFontSizeFromProps(prevProps, prevPhase);
```
steveklebanoff: minor but I'd prefer for the `prevFontSize` and `curFontSize` assignments to be moved down to be right above `if (prevFontSize !== curFontSize) {...`, I think this would make it more clear that they aren't used until then.  

That being said, not sure if you have an established pattern of doing all the variable assignments at the beginning of a function

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,153 @@
+import * as _ from 'lodash';
```
steveklebanoff: Nice work on this component -- although it took a little bit of time for me to mentally parse all the logic, you did a good job separating out functionality into small functions that have proper scopes and names 💯

packages/instant/src/components/scaling_amount_input.tsx
```diff
@@ -0,0 +1,53 @@
+import { BigNumber } from '@0x/utils';
+import * as _ from 'lodash';
+import * as React from 'react';
+
+import { ColorOption } from '../style/theme';
+import { util } from '../util/util';
+
+import { ScalingInput } from './scaling_input';
+
+export interface ScalingAmountInputProps {
```
steveklebanoff: Quite a few of these are in `ScalingInputProps`, we could `pick` out the ones we care about for this interface instead of having to duplicate the type definitions.  Could go either way on this tho.

packages/instant/src/components/asset_amount_input.tsx
```diff
@@ -7,26 +7,51 @@ import { ERC20Asset } from '../types';
...
+        this.state = {
+            currentFontSizePx: props.startingFontSizePx,
+        };
+    }
     public render(): React.ReactNode {
         const { asset, onChange, ...rest } = this.props;
         return (
             <Container>
-                <AmountInput {...rest} onChange={this._handleChange} />
+                <Container borderBottom="1px solid rgba(255,255,255,0.3)" display="inline-block">
```
steveklebanoff: Should we be using a `ColorOption` instead of hard-coding `rgba(255,255,255,0.3)` here?

packages/instant/src/components/ui/container.tsx
```diff
@@ -29,11 +28,7 @@ export interface ContainerProps {
     zIndex?: number;
 }
 
-const PlainContainer: React.StatelessComponent<ContainerProps> = ({ children, className }) => (
-    <div className={className}>{children}</div>
-);
-
-export const Container = styled(PlainContainer)`
+export const Container = styled<ContainerProps, 'div'>('div')`
```
steveklebanoff: I prefer the syntax:`export const Container = styled.div <ContainerProps>` , although I'm curious if they are equivalent?

packages/instant/src/style/fonts.ts
```diff
@@ -1,10 +1,10 @@
-import { injectGlobal } from './theme';
-
 export const fonts = {
     include: () => {
         // Inject the inter-ui font into the page
-        return injectGlobal`
-            @import url('https://rsms.me/inter/inter-ui.css');
-        `;
+        const head = document.head || document.getElementsByTagName('head')[0];
```
steveklebanoff: do you think it's possible that an integrator's page may not have a `head` tag?  i guess its unlikely

steveklebanoff:
A few minor comments, overall looks good.  CI failing and has a conflict though

packages/instant/src/components/ui/container.tsx
fragosti: Does that one work? It seemed to be producing an `any` type?

steveklebanoff: Looks like it's working on my side: 

image

fragosti: @steveklebanoff are the interpolated `props` also typed?

packages/instant/src/components/scaling_input.tsx
fragosti: Adding this to `private readonly _inputRef = React.createRef<HTMLInputElement>();` helped with the other `any` in `_getInputWidthInPx` so thanks!

packages/instant/src/components/scaling_input.tsx
fragosti: In 16 the API has changed significantly. Overall, it's been reduced and it's harder to shoot yourself in the foot.

packages/instant/src/components/scaling_amount_input.tsx
fragosti: I usually like to `extend` but the change in the `onChange` signature makes this difficult. Feel like using `pick` doesn't add a ton of value.

packages/instant/src/components/asset_amount_input.tsx
fragosti: I've resigned to just use a string in this case (and a few others that are opacity-based) because "white with opacity 0.3" is not really a color.

fragosti: Maybe we can add a constant somewhere?

packages/instant/src/components/ui/container.tsx
fragosti: Yea when I use this I find that it cannot infer the type of `props`

packages/instant/src/style/fonts.ts
fragosti: This is a good point.

packages/instant/src/components/scaling_input.tsx
```diff
@@ -0,0 +1,170 @@
...
+    ): void {
+        const prevPhase = ScalingInput.getPhaseFromProps(prevProps);
+        const curPhase = ScalingInput.getPhaseFromProps(this.props);
+        // if we went from anything else to end, fix to the current width as it shouldn't change as we grow
+        if (prevPhase !== ScalingInputPhase.ScalingFontSize && curPhase === ScalingInputPhase.ScalingFontSize) {
+            this.setState({
+                inputWidthPxAtPhaseChange: snapshot.inputWidthPx,
+            });
+        }
+        // if we end from end to to anything else, un-fix the width
```
steveklebanoff: comment needs rewording, also potentially on line 104 too

packages/instant/src/components/asset_amount_input.tsx
steveklebanoff: hypothetically we could do something with this https://github.com/misund/hex-to-rgba#readme but def seems like overkill.

I like the idea of all of our colors being in one file, or at least easily findable.  I'm open to constants, or maybe even a `whiteWithOpacity` function in theme.ts, or something like that: open to diff implementations but overall would prefer to have the colors centralized, and not just one-off defined components

packages/instant/src/util/big_number.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as _ from 'lodash';
+
+/**
+ *  A BigNumber extension that is more flexible about decimal strings.
+ *  Such as allowing:
+ *  new BigNumberInput(0.) => 0
+ *  new BigNumberInput(1.) => 1
+ *  new BigNumberInput(1..) => still throws
+ */
+export class BigNumberInput extends BigNumber {
```
steveklebanoff: I feel like the class name should match the file name

packages/instant/src/util/big_number.ts
```diff
@@ -0,0 +1,29 @@
...
+    private readonly _hasDecimalPeriod: boolean;
+    constructor(bigNumberString: string) {
+        const hasDecimalPeriod = _.endsWith(bigNumberString, '.');
```
steveklebanoff: minor but would prefer this var to be more descriptive, like `isEndingWithDecimal`

packages/instant/src/util/big_number.ts
```diff
@@ -0,0 +1,29 @@
...
+        let internalString = bigNumberString;
+        if (hasDecimalPeriod) {
+            internalString = bigNumberString.slice(0, bigNumberString.length - 1);
```
steveklebanoff: minor but I feel like `str.slice(0, -1);` is more immediately understandable

packages/instant/src/util/big_number.ts
steveklebanoff: extending BigNumber makes me a little nervous, but after thinking about it I understand why you went with this approach, and I can't immediately think of a better one

packages/instant/src/util/asset.ts
```diff
@@ -43,6 +43,16 @@ export const assetUtils = {
...
     },
+    formattedSymbolForAsset: (asset?: ERC20Asset, defaultName: string = '???'): string => {
+        if (_.isUndefined(asset)) {
+            return defaultName;
+        }
+        const symbol = asset.metaData.symbol;
+        if (symbol.length <= 5) {
+            return symbol;
+        }
+        return `${symbol.slice(0, 3)}...`;
```
steveklebanoff: Should we use `&hellip;`?

packages/instant/src/components/ui/container.tsx
```diff
@@ -27,14 +26,14 @@ export interface ContainerProps {
...
+    whiteSpace?: string;
     opacity?: number;
 }
 
-const PlainContainer: React.StatelessComponent<ContainerProps> = ({ children, className }) => (
-    <div className={className}>{children}</div>
-);
-
-export const Container = styled(PlainContainer)`
+export const Container =
```
steveklebanoff: :neckbeard: 👍

packages/instant/src/components/ui/text.tsx
```diff
@@ -23,14 +23,11 @@ export interface TextProps {
...
 
-const PlainText: React.StatelessComponent<TextProps> = ({ children, className, onClick }) => (
-    <div className={className} onClick={onClick}>
-        {children}
-    </div>
-);
-
 const darkenOnHoverAmount = 0.3;
-export const Text = styled(PlainText)`
+export const Text =
```
steveklebanoff: thanks for changing these!  feels cleaner, but bummer that prettier makes it look funky

steveklebanoff:
Overall looks very good, just some minor comments

packages/instant/src/components/ui/container.tsx
BMillman19: Why is this different from how `SlidingAnimation` is defined?

packages/instant/src/components/ui/container.tsx
```diff
@@ -27,14 +26,14 @@ export interface ContainerProps {
...
+    styled.div <
```
BMillman19: I wonder if we can tweak the prettier rules to not do this...

packages/instant/src/util/big_number.ts
```diff
@@ -0,0 +1,29 @@
+import { BigNumber } from '@0x/utils';
```
BMillman19: Is there any way to achieve this using BigNumber's API rather than extending it? I'm a bit wary of us ending up in a situation where we don't know if something is a BigNumber / BigNumberInput, we get unexpected bugs, etc

packages/instant/src/components/asset_amount_input.tsx
fragosti: @BMillman19 you can't really go from rgba to HEX, unless I'm misunderstanding what you mean.

fragosti: I'm just going to make this a constant in `theme.ts`

To get access to the actual injected theme you need to read from context, define a styled-component or use a HOC.

packages/instant/src/util/big_number.ts
fragosti: This is pretty safe IMHO

packages/instant/src/util/asset.ts
fragosti: React will escape that.

packages/instant/src/components/ui/container.tsx
fragosti: because whoops

packages/instant/src/util/big_number.ts
fragosti: There isn't... We are only overriding `toString`... Maybe if a dependency depends on `toString` never returning an invalid number we could be in trouble.

How bout this. I'm going to change all references to `selectedAssetAmount` to be `BigNumberInput` and not override `toString` but add another method that returns the displayable string.

PR #1177, opened 2018-10-23
fix(web3-wrapper): Make getTransactionByHashAsync return the correct type

albrow:
## Description

This PR fixes a bug. Because we were not calling `unmarshalTransaction`, the return value of `getTransactionByHashAsync` did not line up with its type definition. The fields were not getting parsed (e.g. `gasPrice` was a `string` instead of a `BigNumber`).

## Types of changes

## Checklist:

### DIFF
--- a/packages/web3-wrapper/src/web3_wrapper.ts
+++ b/packages/web3-wrapper/src/web3_wrapper.ts
@@ -23,7 +23,13 @@ import {
 import * as _ from 'lodash';
 
 import { marshaller } from './marshaller';
-import { BlockWithoutTransactionDataRPC, BlockWithTransactionDataRPC, NodeType, Web3WrapperErrors } from './types';
+import {
+    BlockWithoutTransactionDataRPC,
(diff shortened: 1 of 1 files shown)


packages/web3-wrapper/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "3.1.1",
```
BMillman19: should this be a major version bump, I can go either way

albrow: In this case, I think we presented users with a pretty clear description of the intended behavior in the type definition and the documentation. The actual behavior did not match that intended behavior. I would classify this as a bug fix.

However I admit the decision is not always clear-cut. There's a [decent discussion on StackOverflow](https://softwareengineering.stackexchange.com/questions/368030/is-a-breaking-change-bug-fix-a-semver-patch-or-a-major) about it.

PR #1179, opened 2018-10-24
[instant] Success and Processing state

steveklebanoff:
## Description

- Shows "Placing order" button when waiting for metamask txn to be confirmed
- Styles success screen and gives a "View Transaction" button on success
- Introduces new "OrderState" type per convo on slack which supports holding the txnhash

tokenbuygif

There is a separate ticket which represents being able to view the txn while processing

## Checklist:


package.json
```diff
@@ -50,7 +50,7 @@
             },
             {
                 "path": "packages/instant/public/main.bundle.js",
-                "maxSize": "500kB"
+                "maxSize": "800kB"
```
steveklebanoff: Will get rid of react-shared and put this back down

PR #1184, opened 2018-10-25
[instant] View Processing Transaction

steveklebanoff:
## Description

When transaction is processing:
- Be able to view transaction on etherscan
- Update top text and show spinner icon

viewtxngif

## Checklist:

### DIFF
--- a/packages/instant/src/components/buy_button.tsx
+++ b/packages/instant/src/components/buy_button.tsx
@@ -13,9 +13,9 @@ export interface BuyButtonProps {
     assetBuyer?: AssetBuyer;
     onAwaitingSignature: (buyQuote: BuyQuote) => void;
     onSignatureDenied: (buyQuote: BuyQuote, preventedError: Error) => void;
-    onBuyProcessing: (buyQuote: BuyQuote, txnHash: string) => void;
-    onBuySuccess: (buyQuote: BuyQuote, txnHash: string) => void;
-    onBuyFailure: (buyQuote: BuyQuote, txnHash?: string) => void; // TODO: make this non-optional
+    onBuyProcessing: (buyQuote: BuyQuote, txHash: string) => void;
+    onBuySuccess: (buyQuote: BuyQuote, txHash: string) => void;
+    onBuyFailure: (buyQuote: BuyQuote, txHash?: string) => void; // TODO: make this non-optional
 }
 
 export class BuyButton extends React.Component<BuyButtonProps> {
@@ -41,10 +41,10 @@ export class BuyButton extends React.Component<BuyButtonProps> {
             return;
         }
 
-        let txnHash: string | undefined;
+        let txHash: string | undefined;
         this.props.onAwaitingSignature(buyQuote);
         try {
-            txnHash = await assetBuyer.executeBuyQuoteAsync(buyQuote);
+            txHash = await assetBuyer.executeBuyQuoteAsync(buyQuote);
         } catch (e) {
             if (e instanceof Error && e.message === AssetBuyerError.SignatureRequestDenied) {
                 this.props.onSignatureDenied(buyQuote, e);
@@ -54,20 +54,20 @@ export class BuyButton extends React.Component<BuyButtonProps> {
         }
 
         // Have to let TS know that txHash is definitely defined now
-        if (!txnHash) {
+        if (!txHash) {
             throw new Error('No txHash available');
         }
 
-        this.props.onBuyProcessing(buyQuote, txnHash);
+        this.props.onBuyProcessing(buyQuote, txHash);
         try {
-            await web3Wrapper.awaitTransactionSuccessAsync(txnHash);
+            await web3Wrapper.awaitTransactionSuccessAsync(txHash);
         } catch (e) {
             if (e instanceof Error && e.message.startsWith('Transaction failed')) {
-                this.props.onBuyFailure(buyQuote, txnHash);
+                this.props.onBuyFailure(buyQuote, txHash);
                 return;
             }
             throw e;
         }
-        this.props.onBuySuccess(buyQuote, txnHash);
+        this.props.onBuySuccess(buyQuote, txHash);
     };
 }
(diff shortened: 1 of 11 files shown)


packages/instant/src/components/buy_button.tsx
```diff
@@ -11,7 +11,8 @@ import { Button, Text } from './ui';
 export interface BuyButtonProps {
     buyQuote?: BuyQuote;
     assetBuyer?: AssetBuyer;
-    onClick: (buyQuote: BuyQuote) => void;
+    onAwaitingSignature: (buyQuote: BuyQuote) => void;
```
BMillman19: What do you think of this naming for the callbacks?

```
    buyQuote?: BuyQuote;
    assetBuyer?: AssetBuyer;
    onAwaitingSignature: (buyQuote: BuyQuote) => void;
    onSignatureDenied: (buyQuote: BuyQuote) => void;
    onBuyProcessing: (buyQuote: BuyQuote, txnHash: string) => void;
    onBuySuccess: (buyQuote: BuyQuote, txnHash: string) => void;
    onBuyFailure: (buyQuote: BuyQuote, tnxHash: string) => void;
}
```

packages/instant/src/components/buy_button.tsx
```diff
@@ -35,21 +36,23 @@ export class BuyButton extends React.Component<BuyButtonProps> {
...
             return;
         }
-        this.props.onClick(this.props.buyQuote);
+        this.props.onAwaitingSignature(buyQuote);
         let txnHash;
         try {
-            txnHash = await this.props.assetBuyer.executeBuyQuoteAsync(this.props.buyQuote);
-            const txnReceipt = await web3Wrapper.awaitTransactionSuccessAsync(txnHash);
-            this.props.onBuySuccess(this.props.buyQuote, txnReceipt.transactionHash);
+            txnHash = await assetBuyer.executeBuyQuoteAsync(buyQuote);
```
BMillman19: If we split this up into multiple try/catches i believe that will give us a bit more control over handling the errors, it will also let us get rid of the optional txHash in the `onBuyFailure` callback:

```
    private readonly _handleClick = async () => {
        // The button is disabled when there is no buy quote anyway.
        const { buyQuote, assetBuyer } = this.props;
        if (_.isUndefined(buyQuote) || _.isUndefined(assetBuyer)) {
            return;
        }
        let txnHash;
        this.props.onAwaitingSignature(buyQuote);
        try {
            txnHash = await assetBuyer.executeBuyQuoteAsync(buyQuote);
        } catch (e) {
            if (e instanceof Error && e.message === AssetBuyerError.SignatureRequestDenied) {
                this.props.onSignatureDenied(buyQuote);
                return;
            }
            throw e;
        }
        this.props.onBuyProcessing(buyQuote, txnHash);
        try {
            await web3Wrapper.awaitTransactionSuccessAsync(txnHash);
        } catch (e) {
            if (e instanceof Error && e.message.startsWith('Transaction failed')) {
                this.props.onBuyFailure(buyQuote, txnHash);
                return;
            }
            throw e;
        }
        this.props.onBuySuccess(buyQuote, txnHash);
    };
```

BMillman19: This implementation also re-throws errors that we don't expect

packages/instant/src/types.ts
```diff
@@ -8,18 +8,30 @@ export enum AsyncProcessState {
...
+    txnHash: string;
 }
 interface SuccessfulOrderState {
-    processState: AsyncProcessState.SUCCESS;
+    processState: OrderProcessState.SUCCESS;
     txnHash: string;
 }
 interface FailureOrderState {
-    processState: AsyncProcessState.FAILURE;
+    processState: OrderProcessState.FAILURE;
```
BMillman19: we can get rid of this optional property here and merge Successful and Failure order states?

Also lets use `txHash` instead of `txnHash` since its in line with how we abbreviate "transaction hash" in other parts of the code base (e.g. `Web3Wrapper`)

packages/instant/src/components/buy_button.tsx
steveklebanoff: Look good! will change

packages/instant/src/components/buy_button.tsx
steveklebanoff: I dig it -- it makes the ErrorBoundary more important (will prioritize as separate feature), but helps make our code more explicit

packages/instant/src/components/buy_button.tsx
```diff
@@ -35,21 +36,38 @@ export class BuyButton extends React.Component<BuyButtonProps> {
...
             if (e instanceof Error && e.message === AssetBuyerError.SignatureRequestDenied) {
-                this.props.onBuyPrevented(this.props.buyQuote, e);
+                this.props.onSignatureDenied(buyQuote, e);
+            } else {
+                throw e;
+            }
+        }
+
+        // Have to let TS know that txHash is definitely defined now
+        if (!txHash) {
```
BMillman19: hm i think if we leave out the `string | undefined` on line 44 the type inference will work w/o this

packages/instant/src/components/buy_button.tsx
```diff
@@ -35,21 +36,38 @@ export class BuyButton extends React.Component<BuyButtonProps> {
...
+            throw new Error('No txHash available');
+        }
+
+        this.props.onBuyProcessing(buyQuote, txHash);
+        try {
+            await web3Wrapper.awaitTransactionSuccessAsync(txHash);
+        } catch (e) {
+            if (e instanceof Error && e.message.startsWith('Transaction failed')) {
```
BMillman19: maybe we can put the "Transaction failed" string in the constants, `WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG` or something like that

packages/instant/src/components/buy_button.tsx
steveklebanoff: this doesn't work unfortunately, gives the error "Variable 'txHash' is used before being assigned"

PR #1185, opened 2018-10-25
[order_utils.py] tweaks for first publish

feuGeneA:
## Description

The first version of 0x Python Order Utilities has been published! :tada:

[https://pypi.org/project/0x-order-utils/](https://pypi.org/project/0x-order-utils/)

This PR is for the changes that were needed to facilitate this publication.

## Testing instructions
Explained in README.md changes.

## Types of changes

## Checklist:

### DIFF
--- a/python-packages/order_utils/README.md
+++ b/python-packages/order_utils/README.md
@@ -1,0 +1,47 @@
+API documentation hosted at [https://0x.readthedocs.io/projects/order-utils/en/latest/](https://0x.readthedocs.io/projects/order-utils/en/latest/)
+
+## Development
+
+### Edit/test cycle
+
+`pip install -e .[dev]`
+
+`./setup.py build`
+
+`./setup.py build_sphinx`
+
+`./setup.py test`
+
+`./setup.py lint`
+
+`./setup.py clean --all`
+
+See `./setup.py --help-commands` for more info.
+
+### Publishing
+
+`./setup.py sdist bdist_wheel` will build the distribution into the `dist` folder. Once you're happy with it, you can use `./setup.py` to upload it to the package index, as described below.
+
+If you're just experimenting, and want to try out a publish without committing to it, consider using a "staging package;" that is, rename the package, in `setup(name=...)` in `setup.py`, to something like "0x-order-utils-staging". The reason is that PyPI will never let you re-publish to an already-used file name (which includes the package name and version), even if you delete that published version from the repo. With an alternate package name, you can increase the version number (required for a successful upload) for every little experiment, without "stealing" version numbers from the main package name.
+
+#### Uploading to test.pypi.org
+
+Create an account at test.pypi.org (this is independent of any account at pypi.org).
+
+With an `sdist` folder populated as described above, do `./setup.py test_publish`.
+
+To run tests against the package deployed to test.pypi.org, run `tox -erun_tests_against_test_deployment`. If you need to run against a staging package, modify `tox.ini` to change the package name.
+
+#### Uploading to (the real) pypi.org
+
+Create an account at pypi.org (this is independent of any account at test.pypi.org).
+
+With an `sdist` folder populated as described above, do `./setup.py publish`.
+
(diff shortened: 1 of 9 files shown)


python-packages/order_utils/README.md
```diff
@@ -0,0 +1,47 @@
+API documentation hosted at [https://0x.readthedocs.io/projects/order-utils/en/latest/](https://0x.readthedocs.io/projects/order-utils/en/latest/)
```
fabioberger:
```suggestion
## order_utils

0x order-related utilities for those developing on top of 0x protocol.

### Read the [Documentation](https://0x.readthedocs.io/projects/order-utils/en/latest/](https://0x.readthedocs.io/projects/order-utils/en/latest/).
```

fabioberger: Standardize README header.

python-packages/order_utils/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+`./setup.py build_sphinx`
+
+`./setup.py test`
+
+`./setup.py lint`
+
+`./setup.py clean --all`
+
+See `./setup.py --help-commands` for more info.
```
fabioberger: Can we also copy the TS README structure for these? Build, clean, lint are separate sections (https://github.com/0xProject/0x-monorepo/tree/development/packages/order-utils#build)

python-packages/order_utils/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Publishing
+
+`./setup.py sdist bdist_wheel` will build the distribution into the `dist` folder. Once you're happy with it, you can use `./setup.py` to upload it to the package index, as described below.
+
+If you're just experimenting, and want to try out a publish without committing to it, consider using a "staging package;" that is, rename the package, in `setup(name=...)` in `setup.py`, to something like "0x-order-utils-staging". The reason is that PyPI will never let you re-publish to an already-used file name (which includes the package name and version), even if you delete that published version from the repo. With an alternate package name, you can increase the version number (required for a successful upload) for every little experiment, without "stealing" version numbers from the main package name.
```
fabioberger: Nit: Intentional `staging package;"`?

fabioberger: Hm... Since we don't expect end-developers to publish this library, let's move all sections here related to publishing to a Quip doc in the `documentation` folder about how publish the Python package.

python-packages/order_utils/setup.py
```diff
@@ -73,31 +75,64 @@ class CleanCommandExtension(clean):
...
+
+with open("README.md", "r") as file_handle:
+    README_MD = file_handle.read()
 
 
 setup(
-    name="order_utils",
-    version="1.0.0",
+    name="0x-order-utils",
+    version="0.1.0",
```
fabioberger: Why did we have to change to version `0.1.0`? I'd rather we start at v1 due to SemVer rules about pre-1.0 releases.

feuGeneA: We didn't HAVE to, I just did it because it felt right. I was thinking that we would bump to 1.0 upon completion of what we set out as the MVP in our spec. I'm no expert in sever, but I just google it and it seems like my idea is consistent with the rules in its spec: "Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable."  Having said all that, I don't have strong feelings about it, so I'm happy to do whatever you think is right.

python-packages/order_utils/README.md
feuGeneA: Yes that was intentional. Looks weird now that you mention it, but it's grammatically correct 🙂.

python-packages/order_utils/setup.py
fabioberger: Ah, I see. Ok, that's fine then.

python-packages/order_utils/README.md
```diff
@@ -0,0 +1,45 @@
+## 0x-order-utils
+
+0x order-related utilities for those developing on top of 0x protocol.
+
+Read the [documentation](https://0x.readthedocs.io/projects/order-utils/en/latest/)
+
+## Installing
+
+```bash
+pip install 0x-order-utils`
```
fabioberger: Extra `

PR #1187, opened 2018-10-25
[instant] Add ability to toggle render settings through URL, flash error on incorrect network, provided liquidity

fragosti:
## Description

- Made a small change to asset-buyer provided order constructor.
- Allow configuring instant from URL query params.
- Flash error when wrong network is detected. 
- Support defaultAssetBuyAmountSetting.
- Refactor how error messages are flashed

## Testing instructions

http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com/?networkId=42&assetData=0xf47261b00000000000000000000000002002d3812f58e35f0ea1ffbf80a75a38c32175fa&liquiditySource=provided

## Checklist:


packages/instant/src/components/zero_ex_instant.tsx
```diff
@@ -21,28 +25,34 @@ export type ZeroExInstantProps = ZeroExInstantRequiredProps & Partial<ZeroExInst
...
 export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
     private readonly _store: Store;
     private static _mergeInitialStateWithProps(props: ZeroExInstantProps, state: State = INITIAL_STATE): State {
-        // Create merged object such that properties in props override default settings
-        const optionalPropsWithDefaults: ZeroExInstantOptionalProps = {
-            additionalAssetMetaDataMap: props.additionalAssetMetaDataMap || {},
-            network: props.network || state.network,
-        };
-        const { network } = optionalPropsWithDefaults;
+        const network = props.network || state.network;
```
BMillman19: can we rename `network` to `networkId` in the props definition

BMillman19: on another note, we probably want to add asserstions for the props since this wil be a public facing API, perhaps in the `render` method in `index.umd.ts` ?

packages/instant/src/components/zero_ex_instant.tsx
```diff
@@ -52,15 +62,24 @@ export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
...
         return storeStateFromProps;
     }
     constructor(props: ZeroExInstantProps) {
         super(props);
-        this._store = store.create(ZeroExInstant._mergeInitialStateWithProps(this.props, INITIAL_STATE));
+        const initialAppState = ZeroExInstant._mergeInitialStateWithProps(this.props, INITIAL_STATE);
+        this._store = store.create(initialAppState);
+    }
+
+    public componentDidMount(): void {
```
BMillman19: do we want to do this for will / did change props lifecycle method as well?

fragosti: I think our current position is that we do not support these settings changing after an initial render (had this discussion in https://github.com/0xProject/0x-monorepo/pull/1159). However, I do see how it could be useful. Would rather it be done in a follow-up branch however.

packages/instant/src/util/error.ts
```diff
@@ -60,6 +60,9 @@ export const errorUtil = {
         if (error instanceof Error) {
             bestMessage = humanReadableMessageForError(error, asset);
         }
+        if (_.isString(error)) {
```
steveklebanoff: I personally like the pattern of raising an error with a constant, and then turning into a human readable version here. This way we have one place to look at all the human readable errors we produce.  See https://github.com/0xProject/0x-monorepo/pull/1193/files for an example of this.

packages/instant/src/containers/selected_asset_buy_button.ts
```diff
@@ -7,6 +7,7 @@ import { Dispatch } from 'redux';
 import { Action, actions } from '../redux/actions';
```
steveklebanoff: In development, the buy button logic is now in `selected_asset_buy_order_state_buttons`

BMillman19: Lol git figured it out

steveklebanoff: Nice :)

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
```diff
@@ -79,11 +80,24 @@ const updateBuyQuoteAsync = async (
         newBuyQuote = await assetBuyer.getBuyQuoteAsync(asset.assetData, baseUnitValue);
     } catch (error) {
         dispatch(actions.setQuoteRequestStateFailure());
-        errorUtil.errorFlasher.flashNewError(dispatch, error);
+        let errorMessage;
+        if (error.message === AssetBuyerError.InsufficientAssetLiquidity) {
```
steveklebanoff: I think ideally this would still be in a separate util method and we would keep the tests around

PR #1188, opened 2018-10-26
[instant] Prevent older quote requests from overriding newer requests

BMillman19:
## Description

* Every time we dispatch an `UPDATE_LATEST_BUY_QUOTE` action, include the time that we made the request
* The reducer will now check if the new quote was requested more recently than the latest quote before overriding it
* Also removed the debounce from `selected_asset_amount_input.ts` because we are already cacheing at the `AssetBuyer` level, so we are not making a request more that once every 10s (this is configurable on our end)
* Overall, this provides a better UX because the form looks more responsive

## Testing instructions

Mash the token input

## Types of changes

## Checklist:


packages/instant/src/redux/reducer.ts
```diff
@@ -121,3 +130,24 @@ export const reducer = (state: State = INITIAL_STATE, action: Action): State =>
             return state;
     }
 };
+
+const doesBuyQuoteMatchState = (buyQuote: BuyQuote, state: State): boolean => {
```
fragosti: maybe put this in a util file? don't feel strongly about it.

packages/instant/src/redux/reducer.ts
```diff
@@ -121,3 +130,24 @@ export const reducer = (state: State = INITIAL_STATE, action: Action): State =>
...
+    // if buyQuote's assetData does not match that of the current selected asset, return false
+    if (selectedAssetIfExists.assetData !== buyQuote.assetData) {
+        return false;
+    }
+    // if buyQuote's assetBuyAmount does not match selectedAssetAmount, return false
+    const selectedAssetMetaData = selectedAssetIfExists.metaData;
+    const selectedAssetAmountBaseUnits =
+        selectedAssetMetaData.assetProxyId === AssetProxyId.ERC20
+            ? Web3Wrapper.toBaseUnitAmount(selectedAssetAmountIfExists, selectedAssetMetaData.decimals)
+            : new BigNumber(1);
```
fragosti: Does ERC721 have the concept of base units? I guess the `BuyQuote` would just have the correct amount? This doesn't seem to cover the case where the user types "0" but we're not planning to show an input and automatically put selected amount to 1 I guess.

BMillman19: I'm going to amend this to be more flexible, right now we're making too many assumptions about the nature of future non erc20 assets

packages/instant/src/redux/reducer.ts
BMillman19: will just leave it

PR #1189, opened 2018-10-26
[instant] View txn on failure

steveklebanoff:
## Description

Allows users to view transaction details when transaction fails.

Also, converts to the SelectedBuyOrderStateButton to be one big connected component. This refactor made things much easier with less boilerplate and indirection to implement this feature.

image

## Checklist:


packages/instant/src/containers/selected_asset_buy_order_state_button.tsx
```diff
@@ -1,20 +1,78 @@
+import { AssetBuyer, BuyQuote } from '@0x/asset-buyer';
```
BMillman19: I think this should be a `.ts` file because we don't have any tsx in the file, this is the same for any other connected containers that we might write

steveklebanoff: Good point, fixed in 68182fb6c41256f9886c7aa0269746b3a9ac632f

packages/instant/src/components/buy_order_state_button.tsx
```diff
@@ -1,26 +1,63 @@
...
+    onViewTransaction: () => void;
+    onAwaitingSignature: (buyQuote: BuyQuote) => void;
+    onSignatureDenied: (buyQuote: BuyQuote, error: Error) => void;
+    onBuyProcessing: (buyQuote: BuyQuote, txHash: string) => void;
+    onBuySuccess: (buyQuote: BuyQuote, txHash: string) => void;
+    onBuyFailure: (buyQuote: BuyQuote, txHash: string) => void;
+    onRetry: () => void;
 }
 
+// TODO: rename to buttons
```
fragosti: You still want to do this?

steveklebanoff: Yes, thanks! will do

PR #1193, opened 2018-10-26
[instant] Not enough ETH error message

steveklebanoff:
## Description

Shows error message when there is not enough ETH in a wallet to perform txn (doesnt include gas calculation)

Also renames ethDecimals to ETH_DECIMALS

## Checklist:


packages/instant/src/containers/selected_erc20_asset_amount_input.ts
```diff
@@ -10,12 +10,15 @@ import { Dispatch } from 'redux';
 import { Action, actions } from '../redux/actions';
 import { State } from '../redux/reducer';
 import { ColorOption } from '../style/theme';
-import { ERC20Asset, OrderProcessState } from '../types';
+import { getBestAddress } from '../util/address';
 import { BigNumberInput } from '../util/big_number_input';
 import { errorUtil } from '../util/error';
+import { web3Wrapper } from '../util/web3_wrapper';
 
 import { ERC20AssetAmountInput } from '../components/erc20_asset_amount_input';
```
BMillman19: I believe this and the import from `../types` should go in the grouping above

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
```diff
@@ -77,6 +80,14 @@ const updateBuyQuoteAsync = async (
     errorUtil.errorFlasher.clearError(dispatch);
     // invalidate the last buy quote.
     dispatch(actions.updateLatestBuyQuote(newBuyQuote));
+
+    // set error if user doesn't have appropriate balance
```
BMillman19: we should also either disable the buy button in this case or flash the same message when the user tries to click buy

packages/instant/src/util/address.ts
```diff
@@ -0,0 +1,6 @@
+import { web3Wrapper } from '../util/web3_wrapper';
+
+export const getBestAddress = async (): Promise<string> => {
+    const addresses = await web3Wrapper.getAvailableAddressesAsync();
+    return addresses[0];
```
BMillman19: what happens if `addresses` is empty

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
steveklebanoff: good point -- i think it'd be easier just to flash again if they click buy

packages/instant/src/util/address.ts
steveklebanoff: is this a possibility?  if so, I could have this thrown and flash an error

packages/instant/src/components/buy_button.tsx
```diff
@@ -17,6 +19,7 @@ export interface BuyButtonProps {
     onBuyProcessing: (buyQuote: BuyQuote, txHash: string) => void;
     onBuySuccess: (buyQuote: BuyQuote, txHash: string) => void;
     onBuyFailure: (buyQuote: BuyQuote, txHash: string) => void;
+    validateWalletBeforeBuy: (buyQuote: BuyQuote, takerAddress: string | undefined) => Promise<boolean>;
```
BMillman19: we usually just write `takerAddress?: string` for optional params we want to add an "async" suffix to the prop name to signal that it is awaitable, i.e. `validateAddressBeforeBuyAsync`

packages/instant/src/util/address.ts
steveklebanoff: This should be handled by `hasSufficientFunds` returning `false`

steveklebanoff:
@BMillman19 This is now updated.  It was a lot simpler to just validate this when clicking the buy button, we can add synchronous checking in the future if we'd like.

packages/instant/src/components/buy_button.tsx
```diff
@@ -42,10 +46,18 @@ export class BuyButton extends React.Component<BuyButtonProps> {
...
+            this.props.onValidationFail(buyQuote, ZeroExInstantError.InsufficientBalance);
+            return;
+        }
+
         let txHash: string | undefined;
-        this.props.onAwaitingSignature(buyQuote);
         try {
-            txHash = await assetBuyer.executeBuyQuoteAsync(buyQuote);
+            txHash = await assetBuyer.executeBuyQuoteAsync(buyQuote, { takerAddress });
         } catch (e) {
```
steveklebanoff: @BMillman19 I think you mentioned there could be validation errors from `executeBuyQuoteAsync`, is that correct?  If so -- we should we identify them and send them to onValidationFail?

packages/instant/src/containers/selected_asset_buy_order_state_buttons.ts
```diff
@@ -68,7 +70,11 @@ const mapDispatchToProps = (
         dispatch(actions.updateBuyOrderState({ processState: OrderProcessState.FAILURE, txHash })),
     onSignatureDenied: (buyQuote, error) => {
         dispatch(actions.resetAmount());
-        dispatch(actions.setError(error));
+        errorUtil.errorFlasher.flashNewError(dispatch, error);
```
steveklebanoff: This will have conflicts with https://github.com/0xProject/0x-monorepo/pull/1187-- will have to update whichever branch gets merged second

packages/instant/src/components/buy_button.tsx
```diff
@@ -12,7 +15,8 @@ import { Button, Text } from './ui';
 export interface BuyButtonProps {
     buyQuote?: BuyQuote;
     assetBuyer?: AssetBuyer;
-    onAwaitingSignature: (buyQuote: BuyQuote) => void;
+    onPendingValidation: (buyQuote: BuyQuote) => void;
```
BMillman19: can we reverse this to `onValidationPending` so that it is consistent with the other prop names?

packages/instant/src/util/balance.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Web3Wrapper } from '@0x/web3-wrapper';
+import * as _ from 'lodash';
+
+export const balanceUtil = {
+    hasSufficientFunds: async (takerAddress: string | undefined, buyQuote: BuyQuote, web3Wrapper: Web3Wrapper) => {
+        if (_.isUndefined(takerAddress)) {
+            return false;
+        }
+        const balanceWei = await web3Wrapper.getBalanceInWeiAsync(takerAddress);
+        return balanceWei >= buyQuote.worstCaseQuoteInfo.totalEthAmount;
```
BMillman19: we might want to keep this generalized by passing in a BigNumber representing the "sufficient" balance. Also the name of the method could be improved since fund can refer to eth or tokens. What about `hasSufficientEth`?

packages/instant/src/components/buy_button.tsx
BMillman19: yes, it is `AssetBuyerError.TransactionValueTooLow`

BMillman19: If this pops up it can mean quote is stale so we what a slightly different behavior

packages/instant/src/util/address.ts
BMillman19: Ok, I was playing around and a locked metamask will produce an empty array here so we should handle it by flashing an error

steveklebanoff: Ah cool, didn't know this is what would happen when metamask was is locked. It will show "Not enough ETH" for now, I wonder if we should make a better message in a future PR?

packages/instant/src/components/buy_button.tsx
steveklebanoff: OK, will treat this as a validation fail for now

packages/instant/src/util/balance.ts
steveklebanoff: I like the name change.  I'd prefer to keep the logic as-is, as this helper function would hardly do anything if you sent in the balance.  I'd like this function to be helpful if we validate balance on typing (instead of just when clicking 'Buy') in the future

packages/instant/src/util/address.ts
BMillman19: yea let me make a ticket for that

PR #1198, opened 2018-10-29
[instant] Gas price estimate

steveklebanoff:
## Description

Hit eth gas station to get best gas price

## Checklist:

### DIFF
--- a/packages/instant/src/components/buy_button.tsx
+++ b/packages/instant/src/components/buy_button.tsx
@@ -7,6 +7,7 @@ import { ColorOption } from '../style/theme';
 import { ZeroExInstantError } from '../types';
 import { getBestAddress } from '../util/address';
 import { balanceUtil } from '../util/balance';
+import { gasPriceEstimator } from '../util/gas_price_estimator';
 import { util } from '../util/util';
 import { web3Wrapper } from '../util/web3_wrapper';
 
@@ -56,9 +57,9 @@ export class BuyButton extends React.Component<BuyButtonProps> {
         }
 
         let txHash: string | undefined;
+        const gasPrice = await gasPriceEstimator.getFastAmountInWeiAsync();
         try {
-            const gasPrice = DEFAULT_GAS_PRICE;
-            txHash = await assetBuyer.executeBuyQuoteAsync(buyQuote, { gasPrice, takerAddress });
+            txHash = await assetBuyer.executeBuyQuoteAsync(buyQuote, { takerAddress, gasPrice });
         } catch (e) {
             if (e instanceof Error) {
                 if (e.message === AssetBuyerError.SignatureRequestDenied) {
--- a/packages/instant/src/components/buy_button.tsx
+++ b/packages/instant/src/components/buy_button.tsx
@@ -2,7 +2,7 @@ import { AssetBuyer, AssetBuyerError, BuyQuote } from '@0x/asset-buyer';
 import * as _ from 'lodash';
 import * as React from 'react';
 
-import { DEFAULT_GAS_PRICE, WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG_PREFIX } from '../constants';
+import { WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG_PREFIX } from '../constants';
 import { ColorOption } from '../style/theme';
 import { ZeroExInstantError } from '../types';
 import { getBestAddress } from '../util/address';
--- a/packages/instant/src/components/zero_ex_instant.tsx
+++ b/packages/instant/src/components/zero_ex_instant.tsx
@@ -13,6 +13,7 @@ import { AssetMetaData, Network } from '../types';
 import { assetUtils } from '../util/asset';
 import { BigNumberInput } from '../util/big_number_input';
 import { errorFlasher } from '../util/error_flasher';
+import { gasPriceEstimator } from '../util/gas_price_estimator';
 import { getProvider } from '../util/provider';
 import { web3Wrapper } from '../util/web3_wrapper';
 
@@ -78,6 +79,7 @@ export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
     public componentDidMount(): void {
         // tslint:disable-next-line:no-floating-promises
         asyncData.fetchAndDispatchToStore(this._store);
+        gasPriceEstimator.getFastAmountInWeiAsync();
         // tslint:disable-next-line:no-floating-promises
         this._flashErrorIfWrongNetwork();
     }
--- a/packages/instant/src/components/zero_ex_instant.tsx
+++ b/packages/instant/src/components/zero_ex_instant.tsx
@@ -79,10 +79,12 @@ export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
     public componentDidMount(): void {
         // tslint:disable-next-line:no-floating-promises
         asyncData.fetchAndDispatchToStore(this._store);
-        // tslint:disable-next-line:no-floating-promises
+
         // warm up the gas price estimator cache just in case we can't
         // grab the gas price estimate when submitting the transaction
+        // tslint:disable-next-line:no-floating-promises
         gasPriceEstimator.getFastAmountInWeiAsync();
+
         // tslint:disable-next-line:no-floating-promises
         this._flashErrorIfWrongNetwork();
     }
--- a/packages/instant/src/components/zero_ex_instant.tsx
+++ b/packages/instant/src/components/zero_ex_instant.tsx
@@ -79,6 +79,7 @@ export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
     public componentDidMount(): void {
         // tslint:disable-next-line:no-floating-promises
         asyncData.fetchAndDispatchToStore(this._store);
+        // tslint:disable-next-line:no-floating-promises
         gasPriceEstimator.getFastAmountInWeiAsync();
         // tslint:disable-next-line:no-floating-promises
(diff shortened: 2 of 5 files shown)


packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+    speed: number;
+    fastest: number;
+    safeLow: number;
+}
+
+const endpointUrl = 'https://ethgasstation.info/json/ethgasAPI.json';
+const fetchFastAmountInWei = async () => {
+    const res = await fetch(endpointUrl);
+    const gasInfo = (await res.json()) as GasStationResult;
+    const gasPriceInGwei = new BigNumber(gasInfo.fast / 10);
```
BMillman19: why are we dividing by ten here? let's leave a comment

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+    return gasPriceInGwei.mul(1000000000);
```
BMillman19: this constant exists in `constants.ts` already as `GWEI_IN_WEI`

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+    avgWait: number;
+    block_time: number;
+    speed: number;
+    fastest: number;
+    safeLow: number;
+}
+
+const endpointUrl = 'https://ethgasstation.info/json/ethgasAPI.json';
+const fetchFastAmountInWei = async () => {
+    const res = await fetch(endpointUrl);
```
BMillman19: I see that we're using `fetch` for grabbing the ETH-USD price from coinbase but by using `fetch` we're dropping support for certain browsers: https://caniuse.com/#feat=fetch. So we should either use `fetchAsync` in `@0x/utils` which uses a fetch polyfill. Or make a note and follow up, @fragosti any concerns?

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+    const gasInfo = (await res.json()) as GasStationResult;
+    const gasPriceInGwei = new BigNumber(gasInfo.fast / 10);
+    return gasPriceInGwei.mul(1000000000);
+};
+
+export class GasPriceEstimator {
+    private _lastFetched?: BigNumber;
+    public async getFastAmountInWeiAsync(): Promise<BigNumber> {
```
BMillman19: `getFastGasPriceInWeiAsync`

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+        let fetchedAmount: BigNumber | undefined;
+        try {
+            fetchedAmount = await fetchFastAmountInWei();
+        } catch {
+            fetchedAmount = undefined;
+        }
+
+        if (fetchedAmount) {
+            this._lastFetched = fetchedAmount;
```
BMillman19: can we just put this in the try?

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        return fetchedAmount || this._lastFetched || DEFAULT_GAS_PRICE;
```
BMillman19: If I'm reading this correctly, the cached value is just for cases when the latest request fails? If so we probably don't need the warmup step

packages/instant/src/components/zero_ex_instant.tsx
```diff
@@ -79,6 +80,8 @@ export class ZeroExInstant extends React.Component<ZeroExInstantProps> {
         // tslint:disable-next-line:no-floating-promises
         asyncData.fetchAndDispatchToStore(this._store);
         // tslint:disable-next-line:no-floating-promises
+        gasPriceEstimator.getFastAmountInWeiAsync();
```
fragosti: Just some thoughts  -- there is the question of whether this information should be dispatched to the store. However, since we are not displaying this value I think this is a good approach.

BMillman19:
Also the PR description should say 'eth gas station' not 'etherscan'

packages/instant/src/util/gas_price_estimator.ts
fragosti: This logic should be cleaned up or moved to a util? I'm not really sure what `fast` is denominated in.

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+const endpointUrl = 'https://ethgasstation.info/json/ethgasAPI.json';
+const fetchFastAmountInWei = async () => {
+    const res = await fetch(endpointUrl);
+    const gasInfo = (await res.json()) as GasStationResult;
+    const gasPriceInGwei = new BigNumber(gasInfo.fast / 10);
+    return gasPriceInGwei.mul(1000000000);
+};
+
+export class GasPriceEstimator {
+    private _lastFetched?: BigNumber;
```
fragosti: Should we have a way of invalidating this state? Ie. a concept of staleness?

packages/instant/src/util/gas_price_estimator.ts
fragosti: This is a good point.

packages/instant/src/util/gas_price_estimator.ts
BMillman19: also we might want to be safe and grab `fast` using `_.get` just in case there is a breaking change to the API

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: I assumed since we were using `fetch` without a polyfill for coinbase there was already an existing decision that this was OK. 
I think we either need to add a polyfill, or give an explicit deprecated warning for unsupported browsers.  I'm more in favor of the polyfill

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: I'm down to clean this up to be more understandable, but FWIW this was straight up ported from code already in website 😆 https://github.com/0xProject/0x-monorepo/blob/development/packages/website/ts/blockchain.ts#L941

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: This was a straight up port from https://github.com/0xProject/0x-monorepo/blob/development/packages/website/ts/blockchain.ts#L941 

Ideally all of this would be moved to a new package and shared between the two, but not sure if that's worth it right now

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: I excluded this from the `try` because I wanted to be explicit about things that I would expect to possibly `throw`

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: I think it's okay to be naive in this scenario.  The flow is:

```
- On startup, get best gas price
- When submitting txn, try to fetch fresh price, but if fails, fall back to last known price or resort to default
```

I don't think staleness is necessary in the scenario, because I would expect the last fetched price to be more accurate than falling back to the default. Open to feedback though

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: Yes.  I like the warmup step though, because if the fetch fails on our first time submitting a txn, we have a value to fall back to instead of having to rely on the default.

steveklebanoff: Thinking about this a bit more, I'm wondering if the warmup is worth it, esp. if this gets a lot of txn and starts hammering ethgasstation

packages/instant/src/util/gas_price_estimator.ts
steveklebanoff: If the API changes this should just raise an error and fallback to the `catch` in the function that uses this, so not sure if adding a `_.get` is necessary, I kind of like the "let it fail" approach here.
That being said, in the future might be good to report to rollbar or some sort of metric tracking every time a failure happens here, so we would know if the API changed

packages/instant/src/util/gas_price_estimator.ts
fragosti: Website code should not be used as the bar lol

packages/instant/src/components/buy_button.tsx
```diff
@@ -56,9 +57,9 @@ export class BuyButton extends React.Component<BuyButtonProps> {
         }
 
         let txHash: string | undefined;
+        const gasPrice = await gasPriceEstimator.getFastAmountInWeiAsync();
```
PirosB3: Is there any way we can set the highest (as hard-coded value) gas price that we are willing to pay?

packages/instant/src/util/gas_price_estimator.ts
```diff
@@ -0,0 +1,44 @@
...
+    blockNum: number;
+    avgWait: number;
+    block_time: number;
+    speed: number;
+    fastest: number;
+    safeLow: number;
+}
+
+const fetchFastAmountInWeiAsync = async () => {
+    const res = await fetchAsync(`${ETH_GAS_STATION_API_BASE_URL}/json/ethgasAPI.json`);
```
PirosB3: If this fails, does our transaction fail? should we use a retry mechanism?

PR #1199, opened 2018-10-30
[website] Prepare website for breaking MetaMask change for EIP 1102

BMillman19:
## Description

* Update our injected web3 handling to now handle injected provider instead, and request user address information using the new `provider.enable()` method 
* Learn about MetaMask privacy mode here: https://medium.com/metamask/https-medium-com-metamask-breaking-change-injecting-web3-7722797916a8

## Testing instructions

* Download custom build of MetaMask and install as the medium article above describes
* Enable privacy mode
* Visit http://staging-0xproject.s3-website-us-east-1.amazonaws.com/portal
* Observe MetaMask asking for approval

## Types of changes

## Checklist:


packages/website/ts/blockchain.ts
```diff
@@ -611,6 +576,45 @@ export class Blockchain {
 
         this._dispatcher.updateBlockchainIsLoaded(true);
     }
+    private async _getInjectedProviderIfExistsAsync(): Promise<InjectedProvider | undefined> {
+        if (!_.isUndefined(this._injectedProviderIfExists)) {
+            return this._injectedProviderIfExists;
+        }
+        let injectedProviderIfExists = (window as any).ethereum;
+        if (!_.isUndefined(injectedProviderIfExists)) {
+            if (!_.isUndefined(injectedProviderIfExists.enable)) {
```
fragosti: if `injectedProvider` is not undefined but it does not contain `enable`, are we intentionally setting `this._injectedProviderIfExists = injectedProviderIfExists;` at the end of this function?

packages/website/ts/blockchain.ts
```diff
@@ -611,6 +576,45 @@ export class Blockchain {
...
+        return injectedProviderIfExists;
+    }
+    private async _getInjectedProviderNetworkIdIfExistsAsync(): Promise<number | undefined> {
+        // If the user has an injectedWeb3 instance that is disconnected from a backing
+        // Ethereum node, this call will throw. We need to handle this case gracefully
+        const injectedProviderIfExists = await this._getInjectedProviderIfExistsAsync();
+        let networkIdIfExists: number;
+        if (!_.isUndefined(injectedProviderIfExists)) {
+            try {
+                const injectedWeb3Wrapper = new Web3Wrapper(injectedProviderIfExists);
```
fragosti: What about the comment from before? Is this no longer the case? 

`// We need to know the networkId the injectedWeb3 is connected to (if it is defined) in	
        // order to properly instantiate the web3Wrapper. Since we must use the async call, we cannot	
        // retrieve it from within the web3Wrapper constructor. This is and should remain the only	
        // call to a web3 instance outside of web3Wrapper in the entire dapp.`

packages/website/ts/blockchain.ts
```diff
@@ -611,6 +576,45 @@ export class Blockchain {
...
+        if (!_.isUndefined(this._injectedProviderIfExists)) {
+            return this._injectedProviderIfExists;
+        }
+        let injectedProviderIfExists = (window as any).ethereum;
+        if (!_.isUndefined(injectedProviderIfExists)) {
+            if (!_.isUndefined(injectedProviderIfExists.enable)) {
+                try {
+                    await injectedProviderIfExists.enable();
+                } catch (err) {
+                    errorReporter.report(err);
```
steveklebanoff: do we want to do something nice in the UX if the user declines?

BMillman19: We could pop up a toast or something, we have the infrastructure to do it but the design is legacy, from a previous version of portal and doesn't fit in with the current design. I don't want to take up more bandwidth for the team / designers for this I mostly just not want it to completely break when the MetaMask change lands.

packages/website/ts/blockchain.ts
BMillman19: We no longer need the networkId to instantiate web3wrapper and we're actually removing the raw web3 call with this PR so the comment no longer applies.

packages/website/ts/blockchain.ts
steveklebanoff: 👍 makes sense

packages/website/ts/blockchain.ts
fragosti: I get that the fall-through is intentional, but don't get why there are two different if-statements here. Anyway I don't think it makes a difference.

PR #1200, opened 2018-10-30
[instant] Progress Bar

steveklebanoff:
## Description

At a high level, this associates a `SimulatedProgress` object with the `OrderState`.  This keeps track of when we started the txn and when we expect to finish the txn.  Then, the `TimedProgressBar` uses this information to figure out how to display the progress bar, including speeding up to finish when the process is done regardless of what the estimated wait time was.

## Checklist:


packages/instant/src/components/buy_order_state_buttons.tsx
```diff
@@ -20,13 +21,12 @@ export interface BuyOrderStateButtonProps {
     onValidationPending: (buyQuote: BuyQuote) => void;
     onValidationFail: (buyQuote: BuyQuote, errorMessage: AssetBuyerError | ZeroExInstantError) => void;
     onSignatureDenied: (buyQuote: BuyQuote) => void;
-    onBuyProcessing: (buyQuote: BuyQuote, txHash: string) => void;
-    onBuySuccess: (buyQuote: BuyQuote, txHash: string) => void;
-    onBuyFailure: (buyQuote: BuyQuote, txHash: string) => void;
+    onBuyProcessing: (buyQuote: BuyQuote, txHash: string, startTimeUnix: number, expectedEndTimeUnix: number) => void;
+    onBuySuccess: (buyQuote: BuyQuote, txHash: string, startTimeUnix: number, expectedEndTimeUnix: number) => void;
```
steveklebanoff: i think ideally you should only have to send in the start and expected times in `onBuyProcessing`, will experiment with a refactor for this

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -25,6 +26,7 @@ export const ZeroExInstantContainer: React.StatelessComponent<ZeroExInstantConta
         >
             <Flex direction="column" justify="flex-start">
                 <SelectedAssetInstantHeading />
+                <SelectedAssetSimulatedProgressBar />
```
BMillman19: this should only show if we're in a processing, success, or failure state right?

packages/instant/src/components/buy_order_state_buttons.tsx
BMillman19: yup i agree

packages/instant/src/components/zero_ex_instant_container.tsx
steveklebanoff: Yes, see https://github.com/0xProject/0x-monorepo/blob/229f11f164ce5109b771295d9aee8ebb74314181/packages/instant/src/containers/selected_asset_simulated_progress_bar.tsx#L18

packages/instant/src/components/simulated_progress_bar.tsx
```diff
@@ -0,0 +1,194 @@
...
+        const intervalId = window.setInterval(this._tick.bind(this), PROGRESS_TICK_INTERVAL_MS);
+        return {
+            percentageDone: 0,
+            intervalId,
+            tickingStatus: { runState: TickingRunState.Running },
+            elapsedTimeMs: 0,
+        };
+    }
+
+    private _tick(): void {
```
steveklebanoff: I feel like there it is probably possible to do an alternative approach that doesn't need this concept of "ticking" every set interval, instead just use CSS animations to show 0% width -> 95% width with an animation time of `expectedTimeMs`, and when done, somehow use just CSS animations to go from `currentWidth` to `100%`.  

This would be cleaner and nicer, That being said, I'm not 100% sure how to do this, and not sure how much more time is worth investing into this, considering the current solution seems to work.

packages/instant/src/components/buy_order_state_buttons.tsx
steveklebanoff: Done, see d675547

packages/instant/src/redux/actions.ts
```diff
@@ -25,7 +25,11 @@ function createAction<T extends string, P>(type: T, data?: P): PlainAction<T> |
 export enum ActionTypes {
     UPDATE_ETH_USD_PRICE = 'UPDATE_ETH_USD_PRICE',
     UPDATE_SELECTED_ASSET_AMOUNT = 'UPDATE_SELECTED_ASSET_AMOUNT',
-    UPDATE_BUY_ORDER_STATE = 'UPDATE_BUY_ORDER_STATE',
+    SET_BUY_ORDER_STATE_NONE = 'SET_BUY_ORDER_STATE_NONE',
```
steveklebanoff: I know we spoke about the pattern of `UPDATE`-prefixed actions taking a parameter and `SET`-prefixed actions not taking any parameters, but `UPDATE_BUY_ORDER_STATE_PROCESSING` reads a lot less naturally than `SET_BUY_ORDER_STATE_PROCESSING` IMHO.  Also, I like having the same prefix for all actions that have to do with buy order state, even if some take parameters and others don't.

packages/instant/tslint.json
```diff
@@ -3,6 +3,7 @@
     "rules": {
         "custom-no-magic-numbers": false,
         "semicolon": [true, "always", "ignore-bound-class-methods"],
-        "max-classes-per-file": false
+        "max-classes-per-file": false,
+        "switch-default": false
```
steveklebanoff: I did this as part of cc2719492d72602ecd1c503cf0ecee95cd05db33

Removing the `default` in the `reducer` function allows the compiler to ensure that we explicitly switch on every `ActionType`.  It prevents us from adding a new `ActionType` and forgetting to have the reducer handle it.

packages/instant/src/components/simulated_progress_bar.tsx
```diff
@@ -0,0 +1,194 @@
...
+}
+interface TickingFinishingStatus {
+    runState: TickingRunState.Finishing;
+    increasePercentageEveryTick: number;
+}
+type TickingStatus = TickingNoneStatus | TickingRunningStatus | TickingFinishingStatus;
+
+export interface SimulatedProgressState {
+    percentageDone: number;
+    intervalId?: number;
```
fragosti: probably doesn't need to be in state. can just be an instance var

packages/instant/src/components/simulated_progress_bar.tsx
fragosti: This approach makes a lot more sense to me. Are you sure it's not worth trying? Re-rendering to move the animation forward doesn't seem ideal.

steveklebanoff: I looked into it a bit and couldn't figure out the right way to do it.  That being said I agree it definitely would be nicer and worthwhile of some more time:  I'll time-box spending some more on time on this and see if I can get it figured out.

packages/instant/src/redux/actions.ts
fragosti: Don't feel strongly about this. The reasoning was that `UPDATE_BUY_ORDER_STATE_PROCESSING` would never really be a name used -- rather it would be `UPDATE_BUY_ORDER_STATE` or `SET_BUY_ORDER_STATE_PROCESSING`

packages/instant/src/components/time_counter.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+
+import { timeUtil } from '../util/time';
+
+import { Flex } from './ui/flex';
+import { Text } from './ui/text';
```
fragosti: still discussing the convention for these imports but just leaving a note here.

packages/instant/src/components/time_counter.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+export interface TimeCounterProps {
+    estimatedTimeMs: number;
+    ended: boolean;
```
fragosti: nit: `isEnded`

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,102 @@
...
+import * as React from 'react';
+
+import { PROGRESS_FINISH_ANIMATION_TIME_MS, PROGRESS_STALL_AT_PERCENTAGE } from '../constants';
+import { ColorOption, keyframes, styled } from '../style/theme';
+
+import { Container } from './ui/container';
+
+export interface TimedProgressBarProps {
+    expectedTimeMs: number;
+    ended: boolean;
```
fragosti: same nit: `isEnded`

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,102 @@
...
+}
+
+interface TimedProgressBarState {
+    animationTimeMs: number;
+    animationStartingWidth: string;
+    maxWidthPercent: number;
+}
+
+export const beginningState = (props: TimedProgressBarProps): TimedProgressBarState => {
```
fragosti: nit: need to export this?

packages/instant/src/constants.ts
```diff
@@ -5,5 +5,8 @@ export const DEFAULT_ZERO_EX_CONTAINER_SELECTOR = '#zeroExInstantContainer';
 export const WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG_PREFIX = 'Transaction failed';
 export const GWEI_IN_WEI = new BigNumber(1000000000);
 export const DEFAULT_GAS_PRICE = GWEI_IN_WEI.mul(6);
+export const DEFAULT_ESTIMATED_TRANSACTION_TIME_MS = 2 * 60 * 1000; // 2 minutes
```
fragosti: maybe make `ONE_SECOND_MS` and use that for the `setInterval` as well.

packages/instant/src/components/buy_button.tsx
```diff
@@ -73,7 +73,9 @@ export class BuyButton extends React.Component<BuyButtonProps> {
             throw e;
         }
 
-        this.props.onBuyProcessing(buyQuote, txHash);
+        const startTimeUnix = new Date().getTime();
+        const expectedEndTimeUnix = startTimeUnix + gasInfo.estimatedTimeMs;
```
fragosti: If it weren't for us only having access to `estimatedTimeMs` here, I would say this kind of logic belongs in the reducer. How do you feel about just passing `estimatedTime` up through the callbacks, and having the `Date.now()` call in the reducer? I guess `startTimeUnix` is more accurate if done here..

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,102 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+
+import { PROGRESS_FINISH_ANIMATION_TIME_MS, PROGRESS_STALL_AT_PERCENTAGE } from '../constants';
```
fragosti: Can these just be props with default values?

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { ColorOption, keyframes, styled } from '../style/theme';
+
+import { Container } from './ui/container';
+
+export interface TimedProgressBarProps {
+    expectedTimeMs: number;
+    ended: boolean;
+}
+
+interface TimedProgressBarState {
```
fragosti: Do you need state? Seems like everything can be computed from `props`.
Something like (pseudoCode):
```
interface AnimationSettings / TimedProgressProps { 
  animationTimeMs: number;
  animationStartingWidth: string;
  maxWidthPercent: number;
}

private readonly _calculateAnimationSettings(): AnimationSettings => {
  if (this.props.isEnded) {
    return {
      animationTimeMs: props.endingAnimationTime,
      animationStartingWidth: this._readWidthFromDOM(),
      maxWidthPercent: 100,
    }
  }
  return {
    animationTimeMs: props.estimatedTime;
    animationStartingWidth: 0;
    maxWidthPercent: props.maxWidthToStallAtPercent;
  } 
}

  return <ProgressBar {{...this._calculateAnimationSettings()}}/>
}

```
I think this is basically what you have but you're saving to state.

Also would it be possible to have start width and end width use the same units?
Don't think there should be both `TimedProgressProps` and `AnimationSettings` or whatever it ends up being.

packages/instant/src/constants.ts
```diff
@@ -5,5 +5,8 @@ export const DEFAULT_ZERO_EX_CONTAINER_SELECTOR = '#zeroExInstantContainer';
...
 export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
+export const PROGRESS_STALL_AT_PERCENTAGE = 95;
```
fragosti: nit: seems like this should just belong in the component file and be default props.

packages/instant/src/containers/selected_asset_progress.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { TimeCounter } from '../components/time_counter';
+import { Container } from '../components/ui';
+import { State } from '../redux/reducer';
+import { OrderProcessState, OrderState, SimulatedProgress } from '../types';
+
+interface SelectedAssetProgressComponentProps {
+    buyOrderState: OrderState;
+}
+// TODO: rename this component and move to seperate file, and get props using mapStateToProps
```
fragosti: 👍

packages/instant/src/containers/selected_asset_progress.tsx
```diff
@@ -0,0 +1,49 @@
...
+> = props => {
+    const { buyOrderState } = props;
+
+    if (
+        buyOrderState.processState === OrderProcessState.PROCESSING ||
+        buyOrderState.processState === OrderProcessState.SUCCESS ||
+        buyOrderState.processState === OrderProcessState.FAILURE
+    ) {
+        const progress = buyOrderState.progress;
+        const ended = buyOrderState.processState !== OrderProcessState.PROCESSING;
```
fragosti: What happens to the bar when it fails? Would be cool if it went to zero? Or freeze?

packages/instant/src/redux/reducer.ts
```diff
@@ -83,11 +83,70 @@ export const reducer = (state: State = INITIAL_STATE, action: Action): State =>
...
+                        ...state,
+                        buyOrderState: {
+                            processState: OrderProcessState.SUCCESS,
+                            txHash: state.buyOrderState.txHash,
+                            progress: successProgress,
+                        },
+                    };
+                }
+            }
+            return state;
```
fragosti: A helper might be useful to DRY up this code

packages/instant/test/util/time.test.ts
```diff
@@ -0,0 +1,48 @@
+import { timeUtil } from '../../src/util/time';
+
+describe('assetDataUtil', () => {
```
fragosti: typo

packages/instant/src/components/time_counter.tsx
fragosti: Or `hasEnded`

fragosti: ... or something

packages/instant/src/components/simulated_progress_bar.tsx
steveklebanoff: Just adding a note: was able to figure this out & do this via CSS, see new `TimedProgressBar`file

packages/instant/src/components/buy_button.tsx
steveklebanoff: I'd prefer to leave as-is as it's a simple & quick calculation, it's nice to grab the startTime immediately, and this is the only place where `onBuyProcessing` is triggered, but can change if you feel strongly

packages/instant/src/components/time_counter.tsx
steveklebanoff: ah!  the boolean prefix convention still hasn't been wired into my thinking yet :)

steveklebanoff: it looks like there is an `eslint` rule for this, but not a `tslint` rule: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/boolean-prop-naming.md 😢

packages/instant/src/components/timed_progress_bar.tsx
steveklebanoff: I'm a little mixed on this -- this is used only in one place, and we don't expect anyone to ever use values other than these defaults, so I'm wondering if making these props is a bit of over-engineering?

packages/instant/src/components/time_counter.tsx
fragosti: I'm a bit confused about tslint. We definitely(?) have this in website, and also a rule about adding `Async` at the end of async functions. Is that not working?

packages/instant/src/components/buy_button.tsx
fragosti: Yea agreed

packages/instant/src/components/timed_progress_bar.tsx
fragosti: Don't feel strongly about this.

packages/instant/src/components/timed_progress_bar.tsx
steveklebanoff: I was unsure about how grabbing the ref width would work outside of the `componentDidUpdate` callback, but it turns out you are so right: I didn't need state here, and it is much cleaner now without it  💯

You deserve a guest post on https://you-dont-need.com/ :)

fragosti: Lol our code is not a good example of "you-dont-need"

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,78 @@
...
+    public render(): React.ReactNode {
+        const timedProgressProps = this._calculateTimedProgressProps();
+        return (
+            <Container width="100%" backgroundColor={ColorOption.lightGrey} borderRadius="6px">
+                <TimedProgress {...timedProgressProps} ref={this._barRef as any} />
+            </Container>
+        );
+    }
+
+    private _calculateTimedProgressProps(): TimedProgressProps {
```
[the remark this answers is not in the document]
steveklebanoff: yea man.  thanks for your help on this, I learned a lot making it!

packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -0,0 +1,78 @@
...
+        if (this.props.hasEnded) {
+            if (!this._barRef.current) {
+                throw new Error('ended but no reference');
+            }
+            const fromWidth = `${this._barRef.current.offsetWidth}px`;
+            return {
+                timeMs: PROGRESS_FINISH_ANIMATION_TIME_MS,
+                fromWidth,
+                toWidth: '100%',
```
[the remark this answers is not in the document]
steveklebanoff: I didn't like when one was a `number` which was a percentage and another one was `string`.  I agree it'd be nice for them to always be the same units, but I'm ok with it as-is, as I feel like it's easy enough to understand as it exists

steveklebanoff:
I had an issue where github was having issues showing the updated version of the PR without conflicts, I merged manually into development through the CLI and pushed up

PR #1201, opened 2018-10-31
[instant] Add `Select Token` mode, animation abstractions, and basic token selection panel

fragosti:
## Description

This is a purely UI branch... didn't add a lot of feature functionality. 
screen shot 2018-10-29 at 6 36 02 pm
screen shot 2018-10-30 at 5 01 29 pm

Does:
- Add abstractions for animation `PositionAnimation` and `SlideAnimation`. 
- Add the 'Select Token' UI.
- Add chevron icon.
- Add click handlers for clicking on symbol or 'Select Token' UI.
- Create basic (not representative) panel component.
- Open and close panel with handlers.

Does not (for another branch):
- Actually "work" if assetData is undefined.
- Implement the token selection dropdown.

## Testing instructions

## Checklist:


packages/instant/src/components/sliding_panel.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+import { ColorOption } from '../style/theme';
+import { zIndex } from '../style/z_index';
+
+import { PositionAnimationSettings } from './animations/position_animation';
+import { SlideAnimation, SlideAnimationState } from './animations/slide_animation';
+import { Button, Container, Text } from './ui';
+
+export interface PanelProps {
+    onClose?: () => void;
```
BMillman19: do we need a default value here?

packages/instant/src/components/animations/slide_animation.tsx
```diff
@@ -0,0 +1,23 @@
+import * as React from 'react';
+
+import { PositionAnimation, PositionAnimationSettings } from './position_animation';
+
+export type SlideAnimationState = 'slidIn' | 'slidOut' | 'none';
+export interface SlideAnimationProps {
+    position: string;
+    animationState: SlideAnimationState;
+    slideIn: PositionAnimationSettings;
```
BMillman19: should `slideIn` and `slideOut` be a bit more descriptive?

packages/instant/src/components/animations/position_animation.tsx
```diff
@@ -0,0 +1,76 @@
...
+        ${generateTransitionInfoCss('from', top, bottom, left, right)}
+    }
+
+    to {
+        position: ${position};
+        ${generateTransitionInfoCss('to', top, bottom, left, right)}
+    }
+`;
+
+export interface TransitionInfo {
```
steveklebanoff: minor but can we move this interface definition to the top of the file? Felt a little confusing to see it used without being defined or imported first

packages/instant/src/components/animations/position_animation.tsx
```diff
@@ -0,0 +1,76 @@
+import { Keyframes } from 'styled-components';
```
steveklebanoff: Sweet helper! 🎉

packages/instant/src/components/animations/position_animation.tsx
```diff
@@ -0,0 +1,76 @@
...
+    top?: TransitionInfo;
+    bottom?: TransitionInfo;
+    left?: TransitionInfo;
+    right?: TransitionInfo;
+    timingFunction: string;
+    duration?: string;
+}
+
+export interface PositionAnimationProps extends PositionAnimationSettings {
+    position: string;
```
steveklebanoff: Can we default to `relative` instead of requiring?

packages/instant/src/components/erc20_asset_amount_input.tsx
```diff
@@ -36,10 +37,18 @@ export class ERC20AssetAmountInput extends React.Component<ERC20AssetAmountInput
         };
     }
     public render(): React.ReactNode {
-        const { asset, onChange, ...rest } = this.props;
-        const amountBorderBottom = this.props.isDisabled ? '' : `1px solid ${transparentWhite}`;
+        const { asset } = this.props;
         return (
             <Container whiteSpace="nowrap">
+                {!_.isUndefined(asset) ? this._renderContentForAsset(asset) : this._renderBackupContent()}
```
steveklebanoff: minor but `{_.isUndefined(asset) ? this._renderBackupContent() : this._renderContentForAsset(asset)}` reads much more naturally to me

packages/instant/src/components/erc20_asset_amount_input.tsx
```diff
@@ -8,13 +8,14 @@ import { BigNumberInput } from '../util/big_number_input';
...
 import { ScalingAmountInput } from './scaling_amount_input';
-import { Container, Text } from './ui';
+import { Container, Flex, Icon, Text } from './ui';
 
 // Asset amounts only apply to ERC20 assets
 export interface ERC20AssetAmountInputProps {
     asset?: ERC20Asset;
     value?: BigNumberInput;
     onChange: (value?: BigNumberInput, asset?: ERC20Asset) => void;
+    onSymbolClick?: (asset?: ERC20Asset) => void;
```
steveklebanoff: `onSymbolClick` feels like a bit of a generic name for me, as `Symbol` could mean many things.   What do you think about something like `onChangeAssetClick` or something that describes more what the user is intending to do?

packages/instant/src/components/erc20_asset_amount_input.tsx
```diff
@@ -36,10 +37,18 @@ export class ERC20AssetAmountInput extends React.Component<ERC20AssetAmountInput
...
             <Container whiteSpace="nowrap">
+                {!_.isUndefined(asset) ? this._renderContentForAsset(asset) : this._renderBackupContent()}
+            </Container>
+        );
+    }
+    private readonly _renderContentForAsset = (asset: ERC20Asset): React.ReactNode => {
+        const { onChange, ...rest } = this.props;
+        const amountBorderBottom = this.props.isDisabled ? '' : `1px solid ${transparentWhite}`;
+        return (
+            <React.Fragment>
```
steveklebanoff: Didn't know about `React.Fragment`.  Love that we don't need to introduce extra divs b/c of this 🆒

packages/instant/src/components/erc20_asset_amount_input.tsx
```diff
@@ -49,18 +58,50 @@ export class ERC20AssetAmountInput extends React.Component<ERC20AssetAmountInput
...
+                >
+                    Select Token
+                </Text>
+                {this._renderChevronIcon()}
+            </Flex>
+        );
+    };
+    private readonly _renderChevronIcon = (): React.ReactNode => {
+        return (
+            <Container marginLeft="5px" onClick={this._handleSymbolClick}>
```
steveklebanoff: Correct me if I'm wrong, but I believe that we want to handle the case where the widget is for just one specific ERC-20 token, and doesn't allow selecting other tokens.  In this case, will `_handleSymbolClick` be undefined and we should not render the chevron icon?

packages/instant/src/components/instant_heading.tsx
```diff
@@ -94,7 +96,7 @@ export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
         return 'I want to buy';
     }
 
-    private _placeholderOrAmount(amountFunction: () => React.ReactNode): React.ReactNode {
+    private _renderPlaceholderOrAmount(amountFunction: () => React.ReactNode): React.ReactNode {
```
steveklebanoff: thanks for this rename!

packages/instant/src/style/z_index.ts
```diff
@@ -0,0 +1,5 @@
+export const zIndex = {
```
steveklebanoff: I like this abstraction 👍

packages/instant/src/components/ui/icon.tsx
```diff
@@ -1,17 +1,21 @@
...
-
 type svgRule = 'evenodd' | 'nonzero' | 'inherit';
 interface IconInfo {
     viewBox: string;
+    path: string;
     fillRule?: svgRule;
     clipRule?: svgRule;
-    path: string;
+    stroke?: string;
+    strokeOpacity?: number;
```
steveklebanoff: I didn't anticipate all these other properties.  With these additions, I'm wondering if this Icon abstraction still feels good -- what do you think?

packages/instant/src/components/ui/index.ts
```diff
@@ -3,3 +3,5 @@ export { Button } from './button';
 export { Flex } from './flex';
 export { Container } from './container';
 export { Input } from './input';
+export { Icon } from './icon';
+export { Spinner } from './spinner';
```
steveklebanoff: Hrm, I haven't been importing from this `index` file, just importing directly from the components themselves.  What is the accepted pattern here?

packages/instant/src/components/erc20_asset_amount_input.tsx
fragosti: Yea I guess I was trying to keep it generic but by including the text "select token" in the component code that was a bit of a lie anyway.

packages/instant/src/components/erc20_asset_amount_input.tsx
fragosti: Yea will not do in this branch. Mostly was concerned with getting this animation working and hooking it up in a very basic way.

packages/instant/src/components/ui/icon.tsx
fragosti: I think we're good for now... worth talking about if it keeps on getting bigger though.

packages/instant/src/components/ui/index.ts
fragosti: I can see how it seems arbitrary... I don't have great reasons for why I went with this pattern. I guess I saw it as kind of a package (like a styleguide) within this package, so structured it that way. Also it's convenient because they are our most used components, and are often used together. Think we should be consistent with this so fixed the other imports. Let me know what you think.

steveklebanoff: I could understand this being convenient if the ui components were exposed as a npm package that someone else would be importing from.

But, upon thinking about this more, I worry that this approach could be bad for code splitting and tree shaking.  Let's say we only use the Spinner and no other ui components in our entry point.  I worry that the code splitting and tree shaking tooling would think that there could be side effects in `ui/index.ts`, and would include all the other ui components in the build.

For that reason, I'd refer that we import directly from the individual component files instead of using this `index` pattern.  It also reduces some boilerplate

steveklebanoff:
Everything looks good to me with the exception of the outstanding discussion about how we import `ui` packages

packages/instant/src/components/ui/index.ts
fragosti: Ok before going deeper down this "if it was an npm package" hypothetical, I'd just like to say it's not and probably won't be :P. As mentioned I don't have great reasons except that I was thinking about it that way and that it is a bit nicer to import them this way as they are our building blocks for the other things in `components/`. It really is mostly a stylistic / convenience thing. 

In terms of the tree shaking discussion, I think it will work as long as your use the `export` `import` syntax which we are in this case (but not in the case where we are exporting a single object, which I agree is not great for tree shaking although I am a fan of the syntax / style). Also I don't think webpack tries to figure out if there are side-effects, you need to tell it whether there are or not.

steveklebanoff: Def agree RE: npm package, and that is a good point about the webpack `sideEffects` config!

Overall I relate to the thought that you've brought up in other scenarios, in which there should be one accepted way to do things.  I'd prefer not to have to think about: "Do I import this component directly from the file, or do I import it from its' parent `index`? " 

I personally don't feel like this index pattern makes things really any more convenient, and it introduces inconsistencies with how we import objects, and adds additional (although small) overhead to introducing a new UI component.

For the sake of keeping things simple and consistent, I feel pretty strongly that we should just import the components directly.

fragosti: Chatted in Slack. While I do prefer the `./iui` pattern, @steveklebanoff feels pretty strongly it should be a direct import, brings up some good points, and is willing to make the change, so we will follow up on this in a follow-up branch :)

PR #1205, opened 2018-11-01
[instant] Dismissible overlay

BMillman19:
## Description

* Export a new component that encases the instant component in a dismissible overlay
* Update umd bundle to use the new ZeroExInstantOverlay component

## Testing instructions

## Types of changes

## Checklist:

### DIFF
--- a/packages/instant/src/components/ui/button.tsx
+++ b/packages/instant/src/components/ui/button.tsx
@@ -3,6 +3,11 @@ import * as React from 'react';
 
 import { ColorOption, styled } from '../../style/theme';
 
+export enum ButtonHoverStyle {
+    Darken = 0,
+    Opacity,
+}
+
 export interface ButtonProps {
     backgroundColor?: ColorOption;
     borderColor?: ColorOption;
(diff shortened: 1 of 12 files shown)


packages/instant/src/components/ui/button.tsx
```diff
@@ -3,6 +3,11 @@ import * as React from 'react';
 
 import { ColorOption, styled } from '../../style/theme';
 
+export enum ButtonHoverStyle {
```
fragosti: Is this only to support the close `Icon`? I would bake click functionality directly into `Icon`. This type of configuration for the button seems like too much, all buttons should just behave the same way I think...

packages/instant/src/components/ui/overlay.tsx
```diff
@@ -0,0 +1,49 @@
...
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: ${props => props.zIndex}
+    background-color: ${overlayBlack};
+`;
+
+Overlay.defaultProps = {
+    onClose: util.boundNoop,
```
fragosti: nit: don't think you need this.

packages/instant/src/style/theme.ts
```diff
@@ -26,8 +27,10 @@ export const theme: Theme = {
     white: 'white',
     lightOrange: '#F9F2ED',
     darkOrange: '#F2994C',
+    clear: 'rgba(0, 0, 0, 0.0)',
```
fragosti: I think this just belongs with the other opacity-based values below?

fragosti: We need to figure out this color stuff... apologies if it's a bit confusing right now.

PR #1206, opened 2018-11-02
[instant] Fix decimal input issues on instant

steveklebanoff:
## Description

This fixes the decimal issues with the input in instant.  It gets rid of the `BigNumberInput`.  It changes `ScalingAmountInput` to internally keep track of the raw string value the user put in, and triggers `onChange` with either a `BigNumber` if the users input parses to a valid (non-NaN) BigNumber, or `undefined`.  

I like this approach because it allows us to keep the interface of having `selectedAssetAmount` always being a `BigNumber | undefined`.  I don't like this approach because we have to have a silly safeguard to ensure that the prop `value` isn't out of sync with the state's `stringValue`.  These values should never get out of sync, but I wanted to take extra care since we're dealing with important numbers.  This approach feels like a bit of a react anti-pattern though.

The first approach I prototyped was storing `selectedAssetAmount` as a `string`, and parsed into a `BigNumber` when necessary.  That approach did not suffer from the `state` issue mentioned, but felt a little unnatural,  because outside of this component we only cared if it was a parseable `BigNumber` or `undefined`.

Another approach would be expanding `BigNumberInput` to store the whole raw input as an attribute. Or to store and create a different type on the global state, like
```
interface BigNumberWithInput {
  parsedBigNumber?: BigNumber;
  rawInput: string;
}
```

It feels like there are two kind of "buckets" of approaches.  There is this PR, which localizes the raw input string concept to `ScalingAmountInput`, and leaves the global state clean and unaware of raw input strings.  It does this at the expense of the code in this `ScalingAmountInput` component being kind of funky though.   Then there are the other approaches, which make the code in  `ScalingAmountInput` more straightforward, but keep the raw string somewhere in the global state, even though it is really only needed in a small part of the code.  Both approaches have trade-offs.

## Checklist:


packages/instant/src/components/scaling_amount_input.tsx
```diff
@@ -12,17 +13,56 @@ export interface ScalingAmountInputProps {
...
-    onChange: (value?: BigNumberInput) => void;
+    value?: BigNumber;
+    onChange: (value?: BigNumber) => void;
     onFontSizeChange: (fontSizePx: number) => void;
 }
+interface ScalingAmountInputState {
+    stringValue: string;
+}
+
+const stringToMaybeBigNumber = (stringValue: string): MaybeBigNumber => {
```
steveklebanoff: could move this and function below into helper if we decide on this approach

packages/instant/src/components/scaling_amount_input.tsx
```diff
@@ -12,17 +13,56 @@ export interface ScalingAmountInputProps {
...
+        };
+    }
+    public componentDidUpdate(prevProps: ScalingAmountInputProps): void {
+        const parsedStateValue = stringToMaybeBigNumber(this.state.stringValue);
+        const currentValue = this.props.value;
+
+        if (!areMaybeBigNumbersEqual(parsedStateValue, currentValue)) {
+            // we somehow got into the state in which the value passed in and the string value
+            // in state have differed, reset state
+            this.setState({
```
steveklebanoff: could alternatively raise here

packages/instant/src/components/scaling_amount_input.tsx
```diff
@@ -12,17 +13,56 @@ export interface ScalingAmountInputProps {
...
+}
+
+const stringToMaybeBigNumber = (stringValue: string): MaybeBigNumber => {
+    let maybeBigNumber: MaybeBigNumber;
+    try {
+        maybeBigNumber = new BigNumber(stringValue);
+    } catch {
+        maybeBigNumber = undefined;
+    }
+    return _.isNaN(maybeBigNumber) ? undefined : maybeBigNumber;
```
steveklebanoff: should use BigNumber's isNaN here instead of lodash

steveklebanoff:
@fragosti cool, thanks for the review.  I agree that the input is purely based on the redux value is a downside, but we have the safeguard in `componentDidUpdate` to make sure the string value and the parsed big number never get out of sync (which should never happen).  I made some improvements on this branch if you want to take another gander.

steveklebanoff:
@BMillman19 I like the concept of "Maybe Big Number" here as in my mind it makes it easier to label things and reason about, and lends itself nicely to something like the `maybeBigNumberUtil`.  That being said, I like your generic `Maybe` idea, and added that.  If you're cool w/ it as-is, would like to merge.  Open to further discussion if you're feeling against the Maybe pattern in general though.

PR #1207, opened 2018-11-02
[instant][asset-buyer] Implement affiliateInfo prop

BMillman19:
## Description

* Add `affiliateFeeInfo` as a prop to instant and option for the umd entry point
* Also fix a couple bugs in asset buyer where we were accidentally overriding default values for some methods and also not rounding when calculating the affiliate fee

## Testing instructions

```
http://localhost:5000/?feeRecipient=0x14e2f1f157e7dd4057d02817436d628a37120fd1&feePercentage=0.02
```

## Types of changes

## Checklist:

### DIFF
--- a/packages/instant/public/index.html
+++ b/packages/instant/public/index.html
@@ -77,11 +77,21 @@
             onClose: () => { console.log('0x Instant Closed') }
         }
         const liquiditySourceOverride = queryParams.getQueryParamValue('liquiditySource');
+        const feeRecipientOverride = queryParams.getQueryParamValue('feeRecipient');
+        const feePercentageOverride = +queryParams.getQueryParamValue('feePercentage');
+        let affiliateInfoOverride;
+        if (feeRecipientOverride !== undefined && feePercentageOverride !== undefined) {
+            affiliateInfoOverride = {
+                feeRecipient: feeRecipientOverride,
+                feePercentage: feePercentageOverride
+            };
+        }
         const renderOptionsOverrides = {
             liquiditySource: liquiditySourceOverride === 'provided' ? providedOrders : liquiditySourceOverride,
             assetData: queryParams.getQueryParamValue('assetData'),
             networkId: +queryParams.getQueryParamValue('networkId') || undefined,
             defaultAssetBuyAmount: +queryParams.getQueryParamValue('defaultAssetBuyAmount') || undefined,
+            affiliateInfo: affiliateInfoOverride,
         }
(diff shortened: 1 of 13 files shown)


packages/instant/public/index.html
```diff
@@ -74,6 +74,10 @@
         const renderOptionsDefaults = {
             liquiditySource: 'https://api.radarrelay.com/0x/v2/',
             assetData: '0xf47261b0000000000000000000000000e41d2489571d322189246dafa5ebde1f4699f498',
+            affiliateFeeInfo: {
+                recipientAddress: '0x14e2f1f157e7dd4057d02817436d628a37120fd1',
+                percentage: 0.02,
+            },
```
fragosti: no option to override these in the url ?

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
```diff
@@ -120,7 +134,7 @@ const mapDispatchToProps = (
             // even if it's debounced, give them the illusion it's loading
             dispatch(actions.setQuoteRequestStatePending());
             // tslint:disable-next-line:no-floating-promises
-            debouncedUpdateBuyQuoteAsync(assetBuyer, dispatch, asset, value);
+            debouncedUpdateBuyQuoteAsync(assetBuyer, dispatch, asset, value, affiliateFeeInfo);
```
fragosti: I thought we got rid of `debounced` in another branch? Is it back because of a bad merge?

packages/instant/src/index.umd.ts
```diff
@@ -24,6 +24,9 @@ export const render = (props: ZeroExInstantOverlayProps, selector: string = DEFA
     if (!_.isUndefined(props.zIndex)) {
         assert.isNumber('props.zIndex', props.zIndex);
     }
+    if (!_.isUndefined(props.affiliateFeeInfo)) {
+        assert.isValidAffiliateFeeInfo('props.affiliateFeeInfo', props.affiliateFeeInfo);
```
fragosti: Is there no option to pass this in from the URL?

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
BMillman19: I brought it back we cause while the first AssetBuyer request is going we can still run unto a case where we make a bunch of parallel requests

BMillman19:
I'll make a ticket for your AffiliateInfo on constructor suggestion

PR #1216, opened 2018-11-06
[order_utils.py] is_signature_valid, via Exchange contract

feuGeneA:
## Description

Initial support for signature validation, done via Exchange contract's isValidSignature() method.

## Testing instructions

`./setup.py ganache`
`./setup.py test`

Also there are test runs integrated with CI.

## Types of changes

## Checklist:


python-packages/order_utils/setup.py
```diff
@@ -62,6 +62,15 @@ def run(self):
         with open(path.join(eth_abi_dir, "py.typed"), "a"):
             pass
 
+        # HACK(gene): until eth_utils fixes
+        # https://github.com/ethereum/eth-utils/issues/140 , we need to simply
+        # create an empty file `py.typed` in the eth_abi package directory.
+        import eth_utils
+
+        eth_utils_dir = path.dirname(path.realpath(eth_utils.__file__))
+        with open(path.join(eth_utils_dir, "py.typed"), "a"):
```
LogvinovLeon:
```python
from pathlib import Path
```
```suggestion
        Path(path.join(eth_utils_dir, "py.typed")).touch()
```

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -1 +1,11 @@
-"""Order utilities for 0x applications."""
+"""Order utilities for 0x applications.
+
+Some methods require the caller to pass in a `Web3.HTTPProvider` object.  For
+local testing one may construct such a provider pointing at an instance of
+`ganache-cli <https://www.npmjs.com/package/ganache-cli>`_ which has the 0x
+contracts deployed in it.  For convenience, a docker container is provided for
```
fabioberger:
```suggestion
contracts deployed on it.  For convenience, a docker container is provided for
```

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -1 +1,11 @@
...
+just such this purpose.  To start it: ``docker run -d -p 8545:8545
```
fabioberger:
```suggestion
just this purpose.  To start it: ``docker run -d -p 8545:8545
```

python-packages/order_utils/src/conf.py
```diff
@@ -13,6 +13,9 @@
 copyright = "2018, ZeroEx, Intl."
 author = "F. Eugene Aumson"
 version = "0.1.0"  # The short X.Y version
+# we should modify this version assignment to pull from setup.py like for
+# __version__ in in
```
fabioberger:
```suggestion
# __version__ in
```

python-packages/order_utils/src/zero_ex/order_utils/signature_utils.py
```diff
@@ -0,0 +1,86 @@
...
+
+    >>> is_valid_signature(
+    ...     Web3.HTTPProvider("http://127.0.0.1:8545"),
+    ...     '0x6927e990021d23b1eb7b8789f6a6feaf98fe104bb0cf8259421b79f9a34222b0',
+    ...     '0x1B61a3ed31b43c8780e905a260a35faefcc527be7516aa11c0256729b5b351bc3340349190569279751135161d22529dc25add4f6069af05be04cacbda2ace225403',
+    ...     '0x5409ed021d9299bf6814279a6a1411a7e866a631',
+    ... )
+    (True, '')
+    """  # noqa: E501 (line too long)
+    if not isinstance(provider, Web3.HTTPProvider):
```
fabioberger: Hm... is there a way to make this more flexible? Literally we should accept any class instance with a public `def make_request(self, method, params)` method used to send JSON RPC calls. Perhaps we can define an abstract class for this?

feuGeneA: I love the idea but I'm not sure it's that simple.  I removed this type check, and coded up a test of what you're describing, and the test fails.  [Details here](https://gist.github.com/feuGeneA/eed20d7e62bebb93957adc8335006df3).  Maybe the method to implement is actually `request_func()`?  What made you suggest `make_request`?  If it's `request_func()` I'd like to have a test for it (like this one) but I'd have to figure out what's up with that `middleware_stack` argument and how to construct one in the test.  For more context, that `web3/manager.py:89` cited by my gist is [here](https://github.com/ethereum/web3.py/blob/4c52f0d592571077ad2573cefe0587a336685d44/web3/manager.py#L89).

python-packages/order_utils/src/conf.py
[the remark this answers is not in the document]
feuGeneA: Sorry, that was more of a TODO, which I meant to clean up or implement before committing.  I cleaned it up.

python-packages/order_utils/src/zero_ex/order_utils/signature_utils.py
fabioberger: I read through Web3.py source and saw that method in both the HTTP and IPC providers.

fabioberger: For now, I'm ok with adding a `TODO` comment here for  adding an assertion and shipping this.

PR #1219, opened 2018-11-06
[instant] Mobile full view

steveklebanoff:
## Description

Shows widget in "full bleed" display when on a mobile device (or browser w/ small width)

lhbuy

responsiveinstant

## Checklist:


packages/instant/src/components/ui/container.tsx
```diff
@@ -33,21 +35,20 @@ export interface ContainerProps {
     cursor?: string;
     overflow?: string;
     darkenOnHover?: boolean;
+    flexGrow?: string | number;
```
steveklebanoff: I originally thought to only add this to the `Flex` component, but replacing the container that needed this with a `Flex` messed things up.  I ended up feeling as though this was reasonable to add onto a `Container`

packages/instant/src/components/erc20_token_selector.tsx
```diff
@@ -35,7 +35,7 @@ export class ERC20TokenSelector extends React.Component<ERC20TokenSelectorProps>
                     value={this.state.searchQuery}
                     onChange={this._handleSearchInputChange}
                 />
-                <Container overflow="scroll" height="275px" marginTop="10px">
+                <Container overflow="scroll" height={{ default: '275px', sm: '75vh' }} marginTop="10px">
```
steveklebanoff: `75vh` feels like an arbitrary number here, I essentially want this to be "100%" but if I set it to that, then the  `overflow: scroll;` doesn't work.

steveklebanoff:
This isn't unique to this branch, but it is more prominent.  I think when we use prop names that are the same as CSS properties, they get set as attributes in the DOM?

image

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -38,8 +42,9 @@ export class ZeroExInstantContainer extends React.Component<ZeroExInstantContain
                     borderRadius="3px"
                     hasBoxShadow={true}
                     overflow="hidden"
+                    height="100%"
                 >
-                    <Flex direction="column" justify="flex-start">
+                    <Flex direction="column" height={'100%'} justify="flex-start">
```
fragosti: nit: can just be string.

packages/instant/src/components/ui/container.tsx
```diff
@@ -65,6 +65,9 @@ export const Container =
     ${props => cssRuleIfExists(props, 'cursor')}
     ${props => cssRuleIfExists(props, 'overflow')}
     ${props => (props.hasBoxShadow ? `box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.1)` : '')};
+    ${props => stylesForMedia('display', props.display || 'initial')}
```
fragosti: nit: should we use defaultProps for this `||` functionality ?

packages/instant/src/components/ui/container.tsx
fragosti: Yea we considered creating a `FlexItem` but I think this works.

packages/instant/src/components/erc20_token_selector.tsx
fragosti: Yea... still not sure whether it's possible to make scrolling work without a fixed height here?  

Using`vh` here makes sense to me. If we know the exact offset from the bottom maybe something like `calc(100vh - 100px)` makes sense too.

BMillman19:
Hmm, as a user of instant on mobile, how do I go back to the original app I was using?

steveklebanoff:
@BMillman19 I spoke to Chris about this, and here was his feedback:

> Since it is a full screen experience and may not feel like a modal or an overlay. I was thinking the user could just use the browser's back button or swipe to go back like you normally do on mobile web.

If we wanted to, we could add a "Close" button on the bottom, or figure out how to make the "X" in the top right look nice.   The only mobile experiences we have in the pipeline are dedicated Instant dapps which this doesn't apply to, so I think we could leave this as a future task.

PR #1220, opened 2018-11-07
[instant] Prevent CSS leakage

fragosti:
## Description

This approach is not bullet-proof. If an integrator wanted to, they could use `!important` on a CSS rule targeting `button` and still mess with the styles of instant. However I think this will handle basically all cases where the integrator may unintentionally be affecting the styles of instant.

Looking around at how other people do widget style isolation -- the most popular option seems to be an `iframe`. I didn't dig into this too much, but if this solution is not enough we may have to figure out how to use the `iframe` solution.

Approach:
- Add `external.css` file with some cases that we may encounter. 
- Make instant look normal again. 

Changes:
- Add `CSSReset` component that applies a **scoped** CSS reset to HTML elements at the root of the app.
- Selectively use `all: initial` to basically turn off CSS inheritance for certain components. 
- Use styled-components `&&` syntax to increase specificity of our UI components (https://www.styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting). This just results in the class name being repeated in the CSS selector.
- Make the `Button` component not rely on nested `Text` component for text styles.

## Checklist:


packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -27,35 +28,37 @@ export class ZeroExInstantContainer extends React.Component<ZeroExInstantContain
...
-                        <LatestBuyQuoteOrderDetails />
-                        <Container padding="20px" width="100%">
-                            <SelectedAssetBuyOrderStateButtons />
-                        </Container>
-                    </Flex>
-                    <SlidingPanel
-                        title="Select Token"
-                        animationState={this.state.tokenSelectionPanelAnimationState}
-                        onClose={this._handlePanelClose}
+            <CSSReset>
```
BMillman19: We were discussing a bit yesterday about if this would go here or in the umd entry point. This doesn't cover the overlay, correct?

fragosti: yea need to think about this a bit more. However, I think if we are taking this CSS approach to isolation, instead of the iFrame one, we should apply it to both bundles as it is one of the main advantages

PR #1221, opened 2018-11-07
[instant] Ensure we always have a provider from initial state

BMillman19:
## Description

Updates to redux `State`:
* Decomposed the `State` type into `DefaultState`, `PropsDerivedState`, and `OptionalState`
* `DefaultState`: state that is required AND we have defaults for before we have the initial state
* `PropsDerivedState`: state that is required, but is derived from the props
* `OptionalState`: state that is optional
* We now only merge the initial props with some `DefaultState` object
* This gives us a better understanding about the starting state of our component and lets use guarantee the presence of a provider on the props (less optionality)
* Note: when merging the initial props, we still might override defaults. `PropsDerivedState` is a bit confusing to I'm open to renaming

Provider / Account Handling:
* Added a new property `providerState` to `PropsDerivedState` 
* This object contains all state derived from the provider and must be updated if the provider updates
* The object also has an `account` property which can be in one of 5 states: None, Loading, Ready, Locked, and Error.
* Added a `providerStateFactory` utility to house logic for creating the initial `providerState`
* Added a `providerFactory` utility to house logic for creating different providers
* `providerFactory.getFallbackNoSigningProvider` creates a web3 provider engine composed of `EmptyWalletSubprovider` and `RPCSubprovider`. We rely on infura's default api in this scenario
* Note: There is a TODO item to make us less subject to infura failures / outages

Other:
* Updated the casing on some enums in `types.ts` to make them more conventional
* Cleaned up a lot isUndefined checks now that we are guaranteed to have an `AssetBuyer` object in all components

## Testing instructions

* Try in incognito mode

## Types of changes

## Checklist:


BMillman19:
@fragosti the state transitions are coming in another branch

package.json
```diff
@@ -50,7 +50,7 @@
             },
             {
                 "path": "packages/instant/public/main.bundle.js",
-                "maxSize": "500kB"
+                "maxSize": "1000kB"
```
steveklebanoff: 😢

packages/instant/src/constants.ts
```diff
@@ -13,3 +16,8 @@ export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
 export const PROGRESS_STALL_AT_WIDTH = '95%';
 export const PROGRESS_FINISH_ANIMATION_TIME_MS = 200;
+export const ETHEREUM_NODE_URL_BY_NETWORK = {
+    [Network.Mainnet]: 'https://mainnet.infura.io/',
+    [Network.Kovan]: 'https://kovan.infura.io/',
+};
+export const BLOCK_POLLING_INTERVAL_MS = 10000; // 10s
```
steveklebanoff: would be good to use the `ONE_SECOND_MS` constant here, i.e. `10 * ONE_SECOND_MS`

PR #1222, opened 2018-11-07
[WIP] Add two helpful utility methods for `SignedOrder` and JSON `SignedOrder`

TravisGDev:
##Requirements
Add two helpful utility method to `@0x/order-utils`. One that takes in a JSON SignedOrder and returns a SignedOrder where the numerical values have been cast to BigNumber's. Another that takes in a SignedOrder and returns it's JSON representation.

You can use the `convertOrderStringFieldsToBigNumber` method in `parsing_utils` as a starting point. It takes in a a JSON signedOrder and returns a signedOrder. Ideally we would make sure what is passed in is of type: `SignedOrder`, including an assertion check that it conforms to the `signedOrderSchema` and rename this method to `convertOrderJsonToOrder` and then write a second one that does the reverse.

## Checklist:

*   [x ] Prefix PR title with `[WIP]` if necessary.
*   [x ] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.

### DIFF
--- a/packages/order-utils/test/assert_test.ts
+++ b/packages/order-utils/test/assert_test.ts
@@ -13,7 +13,7 @@ describe('Assertion library', () => {
     describe('#isSenderAddressHexAsync', () => {
         it('throws when address is invalid', async () => {
             const address = '0xdeadbeef';
-            const varName = 'address';
+            const varName = 'addresss';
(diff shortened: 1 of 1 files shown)


LogvinovLeon:
@TravisGDev I can't see the actual changes. You've probably chosen the wrong branch.

PR #1229, opened 2018-11-08
[instant] Dismissible overlay for mobile errors

steveklebanoff:
## Description

Shows a dark overlay over the widget when an error occurs in mobile.  Upon clicking the overlay, the error is dismissed.

⚠️  Should squash & merge since there are a lot of experimental commits here ⚠️ 

dismissibleerrorgif

## Checklist:


steveklebanoff:
`ScalingInput` issue should be fixed with 38ed866c5a80ae83e67c01de7249956877c3191a

packages/instant/src/components/ui/overlay_container.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { generateMediaWrapper, ScreenWidths } from '../../style/media';
+import { styled } from '../../style/theme';
+import { zIndex } from '../../style/z_index';
+
+import { Container, ContainerProps } from './container';
+
+interface OverlayProps {
+    showMaxWidth?: ScreenWidths;
+}
+const Overlay =
```
steveklebanoff: When I first started out I was using an `:after` selector, but this has evolved into looking similar enough that it might make sense to modify the existing `Overlay` component instead of generating this new one.

steveklebanoff: This might make sense as a refactor once https://github.com/0xProject/0x-monorepo/pull/1230 is merged, as 1230 is what makes the more generic Overlay

packages/instant/src/style/z_index.ts
```diff
@@ -2,5 +2,6 @@ export const zIndex = {
     errorPopBehind: 1,
     mainContainer: 2,
     panel: 3,
-    errorPopUp: 4,
+    containerOverlay: 4,
```
steveklebanoff: This will have conflicts with https://github.com/0xProject/0x-monorepo/pull/1230/

steveklebanoff:
Updated w/ development and ready for review

packages/instant/src/containers/latest_error_overlay.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import { OverlayContainer, OverlayContainerProps } from '../components/ui/overlay_container';
+
+import { Action } from '../redux/actions';
+import { State } from '../redux/reducer';
+import { ScreenWidths } from '../style/media';
+import { DisplayStatus } from '../types';
+import { errorFlasher } from '../util/error_flasher';
+
+type ConnectedState = Pick<OverlayContainerProps, 'showOverlay' | 'showMaxWidth'>;
```
fragosti: Oh I like this.

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -45,14 +46,16 @@ export class ZeroExInstantContainer extends React.Component<ZeroExInstantContain
...
                     >
-                        <Flex direction="column" justify="flex-start" height="100%">
-                            <SelectedAssetInstantHeading onSelectAssetClick={this._handleSymbolClick} />
-                            <SelectedAssetBuyOrderProgress />
-                            <LatestBuyQuoteOrderDetails />
-                            <Container padding="20px" width="100%">
-                                <SelectedAssetBuyOrderStateButtons />
-                            </Container>
-                        </Flex>
+                        <LatestErrorOverlay>
```
fragosti: Can it just be a sibling?

steveklebanoff: Yes, thanks for calling this out.  This hierarchy was a relic of when the overlay was implemented with an `:after` css selector

packages/instant/src/containers/latest_error_overlay.tsx
steveklebanoff: Yea, I think we could benefit from adopting this pattern for all of our containers.  Would help reduce work when interfaces change.

steveklebanoff:
@fragosti thanks for the callout about this being able to be done all in `LatestError` -- a lot of this design originated from the first pass which was before the more generic `Overlay`, and was based on an `:after` css effect.  I've updated this to just be handled in `LatestError` and the code feels a lot cleaner now 🎉 

I think this overlay would benefit from a fade-in/fade-out, and it looks like that is the way Chris designed it in his mockup, but I would like to get this in as-is, and add the transition as a follow up step

PR #1232, opened 2018-11-08
[instant] Request account address and balance at mount

BMillman19:
## Description

* Add a new function `asyncData.fetchAccountInfoAndDispatchToStore()` that kicks off a request for the account address and balance
* This function drives state updates to the `state.providerState.account` object in redux

## Testing instructions

* Observe appropriate account states in redux inspector

## Types of changes

## Checklist:


packages/instant/src/redux/async_data.ts
```diff
@@ -33,4 +34,40 @@ export const asyncData = {
...
+            store.dispatch(actions.setAccountStateLoading());
+        }
+        let availableAddresses: string[];
+        try {
+            availableAddresses = await web3Wrapper.getAvailableAddressesAsync();
+        } catch (e) {
+            store.dispatch(actions.setAccountStateError());
+            return;
+        }
+        if (!_.isEmpty(availableAddresses)) {
```
steveklebanoff: super nit (not necessary to change) but I'd prefer for this to follow the `return` pattern above, i.e.

```
if (_.isEmpty(availableAddresses)) {
   store.dispatch(actions.setAccountStateLocked());
   return;
}

... setAccountStateReady code ..
```

packages/instant/src/redux/async_data.ts
```diff
@@ -33,4 +34,40 @@ export const asyncData = {
...
+            const activeAddress = availableAddresses[0];
+            store.dispatch(actions.setAccountStateReady(activeAddress));
+            await asyncData.fetchAccountBalanceAndDispatchToStore(store);
+        } else {
+            store.dispatch(actions.setAccountStateLocked());
+        }
+    },
+    fetchAccountBalanceAndDispatchToStore: async (store: Store) => {
```
steveklebanoff: Is it possible that we'd have a concurrency issue in which we'd dispatch a balance for an account that they have since changed?   I feel like we may need a similar check here as we have for the buy quote.  

Alternatively, part of me feels like balance fetching could be paired with address fetching.  This could be easier to reason about -- we always get the address and the balance together, or the whole thing is considered "failed".  I'm not married to this though, what do you think?

BMillman19:
@fragosti im inclined to leave the balance as optional in the AccountReady state because it is not core to the functionality of instant. Even if we did not have the balance, we would still be able to function normally, we would just not be able to do the "do you have enough ETH" validation, which is not a big deal because it will most likely be caught at the wallet layer

packages/instant/src/redux/async_data.ts
BMillman19: Yes, you are correct wrt to the balance update coming late issue

steveklebanoff:
Thanks for the concurrency fix 🎉 !

Issue #1218, opened 2018-11-06
validateOrderFillableOrThrowAsync swallows errors and returns incorrect RevertReason


prettymuchbryce:
## Expected Behavior

Calling `zeroExContractWrappers.exchange.validateOrderFillableOrThrowAsync` should return an accurate `RevertReason`, or throw a relevant error if the underlying Ethereum node is unreachable. 

## Current Behavior

Instead `validateFillOrderBalancesAllowancesThrowIfInvalidAsync` in `order_validation_utils` will swallow any connectivity errors and instead return the revert reason as `TRANSFER_FAILED`.

https://github.com/0xProject/0x-monorepo/blob/development/packages/order-utils/src/order_validation_utils.ts#L109

## Possible Solution

Don't swallow errors, and don't ignore the possibility of connectivity issues when dealing with requests over a network. Check the type of error and return an accurate error to the client so that the client can decide how to handle the error.

## Context

We see this when there is any sort of connectivity issue, we have no way to discern it from a legitimate TRANSFER_FAILED error. It potentially could cause clients to mark an order as unfillable when in fact there was just a temporary connection issue.

## Your Environment

|             Package | Version |
| ------------------: | :------ |
|             `0x.js` | ^1.0.7  |

| Network |
| ------- |
| mainnet |

fabioberger:
Thanks for the report @prettymuchbryce, will look into fixing this now.

fabioberger:
@prettymuchbryce we've fixed this issue and published a new version of `@0x/contract-wrappers`.

PR #1235, opened 2018-11-09
[order-utils] Fix order validation method

fabioberger:
## Description

Artek from BambooRelay notified us on Rocket.chat #dev that our validation method was missing some critical checks. This PR fixes that. We now check:

If an:
- Order signature is valid
- Order was cancelled (with a regular cancel or via cancelledOrdersUpTo)

This is a short-term fix. Long term, we are hoping to improve the error messages returned from our smart contracts, such that validation can be done by "calling" the method that will be filling/cancelling the order/orders.

Additionally, this addresses the network errors being swallowed during validation issue raised here: https://github.com/0xProject/0x-monorepo/issues/1218

## Checklist:


packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1114,6 +1117,20 @@ export class ExchangeWrapper extends ContractWrapper {
         signedOrder: SignedOrder,
         opts: ValidateOrderFillableOpts = {},
     ): Promise<void> {
+        assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
+        assert.doesConformToSchema('opts', opts, validateOrderFillableOptsSchema);
+
+        const orderHash = await orderHashUtils.getOrderHashHex(signedOrder);
+        const isValidSignature = await signatureUtils.isValidSignatureAsync(
```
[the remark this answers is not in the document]
fabioberger: Did it although this whole class is going to disappear very soon.

packages/contracts/test/utils/exchange_wrapper.ts
```diff
@@ -219,6 +219,10 @@ export class ExchangeWrapper {
         const isCancelled = await this._exchange.cancelled.callAsync(orderHashHex);
         return isCancelled;
     }
+    public async getOrderEpochAsync(makerAddress: string, takerAddress: string): Promise<BigNumber> {
```
dekz: This should be `senderAddress`

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -146,12 +139,29 @@ export class OrderValidationUtils {
...
-            filledTakerTokenAmount,
+        const isValidSignature = await signatureUtils.isValidSignatureAsync(
+            this._provider,
+            orderHash,
+            signedOrder.signature,
+            signedOrder.makerAddress,
         );
-        OrderValidationUtils._validateOrderNotExpiredOrThrow(signedOrder.expirationTimeSeconds);
+        if (!isValidSignature) {
+            throw new Error('INVALID_ORDER_SIGNATURE');
```
dekz: Can we map these to RevertReasons where possible? 

```
RevertReason.InvalidOrderSignature
```

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -146,12 +139,29 @@ export class OrderValidationUtils {
...
+        }
+
+        const isCancelled = await this._orderFilledCancelledFetcher.isOrderCancelledAsync(signedOrder);
+        if (isCancelled) {
+            throw new Error('CANCELLED');
```
dekz: Since Exchange will revert with an `ORDER_UNFILLABLE` generic error we should refactor some of this to `getOrderInfo` and map the states to a enum of errors.

Happy for this to be a later TODO item, as it would be a larger refactor, replacing some of the balance/allowance cancel fetchers.

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -210,13 +219,30 @@ export class OrderValidationUtils {
...
+                zrxAssetData,
+            );
+        } catch (err) {
+            const transferFailedErrorMessages = [
+                ExchangeContractErrs.InsufficientMakerBalance,
+                ExchangeContractErrs.InsufficientMakerFeeBalance,
+                ExchangeContractErrs.InsufficientTakerBalance,
+                ExchangeContractErrs.InsufficientTakerFeeBalance,
+                ExchangeContractErrs.InsufficientMakerAllowance,
+                ExchangeContractErrs.InsufficientMakerFeeAllowance,
```
dekz: As part of the larger refactor we should also remember to remove the older `ExchangeContractErrs` which are no longer applicable.

packages/contract-wrappers/test/exchange_wrapper_test.ts
```diff
@@ -282,6 +282,18 @@ describe('ExchangeWrapper', () => {
...
+    describe('#validateOrderFillableOrThrowAsync', () => {
+        it('should throw if signature is invalid', async () => {
+            const signedOrderWithInvalidSignature = {
+                ...signedOrder,
+                signature: '0xdeadbeef',
+            };
+
+            expect(
+                contractWrappers.exchange.validateOrderFillableOrThrowAsync(signedOrderWithInvalidSignature),
+            ).to.eventually.to.be.rejected();
```
LogvinovLeon: Can we test for the exact rejection reason? `rejectedWith`

packages/contracts/test/utils/exchange_wrapper.ts
```diff
@@ -219,6 +219,10 @@ export class ExchangeWrapper {
...
+        const orderEpoch = new BigNumber(await this._exchange.orderEpoch.callAsync(makerAddress, takerAddress));
```
LogvinovLeon: Is there a reason to convert to BigNumber here? As far as I remember all contract wrappers already return our BigNumbers

packages/contracts/test/utils/exchange_wrapper.ts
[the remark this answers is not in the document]
fabioberger: @dekz in the future can you use the new "suggestion" feature on Github reviews? Esp. for small one-line changes like this.

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -146,12 +139,29 @@ export class OrderValidationUtils {
...
+            throw new Error('INVALID_ORDER_SIGNATURE');
+        }
+
+        const isCancelled = await this._orderFilledCancelledFetcher.isOrderCancelledAsync(signedOrder);
+        if (isCancelled) {
+            throw new Error('CANCELLED');
+        }
+        const filledTakerTokenAmount = await this._orderFilledCancelledFetcher.getFilledTakerAmountAsync(orderHash);
+        if (signedOrder.takerAssetAmount.eq(filledTakerTokenAmount)) {
+            throw new Error('FULLY_FILLED');
```
LogvinovLeon: We shouldn't have errors as strings here. They should be a part of some enum. That way we can display them in the docs and test for them

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -198,10 +208,9 @@ export class OrderValidationUtils {
             throw new Error(OrderError.InvalidSignature);
         }
         const filledTakerTokenAmount = await this._orderFilledCancelledFetcher.getFilledTakerAmountAsync(orderHash);
-        OrderValidationUtils._validateRemainingFillAmountNotZeroOrThrow(
-            signedOrder.takerAssetAmount,
-            filledTakerTokenAmount,
-        );
+        if (signedOrder.takerAssetAmount.eq(filledTakerTokenAmount)) {
+            throw new Error(RevertReason.OrderUnfillable);
```
LogvinovLeon: Let's not throw `OrderUnfillable`. It's cryptic. How about `REMAINING_FILL_AMOUNT_ZERO`?

packages/order-utils/src/order_validation_utils.ts
fabioberger: @dekz in a later PR this entire class is going to be removed. Don't think it's worth making this change.

packages/order-utils/src/order_validation_utils.ts
fabioberger: Ditto above.

packages/order-utils/src/store/order_filled_cancelled_lazy_store.ts
```diff
@@ -58,9 +60,10 @@ export class OrderFilledCancelledLazyStore implements AbstractOrderFilledCancell
      * @param orderHash OrderHash from order of interest
      * @return Whether the order has been cancelled
      */
-    public async getIsCancelledAsync(orderHash: string): Promise<boolean> {
+    public async getIsCancelledAsync(signedOrder: SignedOrder): Promise<boolean> {
```
LogvinovLeon: Is this a new convention? That no function should accept a hash but should recompute it on it's own?

packages/order-utils/src/order_validation_utils.ts
fabioberger: Yes, ideally they would be part of the `RevertReason` enum, however right now they are part of the `OrderStatus` enum, but this method does not throw all the values in that enum. Since this will get removed soon, I think we can deal with this inconsistency for a bit longer.

packages/order-utils/src/order_validation_utils.ts
fabioberger: This method is _only_ used by our combinatorial tests and as such they must mimick the behavior of the actual contract. The validation method exposed in `@0x/contract-wrappers` does not throw this cryptic error.

packages/order-utils/src/store/order_filled_cancelled_lazy_store.ts
fabioberger: No, it is not. The reason we need it is because `isOrderCancelledAsync` in `OrderFilledCancelledFetcher` needs not only the OrderHash, but also the order's `makerAddress` and `signerAddress` to check both the `cancelled` and `cancelledOrdersUpTo` mappings, and so we just pass in the order.

PR #1237, opened 2018-11-09
[instant] Heartbeats for account info and buy quotes

steveklebanoff:
## Description

Introduces generic heartbeat, and sets up heartbeat for updating account info and fetching buy quotes.

## Testing instructions

Watch redux for heartbeat actions for account info and buy quotes

## Checklist:


packages/instant/src/util/buy_quote_fetcher.ts
```diff
@@ -0,0 +1,73 @@
...
+import { oc } from 'ts-optchain';
+
+import { Action, actions } from '../redux/actions';
+import { State } from '../redux/reducer';
+import { AffiliateInfo, ERC20Asset } from '../types';
+import { assetUtils } from '../util/asset';
+
+import { errorFlasher } from './error_flasher';
+
+export const updateBuyQuoteOrFlashErrorAsync = async (
```
steveklebanoff: Need to update to use https://github.com/0xProject/0x-monorepo/blob/development/packages/instant/src/util/buy_quote_updater.ts instead

steveklebanoff:
Going to merge development in and do some fixups, review can wait until then

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
```diff
@@ -69,7 +69,7 @@ const mapStateToProps = (state: State, _ownProps: SelectedERC20AssetAmountInputP
 
 const debouncedUpdateBuyQuoteAsync = _.debounce(buyQuoteUpdater.updateBuyQuoteAsync.bind(buyQuoteUpdater), 200, {
     trailing: true,
-});
+}) as typeof buyQuoteUpdater.updateBuyQuoteAsync;
```
steveklebanoff: without this, `debouncedUpdateBuyQuoteAsync` was `any` 😱

packages/instant/src/redux/async_data.ts
```diff
@@ -36,10 +36,11 @@ export const asyncData = {
             store.dispatch(actions.setAvailableAssets([]));
         }
     },
-    fetchAccountInfoAndDispatchToStore: async (store: Store) => {
+    fetchAccountInfoAndDispatchToStore: async (options: { store: Store; setLoading: boolean }) => {
```
steveklebanoff: I changed to this syntax because for this and other methods, sending in booleans w/o this "psuedo-named parameter" approach felt indirect -- I found myself always looking up the function that was being invoked to understand what the boolean parameter represented.  With this approach, you can tell by the key of the object what the caller's intention is

packages/instant/src/redux/async_data.ts
```diff
@@ -74,19 +75,22 @@ export const asyncData = {
...
-    fetchCurrentBuyQuoteAndDispatchToStore: async (store: Store) => {
-        const { providerState, selectedAsset, selectedAssetAmount, affiliateInfo } = store.getState();
+    fetchCurrentBuyQuoteAndDispatchToStore: async (options: { store: Store; setPending: boolean }) => {
+        const { store, setPending } = options;
+        const { buyOrderState, providerState, selectedAsset, selectedAssetAmount, affiliateInfo } = store.getState();
         const assetBuyer = providerState.assetBuyer;
         if (
             !_.isUndefined(selectedAssetAmount) &&
             !_.isUndefined(selectedAsset) &&
+            buyOrderState.processState === OrderProcessState.None &&
```
steveklebanoff: Making sure we don't update the buy quote if they are in the middle of a purchase

steveklebanoff:
Merged in development and fixed this branch up, this is now ready for review

packages/instant/src/constants.ts
```diff
@@ -11,6 +11,8 @@ export const WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG_PREFIX = 'Transaction fa
 export const GWEI_IN_WEI = new BigNumber(1000000000);
 export const ONE_SECOND_MS = 1000;
 export const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
+export const ACCOUNT_UPDATE_INTERVAL_TIME_MS = ONE_SECOND_MS * 15;
+export const BUY_QUOTE_UPDATE_INTERVAL_TIME_MS = ONE_SECOND_MS * 15;
```
steveklebanoff: open to ideas for best numbers for these

BMillman19: I think we might want to go lower for the account, possibly 5s? This is what we have on the website now. That way we can better respond to changes in metamask (i.e. user changes address)

packages/instant/src/containers/selected_erc20_asset_amount_input.ts
BMillman19: 🙀

packages/instant/src/util/heartbeater.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (!_.isUndefined(this._intervalId)) {
+            throw new Error('Heartbeat is running, please stop before restarting');
+        }
+
+        if (this._performImmediatelyOnStart) {
+            // tslint:disable-next-line:no-floating-promises
+            this._trackAndPerformAsync();
+        }
+
+        this._intervalId = window.setInterval(this._trackAndPerformAsync.bind(this), intervalTimeMs);
```
BMillman19: we can avoid the bind by making `_trackAndPerformAsync` an arrow function

packages/instant/src/util/heartbeater.ts
```diff
@@ -0,0 +1,49 @@
...
+type HeartbeatableFunction = () => Promise<void>;
+export class Heartbeater {
+    private _intervalId?: number;
+    private _hasPendingRequest: boolean;
+    private readonly _performImmediatelyOnStart: boolean;
+    private readonly _performFunction: HeartbeatableFunction;
+
+    public constructor(performingFunctionAsync: HeartbeatableFunction, performImmediatelyOnStart: boolean) {
+        this._performFunction = performingFunctionAsync;
+        this._hasPendingRequest = false;
```
BMillman19: we have `intervalUtils.setAsyncExcludingInterval` in `@0x/utils` that takes care of this concurrent request handling

packages/instant/src/util/heartbeater_factory.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { Heartbeater } from './heartbeater';
+
+export interface HeartbeatFactoryOptions {
+    store: Store;
+    shouldPerformImmediatelyOnStart: boolean;
+}
+export const generateAccountHeartbeater = (options: HeartbeatFactoryOptions): Heartbeater => {
+    const { store, shouldPerformImmediatelyOnStart } = options;
+    return new Heartbeater(async () => {
```
BMillman19: There's a small bug in my branch that i fixed that will have a conflict here. but we should not fetch account info if `state.providerState.account.state === AccountState.None` perhaps this check belongs in the provider component and we just don't start the heartbeater, or this function returns undefined in that case and isolate reach into the state in here

packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -37,6 +40,9 @@ export interface ZeroExInstantProviderOptionalProps {
 
 export class ZeroExInstantProvider extends React.Component<ZeroExInstantProviderProps> {
     private readonly _store: Store;
+    private _accountUpdateHeartbeat?: Heartbeater;
```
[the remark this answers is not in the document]
steveklebanoff: We are assigning these in `componentDidMount`, and not in the class's `constructor`, so it looks like we need these to be optional unfortunately

packages/instant/src/util/heartbeater.ts
steveklebanoff: Ah this is nice!  Wish I knew this before writing up my own though.  When possible, we should try to bring up any knowledge like this when "grooming" stories

PR #1242, opened 2018-11-09
[instant] Install/Unlock MetaMask, connect PaymentDropdown to redux state

fragosti:
## Description

Apologies for the big PR. 

Changes:
- Connect `ZeroExInstantContainer` to store such that it can decide between the payment dropdown or the progress bar. 
- Add some progress bar code to export `ProgressBar` (instead of only `TimedProgressBar`) such that we have a not-undefined state for a progress bar that is not processing an order (it's just empty). Ie. make it possible to display a non-loading progress bar. 
- Connect `PaymentMethod` to the store such that it can handle all account states and open the standard sliding panel (below).
- Create `StandardSlidingPanel` and add `standardSlidingPanelSettings` to state such that we can always show panels (useful for errors). 
- Create `StandardPanelContent` as the template for content that shows up in the panel. 
- Create `InstallWalletPanelContent` based on the above ^
- Add `href?` prop to both `Button` and `Text`.
- Change default padding of `Button`.

screen shot 2018-11-09 at 3 29 28 pm
screen shot 2018-11-09 at 3 39 03 pm

## Testing instructions

Since there is a bug with the account setting code, the state the account is "Locked" when MetaMask is not installed.

## Checklist:


packages/instant/src/components/buy_order_progress.tsx
```diff
@@ -21,15 +22,25 @@ export const BuyOrderProgress: React.StatelessComponent<BuyOrderProgressProps> =
...
+            <React.Fragment>
                 <Container marginBottom="5px">
                     <TimeCounter estimatedTimeMs={expectedTimeMs} hasEnded={hasEnded} key={progress.startTimeUnix} />
                 </Container>
                 <TimedProgressBar expectedTimeMs={expectedTimeMs} hasEnded={hasEnded} key={progress.startTimeUnix} />
+            </React.Fragment>
+        );
+    } else {
+        // Just show a static progress bar if we aren't processing or in an end state
+        content = (
```
steveklebanoff: Per slack convo, this feels a bit unnecessary, esp. since we never actually hit this scenario

packages/instant/src/components/standard_sliding_panel.tsx
```diff
@@ -0,0 +1,29 @@
...
+            <SlidingPanel animationState={animationState} onClose={onClose}>
+                {this._getNodeForContent(content)}
+            </SlidingPanel>
+        );
+    }
+    private readonly _getNodeForContent = (content: StandardSlidingPanelContent): React.ReactNode => {
+        switch (content) {
+            case StandardSlidingPanelContent.InstallWallet:
+                return <InstallWalletPanelContent />;
+            default:
```
steveklebanoff: I'd like to get rid of this `default:` to ensure that if we add a new `StandardSlidingPanelContent.` there is a proper React.ReactNode returned here. Thoughts?

packages/instant/src/components/standard_sliding_panel.tsx
```diff
@@ -0,0 +1,29 @@
...
+        return (
+            <SlidingPanel animationState={animationState} onClose={onClose}>
+                {this._getNodeForContent(content)}
+            </SlidingPanel>
+        );
+    }
+    private readonly _getNodeForContent = (content: StandardSlidingPanelContent): React.ReactNode => {
+        switch (content) {
+            case StandardSlidingPanelContent.InstallWallet:
+                return <InstallWalletPanelContent />;
```
steveklebanoff: At first glance this felt a little over-engineered to me -- why not just take the content as a child, instead of sending in a `StandardSlidingPanelContent` and rendering based on that here.  But given how the code works with dispatching the sliding in and out of these, this makes sense

packages/instant/src/components/payment_method.tsx
```diff
@@ -1,45 +1,139 @@
...
+            );
+        }
+        return null;
+    };
+    private readonly _renderMainContent = (): React.ReactNode => {
+        const { account, network } = this.props;
+        switch (account.state) {
+            case AccountState.Loading:
+                // Just take up the same amount of space as the other states.
+                return <Container height="52px" />;
```
steveklebanoff: Is there anyway to avoid this magic number?  OK if not

packages/instant/src/components/payment_method.tsx
```diff
@@ -1,45 +1,139 @@
...
+            case AccountState.Ready:
+                return (
+                    <PaymentMethodDropdown
+                        accountAddress={account.address}
+                        accountEthBalanceInWei={account.ethBalanceInWei}
+                        network={network}
+                    />
+                );
+            default:
+                return null;
```
steveklebanoff: Same comment about wanting to take out the `default:` to have typesafety ensure we are explicit about handling every `AccountState`

packages/instant/src/components/payment_method.tsx
```diff
@@ -1,45 +1,139 @@
...
-            network={Network.Mainnet}
-        />
+                </React.Fragment>
+            );
+        }
+        return null;
+    };
+    private readonly _renderMainContent = (): React.ReactNode => {
+        const { account, network } = this.props;
+        switch (account.state) {
```
steveklebanoff: nice clear logic here 💯

packages/instant/src/components/buy_button.tsx
```diff
@@ -43,7 +43,7 @@ export class BuyButton extends React.Component<BuyButtonProps> {
                 onClick={this._handleClick}
                 isDisabled={shouldDisableButton}
                 fontColor={ColorOption.white}
-                fontSize="20px"
+                fontSize="16px"
```
steveklebanoff: Since we went from 20 -> 16 in a few places, I'm wondering if we should introduce this number as a constant so we only have to change it one place in the future?  Not a huge deal tho

packages/instant/src/util/util.ts
```diff
@@ -2,4 +2,5 @@ import * as _ from 'lodash';
 
 export const util = {
     boundNoop: _.noop.bind(_),
+    createHrefOnClick: (href: string) => () => window.open(href, '_blank'),
```
steveklebanoff: nit but this feels like a weird name: `createHrefOnClick` makes me think it will generate an `a` tag or something.  Really this is just something like `openUrlInNewWindow`, and `url` feels like a more natural parameter name than `href` here

packages/instant/src/components/ui/button.tsx
```diff
@@ -12,15 +15,26 @@ export interface ButtonProps {
     padding?: string;
     type?: string;
     isDisabled?: boolean;
-    onClick?: (event: React.MouseEvent<HTMLElement>) => void;
+    href?: string;
+    onClick?: ButtonOnClickHandler;
```
steveklebanoff: nit but would be nice for the type to ensure that we don't define both an `href` and an `onclick`

packages/instant/src/containers/connected_account_payment_method.ts
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
```
steveklebanoff: o0o I like the "connected_" prefix

packages/instant/src/containers/connected_zero_ex_instant_container.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
```
steveklebanoff: I don't love adding another layer of indirection to get to the meat of the `ZeroExInstantContainer` here.  What do you think about putting the `_renderPaymentMethodOrBuyOrderProgress` code into its own connected component, instead of creating this `ConnectedZeroExInstantContainer` here? 

 If we did this, I'd be OK with the `_renderPaymentMethodOrBuyOrderProgress` just going into the container

packages/instant/src/constants.ts
```diff
@@ -17,6 +17,10 @@ export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
 export const PROGRESS_STALL_AT_WIDTH = '95%';
 export const PROGRESS_FINISH_ANIMATION_TIME_MS = 200;
+export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
```
steveklebanoff: It doesn't look like this is ever used, perhaps a relic of old code?

packages/instant/src/containers/connected_account_payment_method.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export interface ConnectedAccountPaymentMethodProps {}
+
+interface ConnectedState {
+    account: Account;
+    network: Network;
+}
+
+interface ConnectedDispatch {
+    openStandardSlidingPanel: (content: StandardSlidingPanelContent) => void;
```
steveklebanoff: Want to start using the pattern of `Pick`-ing `openStandardSlidingPanel` from `PaymentMethod` here?

packages/instant/src/components/standard_sliding_panel.tsx
fragosti: Yea we have these defaults all over the place because of the tslint config. However, I think for instant it makes sense to turn this off since it is useful not to specify a default in certain cases. The only gotcha is that if we have a value coming externally (ie. non typescript-land) we need to consider that and include the default. However, internally it is super useful to switch on typescript enums without providing a default so went ahead and did that.

packages/instant/src/components/standard_sliding_panel.tsx
fragosti: I thought about that. Some reasons I went this direction:
1. Would rather have redux actions be pure data when possible.
2. Imagined that all future standard panels would want access to `onClose` and it would be a pain to have to connect them.

packages/instant/src/components/ui/button.tsx
fragosti: I played around with this and couldn't find a solution I loved. Should `onClick` also be able to take in a `string`?

packages/instant/src/containers/connected_zero_ex_instant_container.ts
fragosti: I feel like it's ok for this logic to be in here? It seems overkill to create another component for this.

packages/instant/src/components/buy_order_progress.tsx
```diff
@@ -1,3 +1,4 @@
+import * as _ from 'lodash';
```
BMillman19: is this fixing a regression?

packages/instant/src/components/scaling_input.tsx
```diff
@@ -156,8 +156,6 @@ export class ScalingInput extends React.Component<ScalingInputProps, ScalingInpu
                     return `${width}px`;
                 }
                 return `${textLengthThreshold}ch`;
-            default:
```
BMillman19: Is this fixing something? I noticed that sometimes the text in the input jumps around a bit, is this related?

packages/instant/src/components/standard_panel_content.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+export interface StandardPanelContentProps {
+    image: React.ReactNode;
+    title: string;
+    description: string;
+    moreInfoSettings?: MoreInfoSettings;
+    action: React.ReactNode;
+}
+
+const spacingBetweenPx = '20px';
```
BMillman19: This is different than how we usually format constants. We would usually do `SPACING_BETWEEN_PX`

packages/instant/src/components/zero_ex_instant_overlay.tsx
```diff
@@ -31,7 +31,7 @@ export const ZeroExInstantOverlay: React.StatelessComponent<ZeroExInstantOverlay
                         />
                     </Container>
                     <Container width={{ default: 'auto', sm: '100%' }} height={{ default: 'auto', sm: '100%' }}>
-                        <ZeroExInstantContainer />
+                        <ConnectedZeroExInstantContainer />
```
BMillman19: Should we update all connected components / container to follow this convention?

packages/instant/src/data/asset_meta_data_map.ts
```diff
@@ -8,42 +8,42 @@ export const assetMetaDataMap: ObjectMap<AssetMetaData> = {
     '0xf47261b0000000000000000000000000e41d2489571d322189246dafa5ebde1f4699f498': {
         assetProxyId: AssetProxyId.ERC20,
         decimals: 18,
-        primaryColor: 'rgb(54, 50, 60)',
+        primaryColor: '#333333',
```
BMillman19: This is part of my branch

packages/instant/src/components/buy_order_progress.tsx
fragosti: hmmm I saw this but assumed running `yarn lint --fix` would fix it. Clearly it doesn't catch this case.

packages/instant/src/components/scaling_input.tsx
fragosti: This is related to @steveklebanoff's comment that we shouldn't be always setting a default case for switch statements. I've not noticed the text jumping around.

packages/instant/src/components/zero_ex_instant_overlay.tsx
fragosti: I feel like I've gotten some feedback on the naming here from @steveklebanoff as well so might be good to explain my reasoning. 

Typically, `connected_` is the catch-all prefix you could give to a connected component, but when possible I like to be more specific and specify what it's connected to, especially when you could feasibly have the same component connected to two different parts of the store. For example, `SelectedAssetBuyOrderProgress` is a progress bar that tracks the progress of the buy on `selecteedAsset`, but could feasibly be used to track any buy order.

Let me know if this is too confusing and you think we should just stick with `connected_`

(in `connected_account_payment_method` I mean "the account you connected to")

packages/instant/src/containers/connected_zero_ex_instant_container.ts
steveklebanoff: I still am not a huge fan of this -- I felt like the existing `ZeroExInstantContainer` was nicer, because the separation of concerns felt more clear and straight forward to me:  ZeroExInstantContainer` was the big parent container, and it is just responsible for setting up the hierarchy of its child components, and any connecting is done within the child components.  

Also, for a new developer, it feels indirect to navigate to the "meat" of the instant markup.  You go `zero_ex_instant` -> `connected_zero_ex_instant_container` -> `zero_ex_instant_container`.

packages/instant/src/components/zero_ex_instant_overlay.tsx
steveklebanoff: I prefer just calling everything `connected_`, as names like `SelectedAssetBuyOrderProgress` feel overly verbose to me, but I understand your reasoning and am cool w/ it

packages/instant/src/components/standard_sliding_panel.tsx
steveklebanoff: Awesome, thx!

packages/instant/src/components/ui/button.tsx
steveklebanoff: `onClick` doing double duty of being able to take a string or function feels a little funky, i'm ok w/ this as-is

packages/instant/src/containers/connected_zero_ex_instant_container.ts
steveklebanoff: This is resolved w/ https://github.com/0xProject/0x-monorepo/pull/1255

PR #1252, opened 2018-11-13
[asset-buyer][instant] Fix incorrect token prices for non 18-decimal tokens

BMillman19:
## Description

- Update `AssetBuyer` to return information about total eth spent on the target asset instead of per-unit token prices. 
- Update `AssetBuyer` to include eth spent on zrx tokens to fund takerFees as part of the `feeEthAmount` returned by `getBuyQuoteAsync`. Previously this cost was being reported as part of the base amount being paid for an asset and the `feeEthAmount` represented only the affiliate fee.
- Update `Instant` to calculate per-unit token prices using the amount of units that is provided by the user
- Update variable names in `Instant` to be clearer what amounts represent normal units vs base units

## Testing instructions

Unit tests

## Types of changes

## Checklist:


packages/instant/src/containers/latest_buy_quote_order_details.ts
```diff
@@ -14,13 +14,15 @@ export interface LatestBuyQuoteOrderDetailsProps {}
 
 interface ConnectedState {
     buyQuoteInfo?: BuyQuoteInfo;
+    selectedAssetAmount?: BigNumber;
```
steveklebanoff: Per convo, use word "unit" to describe selected asest amount

packages/asset-buyer/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.3.0",
```
steveklebanoff: Should this be a full version bump (3)?

PR #1254, opened 2018-11-13
feat(order_utils.py): sign_hash()

feuGeneA:
Implemented `zero_ex.order_utils.sign_hash()`.

Also moved is_valid_signature() into main package module, for simplicity.

Also consolidated a handul of in-line pylint disable directives into the .pylintrc config file.

## Testing instructions

`./setup.py test`
`./setup.py lint`

Both of which are integrated with CI.

## Types of changes

## Checklist:


python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -159,7 +182,209 @@ def int_to_32_big_endian_bytes(i: int):
...
+    ...     '0x34decbedc118904df65f379a175bb39ca18209d6ce41d5ed549d54e6e0a95004',
+    ... )
+    '0x1b117902c86dfb95fe0d1badd983ee166ad259b27acb220174cbb4460d872871137feabdfe76e05924b484789f79af4ee7fa29ec006cedce1bbf369320d034e10b03'
+    """  # noqa: E501 (line too long)
+    # TODO: make this provider check more flexible.
+    # https://app.asana.com/0/684263176955174/901300863045491/f
+    if not isinstance(provider, Web3.HTTPProvider):
+        raise TypeError("provider is not a Web3.HTTPProvider")
+    if not is_address(signer_address):
+        raise ValueError("signer_address is not a valid address")
```
fabioberger: Let's wrap this into `asset_is_address()` function

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -159,7 +182,209 @@ def int_to_32_big_endian_bytes(i: int):
...
+    ...     Web3.HTTPProvider("http://127.0.0.1:8545"),
+    ...     Web3(Web3.HTTPProvider("http://127.0.0.1:8545")).personal.listAccounts[0],
+    ...     '0x34decbedc118904df65f379a175bb39ca18209d6ce41d5ed549d54e6e0a95004',
+    ... )
+    '0x1b117902c86dfb95fe0d1badd983ee166ad259b27acb220174cbb4460d872871137feabdfe76e05924b484789f79af4ee7fa29ec006cedce1bbf369320d034e10b03'
+    """  # noqa: E501 (line too long)
+    # TODO: make this provider check more flexible.
+    # https://app.asana.com/0/684263176955174/901300863045491/f
+    if not isinstance(provider, Web3.HTTPProvider):
+        raise TypeError("provider is not a Web3.HTTPProvider")
```
fabioberger: Let's wrap this into a `assert_is_provider()` function

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -159,7 +182,209 @@ def int_to_32_big_endian_bytes(i: int):
...
+        )
+        (valid, _) = is_valid_signature(
+            provider, order_hash_hex, signature_as_vrst_hex, signer_address
+        )
+
+        if valid is True:
+            return signature_as_vrst_hex
+
+    raise RuntimeError(
+        "Signature returned from web3 provider is in an unknown format."
```
fabioberger: Let's add: `Attempted parsing as RSV and VRS.`

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -10,19 +10,22 @@
...
 from web3 import Web3
-from web3.utils import datatypes
 import web3.exceptions
+from web3.utils import datatypes
 
+from zero_ex.dev_utils.type_assertions import assert_is_hex_string
 
-class Constants:  # pylint: disable=too-few-public-methods
+
+class _Constants:
```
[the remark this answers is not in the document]
feuGeneA: It implies that it's "private/internal."  More concretely, though, it means that the automatic documentation generation will not include this class in the doc page.

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -159,7 +182,209 @@ def int_to_32_big_endian_bytes(i: int):
...
+
+    :param provider: A Web3 provider.
+    :param signer_address: The address of the signing account.
+    :param order_hash_hex: A hex string representing the order hash, like that
+        returned from `generate_order_hash_hex()`.
+    :rtype: A string, of ASCII hex digits, representing the signature.
+
+    >>> sign_order_hash(
+    ...     Web3.HTTPProvider("http://127.0.0.1:8545"),
+    ...     Web3(Web3.HTTPProvider("http://127.0.0.1:8545")).personal.listAccounts[0],
```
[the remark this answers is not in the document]
feuGeneA: Thanks for noticing this.  It was indeed weird.  I had tried to do this once and it didn't work but I'm glad I tried again because now it's working great and I'm happy with it.

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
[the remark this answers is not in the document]
feuGeneA: I went with a class because member variables that are initialized in the declaration are considered "static", meaning that they're only initialized once.  (So, for us, the json artifacts are only read from disk once, at declaration time.)  I went back and forth on whether to call the class `_StaticData` instead.  Thinking more, maybe it's appropriate to have a `class Constants` (no underscore, so publicly documented) with things like `contract_name_to_abi` and `null_address`, things that other people might want to use; and to have a separate `class _StaticData` (so "private," not documented) that contains things that are only used internally, like precomputed keccak hashes and stuff.  Just brainstorming...  Dunno how much people would re-use that stuff right now, and in the future we'll probably have separate public packages to more properly expose things like ABI's and contract addresses.

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -159,7 +187,202 @@ def int_to_32_big_endian_bytes(i: int):
...
+        + signature["v"].to_bytes(1, byteorder="big").hex()
+        + signature["r"]
+        + signature["s"]
+    )
+
+
+def sign_hash(
+    provider: BaseProvider, signer_address: str, order_hash_hex: str
+) -> str:
+    """Sign a message with the order hash, and return the signature.
```
fabioberger: Update so it doesn't mention `orderHash` definitively. It should be an example of the type of hash one could sign.

PR #1255, opened 2018-11-13
[instant] Fix input bug

steveklebanoff:
## Description

⚠️ This PR is into Francesco's branch https://github.com/0xProject/0x-monorepo/pull/1242 ⚠️ 

Remove ConnectedZeroExInstantContainer, introduce `ConnectedBuyOrderProgressOrPaymentMethod`.  This fixes the issue with ".005" input.

Not 100% sure why this fixes the input issue, but I suggested this pattern even before I knew it would fix the regression: https://github.com/0xProject/0x-monorepo/pull/1242#discussion_r232431188

## Testing instructions

Type in ".005", watch it work

## Checklist:


packages/instant/src/containers/connected_buy_order_progress_or_payment_method.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { State } from '../redux/reducer';
+import { OrderProcessState } from '../types';
+
+import { ConnectedAccountPaymentMethod } from './connected_account_payment_method';
+import { SelectedAssetBuyOrderProgress } from './selected_asset_buy_order_progress';
+
+interface BuyOrderProgressOrPaymentMethodProps {
+    orderProcessState: OrderProcessState;
+}
+export const BuyOrderProgressOrPaymentMethod = (props: BuyOrderProgressOrPaymentMethodProps) => {
```
steveklebanoff: I think this a scenario where it's OK to break the pattern of putting this component into it's own file, since it's just switching on a variable to decide which component to render

steveklebanoff:
My guess is that without this fix, the ZeroExInstantContainer was re-rendering too aggressively when `OrderProcessState` changed, and this caused the input bug.  The mobile performance feels better on mobile too w/ this change.  Not perfect, but not nearly as laggy as it was before, might be 'good enough' for now.

PR #1266, opened 2018-11-15
[instant] Fix mobile specific content

fragosti:
## Description
- Fix issue where we were using wrong version of `bowser`.
- Design feedback: make prompt blue. 
- Design feedback: skip panel altogether for mobile, so remove that panel and don't show it for mobile -> go straight to ios / google store. 

screen shot 2018-11-14 at 4 30 26 pm

## Checklist:


packages/instant/src/components/payment_method.tsx
```diff
@@ -107,4 +116,24 @@ export class PaymentMethod extends React.Component<PaymentMethodProps> {
                 );
         }
     };
+    private readonly _handleInstallWalletClick = (): void => {
```
steveklebanoff: This is feeling kind of unnatural to me -- triggering `onInstallWalletClick` only when we are on desktop.  

If we keep this name, I would think we would trigger something like `onInstallWalletClick(walletType: 'coinbaseWallet' | 'metamask')`.  

If we want to keep this pattern, I think we should rename this to something like `onInstallMetamaskClick` or `onInstallDesktopWalletClick`

packages/instant/src/containers/connected_account_payment_method.ts
```diff
@@ -49,7 +55,24 @@ const mergeProps = (
...
     walletName: connectedState.providerState.name,
     onUnlockWalletClick: () => {
-        connectedDispatch.unlockWalletAndDispatchToStore(connectedState.providerState);
+        const isMobile = envUtil.isMobileOperatingSystem();
+        if (!isMobile) {
+            connectedDispatch.unlockWalletAndDispatchToStore(connectedState.providerState);
+            return;
+        }
+        const operatingSystem = envUtil.getOperatingSystem();
+        let url = COINBASE_WALLET_SITE_URL;
```
steveklebanoff: it looks like we're doing this type of logic in 2 places -- can we introduce a helper ?

packages/instant/src/containers/connected_account_payment_method.ts
```diff
@@ -49,7 +55,24 @@ const mergeProps = (
     onInstallWalletClick: connectedDispatch.onInstallWalletClick,
     walletName: connectedState.providerState.name,
     onUnlockWalletClick: () => {
-        connectedDispatch.unlockWalletAndDispatchToStore(connectedState.providerState);
+        const isMobile = envUtil.isMobileOperatingSystem();
+        if (!isMobile) {
```
steveklebanoff: what's going on here? -- it looks like we're accounting for a situation in which we display the wallet as 'locked' if they are on mobile, even if they are not on a mobile dapp app like coinbase wallet?

packages/instant/src/components/payment_method.tsx
BMillman19: I think its fine since we already have mobile specific logic when rendering, I think we can either move all the mobile specific logic out and up into a higher component or keep it as is

fragosti: Oh I actually changes this.

packages/instant/src/containers/connected_account_payment_method.ts
fragosti: Whoops was in a rush when I wrote this -- this is wrong.

packages/instant/src/containers/connected_account_payment_method.ts
fragosti: This should be fixed now

packages/instant/src/components/install_wallet_panel_content.tsx
```diff
@@ -1,19 +1,15 @@
 import * as React from 'react';
```
steveklebanoff: Since this is no longer generic, we could rename this to be metamask specific.  Not a huge deal though

PR #1269, opened 2018-11-15
[WIP] Upgrade the BigNumber to a newest version

### LINKED ISSUE #1136: Old Version of bignumber.js
## Current Behavior

0x-monorepo/packages/types/ uses "bignumber.js": "~4.1.0". This is a very old version of the bignumber library and is not forward compatible with the latest bignumber.js "7.2.1". If a project has a dependency a more recent version, it wreaks havoc on the code.
## Possible Solution

Upgrade the types package to the latest bignumber.js.

LogvinovLeon:
## Description

Fixes #1136 

## Checklist:


LogvinovLeon:
After a discussion with @fabioberger We've decided to close this PR for now and go with a different approach. We'll describe it later in a design doc

WyseNynja:
Has that design doc been written yet? If so, can you share a link to it?

fabioberger:
@WyseNynja the design we've agreed on is similar to what ethers.js does. We will have our own "bignumber" type, which will be a wrapper around BigNumber.js. That way we can easily upgrade to future versions of BigNumber, and ever different BigNumber libraries by simply updating the wrapper. 

For additional reasons, check out this issue: https://github.com/ethers-io/ethers.js/issues/312

WyseNynja:
That sounds like a good plan. Thanks for the update.

PR #1270, opened 2018-11-15
[WIP] [sol-resolver] Fix *nix-like hardcoded paths in NPM Resolver

jvanoers:
## Description

The NPM resolver currently assumes a hardcoded root directory
of '/' and a path separator of '/'.

Currently it uses a while loop to recursively move up directories
from the current project working directory to the root directory,
searching for dependencies in possible node_modules locations.

The while loop breaks if 'trial directory === the hardcoded root'
i.e. trial as a result of moving up directories  === '/'

On windows machines this statement will never be true, as the root
path of the file system is 'X:/' (where X is any letter [A-Z]), thus this 
loop is infinitely recursive.

Additionally, it hardcodes a path separator of '/' which on Windows
is instead '\'.

This fix utilises Node's path library to gather the current system root 
directory and separator.

It uses 'path.parse(...).root' for the system root and 'path.sep'.

## Testing instructions

- Use `sol-compiler --contracts-dir contracts` on Windows and *nix
    - _Previous_: Observe recursive hang on Windows
    - _Now_: Contracts compile correctly on Windows

## Types of changes

## Checklist:


LogvinovLeon:
@jvanoers Why is this PR a WIP? What are the changes that still need to be made?

jvanoers:
@LogvinovLeon it's WIP because I was naive in thinking the problem with Unix-hardcoded paths was isolated to the `sol-resolver` package, for example `_recursivelyGatherDependencySources()` in `sol-compiler/src/compiler.ts` has problems on Windows too. I will change the scope of this issue/PR to include the additional resolution problems as I discover them.

I will eventually get around to ensuring Windows is supported by the compiler toolchain, but at the moment with time focused elsewhere I haven't been able to finish ensuring all the use/edge-cases of resolution are tested across the board.

Hopefully more to come soon!

LogvinovLeon:
I've removed the stale label cause I still want to look into that. I have one PR coming that will change some things in the resolver and I will give this one a look after that. Unless it's urgent for you.

jvanoers:
Much appreciated @LogvinovLeon ! 

To add some notes IIRC, I ended up resorting to modifying the NameResolver to act as a catch-all in the many path translations that happen in/out/around the solc compiler output. It was quite a mess to follow the trail of all of the path modifications in a non-trivial number of contracts/import paths, so I cheated a bit which worked in my case, but didn't end up working in all scenarios.

I also added in some logic to support absolute imports of the nature:
```
import 'zeppelin-solidity/contracts/token/ERC20/StandardToken.sol';
```

This led to a successful compilation process for my purposes (12 inter-related contracts with relative and absolute imports), however it couldn't compile the more complex 0x contracts due to some quirky path resolution errors when dealing with relative imports a few directories deep. I unfortunately ran out of time debugging this before needing to move on. It's a shame to say I haven't had an opportunity to go back to give it another go, so I hope the incoming PR you mention helps things along!

PR #1271, opened 2018-11-15
[instant] Add icons to token selector

BMillman19:
## Description

* Add a new `assets/icons` folder with token icons
* Update the webpack config to bundle images

## Testing instructions

## Types of changes

## Checklist:


packages/instant/src/components/erc20_token_selector.tsx
```diff
@@ -95,9 +96,13 @@ class TokenSelectorRow extends React.Component<TokenSelectorRowProps> {
                         <Container marginRight="10px">
                             <Circle diameter={30} rawColor={token.metaData.primaryColor}>
                                 <Flex height="100%">
-                                    <Text fontColor={ColorOption.white} fontSize="8px">
-                                        {displaySymbol}
-                                    </Text>
+                                    {!_.isUndefined(iconUrlIfExists) ? (
```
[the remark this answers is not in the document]
BMillman19: I understand where your coming from but this is the convention we use all over the codebase and I'd rather be consistent

packages/instant/src/components/erc20_token_selector.tsx
```diff
@@ -79,6 +79,7 @@ class TokenSelectorRow extends React.Component<TokenSelectorRowProps> {
     public render(): React.ReactNode {
         const { token } = this.props;
         const displaySymbol = assetUtils.bestNameForAsset(token);
+        const TokenIcon = require(`../assets/icons/${token.metaData.symbol}.svg`);
```
fragosti: apparently `import()` is a thing? Does that work?

Also this doesn't allow people to pass icons in which is kind of a bummer. Any way we can get that functionality?

BMillman19:
@fragosti the icons are converted into react components and bundled into the main bundle, they are all present with the first load: https://github.com/jhamlet/svg-react-loader

packages/instant/src/components/erc20_token_selector.tsx
BMillman19: hmm, it works with `require` but not `import`

fragosti: @BMillman19 what you said makes sense about `iconUrl`

PR #1272, opened 2018-11-15
[instant] Base heap integration

steveklebanoff:
## Description

Introduces base heap integration with instant.  Starts off with reporting on "Widget - Opened" and "Wallet - Ready", and sending in some details about network, where it was embedded from, and what the provider is.

There are some TODOs which will get resolved when we introduce the Rollbar branch which allows us to know if we are on a dev environment or not.  I made an Asana task for these.

## Testing instructions

- Check out this branch
- Log into Heap, chose "Instant - Development" project
- View live action tracking ( https://heapanalytics.com/app/live )
- Look for "Widget - Opened" event
- Watch for "Wallet - Ready" event (should only trigger on initial render and on changing wallet -- should not trigger w/ every heart beat)

## Checklist:


packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -120,6 +121,15 @@ export class ZeroExInstantProvider extends React.Component<ZeroExInstantProvider
         gasPriceEstimator.getGasInfoAsync();
         // tslint:disable-next-line:no-floating-promises
         this._flashErrorIfWrongNetwork();
+
+        // Analytics
+        analytics.addEventProperties({
```
steveklebanoff: Instead of sprinkling in `.track`s with arbitrary strings throughout the code base, I'm thinking it might make sense to have a centralized helper so we have all event names and properties in one place.  Does this sound reasonable, or like overkill?

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,85 @@
...
+                c < p.length;
+                c++
+            )
+                (window as any).heap[p[c]] = o(p[c]);
+        });
+    // TODO: use production heap id once environment utils merged
+    (window as any).heap.load(HEAP_ANALYTICS_DEVELOPMENT_APP_ID);
+    /* tslint:enable */
+
+    return curWindow.heap as HeapAnalytics;
```
fragosti: This might now work... did you see the heap GOTCHA in website?

``
    /**
     * Heap is not available as a UMD module, and additionally has the strange property of replacing itself with
     * a different object once it's loaded.
     * Instead of having an await call before every analytics use, we opt to have the awaiting logic in here by
     * guarding every API call with the guard below.
     */
    private async _heapLoadedGuardAsync(callback: () => void): Promise<void> {
        if (this._heap.loaded) {
            callback();
            return undefined;
        }
        await utils.onPageLoadPromise;
        // HACK: Reset heap to loaded heap
        this._heap = (window as any).heap;
        callback();
    }
```

File: packages/instant/src/components/zero_ex_instant_provider.tsx (hunk 1 above)
Comment (fragosti) [reply packages/instant/src/components/zero_ex_instant_provider.tsx]: Don't feel strongly about this

File: packages/instant/src/util/heap.ts (hunk 2 above)
Comment (fragosti) [reply packages/instant/src/util/heap.ts]: Never mind this works.

File: packages/instant/src/redux/analytics_middleware.ts [hunk 3]
```diff
@@ -0,0 +1,45 @@
...
+export const analyticsMiddleware: Middleware = store => next => middlewareAction => {
+    const prevState = store.getState() as State;
+    const prevAccount = prevState.providerState.account;
+
+    const nextAction = next(middlewareAction) as Action;
+
+    const curState = store.getState() as State;
+    const curAccount = curState.providerState.account;
+
+    switch (nextAction.type) {
```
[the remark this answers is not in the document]
steveklebanoff: Yes, this is just a POC to provide a framework for logging events, we will have a separate ticket(s) which we implement a bunch of events based on the quip document we're working on

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,85 @@
...
+
+const setupZeroExInstantHeap = () => {
+    const curWindow = getWindow();
+    // Set property to specify that this is zeroEx's heap
+    curWindow.zeroExInstantLoadedHeap = true;
+
+    // Typescript-compatible version of https://docs.heapanalytics.com/docs/installation
+    /* tslint:disable */
+    ((window as any).heap = (window as any).heap || []),
+        ((window as any).heap.load = function(e: any, t: any) {
```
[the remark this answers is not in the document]
steveklebanoff: was pretty quick -- felt a little dirty, but ultimately decided to this instead of `require`ing a plain JS file

packages/instant/src/components/zero_ex_instant_provider.tsx
steveklebanoff: After writing out a bunch of events we want to track in the quip doc, I think it'd be worthwhile to centralize all the event names and their properties -- I've seen this be helpful in past projects when dealing with a large amount of events.  I'll work on a commit

packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ETH_DECIMALS } from '../constants';
+import { AccountState } from '../types';
+import { analytics } from '../util/analytics';
+
+import { Action, ActionTypes } from './actions';
+
+import { State } from './reducer';
+
+export const analyticsMiddleware: Middleware = store => next => middlewareAction => {
+    const prevState = store.getState() as State;
```
BMillman19: prevState and curState are the same?

packages/instant/src/util/analytics.ts
```diff
@@ -0,0 +1,35 @@
...
+        return;
+    }
+
+    const curHeap = heapUtil.getHeap();
+    if (curHeap) {
+        try {
+            heapFunctionCall(curHeap);
+        } catch (e) {
+            // We never want analytics to crash our React component
+            // TODO: error reporter here
```
BMillman19: I've been trying to use `// TODO(bmillman): blah blah` as a convention

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,85 @@
...
+        });
+    // TODO: use production heap id once environment utils merged
+    (window as any).heap.load(HEAP_ANALYTICS_DEVELOPMENT_APP_ID);
+    /* tslint:enable */
+
+    return curWindow.heap as HeapAnalytics;
+};
+
+export const heapUtil = {
+    getHeap: (): HeapAnalytics | null => {
```
BMillman19: is there a specific reason you used null here instead of undefined? we always prefer to use undefined unless we have a null React element

BMillman19: `getHeapIfExists` ?

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+const setupZeroExInstantHeap = () => {
+    const curWindow = getWindow();
+    // Set property to specify that this is zeroEx's heap
+    curWindow.zeroExInstantLoadedHeap = true;
+
+    // Typescript-compatible version of https://docs.heapanalytics.com/docs/installation
+    /* tslint:disable */
+    ((window as any).heap = (window as any).heap || []),
```
BMillman19: why aren't we using curWindow here?

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,85 @@
...
+        ((window as any).heap.load = function(e: any, t: any) {
+            ((window as any).heap.appid = e), ((window as any).heap.config = t = t || {});
+            var r = t.forceSSL || 'https:' === (document.location as Location).protocol,
```
BMillman19: do we need to be using var's here?

packages/instant/src/util/heap.ts
BMillman19: why doesn't `load` appear in the `HeapAnalytics` interface definition above?

packages/instant/src/redux/analytics_middleware.ts
BMillman19: also what happens without the `as State` and `as Action`

packages/instant/src/components/zero_ex_instant_provider.tsx
BMillman19: we should at least have one file with all the eventNames as constants and maybe some way of typing the corresponding eventProperties

steveklebanoff: coo yea i agree, working on this

packages/instant/src/redux/analytics_middleware.ts
steveklebanoff: - it's a little funky, but calling `next(middlewareAction)` causes the getState() result to change.  I can add a comment for this
- without `as State` or `as Action` they are cast as `any`

packages/instant/src/util/heap.ts
steveklebanoff: no particular reason, will use undefined

steveklebanoff: hrm, this generates heap if it doesn't exist, so `getHeapIfExists` feels a little wrong

packages/instant/src/util/heap.ts
steveklebanoff: This is just a copy-paste of heap's integration code (https://docs.heapanalytics.com/docs/installation) with small mods to make TS compile it

packages/instant/src/util/heap.ts
steveklebanoff: This is just a copy-paste of heap's integration code (https://docs.heapanalytics.com/docs/installation) with small mods to make TS compile it

packages/instant/src/util/heap.ts
steveklebanoff: ah, I just copy-pasted the interface from website, and it was excluded there, likely b/c we don't actually call or use it

packages/instant/src/util/analytics.ts
```diff
@@ -0,0 +1,35 @@
+import { ObjectMap } from '@0x/types';
+import { logUtils } from '@0x/utils';
+
+import { ANALYTICS_ENABLED } from '../constants';
+
+import { HeapAnalytics, heapUtil } from './heap';
+
+const evaluteHeapCall = (heapFunctionCall: (heap: HeapAnalytics) => void): void => {
+    if (!ANALYTICS_ENABLED) {
```
steveklebanoff: per convo w/ @BMillman19 we should also guard against our heap being overwritten.  I could do this by checking the `appId` on the window's heap object

packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -120,6 +121,15 @@ export class ZeroExInstantProvider extends React.Component<ZeroExInstantProvider
...
+            embeddedHost: window.location.host,
+            embeddedUrl: window.location.href,
+            networkId: state.network,
+            providerName: state.providerState.name,
```
steveklebanoff: per convo w/ brandon, let's include the versino of instant here

BMillman19: as discussed offline, we need some kind of versioning here or else we will not be able to tell if the absence of an event is attributable to an old version or user inaction

packages/instant/src/components/zero_ex_instant_provider.tsx
steveklebanoff: this is done

packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -121,6 +122,17 @@ export class ZeroExInstantProvider extends React.Component<ZeroExInstantProvider
...
```
BMillman19: What does the providerName look like for fallback?

packages/instant/webpack.config.js
```diff
@@ -1,7 +1,15 @@
...
+const webpack = require('webpack');
+
 // The common js bundle (not this one) is built using tsc.
 // The umd bundle (this one) has a different entrypoint.
+
+const GIT_SHA = childProcess
+    .execSync('git rev-parse HEAD')
+    .toString()
+    .trim();
+const ip = require('ip');
```
BMillman19: why is this down here now?

packages/instant/webpack.config.js
```diff
@@ -10,6 +18,16 @@ const config = {
         library: 'zeroExInstant',
         libraryTarget: 'umd',
     },
+    plugins: [
+        new webpack.DefinePlugin({
+            'process.env': {
+                NODE_ENV: JSON.stringify(process.env.NODE_ENV),
```
BMillman19: Do we need this node_env line here? I recently removed this from the website because it was causing that error that fabio had mentioned? I believe webpack will take care of it for us

packages/instant/src/util/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+    npmVersion?: string;
+}
+export const analytics = {
+    addUserProperties: (properties: AnalyticsUserOptions): void => {
+        heapUtil.evaluateHeapCall(heap => heap.addUserProperties(properties));
+    },
+    addEventProperties: (properties: AnalyticsEventOptions): void => {
+        heapUtil.evaluateHeapCall(heap => heap.addEventProperties(properties));
+    },
+    walletOpened: trackingEventFnWithoutPayload(EventNames.WALLET_OPENED),
```
BMillman19: hm maybe these can be like `trackWalletOpened` or `walletOpenedEvent` to be more clear that these are tracking events vs adding properties like the functions above

packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    const curState = store.getState() as State;
+    const curAccount = curState.providerState.account;
+
+    switch (nextAction.type) {
+        case ActionTypes.SET_ACCOUNT_STATE_READY:
+            if (curAccount.state === AccountState.Ready && prevAccount.state !== AccountState.Ready) {
+                const ethAddress = curAccount.address;
+                analytics.addUserProperties({ ethAddress });
+                analytics.walletReady();
```
BMillman19: I think we want something here to track if a user changes their active address

packages/instant/src/constants.ts
```diff
@@ -16,6 +16,9 @@ export const BUY_QUOTE_UPDATE_INTERVAL_TIME_MS = ONE_SECOND_MS * 15;
 export const DEFAULT_GAS_PRICE = GWEI_IN_WEI.mul(6);
 export const DEFAULT_ESTIMATED_TRANSACTION_TIME_MS = ONE_MINUTE_MS * 2;
 export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
+export const ANALYTICS_ENABLED = process.env.NODE_ENV === 'production' || process.env.ENABLE_HEAP;
+export const HEAP_ANALYTICS_DEVELOPMENT_APP_ID = '507265531';
```
BMillman19: can we move these into our build process? If some one forks this, they could be logging garbage into our events right?

packages/instant/src/util/analytics.ts
```diff
@@ -0,0 +1,49 @@
...
+    walletReady: trackingEventFnWithoutPayload(EventNames.WALLET_READY),
+    widgetOpened: trackingEventFnWithoutPayload(EventNames.WIDGET_OPENED),
```
BMillman19: Can we change to `instantOpened` ?

packages/instant/src/util/analytics.ts
BMillman19: Also, what is the wallet opened event mean?

packages/instant/src/components/zero_ex_instant_provider.tsx
fragosti:
```
    getProviderName(provider: Provider): string {
        const providerTypeIfExists = envUtil.getProviderType(provider);
        if (_.isUndefined(providerTypeIfExists)) {
            return provider.constructor.name;
        }
        return PROVIDER_TYPE_TO_NAME[providerTypeIfExists];
    },
```

packages/instant/src/util/analytics.ts
```diff
@@ -0,0 +1,49 @@
+import { ObjectMap } from '@0x/types';
+
+import { heapUtil } from './heap';
+
+enum EventNames {
```
BMillman19: We maybe want something like name spacing for events so we can have events that are like `Wallet-Dropdown-Opened`  and we can enforce consistency for "open" events, that should make navigating the data easier down the road

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,115 @@
...
+                    c = 0;
+                c < p.length;
+                c++
+            )
+                (window as any).heap[p[c]] = o(p[c]);
+        });
+    (window as any).heap.load(getHeapAppId());
+    /* tslint:enable */
+
+    return curWindow.heap as HeapAnalytics;
```
fragosti: There is a bit of a race condition here. What I remember happening is that `heap` is initialized to this thing with the `load` method on it, and then you call load, and then that method will replace `heap` with something that does adhere to `HeapAnalytics` at some point in the future. However, in the mean-time, it does not adhere to the interface and only has the `loaded` boolean I believe(?).

packages/instant/src/util/heap.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        const zeroExInstantHeapIntegration = curWindow.zeroExInstantLoadedHeap && curWindow.heap;
+        if (zeroExInstantHeapIntegration) {
+            return zeroExInstantHeapIntegration;
+        }
+
+        return setupZeroExInstantHeap();
+    },
+    evaluateHeapCall: (heapFunctionCall: (heap: HeapAnalytics) => void): void => {
+        if (!ANALYTICS_ENABLED) {
```
steveklebanoff: per convo w/ Brandon - let's add an option for the integrator to be able to turn tracking off

packages/instant/webpack.config.js
steveklebanoff: Happened in a merge, fixed with 21ae0c4

packages/instant/webpack.config.js
steveklebanoff: Ah you are right that it looks like this isn't needed, good call. 

Just so I know though: can you specify which error and PR you are referencing?

packages/instant/src/util/provider_factory.ts
```diff
@@ -29,6 +29,10 @@ export const providerFactory = {
         providerEngine.addProvider(new RPCSubprovider(rpcUrl));
         // // Start the Provider Engine
         providerEngine.start();
+        // This feels a bit dirty, but was the only way I could think of
+        // checking to see if this engine is our fallback engine, and not
+        // another Web3Provider engine provided by some dapp browser
+        (providerEngine as any).zeroExInstantFallbackEngine = true;
```
steveklebanoff: I did this to ensure we are only reporting as the "Fallback" provider if the provider was constructed here.  This felt a bit dirty though.  

Is it possible that another dapp browser's constructor's name is "Web3ProviderEngine"? If so -- I think we need something like this to differentiate ours, unless someone can think of something better

packages/instant/webpack.config.js
BMillman19: https://github.com/0xProject/0x-monorepo/pull/1263

screen shot 2018-11-14 at 23 02 10

packages/instant/src/util/provider_factory.ts
BMillman19: hmm we can just set the name of the providerState in `providerStateFactory.getInitialProviderStateFallback` to some constant instead of calling `envUtil.getProviderName(provider)`

packages/instant/src/util/heap.ts
steveklebanoff: I think this is safe.  Here's what I observed from experimenting with this.

My code was as follows:

```js
<...heap integration snippet>
console.log("beforeLoad", heap.addEventProperties);
heap.load("YOUR_APP_ID");
console.log("afterLoad", heap.addEventProperties);
```

The last `console.log` was executed immediately after the `.load` call was made, and before heap actually loaded their full JS.  The output was as follows:

image

So it looks like it's pretty smart, which is what I would expect from an analytics snippet like this: immediately after `.load` it makes the public functions available: they push the data into an array which is later consumed by the full script once it loads from the CDN.

Let me know if I'm missing something

Source: https://gist.github.com/steveklebanoff/3de834cf8f8d2e30cceace4a87cf3aed

packages/instant/src/util/provider_factory.ts
steveklebanoff: Good call -- done in 097bfe5

packages/instant/src/util/heap.ts
steveklebanoff: This is done w/ 3f0d94c

packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -36,6 +37,7 @@ export interface ZeroExInstantProviderOptionalProps {
     additionalAssetMetaDataMap: ObjectMap<AssetMetaData>;
     networkId: Network;
     affiliateInfo: AffiliateInfo;
+    disableAnalyticsTracking: boolean;
```
steveklebanoff: If this name and approach seems good, I can add to the public spec

packages/instant/src/constants.ts
steveklebanoff: Per slack convo, I will move these into ENV vars

packages/instant/src/components/zero_ex_instant_provider.tsx
steveklebanoff: TODO: Need to add an `assert` to the render function for this

packages/instant/src/components/zero_ex_instant_provider.tsx
```diff
@@ -121,6 +123,20 @@ export class ZeroExInstantProvider extends React.Component<ZeroExInstantProvider
         gasPriceEstimator.getGasInfoAsync();
         // tslint:disable-next-line:no-floating-promises
         this._flashErrorIfWrongNetwork();
+
+        // Analytics
+        if (this.props.disableAnalyticsTracking) {
+            disableAnalytics();
```
steveklebanoff: TODO: Need to have this get reset if they call a subsequent render to enable analytics

packages/instant/src/components/zero_ex_instant_provider.tsx
steveklebanoff: This is done now

steveklebanoff:
This is updated and ready for another review

steveklebanoff:
@fabioberger I spoke with @BMillman19 about the ENV variable approach, and he mentioned it'd be good to get your :eyes: on this

Background: We want to log events to heap, and to do so we need an app id to configure the heap js with.  

Since it will be exposed to the browser anyway, we could just commit the production heap app id to the code base.  But it generally doesn't feel right to commit this kind of thing, and if someone forks and builds their own version, they could be sending heap data to us which may be garbage.  

So, we introduced this id being set by an environment variable.  We want to make sure that the published bundle always has a production heap app id set, so I started with throwing an error in `webpack.config.js` if you were creating a production build but didn't set the ENV variable.  But, this introduced some friction, because the ci step was failing because of this, *and* if you tried to build the entire monorepo it would error out here.

So, we removed the check from `webpack.config.js`, so ci and devs can build the package without the prod heap id.  Instead, we added an environment variable check in `prepublish_checks` that ensures that this env var is set.  

This feels like a good approach because it lets everyone build w/o hassle, but ensures that the published version has the env var set.  It'd be good to have your buy-in on the approach, as it will likely become a pattern for other integrations (i.e. rollbar).   Let us know what ya think.

fabioberger:
Yep, this sounds like a good approach to me! Thanks for roping me in. Does this mean that any dev that publishes our packages needs to have this environment set locally? If so, let's add the envs and instructions on adding them to dev environment to our Engineering onboarding doc.

steveklebanoff:
@fabioberger yes, per Slack convo: every dev who publishes will need this key set.  I will add it to the publishing doc and alert devs.

PR #1273, opened 2018-11-15
[instant] Don't show errors from heartbeat buy quote calls

steveklebanoff:
## Description

Don't trigger errors when buy quotes are updated from heartbeat

## Testing instructions

Type in large amount of asset, watch error come up .  Wait for heartbeat to trigger, and you should not see error pop up again.

## Checklist:

### DIFF
--- a/packages/instant/src/util/buy_quote_updater.ts
+++ b/packages/instant/src/util/buy_quote_updater.ts
@@ -16,8 +16,7 @@ export const buyQuoteUpdater = {
         dispatch: Dispatch<Action>,
         asset: ERC20Asset,
         assetUnitAmount: BigNumber,
-        options: { setPending: boolean; dispatchErrors: boolean },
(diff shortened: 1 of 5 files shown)


packages/instant/src/util/buy_quote_updater.ts
```diff
@@ -16,12 +16,12 @@ export const buyQuoteUpdater = {
         dispatch: Dispatch<Action>,
         asset: ERC20Asset,
         assetUnitAmount: BigNumber,
-        setPending = true,
+        options: { setPending: boolean; dispatchErrors: boolean },
         affiliateInfo?: AffiliateInfo,
```
fragosti: nit: `affiliateInfo` should also be part of options?

steveklebanoff: updated with 33a156d

PR #1274, opened 2018-11-15
chore: Add higher gas limit for global ganache

dekz:
## Description

Default in ganache is `6721975` which is below the current mainnet `8000000`. Running migrations with the `yarn ganache` in our top level directory will fail to deploy due to this limit.

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
         "packages/*"
     ],
     "scripts": {
-        "ganache": "ganache-cli -p 8545 --networkId 50 -m \"${npm_package_config_mnemonic}\"",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
         "packages/*"
     ],
     "scripts": {
-        "ganache": "ganache-cli -p 8545 --networkId 50 -m \"${npm_package_config_mnemonic}\"",
+        "ganache": "ganache-cli -p 8545 --networkId 50 --gasLimit 8000000 -m \"${npm_package_config_mnemonic}\"",
```
BMillman19: nit: perhaps it can go in the config below

PR #1280, opened 2018-11-16
[instant] Rollbar integration

steveklebanoff:
## Description

Report Instant errors to Rollbar

⚠️ There are a lot of commits in here, would like to squash to merge ⚠️ 

⚠️ Before Merge: ⚠️ 
- [ ] Figure out how to publish sourcemaps on npm publish
- [ ] Fix `test-publish` error -- not sure what's happening here, but may be related to the webpack config change?

## Testing instructions

- Add change to a react component to make it easier to trigger an error.  (i.e. Typing in a magic number to the input makes an error trigger)
- Deploy to dogfood
- Trigger error and see it reported to rollbar w/ source maps and the correct environment set

## Checklist:


packages/instant/.dogfood.discharge.json
```diff
@@ -1,12 +1,12 @@
...
     "domain": "0x-instant-dogfood",
-    "build_command": "WEBPACK_OUTPUT_PATH=public yarn build:umd:prod",
+    "build_command": "WEBPACK_OUTPUT_PATH=public yarn build:umd:prod --env.dogfood",
     "upload_directory": "public",
     "index_key": "index.html",
     "error_key": "index.html",
     "trailing_slashes": true,
     "cache": 3600,
-    "aws_profile": "default",
+    "aws_profile": "0xproject",
```
steveklebanoff: I like to be conservative and not have 0x credentials in my default profile -- I think it's good for folks to have a separate "0xproject" aws profile. Being this explicit could help prevent accidents when experimenting or prototyping w/ aws

packages/instant/webpack.config.js
```diff
@@ -1,42 +1,95 @@
...
+const GIT_SHA = childProcess
+    .execSync('git rev-parse HEAD')
+    .toString()
+    .trim();
+
+const ROLLBAR_ACCESS_KEY_ENV_NAME = 'ZEROEX_INSTANT_ROLLBAR_POST_ACCESS_KEY';
+function getPlugins(webpackEnv) {
+    const basePlugins = [
+        new webpack.DefinePlugin({
+            'process.env': {
```
steveklebanoff: Will have to update this to have the new env vars in https://github.com/0xProject/0x-monorepo/pull/1272

packages/instant/webpack.config.js
```diff
@@ -1,42 +1,95 @@
...
+        rollbarPublicPath = 'http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com';
+    } else if (webpackEnv.staging) {
+        rollbarPublicPath = 'http://0x-instant-staging.s3-website-us-east-1.amazonaws.com';
+    }
+
+    if (!rollbarPublicPath) {
+        return basePlugins;
+    }
+
+    const rollbarAccessKey = process.env[ROLLBAR_ACCESS_KEY_ENV_NAME];
```
steveklebanoff: I think it's good practice to avoid having the Rollbar access key with "post_server_item" permissions be public.  But, this means we will need to use this ENV variable when publishing to a remote server

packages/instant/webpack.config.js
```diff
@@ -1,42 +1,95 @@
...
-                    `webpack-dev-server can be accessed externally at: http://${ip.address()}:${config.devServer.port}`,
-                );
-            }
+        }),
+    ];
+
+    let rollbarPublicPath;
+    if (webpackEnv.dogfood) {
+        rollbarPublicPath = 'http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com';
+    } else if (webpackEnv.staging) {
```
steveklebanoff: ⚠️ TODO: Need to put an item in here for our production build ⚠️

steveklebanoff:
@BMillman19 you have a todo here: https://github.com/0xProject/0x-monorepo/blob/development/packages/asset-buyer/src/utils/order_provider_response_processor.ts#L63 . But, this is catching the error in asset buyer package, not in Instant.  We need to figure out how to report this to Rollbar -- would it make sense to surface the error somehow to Instant?

packages/instant/src/util/script_environment.ts
```diff
@@ -0,0 +1,60 @@
...
+        ];
+        for (const curEnvironment of allEnvironments) {
+            const domains = ENVIRONMENTS_TO_DOMAINS[curEnvironment];
+            if (_.some(domains, (domain: string) => url.includes(domain))) {
+                return curEnvironment;
+            }
+        }
+        return undefined;
+    },
+    getEnvironment: (defaltValue = Environment.Production): Environment => {
```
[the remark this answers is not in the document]
steveklebanoff: On second look, if we can just identify the environment in the build process (i.e. https://github.com/0xProject/0x-monorepo/pull/1272 ), we could use that instead of doing all of this logic to identify the environment

packages/instant/src/util/error_reporter.ts
```diff
@@ -0,0 +1,61 @@
...
+import { scriptEnvironment } from './script_environment';
+
+// Import version of Rollbar designed for embedded components
+// See https://docs.rollbar.com/docs/using-rollbarjs-inside-an-embedded-component
+// tslint:disable-next-line:no-var-requires
+const Rollbar = require('rollbar/dist/rollbar.noconflict.umd');
+
+const shouldReportErrors = () => {
+    const isNonDev = scriptEnvironment.getEnvironment() !== Environment.Development;
+    return isNonDev || process.env.FORCE_REPORT_ROLLBAR ? true : false;
```
BMillman19:
```suggestion
    return isNonDev || process.env.FORCE_REPORT_ROLLBAR;
```

packages/instant/src/constants.ts
```diff
@@ -19,6 +20,13 @@ export const ETH_GAS_STATION_API_BASE_URL = 'https://ethgasstation.info';
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
 export const PROGRESS_STALL_AT_WIDTH = '95%';
 export const PROGRESS_FINISH_ANIMATION_TIME_MS = 200;
+export const ROLLBAR_ACCESS_TOKEN = '61aed26a8c4d49aa9d10509a31a843d4';
```
BMillman19: do we want to inject this as a env variable as well?

packages/instant/src/util/script_environment.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export type EnvironmentToDomain = { [key in Environment]: string[] };
+
+export const scriptEnvironment = {
+    // Attempts to figure out full URL of where this JS is being served from
+    getFullScriptLocation: (): string | null => {
+        const currentScript = document.currentScript;
+
+        if (!currentScript) {
+            return null;
```
BMillman19: any reason we're using null and not undefined?

packages/instant/src/util/script_environment.ts
```diff
@@ -0,0 +1,60 @@
+import * as _ from 'lodash';
+
+import { ENVIRONMENTS_TO_DOMAINS } from '../constants';
+import { Environment } from '../types';
+
+export type EnvironmentToDomain = { [key in Environment]: string[] };
```
BMillman19: conventionally we put this in the `types` file

steveklebanoff:
I'm going to put up a new version based on the ENV var approach in https://github.com/0xProject/0x-monorepo/pull/1272

PR #1281, opened 2018-11-17
[instant] Close Instant on "back" and re-open instant on "forward by default. Provide option to disable.

fragosti:
## Description

- Changes the `render` method such that it calls `pushState` before it renders. This allows the user to press "back" without us having to worry about them leaving the page. 
- Listens to `window.onpopstate` to know when to open and close instant when the feature is not disabled. When user presses back, instant is removed from the DOM. When he presses forward again, instant is re-rendered.
- Calls `history.back` instead of directly moving instant from the DOM when the closing "X" is pressed.
- Add `shouldDisablePushToHistory` to disable the above behavior. 

Worth noting that this works jankily with the hot reloading in webpack-dev-server, and so we might want to disable it by default.

## Checklist:

### DIFF
--- a/packages/instant/src/index.umd.ts
+++ b/packages/instant/src/index.umd.ts
@@ -53,14 +53,19 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
         injectedDiv.setAttribute('id', INJECTED_DIV_ID);
         injectedDiv.setAttribute('class', INJECTED_DIV_CLASS);
         appendTo.appendChild(injectedDiv);
-        const removeFromDom = () => appendTo.removeChild(injectedDiv);
+        const closeInstant = () => {
+            if (config.onClose) {
+                config.onClose();
+            }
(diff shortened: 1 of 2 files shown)


packages/instant/src/index.umd.ts
```diff
@@ -5,52 +5,98 @@ import * as ReactDOM from 'react-dom';
...
-    injectedDiv.setAttribute('class', INJECTED_DIV_CLASS);
-    appendTo.appendChild(injectedDiv);
-    const instantOverlayProps = {
-        ...props,
-        onClose: () => {
-            appendTo.removeChild(injectedDiv);
-            if (!_.isUndefined(props.onClose)) {
-                props.onClose();
+    // Render instant and return a callback that allows you to remove it from the DOM.
+    const renderInstant = () => {
```
steveklebanoff: `render` has gotten pretty big.  I would prefer for this function to be defined as it's own top-level function instead of defined in here

PR #1283, opened 2018-11-17
[instant] Make package public and fix publishing

fragosti:
## Description

The biggest problem is that we were building to esnext, and not commonjs. However, more generally we weren't following monorepo conventions in `tsconfig.json`, but now we should be. 

## Checklist:


packages/instant/tsconfig.json
```diff
@@ -2,16 +2,11 @@
...
-        "module": "ESNext",
-        "moduleResolution": "node",
-        "lib": ["es2015", "dom"],
-        "target": "es5",
-        "sourceMap": true
+        "allowSyntheticDefaultImports": true
     },
-    "include": ["./src/**/*", "./test/**/*"],
+    "include": ["./src/**/*"],
     "exclude": ["./src/index.umd.ts"]
```
BMillman19: does the `exclude` need to be amended as well?

fragosti: This is just for the commonjs build, the `index.umd.ts` file is the entry-point for the UMD file.

PR #1287, opened 2018-11-20
[instant] Add "Powered By 0x"

steveklebanoff:
## Description

Add "Powered by 0x" to Instant

image

## Checklist:


packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -28,6 +30,8 @@ export class ZeroExInstantContainer extends React.Component<{}, ZeroExInstantCon
         tokenSelectionPanelAnimationState: 'none' as SlideAnimationState,
     };
     public render(): React.ReactNode {
+        const PoweredByLogo = require('../assets/powered_by_0x.svg');
```
fragosti: No need to import this every render.

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -64,6 +68,17 @@ export class ZeroExInstantContainer extends React.Component<{}, ZeroExInstantCon
...
                         <CurrentStandardSlidingPanel />
                     </Container>
+                    <Container
+                        display={{ sm: 'none', default: 'block' }}
+                        marginTop="10px"
+                        marginLeft="auto"
+                        marginRight="auto"
+                        width="140px"
+                    >
+                        <Text href={ZERO_EX_SITE_URL}>
```
fragosti: Are we using the text for the hover styles too? Otherwise an anchor might just suffice.

packages/instant/src/components/zero_ex_instant_container.tsx
steveklebanoff: I tried that at first, but I get the tslint complaint `require statement not part of an import statement [no-var-requires]`  if I put it outside of a function

packages/instant/src/components/zero_ex_instant_container.tsx
steveklebanoff: We're not using the hover style.  It felt kind of unnatural to use `Text` here but wasn't sure the accepted pattern.

steveklebanoff: Changed to `a`

packages/instant/src/components/zero_ex_instant_container.tsx
steveklebanoff: If I try to import, I get `TS2307: Cannot find module '../assets/powered_by_0x.svg'.`

BMillman19:
@steveklebanoff you can use:

`import PoweredByLogo from '../assets/powered_by_0x.svg';`

if you add this to `globals.d.ts`:
```
declare module '*.svg' {
    const content: any;
    /* tslint:disable */
    export default content;
    /* tslint:enable */
}
```

https://stackoverflow.com/questions/44717164/unable-to-import-svg-files-in-typescript

steveklebanoff:
Good call @BMillman19 , updated.

PR #1288, opened 2018-11-20
[pipeline] pull token metadata re trusted tokens

wakkadojo:
## Description

Added trusted token source metadata using 0x and metamask sources

## Testing instructions

Migrate to test db, run node lib/src/scripts/pull_trusted_tokens.js

## Checklist:


packages/pipeline/src/entities/token_metadata.ts
```diff
@@ -14,3 +14,9 @@ export class TokenOnChainMetadata {
     @Column({ type: 'nvarchar', nullable: false })
     public name!: string;
 }
+
+@Entity({ name: 'trusted_tokens', schema: 'raw' })
```
albrow: I imagine it is common for different authorities to use different names for the same token (e.g. "0x", "ZeroEx", "0x Protocol", etc.). Our schema should allow for this, but it looks like it currently does not because `address` is the only primary column for `TokenOnChainMetadata`.

What do you think about removing the `TrustedTokens` entity and simply including `authority` as the second primary column on `TokenOnChainMetadata`?

So `TokenOnChainMetadata` would look like this:

```ts
export class TokenOnChainMetadata {
    @PrimaryColumn()
    public address!: string;

    @PrimaryColumn()
    public authority!: string;

    // ...
}
```

wakkadojo: The reason I did a separate table was to reduce dependencies in the raw layer esp since I envision token_on_chain_metadata coming from the blockchain and trusted sources coming from other providers. 

The thing you're describing, however makes a lot of sense to have in the "concepts" layer (the curated core analytics building blocks). What do you think of producing the combined table there instead?

packages/pipeline/src/entities/ohlcv_external.ts
```diff
@@ -0,0 +1,20 @@
...
+    @PrimaryColumn() public from_symbol!: string;
+    @PrimaryColumn() public to_symbol!: string;
+    @PrimaryColumn() public start_time!: number;
+    @PrimaryColumn() public end_time!: number;
+    
+    @Column() public open!: number;
+    @Column() public close!: number;
+    @Column() public low!: number;
+    @Column() public high!: number;
+    @Column() public volume_from!: number;
```
albrow: We usually use `camelCase` for properties in TypeScript. Can we change this so it looks like the other entities? You will need to specify the `name` in the `@Colum` decorator since it is different than the property name.

For example:

```ts
@Column({name: 'volume_from'})
public volumeFrom!: number;
```

wakkadojo: yeah

packages/pipeline/src/data_sources/trusted_tokens/index.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export interface MetamaskTrustedTokenMeta {
+    address: string;
+    name: string;
+    erc20: boolean;
+    symbol: string;
+    decimals: number;
+}
+
+export class TrustedTokenSource<T> {
```
albrow: nit: Since the type parameter `T` is only used in the `getTrustedTokenMetaAsync` method, we could declare it there instead of at the class level.

packages/pipeline/src/parsers/token_metadata/index.ts
```diff
@@ -0,0 +1,47 @@
+import * as R from 'ramda';
+
+import { MetamaskTrustedTokenMeta, ZeroExTrustedTokenMeta } from '../../data_sources/trusted_tokens';
+import { TokenMetadata } from '../../entities';
+
+/**
+ * Parses Metamask's trusted tokens list.
+ * @param rawResp raw response from the metamask json file.
```
albrow: nit: The raw response comes from an HTTP request, not a JSON file.

packages/pipeline/src/utils/index.ts
```diff
@@ -49,3 +49,12 @@ class BigNumberTransformer implements ValueTransformer {
 }
 
 export const bigNumberTransformer = new BigNumberTransformer();
+
+/**
+ * Returns the unix timestamp of the current hour
+ */
+export function getHourInUnixTime(): number {
```
albrow: This doesn't seem to be used anywhere. Should we remove it?

packages/pipeline/src/entities/ohlcv_external.ts
```diff
@@ -0,0 +1,20 @@
...
+
+@Entity({ name: 'ohlcv_external', schema: 'raw' })
+export class OHLCVExternal {
+    @PrimaryColumn() public exchange!: string;
+    @PrimaryColumn() public fromSymbol!: string;
+    @PrimaryColumn() public toSymbol!: string;
+    @PrimaryColumn() public startTime!: number;
+    @PrimaryColumn() public endTime!: number;
+
+    @Column() public open!: number;
```
albrow: Should we use the `BigNumber` type here just to be safe? `number` can sometimes lead to precision issues.

packages/pipeline/src/utils/index.ts
datamongers: I had already had a pretty significant amt of design done for CC price/volume scaping, this was part of that until priority shifted last week. Going to sync with xianny but 90% sure we'll need this function to interface with cryptocompare

albrow: 👍 okay with keeping it then.

packages/pipeline/src/parsers/token_metadata/index.ts
datamongers: k i'll edit in-line

packages/pipeline/src/entities/ohlcv_external.ts
wakkadojo: number for prices is fine since these are not precision sensitive generally, but bignumber for times is probably a good standard practice since those have like 15 digits. i think we should do this in a separate PR since it applies to all of our timestamp entities?

wakkadojo: <as per discussion, timestamp ms has plenty of runway b/c number has 15 decimals of precision vs 12 decimals of ms since 1970, and pricing data via cryptocompare and similar api's are not precision-sensitive nor reported as long strings of precise values>

PR #1290, opened 2018-11-20
[instant] Don't allow changing of assets when order in progress

steveklebanoff:
## Description

Don't allow user to change assets when an order is happening.

## Checklist:

### DIFF
--- a/packages/instant/src/components/erc20_asset_amount_input.tsx
+++ b/packages/instant/src/components/erc20_asset_amount_input.tsx
@@ -23,6 +23,7 @@ export interface ERC20AssetAmountInputProps {
     startingFontSizePx: number;
     fontColor?: ColorOption;
     isDisabled: boolean;
+    canSelectOtherAsset: boolean;
     numberOfAssetsAvailable?: number;
 }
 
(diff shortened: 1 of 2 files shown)


steveklebanoff:
Thinking about it more, I wonder if we should keep the asset text and icon clickable for now - as it's a way to restart the process.  The code wouldn't be as clean though, as there's no longer a one-to-one mapping, it's nice right now to be able to say "Disabled input means no token selection"

steveklebanoff:
I updated the code to re-enable the token selection once the order is successful, even though the code isn't as clean and it's a non-obvious UX interaction, I think it's good to have _some way_ to restart the process on success right now.

packages/instant/src/components/erc20_asset_amount_input.tsx
```diff
@@ -23,6 +23,7 @@ export interface ERC20AssetAmountInputProps {
     startingFontSizePx: number;
     fontColor?: ColorOption;
     isDisabled: boolean;
```
BMillman19: Lets rename this prop into something more specific like `isInputDisabled`

steveklebanoff: Done in b7eb2e8

PR #1292, opened 2018-11-20
Add `nonce` to `TxOpts`

LogvinovLeon:
## Description

Sometimes transactions get stuck because of insufficient gas price and you need to rebroadcast them.
For that, you need to be able to specify the nonce. That PR allows that.

## Checklist:

### DIFF
--- a/packages/contract-wrappers/CHANGELOG.json
+++ b/packages/contract-wrappers/CHANGELOG.json
@@ -1,5 +1,15 @@
 [
     {
+        "version": "4.1.0",
+        "changes": [
+            {
+                "note":
+                    "Add a `nonce` field for `TxOpts` so that it's now possible to re-broadcast stuck transactions with higher gas cost",
+                "pr": 1292
+            }
+        ]
+    },
+    {
         "timestamp": 1542208198,
         "version": "4.0.2",
         "changes": [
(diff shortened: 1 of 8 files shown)


packages/contract-wrappers/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "4.1.0",
+        "changes": [
+            {
+                "note":
+                    "Add a `nonce` field for `TxOpts` so that it's now possible to re-broadcast stuck transactions with higher gas cost",
```
fabioberger:
```suggestion
                    "Add a `nonce` field for `TxOpts` so that it's now possible to re-broadcast stuck transactions with a higher gas amount",
```

packages/contract-wrappers/src/types.ts
```diff
@@ -142,10 +142,12 @@ export interface MethodOpts {
 /**
  * gasPrice: Gas price in Wei to use for a transaction
  * gasLimit: The amount of gas to send with a transaction (in Gwei)
+ * nonce: The nonce to use for a transaction
```
fabioberger:
```suggestion
 * nonce: The nonce to use for a transaction. If not specified, it defaults to the next incremented nonce.
```

PR #1294, opened 2018-11-20
[instant] Header tweaks

steveklebanoff:
## Description

- Right align amounts
- Autofocus input

## Checklist:


BMillman19:
Looks like it stops autofocusing if you switch to a different asset.

Repro steps:

* select a token
* observe the input is focused
* open token picker
* select a different token
* observe the input is no longer focused

steveklebanoff:
Ah snap, thought this was working -- will fix.  Thanks for testing it out!

steveklebanoff:
@BMillman19  This should be fixed now, see https://github.com/0xProject/0x-monorepo/pull/1294/commits/10c9d0b723ba190a7f1c4df112ae0c5b37887bb9 . Approach is a little funky, but discussed with @fragosti and he was on board with it

packages/instant/src/components/scaling_input.tsx
```diff
@@ -136,6 +138,7 @@ export class ScalingInput extends React.Component<ScalingInputProps, ScalingInpu
                 width={this._calculateWidth(phase)}
                 maxLength={maxLength}
                 disabled={isDisabled}
+                autoFocus={hasAutofocus}
```
fragosti: 👍

PR #1295, opened 2018-11-20
[instant] Rollbar env

steveklebanoff:
## Description

Sets up rollbar tracking.  Like https://github.com/0xProject/0x-monorepo/pull/1280 but with a simpler env set up

⚠️  When we decide on a js cdn, we will have to update the publicPath for the JS source map ⚠️ 
⚠️  This introduces new ENV vars which need to be set to publish, doc will have to be updated ⚠️ 

## Testing instructions

Add a change which forces an error, i.e:

```diff
--- a/packages/instant/src/components/scaling_amount_input.tsx
+++ b/packages/instant/src/components/scaling_amount_input.tsx
@@ -73,6 +73,10 @@ export class ScalingAmountInput extends React.Component<ScalingAmountInputProps,
             stringValue: sanitizedValue,
         });

+        if (sanitizedValue === '101010') {
+            throw new Error('TEST: another input error');
+        }
+
         // Trigger onAmountChange with a valid BigNumber, or undefined if the sanitizedValue is invalid or empty
         const bigNumberValue: Maybe<BigNumber> = _.isEmpty(sanitizedValue)
             ? undefined
➜
```

Deploy to dogfood or staging with `INSTANT_ROLLBAR_CLIENT_TOKEN` and `INSTANT_ROLLBAR_PUBLISH_TOKEN` set.  Trigger error and see it is reported, with correct environment and source map

## Checklist:

### DIFF
--- a/packages/instant/package.json
+++ b/packages/instant/package.json
@@ -64,6 +64,7 @@
         "react-redux": "^5.0.7",
         "redux": "^4.0.0",
         "redux-devtools-extension": "^2.13.5",
+        "rollbar": "^2.5.0",
         "styled-components": "^4.0.2",
         "ts-optchain": "^0.1.1"
     },
@@ -88,7 +89,9 @@
         "make-promises-safe": "^1.1.0",
         "npm-run-all": "^4.1.2",
         "nyc": "^11.0.1",
+        "rollbar-sourcemap-webpack-plugin": "^2.4.0",
         "shx": "^0.2.2",
+        "source-map-loader": "^0.2.4",
         "svg-react-loader": "^0.4.6",
         "ts-jest": "^23.10.3",
         "tslint": "5.11.0",
--- a/packages/instant/src/components/zero_ex_instant_container.tsx
+++ b/packages/instant/src/components/zero_ex_instant_container.tsx
@@ -12,11 +12,14 @@ import { SelectedAssetInstantHeading } from '../containers/selected_asset_instan
 import { ColorOption } from '../style/theme';
 import { zIndex } from '../style/z_index';
 import { OrderProcessState, SlideAnimationState } from '../types';
+import { setupRollbar } from '../util/error_reporter';
 
 import { CSSReset } from './css_reset';
 import { SlidingPanel } from './sliding_panel';
 import { Container } from './ui/container';
 import { Flex } from './ui/flex';
+
+setupRollbar();
 
 export interface ZeroExInstantContainerProps {
     orderProcessState: OrderProcessState;
--- a/packages/instant/src/constants.ts
+++ b/packages/instant/src/constants.ts
@@ -20,6 +20,13 @@ export const HEAP_ANALYTICS_ID = process.env.HEAP_ANALYTICS_ID;
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
 export const PROGRESS_STALL_AT_WIDTH = '95%';
 export const PROGRESS_FINISH_ANIMATION_TIME_MS = 200;
+export const ROLLBAR_CLIENT_TOKEN = process.env.ROLLBAR_CLIENT_TOKEN;
+export const ROLLBAR_ENVIRONMENT = process.env.ROLLBAR_ENVIRONMENT as
+    | 'dogfood'
+    | 'staging'
+    | 'development'
+    | 'production'
+    | undefined;
 export const COINBASE_WALLET_IOS_APP_STORE_URL = 'https://itunes.apple.com/us/app/coinbase-wallet/id1278383455?mt=8';
 export const COINBASE_WALLET_ANDROID_APP_STORE_URL = 'https://play.google.com/store/apps/details?id=org.toshi&hl=en';
 export const COINBASE_WALLET_SITE_URL = 'https://wallet.coinbase.com/';
--- a/packages/instant/src/util/buy_quote_updater.ts
+++ b/packages/instant/src/util/buy_quote_updater.ts
@@ -9,6 +9,7 @@ import { Action, actions } from '../redux/actions';
 import { AffiliateInfo, ERC20Asset } from '../types';
 import { assetUtils } from '../util/asset';
 import { errorFlasher } from '../util/error_flasher';
+import { errorReporter } from '../util/error_reporter';
 
 export const buyQuoteUpdater = {
     updateBuyQuoteAsync: async (
@@ -49,8 +50,9 @@ export const buyQuoteUpdater = {
                 } else {
                     throw error;
                 }
+            } else {
+                errorReporter.report(error);
             }
-            // TODO: report to error reporter on else
 
             return;
         }
--- a/packages/instant/src/util/error_reporter.ts
+++ b/packages/instant/src/util/error_reporter.ts
@@ -1,0 +1,68 @@
+import { logUtils } from '@0x/utils';
+
+import { ROLLBAR_CLIENT_TOKEN, ROLLBAR_ENVIRONMENT } from '../constants';
+
+// Import version of Rollbar designed for embedded components
+// See https://docs.rollbar.com/docs/using-rollbarjs-inside-an-embedded-component
+// tslint:disable-next-line:no-var-requires
+const Rollbar = require('rollbar/dist/rollbar.noconflict.umd');
+
+const shouldAllowRollbar = () => {
+    if (ROLLBAR_ENVIRONMENT === 'development') {
+        return process.env.ROLLBAR_FORCE_DEVELOPMENT_REPORT ? true : false;
+    }
+    return true;
+};
+
+let rollbar: any;
+if (ROLLBAR_CLIENT_TOKEN && ROLLBAR_ENVIRONMENT && shouldAllowRollbar()) {
+    rollbar = new Rollbar({
+        accessToken: ROLLBAR_CLIENT_TOKEN,
+        captureUncaught: true,
+        captureUnhandledRejections: true,
+        enabled: true,
+        itemsPerMinute: 10,
+        maxItems: 500,
+        payload: {
+            environment: ROLLBAR_ENVIRONMENT,
+            client: {
+                javascript: {
+                    source_map_enabled: true,
+                    code_version: process.env.GIT_SHA,
+                    guess_uncaught_frames: true,
+                },
+            },
+        },
+        uncaughtErrorLevel: 'error',
+        ignoredMessages: [
+            // Errors from the third-party scripts
+            'Script error',
+            // Network errors or ad-blockers
+            'TypeError: Failed to fetch',
+            'Exchange has not been deployed to detected network (network/artifact mismatch)',
+            // Source: https://groups.google.com/a/chromium.org/forum/#!topic/chromium-discuss/7VU0_VvC7mE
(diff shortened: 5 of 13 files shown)


packages/instant/src/util/buy_quote_updater.ts
```diff
@@ -49,8 +50,9 @@ export const buyQuoteUpdater = {
                 } else {
                     throw error;
                 }
+            } else {
+                errorReporter.report(error);
```
BMillman19: hmm do we want to report here if we are intentionally suppressing errors? These are errors that we would be catching and swallowing otherwise

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -12,12 +12,15 @@ import { SelectedAssetInstantHeading } from '../containers/selected_asset_instan
...
 import { zIndex } from '../style/z_index';
 import { OrderProcessState, SlideAnimationState } from '../types';
+import { setupRollbar } from '../util/error_reporter';
 
 import { CSSReset } from './css_reset';
 import { SlidingPanel } from './sliding_panel';
 import { Container } from './ui/container';
 import { Flex } from './ui/flex';
 
+setupRollbar();
```
fragosti: I think this type of "set up" logic has typically gone into the provider file.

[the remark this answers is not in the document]
fragosti: Also this doesn't appear to be doing anything?

packages/instant/src/util/error_reporter.ts
```diff
@@ -0,0 +1,68 @@
...
+// See https://docs.rollbar.com/docs/using-rollbarjs-inside-an-embedded-component
+// tslint:disable-next-line:no-var-requires
+const Rollbar = require('rollbar/dist/rollbar.noconflict.umd');
+
+const shouldAllowRollbar = () => {
+    if (ROLLBAR_ENVIRONMENT === 'development') {
+        return process.env.ROLLBAR_FORCE_DEVELOPMENT_REPORT ? true : false;
+    }
+    return true;
+};
```
fragosti: I find the current function of `configs => rollbar state` a bit confusing. Can't there just be a `ROLLBAR_ENABLED` var as far as Instant is concerned?

Also `ROLLBAR_ENVIRONMENT` just means `ENVIRONMENT` right?

packages/instant/webpack.config.js
```diff
@@ -11,22 +12,107 @@ const GIT_SHA = childProcess
...
+
+    if (!process.env[ROLLBAR_CLIENT_TOKEN_ENV_NAME]) {
+        throw new Error(`${ROLLBAR_CLIENT_TOKEN_ENV_NAME} must be set for ${environmentName}`);
+    }
+
+    if (!rollbarPlugin) {
+        if (environmentName === 'dogfood' || environmentName === 'staging') {
+            throw new Error(
+                `Please set rollbar env var ${ROLLBAR_PUBLISH_TOKEN_ENV_NAME} to a Rollbar project access token with post_server_item permissions to deploy source maps to ${environmentName}`,
+            );
```
fragosti: The final boolean check is redundant because we already know we are in a state of `requiresRollbar` I think.

packages/instant/webpack.config.js
```diff
@@ -11,22 +12,107 @@ const GIT_SHA = childProcess
...
+    ];
+
+    const rollbarPlugin = getRollbarPlugin(environmentName);
+    if (rollbarPlugin) {
+        console.log('Using rollbar plugin');
+        plugins = plugins.concat(rollbarPlugin);
+    } else {
+        console.log('Not using rollbar plugin');
+    }
+    validateRollbar(environmentName, rollbarPlugin);
```
fragosti: Should we validate it if we are not using it?

packages/instant/webpack.config.js
```diff
@@ -11,22 +12,107 @@ const GIT_SHA = childProcess
...
+    let plugins = [
+        new webpack.DefinePlugin({
+            'process.env': {
+                GIT_SHA: JSON.stringify(GIT_SHA),
+                NPM_PACKAGE_VERSION: JSON.stringify(process.env.npm_package_version),
+                HEAP_ANALYTICS_ID: getHeapAnalyticsId(environmentName),
+                ROLLBAR_ENVIRONMENT: JSON.stringify(environmentName),
+                ROLLBAR_CLIENT_TOKEN: JSON.stringify(process.env[ROLLBAR_CLIENT_TOKEN_ENV_NAME]),
+                ROLLBAR_FORCE_DEVELOPMENT_REPORT: JSON.stringify(process.env.INSTANT_ROLLBAR_FORCE_DEVELOPMENT_REPORT),
+            },
```
fragosti: Feel like we can get away with just `ENVIRONMENT`, `ROLLBAR_CLIENT_TOKEN`, `ROLLBAR_ENABLED`

packages/instant/src/util/buy_quote_updater.ts
steveklebanoff: Good callout.  I can update the code, but wanted to make sure we are on the same page wrt to the desired implementation here:

When this `catch` is hit, there are two types of errors: "Known / Expected" errors (i.e. not enough ZRX available), which we want to swallow, not report to rollbar, and flash to the user if we are not in a heartbeat.  Then there are "Unexpected" errors, which we want to always report to rollbar, regardless of if we are in a heartbeat or not.

packages/instant/webpack.config.js
steveklebanoff: Looking at this again, I agree that the `ROLLBAR_FORCE_DEVELOPMENT_REPORT` feels a little unnatural.  I think just sending in a `ROLLBAR_ENABLED` var feels more clear

packages/instant/webpack.config.js
steveklebanoff: Ah good call, didn't remove this during a refactor.  Thanks for the sharp  👀

packages/instant/webpack.config.js
steveklebanoff: This function will throw an error if we are an env that requires sourcemaps, but there is no `rollbarPlugin` set.  Perhaps renaming the function to something like `validateRollbarPresence` may make things more clear.

packages/instant/src/components/zero_ex_instant_container.tsx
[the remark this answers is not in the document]
steveklebanoff: Done in 22a3124

packages/instant/src/util/error_reporter.ts
steveklebanoff: Yes, renamed

steveklebanoff: LMK if the env var logic still seems confusing to ya

packages/instant/webpack.config.js
```diff
@@ -11,22 +12,104 @@ const GIT_SHA = childProcess
...
+
+    const envVars = {
+        GIT_SHA: JSON.stringify(GIT_SHA),
+        NPM_PACKAGE_VERSION: JSON.stringify(process.env.npm_package_version),
+        HEAP_ANALYTICS_ID: getHeapAnalyticsId(environmentName),
+        INSTANT_ENVIRONMENT: JSON.stringify(environmentName),
+        ROLLBAR_CLIENT_TOKEN: JSON.stringify(process.env[ROLLBAR_CLIENT_TOKEN_ENV_NAME]),
+    };
+
+    const canRollbarBeEnabled =
```
steveklebanoff: The idea behind `INSTANT_ROLLBAR_FORCE_DEVELOPMENT_REPORT` is guarding against having  `INSTANT_ROLLBAR_CLIENT_TOKEN` set in your environment and accidentally reporting in the development environment

steveklebanoff:
@fragosti I agree that the webpack config is getting a bit big, and could benefit from splitting into multiple env-based configs.  That being said, this feature has taken more time than expected already, and I feel like the current webpack config is 'good enough for now' -- we could split up as a tech debt cleanup ticket, or if we start feeling the pain by adding more complex modifications to the config like this.

steveklebanoff:
@BMillman19 @fragosti Made a bunch of changes and replied to comments, ready for another review when ya'll have time

packages/instant/src/util/error_reporter.ts
fragosti: much better now!

packages/instant/src/util/error_reporter.ts
```diff
@@ -0,0 +1,62 @@
...
+import { EMBEDDED_DOMAINS, INSTANT_ENVIRONMENT, ROLLBAR_CLIENT_TOKEN, ROLLBAR_ENABLED } from '../constants';
+
+// Import version of Rollbar designed for embedded components
+// See https://docs.rollbar.com/docs/using-rollbarjs-inside-an-embedded-component
+// tslint:disable-next-line:no-var-requires
+const Rollbar = require('rollbar/dist/rollbar.noconflict.umd');
+
+let rollbar: any;
+// Configures rollbar and sets up error catching
+export const setupRollbar = (): any => {
```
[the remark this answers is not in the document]
steveklebanoff: I used to have this logic defined at the top of the file, outside of a function. My intention was to have it loaded "on import", as you suggested.

*But*, this is what I believe was happening: because the compiler tries to be smart about not import unused code, the code was not executed unless I called a function in the file.  Thus, I created a function called `setupRollbar` which just returned the rollbar instance, forcing the code to be executed.  This was confusing for the reader though, per your comment: ["Also this doesn't appear to be doing anything?"](https://github.com/0xProject/0x-monorepo/pull/1295#discussion_r235429735).  

So, I decided to put all this instantiation logic into the setup function,  this change happened in https://github.com/0xProject/0x-monorepo/pull/1295/commits/22a31246622e7139185d080cae9bef275087e245

packages/instant/src/constants.ts
```diff
@@ -20,6 +20,23 @@ export const HEAP_ANALYTICS_ID = process.env.HEAP_ANALYTICS_ID;
 export const COINBASE_API_BASE_URL = 'https://api.coinbase.com/v2';
 export const PROGRESS_STALL_AT_WIDTH = '95%';
 export const PROGRESS_FINISH_ANIMATION_TIME_MS = 200;
+export const EMBEDDED_DOMAINS = [
```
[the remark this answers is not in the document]
steveklebanoff: I think _ideally_ we would report even if they embedded via the npm package and not from the umd module.  But, we also don't want to clog our rollbar with exceptions unrelated to our code. Only reporting errors that from JS hosted on our known CDN seemed like the only straightforward way to achieve this naively, but I'm open to other thoughts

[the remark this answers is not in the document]
fragosti: No longer an issue now I guess :D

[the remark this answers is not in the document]
steveklebanoff: I'm pretty stoked on our CDN decision, awesome that it simplifies things like this.

packages/instant/package.json
```diff
@@ -7,7 +7,6 @@
     "private": true,
     "description": "0x Instant React Component",
     "main": "umd/instant.js",
-    "private": true,
```
BMillman19: we want to keep this private

steveklebanoff: `private` was defined twice, see line 7 above ^

PR #1296, opened 2018-11-20
[instant] Replace "Powered By 0x" svg

steveklebanoff:
## Description

Without this, React gives the warning:

>Warning: React does not recognize the `maskType` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `masktype` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

I think figma accidentally exported a svg property which is in valid (maskType).  Chris exported a new one and it didn't include any masks.

## Checklist:


BMillman19:
@steveklebanoff we might be able to play with the webpack config to transform / sanitize these properties for the future: https://github.com/jhamlet/svg-react-loader

Otherwise this looks good to me for now

Issue #1297, opened 2018-11-20
Interface for artifacts mismatch


fabioberger:
When trying to generate the Documentation JSON for `@0x/migrations` I got this error:

```
{...} is not assignable to parameter of type \'ContractArtifact\'.\n  Property \'compiler\' is missing in type {...}
```

Currently, `@0x/contract-addresses` return an succinct artifact object (without the `compiler` key), but it gets passed into methods expecting the full artifact type defined by `ContractArtifact` (where `compiler` is a required key).

## Expected Behavior

Places that are being passed succinct artifacts should specifically allow them to be passed in. Perhaps accepting the union type: `ContractArtifact | SuccinctContractArtifact`.

## Current Behavior

They only accept full artifacts, despite not needing the missing fields.

## Steps to Reproduce (for bugs)

```
1.
2.
3.
```

## Your Environment

|             Package | Version |
| ------------------: | :------ |

| Network |
| ------- |
| NAME    |

PR #1298, opened 2018-11-20
Fix Artifact Interface Type Issue

fabioberger:
Fixes: https://github.com/0xProject/0x-monorepo/issues/1297

## Checklist:

### DIFF
--- a/packages/abi-gen-wrappers/CHANGELOG.json
+++ b/packages/abi-gen-wrappers/CHANGELOG.json
@@ -1,5 +1,15 @@
 [
     {
+        "version": "1.1.0",
+        "changes": [
+            {
+                "note":
+                    "`deployFrom0xArtifactAsync` additionally accepts artifacts that conform to the `SimpleContractArtifact` interface",
+                "pr": 1298
+            }
+        ]
+    },
+    {
         "timestamp": 1542208198,
         "version": "1.0.5",
         "changes": [
--- a/packages/abi-gen-wrappers/package.json
+++ b/packages/abi-gen-wrappers/package.json
@@ -32,6 +32,7 @@
     "devDependencies": {
         "@0x/abi-gen": "^1.0.16",
         "@0x/tslint-config": "^1.0.10",
+        "@0x/types": "^1.2.1",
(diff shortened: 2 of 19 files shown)


fabioberger:
@LogvinovLeon another way to solve this issue is to change the interface for `deployFrom0xArtifactAsync` to be:

```
public static async deployFrom0xArtifactAsync(
        abi: ContractAbi,
        bytecode: string,
        provider: Provider,
        txDefaults: Partial<TxData>) {
```

Which I actually prefer since it means developers don't need to create an object which conforms to our `artifacts` format. Instead, they just need the `abi` and `bytecode` and are ready to go. 

Thoughts?

LogvinovLeon:
Contract wrappers already hae 2 functions:
* `deployAsync`
* `deployFrom0xArtifactAsync`
So I don't see the benefit to change the first one to be like a second one.
Maybe we should just remove the first one and use the second one exclusively.
I feel like if we need just two fields than there is no need for those wrappers to even know about the whole artifact. Slightly more verbose, but less coupling.

Other solution is to have an artifact types tree and have them inherit from each other or use mixins.
Then each part in code specifies which parts of the artifact i wants. That could be cool.

`profiler(artifacts: BaseArtifact | SourceMapArtifact | ByteCodeArtifact)`
`deploy(artifac: BaseArtifact | ByteCodeArtifact | AbiArtifact)`
`useInWebsite(artifac: BaseArtifact | AddressArtifact | AbiArtifact)`

And then we can have validation functions that validate those pieces. And nice error messages like:
Can't use artifact s in website because the addresses are missing.

Overall I think this PR goes in a good direction. We have multiple artifact formats which are quite simmilar and differ only in required modifiers, but if we make everything optional it'll become a mess.

fabioberger:
Ah, I missed `deployAsync`, looks like we thought about this strong coupling in-advance for external users. In that case, we can leave it as-is.

Hm... I feel like there is almost endless customization possible with artifacts. I think we'll need to define a handful that we use (currently `ContractArtifact` and `SimpleContractArtifact`) and be careful not to pass these into publicly exported methods.

PR #1299, opened 2018-11-20
[instant] Add more event properties

steveklebanoff:
## Description

Add more event properties (orderSource, affiliateInfo )

## Checklist:


packages/instant/src/util/analytics.ts
```diff
@@ -59,6 +65,29 @@ export const analytics = {
             heapUtil.evaluateHeapCall(heap => heap.addEventProperties(properties));
         });
     },
+    generateEventProperties: (state: State, window: Window): AnalyticsEventOptions => {
+        let orderSource = 'unknown';
+        const orderProvider = state.providerState.assetBuyer.orderProvider;
+        if (orderProvider instanceof StandardRelayerAPIOrderProvider) {
```
BMillman19: hmm this works, but we can also pass in the props into this function in order to avoid reaching into asset buyer this far, also we will have this implicit dependency on the implementation of asset buyer here with this approach

steveklebanoff: Some things are needed from props, while others from state.  I updated this to take in all params explicitly in 9c86d7f250d652bd46ef0cad1b02c81f4e211c96 -- feels a bit more verbose.   I prefer the original implementation, esp. since we have typesafety for the asset buyer reliance, but I don't feel uber strongly about it

PR #1300, opened 2018-11-21
[instant] Implement account related events

BMillman19:
## Description

Adds the following events
```
    ACCOUNT_LOCKED = 'Account - Locked',
    ACCOUNT_READY = 'Account - Ready',
    ACCOUNT_UNLOCK_REQUESTED = 'Account - Unlock Requested',
    ACCOUNT_UNLOCK_DENIED = 'Account - Unlock Denied',
    ACCOUNT_ADDRESS_CHANGED = 'Account - Address Changed',
```

## Testing instructions

Check out live capture on heap

## Checklist:

### DIFF
--- a/packages/instant/src/redux/analytics_middleware.ts
+++ b/packages/instant/src/redux/analytics_middleware.ts
@@ -27,19 +27,17 @@ export const analyticsMiddleware: Middleware = store => next => middlewareAction
                     prevAccount.state === AccountState.Ready && prevAccount.address !== curAccount.address;
                 const ethAddress = curAccount.address;
                 if (didJustTurnReady) {
-                    analytics.trackAccountReady(ethAddress);
-                    analytics.addUserProperties({ ethAddress });
+                    analytics.trackAccountReady({ address: ethAddress });
+                    analytics.addUserProperties({ lastKnownEthAddress: ethAddress });
                 } else if (didJustUpdateAddress) {
-                    analytics.trackAccountAddressChanged(ethAddress);
-                    analytics.addUserProperties({ ethAddress });
+                    analytics.trackAccountAddressChanged({ address: ethAddress });
+                    analytics.addUserProperties({ lastKnownEthAddress: ethAddress });
                 }
             }
             break;
         case ActionTypes.SET_ACCOUNT_STATE_LOCKED:
             if (prevAccount.state !== AccountState.Locked && curAccount.state === AccountState.Locked) {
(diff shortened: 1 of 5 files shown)


packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -35,10 +21,26 @@ export const analyticsMiddleware: Middleware = store => next => middlewareAction
 
     switch (nextAction.type) {
         case ActionTypes.SET_ACCOUNT_STATE_READY:
-            if (curAccount.state === AccountState.Ready && shouldTriggerWalletReady(prevAccount, curAccount)) {
+            if (prevAccount.state !== AccountState.Ready && curAccount.state === AccountState.Ready) {
```
steveklebanoff: I feel like there's a lot of logic here, and is a bit hard to digest.

I would prefer to use some helper functions in here if possible -- like `accountTurnedReady(prevAccount, curAccount)` instead of this if conditional here, and on line 29 a helper function for `accountAddressChanged(prevAccount, curAccount)`.  

We will still need a `curAccount.state === AccountState.Ready` in the `if` for the typechecking to know that the properties we expect are there though.

What do you think?

BMillman19: yea its a bit weird and a bit more confusing to add the extra check for type checking if we go the helper function route.

tbh i don't think it is that difficult to follow with the comments there

steveklebanoff: An alternative middle of the road would be something like:

```ts
        case ActionTypes.SET_ACCOUNT_STATE_READY:
            if (curAccount.state === AccountState.Ready) {
                const didJustTurnReady = prevAccount.state !== AccountState.Ready;
                const didJustUpdateAddress =
                    prevAccount.state === AccountState.Ready && prevAccount.address !== curAccount.address;

                const ethAddress = curAccount.address;
                if (didJustTurnReady) {
                    analytics.trackAccountReady(ethAddress);
                    analytics.addUserProperties({ ethAddress });
                } else if (didJustUpdateAddress) {
                    analytics.trackAccountAddressChanged(ethAddress);
                    analytics.addUserProperties({ ethAddress });
                }
            }
            break;
```

This reads a little cleaner to me: we don't need to check `prevAccount.state === AccountState.Ready` in both conditions, and we have self-documenting variable names instead of relying on comments that could go out of date.

That being said, I recognize this is largely personal preference, if you feel strongly about leaving it as-is I'm OK w/ that

PR #1304, opened 2018-11-22
Add a POC for callData-like structured revert reasons with additional data

LogvinovLeon:
## Description

If we want to make 0x multi-language we want to have better error messages in smart contracts.
That requires adding some additional data to them. For example `orderHash` or `transactionHash`. That allows users to know which order fails (especially important for batch fills/cancels). We can encode that data into revert reason in different ways.

The way proposed in this PR is similar to the way asset data/call data encoding works.
We just encode the params and the revert reason into bytes and throw that as a string. Later in tests or in client code we decode it as `callData`. Look at the PR for an example.

For now I've transformed just one error. `INVALID_ORDER_SIGNATURE`.

There is some magic happening in the matching tests for computing order hashes. That's because of our `callData` optimizations and overriding order parts with parts from other orders.

In tests we need to do that because we deliberately submit incorrect inputs and the stuff hashed on the bockchain is not the same as the stuff we submit.

In a client-side library that will not happen. We'll have some basic validation.

This is a first PR of a series and it doesn't go to development. It goes to my prototype branch `feature/error-abi`

## Testing instructions

Run smart contracts tests

## Types of changes

## Checklist:


packages/contracts/test/exchange/match_orders.ts
```diff
@@ -1168,14 +1168,22 @@ describe('matchOrders', () => {
...
                 exchangeWrapper.matchOrdersAsync(signedOrderLeft, signedOrderRight, takerAddress),
                 // We are assuming assetData fields of the right order are the
                 // reverse of the left order, rather than checking equality. This
                 // saves a bunch of gas, but as a result if the assetData fields are
                 // off then the failure ends up happening at signature validation
-                RevertReason.InvalidOrderSignature,
+                {
+                    reason: RevertReason.InvalidOrderSignature,
+                    params: { orderHash: orderHashRightPrim },
+                },
```
fabioberger: Nice, this is a way more precise assert then what we had about this. Feel these error messages are going to greatly improve the quality of our tests as well 🚀

packages/contracts/test/utils/assertions.ts
```diff
@@ -1,16 +1,34 @@
...
+);
+const errorAbiBySelector = abisToAbiBySelector(ERROR_ABIS);
+interface InvalidOrderSignatureError {
+    reason: RevertReason.InvalidOrderSignature;
+    params: {
+        orderHash: string;
+    };
+}
+
+type SmartContractError = InvalidOrderSignatureError;
```
fabioberger: Will this eventually be a union type?

packages/contracts/test/utils/assertions.ts
```diff
@@ -80,6 +98,10 @@ export async function expectInsufficientFundsAsync<T>(p: Promise<T>): Promise<vo
     return expect(p).to.be.rejectedWith(errMessage);
 }
 
+function escapeRegExp(str: string): string {
+    return str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'); // $& means the whole matched string
```
fabioberger: Not sure I fully understand what's going on here. Can you explain?

packages/contracts/test/utils/assertions.ts
LogvinovLeon: Yes

packages/contracts/test/utils/assertions.ts
LogvinovLeon: If I want to match exactly the string s with a regex how do I do that?

`new RegExp(s)`? Not too fast cowboy. Some characters are special within the regexp context. Like back slashes, brackets, comas, questions marks etc. This function escapes all of that so if your string is `'.'` Your regex will match exactly that strings and not all strings with one character.
`'.'` -> `'\\.'`

packages/contracts/test/exchange/match_orders.ts
LogvinovLeon: If you have a better naming idea than _Prim I'm happy to adapt it. It's just a placeholder while I'm thinking about a better one.

packages/contracts/test/utils/assertions.ts
LogvinovLeon: No worries, I didn't write that myself. it's from Mozilla docs. Scroll slightly up from here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Using_parentheses

packages/contracts/test/exchange/match_orders.ts
LogvinovLeon: I've meant `Prime` as [here](https://en.wikipedia.org/wiki/Prime_(symbol)).
_In mathematics, the prime is generally used to generate more variable names for things which are similar, without resorting to subscripts – `x′` generally means something related to or derived from `x`._

It's not the hash of the optimized order. it's the hash of the order that we get when we try to deoptimize the order on-chain assuming the optimization was done correctly. So if it was - that's the hash of the right order. If it wasn't - it's a hash of the fake order that we've never seen before even during optimization phase.

fabioberger: gotcha

LogvinovLeon:
Test are failing on development too. We're working on fixing them

LogvinovLeon:
I just reopened it so that autolabeler kicks in.

packages/contracts/contracts/protocol/Exchange/MixinExchangeCore.sol
```diff
@@ -354,7 +354,7 @@ contract MixinExchangeCore is
                     order.makerAddress,
                     signature
                 ),
-                "INVALID_ORDER_SIGNATURE"
+                string(abi.encodeWithSignature("INVALID_ORDER_SIGNATURE(bytes32)", orderInfo.orderHash))
```
Recmo: This is a good hack to add data to the existing revert reasons. Remember that Solidity encodes those as `abi.encodeWithSignature("Error(string)", <revert reason>)`. So now we effectively get

```
abi.encodeWithSignature("Error(string)", string(
    abi.encodeWithSignature("INVALID_ORDER_SIGNATURE(bytes32)",
        orderInfo.orderHash)))
```

This can lead to zero bytes causing C-style string termination. Please test to make sure all this is handled correctly by all libraries involved.

I expect some future version of Solidity to support something natively. There is currently no open issue for it, so we could propose one.

Whatever the future standard is going to be, it is unlikely to be our current double-encoded one. Rather it would be something like:

```
abi.encodeWithSignature("ErrorInvalidOrderSignature(bytes32)", orderInfo.orderHash)
```

We could implement this already by writing it in assembly and it would be cleaner and cheaper, but that doesn't seem worth (1) the effort and (2) breaking compatibility with the existing standard.

Speaking of compatibility, the current implementation does not produce a readable string, which is what is expected. I do not think this is a major concern.

packages/contracts/test/utils/assertions.ts
```diff
@@ -104,7 +129,74 @@ export async function expectTransactionFailedAsync(p: sendTransactionResult, rea
...
+    // make-promises-safe.
+    p.catch(e => {
+        _.noop(e);
+    });
+
+    if (_.isUndefined(nodeType)) {
+        nodeType = await web3Wrapper.getNodeTypeAsync();
+    }
+    switch (nodeType) {
+        case NodeType.Ganache:
```
Recmo: This feels like the wrong place to do it. Can't web3Wrapper take care of abstracting over the node type? It can expose a flag indicating whether the node supports revert reasons or not, which this function would use.

packages/contracts/test/utils/assertions.ts
```diff
@@ -104,7 +129,74 @@ export async function expectTransactionFailedAsync(p: sendTransactionResult, rea
...
+                        throw new Error('Expected just one transaction hash in Ganache error object');
+                    }
+                    const txHash = err.hashes[0];
+                    const ganacheError = err.results[txHash];
+                    const returnData: string = ganacheError.return;
+                    const REVERT_ERROR_SELECTOR_BYTES_LENGTH = 4;
+                    const [revertReasonHex] = ethers.utils.defaultAbiCoder.decode(
+                        ['bytes'],
+                        ethers.utils.hexDataSlice(returnData, REVERT_ERROR_SELECTOR_BYTES_LENGTH),
+                    );
```
Recmo: Re-use `decodeCallData` here with `Error(string)` as ABI to get `revertReasonHex`.

packages/order-utils/src/call_data_utils.ts
```diff
@@ -0,0 +1,86 @@
...
+        {},
+    );
+    return abiBySelector;
+}
+
+/**
+ * Computes the 4-byte selector of the method.
+ * @param methodAbi Method ABI
+ */
+export function getMethodSelector(methodAbi: MethodAbi): string {
```
Recmo: This should be pushed upstream to ethersjs.

packages/order-utils/src/call_data_utils.ts
```diff
@@ -0,0 +1,86 @@
...
+            .toString('hex'),
+    );
+    return selector;
+}
+/**
+ * Decoded call data into function name and params using the provided ABIs.
+ * @param abiBySelector Hash of ABI indexed by 4-byte selector
+ * @param callDataHex Call data to decode
+ */
+export function decodeCallData(abiBySelector: AbiBySelector, callDataHex: string): DecodedCallData {
```
Recmo: Same here.

packages/contracts/contracts/protocol/Exchange/MixinExchangeCore.sol
LogvinovLeon: Very good catch. I think the initial EIP about revert reasons had some discussions about additional data but they've decided to not do that for now because of 1. It's easier. 2. It's faster to decide on smth easier 3. Not everyone needs additional data (debatable).
I think in future they might agree on smth but there is no way for us to guess. Assuming they will not break the basic functionality - there will always be a way for our users to decode our error messages. If you can propose a better scheme - I'm all about it. I think I'm against the assembly solution because it's harder for users to see what's happening. We might still write it and check the gas  costs though. It might break the existing libraries trying to decode the standard revert reasons though. Do I understand it correctly, that we'll need to write a separate assembly code for each revert reason type? They have different payloads.

packages/order-utils/src/call_data_utils.ts
LogvinovLeon: Will do

packages/order-utils/src/call_data_utils.ts
LogvinovLeon: Will do

packages/contracts/contracts/protocol/Exchange/MixinExchangeCore.sol
Recmo: > It might break the existing libraries trying to decode the standard revert reasons though.

Yeah. That's why we should stick with what you made for now. We could put up an EIP and push some standard through, then switch in v.3.

> We might still write it and check the gas costs though.

Generally you would not optimize the error cases. Do we need cheap reverts for batchfills?

> Do I understand it correctly, that we'll need to write a separate assembly code for each revert reason type?

Just one assembly function would be required:

```solidity
function rawRevert(bytes data) internal {
     assembly {
        revert(add(data, 0x20), mload(data))
     }
}
```

which can then be used as:

```
rawRevert(abi.encodeWithSignature("ErrorInvalidOrderSignature(bytes32)", orderInfo.orderHash))
```

Recmo: > This can lead to zero bytes causing C-style string termination. Please test to make sure all this is handled correctly by all libraries involved.

Another thing to watch out for are byte sequences that are not valid UTF8. Casting binary to strings is a dangerous.

packages/contracts/test/utils/assertions.ts
LogvinovLeon: It required some hacks because of the selector mismatch. But I've added an option to override the selector while decoding.

packages/order-utils/src/call_data_utils.ts
LogvinovLeon: Removed. Using `AbiEncoder` now

contracts/test-utils/src/assertions.ts
```diff
@@ -1,16 +1,50 @@
...
 
 const expect = chai.expect;
 
 let nodeType: NodeType | undefined;
 
+const ERROR_ABIS = _.map(
+    constants.ERROR_ABI_STRINGS,
+    (errorAbiString: string) => ethers.utils.parseSignature(errorAbiString) as MethodAbi,
+);
+const errorAbiBySelector = abisToAbiBySelector(ERROR_ABIS);
```
fabioberger: In-line:

```
const selectorToAbi = {};
_.map(abis, abi => {
const selector = new AbiEncoder.Method(methodAbi).getSelector();
selectorToAbi[selector] = abi;
});
```

contracts/test-utils/src/constants.ts
```diff
@@ -52,6 +53,9 @@ export const constants = {
     WORD_LENGTH: 32,
     ZERO_AMOUNT: new BigNumber(0),
     PERCENTAGE_DENOMINATOR: new BigNumber(10).pow(18),
+    SOLIDITY_REVERT_ABI: 'function Error(string reason)',
+    BYTES_REVERT_ABI: 'function Error(bytes encodedRevertReasonWithAssociatedData)',
```
fabioberger:
```suggestion
    BYTES_REVERT_ABI_STRING: 'function Error(bytes encodedRevertReasonWithAssociatedData)',
```

contracts/test-utils/src/constants.ts
```diff
@@ -52,6 +53,9 @@ export const constants = {
     WORD_LENGTH: 32,
     ZERO_AMOUNT: new BigNumber(0),
     PERCENTAGE_DENOMINATOR: new BigNumber(10).pow(18),
+    SOLIDITY_REVERT_ABI: 'function Error(string reason)',
```
fabioberger:
```suggestion
    SOLIDITY_REVERT_ABI_STRING: 'function Error(string reason)',
```

packages/order-utils/src/call_data_utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    const selector = ethers.utils.hexDataSlice(callDataHex, 0, constants.SELECTOR_LENGTH);
+    const abi = abiBySelector[selector];
+    if (_.isUndefined(abi)) {
+        throw new Error(`Unable to decode call data. Unknown selector ${selector}`);
+    }
+    const method = new AbiEncoder.Method(abi);
+    /**
+     * HACK(leo) Sometimes we want to decode smth as smth else. e.g. `Error(string)` as `Error(bytes)`.
+     * In that case `abiBySelector` has the ABI of the later and the `selector` of the first.
```
fabioberger:
```suggestion
     * In that case `abiBySelector` has the ABI of the latter and the `selector` of the first.
```

contracts/test-utils/src/assertions.ts
```diff
@@ -1,16 +1,50 @@
...
+const solidityRevertSelector = new AbiEncoder.Method(ethers.utils.parseSignature(
+    constants.SOLIDITY_REVERT_ABI,
+) as MethodAbi).getSelector();
+const bytesRevertAbi = ethers.utils.parseSignature(constants.BYTES_REVERT_ABI) as MethodAbi;
+/**
+ * HACK: By default solidity encodes revert reasons as `Error(string)` we want to throw additional data
+ * so we encode that data using an error signature e.g. `function INVALID_ORDER_SIGNATURE(bytes32 orderHash)`.
+ * After that we treat those bytes as string and use it instead of a revert reason. When we try to decode that malformed revert reason - we have a selector - ABI mismatch.
```
fabioberger:
```suggestion
 * We pretend those encoded bytes are a string (Solidity lets us do that) and use it instead of a string revert reason. 
 * In order for us to decode the revert properly, we need to use ABI `Error(bytes)` NOT `Error(string)`. Solidity by default
 * uses the selector for `Error(string)`, so we introduce a hack where we map that selector to the ABI definition for `Error(bytes)`. 
 * Doing so let's us decode the revert string AS BYTES! Which is what we need to do to avoid a selector <-> ABI mismatch.
 
```

contracts/test-utils/src/assertions.ts
```diff
@@ -1,16 +1,50 @@
...
+ * That's why we need this hack, so that the decoding function uses `bytesRevertAbi` for revert reasons.
```
fabioberger:
```suggestion

```

contracts/test-utils/src/assertions.ts
```diff
@@ -1,16 +1,50 @@
...
+ */
+errorAbiBySelector[solidityRevertSelector] = bytesRevertAbi;
+interface InvalidOrderSignatureError {
```
fabioberger:
```suggestion

interface InvalidOrderSignatureError {
```

packages/order-utils/src/call_data_utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    callParams: CallParams;
+}
+
+export interface AbiBySelector {
+    [selector: string]: MethodAbi;
+}
+/**
+ * Converts the list of ABIs to the hash indexed by the 4-byte selector.
+ */
+export function abisToAbiBySelector(abis: MethodAbi[]): AbiBySelector {
```
fabioberger: Remove this

PR #1305, opened 2018-11-22
[@0x/contract-templates] Make contract-templates an npm package

LogvinovLeon:
## Description
We want to version our handlebars templates and therefore make them an npm package.

## Checklist:


package.json
```diff
@@ -22,9 +22,9 @@
         "install:all": "yarn install",
         "wsrun": "wsrun",
         "lerna": "lerna",
-        "build": "wsrun build $PKG --fast-exit -r --stages",
-        "build:no_website": "wsrun build $PKG --fast-exit -r --stages --exclude @0x/website",
-        "build:ci:no_website": "wsrun build:ci $PKG --fast-exit -r --stages --exclude @0x/website",
+        "build": "wsrun build $PKG --fast-exit -r --stages --exclude-missing",
+        "build:no_website": "wsrun build $PKG --fast-exit -r --stages --exclude @0x/website --exclude-missing",
+        "build:ci:no_website": "wsrun build:ci $PKG --fast-exit -r --stages --exclude @0x/website --exclude-missing",
```
fabioberger: What does `exclude-missing` do?

packages/contract-templates/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "@0x/contract-templates",
```
fabioberger: What about naming this `abi-gen-templates`? I like having all the `abi-gen` packages, grouped together when packages are sorted alphabetically, and `abi` alludes to `contracts`, so we don't really need to say it explicitly.

PR #1312, opened 2018-11-23
[@0x/instant] Fix CI build broken by instant installation tests

LogvinovLeon:
## Description

Currently the CI build is broken: [Build log](https://circleci.com/gh/0xProject/0x-monorepo/43271)
After fixing that parameter naming issue I've encountered the second one that was also present before: [Build log](https://circleci.com/gh/0xProject/0x-monorepo/42949)

This PR fixes both. @fragosti Please look at this PR as it's related to instant and I'm not sure my fix is a correct one.

## Checklist:

*   [ ] Prefix PR title with `[WIP]` if necessary.
*   [ ] Prefix PR title with bracketed package name(s) corresponding to the changed package(s). For example: `[sol-cov] Fixed bug`.
*   [ ] Add tests to cover changes as needed.
*   [ ] Update documentation as needed.
*   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #1313, opened 2018-11-23
[WIP] Contracts monorepo III

LogvinovLeon:
## Description

Refactors `@0x/contracts-core` into two packages:
* `@0x/contracts-examples` Example contracts
* `@0x/contracts-core` All other contracts

`@0x/contracts-examples` has a dependency on `@0x/contracts-core` and imports those contracts using `NPMResolver`.
`@0x/contracts-core` has a dev dependency on `@0x/contracts-examples` because those contracts are used in tests. We use wrappers and artifacts exported by that package. It's not a dep cycle for yarn because it's a dev dependency. Unfortunately wsrun doesn't share that view and it makes sense. It's a dep cycle and we can't build it. I'm still figuring it out.

Ok, all the things that examples contracts need from the core contracts are either interfaces or libraries. I can solve that dep cycle by refactoring out the interfaces!

Those contracts don't have any tests yet. Oops!

Publishing tests fail because of that [PR](https://github.com/0xProject/0x-monorepo/pull/1312)

## Types of changes

## Checklist:


LogvinovLeon:
I'll close this PR and open a new one. Gonna be faster that way. It should be done from the other side.

Issue #1316, opened 2018-11-25
[order_utils.py] allow LocalRefResolver to take advantage of cache + add additional schema for signed order


PirosB3:
## Expected Behavior
`LocalRefResolver()` only fetches the JSON schema files once per id, regardless of how many schema validations happen via `assert_valid()` function call.

## Current Behavior

a new `LocalRefResolver()` instance is created upon every call to `assert_valid()`, and `assert_valid()` calls the instance method `LocalRefResolver().resolve_from_url()` that does not take advantage of cache. This causes 1 or more files to be fetched from disk every time the `assert_valid()` operation is called. 

## Possible Solution

Below are the 2 suggestions I propose (and I have also implemented, to work on a project that heavily relies on order schema validation)

1. Remove the `LocalRefResolver` class definition from within `assert_valid()`, and define it at a module level
2. Define a module instance of `LocalRefResolver` called `LOCAL_RESOLVER`
3. Use `LOCAL_RESOLVER` inside of `assert_valid()` instead of instantiating `LocalRefResolver()`
4. Instead of calling `LocalRefResolver().resolve_from_url()` directly, it looks like we can take advantage of a LRU cached version of that function by using `LocalRefResolver().resolve()`. Please see [here](https://github.com/Julian/jsonschema/blob/fe72efe57d60d3d5083493bc15cb56a932409b15/jsonschema/validators.py#L606) and [here](https://github.com/Julian/jsonschema/blob/fe72efe57d60d3d5083493bc15cb56a932409b15/jsonschema/validators.py#L704) 

## Context

I'm working on an order validator implementation in Python. I would love to re-use your `LocalRefResolver`, and while reading the implementation I noticed these potential changes that would potentially improve `LocalRefResolver` as well as make the component reusable (the class cannot be imported from outside because it's defined inside a function).

## Your Environment
- Python 3.6.5
- `0x-order-utils==1.0.1`

feuGeneA:
Daniel, thank you very much for the contribution!!

I would be thrilled if we fixed the performance problems you've pointed out.

As for `LocalRefResolver`, I intentionally chose to make it "private."  To me it seems an implementation detail and an offering that's out of scope for this component.  But more importantly, I sort of hacked it together and didn't fully understand how the `jsonschema` package was using it, and I'm certainly not comfortable exposing such a hack for external re-use.  But I am open to having my mind changed!  I will review your PR and reconsider my position.

PR #1317, opened 2018-11-25
[order_utils.py] Feature/schema resolver cache

PirosB3:
## Description

This PR is a proposed solution for https://github.com/0xProject/0x-monorepo/issues/1316

## Testing instructions
- Perform regression on schema validation logic

## Types of changes

- Extracts `LocalRefResolver()` to the module level
- Creates an instance variable `LOCAL_RESOLVER` that is called by `assert_valid()`
- Changes the method that the resolver uses to fetch the JSON schema: from `resolve_from_url` I changed it to `resolve`, because the latter uses a LRU cache.
- Added `signedOrderSchema` to the `LocalRefResolver()` IDs.
- Added some new unit tests for my work
- Changed the unit test that runs all docstrings to use `importlib` causes my new test to fail sporadically. This is because the old  `load_module()` will also reload modules that are already loaded. Aside from this, `load_module()` is also deprecated in Python 3.4

## Checklist:

### DIFF
--- a/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
+++ b/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
@@ -8,11 +8,11 @@ from pkg_resources import resource_string
 import jsonschema
 
 
-class LocalRefResolver(jsonschema.RefResolver):
+class _LocalRefResolver(jsonschema.RefResolver):
     """Resolve package-local JSON schema id's."""
 
     def __init__(self):
-        """Initialize a new LocalRefResolver instance."""
+        """Initialize a new instance."""
         self.ref_to_file = {
             "/addressSchema": "address_schema.json",
             "/hexSchema": "hex_schema.json",
@@ -32,7 +32,7 @@ class LocalRefResolver(jsonschema.RefResolver):
         :param url: a string representing the URL of the JSON schema to fetch.
         :returns: a string representing the deserialized JSON schema
         :raises jsonschema.ValidationError: when the resource associated with
-                                            `url` does not exist.
+                   `url` does not exist.
         """
         ref = url.replace("file://", "")
         if ref in self.ref_to_file:
@@ -47,10 +47,10 @@ class LocalRefResolver(jsonschema.RefResolver):
         )
 
 
-# Instantiate the `LocalRefResolver()` only once so that `assert_valid()` can
+# Instantiate the `_LocalRefResolver()` only once so that `assert_valid()` can
 # perform multiple schema validations without reading from disk the schema
 # every time.
-LOCAL_RESOLVER = LocalRefResolver()
+_LOCAL_RESOLVER = _LocalRefResolver()
 
 
 def assert_valid(data: Mapping, schema_id: str) -> None:
@@ -70,5 +70,5 @@ def assert_valid(data: Mapping, schema_id: str) -> None:
     """
     # noqa
 
-    _, schema = LOCAL_RESOLVER.resolve(schema_id)
-    jsonschema.validate(data, schema, resolver=LOCAL_RESOLVER)
+    _, schema = _LOCAL_RESOLVER.resolve(schema_id)
+    jsonschema.validate(data, schema, resolver=_LOCAL_RESOLVER)
--- a/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
+++ b/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
@@ -12,6 +12,7 @@ class LocalRefResolver(jsonschema.RefResolver):
     """Resolve package-local JSON schema id's."""
 
     def __init__(self):
+        """Initialize a new LocalRefResolver instance."""
         self.ref_to_file = {
             "/addressSchema": "address_schema.json",
             "/hexSchema": "hex_schema.json",
@@ -30,7 +31,8 @@ class LocalRefResolver(jsonschema.RefResolver):
 
         :param url: a string representing the URL of the JSON schema to fetch.
         :returns: a string representing the deserialized JSON schema
-        :raises jsonschema.ValidationError: when the resource associated with `url` does not exist.
+        :raises jsonschema.ValidationError: when the resource associated with
+                                            `url` does not exist.
         """
         ref = url.replace("file://", "")
         if ref in self.ref_to_file:
@@ -45,8 +47,9 @@ class LocalRefResolver(jsonschema.RefResolver):
         )
 
 
-# Instantiate the `LocalRefResolver()` only once so that `assert_valid()` can perform multiple schema validations without
-# reading from disk the schema every time.
+# Instantiate the `LocalRefResolver()` only once so that `assert_valid()` can
+# perform multiple schema validations without reading from disk the schema
+# every time.
 LOCAL_RESOLVER = LocalRefResolver()
 
 
--- a/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
+++ b/python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
@@ -8,6 +8,48 @@ from pkg_resources import resource_string
 import jsonschema
 
 
+class LocalRefResolver(jsonschema.RefResolver):
+    """Resolve package-local JSON schema id's."""
+
+    def __init__(self):
+        self.ref_to_file = {
+            "/addressSchema": "address_schema.json",
+            "/hexSchema": "hex_schema.json",
+            "/orderSchema": "order_schema.json",
+            "/wholeNumberSchema": "whole_number_schema.json",
+            "/ECSignature": "ec_signature_schema.json",
+            "/signedOrderSchema": "signed_order_schema.json",
+            "/ecSignatureParameterSchema": (
+                "ec_signature_parameter_schema.json" + ""
+            ),
+        }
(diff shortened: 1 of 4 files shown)


PirosB3:
Hi, @feuGeneA thanks for the response.

From a performance perspective, I can see how we can still hide `LocalRefResolver` and create an instance of the class that can be used by one or more runs of `assert_valid()`. The implementation I'm thinking, unfortunately, requires a global and could potentially cause race conditions if it's used in multithread envirornment. Would you mind telling me the idea that you had?

Aside from this, I'm building an order validator (both schema and on-chain) and, as part of this change, I wanted to take advantage of new schema IDs inside the 0x repository (in this case `signedOrder`). I noticed that instead of re-building a similar component to yours I could re-use `LocalRefResolver` and simply make it compatible with more schemas, and also add improved cache so it does not read from disk.

For some reason I am not able to access CI? if I open Circle I see `Error response from daemon: unauthorized: authentication required`. Is this expected?

Thanks!
Dan

feuGeneA:
Okay, I myself am still getting a feel for what's "pythonic" and what isn't, and I was confusing myself.  The idea I had does not make sense.

What do you think about a solution like https://stackoverflow.com/a/279597/406289 ?  That is, do something like
```
def assert_valid(...):
    class LocalRefResolver...
    if not hasattr(assert_valid, "resolver"):
        assert_valid.resolver = LocalRefResolver()
    ...
    jsonschema.validate(data, schema, resolver=assert_valid.resolver)
```

I do love the encapsulation this provides.  However, if this is too anti-idiomatic, then I guess I'm okay with the module-level declaration and instantiation, but if we do that then we should definitely prefix the class and instance identifiers with an underscore, to scare away anyone wanting to import them.

FYI, when I wrote `zero_ex.json_schemas` I only added support for the schemas that I needed to use at that time.  I expected to add more as we need them, so we can most definitely add whichever ones you need, regardless of where we go with the cache/optimization.

Regarding your CI error, no, that's not expected.  I can't tell what's wrong just from that message, but if you want to ping me (@gene) on [our community chat](https://chat.0xproject.com/), I'd be happy to work with you to figure it out.

PirosB3:
Hi @feuGeneA thanks again for the lengthy response!

While hiding attributes within functions aren't necessarily Pythonic, it's definitely done in some frameworks like Django. I personally like the module level instance definition pattern because it's less code and has less logic.

Are we sure we don't want `LocalRefResolver` to be importable externally? since it's a rather small, useful, and fairly generic component, I would imagine it being useful to other developers too? Of course the trade-off we make here is that it's something that we will need to maintain.

I made most of the changes you suggested. For some reason, I'm not getting `mypy` to run because it's not able to discover modules inside of `jsonschema`. Has this ever happened to you?

python-packages/order_utils/test/test_doctest.py
```diff
@@ -8,7 +8,13 @@
 
 
 def test_all_doctests():
-    """Gather zero_ex.* modules and doctest them."""
+    """Gather zero_ex.* modules and doctest them.
+
+    NOTE: we now use `importlib` instead of `imp` because `imp` is deprecated
```
feuGeneA: I appreciate this content, but I think its too noisy for a code comment, and instead should be in a comment on the PR or in the commit message.  And, I think you could just say "`imp` is deprecated" and leave it at that.

python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
```diff
@@ -30,7 +31,8 @@ def resolve_from_url(self, url: str) -> str:
 
         :param url: a string representing the URL of the JSON schema to fetch.
         :returns: a string representing the deserialized JSON schema
-        :raises jsonschema.ValidationError: when the resource associated with `url` does not exist.
+        :raises jsonschema.ValidationError: when the resource associated with
+                                            `url` does not exist.
```
feuGeneA: Reduce indent to be 4 spaces more than `:raises` on the line above.  Anal, yes, but this would make it consistent with the rest of the code.

python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
```diff
@@ -8,6 +8,51 @@
...
+        """Initialize a new LocalRefResolver instance."""
+        self.ref_to_file = {
+            "/addressSchema": "address_schema.json",
+            "/hexSchema": "hex_schema.json",
+            "/orderSchema": "order_schema.json",
+            "/wholeNumberSchema": "whole_number_schema.json",
+            "/ECSignature": "ec_signature_schema.json",
+            "/signedOrderSchema": "signed_order_schema.json",
+            "/ecSignatureParameterSchema": (
+                "ec_signature_parameter_schema.json" + ""
```
feuGeneA: While you're in here would you mind removing this redundant string concatenation and the parenthesis and putting this all on one line?

python-packages/order_utils/test/test_doctest.py
feuGeneA: Though, there never was any mention of "`imp`" in the code; care to clarify?

PirosB3: Of course. So, `walk_packages` returns a generator that crawls the module path. Eventually, it will yield a `ModuleInfo` tuple. The first argument (what is defined as `loader` in your for loop) is of type `importlib.abc.FileLoader`. This loader inherits from [Loader](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) whose documentation specifies that `load_module` is outdated (the reason this happens is that it uses `imp` module internally).

Documentation [here](https://docs.python.org/3/library/importlib.html#importlib.abc.Loader.load_module) does suggest to use `exec_module` instead but I failed in getting that to work (I tried `exec_module` and `create_module()` with no success). What did work is `importlib.import_module` so I ended up using that. Happy to try alternatives if you prefer!

python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
PirosB3: absolutely! it's important to keep the style consistent. Will make the change

python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
PirosB3: I have a feeling it was added there by `black` because it passes the max char length? I will take a look and make the change (or let you know)

python-packages/order_utils/test/test_doctest.py
feuGeneA: Whatever works!  I just want to make sure that there's some logical continuity between the code we had in place and what we're saying is deprecated.

PirosB3:
Thanks for the advice! I will proceed as you suggested. I will go with `_LOCAL_RESOLVER` approach.

`mypy` is complaining now because I added type hints to the `resolve_from_url` function that was previously untyped. By default, `mypy` does not do static analysis on untyped functions. I will try your suggestion of updating the `pyi` file

python-packages/order_utils/test/test_doctest.py
PirosB3: sounds great! I will remove the comment from here, and I will ensure it's documented in the PR. In that comment, instead of mentioning `imp` which is internal, I'll simply specify that `load_module` is deprecated so that it's clearer.

python-packages/order_utils/src/zero_ex/json_schemas/__init__.py
PirosB3: I confirm this is black

PirosB3:
Hi @feuGeneA 

I'm not sure what happened here, but I did a rebase and automagically that ended up adding a lot of people to review and a lot of tags. Sorry about that!
On the bright side, Python tests are successful now and I've addressed the changes we discussed. More specifically:
- Refactored `LocalRefResolver` to `_LocalRefResolver`
- Removed comment about deprecation in `test_all_doctests` and instead added it to the PR
- Did other style changes.

Let me know :)

python-packages/order_utils/test/test_json_schemas.py
```diff
@@ -0,0 +1,23 @@
...
+    on _LOCAL_RESOLVER
+    """
+    _LOCAL_RESOLVER._remote_cache.cache_clear()  # pylint: disable=W0212
+
+    assert_valid(make_empty_order(), "/orderSchema")
+    cache_info = (
+        _LOCAL_RESOLVER._remote_cache.cache_info()  # pylint: disable=W0212
+    )
+    assert cache_info.currsize == 4
+    assert cache_info.hits == 10
```
feuGeneA: What are these magic numbers `4` and `10`?  Are they coincident with our usage, or are they reflective of some implementation detail of the 3rd party `jsonschemas` package?  If the latter, I'm wary, as a change there would break this test.

feuGeneA:
> It looks like everything is passing except the `submit-coverage` job?
> 
> ```
> home/circleci/repo/node_modules/coveralls/bin/coveralls.js:18
>         throw err;
>         ^
> Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
> ```
> Do you know why this is happening?

Probably this: https://github.com/0xProject/0x-monorepo/blob/development/CONTRIBUTING.md#fix-submit-coverage-ci-failure

PirosB3:
All conflicts resolved! let me know if you need anything from me @feuGeneA

PR #1322, opened 2018-11-26
fix(order_utils.py): work aroud pylint bug

feuGeneA:
## Description

Fix error in runs of `static-tests-python`.  Apparently there's a bug in the new version of pylint that was released yesterday, which [I've reported](https://github.com/PyCQA/pylint/issues/2612).  While we await the fix for that, this PR pins pylint to the previous version.

## Testing instructions

Watch CI :smile: 

## Types of changes

## Checklist:


LogvinovLeon:
So jelly that you can add comments there. Package.json doesn't allow that ;) Thanks

PR #1328, opened 2018-11-26
[pipeline] Add additional documentation to the README

albrow:
## Description

This PR adds some additional documentation to the README. We can probably add even more information over time and possibly move some info to a different doc.

## Checklist:


packages/pipeline/README.md
```diff
@@ -45,6 +45,27 @@ environment variable `ZEROEX_DATA_PIPELINE_DB_URL` which should be a properly
 formatted
 [PostgreSQL connection url](https://stackoverflow.com/questions/3582552/postgresql-connection-url).
 
+## Test environment
```
wakkadojo: Is it cleaner to have containerized test environments? should we have simple instructions for a docker psql server?

albrow: 👍 I added Docker instructions as the preferred method. I also mentioned that there are other possible ways of installing Postgres that would work fine.

PR #1329, opened 2018-11-26
[instant] Token Selector events

steveklebanoff:
## Description

Adds all Token Selector events, and sets `numberAvailableAssets`, `selectedAssetName`, `selectedAssetData` event properties

## Checklist:


packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -29,6 +30,13 @@ export class ZeroExInstantContainer extends React.Component<{}, ZeroExInstantCon
     public state = {
         tokenSelectionPanelAnimationState: 'none' as SlideAnimationState,
     };
+    private _handlePanelCloseClickedX: () => void;
+    private _handlePanelCloseAfterChose: () => void;
+    public constructor(props: {}) {
```
BMillman19: can we just name the props but keep it empty? this is what we usually do for other components.

packages/instant/src/components/zero_ex_instant_container.tsx
```diff
@@ -29,6 +30,13 @@ export class ZeroExInstantContainer extends React.Component<{}, ZeroExInstantCon
...
+        super(props);
+        this._handlePanelCloseClickedX = this._handlePanelClose.bind(this, TokenSelectorClosedVia.ClickedX);
+        this._handlePanelCloseAfterChose = this._handlePanelClose.bind(this, TokenSelectorClosedVia.TokenChose);
```
BMillman19: can we just make these they're own functions with implementations below? i.e.

```javascript
private readonly _handlePanelCloseClickedX = (): void => {
    this._handlePanelClose(TokenSelectorClosedVia.ClickedX);
};
```

I this is clearer

packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -53,6 +53,30 @@ export const analyticsMiddleware: Middleware = store => next => middlewareAction
                 ).toString();
                 analytics.addUserProperties({ ethBalanceInUnitAmount });
             }
+            break;
+        case ActionTypes.UPDATE_SELECTED_ASSET:
+            const selectedAsset = curState.selectedAsset;
+            if (selectedAsset) {
+                const assetName = selectedAsset.metaData.name;
```
BMillman19: I think asset data only is ok? right now by coincidence we have the `name` property available in both `ERC20AssetMetaData` and `ERC721AssetMetaData` but I can see future asset types not necessarily having a "name". Assetdata should be good enough for us to differentiate the assets.

packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -53,6 +53,30 @@ export const analyticsMiddleware: Middleware = store => next => middlewareAction
                 ).toString();
                 analytics.addUserProperties({ ethBalanceInUnitAmount });
             }
+            break;
+        case ActionTypes.UPDATE_SELECTED_ASSET:
```
BMillman19: What if we default open to a certain asset?

BMillman19: I see it should be handled by the `zero_ex_instant_provider.tsx` change above, correct?

steveklebanoff: Yes, correct

packages/instant/src/redux/analytics_middleware.ts
steveklebanoff: I think being able to easily view the name in heap instead of having to reference an asset data mapping would be a huge benefit

packages/instant/src/components/zero_ex_instant_container.tsx
steveklebanoff: I can do this, but they both would have to call 
```
        this.setState({
            tokenSelectionPanelAnimationState: 'slidOut',
        });
```
or a helper function which does this, which feels a little repetitive.  Would prefer to keep the existing implementation, but can change if you feel strongly

PR #1342, opened 2018-11-27
[instant] Close and View Transaction analytics events

steveklebanoff:
## Description

Analytics events for closing Instant and viewing txn

## Checklist:


packages/instant/src/index.umd.ts
```diff
@@ -57,6 +58,7 @@ const renderInstant = (config: ZeroExInstantConfig, selector: string) => {
     injectedDiv.setAttribute('class', INJECTED_DIV_CLASS);
     appendTo.appendChild(injectedDiv);
     const closeInstant = () => {
+        analytics.trackInstantClosed();
```
BMillman19: Do we want to put this in the `instantOverlayProps` below? otherwise we won't get a close event if `config.shouldDisablePushToHistory` is `false`

steveklebanoff: I checked and it seems to be triggering even w/ `shouldDisablePushToHistory` set to `false`.  lmk if I'm missing something though

Here's my understanding of why this works:
- `renderInstant` returns the `closeInstant` function
- The `onPopStateHandler` holds onto the return value of `renderInstant` in a function called `removeInstant`, and triggers this when the user hits the back button

fragosti: I think @BMillman19 is right, are you sure it's triggering?

steveklebanoff: I'll take another look

steveklebanoff: So this is what I did:

```diff
diff --git a/packages/instant/public/index.html b/packages/instant/public/index.html
index df39994ef..01e3e49d8 100644
--- a/packages/instant/public/index.html
+++ b/packages/instant/public/index.html
@@ -174,7 +174,7 @@
                     availableAssetDatas: availableAssetDatasString ? JSON.parse(availableAssetDatasString) : undefined,
                     defaultSelectedAssetData: queryParams.getQueryParamValue('defaultSelectedAssetData'),
                     affiliateInfo: affiliateInfoOverride,
-                    shouldDisablePushToHistory: !!queryParams.getQueryParamValue('shouldDisablePushToHistory'),
+                    shouldDisablePushToHistory: false,
                 };
                 return renderOptionsOverrides;
             };
diff --git a/packages/instant/src/index.umd.ts b/packages/instant/src/index.umd.ts
index 667daf441..da57b4e3c 100644
--- a/packages/instant/src/index.umd.ts
+++ b/packages/instant/src/index.umd.ts
@@ -50,6 +50,7 @@ const validateInstantRenderConfig = (config: ZeroExInstantConfig, selector: stri

 // Render instant and return a callback that allows you to remove it from the DOM.
 const renderInstant = (config: ZeroExInstantConfig, selector: string) => {
+    console.log('shouldDisablePushToHistory', config.shouldDisablePushToHistory);
     const appendToIfExists = document.querySelector(selector);
     assert.assert(!_.isNull(appendToIfExists), `Could not find div with selector: ${selector}`);
     const appendTo = appendToIfExists as Element;
@@ -58,6 +59,7 @@ const renderInstant = (config: ZeroExInstantConfig, selector: string) => {
     injectedDiv.setAttribute('class', INJECTED_DIV_CLASS);
     appendTo.appendChild(injectedDiv);
     const closeInstant = () => {
+        console.log('*track closed*');
         analytics.trackInstantClosed();
         if (!_.isUndefined(config.onClose)) {
             config.onClose();
```

I loaded the new public/index.html, closed via clicking the "X", and saw the following in the console:
```
shouldDisablePushToHistory false
*track closed*
```

So, in my understanding: this works.  But it sounds like there may be another piece to this I'm missing?

PR #1343, opened 2018-11-27
[instant] Quote analytics

steveklebanoff:
## Description

Track successful quote fetches, and errors in fetching quotes

## Checklist:


packages/instant/src/util/analytics.ts
```diff
@@ -177,4 +180,24 @@ export const analytics = {
...
+    trackQuoteError: (
+        errorMessage: string,
+        assetName: string,
+        assetData: string,
+        assetAmount: BigNumber,
+        fetchedVia: QuoteFetchedVia,
+    ) => {
+        trackingEventFnWithPayload(EventNames.QUOTE_ERROR)({
+            errorMessage,
+            assetName,
```
BMillman19: Do we need `assetName` and `assetData` here?

packages/instant/src/util/buy_quote_updater.ts
```diff
@@ -31,6 +37,13 @@ export const buyQuoteUpdater = {
         } catch (error) {
             if (options.dispatchErrors) {
                 dispatch(actions.setQuoteRequestStateFailure());
+                analytics.trackQuoteError(
+                    error.message ? error.message : 'other',
+                    asset.metaData.name,
+                    asset.assetData,
+                    assetUnitAmount,
```
BMillman19: Let's log the asset amount in wei (base units) because we're logging wei amounts for buyQuoteProperties

packages/instant/src/util/analytics.ts
steveklebanoff: ah good call, we should be able to rely on `selectedAsset` event properties here, unless there's a race condition

packages/instant/src/util/buy_quote_updater.ts
steveklebanoff: Good call that these should be consistent. I'm wondering though: would it be better to instead convert the `buyQuoteProperties` to report in unit amounts instead?  I feel like we usually think in these amounts, and it'd be easier to not have to do the math when viewing items.  Let me know what you think.

packages/instant/src/types.ts
```diff
@@ -21,6 +21,11 @@ export enum OrderProcessState {
     Failure = 'FAILURE',
 }
 
+export enum QuoteFetchedVia {
+    Manual = 'Manual',
+    Heartbeat = 'Heartbeat',
+}
```
fragosti: `QuoteFetchMethod`, `QuoteFetchOrigin`, `QuoteFetchReason`?

packages/instant/src/redux/async_data.ts
```diff
@@ -99,7 +99,12 @@ export const asyncData = {
                 dispatch,
                 selectedAsset as ERC20Asset,
                 selectedAssetUnitAmount,
-                { setPending: !options.updateSilently, dispatchErrors: !options.updateSilently, affiliateInfo },
+                {
+                    setPending: !options.updateSilently,
+                    dispatchErrors: !options.updateSilently,
+                    fetchedVia: options.fetchedVia,
```
fragosti: nit: it doesn't seem like this is an optional parameter right? There isn't a sensible default value.

packages/instant/src/util/buy_quote_updater.ts
BMillman19: I think its better to report base units because we will need additional context of token decimals to get the true amount being bought.

packages/instant/src/util/analytics.ts
```diff
@@ -190,4 +201,16 @@ export const analytics = {
...
         trackingEventFnWithPayload(EventNames.TOKEN_SELECTOR_SEARCHED)({ searchText }),
+    trackQuoteFetched: (buyQuote: BuyQuote, fetchOrigin: QuoteFetchOrigin) =>
+        trackingEventFnWithPayload(EventNames.QUOTE_FETCHED)({
+            ...buyQuoteEventProperties(buyQuote),
+            fetchOrigin,
+        }),
+    trackQuoteError: (errorMessage: string, assetAmount: BigNumber, fetchOrigin: QuoteFetchOrigin) => {
+        trackingEventFnWithPayload(EventNames.QUOTE_ERROR)({
+            errorMessage,
+            assetAmount: assetAmount.toString(),
```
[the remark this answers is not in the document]
steveklebanoff: yea, good idea

PR #1345, opened 2018-11-28
[instant] Progress bar background

steveklebanoff:
## Description

Make progress bar background 10% opacity of primary color

## Checklist:

### DIFF
--- a/packages/instant/src/components/timed_progress_bar.tsx
+++ b/packages/instant/src/components/timed_progress_bar.tsx
@@ -1,8 +1,9 @@
 import * as _ from 'lodash';
+import { transparentize } from 'polished';
 import * as React from 'react';
 
 import { PROGRESS_FINISH_ANIMATION_TIME_MS, PROGRESS_STALL_AT_WIDTH } from '../constants';
-import { ColorOption, css, keyframes, styled } from '../style/theme';
+import { ColorOption, css, keyframes, styled, ThemeConsumer } from '../style/theme';
 
 import { Container } from './ui/container';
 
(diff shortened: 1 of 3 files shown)


packages/instant/src/components/timed_progress_bar.tsx
```diff
@@ -89,11 +91,23 @@ export const Progress =
...
+        opacity: 0.1;
+        width: 100%;
+    }
+`;
+
 export interface ProgressBarProps extends ProgressProps {}
 
 export const ProgressBar: React.ComponentType<ProgressBarProps & React.ClassAttributes<{}>> = React.forwardRef(
     (props, ref) => (
-        <Container width="100%" backgroundColor={ColorOption.lightGrey} borderRadius="6px">
```
BMillman19: Why don't we just transform the color and set it as the background color of this container instead of creating a new div?

steveklebanoff: That is what I wanted to do, but couldn't find a good way.  There may be an approach I'm missing tho, lemme know:
- If I set an opacity on this container (line 96), it affects the opacity of the moving progress bar too
- There doesn't seem to be a good way to just set the background color opacity to a value
- The primary color could be a hex value and not rgba, and I figured I'd have to bring in additional dependency to generate a new value w/ a specificed opacity

BMillman19: https://polished.js.org/docs/#opacify

BMillman19: https://polished.js.org/docs/#transparentize rather

steveklebanoff: Done!

PR #1348, opened 2018-11-28
[instant] Remove commonjs bundle, productionize UMD bundle

fragosti:
## Description

Should I change the CDN settings to redirect HTTP to HTTPS?

https://instant.0xproject.com/instant.js

## Checklist:

### DIFF
--- a/packages/instant/.production.discharge.json
+++ b/packages/instant/.production.discharge.json
@@ -1,0 +1,13 @@
+{
+    "domain": "instant.0xproject.com",
+    "build_command": "yarn build",
+    "upload_directory": "umd",
+    "index_key": "instant.js",
+    "error_key": "404.html",
+    "trailing_slashes": true,
+    "cache": 3600,
+    "aws_profile": "default",
+    "aws_region": "us-east-1",
+    "cdn": true,
+    "dns_configured": true
+}
--- a/packages/instant/README.md
+++ b/packages/instant/README.md
@@ -41,7 +41,7 @@ Finally, we have our live production bundle that is only meant to be updated wit
 To build and deploy to this bundle, run
 
 ```
-yarn deploy_live
+yarn deploy_production
 ```
 
 **NOTE: On deploying the site to staging and dogfood, it will say the site is available at a non-existent URL. Please ignore and use the (now updated) URL above.**
--- a/packages/instant/README.md
+++ b/packages/instant/README.md
@@ -22,18 +22,26 @@ The package is available as a UMD module named `zeroExInstant`.
 
 You can deploy a work-in-progress version of 0x Instant at http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com for easy sharing.
 
-To build and deploy the site run
+To build and deploy the bundle run
 
 ```
 yarn deploy_dogfood
 ```
 
-We also have a staging bucket that is to be updated less frequently can be used to share instant externally: http://0x-instant-staging.s3-website-us-east-1.amazonaws.com/
+We also have a staging bucket that is to be updated less frequently can be used to share a beta version of instant externally: http://0x-instant-staging.s3-website-us-east-1.amazonaws.com/
 
-To build and deploy to this bucket, run
+To build and deploy to this bundle, run
 
 ```
 yarn deploy_staging
+```
+
+Finally, we have our live production bundle that is only meant to be updated with stable, polished releases.
+
+To build and deploy to this bundle, run
+
+```
+yarn deploy_live
 ```
 
 **NOTE: On deploying the site, it will say the site is available at a non-existent URL. Please ignore and use the (now updated) URL above.**
--- a/packages/instant/README.md
+++ b/packages/instant/README.md
@@ -2,11 +2,11 @@
 
 ## Installation
 
-The package is available as a UMD module named `zeroExInstant`.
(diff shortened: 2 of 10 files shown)


packages/instant/src/util/version.ts
```diff
@@ -0,0 +1,8 @@
+const anyWindow = window as any;
+
+export const versionUtil = {
+    writeVersionInfoToWindow: () => {
+        anyWindow.__zeroExInstantGitSha = process.env.GIT_SHA;
```
steveklebanoff: Can we write this on the `zeroExInstant` window object? I feel like that'd be much cleaner.

packages/instant/.production.discharge.json
```diff
@@ -0,0 +1,13 @@
+{
+    "domain": "instant.0xproject.com",
+    "build_command": "yarn build",
```
steveklebanoff: I think we should change add  `--env.production` but I can do this when I update my heap PR ( see staging and dogfood env setting here: https://github.com/0xProject/0x-monorepo/pull/1295/files )

packages/instant/README.md
```diff
@@ -2,35 +2,7 @@
...
-```json
-"compilerOptions": {
-    "typeRoots": ["node_modules/@0x/typescript-typings/types", "node_modules/@types"],
-}
-```
-
-**UMD Module**
-
-The package is also available as a UMD module named `zeroExInstant`.
+The package is available as a UMD module named `zeroExInstant`.
```
steveklebanoff: Time to add the production URL in the README here?

packages/instant/package.json
```diff
@@ -5,13 +5,10 @@
...
-    "main": "lib/index.js",
-    "types": "lib/index.d.ts",
+    "main": "umd/instant.js",
+    "private": true,
     "scripts": {
-        "build": "yarn build:all",
-        "build:all": "run-p build:umd:prod build:commonjs",
-        "build:umd:prod": "webpack --mode production",
-        "build:commonjs": "tsc -b",
+        "build": "webpack --mode production",
```
steveklebanoff: simpler build process 💯

packages/instant/src/util/version.ts
fragosti: This is a great idea

packages/instant/.production.discharge.json
```diff
@@ -0,0 +1,13 @@
...
+    "upload_directory": "umd",
+    "index_key": "instant.js",
+    "error_key": "404.html",
+    "trailing_slashes": true,
+    "cache": 3600,
+    "aws_profile": "default",
```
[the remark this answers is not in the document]
BMillman19: ok just looked at the other discharge configs and it looks like we left them using `default`

we need to update the other ones with the correct build command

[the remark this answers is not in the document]
steveklebanoff: the 0xproject change is in the rollbar PR which I haven't merged yet ( https://github.com/0xProject/0x-monorepo/pull/1295 )

packages/instant/src/index.umd.ts
```diff
@@ -110,3 +110,7 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
     };
     window.onpopstate = onPopStateHandler;
 };
+
+// Write version info to the exported object for debugging
+export const GitSha = process.env.GIT_SHA;
```
steveklebanoff: nit but I would expect these to be in all caps constant-case `GIT_SHA`, but there may be an accepted pattern for this that 0x uses elsewhere?

packages/instant/package.json
```diff
@@ -5,13 +5,10 @@
...
         "build:ci": "yarn build",
         "watch_without_deps": "tsc -w",
```
steveklebanoff: Is this command relevant any more?

packages/instant/.production.discharge.json
```diff
@@ -0,0 +1,13 @@
+{
+    "domain": "instant.0xproject.com",
+    "build_command": "yarn build",
+    "upload_directory": "umd",
+    "index_key": "instant.js",
+    "error_key": "404.html",
+    "trailing_slashes": true,
+    "cache": 3600,
```
steveklebanoff: Do we feel comfortable with this cache setting? Is there a way we could kill the cache if we had an urgent bug?

steveklebanoff:
Also, we should remove the comment about the common js bundle in `webpack.config.js`

packages/instant/.production.discharge.json
fragosti: discharge also does it when you deploy a new version

steveklebanoff: ok, sweet

packages/instant/package.json
```diff
@@ -23,6 +19,7 @@
         "clean": "shx rm -rf lib coverage scripts",
         "deploy_dogfood": "discharge deploy -c .dogfood.discharge.json",
         "deploy_staging": "discharge deploy -c .staging.discharge.json",
+        "deploy_live": "discharge deploy -c .production.discharge.json",
```
steveklebanoff: nit but I would have expected this to be called `deploy_production`

fragosti: yea... it's meant to match the website but on second thought that isn't a good enough reason.

steveklebanoff:
@fragosti Have you tried setting the compilerOption `declaration` to `false` to exclude the building of `.d.ts`?

PR #1349, opened 2018-11-28
[WIP] pull OHLCV records from Crypto Compare

xianny:
## Description

It scrapes OHLCV records from Crypto Compare and saves to postgres. 

There are some constants of interest that we could convert into env variables:
* EARLIEST_BACKFILL_TIME; this is how far back the script will try to backfill in the absence of any previous data. It's hardcoded to Sept 2010 for now. Ideally we would run once with that, then set it to something more recent. If there are any pairs that don't have data on CC, the script will always try to fetch from this date.

* MAX_CONCURRENT_REQUESTS; this is hardcoded to 18. Should tweak to make better use of the CC rate limit

* TO_CURRENCIES: fiat/major currencies to compare tokens with. It's ETH, USD, and EUR right now.

Workflow:
1. Query events table for token addresses
2. Query Crypto Compare for token symbols
3. Generate trading pairs as {from: token, to: fiat}
4. Fetch latest scraped time for all pairs or set to EARLIEST_BACKFILL_TIME
5. For each pair, fetch from Crypto Compare and save in chronologically ascending order

## Testing instructions

There are some unit tests but I would like to add more test coverage. 
You can build and run (in pipeline package) 
```
$ node lib/src/scripts/pull_ohlcv_cryptocompare.js
```

Then check your postgres db for data. Recommend setting EARLIEST_BACKFILL_DATE to something earlier, e.g. a few months back. Run twice to test that backfill functionality works as expected.

## Types of changes

## Checklist:


xianny:
Updated OP with details.

I would really like to add more unit tests, but this runs and works and we can test by running the script locally. 

Please review and be extra nitpicky! First code contribution to this repo :)

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function parseRecords(rawRecords: CryptoCompareOHLCVRecord[], metadata: OHLCVMetadata): OHLCVExternal[] {
+    if (rawRecords.length > 1) {
+        const intervalInSeconds = Math.abs(rawRecords[0].time - rawRecords[1].time);
+        return rawRecords.map(rec => {
+            const ohlcvRecord = new OHLCVExternal();
+            ohlcvRecord.exchange = metadata.exchange;
+            ohlcvRecord.fromSymbol = metadata.fromSymbol;
+            ohlcvRecord.toSymbol = metadata.toSymbol;
+            ohlcvRecord.startTime = (rec.time - intervalInSeconds) * ONE_SECOND;
```
fabioberger: We use the same `interval` as observed in the first two records for all records?

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,92 @@
...
+const EARLIEST_BACKFILL_TIME = new Date('2010-09-01').getTime(); // the time when BTC/USD info starts appearing on Crypto Compare
+
+let connection: Connection;
+
+(async () => {
+    connection = await createConnection(ormConfig as ConnectionOptions);
+    const repository = connection.getRepository(OHLCVExternal);
+    const source = new CryptoCompareOHLCVSource(MAX_CONCURRENT_REQUESTS);
+
+    const tradingPairs = await fetchOHLCVTradingPairs(connection, SOURCE_NAME, EARLIEST_BACKFILL_TIME);
```
fabioberger: Always suffix methods that are async with `Async`, e.g `fetchOHLCVTradingPairsAsync`

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,92 @@
...
+    console.log(`Starting ${tradingPairs.length} job(s) to scrape Crypto Compare for OHLCV records...`);
+
+    const fetchAndSavePromises = tradingPairs.map(async pair => {
```
fabioberger: `_.map`

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,92 @@
...
+        const pairs = source.generateBackfillIntervals(pair);
+        return fetchAndSaveAsync(source, repository, pairs);
+    });
+    await Promise.all(fetchAndSavePromises);
```
fabioberger: Do we need to limit the number of concurrent requests here?

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,92 @@
...
+            fromSymbol: p.fromSymbol,
+            toSymbol: p.toSymbol,
+            source: SOURCE_NAME,
+            observedTimestamp: new Date().getTime(),
+        };
+        const parsedRecords = parseRecords(rawRecords, metadata);
+        try {
+            await saveRecordsAsync(repository, parsedRecords);
+            i++;
+        } catch (e) {
```
fabioberger: Our convention is to name `error` variables as `err` everywhere.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+        SELECT DISTINCT(taker_token_address) as tokenaddress FROM raw.exchange_fill_events
+        LIMIT 1`,
+    );
+    const tokenAddresses = rawTokenAddresses.map(obj => obj.tokenaddress);
+
+    // get token symbols used by Crypto Compare
+    const allCoins = await axios.get<CryptoCompareCoinListResp>(COINLIST_API);
+    const erc20Coins: Map<string, string> = new Map();
+    Object.entries(allCoins.data.Data).forEach(pair => {
+        const [k, v] = pair;
```
fabioberger: What do `k` and `v` stand for? Please try to use descriptive variable names to improve readability. Is `k` actually the coin symbol? And `v` the coin data?

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+        if (v.BuiltOn === '7605' && v.SmartContractAddress !== 'N/A') {
```
fabioberger: Let's set `'7695'` to a constant called `ETHEREUM_IDENTIFIER` and use that everywhere, instead of hard-coding this very strange constant.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+              to_symbol = '${pair.toSymbol}' AND
+              source = '${source}'`;
+        })
+        .join('\nUNION\n');
+
+    const latestTradingPairsResult: Array<{
+        from_symbol: string;
+        to_symbol: string;
+        latest: string;
+    }> = await conn.query(tradingPairsLatestQuery);
```
fabioberger: Any chance we can wrap `conn.query` so that we can append `Async` to the method name? It's a really nasty bug to forget to `await` an async method, and also really easy to make that mistake. It really helps to have the suffix convention that we do, but for external libraries without the convention, it's worth wrapping them to keep our codebase consistent.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const latestTradingPairs = latestTradingPairsResult.map(p => ({
+        fromSymbol: p.from_symbol,
+        toSymbol: p.to_symbol,
+        // tslint:disable:custom-no-magic-numbers
+        latest: parseInt(p.latest, 10),
```
fabioberger: Use `_.parseInt` and omit the `10`

packages/pipeline/test/data_sources/ohlcv_external/crypto_compare_test.ts
```diff
@@ -0,0 +1,51 @@
...
+            const expected = [pair];
+
+            const actual = source.generateBackfillIntervals(pair);
+            expect(actual).deep.equal(expected);
+        });
+    });
+
+    describe('getAsync', () => {
+        // tslint:disable:no-empty
+        // todo
```
fabioberger: Still open TODO?

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,88 @@
...
+            toTs: Math.floor((pair.latest + this.interval) / ONE_SECOND), // CryptoCompare uses timestamp in seconds. not ms
+        };
+
+        const fetchPromise = this._plimit(() => {
+            // tslint:disable:no-console
+            console.log(`Scraping Crypto Compare with ${JSON.stringify(params)}`);
+            return axios.get<CryptoCompareOHLCVResponse>(this._url, { params });
+        });
+
+        const resp = await Promise.resolve(fetchPromise);
```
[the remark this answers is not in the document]
xianny: I'm referencing a single instance of the plimit function as a private member scoped to the whole class, so that when we instantiate this class it enforces rate limiting for all reqs to the API.

I think your understanding is correct. `this._plimit(() => asyncFunction)` returns a promise wrapping the result of async function. The plimit example code is structured as:
```
Promise.all([
    plimit(fn1),
    plimit(fn2),
    plimit(fn3),
...])
```

But since I'm only calling one function here, I'm using `Promise.resolve` instead of `Promise.all`. 

Does that make sense? I'm open to renaming stuff to make the code more understandable too

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,88 @@
...
+
+        if (Object.keys(resp.data.Data).length < 2) {
+            console.log(`No data found for ${JSON.stringify(params)}`);
```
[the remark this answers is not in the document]
xianny: I disabled no-console - didn't know about logUtils! Thank you!

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
xianny: Yeah. This is because we call the specific CC endpoint for the interval we want. E.g. there's a different endpoint for hourly, weekly, etc. 

On second thought, I should just put the interval field in the metadata to simplify things.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,88 @@
...
+            e: this.default_exchange,
+            fsym: pair.fromSymbol,
+            tsym: pair.toSymbol,
+            toTs: Math.floor((pair.latest + this.interval) / ONE_SECOND), // CryptoCompare uses timestamp in seconds. not ms
+        };
+
+        const fetchPromise = this._plimit(() => {
+            // tslint:disable:no-console
+            console.log(`Scraping Crypto Compare with ${JSON.stringify(params)}`);
+            return axios.get<CryptoCompareOHLCVResponse>(this._url, { params });
```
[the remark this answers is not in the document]
albrow: I just didn't know about `fetchAsync`. Does it handle URL-encoded parameters?

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
xianny: Yep! We are limiting in the `source` instance using `plimit`.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: sorry, my bad! yes, renamed.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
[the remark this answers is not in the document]
albrow: I would argue that `logUtils.log` should not exist. Let me explain.

Our goal with the `no-console` TSLint rule is to disallow logging in code where it does not belong. The existence of `logUtils.log` is counteractive to this goal because it *implicitly* bypasses the linter rule. In other words, if you call `logUtils.log` in your code, the linter will not warn you. As we add more and more developers on the team, someone is bound to call `logUtils.log` in a place where it really does not belong and they will have no idea that it is against our conventions. They will probably just think that `logUtils.log` is what we use for logging.

Here's what I propose instead:

Just use `console.log`. Normally the linter will warn you, but you can explicitly disable it if you feel that you are working on a part of the codebase where logging is appropriate. That *explicit* step of adding a `// tslint:disable` comment is important because it makes it less likely that someone will do this by accident.

@fabioberger what do you think?

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
albrow: @fabioberger this doesn't seem to be a convention that we follow throughout our codebase. It seems like forgetting to add `await` is about as likely as forgetting to wrap the function and append `Async`, so I'm not sure what this buys us.

Instead I would propose we improve our linter rules to do either one of the following:

1. If a function returns a promise, you must call it `await`. This is what we want to do 99% of the time and you can always disable the linter rule in the rare case where you intended to call the function without `await`. There may even already be a rule that we could copy/use.
2. If the first option is not possible, we can add a linter rule to enforce wrapping of external functions with an `Async` prefix.

What do you think?

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,42 @@
...
+            ohlcvRecord.endTime = rec.time * ONE_SECOND;
```
wakkadojo: I'm pretty sure you got the start time / end time correct wrt what the time in CC means, just wanted to get confirm from you that end time in CC = the timestamp they post for an interval

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+                return [];
+            }
+        }, TO_CURRENCIES);
+    };
+    const tradingPairs: StaticPair[] = chain(mapFn, tokenAddresses);
+
+    const tradingPairsLatestQuery: string = tradingPairs
+        .map(pair => {
+            return `SELECT
+              case COUNT(*) when 0 then ${earliestBackfillTime} else MAX(end_time) end AS latest,
```
wakkadojo: How does CC deal with start time /etc? Will this leave hour gaps or cause hour overlaps?

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+                ];
+            } else {
+                return [];
+            }
+        }, TO_CURRENCIES);
+    };
+    const tradingPairs: StaticPair[] = chain(mapFn, tokenAddresses);
+
+    const tradingPairsLatestQuery: string = tradingPairs
+        .map(pair => {
```
wakkadojo: CC's latest data points are often revised over the course of 30 mins-sh. I don't think your pull script is handling this. i.e. maybe throw away the last data point before writing to avoid this hairy edge case

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+    Data: Map<string, CryptoCompareCoin>;
+}
+
+interface CryptoCompareCoin {
+    Symbol: string;
+    BuiltOn: string;
+    SmartContractAddress: string;
+}
+
+const TO_CURRENCIES = ['USD', 'EUR', 'ETH'];
```
wakkadojo: Can you include USDT here? So much token / token volume happens on binance and i think USDT is critical to look at.

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,42 @@
...
+    source: string;
+    observedTimestamp: number;
+}
+/**
+ * Parses OHLCV records from Crypto Compare into an array of OHLCVExternal entities
+ * @param rawRecords an array of OHLCV records from Crypto Compare (not the full response)
+ */
+export function parseRecords(rawRecords: CryptoCompareOHLCVRecord[], metadata: OHLCVMetadata): OHLCVExternal[] {
+    if (rawRecords.length > 1) {
+        const intervalInSeconds = Math.abs(rawRecords[0].time - rawRecords[1].time);
```
wakkadojo: i wrote a utility function to convert current time to unix epoch time to interface w/ cc. since you don't need the utility function, can you delete it in this PR?

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
xianny: Yup, that was my understanding. The time in the API response is the end-time we save.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,88 @@
...
+// tslint:disable:custom-no-magic-numbers
+const ONE_WEEK = 7 * 24 * 60 * 60 * 1000;
+const ONE_SECOND = 1000;
+
+const HISTORICAL_HOURLY_OHLCV_URI = 'https://min-api.cryptocompare.com/data/histohour';
+
+export class CryptoCompareOHLCVSource {
+    public readonly interval = ONE_WEEK; // the hourly API returns data for one week at a time
+    public readonly default_exchange = 'CCCAGG';
+    private readonly _url: string = HISTORICAL_HOURLY_OHLCV_URI;
```
albrow: Is there a reason to set `_url` as an instance variable when it is already defined at the top of the file?

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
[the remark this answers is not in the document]
albrow: @xianny I think renaming some things and possibly adding some comments would greatly help comprehensibility. Most people on the team will have never encountered `p-limit` before.

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,92 @@
...
+async function saveRecordsAsync(repository: Repository<OHLCVExternal>, records: OHLCVExternal[]): Promise<void> {
+    const metadata = [
+        records[0].fromSymbol,
+        records[0].toSymbol,
+        new Date(records[0].startTime),
+        new Date(records[records.length - 1].endTime),
+    ];
+
+    console.log(`Saving ${records.length} records to ${repository.metadata.name}... ${JSON.stringify(metadata)}`);
+    await repository.insert(records);
```
albrow: 1. Should we insert these records in batches instead of all at once? See [here for an example](https://github.com/0xProject/0x-monorepo/blob/73ad482e14b00fe37a33d931d781bcb083f11077/packages/pipeline/src/scripts/pull_competing_dex_trades.ts#L36).
2. Do we really mean to do an `insert` here instead of a `save`? The latter will handle primary key constraint violations gracefully but the former will not. Is there any scenario where we might get the same data more than once from subsequent script runs?

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+        const [k, v] = pair;
+        if (v.BuiltOn === '7605' && v.SmartContractAddress !== 'N/A') {
+            erc20Coins.set(v.SmartContractAddress, k);
+        }
+    });
+
+    const mapFn = (tokenAddress: string): StaticPair[] => {
+        return chain(fiat => {
+            const fromSymbol = erc20Coins.get(tokenAddress);
+            if (!!fromSymbol) {
```
albrow: `!!fromSymbol` implicitly coerces `fromSymbol` into a boolean type. This can lead to some surprising and confusing behavior because of how JavaScript treats "truthiness". Can we do an explicit comparison instead? Something like:

```
if (fromSymbol !== '')
```

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -0,0 +1,100 @@
...
+    const allCoins = await axios.get<CryptoCompareCoinListResp>(COINLIST_API);
+    const erc20Coins: Map<string, string> = new Map();
+    Object.entries(allCoins.data.Data).forEach(pair => {
+        const [k, v] = pair;
+        if (v.BuiltOn === '7605' && v.SmartContractAddress !== 'N/A') {
+            erc20Coins.set(v.SmartContractAddress, k);
+        }
+    });
+
+    const mapFn = (tokenAddress: string): StaticPair[] => {
```
albrow: Instead of `mapFn` can we rename this to something more descriptive?

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
albrow: This kind of query construction makes me nervous and seems prone to error. Can we:

1. Potentially split this into multiple queries (maybe even executing them in a single transaction)?
2. Use prepared statements and parameters instead of string interpolation? As is, the strings that are part of our query are totally unescaped. This leaves open the possibility of SQL injection. Or in less nefarious scenarios, unexpected symbols in the query could cause accidental data loss or incorrect results. Let me know if you have any questions about how to do this in TypeORM.

packages/pipeline/test/data_sources/ohlcv_external/crypto_compare_test.ts
albrow: We don't have any infra/tooling in place to test functions which send HTTP requests. I think the plan is to eventually use something like [nock](https://www.npmjs.com/package/nock).

We should either clarify this TODO comment or just remove the `getAsync` test. When we add `nock` or some other tooling we can go through and add similar tests for the other files in __data_sources/__.

packages/pipeline/test/entities/ohlcv_external_test.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { OHLCVExternal } from '../../src/entities';
+import { createDbConnectionOnceAsync } from '../db_setup';
+import { chaiSetup } from '../utils/chai_setup';
+
+import { testSaveAndFindEntityAsync } from './util';
+
+chaiSetup.configure();
+
+const ohlcvExternal = {
```
albrow: Small tip that I just recently discovered, if you add the `OHLCVExternal` type to `ohlcvExternal`, the compiler will alert you if you are missing any fields or have any types that are incorrect 😄 . To take advantage of this you would just write:


```ts
const ohlcvExternal: OHLCVExternal = {
    // ...
}
```

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,88 @@
+// tslint:disable:no-duplicate-imports
+import { AxiosResponse } from 'axios';
+import axios from 'axios';
+import promiseLimit = require('p-limit');
+
+import { merge, unfold } from 'ramda';
```
albrow: There are some functions in Ramda that conflict with built-in functions or keywords. So sometimes we will have to import it with `import * as R from 'ramda'`. Could we also do that here to keep the name-spacing consistent? I also like being able to more easily see where the functions came from.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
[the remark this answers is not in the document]
xianny: I added some type annotations and comments :)

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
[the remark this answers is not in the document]
xianny: I'm implementing with `fetchAsync` but had to do some extra stuff with parsing query params and the response. LMK if 👍 to keep this

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: Didn't know that! Thanks for the heads up. I'll do that.

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
xianny: @wakkadojo I looked around and deleted `getHourInUnixTime`

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: We should maybe turn this into a param at some point but for now I will just add USDT.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
xianny: I was thinking of parameterising the url, but won't this time.

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
xianny: 1. The size of records is limited by the API call so I think we are safe.

2. Nope, just didn't know about `save`! Changing it to avoid unhandled errors, but also want to note that I'm not sure if primary key constraints apply in this table. We don't have any unique identifiers, unless we want to generate a composite key of all the primary columns (defined with typeorm annotations). And if we do have a duplicate of all those fields, we would probably want to ignore the dupes.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: sounds like we decided in slack to keep lodash out of this package for now so leaving this

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: refactored!

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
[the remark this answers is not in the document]
fabioberger: Yeah, I'd be fine with that, although I feel like the argument is similar. The reason we have `logUtils.log` is because most dev's use `console.log` for debugging and might accidentally leave it in the code. `logUtils.log` is not "front-of-mind" and therefore less likely to be left accidentally (although still a possibility). I do think that adding the `// tslint:disable-next-line` adds even more friction so it could be even better at achieving our stated aim. Let's bring this up at the dev meeting next week, and if the majority agrees, let's replace `logUtils.log` with `console.log` & tslint disable-next-line.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
fabioberger: @albrow let me backtrack a bit.

We already have the `no-floating-promises` which should watch missing awaits within our codebase. We didn't always have this TSLint rule which is why we had the `async` prefix convention across our codebase. We do however want to keep the `async` prefix convention for our publicly exposed interface, since the users of our libraries might not use a linter and this way it's very easy for them to immediately tell which methods return promises and which don't. 

In this specific case then, we can leave it as it is.

packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
xianny: I refactored to a different implementation, but still a good question. I think the new logic is good. The script scrapes CC and errs on the side of overlapping data, and then we filter on save. 

Results:
screen shot 2018-12-04 at 11 17 28 am

810 records means 809 divisors. 
2912400000 ms/ 809 = 3,600,000 ms = 3600 s = 1hr

So I'm reasoning that we aren't getting any gaps or overlaps.

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -9,8 +9,10 @@ import { handleError } from '../utils';
 import { fetchOHLCVTradingPairsAsync, TradingPair } from '../utils/get_ohlcv_trading_pairs';
 
 const SOURCE_NAME = 'CryptoCompare';
+// tslint:disable:custom-no-magic-numbers
```
albrow: Can you change this to `// tslint:disable-next-line:custom-no-magic-numbers`? That disables it for the next line only instead of the entire file.

packages/pipeline/src/parsers/ohlcv_external/crypto_compare.ts
```diff
@@ -0,0 +1,38 @@
...
+
+        ohlcvRecord.open = rec.open;
+        ohlcvRecord.close = rec.close;
+        ohlcvRecord.low = rec.low;
+        ohlcvRecord.high = rec.high;
+        ohlcvRecord.volumeFrom = rec.volumefrom;
+        ohlcvRecord.volumeTo = rec.volumeto;
+
+        ohlcvRecord.source = metadata.source;
+        ohlcvRecord.observedTimestamp = metadata.observedTimestamp;
```
wakkadojo: Can you pull observed timestamp out of the map calls? We want all the records queried to have the same exact timestamp

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -0,0 +1,99 @@
...
+    const sortAscTimestamp = (a: TradingPair, b: TradingPair): number => {
+        if (a.latest < b.latest) {
+            return -1;
+        } else if (a.latest > b.latest) {
+            return 1;
+        } else {
+            return 0;
+        }
+    };
+    pairs.sort(sortAscTimestamp);
```
albrow: Optionally, you could replace this with: `const sortedPairs = R.sortBy(R.prop('latest'), pairs);`. Not a priority though.

PR #1351, opened 2018-11-28
[instant] Fix instant bouncing around / height changes, don't allow labels to wrap

fragosti:
## Description

- Render a hidden `SlidingError` when there is no error to maintain the correct layout. 
- Make the height of `PaymentMethod` fixed such that it doesn't jump around (ie. going from loading to loaded). 
- Don't allow labels to wrap (it will happen in certain edge cases).

## Checklist:


packages/instant/src/components/ui/container.tsx
```diff
@@ -70,6 +71,7 @@ export const Container =
         ${props => props.width && stylesForMedia<string>('width', props.width)}
         ${props => props.height && stylesForMedia<string>('height', props.height)}
         ${props => props.borderRadius && stylesForMedia<string>('border-radius', props.borderRadius)}
+        ${props => (props.isHidden ? 'visibility: hidden;' : '')}
```
steveklebanoff: TIL about `visibility: hidden;` VS `display: none;` 😄

packages/instant/src/containers/latest_error.tsx
```diff
@@ -23,7 +24,12 @@ export interface LatestErrorComponentProps {
 
 export const LatestErrorComponent: React.StatelessComponent<LatestErrorComponentProps> = props => {
     if (!props.latestErrorMessage) {
-        return <div />;
+        // Render a hidden SlidingError such that instant does not move when a real error is rendered.
+        return (
+            <Container isHidden={true}>
```
steveklebanoff: Any reason for this approach as opposed to just setting a constant height on the container like you did for `payment_method.tsx`?

fragosti: Seems more flexible... you can now freely change the SlidingError component and not break this logic. 

The reason I used a fixed height `payment_method` is because the height of the children is slightly different, so I wanted to fix the height of the parent which was ultimately the thing making instant bounce around. 

I guess you could argue for the same thing here, but this seemed simpler (I don't like fixed heights)

PR #1354, opened 2018-11-29
WIP token_orderbook_snapshots

zichongkao:
## Description

1. Sets up token_orderbook_snapshot migration.
2. Scrapes orderbook data from Ddex and inserts it into the table.

## Testing instructions

Added tests for parser. These work fine.
Ran script and saved data into DB:

image

## Types of changes

Most important functions have been documented.

## Checklist:

### DIFF
--- a/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
+++ b/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
@@ -3,17 +3,18 @@ import { MigrationInterface, QueryRunner, Table } from 'typeorm';
 const tokenOrderbookSnapshots = new Table({
     name: 'raw.token_orderbook_snapshots',
     columns: [
-        { name: 'source', type: 'varchar', isPrimary: true },
-        { name: 'retrieval_timestamp', type: 'bigint', isPrimary: true },
-        { name: 'order_type', type: 'order_t', isPrimary: true },
+        { name: 'source', type: 'varchar' },
+        { name: 'observed_timestamp', type: 'bigint', isPrimary: true },
+        { name: 'order_type', type: 'order_t' },
+        { name: 'price', type: 'numeric', isPrimary: true },
 
         { name: 'base_asset_symbol', type: 'varchar', isPrimary: true },
         { name: 'base_asset_address', type: 'char(42)' },
-        { name: 'base_volume', type: 'numeric', isPrimary: true },
+        { name: 'base_volume', type: 'numeric' },
 
         { name: 'quote_asset_symbol', type: 'varchar', isPrimary: true },
         { name: 'quote_asset_address', type: 'char(42)' },
-        { name: 'quote_volume', type: 'numeric', isPrimary: true },
+        { name: 'quote_volume', type: 'numeric' },
     ],
 });
 
--- a/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
+++ b/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
@@ -3,8 +3,8 @@ import { MigrationInterface, QueryRunner, Table } from 'typeorm';
 const tokenOrderbookSnapshots = new Table({
     name: 'raw.token_orderbook_snapshots',
     columns: [
-        { name: 'source', type: 'varchar' },
         { name: 'observed_timestamp', type: 'bigint', isPrimary: true },
+        { name: 'source', type: 'varchar', isPrimary: true },
         { name: 'order_type', type: 'order_t' },
         { name: 'price', type: 'numeric', isPrimary: true },
 
--- a/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
+++ b/packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
@@ -1,0 +1,29 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+const tokenOrderbookSnapshots = new Table({
+    name: 'raw.token_orderbook_snapshots',
+    columns: [
+        { name: 'source', type: 'varchar', isPrimary: true },
+        { name: 'retrieval_timestamp', type: 'bigint', isPrimary: true },
+        { name: 'order_type', type: 'order_t', isPrimary: true },
+
+        { name: 'base_asset_symbol', type: 'varchar', isPrimary: true },
+        { name: 'base_asset_address', type: 'char(42)' },
+        { name: 'base_volume', type: 'numeric', isPrimary: true },
+
+        { name: 'quote_asset_symbol', type: 'varchar', isPrimary: true },
+        { name: 'quote_asset_address', type: 'char(42)' },
+        { name: 'quote_volume', type: 'numeric', isPrimary: true },
+    ],
+});
+
+export class TokenOrderbookSnapshots1543434472116 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.query(`CREATE TYPE order_t AS enum('bid', 'ask');`);
+        await queryRunner.createTable(tokenOrderbookSnapshots);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<any> {
+        await queryRunner.dropTable(tokenOrderbookSnapshots.name);
+    }
+}
--- a/packages/pipeline/src/data_sources/ddex/index.ts
+++ b/packages/pipeline/src/data_sources/ddex/index.ts
@@ -1,5 +1,4 @@
-// tslint:disable:no-console
-import axios from 'axios';
+import { fetchAsync, logUtils } from '@0x/utils';
 
 const DDEX_BASE_URL = 'https://api.ddex.io/v2';
 const ACTIVE_MARKETS_URL = DDEX_BASE_URL + '/markets';
@@ -56,10 +55,11 @@ export class DdexSource {
      * Call Ddex API to find out which markets they are maintaining orderbooks for.
      */
     public async getActiveMarketsAsync(): Promise<DdexMarket[]> {
-        console.log('Getting all active DDEX markets');
-        const resp = await axios.get<DdexActiveMarketsResponse>(ACTIVE_MARKETS_URL);
-        const markets = resp.data.data.markets;
-        console.log(`Got ${markets.length} markets.`);
+        logUtils.log('Getting all active DDEX markets');
+        const resp = await fetchAsync(ACTIVE_MARKETS_URL);
+        const respJson: DdexActiveMarketsResponse = await resp.json();
+        const markets = respJson.data.markets;
+        logUtils.log(`Got ${markets.length} markets.`);
         return markets;
     }
 
@@ -68,9 +68,10 @@ export class DdexSource {
      * @param marketId String identifying the market we want data for.
      */
     public async getMarketOrderbookAsync(marketId: string): Promise<DdexOrderbook> {
-        console.log(`${marketId}: Retrieving orderbook.`);
+        logUtils.log(`${marketId}: Retrieving orderbook.`);
         const marketOrderbookUrl = ACTIVE_MARKETS_URL + '/' + marketId + ORDERBOOK_ENDPOINT;
-        const resp = await axios.get<DdexOrderbookResponse>(marketOrderbookUrl);
-        return resp.data.data.orderBook;
+        const resp = await fetchAsync(marketOrderbookUrl);
+        const respJson: DdexOrderbookResponse = await resp.json();
+        return respJson.data.orderBook;
     }
 }
--- a/packages/pipeline/src/data_sources/ddex/index.ts
+++ b/packages/pipeline/src/data_sources/ddex/index.ts
@@ -1,8 +1,9 @@
 import { fetchAsync, logUtils } from '@0x/utils';
 
 const DDEX_BASE_URL = 'https://api.ddex.io/v2';
-const ACTIVE_MARKETS_URL = DDEX_BASE_URL + '/markets';
-const ORDERBOOK_ENDPOINT = '/orderbook?level=3';
+const ACTIVE_MARKETS_URL = `${DDEX_BASE_URL}/markets`;
+const NO_AGGREGATION_LEVEL = 3; // See https://docs.ddex.io/#get-orderbook
+const ORDERBOOK_ENDPOINT = `/orderbook?level=${NO_AGGREGATION_LEVEL}`;
 export const DDEX_SOURCE = 'ddex';
 
 export interface DdexActiveMarketsResponse {
@@ -65,11 +66,11 @@ export class DdexSource {
 
     /**
      * Retrieve orderbook from Ddex API for a given market.
-     * @param marketId String identifying the market we want data for.
+     * @param marketId String identifying the market we want data for. Eg. 'REP/AUG'
      */
     public async getMarketOrderbookAsync(marketId: string): Promise<DdexOrderbook> {
         logUtils.log(`${marketId}: Retrieving orderbook.`);
-        const marketOrderbookUrl = ACTIVE_MARKETS_URL + '/' + marketId + ORDERBOOK_ENDPOINT;
+        const marketOrderbookUrl = `${ACTIVE_MARKETS_URL}/${marketId}${ORDERBOOK_ENDPOINT}`;
         const resp = await fetchAsync(marketOrderbookUrl);
         const respJson: DdexOrderbookResponse = await resp.json();
         return respJson.data.orderBook;
--- a/packages/pipeline/src/data_sources/ddex/index.ts
+++ b/packages/pipeline/src/data_sources/ddex/index.ts
@@ -1,0 +1,76 @@
+// tslint:disable:no-console
+import axios from 'axios';
+
+const DDEX_BASE_URL = 'https://api.ddex.io/v2';
+const ACTIVE_MARKETS_URL = DDEX_BASE_URL + '/markets';
+const ORDERBOOK_ENDPOINT = '/orderbook?level=3';
+export const DDEX_SOURCE = 'ddex';
+
+export interface DdexActiveMarketsResponse {
+    status: number;
+    desc: string;
+    data: {
+        markets: DdexMarket[];
+    };
+}
+
+export interface DdexMarket {
+    id: string;
+    quoteToken: string;
+    quoteTokenDecimals: number;
+    quoteTokenAddress: string;
+    baseToken: string;
+    baseTokenDecimals: number;
+    baseTokenAddress: string;
+    minOrderSize: string;
+    maxOrderSize: string;
+    pricePrecision: number;
+    priceDecimals: number;
+    amountDecimals: number;
+}
+
+export interface DdexOrderbookResponse {
+    status: number;
+    desc: string;
+    data: {
+        orderBook: DdexOrderbook;
+    };
+}
+
+export interface DdexOrderbook {
+    marketId: string;
+    bids: DdexOrder[];
+    asks: DdexOrder[];
+}
+
+export interface DdexOrder {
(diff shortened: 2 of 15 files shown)


packages/pipeline/src/parsers/ddex_orders/index.ts
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Aggregates orders by price point for consistency with other exchanges.
+ * Querying the Ddex API at level 3 setting returns a breakdown of
+ * individual orders at each price point. Other exchanges only give total amount
+ * at each price point. Returns an array of <price, amount> tuples.
+ * @param ddexOrders A list of Ddex orders awaiting aggregation.
+ */
+export function aggregateOrders(ddexOrders: DdexOrder[]): Array<[string, number]> {
+    const aggregatedPricePoints =  ddexOrders.reduce((orders: Map<string, number>, order: DdexOrder): Map<string, number> => {
```
albrow: Maybe we could use Ramda to simplify this code a bit? 

Here's a [slightly simplified example](https://goo.gl/iZUfdn) which makes use of [`R.reduceBy`](https://ramdajs.com/docs/#reduceBy). If that looks too foreign we could split it into separate steps.

packages/pipeline/src/parsers/ddex_orders/index.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param ddexOrderbook A raw orderbook that we pull from the Ddex API.
+ * @param ddexMarket An object containing market data also directly from the API.
+ * @param retrievalTimestamp Time at which the orders for the market were pulled.
+ * @param source The exchange where these orders are placed. In this case 'ddex'.
+ */
+export function parseDdexOrders(ddexOrderbook: DdexOrderbook, ddexMarket: DdexMarket, retrievalTimestamp: BigNumber, source: string): TokenOrder[] {
+    const aggregatedBids = aggregateOrders(ddexOrderbook.bids);
+    const aggregatedAsks = aggregateOrders(ddexOrderbook.asks);
+    // Might want to shift 'bid', 'ask' enums to 0x/types and retrieve from there.
+    const parsedBids = R.map(parseDdexOrder(ddexMarket, retrievalTimestamp, 'bid', source), aggregatedBids) ;
```
albrow: nit: There's a space before the semicolon here.

This sort of formatting issue is best suited for robots, not humans. Were you able to set up [Prettier and TSLint](https://github.com/0xProject/0x-monorepo/blob/development/CONTRIBUTING.md#development-tooling)? They can automatically format your code and save you a ton of time.

packages/pipeline/src/parsers/ddex_orders/index.ts
```diff
@@ -0,0 +1,73 @@
...
+    const aggregatedPricePoints =  ddexOrders.reduce((orders: Map<string, number>, order: DdexOrder): Map<string, number> => {
+        const price = order.price;
+        const amount = orders.get(price) || 0;
+        orders.set(price, amount + Number(order.amount));
+        return orders;
+    }, new Map());
+    return [ ...aggregatedPricePoints.entries()];
+}
+
+const parseDdexOrder = R.curry((ddexMarket: DdexMarket, retrievalTimestamp: BigNumber, orderType: string, source: string, ddexOrder: [string, number]): TokenOrder => {
```
albrow: It looks like all this function does is call `_parseDdexOrder`. Can we just remove it and call `_parseDdexOrder` directly?

packages/pipeline/src/scripts/pull_ddex_orderbook_snapshots.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as R from 'ramda';
+import { Connection, ConnectionOptions, createConnection, EntityManager } from 'typeorm';
+
+import { DdexActiveMarketsResponse, DdexMarket, DdexOrder, DdexOrderbookResponse } from '../data_sources/ddex';
+import { TokenOrderbookSnapshot as TokenOrder } from '../entities';
+import * as ormConfig from '../ormconfig';
+import { parseDdexOrders } from '../parsers/ddex_orders';
+import { handleError } from '../utils';
+
+const DDEX_BASE_URL = 'https://api.ddex.io/v2';
```
albrow: I was talking to @xianny about this and I think it is more appropriate to move URLs into the *data_sources* directory. In most cases, whenever we send an HTTP request, the response format is tied to the URL. This means that you can't really change the URL without making changes to the *data_sources* directory. They are tightly coupled, so we should keep them in the same place to make that coupling more obvious. What do you think?

Planning to make the same change to our other scripts eventually.

packages/pipeline/src/scripts/pull_ddex_orderbook_snapshots.ts
```diff
@@ -0,0 +1,62 @@
...
+    const marketOrderbookUrl = ACTIVE_MARKETS_URL + '/' + market.id + ORDERBOOK_ENDPOINT;
+    const resp = await axios.get<DdexOrderbookResponse>(marketOrderbookUrl);
+    const retrievedTimestamp = new BigNumber(Date.now());
+
+    console.log(`${market.id}: Parsing orders.`);
+    const orders = parseDdexOrders(resp.data.data.orderBook, market, retrievedTimestamp, DDEX_SOURCE);
+
+    if (orders.length > 0) {
+        console.log(`${market.id}: Saving ${orders.length} orders.`);
+        await connection.transaction(async (manager: EntityManager): Promise<void> => {
```
albrow: If you just want to save the orders, you don't need to create a transaction. You can save some code by doing it this way:

```
ordersRepository.save(orders, { chunk: Math.ceil(blocks.length / BATCH_SAVE_SIZE) });
```

Where `BATCH_SAVE_SIZE` is the number of orders to save per query. Batching saves this way also helps us avoid issues with transactions that are too big for Postgres.

See https://github.com/typeorm/typeorm/blob/master/src/repository/SaveOptions.ts#L30

packages/pipeline/src/entities/token_order.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+import { OrderType } from '../types';
+import { bigNumberTransformer } from '../utils';
+
+@Entity({ name: 'token_orderbook_snapshots', schema: 'raw' })
+export class TokenOrderbookSnapshot {
+    @PrimaryColumn({ name: 'source' })
+    public source!: string;
+    @PrimaryColumn({ name: 'retrieval_timestamp', type: 'numeric', transformer: bigNumberTransformer })
```
albrow: `retrieval_timestamp` doesn't need to have type `BigNumber`/`numeric`. You can use `number` for the JavaScript type and `bigint` for the Postgres type like we do for all other timestamps. The `number` type in JavaScript gives us guaranteed precision for integers up to a maximum up to 2^53-1. This is more than enough to cover Unix timestamps in milliseconds for the next few centuries 😄

packages/pipeline/src/entities/token_order.ts
```diff
@@ -0,0 +1,27 @@
...
+    public retrievalTimestamp!: BigNumber;
+    @PrimaryColumn({ name: 'order_type' })
+    public orderType!: OrderType;
+    @PrimaryColumn({ name: 'base_asset_symbol' })
+    public baseAssetSymbol!: string;
+    @Column({ name: 'base_asset_address' })
+    public baseAssetAddress!: string;
+    @PrimaryColumn({ name: 'base_volume' })
+    public baseVolume!: number;
```
albrow: On the other hand, for volumes we should be using the `BigNumber` and `numeric` types. We don't have any hard guarantees about these numbers and in theory they could be very large or have a high number of significant digits. We want to make absolutely sure that we don't run into any precision-related errors here.

packages/pipeline/src/entities/token_order.ts
zichongkao: But we're going to need this to run for thousands of millennia!!

packages/pipeline/src/parsers/ddex_orders/index.ts
zichongkao: Oh darn. I have TSlint setup, but it didn't pick out this one. Strange eh?

packages/pipeline/src/data_sources/ddex/index.ts
```diff
@@ -0,0 +1,76 @@
...
+        const markets = resp.data.data.markets;
+        console.log(`Got ${markets.length} markets.`);
+        return markets;
+    }
+
+    /**
+     * Retrieve orderbook from Ddex API for a given market.
+     * @param market Object from Ddex API containing market data.
+     */
+    public async getMarketOrderbookAsync(market: DdexMarket): Promise<DdexOrderbook> {
```
albrow: It looks like we only need `market.id` here not the entire `DdexMarket` object?

packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
```diff
@@ -0,0 +1,33 @@
+import { MigrationInterface, QueryRunner, Table } from 'typeorm';
+
+const tokenOrderbookSnapshots = new Table({
+    name: 'raw.token_orderbook_snapshots',
+    columns: [
+        { name: 'source', type: 'varchar', isPrimary: true },
```
albrow: Maybe I'm missing something but are all of these columns supposed to be primary columns? At the very least I would think that `base_volume` and `quote_volume` should not be primary columns.

packages/pipeline/src/scripts/pull_ddex_orderbook_snapshots.ts
```diff
@@ -0,0 +1,43 @@
+// tslint:disable:no-console
+import { Connection, ConnectionOptions, createConnection } from 'typeorm';
+
+import { DDEX_SOURCE, DdexMarket, DdexSource } from '../data_sources/ddex';
+import { TokenOrderbookSnapshot as TokenOrder } from '../entities';
+import * as ormConfig from '../ormconfig';
+import { parseDdexOrders } from '../parsers/ddex_orders';
+import { handleError } from '../utils';
+
+// Number of trades to save at once.
```
albrow: nit: This comment should read "Number of *orders* to save at once".

packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
zichongkao: Hm. This is my same thought. I was under the impression that primary key(s) have to uniquely identify each row. In this case, I think I need all of these columns to uniquely identify a row! The volumes are necessary because for each market, there is one row per price point and the price point is jointly specified by the volumes.

packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
```diff
@@ -0,0 +1,29 @@
...
+        { name: 'retrieval_timestamp', type: 'bigint', isPrimary: true },
```
wakkadojo: Can we use the same convention as in Alex's sra_orderbook table (i.e. call the retrieval_timestamp the observation_timestamp?)

packages/pipeline/migrations/1543434472116-TokenOrderbookSnapshots.ts
wakkadojo: Uniqueness could be in time / base symbol / quote symbol / price. I see you have price data below, can you keep it?

I guess theres no way to avoid having a floating point index if we want ordering in the physical storage, but can we somehow ensure time is the first index when sorting for performance reasons?

albrow: Agree with @wakkadojo. For this level of abstraction, I think it makes sense to store price and use it instead of the volumes as the primary key.

This will also line up better with the types of queries we will run, which will often involve sorting by price.

packages/pipeline/src/parsers/paradex_orders/index.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param retrievalTimestamp Time at which the orders for the market were pulled.
+ * @param source The exchange where these orders are placed. In this case 'paradex'.
+ */
+export function parseParadexOrders(
+    paradexOrderbookResponse: ParadexOrderbookResponse,
+    paradexMarket: ParadexMarket,
+    retrievalTimestamp: number,
+    source: string,
+): TokenOrder[] {
+    // Might want to shift 'bid', 'ask' enums to 0x/types and retrieve from there.
```
albrow: I think this comment can be removed.

packages/pipeline/src/parsers/paradex_orders/index.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param paradexOrder A ParadexOrder object; basically price, amount tuple.
+ */
+export function parseParadexOrder(
+    paradexMarket: ParadexMarket,
+    retrievalTimestamp: number,
+    orderType: string,
+    source: string,
+    paradexOrder: ParadexOrder,
+): TokenOrder {
+    BigNumber.config({ ERRORS: false });
```
albrow: I believe this changes the `BigNumber` configuration globally, which seems really dangerous. Why is this necessary? If it is required is there a way for us to make the configuration change local instead of global?

packages/pipeline/src/parsers/paradex_orders/index.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const tokenOrder = new TokenOrder();
+
+    tokenOrder.source = source;
+    tokenOrder.retrievalTimestamp = retrievalTimestamp;
+    tokenOrder.orderType = orderType as OrderType;
```
albrow: Same comment here re: `orderType as OrderType`.

packages/pipeline/src/scripts/pull_paradex_orderbook_snapshots.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Extend the default ParadexMarket objects with token addresses.
+ * @param markets An array of ParadexMarket objects.
+ * @param tokenInfoResponse An array of ParadexTokenInfo containing the addresses.
+ */
+function addTokenAddresses(
+    markets: ParadexActiveMarketsResponse,
+    tokenInfoResponse: ParadexTokenInfoResponse,
+): ParadexMarket[] {
+    const extractSymbolAddress = (acc: Map<string, string>, tokenInfo: ParadexTokenInfo): Map<string, string> => {
+        acc.set(tokenInfo.symbol, tokenInfo.address);
```
albrow: `acc.set` mutates the map. If we are going to use mutation here, we don't need to use `reduce`. We can just do something like:

```ts
const symbolToAddress = new Map<string, string>();
tokenInfoResponse.forEach(tokenInfo => symbolToAddress.set(tokenInfo.symbol, tokenInfo.address));
```

packages/pipeline/src/scripts/pull_paradex_orderbook_snapshots.ts
```diff
@@ -0,0 +1,93 @@
...
+        return acc;
+    };
+
+    const symbolAddressMapping = R.reduce(extractSymbolAddress, new Map(), tokenInfoResponse);
+
+    markets.forEach((market: ParadexMarket) => {
+        if (symbolAddressMapping.has(market.baseToken)) {
+            market.baseTokenAddress = symbolAddressMapping.get(market.baseToken);
+        } else {
+            console.log(`${market.baseToken}: No address found.`);
```
albrow: How serious/unexpected is this error condition? Should we replace `console.log` with either `console.error` or `throw new Error('your message here')`?

zichongkao: Hm. I feel like in this case it's an inconsistency of the external datasource, not an unexpected result of our system. But I don't feel strongly. I'd be happy to change it if you prefer.

packages/pipeline/src/parsers/paradex_orders/index.ts
albrow: In that case just change:

```ts
 new BigNumber(Number(paradexOrder.price) * Number(paradexOrder.amount))
```

To this:

```ts
 new BigNumber(paradexOrder.price).times(new BigNumber(paradexOrder.amount));
```

In general, we never want to convert a `string` into a `number` because the `number` type might not be able to represent it due to precision constraints. `string` types should always be converted to `BigNumber`.

packages/pipeline/src/scripts/pull_paradex_orderbook_snapshots.ts
albrow: In that case it should either be console.error or nothing. If it happens frequently and we know how to handle it, just don't log anything. If we don't have a good way of handling it, console.error is more appropriate.

packages/pipeline/src/entities/token_order.ts
```diff
@@ -6,22 +6,24 @@ import { bigNumberTransformer, numberToBigIntTransformer } from '../utils';
 
 @Entity({ name: 'token_orderbook_snapshots', schema: 'raw' })
 export class TokenOrderbookSnapshot {
-    @PrimaryColumn({ name: 'source' })
+    @Column({ name: 'source' })
```
albrow: I think we still need `source` to be a primary column because we could theoretically have two different snapshots from different sources at the same time.

packages/pipeline/test/entities/token_order_test.ts
```diff
@@ -0,0 +1,32 @@
...
+    price: new BigNumber(10.1),
+    baseAssetSymbol: 'ETH',
+    baseAssetAddress: '0x818e6fecd516ecc3849daf6845e3ec868087b755',
+    baseVolume: new BigNumber(143),
+    quoteAssetSymbol: 'ABC',
+    quoteAssetAddress: '0x00923b9a074762b93650716333b3e1473a15048e',
+    quoteVolume: new BigNumber(12.3234234),
+};
+
+// tslint:disable:custom-no-magic-numbers
```
fabioberger: Where is the magic number? Also, if you disable a tslint rule, I'd prefer you only do it for the next line with `// tslint-disable-next-line` or also re-enable it after a block of code.

packages/pipeline/src/scripts/pull_ddex_orderbook_snapshots.ts
```diff
@@ -10,13 +11,24 @@ import { handleError } from '../utils';
...
+const MILLISEC_MARKET_ORDERBOOK_REQUEST_DELAY = 5000;
+
 let connection: Connection;
 
 (async () => {
     connection = await createConnection(ormConfig as ConnectionOptions);
     const ddexSource = new DdexSource();
     const markets = await ddexSource.getActiveMarketsAsync();
-    await Promise.all(markets.map(async (market: DdexMarket) => getAndSaveMarketOrderbook(ddexSource, market)));
+    for (const marketsChunk of R.splitEvery(MARKET_ORDERBOOK_REQUEST_BATCH_SIZE, markets)) {
```
albrow: 👍 nice implementation of chunking requests.

packages/pipeline/test/parsers/paradex_orders/index_test.ts
```diff
@@ -52,3 +52,4 @@ describe('paradex_orders', () => {
         });
     });
 });
+// tslint:enable:custom-no-magic-numbers
```
albrow: `tslint:enable` and `tslint:disable` operate on a per-file basis. So you don't need this comment. https://palantir.github.io/tslint/usage/rule-flags/

packages/tslint-config/tslint.json
```diff
@@ -92,6 +92,7 @@
         "prefer-function-over-method": true,
         "prefer-object-spread": true,
         "prefer-readonly": true,
+        "prefer-template": true,
```
albrow: Adding this rule applies it to the entire monorepo. I think it is a good idea but it is probably best to create a separate PR where we also go through and fix all the code that doesn't adhere to the rule.

PR #1355, opened 2018-11-29
Add script for parsing competing dex trades from Bloxy

albrow:
## Description

This PR adds a new script for pulling trades from some competing DEXs like Kyber and Airswap using [the Bloxy API](https://bloxy.info/api_methods#dex).

## Checklist:


packages/pipeline/migrations/1543446690436-CreateDexTrades.ts
```diff
@@ -0,0 +1,41 @@
...
+        { name: 'tx_timestamp', type: 'bigint' },
+        { name: 'tx_date', type: 'varchar' },
+        { name: 'tx_sender', type: 'varchar(42)' },
+        { name: 'smart_contract_id', type: 'bigint' },
+        { name: 'smart_contract_address', type: 'varchar(42)' },
+        { name: 'contract_type', type: 'varchar' },
+        { name: 'maker', type: 'varchar(42)' },
+        { name: 'taker', type: 'varchar(42)' },
+        { name: 'amount_buy', type: 'numeric' },
+        { name: 'maker_fee', type: 'numeric' },
```
datamongers: Can you change names so it's more clear, like maker_fee_bps or maker_fee_pct, since it's not clear to me if maker_fee is in basis points, percent, or if it's a gross amount in some token. I also don't know what buy and sell are. Is the buyer always the market taker (if yes then should this be taker_amount, or is this for a standard market and buy is the person receiving the base of the price quote, like 1 BTC/USD traded at $6k, buyer is the maker if traded on the bid and buyer is the taker if traded on the ask)? 

Or maybe we can just have docs somewhere, inline even w/ table definitions by entities or something, making these clarifications?

packages/pipeline/migrations/1543446690436-CreateDexTrades.ts
```diff
@@ -0,0 +1,41 @@
...
+const dexTrades = new Table({
+    name: 'raw.dex_trades',
+    columns: [
+        { name: 'source_url', type: 'varchar', isPrimary: true },
+        { name: 'tx_hash', type: 'varchar', isPrimary: true },
+
+        { name: 'tx_timestamp', type: 'bigint' },
+        { name: 'tx_date', type: 'varchar' },
+        { name: 'tx_sender', type: 'varchar(42)' },
+        { name: 'smart_contract_id', type: 'bigint' },
```
datamongers: Why do we have a smart_contract_id and a smart_contract_address? Isn't the address a unique ID?

albrow: Take a look at https://bloxy.info/api_methods#dex. Since I'm not sure exactly what we will need, I'm just storing all the data they give us for now.

It looks like there might be a one-to-one relationship between `smart_contract_id` and `smart_contract_address` but I can't be certain. I can imagine it being the case that the address could change e.g. if a new version or hotfix for the contract is released.

Are you okay with keeping it for now until we get some clarification from the Bloxy team?

packages/pipeline/migrations/1543446690436-CreateDexTrades.ts
albrow: Good point. Looking at the data, I am fairly confident that `maker_fee` and `taker_fee` are gross amounts and not basis points or percentages. I renamed them to `maker_fee_amount` and `taker_fee_amount` in order to make this more clear.

Unfortunately the data itself doesn't help me understand whether `buy_amount` and `buy_symbol` correspond to the `maker` or the `taker` as that somewhat arbitrary. I could make a guess about which would make more sense to me, but I think I would rather ask the Bloxy team to be certain.

PR #1357, opened 2018-11-29
[instant] Environment-specific configuration

steveklebanoff:
## Description

⚠️ For review only.  If approved, merge into `development` once https://github.com/0xProject/0x-monorepo/pull/1348 is merged ⚠️ 

⚠️ See https://github.com/0xProject/0x-monorepo/pull/1359 for an alternative approach which may be better ⚠️ 

When we're deploying to dogfood, staging, or the production cdn, we:
   - Want our codebase to know which 'environment' we're in, so we can report it to heap and rollbar
   - Want to ensure that certain ENV vars are set (i.e. heap) so we don't deploy a version without analytics enabled or rollbar enabled

In this PR, I ensure that `--env.<env name>`  is set up for every type of build we can do, and our webpack config knows about the different requirements for each environment.  

I first implemented this using a file per env, like `webpack.config.dogfood.js`, etc - but it felt like a lot of files, and there was duplicated logic in them.  For now, felt cleanest just to centralize this logic into the existing `webpack.config.js`, and abstract the differences into being generated via a config object (i.e. `getConfigForEnv`)

## Checklist:


packages/instant/package.json
```diff
@@ -8,10 +8,11 @@
     "main": "umd/instant.js",
     "private": true,
     "scripts": {
-        "build": "webpack --mode production",
-        "build:ci": "yarn build",
+        "build": "yarn build:prod --env.production_standalone",
```
steveklebanoff: This is a little weird, and I don't love it.  I introduced two production environments:

- **production_standalone** is for CI and the monorepo to build.  It generates with production config, but doesn't require the heap ENV var.  This allows us to not force people to set these env vars locally or in CI in order to build the project
- **production_cdn**: this is for the version that gets deployed to the production cdn, and we force users to have the heap ENV var set, to ensure we never deploy a version w/o analytics enabled

Open to other ideas on how to approach this.  The goal is to not force ENV vars for CI and monorepo builders, but to force ENV vars when deploying to the CDN.  

An alternative approach would be for these both ( and maybe staging and dogfood too?) to be "production" environment, but for "--cdn" to be a separate argument sent to webpack.

steveklebanoff: Alternative approach implemented here: https://github.com/0xProject/0x-monorepo/pull/1359

BMillman19: what about making the default `build` command build development, and then have `production_ci` and `production_cdn`, alternatively if the only difference is whether heap is enabled or not perhaps the naming can be more specific `production_with_heap` / `production_no_heap`

steveklebanoff: in the future the production cdn build will have rollbar too though, and `production_with_heap_and_rollbar` would be a bit verbose

steveklebanoff:
Closing in favor of https://github.com/0xProject/0x-monorepo/pull/1359

PR #1358, opened 2018-11-29
[WIP] Compliant Forwarder

hysz:
## Checklist:


packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,101 @@
...
+    IERC721Token internal COMPLIANCE_TOKEN;
+
+    constructor(address exchange, address complianceToken)
+        public
+    {
+        EXCHANGE = IExchange(exchange);
+        COMPLIANCE_TOKEN = IERC721Token(complianceToken);
+    }
+
+    function fillOrder(
```
hysz: This uses the `executeTransaction` interface. Another option is to use the `fillOrder` interface, then ABI-encode within this function. Did it this way to save gas. Shouldn't be a huge deal if most clients use the TS wrapper we write.

abandeali1: I think it would be valuable to pass in the actual `Order` and `takerAssetFillAmount`, validate them, and then ABI encode the params before passing them into executeTransaction. Even though this would be less gas efficient than what we're currently doing, this is much more extensible to other wrapper functions.

packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,104 @@
...
+            0x1e0     makerAssetData                                                                  
+            *         takerAssetData                                                                  
+            *         signature
+            ------------------------------
+            * Context-dependent offsets; unknown at compile time.
+        */
+        // Add 0x4 to a given offset to account for the fillOrder selector prepended to `signedFillOrderTransaction`.
+        // Add 0xc to the makerAddress since abi-encoded addresses are left padded with 12 bytes.
+        // Putting this together: makerAddress = 0x60 + 0x4 + 0xc = 0x70
+        address makerAddress = signedFillOrderTransaction.readAddress(0x70);
```
abandeali1: Shouldn't we check the location relative to the order offset? (not an issue if we use `Exchange.fillOrder` ABI)

packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,104 @@
...
+        require(
```
abandeali1: If we do use the `Exchange.fillOrder` ABI, we can move this logic into an `isValidOrder` function that is used within each wrapper function.

packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,104 @@
...
+*/
+
+pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
+
+import "../../protocol/Exchange/interfaces/IExchange.sol";
+import "../../tokens/ERC721Token/IERC721Token.sol";
+import "../../utils/LibBytes/LibBytes.sol";
+
+contract CompliantForwarder {
```
abandeali1: I personally like a more descriptive name for the contract, like `ERC721OwnerExchange` (not that I love this name either, but should probably be something a that's a bit more specific).

packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,104 @@
...
+    IExchange internal EXCHANGE;
+    IERC721Token internal COMPLIANCE_TOKEN;
+
+    constructor(address exchange, address complianceToken)
+        public
+    {
+        EXCHANGE = IExchange(exchange);
+        COMPLIANCE_TOKEN = IERC721Token(complianceToken);
+    }
+
```
abandeali1: I do think we need to add _some_ other wrapper functions to this. As is, a user wouldn't even be able to cancel their orders that are tied to this contract. We can probably create a generic mixin that could be used to automate this part of any extension contract (similar to `MixinWrapperFunctions` minus some bells and whistles).

packages/contracts/contracts/tokens/README.md
```diff
@@ -0,0 +1,2 @@
+Contracts from https://github.com/sendwyre/yes-compliance-token
```
abandeali1: This could probably just be added somewhere in the top level contracts README.

packages/contracts/package.json
```diff
@@ -34,7 +34,7 @@
...
+            "generated-artifacts/@(AssetProxyOwner|CompliantForwarder|DummyERC20Token|DummyERC721Receiver|DummyERC721Token|DummyMultipleReturnERC20Token|DummyNoReturnERC20Token|ERC20Token|ERC20Proxy|ERC721Token|ERC721Proxy|Forwarder|Exchange|ExchangeWrapper|IAssetData|IAssetProxy|InvalidERC721Receiver|MixinAuthorizable|MultiAssetProxy|MultiSigWallet|MultiSigWalletWithTimeLock|OrderValidator|ReentrantERC20Token|TestAssetProxyOwner|TestAssetProxyDispatcher|TestConstants|TestExchangeInternals|TestLibBytes|TestLibs|TestSignatureValidator|TestStaticCallReceiver|Validator|Wallet|Whitelist|WETH9|YesComplianceToken|ZRXToken).json"
```
abandeali1: Why not omit this and just use one of our existing ERC721 tokens? We only really need the interface for testing purposes.

packages/contracts/src/artifacts/index.ts
```diff
@@ -36,10 +37,12 @@ import * as Validator from '../../generated-artifacts/Validator.json';
 import * as Wallet from '../../generated-artifacts/Wallet.json';
 import * as WETH9 from '../../generated-artifacts/WETH9.json';
 import * as Whitelist from '../../generated-artifacts/Whitelist.json';
+import * as YesComplianceToken from '../../generated-artifacts/YesComplianceToken.json';
```
abandeali1: Similar to above, we can probably just use `DummyERC721Token` here.

packages/contracts/test/extensions/compliant_forwarder.ts
```diff
@@ -0,0 +1,318 @@
...
+import { TransactionFactory } from '../utils/transaction_factory';
+import { ContractName, ERC20BalancesByOwner, SignedTransaction } from '../utils/types';
+import { provider, txDefaults, web3Wrapper } from '../utils/web3_wrapper';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
+const DECIMALS_DEFAULT = 18;
+
+describe.only(ContractName.CompliantForwarder, () => {
```
abandeali1:
```suggestion
describe(ContractName.CompliantForwarder, () => {
```

packages/contracts/test/extensions/compliant_forwarder.ts
```diff
@@ -0,0 +1,318 @@
...
+            compliantSignedOrderWithoutExchangeAddressData,
+        );
+    });
+    beforeEach(async () => {
+        await blockchainLifecycle.startAsync();
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+    describe.only('fillOrder', () => {
```
abandeali1:
```suggestion
    describe('fillOrder', () => {
```

packages/contracts/test/extensions/compliant_forwarder.ts
```diff
@@ -0,0 +1,318 @@
...
+        /*
+        const compliantForwarderContract = new CompliantForwarderContract(
+            compliantForwarderInstance.abi,
+            compliantForwarderInstance.address,
+            provider,
+        );
+        forwarderWrapper = new ForwarderWrapper(compliantForwarderContract, provider);
+        */
+        // Initialize Yes Token
+        await yesTokenInstance._upgradeable_initialize.sendTransactionAsync({ from: owner });
```
abandeali1: There's a ton of extra logic here that doesn't actually affect our tests and could be initialized to null values (or just use `DummyERC721Token`).

packages/contracts/contracts/extensions/CompliantForwarder/CompliantForwarder.sol
```diff
@@ -0,0 +1,104 @@
...
+        // Validate `signedFillOrderTransaction`
+        bytes4 selector = signedFillOrderTransaction.readBytes4(0);
+        if (selector != EXCHANGE_FILL_ORDER_SELECTOR) {
+            revert("EXCHANGE_TRANSACTION_NOT_FILL_ORDER");
+        }
+        
+        // Taker must be compliant
+        require(
+            COMPLIANCE_TOKEN.balanceOf(signerAddress) > 0,
+            "TAKER_UNVERIFIED"
```
abandeali1: Maybe these revert reasons should also be more generic, like `0_ERC721_MAKER_BALANCE`.

BMillman19:
Is the "compliant forwarder" related to the "forwarder" we're using for instant? we shoulda avoid overloading names

hysz:
This has been superseded by #1383

PR #1362, opened 2018-11-30
[instant] Fix push to history issue on second full render of instant

fragosti:
This fixes

```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at d (https://localhost:3000/zrxinstant.js:37:329752)
    at window.onpopstate (https://localhost:3000/zrxinstant.js:37:331654)
    at window.onpopstate (https://localhost:3000/zrxinstant.js:37:331588)
```

That can occur if:
1. No onpopstate is defined. 
2. Instant is rendered "manually" (not through navigating history) twice in a row. 

Basically we were calling an old version of `removeInstant` which had a closure around an old DOM node that had been deleted because we were mistakenly setting our own `onpopstate` as the "intergrator's" `onpopstate` on the second render... which resulted in us calling the `removeInstant` from the first render...

## Checklist:

### DIFF
--- a/packages/instant/src/index.umd.ts
+++ b/packages/instant/src/index.umd.ts
@@ -89,11 +89,10 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
     // If the integrator defined a popstate handler, save it to __zeroExInstantIntegratorsPopStateHandler
     // unless we have already done so on a previous render.
     const anyWindow = window as any;
-    if (window.onpopstate && !anyWindow.__zeroExInstantIntegratorsPopStateHandler) {
-        anyWindow.__zeroExInstantIntegratorsPopStateHandler = window.onpopstate.bind(window);
-    } else {
-        anyWindow.__zeroExInstantIntegratorsPopStateHandler = util.boundNoop;
-    }
+    const popStateExistsAndNotSetPreviously = window.onpopstate && !anyWindow.__zeroExInstantIntegratorsPopStateHandler;
+    anyWindow.__zeroExInstantIntegratorsPopStateHandler = popStateExistsAndNotSetPreviously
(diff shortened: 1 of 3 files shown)


packages/instant/src/index.umd.ts
```diff
@@ -89,12 +89,12 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
     // If the integrator defined a popstate handler, save it to __zeroExInstantIntegratorsPopStateHandler
     // unless we have already done so on a previous render.
     const anyWindow = window as any;
-    if (window.onpopstate && !anyWindow.__zeroExInstantIntegratorsPopStateHandler) {
-        anyWindow.__zeroExInstantIntegratorsPopStateHandler = window.onpopstate.bind(window);
-    }
-    const integratorsOnPopStateHandler = anyWindow.__zeroExInstantIntegratorsPopStateHandler || util.boundNoop;
+    const popStateExistsAndNotSetPreviously = window.onpopstate && !anyWindow.__zeroExInstantIntegratorsPopStateHandler;
+    anyWindow.__zeroExInstantIntegratorsPopStateHandler = popStateExistsAndNotSetPreviously
```
steveklebanoff: this pattern of setting `__zeroExInstantIntegratorsPopStateHandler` to a noop feels kind of weird to me

would the following work instead?

```
anyWindow.__zeroExInstantIntegratorsPopStateHandler = window.onpopstate && !anyWindow.__zeroExInstantIntegratorsPopStateHandler;

const onPopStateHandler = (e: PopStateEvent) => {
        if (anyWindow.__zeroExInstantIntegratorsPopStateHandler) {
             anyWindow.__zeroExInstantIntegratorsPopStateHandler(e):
        }
       ....
}
```

PR #1363, opened 2018-11-30
Use new ABI encoder in assetDataUtils

abandeali1:
## Checklist:


packages/order-utils/package.json
```diff
@@ -13,12 +13,14 @@
...
-        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --bail --exit",
+        "run_mocha":
+            "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --bail --exit",
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
         "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
         "clean": "shx rm -rf lib generated_docs",
         "lint": "tslint --format stylish --project .",
-        "docs:json": "typedoc --excludePrivate --excludeExternals --target ES5 --tsconfig typedoc-tsconfig.json --json $JSON_FILE_PATH $PROJECT_FILES"
+        "docs:json":
+            "typedoc --excludePrivate --excludeExternals --target ES5 --tsconfig typedoc-tsconfig.json --json $JSON_FILE_PATH $PROJECT_FILES"
```
fabioberger: We ignore `package.json` in `.prettierignore`... Why is your setup running prettier on it?

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -1,10 +1,13 @@
...
-import ethUtil = require('ethereumjs-util');
+import { AssetData, AssetProxyId, ERC20AssetData, ERC721AssetData, MultiAssetData } from '@0x/types';
+import { AbiEncoder, BigNumber } from '@0x/utils';
+import { MethodAbi } from 'ethereum-types';
+import * as _ from 'lodash';
 
 import { constants } from './constants';
 
+const encodingRules: AbiEncoder.EncodingRules = { optimize: true };
+const decodingRules: AbiEncoder.DecodingRules = { structsAsObjects: true };
```
fabioberger: Can we rename these configs to follow our boolean naming conventions: `shouldOptimize` and `shouldConvertStructsToObjects`

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -13,34 +16,25 @@ export const assetDataUtils = {
...
-        const assetProxyId = ethUtil.bufferToHex(data.slice(0, constants.SELECTOR_LENGTH));
-        if (assetProxyId !== AssetProxyId.ERC20) {
-            throw new Error(
-                `Could not decode ERC20 Proxy Data. Expected Asset Proxy Id to be ERC20 (${
-                    AssetProxyId.ERC20
-                }), but got ${assetProxyId}`,
-            );
-        }
-        const [tokenAddress] = ethAbi.rawDecode(['address'], data.slice(constants.SELECTOR_LENGTH));
+        assetDataUtils.validateERC20AssetDataThrow(assetData);
```
fabioberger: Let's rename `validateERC20AssetDataThrow` to `assertIsERC20AssetData`. Assertions through if resolve false.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -13,34 +16,25 @@ export const assetDataUtils = {
...
+        const assetProxyId = assetDataUtils.decodeAssetProxyId(assetData);
+        const abiEncoder = new AbiEncoder.Method(constants.ERC20_METHOD_ABI as MethodAbi);
+        const decodedAssetData = abiEncoder.decode(assetData, decodingRules);
         return {
             assetProxyId,
-            tokenAddress: ethUtil.addHexPrefix(tokenAddress),
+            // TODO(abandeali1): fix decode return types
```
fabioberger: Is this still a TODO for this PR? If not, please elaborate on this TODO, It's not immediately obvious what it's referring to.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -51,42 +45,73 @@ export const assetDataUtils = {
...
     },
     /**
      * Decodes an ERC721 assetData hex string into it's corresponding ERC721 tokenAddress, tokenId & assetProxyId
      * @param assetData Hex encoded assetData string to decode
      * @return An object containing the decoded tokenAddress, tokenId & assetProxyId
      */
     decodeERC721AssetData(assetData: string): ERC721AssetData {
-        const data = ethUtil.toBuffer(assetData);
-        if (data.byteLength < constants.ERC721_ASSET_DATA_MINIMUM_BYTE_LENGTH) {
+        assetDataUtils.validateERC721AssetDataOrThrow(assetData);
```
fabioberger: Ditto comment above about `assert` naming

packages/order-watcher/src/order_watcher/dependent_order_hashes_tracker.ts
```diff
@@ -62,39 +62,20 @@ export class DependentOrderHashesTracker {
...
-        }
+        this._removeAssetDataFromDependentOrderHashes(signedOrder, signedOrder.makerAssetData);
         // If makerToken === ZRX then we already removed it and we don't need to remove it again.
+        const decodedMakerAssetData = assetDataUtils.decodeAssetDataOrThrow(signedOrder.makerAssetData);
         if ((decodedMakerAssetData as ERC20AssetData).tokenAddress !== this._zrxTokenAddress) {
             this._removeFromERC20DependentOrderhashes(signedOrder, this._zrxTokenAddress);
         }
         this._removeFromMakerDependentOrderhashes(signedOrder);
     }
+
```
fabioberger: Remove extra new line.

packages/order-utils/package.json
abandeali1: It's because I open up the packages individually w/ VSCode... will revert this change.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -13,34 +23,25 @@ export const assetDataUtils = {
...
-            throw new Error(
-                `Could not decode ERC20 Proxy Data. Expected Asset Proxy Id to be ERC20 (${
-                    AssetProxyId.ERC20
-                }), but got ${assetProxyId}`,
-            );
-        }
-        const [tokenAddress] = ethAbi.rawDecode(['address'], data.slice(constants.SELECTOR_LENGTH));
+        assetDataUtils.assertIsERC20AssetData(assetData);
+        const assetProxyId = assetDataUtils.decodeAssetProxyId(assetData);
+        const abiEncoder = new AbiEncoder.Method(constants.ERC20_METHOD_ABI as MethodAbi);
```
albrow: Should just declare the type as `MethodAbi` inside of `constants` in order to avoid all these casts?

packages/contract-wrappers/src/fetchers/asset_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -18,60 +19,101 @@ export class AssetBalanceAndProxyAllowanceFetcher implements AbstractBalanceAndP
...
-            });
-            return balance;
-        } else {
-            const decodedERC721AssetData = decodedAssetData as ERC721AssetData;
-            const tokenOwner = await this._erc721Token.getOwnerOfAsync(
-                decodedERC721AssetData.tokenAddress,
-                decodedERC721AssetData.tokenId,
-                {
+        let balance: BigNumber | undefined;
+        switch (decodedAssetData.assetProxyId) {
```
albrow: I see this `switch (decodedAssetData.assetProxyId)` pattern come up a lot. We could optionally use ["type guards"](https://www.typescriptlang.org/docs/handbook/advanced-types.html) to make the types a little easier to manage and avoid type assertions.

For example:

```ts
function isERC20AssetData(assetData: SingleAssetData | MultiAssetData): assetData is ERC20AssetData {
    return assetData.assetProxyId === AssetProxyId.ERC20;
}
```

If we write this type guard, the TypeScript compiler knows that when `isERC20AssetData` returns true the type should be `ERC20AssetData`.

packages/order-utils/src/asset_data_utils.ts
albrow: Should just declare the type as `MethodAbi` inside of `constants` in order to avoid all these type assertions?

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -51,42 +52,110 @@ export const assetDataUtils = {
...
+        const decodedNestedAssetData = _.map(
+            decodedAssetData.nestedAssetData as string[],
+            (nestedAssetDataElement, index) => {
+                const decodedNestedAssetDataElement = assetDataUtils.decodeAssetDataOrThrow(nestedAssetDataElement);
+                if (decodedNestedAssetDataElement.assetProxyId === AssetProxyId.MultiAsset) {
+                    const recursivelyDecodedAssetData = assetDataUtils.decodeMultiAssetDataRecursively(
+                        nestedAssetDataElement,
+                    );
+                    amounts.push(
+                        _.map(recursivelyDecodedAssetData.amounts as BigNumber[], amountElement =>
```
albrow: This `as BigNumber[]` type assertion doesn't seem necessary.

packages/order-utils/src/exchange_transfer_simulator.ts
```diff
@@ -74,24 +75,51 @@ export class ExchangeTransferSimulator {
...
+                await this._decreaseBalanceAsync(assetData, from, amountInBaseUnits);
+                await this._increaseBalanceAsync(assetData, to, amountInBaseUnits);
+                break;
+            case AssetProxyId.MultiAsset:
+                const decodedAssetData = assetDataUtils.decodeMultiAssetData(assetData);
+                for (const [index, nestedAssetDataElement] of decodedAssetData.nestedAssetData.entries()) {
+                    const amountsElement = decodedAssetData.amounts[index];
+                    const totalAmount = amountInBaseUnits.times(amountsElement);
+                    await this.transferFromAsync(
+                        nestedAssetDataElement as string,
```
albrow: This `as string` type assertion is unnecessary.

packages/order-utils/test/asset_data_utils_test.ts
```diff
@@ -10,41 +10,99 @@ import { chaiSetup } from './utils/chai_setup';
...
+    amounts: [new BigNumber(1), new BigNumber(1)],
+    nestedAssetData: [KNOWN_ERC20_ENCODING.assetData, KNOWN_ERC721_ENCODING.assetData],
+    assetData:
+        '0x94cfcdd7000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000024f47261b00000000000000000000000001dc4c1cefef38a777b15aa20260a54e584b16c48000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044025717920000000000000000000000001dc4c1cefef38a777b15aa20260a54e584b16c48000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000',
+};
 
+// tslint:disable:no-unnecessary-type-assertion
```
albrow: Disabling the linter rule for the entire file seems too broad to me. It looks like there are only 3 places in the file where we do type assertions. Could we individually disable the linter rule for those lines instead?

packages/order-watcher/test/order_watcher_test.ts
```diff
@@ -675,5 +675,213 @@ describe('OrderWatcher', () => {
...
+                    orderWatcher.subscribe(callback);
+                    isApproved = false;
+                    await contractWrappers.erc721Token.setProxyApprovalForAllAsync(
+                        makerErc721TokenAddress,
+                        makerAddress,
+                        isApproved,
+                    );
+                })().catch(done);
+            });
+            it('should emit orderStateInvalid when maker moves ERC721 backing watched order', (done: DoneCallback) => {
```
albrow: Mocha supports `async` functions directly, so I don't think you have to use callbacks here. Am I missing something?

packages/pipeline/src/parsers/events/exchange_events.ts
```diff
@@ -1,11 +1,13 @@
...
-import { AssetProxyId, ERC721AssetData } from '@0x/types';
+import { AssetProxyId, ERC721AssetData, SingleAssetData } from '@0x/types';
 import { LogWithDecodedArgs } from 'ethereum-types';
 import * as R from 'ramda';
 
 import { ExchangeCancelEvent, ExchangeCancelUpToEvent, ExchangeFillEvent } from '../../entities';
-import { bigNumbertoStringOrNull } from '../../utils';
+import { bigNumbertoStringOrNull, convertAssetProxyIdToType } from '../../utils';
+
+// tslint:disable:no-unnecessary-type-assertion
```
albrow: Again, I would prefer if we didn't disable this linter rule for the entire file.

packages/pipeline/src/parsers/events/exchange_events.ts
```diff
@@ -59,16 +59,26 @@ export function _convertToExchangeFillEvent(eventLog: LogWithDecodedArgs<Exchang
     exchangeFillEvent.takerFeePaid = eventLog.args.takerFeePaid;
     exchangeFillEvent.orderHash = eventLog.args.orderHash;
     exchangeFillEvent.rawMakerAssetData = eventLog.args.makerAssetData;
-    exchangeFillEvent.makerAssetType = makerAssetType;
+    exchangeFillEvent.makerAssetType = convertAssetProxyIdToType(makerAssetData.assetProxyId as AssetProxyId);
     exchangeFillEvent.makerAssetProxyId = makerAssetData.assetProxyId;
-    exchangeFillEvent.makerTokenAddress = makerAssetData.tokenAddress;
+    // HACK(abandeali1): this event schema currently does not support multiple maker/taker assets, so we store a null byte array when decoding assetData from the MultiAssetProxy
```
albrow: It looks like this comment is wrong and we are actually storing the maker/taker address of the first asset.

Also, it should be possible to change the schema to support this. We would need to write a new migration and update the entity definitions. I don't think we need to do it in this PR since it's already so big, but I'll add a task in Asana.

albrow: Asana task is here: https://app.asana.com/0/682899751319128/962283066289046

packages/pipeline/src/parsers/sra_orders/index.ts
```diff
@@ -1,10 +1,12 @@
...
 import { assetDataUtils, orderHashUtils } from '@0x/order-utils';
-import { AssetProxyId, ERC721AssetData } from '@0x/types';
+import { AssetProxyId, ERC721AssetData, SingleAssetData } from '@0x/types';
 import * as R from 'ramda';
 
 import { SraOrder } from '../../entities';
-import { bigNumbertoStringOrNull } from '../../utils';
+import { bigNumbertoStringOrNull, convertAssetProxyIdToType } from '../../utils';
+
+// tslint:disable:no-unnecessary-type-assertion
```
albrow: Same comment about linter rules.

packages/order-watcher/test/order_watcher_test.ts
abandeali1: I was following the style of the rest of the tests here, but not sure why they're written this way either. Any reason to not update these tests @LogvinovLeon @fabioberger ?

packages/pipeline/src/parsers/events/exchange_events.ts
abandeali1: Whoops, forgot to update the comment.

packages/order-utils/src/order_state_utils.ts
```diff
@@ -336,25 +329,19 @@ export class OrderStateUtils {
...
-                }
-                break;
-            default:
-                throw new Error(`Proxy with id ${decodedAssetData.assetProxyId} not supported`);
+        if (assetDataUtils.isERC20AssetData(decodedAssetData) || assetDataUtils.isERC721AssetData(decodedAssetData)) {
+            const allowance = await this._balanceAndProxyAllowanceFetcher.getProxyAllowanceAsync(
+                assetData,
+                traderAddress,
+            );
+            const tokenAddress = (decodedAssetData as SingleAssetData).tokenAddress;
```
albrow: You don't need this `as SingleAssetData` type assertion any more since we switched to using type guards.

packages/order-utils/src/order_state_utils.ts
```diff
@@ -336,25 +329,19 @@ export class OrderStateUtils {
...
+            allowances[tokenAddress] = _.isUndefined(initialAllowances[tokenAddress])
+                ? allowance
+                : allowances[tokenAddress].add(allowance);
+        } else if (assetDataUtils.isMultiAssetData(decodedAssetData)) {
+            for (const assetDataElement of (decodedAssetData as MultiAssetData).nestedAssetData) {
```
albrow: Same here with the `as MultiAssetData` assertion.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -13,34 +22,25 @@ export const assetDataUtils = {
...
-        }
-        const [tokenAddress] = ethAbi.rawDecode(['address'], data.slice(constants.SELECTOR_LENGTH));
+        assetDataUtils.assertIsERC20AssetData(assetData);
+        const assetProxyId = assetDataUtils.decodeAssetProxyId(assetData);
+        const abiEncoder = new AbiEncoder.Method(constants.ERC20_METHOD_ABI);
+        const decodedAssetData = abiEncoder.decode(assetData, decodingRules);
         return {
             assetProxyId,
-            tokenAddress: ethUtil.addHexPrefix(tokenAddress),
+            // TODO(abandeali1): fix return types for `AbiEncoder.Method.decode` so that we can remove type assertion
```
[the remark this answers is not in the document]
abandeali1: It will be in a new PR.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -95,24 +163,133 @@ export const assetDataUtils = {
...
+                `Could not decode assetData. Expected length of encoded data to be at least 10. Got ${
+                    assetData.length
                 }`,
             );
         }
-        const encodedAssetProxyId = encodedAssetData.slice(0, constants.SELECTOR_LENGTH);
-        const assetProxyId = decodeAssetProxyId(encodedAssetProxyId);
+        const assetProxyId = assetData.slice(0, constants.SELECTOR_CHAR_LENGTH_WITH_PREFIX);
+        if (
+            assetProxyId !== AssetProxyId.ERC20 &&
```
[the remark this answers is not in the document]
abandeali1: This would force us to cast the type of `assetProxyId` to `AssetProxyId`, which is kind of annoying.

PR #1366, opened 2018-12-01
[instant] provide a custom wallet display name option

BMillman19:
## Description

Add new top level option `walletDisplayName` that let's the integrator choose the display name of the wallet.

Add a new query param option in our dev environment to let us test this:

ex: `http://localhost:5000/?walletDisplayName=Ledger`

Also fix misaligned colored circle and wallet display name

## Testing instructions

Try out the link above

## Types of changes

## Checklist:

### DIFF
--- a/packages/instant/src/containers/connected_account_payment_method.ts
+++ b/packages/instant/src/containers/connected_account_payment_method.ts
@@ -20,6 +20,7 @@ export interface ConnectedAccountPaymentMethodProps {}
 interface ConnectedState {
     network: Network;
     providerState: ProviderState;
+    walletDisplayName?: string;
 }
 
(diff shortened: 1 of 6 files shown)


packages/instant/src/containers/connected_account_payment_method.ts
```diff
@@ -56,7 +58,7 @@ const mergeProps = (
     ...ownProps,
     network: connectedState.network,
     account: connectedState.providerState.account,
-    walletName: connectedState.providerState.name,
+    walletDisplayName: connectedState.walletDisplayName || connectedState.providerState.name,
```
fragosti: I feel like this logic should be happening just once in the factory, but oh well :P

PR #1371, opened 2018-12-04
feat(instant): More event properties for tracking

steveklebanoff:
Add eth balance, eth address, token symbol and decimals to event properties so they are logged with every event

## Checklist:


packages/instant/src/redux/analytics_middleware.ts
```diff
@@ -51,7 +54,8 @@ export const analyticsMiddleware: Middleware = store => next => middlewareAction
                     curAccount.ethBalanceInWei,
                     ETH_DECIMALS,
                 ).toString();
-                analytics.addUserProperties({ ethBalanceInUnitAmount });
+                analytics.addUserProperties({ ethBalanceInUnitAmount});
```
BMillman19: should we change this to `lastKnownEthBalanceInUnitAmount`

steveklebanoff: I actually made this change, but I didn't want to change properties that were already associated.  That being said, we haven't really launched live yet, so we _could_ do this

PR #1381, opened 2018-12-04
[instant] Magic input to trigger error for production Rollbar testing

steveklebanoff:
## Description

Allow user to type "0`" into input to trigger error for Rollbar testing

## Checklist:

### DIFF
--- a/packages/instant/src/components/scaling_amount_input.tsx
+++ b/packages/instant/src/components/scaling_amount_input.tsx
@@ -4,6 +4,7 @@ import * as React from 'react';
 
 import { Maybe } from '../types';
 
+import { MAGIC_TRIGGER_ERROR_INPUT } from '../constants';
 import { ColorOption } from '../style/theme';
 import { maybeBigNumberUtil } from '../util/maybe_big_number';
 import { util } from '../util/util';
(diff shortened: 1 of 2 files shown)


packages/instant/src/components/scaling_amount_input.tsx
```diff
@@ -71,6 +72,10 @@ export class ScalingAmountInput extends React.Component<ScalingAmountInputProps,
         );
     }
     private readonly _handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => {
+        if (event.target.value === MAGIC_TRIGGER_ERROR_INPUT) {
+            throw new Error('Triggered error');
```
fragosti: Should be include the npm version and git sha in this string just for sanity?

BMillman19: Perhaps this message should be defined as a constant or as a `ZeroExInstantError` enumeration

packages/instant/src/constants.ts
```diff
@@ -17,6 +17,7 @@ export const ACCOUNT_UPDATE_INTERVAL_TIME_MS = ONE_SECOND_MS * 5;
 export const BUY_QUOTE_UPDATE_INTERVAL_TIME_MS = ONE_SECOND_MS * 15;
 export const DEFAULT_GAS_PRICE = GWEI_IN_WEI.mul(6);
 export const DEFAULT_ESTIMATED_TRANSACTION_TIME_MS = ONE_MINUTE_MS * 2;
+export const MAGIC_TRIGGER_ERROR_INPUT = '0`';
```
fragosti: What about something you'd only be able to copy and paste? instead of being one key away from a valid input?

packages/instant/src/constants.ts
```diff
@@ -13,10 +13,14 @@ export const WEB_3_WRAPPER_TRANSACTION_FAILED_ERROR_MSG_PREFIX = 'Transaction fa
 export const GWEI_IN_WEI = new BigNumber(1000000000);
 export const ONE_SECOND_MS = 1000;
 export const ONE_MINUTE_MS = ONE_SECOND_MS * 60;
+export const GIT_SHA = process.env.GIT_SHA;
```
fragosti: shall we use these in `index.umd.ts` as well?

PR #1385, opened 2018-12-05
Refactor our protocol interfaces into `@0x/contracts-interfaces`

LogvinovLeon:
## Description

Moves protocol interfaces into a separate package so that other packages including those interfaces but being used by the protocol package (extensions) don't create a dependency cycle.

## Testing instructions

Tests still pass.

## Types of changes

## Checklist:


abandeali1:
Not sure I like this change. Why can't extensions just import from the interfaces directory to get around circular dependencies? This also seems like it's less of an issue if the extensions are in their own package.

LogvinovLeon:
@abandeali1 Extensions can't live in the same package as Exchange contract because they need to be versioned separately. Extensions depend on Exchange interfaces. But in order to test that functionality in Exchange, we use those Extensions as example implementations of those interfaces.

abandeali1:
I see, so the issue is basically with the `transactions.ts` tests. Would it solve the problem if we just moved the tests that use any extensions/examples into relevant packages? If there is any way to remove the circular dependencies that doesn't involve moving the interfaces to a separate package, that would be my preference.

abandeali1:
I see, so the issue is basically with the `transactions.ts` tests. Would it solve the problem if we just moved the tests that use any extensions/examples into relevant packages? If there is any way to remove the circular dependencies that doesn't involve moving the interfaces to a separate package, that would be my preference.

abandeali1:
I see, so the issue is basically with the `transactions.ts` tests. Would it solve the problem if we just moved the tests that use any extensions/examples into relevant packages? If there is any way to remove the circular dependencies that doesn't involve moving the interfaces to a separate package, that would be my preference.

contracts/interfaces/README.md
```diff
@@ -0,0 +1,72 @@
+## Contracts interfaces
```
fabioberger:
```suggestion
## Contract interfaces
```

contracts/interfaces/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Smart contract interfaces of the 0x protocol.
+
+## Usage
+
+Contracts that make up and interact with version 2.0.0 of the protocol can be found in the [contracts](./contracts) directory. The contents of this directory are broken down into the following subdirectories:
```
fabioberger: Think we can remove the mention of `2.0.0` specifically in all these packages now.

contracts/interfaces/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+*   [protocol](./contracts/protocol)
+    *   This directory contains the contract interfaces that make up version 2.0.0. A full specification can be found [here](https://github.com/0xProject/0x-protocol-specification/blob/master/v2/v2-specification.md).
+
+## Bug bounty
+
+A bug bounty for the 2.0.0 contracts is ongoing! Instructions can be found [here](https://0xproject.com/wiki#Bug-Bounty).
```
fabioberger: Is this still true?

LogvinovLeon: Yes

PR #1397, opened 2018-12-06
track idex orderbook snapshots

zichongkao:
## Description
track idex orderbook snapshots

## Testing instructions
usual tests

## Types of changes
new feature

## Checklist:

### DIFF
--- a/packages/pipeline/src/data_sources/idex/index.ts
+++ b/packages/pipeline/src/data_sources/idex/index.ts
@@ -1,0 +1,85 @@
+import { fetchAsync, logUtils } from '@0x/utils';
+
+const IDEX_BASE_URL = 'https://api.idex.market';
+const MARKETS_URL = `${IDEX_BASE_URL}/returnTicker`;
+const ORDERBOOK_URL = `${IDEX_BASE_URL}/returnOrderBook`;
+const MAX_ORDER_COUNT = 100; // Maximum based on https://github.com/AuroraDAO/idex-api-docs#returnorderbook
+export const IDEX_SOURCE = 'idex';
+
+export interface IdexMarketsResponse {
+    [marketName: string]: IdexMarket;
+}
+
+export interface IdexMarket {
+    last: string;
+    high: string;
+    low: string;
+    lowestAsk: string;
+    highestBid: string;
+    percentChange: string;
+    baseVolume: string;
+    quoteVolume: string;
+}
+
+export interface IdexOrderbook {
+    asks: IdexOrder[];
+    bids: IdexOrder[];
+}
+
+export interface IdexOrder {
+    price: string;
+    amount: string;
+    total: string;
+    orderHash: string;
+    params: IdexOrderParam;
+}
+
+export interface IdexOrderParam {
+    tokenBuy: string;
+    buySymbol: string;
+    buyPrecision: number;
+    amountBuy: string;
+    tokenSell: string;
+    sellSymbol: string;
+    sellPrecision: number;
+    amountSell: string;
+    expires: number;
+    nonce: number;
+    user: string;
+}
+
+// tslint:disable:prefer-function-over-method
+// ^ Keep consistency with other sources and help logical organization
+export class IdexSource {
+    /**
+     * Call Idex API to find out which markets they are maintaining orderbooks for.
+     */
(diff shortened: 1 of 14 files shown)


packages/pipeline/src/data_sources/idex/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    sellSymbol: string;
+    sellPrecision: number;
+    amountSell: string;
+    expires: number;
+    nonce: number;
+    user: string;
+}
+
+// tslint:disable:prefer-function-over-method
+// ^ Keep consistency with other sources and help logical organization
```
albrow: Just to clarify on this. Originally, I had an idea in the back of my head that we could define data sources as classes with methods and then use some higher-order code to re-use some common functionality between them. (E.g., each class could have a `getData` method and some other code could operate generically on any data source by calling that method). In practice, it hasn't really worked out that way because each data source tends to involve a third-party API with lot of weird quirks. They're too different from one another to allow for straight-forward generalization.

Let's keep it for now, but in the future we should probably just use functions instead of classes wherever it makes sense to do so.

packages/pipeline/src/data_sources/idex/index.ts
```diff
@@ -0,0 +1,85 @@
...
+export class IdexSource {
+    /**
+     * Call Idex API to find out which markets they are maintaining orderbooks for.
+     */
+    public async getMarketsAsync(): Promise<string[]> {
+        logUtils.log('Getting all IDEX markets');
```
albrow: Can we move all logs to the __scripts/___ directory instead of here? It might be important later on when we use nock for testing. We don't want too many logs polluting the test output.

packages/pipeline/src/parsers/idex_orders/index.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+/**
+ * Aggregates orders by price point for consistency with other exchanges.
+ * The Idex API returns a breakdown of individual orders at each price point.
+ * Other exchanges only give total amount at each price point.
+ * Returns an array of <price, amount> tuples.
+ * @param idexOrders A list of Idex orders awaiting aggregation.
+ */
+export function aggregateOrders(idexOrders: IdexOrder[]): Array<[string, BigNumber]> {
```
albrow: It looks like the only difference between this and the `aggregateOrders` function in the Paradex parser is the argument type. Can we generalize this and share the same function for both types of orders?

```ts
// Represents an order for an unspecified arbitrary pair. That is, it just has a price and amount.
interface UnspecifiedOrder {
    amount: BigNumber;
    price: BigNumber;
}

export function aggregateOrdersByPrice(orders: UnspecifiedOrder[]): Array<[string, BigNumber]> {
    // ...
}
```

packages/pipeline/src/parsers/oasis_orders/index.ts
```diff
@@ -0,0 +1,71 @@
...
+    const price = new BigNumber(oasisOrder[0]);
+    const amount = oasisOrder[1];
+
+    tokenOrder.source = source;
+    tokenOrder.observedTimestamp = observedTimestamp;
+    tokenOrder.orderType = orderType;
+    tokenOrder.price = price;
+
+    tokenOrder.baseAssetSymbol = oasisMarket.base;
+    tokenOrder.baseAssetAddress = ''; // Oasis doesn't provide address information
```
albrow: If Oasis doesn't provide addresses for base and quote asset, we should set it to `null` rather than an empty string. We might have to write a new migration to change the column type in the database.

packages/pipeline/src/scripts/pull_oasis_orderbook_snapshots.ts
```diff
@@ -0,0 +1,58 @@
...
+import { handleError } from '../utils';
+
+// Number of orders to save at once.
+const BATCH_SAVE_SIZE = 1000;
+
+// Number of markets to retrieve orderbooks for at once.
+const MARKET_ORDERBOOK_REQUEST_BATCH_SIZE = 50;
+
+// Delay between market orderbook requests.
+const MILLISEC_MARKET_ORDERBOOK_REQUEST_DELAY = 5000;
```
albrow: 5 seconds seems high. Is this the lowest we can go for the Oasis API?

PR #1401, opened 2018-12-07
Refactor event scraping and add support for scraping ERC20 approval events

albrow:
## Description

For now we're just scraping approval events for the WETH token, but it shouldn't be hard to add new ERC20 tokens in the future.

Backfilling all the events with the __pull_erc20_events__ script currently takes ~20 minutes for me locally. We can optimize this in the future by querying multiple block ranges simultaneously instead of doing one request at a time.

When running the __pull_erc20_events__ script, I ran into network timeout errors frequently, so I added some logic to detect this and automatically retry. That logic (along with pagination) is now shared for both exchange events and ERC20 token events.

## Checklist:


packages/pipeline/src/data_sources/contract-wrappers/erc20_events.ts
```diff
@@ -0,0 +1,45 @@
...
+            this._getApprovalEventsForRangeAsync.bind(this) as GetEventsFunc<ERC20TokenApprovalEventArgs>,
+            startBlock,
+            endBlock,
+        );
+    }
+
+    // Gets all approval events of for a specific sub-range. This getter
+    // function will be called during each step of pagination.
+    private async _getApprovalEventsForRangeAsync(
+        fromBlock: number,
```
wakkadojo: Did you mean to rename these variables "startBlock" and "endBlock"?

packages/pipeline/src/data_sources/contract-wrappers/exchange_events.ts
```diff
@@ -8,78 +8,52 @@ import {
...
-            {},
-        );
+    // Returns a getter function which gets all events of a specific type for a
+    // specific sub-range. This getter function will be called during each step
+    // of pagination.
+    private _makeGetterFuncForEventType<ArgsType extends ExchangeEventArgs>(
+        eventType: ExchangeEvents,
+    ): GetEventsFunc<ArgsType> {
+        return async (fromBlock: number, toBlock: number) =>
+            this._exchangeWrapper.getLogsAsync<ArgsType>(eventType, { fromBlock, toBlock }, {});
```
zichongkao: Nice job reusing code here! Makes it easy to understand that the three functions are doing the same things with different event types.

packages/pipeline/src/data_sources/contract-wrappers/utils.ts
```diff
@@ -0,0 +1,60 @@
+import { DecodedLogArgs, LogWithDecodedArgs } from 'ethereum-types';
```
zichongkao: Didn't see tests for these utils. They seem like core functionality which may be worth testing especially under atypical conditions like retryable and non-retryable errors.

packages/pipeline/src/parsers/events/exchange_events.ts
```diff
@@ -0,0 +1,133 @@
...
+
+import { ExchangeCancelEvent, ExchangeCancelUpToEvent, ExchangeFillEvent } from '../../entities';
+import { bigNumbertoStringOrNull } from '../../utils';
+
+/**
+ * Parses raw event logs for a fill event and returns an array of
+ * ExchangeFillEvent entities.
+ * @param eventLogs Raw event logs (e.g. returned from contract-wrappers).
+ */
+export const parseExchangeFillEvents: (
```
zichongkao: Nice design pattern at work here. Would be good for other parsers to reference this, or even build an abstraction around it. Perhaps at a later date. Looks good for now.

packages/pipeline/src/scripts/pull_missing_blocks.ts
```diff
@@ -24,7 +24,7 @@ let connection: Connection;
 (async () => {
     connection = await createConnection(ormConfig as ConnectionOptions);
     const provider = web3Factory.getRpcProvider({
-        rpcUrl: `${INFURA_ROOT_URL}/${process.env.INFURA_API_KEY}`,
+        rpcUrl: INFURA_ROOT_URL,
     });
     const web3Source = new Web3Source(provider);
     await getAllMissingBlocks(web3Source);
```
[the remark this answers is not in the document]
albrow: Yeah I noticed this. Will do in a separate PR.

packages/pipeline/src/scripts/pull_erc20_events.ts
```diff
@@ -0,0 +1,66 @@
+// tslint:disable:no-console
```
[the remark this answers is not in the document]
albrow: I'm not sure I understand what you're suggesting. This script already just does one thing: get approval events for ERC20 tokens. Is there a specific part you think we should split off?

packages/pipeline/src/scripts/pull_missing_blocks.ts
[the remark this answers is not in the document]
albrow: Yeah I noticed this. Will do in a separate PR.

packages/pipeline/src/data_sources/contract-wrappers/erc20_events.ts
albrow: Maybe this is not clear, but my intent was to distinguish between the large total range (`startBlock` to `endBlock`) and each individual sub-range within it (`fromBlock` to `toBlock`). Would a different naming scheme make it more clear?

packages/pipeline/src/data_sources/contract-wrappers/utils.ts
albrow: 👍 Good point. I just added tests in the most recent commit.

packages/pipeline/src/data_sources/contract-wrappers/erc20_events.ts
wakkadojo: i think that makes sense, i just saw you changed many of the other names from fromBlock / ... to startBlock /... and wasn't sure if you meant to change these too.

packages/pipeline/src/scripts/pull_erc20_events.ts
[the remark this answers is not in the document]
wakkadojo: I got it -- I think I mistakenly thought you were designing something more broad (like also including erc20 exchange events, but obviously those are not here).

PR #1403, opened 2018-12-07
Fix/pipeline/ohlcv ratelimit

xianny:
## Description

Use `Bottleneck` instead of `p-limit` to enforce time-based throttling. We're enforcing a wait of at least 67ms between calls (can be adjusted with env variable) and a hard limit of 1800 calls per minute. 

CC rate limit is 50/s or 2000/min. 

There's a few messy commits here so will be squashing commits before merging. 

## Testing instructions

You can run the script locally at lib/src/scripts/pull_ohlcv_cryptocompare.js.

I haven't been able to run on AWS yet (still need to get an AWS account set up). We should run on airflow to see if it works. If we encounter rate limit errors, try setting `process.env.CRYPTOCOMPARE_MAX_REQS_PER_SECOND = 10`.

## Types of changes

## Checklist:


packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
```diff
@@ -11,7 +11,7 @@ import { fetchOHLCVTradingPairsAsync, TradingPair } from '../utils/get_ohlcv_tra
 const SOURCE_NAME = 'CryptoCompare';
 const TWO_HOURS_AGO = new Date().getTime() - 2 * 60 * 60 * 1000; // tslint:disable-line:custom-no-magic-numbers
 
-const MAX_CONCURRENT_REQUESTS = parseInt(process.env.CRYPTOCOMPARE_MAX_CONCURRENT_REQUESTS || '14', 10); // tslint:disable-line:custom-no-magic-numbers
+const MAX_REQS_PER_SECOND = parseInt(process.env.CRYPTOCOMPARE_MAX_REQS_PER_SECOND || '15', 10); // tslint:disable-line:custom-no-magic-numbers
```
wakkadojo: Should the tslint be on its own line?

packages/pipeline/test/data_sources/ohlcv_external/crypto_compare_test.ts
```diff
@@ -13,7 +13,7 @@ const expect = chai.expect;
 describe('ohlcv_external data source (Crypto Compare)', () => {
     describe('generateBackfillIntervals', () => {
         it('generates pairs with intervals to query', () => {
-            const source = new CryptoCompareOHLCVSource();
+            const source = new CryptoCompareOHLCVSource(20);
```
wakkadojo: Jw why you use a different limit here than the default 15 per second

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
```diff
@@ -33,23 +33,28 @@ export interface CryptoCompareOHLCVParams {
...
 const ONE_HOUR_AGO = new Date().getTime() - ONE_HOUR;
 const HTTP_OK_STATUS = 200;
 const CRYPTO_COMPARE_VALID_EMPTY_RESPONSE_TYPE = 96;
+const MAX_PAGE_SIZE = 2000;
 
 export class CryptoCompareOHLCVSource {
-    public readonly interval = ONE_WEEK; // the hourly API returns data for one week at a time
-    public readonly default_exchange = 'CCCAGG';
     public readonly intervalBetweenRecords = ONE_HOUR;
+    public readonly default_exchange = 'CCCAGG';
```
albrow: Shoot I'm sorry I missed this before, but this variable should be in `camelCase`.

packages/pipeline/src/scripts/pull_ohlcv_cryptocompare.ts
xianny: nope, `tslint:disable-line` disables just this line.

packages/pipeline/test/data_sources/ohlcv_external/crypto_compare_test.ts
xianny: aah no reason. I also removed defaults from the class itself. Decided it was better to set defaults in just one place (entrypoint script) to be more obvious.

packages/pipeline/src/data_sources/ohlcv_external/crypto_compare.ts
wakkadojo: This is the cryptocompare name for the exchange, this is their internal "aggregate" across exchanges. maybe lower case? but otherwise i think cccagg or CCCAGG

xianny: ohh I totally missed this! thanks for catching.

PR #1409, opened 2018-12-10
[instant] Don't throw error if can't find icon for token

steveklebanoff:
## Description

Don't throw an error if we can't find an icon for a token

## Testing instructions

Use custom erc20 integration code here: https://gist.github.com/steveklebanoff/b87eef363f6df06a144426120eb731d9

## Checklist:

### DIFF
--- a/packages/instant/src/components/erc20_token_selector.tsx
+++ b/packages/instant/src/components/erc20_token_selector.tsx
@@ -7,7 +7,6 @@ import { analytics } from '../util/analytics';
 import { assetUtils } from '../util/asset';
 
 import { SearchInput } from './search_input';
-
 import { Circle } from './ui/circle';
 import { Container } from './ui/container';
 import { Flex } from './ui/flex';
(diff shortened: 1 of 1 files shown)


packages/instant/src/components/erc20_token_selector.tsx
```diff
@@ -123,10 +122,20 @@ interface TokenSelectorRowIconProps {
     token: ERC20Asset;
 }
 
+const getTokenIcon = (symbol: string): React.StatelessComponent | undefined => {
+    try {
+        return require(`../assets/icons/${symbol}.svg`) as React.StatelessComponent;
```
BMillman19: hmm dunno if we have to go as far as casting this as a stateless component

steveklebanoff: i think it's unlikely that a required svg would have state, and I like the typesafety of the null checking, but I can change back to `any` if you feel strongly

fragosti: I've always thought there should be a separate file where we iterate over all the token symbols in our data and require the icon. This also ensures that we (the developers) can't include a new token without an svg.

steveklebanoff: I think this would be a good CI task

PR #1413, opened 2018-12-10
Contracts monorepo 7

LogvinovLeon:
## Description

Refactors out extensions and examples. Now the core package has only protocol contracts. And is renamed to protocol.

## Types of changes

## Checklist:


contracts/extensions/package.json
```diff
@@ -0,0 +1,95 @@
...
+        "lint": "tslint --format stylish --project . --exclude ./generated-wrappers/**/* --exclude ./generated-artifacts/**/* --exclude **/lib/**/* && yarn lint-contracts",
+        "coverage:report:text": "istanbul report text",
+        "coverage:report:html": "istanbul report html && open coverage/index.html",
+        "profiler:report:html": "istanbul report html && open coverage/index.html",
+        "coverage:report:lcov": "istanbul report lcov",
+        "test:circleci": "yarn test",
+        "lint-contracts": "solhint -c ../.solhint.json contracts/**/**/**/**/*.sol"
+    },
+    "config": {
+        "abis": "generated-artifacts/@(DutchAuction|Forwarder|OrderValidator).json"
```
fabioberger: Hm, `OrderValidator` isn't really an "extension". It doesn't sit in-front of the `exchange` contract and act as a `sender` for orders. I think we should move it to `exchange`, since it'll become the helper function contract that we are going to ship with 2.1, and will become an integral part of our system of smart contracts offering. Thoughts @abandeali1?

packages/instant/package.json
```diff
@@ -96,6 +96,6 @@
         "webpack-dev-server": "^3.1.9"
     },
     "publishConfig": {
-        "access": "private"
+        "access": "public"
```
fabioberger: Why was this changed?

.circleci/config.yml
```diff
@@ -44,7 +44,8 @@ jobs:
             - run: yarn wsrun test:circleci @0x/contracts-utils
             - run: yarn wsrun test:circleci @0x/contracts-libs
             - run: yarn wsrun test:circleci @0x/contracts-tokens
-            - run: yarn wsrun test:circleci @0x/contracts-core
+            - run: yarn wsrun test:circleci @0x/contracts-extensions
+            - run: yarn wsrun test:circleci @0x/contracts-protocol
```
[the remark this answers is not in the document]
LogvinovLeon: They don't have any tests

packages/instant/package.json
LogvinovLeon: Good catch. Should revert

PR #1415, opened 2018-12-11
[instant] Change input to number input, and simplify ScalingInput

fragosti:
## Description

As discussed, adding type="number" does make the numeric keyboard appear on mobile, however it also causes some problems around the `onChange` event. Namely, it will allow you type in "1", "1.", "1.0" but will only trigger `onChange` for "1" and "1.0".

The final solution which I'm quite happy with is to just simplify the ScalingInput logic to measure everything in "ch" instead of moving between "ch" and "px" as before, and simply leaving enough room for the decimal at all times.

The only improvement we may want to add is that this "padding" is present at all time. It isn't very noticeable, but it should be pretty easy to make the text fit *perfectly* for when we have a decimal. We will always have to leave padding for when there is no decimal. On the other hand, I like the consistency of having some padding at all times as it leaves some room for the cursor.

**Scaling is BETTER NOW. People can buy SPANK!**

**Before: horrible SPANK buying experience**
screen shot 2018-12-13 at 1 18 32 pm

**Now: #flawless**
screen shot 2018-12-13 at 1 19 09 pm

## Testing instructions

## Checklist:

### DIFF
--- a/packages/instant/src/components/scaling_input.tsx
+++ b/packages/instant/src/components/scaling_input.tsx
@@ -13,10 +13,10 @@ export enum ScalingInputPhase {
 
 export interface ScalingSettings {
     percentageToReduceFontSizePerCharacter: number;
-    constantPxToIncreaseWidthPerCharacter: number;
 }
 
 export interface ScalingInputProps {
+    type?: string;
     textLengthThreshold: number;
     maxFontSizePx: number;
     value: string;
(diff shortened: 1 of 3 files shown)


steveklebanoff:
lol, I was considering this 'not working' b/c the scaling input didn't adjust the width when we added a decimal -- but maybe I was over-obsessing over that interaction ¯\_(ツ)_/¯

steveklebanoff:
For a little more color, this is what I was observing and was feeling like 'was broken':

- Type in "0", get onChange event with value "0"
- Add a ".", no onChange event is triggered
- Add a "1", onChange event is triggered with value "0.1"

This causes the ScalingInput to feel weird, the 'width' jumps from 1 character to 3 characters, see https://we.tl/t-sxzYasX3q4 and watch the console for the onChange events

packages/instant/src/components/scaling_input.tsx
```diff
@@ -31,32 +37,29 @@ export interface ScalingInputProps {
     hasAutofocus: boolean;
 }
 
-export interface ScalingInputState {
```
steveklebanoff: This feels much cleaner getting rid of this 💯

steveklebanoff:
This is looking awesome!  I played around quite a bit and it is working great.  Only issue I could find, but it is an edge case and I'm not sure of a fix: if I type in "0.00003" and then append a ".1" the input gets really small (b/c it receives an onChange with a blank value) and just shows the end which looks like "3.1"

fragosti:
@steveklebanoff feel like the edge case you pointed out is fine

PR #1417, opened 2018-12-11
Remove stale TODO

feuGeneA:
## Description

Fabio and I discussed the TODO item here and ultimately decided that this code already does what we need it to do.

## Testing instructions

None!  Just comments removed.

## Types of changes

## Checklist:


fabioberger:
Yep, for posterity: We removed it because the BaseProvider class already has a middleware stack of subproviders built into it.

PR #1419, opened 2018-12-12
Fix project references for contracts packages

albrow:
- Removed /packages/contracts
- Removed /contracts/core
- Added `yarn build:ts` to CI tests
- Added each contracts "project" to root tsconfig.json

## Checklist:


.circleci/config.yml
```diff
@@ -18,6 +18,7 @@ jobs:
                   name: yarn
                   command: yarn --frozen-lockfile install || yarn --frozen-lockfile install
             - run: yarn build:ci:no_website
+            - run: yarn build:ts
```
LogvinovLeon: Will that build twice on CI and increase the running time of all tests?

albrow: Yes, but only by about 1-2 seconds. The beauty of project references is that `build:ts` is pretty much instant if there are no changes to any TypeScript code 😄

albrow: This is from the latest CI run:

screen shot 2018-12-11 at 4 46 12 pm

LogvinovLeon:
LGMT after tests pass

LogvinovLeon:
Ok, looks like @fabioberger broke them on deveopment

albrow:
Since this PR fixes a bug, @fabioberger suggested merging it. The `test-publish` failures appear unrelated to the contents of this PR and we will need to address it separately.

PR #1420, opened 2018-12-12
Make pull_missing_blocks script consider all events with block numbers

albrow:
## Description

This PR fixes a bug/oversight in the __pull_missing_blocks__ script. Previously, it only looked at block numbers in the `exchange_fill_events` table. Now it considers all types of events.

## Checklist:


packages/pipeline/src/scripts/pull_missing_blocks.ts
```diff
@@ -53,8 +53,21 @@ async function getAllMissingBlocks(web3Source: Web3Source): Promise<void> {
 
 async function getMissingBlockNumbers(fromBlock: number): Promise<number[]> {
     console.log(`Checking for missing blocks starting at ${fromBlock}...`);
+    // Note(albrow): The easiest way to get all the blocks we need is to
```
albrow: Worth noting that with the current size of our tables, this query takes 295ms for me locally. Seems totally fine for now.

PR #1424, opened 2018-12-12
[instant] ETH/USD toggle

steveklebanoff:
## Description

Allow users to toggle between ETH and USD prices:

ethusdtoggle

⚠️ It'd be good to have a few folks on eyes on this before merging, as it's important these numbers are showing correctly ⚠️ 

## Testing instructions

Get quotes and ensures values are correct.  Make sure to test a few tokens, and also to test with affiliate fees

## Checklist:


packages/instant/src/components/order_details.tsx
```diff
@@ -4,124 +4,223 @@ import * as _ from 'lodash';
...
-}
 
-export interface EthAmountRowProps {
-    rowLabel: string;
-    ethAmount?: BigNumber;
-    isEthAmountInBaseUnits?: boolean;
-    ethUsdPrice?: BigNumber;
-    shouldEmphasize?: boolean;
-    isLoading: boolean;
+    private _renderErrorFetchingUsdPrice(): React.ReactNode {
```
fragosti: 👍

packages/instant/src/components/order_details.tsx
```diff
@@ -4,124 +4,223 @@ import * as _ from 'lodash';
...
+                >
+                    Click here
+                </Text>
+                {' to view ETH prices'}
+            </Text>
+        );
+    }
+
+    private _hadErrorFetchingUsdPrice(): boolean {
+        return this.props.ethUsdPrice === BIG_NUMBER_ZERO;
```
fragosti: Hopefully the name of this method holds true 🤣

packages/instant/src/components/order_details.tsx
```diff
@@ -4,124 +4,223 @@ import * as _ from 'lodash';
...
+    }
+
+    private _totalCostSecondaryValue(): React.ReactNode {
+        const secondaryCurrency = this.props.baseCurrency === BaseCurrency.USD ? BaseCurrency.ETH : BaseCurrency.USD;
+
+        const canDisplayCurrency =
+            secondaryCurrency === BaseCurrency.ETH ||
+            (secondaryCurrency === BaseCurrency.USD &&
+                this.props.ethUsdPrice &&
+                this.props.ethUsdPrice.greaterThan(BIG_NUMBER_ZERO));
```
fragosti: nit: this final expression could just be `!this._hadErrorFetchingUsdPrice()` right?

packages/instant/src/redux/async_data.ts
```diff
@@ -24,7 +24,9 @@ export const asyncData = {
             const errorMessage = 'Error fetching ETH/USD price';
             errorFlasher.flashNewErrorMessage(dispatch, errorMessage);
             dispatch(actions.updateEthUsdPrice(BIG_NUMBER_ZERO));
+            dispatch(actions.updateBaseCurrency(BaseCurrency.ETH));
```
fragosti: nice. this could be weird if we implement the eth price heartbeat but we will deal with that then.

packages/instant/src/components/order_details.tsx
steveklebanoff: haha, yes, I hope the ETH price is never $0.00 😆 

On another note, in the future we may want to have more info about the ethUsdPrice, like an enum for its status, instead of relying on a BIG_NUMBER_ZERO being set during a failure.  Figured this could be cleaned up at another point though, maybe as part of the eth usd heratbeat

packages/instant/src/components/order_details.tsx
steveklebanoff: yes, good point

packages/instant/src/redux/async_data.ts
steveklebanoff: agreed -- in the future we should have a different error handling mechanism for when the fetch is triggered by the heartbeat

packages/instant/src/components/section_header.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+
+import { ColorOption } from '../style/theme';
+
+import { Text } from './ui/text';
+
+export interface SectionHeaderProps {
+    children?: React.ReactNode;
+}
+export const SectionHeader: React.StatelessComponent<{}> = props => {
```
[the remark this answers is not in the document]
steveklebanoff: Ah, whoops, good catch - thx.  Also - we don't want to explicitly call out `children?`

packages/instant/src/components/instant_heading.tsx
```diff
@@ -113,20 +113,23 @@ export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
     }
 
     private readonly _renderEthAmount = (): React.ReactNode => {
+        const ethAmount = format.ethBaseUnitAmount(
+            this.props.totalEthBaseUnitAmount,
+            4,
+            <AmountPlaceholder isPulsating={false} color={PLACEHOLDER_COLOR} />,
+        );
+
+        const fontSize = typeof ethAmount === 'string' && ethAmount.length >= 13 ? '14px' : '16px';
```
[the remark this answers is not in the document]
steveklebanoff: I prefer to use the vanilla JS here as its only a little more verbose, esp. after our eng discussion about potentially removing lodash from packages in the future

packages/instant/src/components/instant_heading.tsx
```diff
@@ -113,20 +113,23 @@ export class InstantHeading extends React.Component<InstantHeadingProps, {}> {
     }
 
     private readonly _renderEthAmount = (): React.ReactNode => {
+        const ethAmount = format.ethBaseUnitAmount(
+            this.props.totalEthBaseUnitAmount,
+            4,
+            <AmountPlaceholder isPulsating={false} color={PLACEHOLDER_COLOR} />,
+        );
```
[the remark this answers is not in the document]
steveklebanoff: I'm a little confused about this comment, would you mind clarifying over Slack when you have time?

packages/instant/src/components/section_header.tsx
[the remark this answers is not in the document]
fragosti: `React.StatelessComponent` adds that

packages/instant/src/components/instant_heading.tsx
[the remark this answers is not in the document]
fragosti: If we removed lodash, we would probably use something else for this check. Also checking for string type in javascript is a bit tricky. https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string-in-javascript 

We would probably never run into the edge cases, but since we are already using `_.isString` in several places, it is technically more correct, and is more readable, I would prefer to use `_.isString`. Also, if and when we move to Ramda, we want to "tag" this instance for easy changing.

packages/instant/src/components/instant_heading.tsx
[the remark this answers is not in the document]
fragosti: Sorry I could have been clearer. 

In our chat I remember we said something along the lines of "lets make this work for ETH up to $10,000". Meaning that, if ETH is worth $10,000, we don't want to show the "less than" sign until the ETH value is worth less than $0.01. As far as I can tell, this is the `minUnitAmountToDisplay` you chose.

All I am saying now is: can we make `minUnitAmountToDisplay` dynamic since we do actually know the price of ETH?

steveklebanoff:
`test-publish` errors seem unrelated

PR #1425, opened 2018-12-12
Check for an Async suffix in functions, not just methods.

LogvinovLeon:
## Description
 We have a custom linter rule `async-suffix`. Before it only worked for methods. Now functions are also checked.

## Testing instructions

It detected two places and they were fixed. GJ everyone. Just two places!!!.

## Types of changes

## Checklist:


steveklebanoff:
Thank you for doing this! 🙌

PR #1426, opened 2018-12-12
Fix bug that occurs when token metadata decimals is undefined.

albrow:
## Description

Title is self-explanatory. This is a bugfix in the token metadata parser.

## Checklist:

### DIFF
--- a/packages/pipeline/src/parsers/token_metadata/index.ts
+++ b/packages/pipeline/src/parsers/token_metadata/index.ts
@@ -1,9 +1,8 @@
-import { BigNumber } from '@0x/utils';
 import * as R from 'ramda';
 
 import { MetamaskTrustedTokenMeta, ZeroExTrustedTokenMeta } from '../../data_sources/trusted_tokens';
 import { TokenMetadata } from '../../entities';
-import {} from '../../utils';
+import { toBigNumberOrNull } from '../../utils';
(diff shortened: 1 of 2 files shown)


packages/pipeline/src/parsers/token_metadata/index.ts
```diff
@@ -26,7 +26,7 @@ function parseMetamaskTrustedToken(resp: MetamaskTrustedTokenMeta, address: stri
     const trustedToken = new TokenMetadata();
 
     trustedToken.address = address;
-    trustedToken.decimals = new BigNumber(resp.decimals);
+    trustedToken.decimals = resp.decimals === undefined ? null : new BigNumber(resp.decimals);
```
albrow: @xianny the only real difference is that we needed to check if `decimals` is `undefined` in the `parseMetamaskTrustedToken` function, not just in `parseZeroExTrustedToken` function.

xianny: 👍 cool! sorry I missed this.

albrow:
@xianny 

> would `resp.decimals` ever be null?

I haven't seen a case like that by we might as well make our solution more robust. Check out the latest changes.

xianny:
Cool. Looks good!

PR #1427, opened 2018-12-12
OrderWatcher WebSocket Server

zichongkao:
## Description
Implements a WebSocket server which wraps OrderWatcher

## Testing instructions
Standard yarn test 

## Types of changes

## Checklist:


packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+import { OnOrderStateChangeCallback, OrderWatcherConfig } from '../types';
+import { assert } from '../utils/assert';
+
+import { OrderWatcher } from './order_watcher';
+
+const DEFAULT_HTTP_PORT = 8080;
+
+const enum OrderWatcherAction {
+    // Actions initiated by the user.
+    getStats = 'getStats',
```
fabioberger: Our enum conventions coincide with recommended Typescript conventions and are capitalized keys, and all-caps snake-case values.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+    addOrderAsync = 'addOrderAsync',
+    removeOrder = 'removeOrder',
+    // These are spontaneous; they are primarily orderstate changes.
+    orderWatcherUpdate = 'orderWatcherUpdate',
```
fabioberger: Just `update` is fine. No need to duplicate `orderWatcher` since it'll appear as `OrderWatcherAction.Update`

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+    params: any;
+}
+
+// Users should expect a json object of this format in the data field
+// of the WebSocket messages that this server sends out.
+interface WebSocketResponseData {
+    action: OrderWatcherAction;
+    success: number;
+    result: any;
+}
```
fabioberger: Can we move these types to the `types.ts` file?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+// Wraps the OrderWatcher functionality in a WebSocket server. Motivations:
+// 1) Users can watch orders via non-typescript programs.
+// 2) Better encapsulation so that users can work
+export class OrderWatcherWebSocketServer {
+    public httpServer: http.Server;
+    public readonly _orderWatcher: OrderWatcher;
+    private readonly _connectionStore: Set<WebSocket.connection>;
+    private readonly _wsServer: WebSocket.server;
+    /**
+     * Instantiate a new web socket server which provides OrderWatcher functionality
```
fabioberger:
```suggestion
     * Instantiate a new WebSocket server which provides OrderWatcher functionality
```

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+     *  @param provider Web3 provider to use for JSON RPC calls (for OrderWatcher)
+     *  @param networkId NetworkId to watch orders on (for OrderWatcher)
+     *  @param contractAddresses Optional contract addresses. Defaults to known
+     *  addresses based on networkId (for OrderWatcher)
+     *  @param partialConfig Optional configurations (for OrderWatcher)
```
fabioberger: Let's omit the `(for OrderWatcher)` here, it's not necessary to distinguish.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        networkId: number,
+        contractAddresses?: ContractAddresses,
+        partialConfig?: Partial<OrderWatcherConfig>,
+    ) {
+        this._orderWatcher = new OrderWatcher(provider, networkId, contractAddresses, partialConfig);
+        this._connectionStore = new Set();
+        this.httpServer = http.createServer();
+        this._wsServer = new WebSocket.server({
+            httpServer: this.httpServer,
+            autoAcceptConnections: false,
```
fabioberger: Can you add a comment why this is hard-coded to `false`?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        });
+
+        this._wsServer.on('request', (request: any) => {
+            // Designed for usage pattern where client and server are run on the same
+            // machine by the same user. As such, no security checks are in place.
+            const connection: WebSocket.connection = request.accept(null, request.origin);
+            logUtils.log(`${new Date()} [Server] Accepted connection from origin ${request.origin}.`);
+            connection.on('message', this._messageCallback.bind(this, connection));
```
fabioberger:
```suggestion
            connection.on('message', this._onMessageCallback.bind(this, connection));
```

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+            connection.on('close', this._closeCallback.bind(this, connection));
```
fabioberger:
```suggestion
            connection.on('close', this._onCloseCallback.bind(this, connection));
```

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        // Have the WebSocket server subscribe to the OrderWatcher to receive updates.
+        // These updates are then broadcast to clients in the _connectionStore.
+        this._orderWatcher.subscribe(this._broadcastCallback);
+    }
+
+    /**
+     * Activates the WebSocket server by having its HTTP server start listening.
+     */
+    public listen(): void {
+        this.httpServer.listen(DEFAULT_HTTP_PORT, () => {
```
fabioberger:
```suggestion
        const port  = process.env.ORDER_WATCHER_HTTP_PORT || DEFAULT_HTTP_PORT;
        this.httpServer.listen(port, () => {
```

fabioberger: Can you also add a WebSocket documentation section to the README, explaining this option of OrderWatcher operation, and mention the available environment variables?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+     */
+    constructor(
+        provider: Provider,
+        networkId: number,
+        contractAddresses?: ContractAddresses,
+        partialConfig?: Partial<OrderWatcherConfig>,
+    ) {
+        this._orderWatcher = new OrderWatcher(provider, networkId, contractAddresses, partialConfig);
+        this._connectionStore = new Set();
+        this.httpServer = http.createServer();
```
fabioberger: Why is the `httpServer` public?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+    /**
+     * Activates the WebSocket server by having its HTTP server start listening.
+     */
+    public listen(): void {
+        this.httpServer.listen(DEFAULT_HTTP_PORT, () => {
+            logUtils.log(`${new Date()} [Server] Listening on port ${DEFAULT_HTTP_PORT}`);
+        });
+    }
+
+    public close(): void {
```
fabioberger: Please add method comment since this is a public, exposed method.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        this.httpServer.close();
+    }
+
+    private _messageCallback(connection: WebSocket.connection, message: any): void {
+        assert.doesConformToSchema('message', message, webSocketUtf8MessageSchema);
+        const requestData: WebSocketRequestData = JSON.parse(message.utf8Data);
```
fabioberger: Sorry if my comment wasn't clear but we actually need to validate that `requestData` has the correct structure using JSON schema. We need to ensure that is has the `action` key, and that depending on the action key, that is has the associated/expected `param`. E.g if the `key` is `addOrderAsync`, then the `param` is a `signedOrder`.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+import { assert } from '../utils/assert';
+
+import { OrderWatcher } from './order_watcher';
+
+const DEFAULT_HTTP_PORT = 8080;
+
+const enum OrderWatcherAction {
+    // Actions initiated by the user.
+    getStats = 'getStats',
+    addOrderAsync = 'addOrderAsync',
```
fabioberger: Let's omit the `async` prefix here, it doesn't make sense.

packages/order-watcher/test/order_watcher_websocket_test.ts
```diff
@@ -0,0 +1,226 @@
...
+            const updateData = JSON.parse(updateMsg.data);
+            const orderState = updateData.result as OrderStateValid;
+            expect(orderState.isValid).to.be.true();
+            expect(orderState.orderRelevantState.makerFeeProxyAllowance).to.be.eq('0');
+        }
+
+        wsClientTwo.close();
+        logUtils.log(`${new Date()} [Client] Closed.`);
+    });
+});
```
fabioberger: Additional tests I'd like to see:

- Client broadcasts payload with `action` `addOrderAsync` but with `params` containing gibberish.
- Client broadcasts payload with `action` `addOrderAsync` but the supplied `signedOrder` is invalid. This will check that the fire-and-forget doesn't cause any problems.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+        this._connectionStore.forEach((connection: WebSocket.connection) => {
+            const responseData: WebSocketResponseData = {
+                action: OrderWatcherAction.orderWatcherUpdate,
+                success: 1,
+                result: orderState || err,
+            };
+            connection.sendUTF(JSON.stringify(responseData));
+        });
+        // tslint:disable-next-line:semicolon
+    }; // tslint thinks this is a class method, It's actally a property that holds a function.
```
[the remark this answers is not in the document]
zichongkao: Hm when I make it a class method, the tslint rule prefer-function-to-method calls me out because these functions don't use `this`. That made sense to me because the functions just wrap a bunch of code instead of mutating aspects of the object (as methods typically do). 

But I don't feel strongly. Should I overwrite the `prefer-function-over-method` lint rule and change these to class methods?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+     * Broadcasts OrderState changes to ALL connected clients. At the moment,
+     * we do not support clients subscribing to only a subset of orders. As such,
+     * Client B will be notified of changes to an order that Client A added.
+     */
+    private readonly _broadcastCallback: OnOrderStateChangeCallback = (err, orderState) => {
+        this._connectionStore.forEach((connection: WebSocket.connection) => {
+            const responseData: WebSocketResponseData = {
+                action: OrderWatcherAction.orderWatcherUpdate,
+                success: 1,
+                result: orderState || err,
```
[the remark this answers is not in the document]
zichongkao: Did not fix this one yet.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,184 @@
...
+            result: undefined,
+        };
+
+        try {
+            logUtils.log(`${new Date()} [Server] Request received: ${requestData.action}`);
+            switch (requestData.action) {
+                case 'addOrderAsync': {
+                    const signedOrder: SignedOrder = this._parseSignedOrder(requestData);
+                    // tslint:disable-next-line:no-floating-promises
+                    this._orderWatcher.addOrderAsync(signedOrder); // Ok to fireNforget
```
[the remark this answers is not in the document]
zichongkao: Ended up switching this to standard await. I guess it might increase latency slightly but better for reliability.

packages/order-watcher/test/order_watcher_websocket_test.ts
zichongkao: Both added!

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,173 @@
...
+    ) {
+        this._orderWatcher = new OrderWatcher(provider, networkId, contractAddresses, partialConfig);
+        this._connectionStore = new Set();
+        this._httpServer = http.createServer();
+        this._wsServer = new WebSocket.server({
+            httpServer: this._httpServer,
+            // Avoid setting autoAcceptConnections to true as it defeats all
+            // standard cross-origin protection facilities built into the protocol
+            // and the browser. Also ensures that a request event is emitted by
+            // the server whenever a new WebSocket request is made.
```
fabioberger: We like to add `// source: ` links when comments include information for a specific place. If you just knew this, then no need to add one 😋

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,173 @@
...
+            autoAcceptConnections: false,
+        });
+
+        this._wsServer.on('request', async (request: any) => {
+            // Designed for usage pattern where client and server are run on the same
+            // machine by the same user. As such, no security checks are in place.
+            const connection: WebSocket.connection = request.accept(null, request.origin);
+            logUtils.log(`${new Date()} [Server] Accepted connection from origin ${request.origin}.`);
+            connection.on('message', await this._onMessageCallbackAsync.bind(this, connection));
```
fabioberger: Keck? Awaiting here does nothing. Why is this here?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,173 @@
...
+            connection.on('close', this._onCloseCallback.bind(this, connection));
+            this._connectionStore.add(connection);
+        });
+
+        // Have the WebSocket server subscribe to the OrderWatcher to receive updates.
+        // These updates are then broadcast to clients in the _connectionStore.
+        const broadcastCallback: OnOrderStateChangeCallback = this._broadcastCallback.bind(this);
```
fabioberger: Why is this line necessary?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,173 @@
...
+        logUtils.log(`${new Date()} [Server] Request received: ${request.action}`);
+        let result = null;
+        switch (request.action) {
+            case OrderWatcherAction.AddOrder: {
+                const signedOrder: SignedOrder = OrderWatcherWebSocketServer._parseSignedOrder(request.signedOrder);
+                await this._orderWatcher.addOrderAsync(signedOrder);
+                break;
+            }
+            case OrderWatcherAction.RemoveOrder: {
+                const orderHash = request.orderHash || '_';
```
fabioberger: This looks like a hack. Can you explain?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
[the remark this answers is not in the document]
fabioberger: Let's move to a [json-rpc](https://www.jsonrpc.org/specification) compliant message format.

packages/order-watcher/src/schemas/websocket_schemas.ts
```diff
@@ -0,0 +1,32 @@
...
+    type: 'object',
+    required: ['utf8Data'],
+};
+
+export const webSocketRequestSchema = {
+    id: '/webSocketRequestSchema',
+    properties: {
+        action: { enum: ['GET_STATS', 'ADD_ORDER', 'REMOVE_ORDER'] },
+        signedOrder: { $ref: '/signedOrderSchema' },
+        orderHash: { type: 'string' },
```
fabioberger: Let's use `hexSchema` here, it's stricter.

packages/order-watcher/src/types.ts
```diff
@@ -31,3 +31,31 @@ export enum InternalOrderWatcherError {
...
+export interface WebSocketRequest {
+    action: OrderWatcherAction;
+    signedOrder?: SignedOrder;
+    orderHash?: string;
+}
+
+// Users should expect a json object of this format in the data field
+// of the WebSocket messages that the server sends out.
+export interface WebSocketResponse {
+    action: OrderWatcherAction | null;
```
fabioberger: When is this expected to be `null`?

packages/order-watcher/src/types.ts
```diff
@@ -31,3 +31,31 @@ export enum InternalOrderWatcherError {
...
+    success: boolean;
+    result: any;
```
fabioberger: Rather then use `any` here, let's use a union type. Similar to this example for elsewhere in the order-watcher:

```
export interface OrderStateValid {
    isValid: true;
    orderHash: string;
    orderRelevantState: OrderRelevantState;
    transactionHash?: string;
}

export interface OrderStateInvalid {
    isValid: false;
    orderHash: string;
    error: ExchangeContractErrs;
    transactionHash?: string;
}

export type OrderState = OrderStateValid | OrderStateInvalid;
```

But instead of hard-coding `isValid`, it would be the different `action` values. Then in the switch statement, once we verify the `action` value, we cast the response object to the appropriate type. This approach will give us better type-safety.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
zichongkao: To tell the compiler that this._broadcastCallback is actually of type OnOrderStateChangeCallback. The underlying types match, but I can't make it recognize that. If I pass _broadcastCallback to subscribe directly, I get a type error. Any better options?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
zichongkao: Yeah. I'm trying to get around the type checker. Because request.orderHash could be null if it's an ADD_ORDER request, the type of request.orderHash is string or null, and this can't be passed to orderHash which expects a string. Ideally the type checker would know we're guarded by OrderWatcherAction.RemoveOrder and so request.orderHash will definitely not be null. Open to better workarounds too. This is a common problem I face with typescript.

packages/order-watcher/src/schemas/websocket_schemas.ts
zichongkao: cool. that's exactly what I was looking for.

packages/order-watcher/src/types.ts
zichongkao: It's null when the user inputs a rubbish action, or worse still, tries to input a request without an action field.  In that case, we respond with an error message, but the error message uses action = null. Do you think this is ok?

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
zichongkao: Yeah hm I thought I had to await async stuff. I guess it sounds like i don't. Will fix.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
fabioberger: You can cast it to `OnOrderStateChangeCallback` when passing it in.

packages/order-watcher/README.md
```diff
@@ -26,6 +29,84 @@ If your project is in [TypeScript](https://www.typescriptlang.org/), add the fol
...
+    and accept connections from. When this is not set, we default to 8080.
+
+**Requests**
+The server accepts three types of requests: `ADD_ORDER`, `REMOVE_ORDER` and `GET_STATS`. These mirror what the underlying OrderWatcher does. You can read more in the [wiki](https://0xproject.com/wiki#0x-OrderWatcher). Unlike the OrderWatcher, it does not expose any subscribe or unsubscribe functionality because the client implicitly subscribes and unsubscribes by connecting to the server.
+
+The first step for making a request is establishing a connection with the server. In Javascript:
+
+```
+var W3CWebSocket = require('websocket').w3cwebsocket;
+wsClient = new WebSocket.w3cwebsocket('ws://127.0.0.1:8080');
```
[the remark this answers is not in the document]
zichongkao: My bad. Will fix.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
fabioberger: Yeah, if you do the union type suggestion I also commented on, this issue will go away.

packages/order-watcher/src/types.ts
fabioberger: Not sure I follow.

packages/order-watcher/src/order_watcher/order_watcher_websocket.ts
```diff
@@ -0,0 +1,173 @@
...
+                this._orderWatcher.removeOrder(orderHash);
+                break;
+            }
+            case OrderWatcherAction.GetStats: {
+                result = this._orderWatcher.getStats();
+                break;
+            }
+            default:
+                // Should never reach here. Should be caught by JSON schema check.
+                throw new Error(`[Server] Invalid request action: ${request.action}`);
```
fabioberger:
```suggestion
                throw new Error(`[Server] Invalid request action: ${request.action || 'undefined'}`);
```

fabioberger: Would this solve the same issue as the `| null` type definition mentioned in your other comment?

packages/order-watcher/src/types.ts
fabioberger: This should not be necessary. I offered an alternative approach in a comment on the error line.

zichongkao:
Alright! This is my last push. Apologies that I couldn't see it through. Hope you can take it from here!
I've tried to make the response and request formats JSON RPC compliant. This meant going back on the earlier decision to move signedOrder and orderHash out of the params field. (So now they are back in). I've also used the union types like you suggested. They worked pretty well!

Current gaps:
- JSON RPC prefers that errors have an error code. I've not implemented error codes.
- Setup script for the WebSocket server has not been implemented and there is an empty section for this in the README.

fabioberger:
Thanks Kao! I'll take care of it from here! 🙏

PR #1428, opened 2018-12-13
add special case to scrape OHLCV for eth/usd

xianny:
## Description
Small fix to start scraping ETH/USD OHLCVs

I also renamed some variables to distinguish where we were getting tokens from. 

## Testing instructions

I've run the script locally but not on QA. Script works locally.

## Types of changes

## Checklist:

### DIFF
--- a/packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
+++ b/packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
@@ -23,6 +23,18 @@ interface CryptoCompareCoin {
 const TO_CURRENCIES = ['USD', 'EUR', 'ETH', 'USDT'];
 const ETHEREUM_IDENTIFIER = '7605';
 const HTTP_OK_STATUS = 200;
+
+interface StaticPair {
+    fromSymbol: string;
+    toSymbol: string;
+}
+const SPECIAL_CASES: StaticPair[] = [
+    {
+        fromSymbol: 'ETH',
+        toSymbol: 'USD',
+    },
+];
+
 /**
  * Get trading pairs with latest scraped time for OHLCV records
  * @param conn a typeorm Connection to postgres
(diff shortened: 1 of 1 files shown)


packages/pipeline/src/utils/get_ohlcv_trading_pairs.ts
```diff
@@ -66,27 +85,29 @@ export async function fetchOHLCVTradingPairsAsync(
...
       SELECT DISTINCT(taker_token_address) as tokenaddress FROM raw.exchange_fill_events`,
     );
-    const tokenAddresses = R.pluck('tokenaddress', rawTokenAddresses);
+    const eventTokenAddresses = R.pluck('tokenaddress', rawEventTokenAddresses);
 
     // join token addresses with CC symbols
-    const allTokenSymbols: string[] = tokenAddresses
+    const eventTokenSymbols: string[] = eventTokenAddresses
         .map(tokenAddress => erc20CoinsIndex.get(tokenAddress.toLowerCase()) || '')
         .filter(x => x);
```
albrow: What's the point of this `filter`?

xianny: this gets rid of the empty strings in the previous step (or any falsey values). I can change it to be more explicit though

albrow: Let's be more explicit. I don't like making assumptions based on JavaScript "truthiness" rules because they can be pretty unintuitive sometimes.

albrow:
@xianny and I talked offline. Approved after one comment addressed and CI passes.

xianny:
Tried a few things;

Compiler complains that this is actually `(string | undefined) []` because it doesn't recognise `defaultTo`:
```
const eventTokenSymbols: string[] = R.chain(tokenAddress => {
        return R.defaultTo([erc20CoinsIndex.get(tokenAddress.toLowerCase())], []);
    }, eventTokenAddresses);
```

Compiler complains about type signature:
```
const eventTokenSymbols: string[] = eventTokenAddresses
    .map(tokenAddress => erc20CoinsIndex.get(tokenAddress.toLowerCase()) || '')
    .filter(!R.isEmpty)
```

This works:
```
    const eventTokenSymbols: string[] = R.chain(tokenAddress => {
        const sym = erc20CoinsIndex.get(tokenAddress.toLowerCase()) || 'undefined';
        return R.defaultTo([sym], []);
    }, eventTokenAddresses);
```

This also works:
```
const eventTokenSymbols: string[] = eventTokenAddresses
    .map(tokenAddress =>  erc20CoinsIndex.get(tokenAddress.toLowerCase()) || '')
    .filter(x => !R.isEmpty(x))
```

my intuition says the last one is most readable, but @albrow interested in what you think?

PR #1430, opened 2018-12-13
Pull approval events for ZRX and DAI

albrow:
## Description

This PR adds two new ERC20 tokens for which we will scrape approval events: ZRX and DAI. We can add more tokens in the future, but this seems like a good starting point.

## Checklist:

### DIFF
--- a/packages/pipeline/src/scripts/pull_erc20_events.ts
+++ b/packages/pipeline/src/scripts/pull_erc20_events.ts
@@ -1,10 +1,10 @@
-// tslint:disable:no-console
 import { getContractAddressesForNetworkOrThrow } from '@0x/contract-addresses';
 import { web3Factory } from '@0x/dev-utils';
 import { Web3ProviderEngine } from '@0x/subproviders';
+import { logUtils } from '@0x/utils';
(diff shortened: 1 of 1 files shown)


packages/pipeline/src/scripts/pull_erc20_events.ts
```diff
@@ -1,10 +1,10 @@
-// tslint:disable:no-console
```
albrow: As per our discussion in the last Eng All Hands, we're going to switch to using `logUtils` for scripts in the pipeline package. I went ahead and did that for this file since I was working on it anyways. Planning to do a pass through all the other scripts in a future PR.

wakkadojo:
Makes sense to me, but I can't speak to the web3 calls. In particular the logic 
```
provider.stop();
(provider as any)._ready.go();
```

albrow:
@wakkadojo that code is a way to stop automatic block polling in `Web3ProviderEngine`. (It sends requests about every second and was mudding up the request logs). It's probably not appropriate to leave it in this PR, so I removed it. I created an Asana task to add the option to disable block polling in a more proper way.

PR #1431, opened 2018-12-13
Feature/subproviders/trezor subprovider

MarcZenn:
## Description

This provides a TrezorSubprovider class within 0x/subproviders package. It helps dapps integrate trezor support in conjunction with web3 provider engine so that users may retrieve their account, sign transactions, & sign personal messages using their Trezor device.

Please note that the `TrezorSubprovider` class MUST be initialized with `trezor-connect` module as such: 

```
import TrezorConnect from 'trezor-connect';
import { TrezorSubprovider } from '@0x/subproviders';

const trezorSubproviderConfig = { trezorConnectClientApi: TrezorConnect  };

const trezorSubprovider - new TrezorSubprovider(trezorSubproviderConfig);

```

## Testing instructions

This could be setup into a codesandbox and tested with a Trezor device so that use is able to retrieve their Trezor ethereum address, sign transactions via `eth_sign` and sign personal messages i.e. `personal_sign`.

## Types of changes

## Checklist:

*   [ ] Add tests to cover changes as needed. - might need some help with this!
*   [ ] Update documentation as needed. - where/how can I update the docs?


packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+        txData.value = txData.value ? txData.value : '0x0';
+        txData.data = txData.data ? txData.data : '0x';
+        txData.gas = txData.gas ? txData.gas : '0x0';
+        txData.gasPrice = txData.gasPrice ? txData.gasPrice : '0x0';
+
+        const accountIndex = this._cachedAccounts.indexOf(txData.from);
+
+        const response: TrezorConnectResponse = await this._trezorConnectClientApi.ethereumSignTransaction({
+            path: this._publicKeyPath + `${accountIndex}`,
```
fabioberger: Our convention is to always use template strings instead of concatenation. Please fix everywhere in this file.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+            transaction: {
+                to: txData.to,
+                value: txData.value,
+                data: txData.data,
+                chainId: 1,
```
fabioberger: Why is this hard-coded? They might be on a testnet. Is this a required property? If so, let's have them pass in the `networkId` to the constructor.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+     * @return TrezorSubprovider instance
+     */
+    constructor(config: TrezorSubproviderConfig) {
+        super();
+        this._publicKeyPath = PRIVATE_KEY_PATH;
+        this._cachedAccounts = [];
+        this._trezorConnectClientApi = config.trezorConnectClientApi;
+    }
+    /**
+     * Retrieve a users Trezor account. The accounts are private key path derived, This method
```
fabioberger:
```suggestion
     * Retrieve a users Trezor account. This method
```

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+    public async signTransactionAsync(txData: PartialTxParams): Promise<string> {
+        if (_.isUndefined(txData.from) || !addressUtils.isAddress(txData.from)) {
+            throw new Error(WalletSubproviderErrors.FromAddressMissingOrInvalid);
+        }
+        txData.value = txData.value ? txData.value : '0x0';
+        txData.data = txData.data ? txData.data : '0x';
+        txData.gas = txData.gas ? txData.gas : '0x0';
+        txData.gasPrice = txData.gasPrice ? txData.gasPrice : '0x0';
+
+        const accountIndex = this._cachedAccounts.indexOf(txData.from);
```
fabioberger: This will be `-1` if they supply a `from` address that is not in `_cachedAccounts`. We should throw with a helpful error message in this case.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+     * @param address Address of the account to sign with
+     * @return Signature hex string (order: rsv)
+     */
+    public async signPersonalMessageAsync(data: string, address: string): Promise<string> {
+        if (_.isUndefined(data)) {
+            throw new Error(WalletSubproviderErrors.DataMissingForSignPersonalMessage);
+        }
+        assert.isHexString('data', data);
+        assert.isETHAddressHex('address', address);
+        const accountIndex = this._cachedAccounts.indexOf(address);
```
fabioberger: Could be `-1`, handle this case.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+        this._cachedAccounts = [];
+        this._trezorConnectClientApi = config.trezorConnectClientApi;
+    }
+    /**
+     * Retrieve a users Trezor account. The accounts are private key path derived, This method
+     * is automatically called when issuing a `eth_accounts` JSON RPC request via your providerEngine
+     * instance.
+     * @return An array of accounts
+     */
+    public async getAccountsAsync(): Promise<string[]> {
```
fabioberger: Let's pass in an amount they wish to fetch and fetch multiple.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+        if (this._cachedAccounts.length) {
+            return this._cachedAccounts;
+        }
+        const accounts: string[] = [];
+        const response: TrezorConnectResponse = await this._trezorConnectClientApi.ethereumGetAddress({
+            path: this._publicKeyPath,
+            showOnTrezor: true,
```
fabioberger: If we default this to `false`, it's less cumbersome to fetch multiple addresses. My understanding is that this doesn't actually make the process more secure, it just gives the user a false sense of security.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+    /**
+     * eth_signTypedData is currently not supported on Trezor devices.
+     * @param address Address of the account to sign with
+     * @param data the typed data object
+     * @return Signature hex string (order: rsv)
+     */
+    // tslint:disable-next-line:prefer-function-over-method
+    public async signTypedDataAsync(address: string, typedData: any): Promise<string> {
+        throw new Error(WalletSubproviderErrors.MethodNotSupported);
```
fabioberger: Let's add a `// TODO` comment here so it's clear that this is a remaining TODO for a future dev.

packages/subproviders/src/types.ts
```diff
@@ -136,3 +135,38 @@ export type NextCallback = (callback?: OnNextCompleted) => void;
...
+export interface TrezorSubproviderConfig {
+    trezorConnectClientApi: any;
+}
+
+export interface TrezorGetAddressResponsePayload {
+    address: string;
+    path: {
+        0: number;
+        1: number;
+        2: number;
```
fabioberger: Are there only 2 paths? If there are more, let's use type: `{[index: number]: number}` instead.

packages/subproviders/src/types.ts
```diff
@@ -136,3 +135,38 @@ export type NextCallback = (callback?: OnNextCompleted) => void;
...
+    serializedPath: string;
+}
+
+export interface TrezorSignTxResponsePayload {
+    v: string;
+    r: string;
+    s: string;
+}
+
+export interface TrezorSignMssgResponsePayload {
```
fabioberger:
```suggestion
export interface TrezorSignMsgResponsePayload {
```

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+    private readonly _trezorConnectClientApi: any;
+    /**
+     * Instantiates a TrezorSubprovider. Defaults to private key path set to `44'/60'/0'`.
+     * Must be initialized with trezor-connect API module https://github.com/trezor/connect.
+     * @param TrezorSubprovider config object containing trezor-connect API
+     * @return TrezorSubprovider instance
+     */
+    constructor(config: TrezorSubproviderConfig) {
+        super();
+        this._publicKeyPath = PRIVATE_KEY_PATH;
```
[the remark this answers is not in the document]
MarcZenn: member var should now be `this._privateKeyPath` and consistent throughout.

packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -0,0 +1,160 @@
...
+        }
+        const accounts: string[] = [];
+        const response: TrezorConnectResponse = await this._trezorConnectClientApi.ethereumGetAddress({
+            path: this._publicKeyPath,
+            showOnTrezor: true,
+        });
+
+        if (response.success) {
+            const payload: TrezorGetAddressResponsePayload = response.payload;
+            accounts.push(payload.address);
```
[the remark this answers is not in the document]
MarcZenn: You're right. We can fetch multiple via public key. Ive updated this and it should fetch multiple addresses but I had to add a dependency `ethereumjs-wallet/hdkey`to do it. I suspect you have a util somewhere that handles this and eliminates the need for that lib?

packages/subproviders/src/subproviders/trezor.ts
MarcZenn: updated

packages/subproviders/src/subproviders/trezor.ts
MarcZenn: Done.

packages/subproviders/src/subproviders/trezor.ts
MarcZenn: Updated. Should either throw an `ADDRESS_NOT_FOUND` subprovider error for signing personal message and `FROM_ADDRESS_MISSING_OR_INVALID` when signing tx.

packages/subproviders/src/subproviders/trezor.ts
MarcZenn: Donezo

packages/subproviders/src/subproviders/trezor.ts
MarcZenn: added

packages/subproviders/src/subproviders/trezor.ts
fabioberger: Thoughts on this @MarcZenn?

MarcZenn:
@fabioberger Take a look now. I've removed caching of addresses and am utilizing wallet_utils given the trezor public key request response payload.

NoahZinsmeister:
Hi @MarcZenn / @fabioberger ! Just wanted to check in before this gets merged.

The private [`_initialDerivedKeyInfoAsync`](https://github.com/0xProject/0x-monorepo/blob/1bff79062839319cd8a557a600aad06c675555e4/packages/subproviders/src/subproviders/trezor.ts#L163) method gets called in each of the core functions: `getAccountsAsync`, `signTransactionAsync`, and `signPersonalMessageAsync`. However, the output of this function doesn't change according to the calling context, and every call after the first unnecessarily prompts the user to agin export their public key, which leads to pretty bad UX.

I'd strongly propose that `_initialDerivedKeyInfoAsync` be modified to cache the result after the first (successful) call, and return appropriately thereafter. I would be more than happy to contribute this (fairly small) fix, then we could move forward with merging?

fabioberger:
Thanks for chiming in @NoahZinsmeister. That change makes sense to me. What do you think @MarcZenn? If I don't hear from him in the next week, I'll comandeer this PR and get it merged. Do you need it soon @NoahZinsmeister?

NoahZinsmeister:
@fabioberger I'm using a ported version for now, so no huge rush. Looking forward to merge though!

NoahZinsmeister:
@fabioberger it'd be nice to have this sooner rather than later, if you ever have a free moment :)

rahul70392:
Hello, I am waiting for this merge. Please !

fabioberger:
Sorry for the delay @rahul70392, working on it right now.

rahul70392:
Thank you!

fabioberger:
@rahul70392, just published a new version of @0x/subproviders to NPM that includes to Trezor subprovider. Happy hacking!

Issue #1433, opened 2018-12-14
0x Instant (buy_button.ts) Unhandled Promise Rejection


hewigovens:
Hi

We tested 0x Instant on a mobile wallet (Trust) and found an error in `buy_button.ts` when canceling an order, below is the screenshot:

![screen shot 2018-12-14 at 09 55 49](https://user-images.githubusercontent.com/360470/49979133-a2fe2580-ff88-11e8-8365-85b6ac711683.png)

We find that `BuyButton` only handles specific error messages, like `AssetBuyerError.SignatureRequestDenied` (https://github.com/0xProject/0x-monorepo/blob/f076cdb8329693e9d7351ec88d2ac01ad42a2a3a/packages/instant/src/components/buy_button.tsx#L85)

And if we change the error message to `SIGNATURE_REQUEST_DENIED`, we can cancel the order correctly, but not every dApp browser emits the same error message.

## Expected Behavior

0x Instant should cancel the order when an error occurred

## Current Behavior

please check the screenshot above, it gets stuck at "Placing Order..."

## Possible Solution

- configurable error messages
- add a default error handler

## Steps to Reproduce (for bugs)

https://0x-instant-test.surge.sh/

```
1. Open Trust Wallet and navigate to the test website
2. Place a order then cancel it
```

## Context

User may also use 0x Instant in mobile wallets

## Your Environment

| Package | Version |
|---------|--------|
| 0x Instant | https://instant.0xproject.com/instant.js |

| Network |
|---------|
| mainnet  |

BMillman19:
@hewigovens thanks for the report, will be looking into it today

BMillman19:
@hewigovens try this link out: http://0x-instant-staging.s3-website-us-east-1.amazonaws.com/

BMillman19:
@hewigovens btw, can we rely on `TrustWeb3Provider` as a stable constructor name for us to identify the provider as trust wallet or is there some recommended way?

hewigovens:
@BMillman19 Yes, `TrustWeb3Provider` is also used in MyCrypto.com to identify Trust Wallet

hewigovens:
And thanks for the fix, It works!

BMillman19:
hey @hewigovens what error do you guys throw when the signature gets denied?

hewigovens:
@BMillman19 right now only a simple "cancelled"

BMillman19:
@hewigovens this should be fixed!

PR #1434, opened 2018-12-14
[instant] fix signature denial error message and add trust wallet detection

BMillman19:
## Description

https://github.com/0xProject/0x-monorepo/issues/1433

* In the eyes of analytics and UX, treat generic errors coming from `executeBuyQuoteAsync` as signature denials
* In the case we don't recognize the error, send it to rollbar

## Testing instructions

http://0x-instant-staging.s3-website-us-east-1.amazonaws.com/

## Types of changes

## Checklist:


steveklebanoff:
@BMillman19 This feels a bit scary to me -- I feel like many different errors could come from "executeBuyQuoteAsync", and it would be weird if we got an unexpected error and told the user they denied their signature.

I wonder if we could be more specific, and look for the words "cancelled" or "denied", and show the denied message then, but otherwise not crash, but instead just flash the message "An error has occurred, please try again"

packages/instant/src/util/env.ts
```diff
@@ -44,6 +44,8 @@ export const envUtil = {
     getProviderType(provider: Provider): ProviderType | undefined {
         if (provider.constructor.name === 'EthereumProvider') {
             return ProviderType.Mist;
+        } else if (provider.constructor.name === 'TrustWeb3Provider') {
+            return ProviderType.TrustWallet;
```
hewigovens: B.T.W, you can also check if `isTrust` is `true`

packages/instant/src/components/buy_button.tsx
```diff
@@ -82,13 +83,17 @@ export class BuyButton extends React.Component<BuyButtonProps> {
...
+                    return;
+                } else if (e.message === AssetBuyerError.SignatureRequestDenied) {
                     analytics.trackBuySignatureDenied(buyQuote);
                     this.props.onSignatureDenied(buyQuote);
                     return;
-                } else if (e.message === AssetBuyerError.TransactionValueTooLow) {
-                    analytics.trackBuySimulationFailed(buyQuote);
-                    this.props.onValidationFail(buyQuote, AssetBuyerError.TransactionValueTooLow);
+                } else {
+                    errorReporter.report(e);
```
steveklebanoff: I think we should still report something to analytics here

PR #1435, opened 2018-12-14
Move json_schemas to its own package

feuGeneA:
## Description

Moved `json_schemas` code out of the `order_utils` package and into its own package.

## Testing instructions

Watch CI.
Review package page at https://test.pypi.org/project/0x-json-schemas-staging/
Review docs at http://0x-json-schemas-py-staging.s3-website-us-east-1.amazonaws.com/
(Ignore version numbers and `-staging` suffix.)

## Types of changes

## Checklist:

### DIFF
--- a/python-packages/json_schemas/src/zero_ex/json_schemas/__init__.py
+++ b/python-packages/json_schemas/src/zero_ex/json_schemas/__init__.py
@@ -59,3 +59,25 @@ def assert_valid(data: Mapping, schema_id: str) -> None:
 
     _, schema = _LOCAL_RESOLVER.resolve(schema_id)
     jsonschema.validate(data, schema, resolver=_LOCAL_RESOLVER)
+
+
+def assert_valid_json(data: str, schema_id: str) -> None:
+    """Validate the given `data` against the specified `schema`.
+
+    :param data: JSON string to be validated.
+    :param schema_id: id property of the JSON schema to validate against.  Must
+        be one of those listed in `the 0x JSON schema files
+        <https://github.com/0xProject/0x-monorepo/tree/development/packages/json-schemas/schemas>`_.
+
+    Raises an exception if validation fails.
+
+    >>> assert_valid_json(
+    ...     r'''{
+    ...         "v": 27,
+    ...         "r": "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
+    ...         "s": "0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"
+    ...     }''',
+    ...     '/ecSignatureSchema',
+    ... )
+    """  # noqa: E501 (line too long)
+    assert_valid(json.loads(data), schema_id)


.circleci/config.yml
```diff
@@ -200,16 +200,26 @@ jobs:
...
             - restore_cache:
                   key: deps9-{{ .Branch }}-{{ .Environment.CIRCLE_SHA1 }}
+            - run:
+                  command: |
+                      cd python-packages/json_schemas
+                      python -m ensurepip
+                      python -m pip install .[dev]
+                      # HACK! installing the package should do the following
+                      # copy for us, but it's not working in CircleCI for some
+                      # reason.  Support ticket raised (#43979).
```
fabioberger: Can you give us a full URL to the ticket?

python-packages/json_schemas/src/zero_ex/json_schemas/__init__.py
```diff
@@ -6,26 +6,33 @@
...
+    def _ref_to_file(ref: str) -> str:
+        """Translate a JSON schema ref to its corresponding file name.
+
+        >>> _LocalRefResolver._ref_to_file("/addressSchema")
+        'address_schema.json'
+        """
+        _ref = ref.lstrip("/")
+
+        # handle weird special cases
+        _ref = _ref.replace("ECSignature", "EcSignature")
```
fabioberger: Is this an inconsistency in our schema ref names? We should fix this in the JS package, rather then special case this here.

python-packages/sra_client/README.md
```diff
@@ -6,6 +6,35 @@ A Python client for interacting with servers conforming to [the Standard Relayer
...
+    'makerFee': "0",
+    'takerFee': "0",
+    'expirationTimeSeconds': "12345",
+    'salt': "12345",
+    'makerAssetData': "0x0000000000000000000000000000000000000000",
+    'takerAssetData': "0x0000000000000000000000000000000000000000",
+    'exchangeAddress': "0x0000000000000000000000000000000000000000",
+}
+
+assert_valid(order, "/orderSchema")
```
fabioberger: Why not just pass in a JSON string?

python-packages/json_schemas/src/zero_ex/json_schemas/__init__.py
feuGeneA: I was thinking that this wasn't an inconsistency, but rather a truly special case, analogous to how the ERC20.sol's generated TypeScript wrappers would get turned into e_r_c_20.ts without intervention.

But looking further now I see that we have a convention of not fully uppercasing acronyms in schema ref ids (eg `/eip712TypedDataSchema` and `/ecSignatureParameterSchema`).  So yes, I'd say it makes sense to change ref id `/ECSignature` to `/ecSignature`.  Does that make sense to you?

But also, now that you mention it, there is also a real inconsistency here that we could address.  It's the second special case in this code block.  It's that most schema refs (37/44) end with the word "Schema", but some (8/44*) don't:

```
schemas$ grep -He 'id":' * | grep -v Schema
asset_pairs_request_opts_schema.json:    "id": "/AssetPairsRequestOpts",
ec_signature_schema.json:    "id": "/ECSignature",
js_number.json:    "id": "/jsNumber",
orderbook_request_schema.json:    "id": "/OrderBookRequest",
order_config_request_schema.json:    "id": "/OrderConfigRequest",
orders_request_opts_schema.json:    "id": "/OrdersRequestOpts",
paged_request_opts_schema.json:    "id": "/PagedRequestOpts",
request_opts_schema.json:    "id": "/RequestOpts",
schemas$ grep -He 'id":' * | grep -v Schema | wc -l
8
```

Also, I see that while most files (43/44) end with `_schema.json`, there's one (`js_number.json`) that doesn't!  That would have broken my code here!

If we fix all of these things in `@0x/json-schemas` then this method will collapse to just `return f"{snakecase(ref.lstrip("/"))}_schema.json"`!  I'll do it.

python-packages/sra_client/README.md
feuGeneA: The reason is that I wrote the validator to take a python dict because I figured clients would be building up an order programmatically.

Would it make sense to add another assert function for validating an actual JSON string?

.circleci/config.yml
fabioberger: I see, can you add the fact that it's a Zendesk ticket? I was going to start search on Github.

python-packages/json_schemas/src/zero_ex/json_schemas/__init__.py
fabioberger: Yep, awesome, let's make both of those changes to @0x/json-schemas.

python-packages/sra_client/README.md
fabioberger: Yeah, I think we should. Some might be receiving orders via SRA, and in that case, it's cumbersome to first convert it to a dict.

PR #1438, opened 2018-12-15
Fix bug in pull_missing_blocks with incorrect start block

albrow:
## Description

Since we now need to pull missing blocks for other types of events, not just exchange events, we need to remove this lower bound on the block number.

## Checklist:

### DIFF
--- a/packages/pipeline/src/scripts/pull_missing_blocks.ts
+++ b/packages/pipeline/src/scripts/pull_missing_blocks.ts
@@ -37,22 +37,19 @@ interface MissingBlocksResponse {
 
 async function getAllMissingBlocksAsync(web3Source: Web3Source): Promise<void> {
     const blocksRepository = connection.getRepository(Block);
-    let fromBlock = 0;
     while (true) {
-        const blockNumbers = await getMissingBlockNumbersAsync(fromBlock);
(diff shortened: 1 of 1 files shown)


packages/pipeline/src/scripts/pull_missing_blocks.ts
```diff
@@ -37,7 +37,7 @@ interface MissingBlocksResponse {
 
 async function getAllMissingBlocksAsync(web3Source: Web3Source): Promise<void> {
     const blocksRepository = connection.getRepository(Block);
-    let fromBlock = EXCHANGE_START_BLOCK;
+    let fromBlock = 0;
```
wakkadojo: Do you need fromBlock as a parameter to the query? It looks like even without it, and even without sorting the blocks, you'd always pull MAX_BLOCKS_PER_QUERY new, distinct blocks. I'm sure sorting the blocks gives you a performance boost, but maybe you can just get rid of the fromBlock parameter entirely?

albrow: 👌 You're right. The `fromBlock` parameter is an artifact of an older implementation and we don't need it anymore. I removed it.

PR #1439, opened 2018-12-17
Add Docker image and Snapshot commands for migrations

dekz:
## Description

Added a Docker image for our migrations. Having this as a docker image is helpful for those who do not want to install an entire nodejs stack to run ganache. 

Ganache runs in memory but writes a leveldb to directory `0x_ganache_snapshot`. 

TODO 
- [x] Decide on best way for publishing docker image
- [ ] Hook into the post publish flow to build and publish the snapshot 
- [x] Update README about building the snapshot and docker image

## Checklist:

### DIFF
--- a/packages/migrations/Dockerfile
+++ b/packages/migrations/Dockerfile
@@ -3,11 +3,13 @@ FROM mhart/alpine-node:10
 WORKDIR /usr/src/app
 
 RUN npm install -g ganache-cli@6.1.6
-COPY 0x_ganache_snapshot ./0x_ganache_snapshot
 
 ENV MNEMONIC "concert load couple harbor equip island argue ramp clarify fence smart topic"
 ENV NETWORK_ID 50
-
+ENV VERSION "2.2.2"
+ENV SNAPSHOT_HOST "http://ganache-snapshots.0x.org.s3-website.us-east-2.amazonaws.com"
+ENV SNAPSHOT_NAME "0x_ganache_snapshot"
 EXPOSE 8545
-CMD [ "sh", "-c", "ganache-cli --gasLimit 10000000 --db 0x_ganache_snapshot --noVMErrorsOnRPCResponse -p 8545 --networkId \"$NETWORK_ID\" -m \"$MNEMONIC\""]
+
+CMD [ "sh", "-c", "wget $SNAPSHOT_HOST/$SNAPSHOT_NAME-$VERSION.zip -O snapshot.zip && unzip snapshot.zip && ganache-cli --gasLimit 10000000 --db $SNAPSHOT_NAME --noVMErrorsOnRPCResponse -p 8545 --networkId \"$NETWORK_ID\" -m \"$MNEMONIC\" -h 0.0.0.0"]
 
--- a/packages/migrations/Dockerfile
+++ b/packages/migrations/Dockerfile
@@ -1,0 +1,13 @@
+FROM mhart/alpine-node:10
+
+WORKDIR /usr/src/app
+
+RUN npm install -g ganache-cli@6.1.6
+COPY 0x_ganache_snapshot ./0x_ganache_snapshot
+
+ENV MNEMONIC "concert load couple harbor equip island argue ramp clarify fence smart topic"
+ENV NETWORK_ID 50
+
+EXPOSE 8545
+CMD [ "sh", "-c", "ganache-cli --gasLimit 10000000 --db 0x_ganache_snapshot --noVMErrorsOnRPCResponse -p 8545 --networkId \"$NETWORK_ID\" -m \"$MNEMONIC\""]
+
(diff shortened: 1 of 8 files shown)


packages/migrations/package.json
```diff
@@ -10,13 +10,22 @@
...
         "lint": "tslint --format stylish --project .",
         "migrate:v2": "run-s build script:migrate:v2",
+        "migrate:v2:snapshot": "run-s build script:migrate:v2:snapshot",
         "script:migrate:v2": "node ./lib/migrate.js",
-        "docs:json": "typedoc --excludePrivate --excludeExternals --target ES5 --tsconfig typedoc-tsconfig.json --json $JSON_FILE_PATH $PROJECT_FILES"
+        "script:migrate:v2:snapshot": "node ./lib/migrate_snapshot.js",
+        "docs:json": "typedoc --excludePrivate --excludeExternals --target ES5 --tsconfig typedoc-tsconfig.json --json $JSON_FILE_PATH $PROJECT_FILES",
+        "build:snapshot": "rm -rf ${npm_package_config_snapshot_name} && yarn migrate:v2:snapshot && zip -r \"$(git rev-parse HEAD).zip\" ${npm_package_config_snapshot_name}",
+        "build:snapshot:docker": "docker build --tag ${npm_package_config_docker_snapshot_name}:${npm_package_version} --tag ${npm_package_config_docker_snapshot_name}:latest .",
+        "publish:snapshot": "aws s3 cp $(git rev-parse HEAD).zip ${npm_package_config_s3_snapshot_bucket}",
```
dekz: Perhaps instead of the SHA we can use the migrations package version so the URL can be guessed. e.g `s3://testrpc-snapshots/0x_ganache_snapshot-2.2.3.zip`.

fabioberger: I'd be down for that.

packages/migrations/package.json
```diff
@@ -10,13 +10,22 @@
     "scripts": {
         "build": "tsc -b",
         "build:ci": "yarn build",
-        "clean": "shx rm -rf lib",
+        "clean": "shx rm -rf lib ${npm_package_config_snapshot_name} *.zip",
```
fabioberger: Can we make this `*.zip` a bit more specific? Worried a future dev might introduce a .zip and be very confused. Is there some suffix we could add to all the snapshots?

packages/migrations/src/migrate_snapshot.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Provider } from 'ethereum-types';
+
+import { runMigrationsAsync } from './migration';
+
+(async () => {
+    let providerConfigs;
+    let provider: Provider;
+    let txDefaults;
+
+    providerConfigs = { shouldUseInProcessGanache: true, ganacheDatabasePath: '0x_ganache_snapshot' };
```
fabioberger: Can we reference `0x_ganache_snapshot` from the `package.json` config? Dedup.

packages/migrations/src/migrate_snapshot.ts
```diff
@@ -0,0 +1,23 @@
...
+    provider = web3Factory.getRpcProvider(providerConfigs);
+    txDefaults = {
+        from: devConstants.TESTRPC_FIRST_ADDRESS,
+    };
+    await runMigrationsAsync(provider, txDefaults);
+    process.exit(0);
+})().catch(err => {
+    logUtils.log(err);
+    process.exit(1);
+});
```
fabioberger: Can we please at instructions to the README about how to generate a snapshot and upload it to S3?

packages/migrations/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN npm install -g ganache-cli@6.1.6
+
+ENV MNEMONIC "concert load couple harbor equip island argue ramp clarify fence smart topic"
+ENV NETWORK_ID 50
+ENV VERSION "latest"
+ENV SNAPSHOT_HOST "http://ganache-snapshots.0x.org.s3-website.us-east-2.amazonaws.com"
+ENV SNAPSHOT_NAME "0x_ganache_snapshot"
+EXPOSE 8545
+
+CMD [ "sh", "-c", "wget $SNAPSHOT_HOST/$SNAPSHOT_NAME-$VERSION.zip -O snapshot.zip && unzip snapshot.zip && ganache-cli --gasLimit 10000000 --db $SNAPSHOT_NAME --noVMErrorsOnRPCResponse -p 8545 --networkId \"$NETWORK_ID\" -m \"$MNEMONIC\" -h 0.0.0.0"]
```
LogvinovLeon: Is it possible to make it multiline for improved readability?

dekz: Dockerfile supports multiline but these arguments need to be supplied as a single argument to `sh`. So I'm not sure of the best way to break it up. We could introduce a shell script and run that but it disconnects it from the Dockerfile.

PR #1443, opened 2018-12-17
[0x.org] fix portal

BMillman19:
## Description

- @fragosti can you take a look at this? Something about how keyframes are supported in styled components v4 is different than in v3, looks like b&b upgraded to v4 as part of the move to 0x.org
- The fix I applied got rid of the run-time error and makes portal usable, from what i can tell the onboarding cards which were using the `Animation` component are not affected

## Checklist:


steveklebanoff:
'keyframes' section of "Breaking Changes" on https://github.com/styled-components/styled-components/releases/tag/v4.0.0 has some insight ,but I'm not 100% confident I understand what's going on here, and this also has me worried about additional warnings we are ignoring for now

PR #1447, opened 2018-12-17
[website] Fix 0x org jank links

fragosti:
## Description

Fixes some of the weird link behaviors and implements the correct links in the header and the footer.

## Checklist:


packages/react-shared/src/components/link.tsx
```diff
@@ -18,11 +18,15 @@ interface BaseLinkProps {
     fontColor?: string;
 }
 
-interface ScrollLinkProps extends BaseLinkProps {
+export interface ScrollLinkProps extends BaseLinkProps {
     onActivityChanged?: (isActive: boolean) => void;
 }
 
-type LinkProps = BaseLinkProps & ScrollLinkProps;
+export interface ReactLinkProps extends BaseLinkProps {
```
steveklebanoff: Do we need to update the changelog for this change?

PR #1450, opened 2018-12-18
[website] 0x org about

fragosti:
## Description

All /about tasks.

## Checklist:


packages/website/ts/@next/pages/about/press.tsx
```diff
@@ -51,7 +56,7 @@ export const NextAboutPress = () => (
         description={
             <>
                 <Paragraph size="medium" marginBottom="60px">
-                    Want to write about 0x? <a href="#">Get in touch</a>, or <a href="#">download our press kit</a>.
+                    Want to write about 0x? <a href="mailto:team@0xproject.com">Get in touch.</a>
```
steveklebanoff: seems like this should be in a constant but no big deal

PR #1455, opened 2018-12-18
Add validation to MAP

abandeali1:
## Checklist:


contracts/protocol/contracts/protocol/AssetProxy/MultiAssetProxy.sol
```diff
@@ -182,7 +182,10 @@ contract MultiAssetProxy is
                     let totalAmount := mul(amountsElement, amount)
 
                     // Revert if multiplication resulted in an overflow
-                    if iszero(eq(div(totalAmount, amount), amountsElement)) {
+                    if and(
+                        gt(amount, 0),
+                        iszero(eq(div(totalAmount, amount), amountsElement))
+                    ) {
```
Recmo: We can save a bit of gas:
```
iszero(or(
    iszero(amount),
    eq(div(totalAmount, amount), amountsElement)
))
```

BTW, in the original we could have done `sub(a, b)` instead of `iszero(eq(a, b))`. Does not work here because `and` is bitwise and the result of `sub` is noisy. To fix that we could change `and` to `mul`. Since the `gt` result is always `0` or `1` this works.

PR #1461, opened 2018-12-19
Sol-compiler watch mode

LogvinovLeon:
## Description

TL;DR. Run `yarn watch` in any of the contracts packages and the contracts will be watched and recompiled on change.

Sol-compiler now has a `-w` flag that runs it in watch mode. In that mode it uses chokidar to watch all the files that were read during the compilation and redoes the compilation if any of those got changed. Including dependencies in node_modules. Also sol-resolver now returns an absolute file path in addition to a short one. That is used in watch mode. In order to detect which files got touched we use a SpyResolver that is a proxy recording resolve calls. It also changed the soloring of errors and warnings to be less painful for the eyes and consistent with tsc. It's not just one word being colored. Warning or error. The whole message is gray.

There is quite a bit of code in this PR that was just moved from the Compiler class to the compiler utils because it used just the resolver from that class. Now those functions are parametrized with the resolver, so there is no need to review those. Nothing changed. The only change is the watchAsync function and the function to get touched paths.

## Testing instructions

Try watch mode!

## Types of changes

## Checklist:


packages/sol-compiler/src/compiler.ts
```diff
@@ -173,6 +133,43 @@ export class Compiler {
         const promisedOutputs = this._compileContractsAsync(this._getContractNamesToCompile(), false);
         return promisedOutputs;
     }
+    public async watchAsync(): Promise<void> {
+        console.clear(); // tslint:disable-line:no-console
+        utils.logWithTime('Starting compilation in watch mode...');
+        const watcher = chokidar.watch('^$', { ignored: /(^|[\/\\])\../ });
```
fabioberger: Please set `'^$'` to a descriptive variable name. It's not immediately clear to the reader what this means. Same goes for the regexp.

packages/sol-compiler/src/compiler.ts
```diff
@@ -173,6 +133,43 @@ export class Compiler {
...
+        const onFileChangedAsync = async () => {
+            watcher.unwatch('*'); // Stop watching
+            try {
+                await this.compileAsync();
+                utils.logWithTime('Found 0 errors. Watching for file changes.');
+            } catch (err) {
+                if (err.typeName === 'CompilationError') {
+                    utils.logWithTime(`Found ${err.errorsCount} ${pluralize('error', err.errorsCount)}. Watching for file changes.`);
+                } else {
+                    utils.logWithTime('Found errors. Watching for file changes.');
```
fabioberger: We should still print out the errors in this case.

packages/sol-compiler/src/compiler.ts
```diff
@@ -173,6 +133,43 @@ export class Compiler {
...
+            utils.logWithTime('File change detected. Starting incremental compilation...');
+            onFileChangedAsync();
+        });
+    }
+    private _getPathsToWatch(): string[] {
+        const contractNames = this._getContractNamesToCompile();
+        const spyResolver = new SpyResolver(this._resolver);
+        for (const contractName of contractNames) {
+            const contractSource = spyResolver.resolve(contractName);
+            getSourceTreeHash(spyResolver, contractSource.path);
```
fabioberger: We get the sourceTreeHash but then do nothing with it? This warrants a comment if it's because the `spyResolver` is storing something during this operation.

packages/sol-compiler/src/compiler.ts
```diff
@@ -301,7 +300,11 @@ export class Compiler {
         // contains listings for for every contract compiled during the compiler invocation that compiled the contract
```
fabioberger:
```suggestion
        // contains listings for every contract compiled during the compiler invocation that compiled the contract
```

packages/sol-compiler/src/utils/utils.ts
```diff
@@ -1,6 +1,12 @@
...
+import chalk from 'chalk';
+
 export const utils = {
     stringifyWithFormatting(obj: any): string {
         const stringifiedObj = JSON.stringify(obj, null, '\t');
         return stringifiedObj;
     },
+    logWithTime(arg: string): void {
+        logUtils.log(`[${chalk.gray(new Date().toLocaleTimeString())}] ${arg}`);
+    },
```
fabioberger: Let's move this to `logUtils`, think others would use it too.

packages/sol-resolver/src/resolvers/spy_resolver.ts
```diff
@@ -0,0 +1,21 @@
+import * as _ from 'lodash';
+
+import { ContractSource } from '../types';
+
+import { Resolver } from './resolver';
+
+export class SpyResolver extends Resolver {
```
fabioberger: Let's a comment explaining what this resolver does.

packages/sol-compiler/src/compiler.ts
LogvinovLeon: Yes. We do that

packages/sol-compiler/src/utils/compiler.ts
```diff
@@ -106,3 +114,208 @@ export function parseDependencies(contractSource: ContractSource): string[] {
...
+        return;
+    }
+    for (const importStatementMatch of importStatementMatches) {
+        const importPathMatches = importStatementMatch.match(/\"([^\"]*)\"/);
+        if (importPathMatches === null || importPathMatches.length === 0) {
+            continue;
+        }
+
+        let importPath = importPathMatches[1];
+        // HACK(ablrow): We have, e.g.:
```
fabioberger:
```suggestion
        // HACK(albrow): We have, e.g.:
```

packages/sol-compiler/src/compiler.ts
```diff
@@ -173,6 +132,53 @@ export class Compiler {
...
+        });
+    }
+    private _getPathsToWatch(): string[] {
+        const contractNames = this._getContractNamesToCompile();
+        const spyResolver = new SpyResolver(this._resolver);
+        for (const contractName of contractNames) {
+            const contractSource = spyResolver.resolve(contractName);
+            // NOTE: We ignore the return value here. We don't want to compute the source tree hash.
+            // We just want to call a SpyResolver on each contracts and it's dependencies and
+            // this is a convinient way to reuse the existing code that does that.
```
fabioberger:
```suggestion
            // this is a convenient way to reuse the existing code that does that.
            // We can then get all the relevant paths from the `spyResolver` below.
```

PR #1466, opened 2018-12-20
[website] Fix TS errors

steveklebanoff:
## Description

* Fixes TS errors
* Makes it so we can't compile w/ TS errors
* Removed some unused files instead of fixing the TS errors

## Checklist:


packages/website/ts/@next/components/button.tsx
```diff
@@ -37,7 +37,7 @@ export const Button = (props: ButtonInterface) => {
         linkElem = ReactRouterLink;
     }
 
-    const Component = linkElem ? ButtonBase.withComponent(linkElem) : ButtonBase;
+    const Component = linkElem ? ButtonBase.withComponent<any>(linkElem) : ButtonBase;
```
steveklebanoff: I tried for a bit to do this without `any`, but couldn't find a good way.  `withComponent` seems to expect a functional react component, but `ReactRouterLink` is typed as a class based react component

PR #1472, opened 2018-12-22
fetch and save copper

xianny:
## Description

Implements fetching, parsing, and saving Copper entities. Does not implement rate limiting yet. Pretty large PR but a lot of it is boilerplate.  This one has the clean commit history.

## Testing instructions

There are entity and parser tests added to `yarn test:all`. You can also run the script: lib/src/scripts/pull_copper.js

Env variables required:
COPPER_ACCESS_TOKEN
COPPER_USER_EMAIL

## Types of changes

## Checklist:


packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+        entity.dateCreated = lead.date_created * ONE_SECOND;
+        entity.dateModified = lead.date_modified * ONE_SECOND;
+
+        entity.email = lead.email === null ? 0 : 1;
+
+        // nullable fields
+        entity.socials = 0;
+        entity.websites = 0;
+        entity.phoneNumbers = 0;
+        try {
```
albrow: It looks like this would result in all of the `socials`, `websites`, and `phoneNumbers` fields being `0` if just one of them is `undefined` in the raw response from the API. That doesn't seem right.

But I have another question that might override this. Is it really all that useful to store *just* the number of `socials`, `websites`, and `phoneNumbers`? It seems like we should either be storing the values themselves (e.g., save each phone number) or nothing.

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+        entity.typeId = activity.type.id;
+        entity.typeCategory = activity.type.category.toString();
+        entity.typeName = activity.type.name;
+
+        entity.userId = activity.user_id;
+        entity.dateCreated = activity.date_created * ONE_SECOND;
+        entity.dateModified = activity.date_modified * ONE_SECOND;
+
+        // nullable fields
+        try {
```
albrow: Similarly, I don't think this is the right way to handle nullable fields. It looks like if any of the fields are `undefined` it will result in all of them being `null` in the database.

Can we use [`R.path`](https://ramdajs.com/docs/#path) or [`R.pathOr`](https://ramdajs.com/docs/#pathOr) instead?

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
+import * as R from 'ramda';
+
+import { CopperActivity, CopperActivityType, CopperCustomField, CopperLead, CopperOpportunity } from '../../entities';
+
+const ONE_SECOND = 1000;
+export type CopperSearchResponse = CopperLeadResponse | CopperActivityResponse | CopperOpportunityResponse;
```
albrow: Response types probably belong in __src/data_sources__ instead of here.

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+        entity.primaryContactId = opp.primary_contact_id;
+        entity.priority = opp.priority;
+        entity.status = opp.status;
+        entity.tags = opp.tags;
+        entity.interactionCount = opp.interaction_count;
+        entity.monetaryValue = opp.monetary_value;
+        entity.winProbability = opp.win_probability;
+        entity.dateCreated = opp.date_created * ONE_SECOND;
+        entity.dateModified = opp.date_modified * ONE_SECOND;
+        entity.customFields = opp.custom_fields.map(f => f.custom_field_definition_id);
```
albrow: It looks like this is a one-to-many relationship where an opportunity can have many custom fields? If that's the case, storing custom field ids in a `jsonb` type like this is going to make it harder to do joins. Typically the way that I've set up one-to-many relationships in the past (e.g. where one "parent" has many "children") is to add an `parent_id` field to the child.

Have you considered adding an `opportunity_id` field to the `custom_fields` table instead? It would make joins faster/simpler.

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+/**
+ * Parse response from Copper API /activity_types/
+ *
+ * @param activityTypeResponse - Activity Types response from the API, keyed by "user" or "system"
+ * @returns Returns an array of Copper Activity Type entities
+ */
+export function parseActivityTypes(
+    activityTypeResponse: Map<CopperActivityTypeCategory, CopperActivityTypeResponse[]>,
+): CopperActivityType[] {
+    const values: CopperActivityTypeResponse[] = R.reduce(
```
albrow: It doesn't look like `reduce` is necessary here. I think you can just write:

```ts
const values = Array.from(activityTypeResponse.values());
```

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+        (acc, val) => acc.concat(val),
+        [] as CopperActivityTypeResponse[],
+        Object.values(activityTypeResponse),
+    );
+    return values.map(activityType => {
+        const parsed: CopperActivityType = {
```
albrow: nit: I think we already have enough context here to understand what's going on without declaring the temporary variable called `parsed`. (The function is called `parseActivityTypes`).

Can we do this instead?

```ts
values.map(activityType => ({
    id: activityType.id,
    name: activityType.name,
    category: activityType.category.toString(),
    isDisabled: activityType.is_disabled,
    countAsInteraction: activityType.count_as_interaction,
}));
```

packages/pipeline/src/parsers/copper/index.ts
```diff
@@ -0,0 +1,274 @@
...
+ */
+export function parseCustomFields(customFieldResponse: CopperCustomFieldResponse[]): CopperCustomField[] {
+    function parseTopLevelField(field: CopperCustomFieldResponse): CopperCustomField[] {
+        const parsed: CopperCustomField = {
+            id: field.id,
+            name: field.name,
+            dataType: field.data_type.toString(),
+        };
+
+        if (field.options !== undefined) {
```
albrow: I would appreciate a comment about what's going on here. It seems pretty weird to pull some properties out of `field` and then concatenate them with in a list which basically already contains `field` itself. Could you clarify?

packages/pipeline/src/scripts/pull_copper.ts
```diff
@@ -0,0 +1,129 @@
...
+async function getMaxAsync(conn: Connection, sortColumn: string, tableName: string): Promise<number> {
+    const queryResult = await conn.query(`SELECT MAX(${sortColumn}) as _max from ${tableName};`);
+    if (R.isEmpty(queryResult)) {
+        return 0;
+    } else {
+        return queryResult[0]._max;
+    }
+}
+
+// (Xianny): Copper API doesn't allow queries to filter by date. To ensure that we are filling in ascending chronological
```
albrow: Thank you for this comment. That's pretty annoying 🙁

packages/pipeline/src/utils/transformers/number_to_bigint.ts
```diff
@@ -9,8 +9,12 @@ const decimalRadix = 10;
 // https://github.com/typeorm/typeorm/issues/2400 for more information.
 export class NumberToBigIntTransformer implements ValueTransformer {
     // tslint:disable-next-line:prefer-function-over-method
-    public to(value: number): string {
-        return value.toString();
+    public to(value: number): string | null {
+        if (typeof value !== 'number') {
```
albrow: If `value` is not a `number` something is spectacularly wrong. IMO we should throw an error instead of just returning `null`.

If `value` is sometimes `null` or `undefined` we should add a more specific check for that.

packages/pipeline/test/entities/copper_test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+chaiSetup.configure();
+
+// tslint:disable:custom-no-magic-numbers
+describe('Copper entities', () => {
+    describe('save and find', async () => {
+        it('Copper lead', async () => {
+            const connection = await createDbConnectionOnceAsync();
+            const repository = connection.getRepository(CopperLead);
+            const entity = (leadsData[0] as unknown) as CopperLead;
```
albrow: It seems like all of the fixture data is an array but we only ever use the first element. We should either use all the elements in the array or just change the fixture data to be an object.

packages/pipeline/test/entities/util.ts
```diff
@@ -15,9 +15,7 @@ const expect = chai.expect;
  * @param entity An instance of a TypeORM entity which will be saved/retrieved from the database.
  */
 export async function testSaveAndFindEntityAsync<T>(repository: Repository<T>, entity: T): Promise<void> {
-    // Note(albrow): We are forced to use an 'as any' hack here because
-    // TypeScript complains about stack depth when checking the types.
-    await repository.save(entity as any);
+    await repository.save<any>(entity);
```
albrow: This is okay but we should still probably include the comment explaining why it is necessary.

packages/pipeline/test/fixtures/copper/api_v1_activity_types.parsed.json
```diff
@@ -0,0 +1,14 @@
+[
```
albrow: I understand the raw response data being JSON, but I think the parsed data should be TypeScript. This would give us the benefit of the compiler helping us out if a field is ever renamed or something in the data structure changes. As it currently stands, we are bypassing the type checker by starting with JSON and then performing a blind type assertion (e.g., `(opportunitiesData[0] as unknown) as CopperOpportunity`). This seems like it could lead to unexpected bugs down the road.

What do you think?

packages/pipeline/test/parsers/copper/index_test.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as customFieldsApiResponse from '../../fixtures/copper/api_v1_custom_field_definitions.json';
+import * as customFieldsParsed from '../../fixtures/copper/api_v1_custom_field_definitions.parsed.json';
+import * as listActivitiesApiResponse from '../../fixtures/copper/api_v1_list_activities.json';
+import * as listActivitiesParsed from '../../fixtures/copper/api_v1_list_activities.parsed.json';
+import * as listLeadsApiResponse from '../../fixtures/copper/api_v1_list_leads.json';
+import * as listLeadsParsed from '../../fixtures/copper/api_v1_list_leads.parsed.json';
+import * as listOpportunitiesApiResponse from '../../fixtures/copper/api_v1_list_opportunities.json';
+import * as listOpportunitiesParsed from '../../fixtures/copper/api_v1_list_opportunities.parsed.json';
+
+interface TestCase {
```
albrow: I really like how these tests are structured. Makes it very easy to read and update later on 😄

packages/pipeline/test/parsers/copper/index_test.ts
```diff
@@ -0,0 +1,84 @@
...
+        input: customFieldsApiResponse,
+        expected: customFieldsParsed,
+        parseFn: parseCustomFields,
+    },
+];
+describe('Copper parser', () => {
+    it('parses API responses', () => {
+        testCases.forEach(testCase => {
+            const actual: CopperEntity[] = testCase.parseFn(testCase.input);
+            expect(JSON.stringify(actual)).equal(JSON.stringify(testCase.expected));
```
albrow: `JSON.stringify` is [not guaranteed to  preserve the order of keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify). Is there some other way to do this assertion?

packages/pipeline/src/data_sources/copper/index.ts
```diff
@@ -0,0 +1,139 @@
...
+    page_number?: number;
+}
+export enum CopperEndpoint {
+    Leads = 'leads',
+    Opportunities = 'opportunities',
+    Activities = 'activities',
+}
+const ONE_SECOND = 1000;
+function getUriForEndpoint(endpoint: CopperEndpoint): string {
+    return ((_endpoint: CopperEndpoint) => {
```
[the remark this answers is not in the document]
xianny: Nope, this needs to be simplified! when I wrote this I was still figuring out how enums in typescript worked. Getting rid of it.

packages/pipeline/src/parsers/copper/index.ts
xianny: > It looks like this would result in all of the socials, websites, and phoneNumbers fields being 0 if just one of them is undefined in the raw response from the API. That doesn't seem right.

Yeah, you are right. Thanks for catching that!

> But I have another question that might override this. Is it really all that useful to store just the number of socials, websites, and phoneNumbers? It seems like we should either be storing the values themselves (e.g., save each phone number) or nothing.

I don't think we should store these details because we would then be subject to GDPR requirements or potential future data protection regulation, and generally it's a privacy and data management concern. If we need this info for business purposes it lives within Copper - we shouldn't need it for analytics. 

I'm cool with not storing it if the numbers aren't of interest to us.

xianny: Checked with Clay, he says numbers aren't useful.

packages/pipeline/src/parsers/copper/index.ts
xianny: I was thinking about this too. I originally put it in /data_sources, but there was a lot of referencing between files when writing the parsing logic. I think it makes development easier when the parser contains all of the logic/context of what a response looks like and how to transform it into an entity. Does that make sense?

packages/pipeline/src/parsers/copper/index.ts
xianny: I had some uncertainty with this one too. It's actually a many-to-many relationship. We could model it with a junction table, but I'm worried about the sustainability of trying to recreate 
(and maintain!) Copper domain models in our code. Right now it's just an array, which I agree makes querying clunky. 

As a real example, we could be interested in opportunities by integration type (a top level category for custom fields). Two possible solutions:

1. We could change it to a key/value store with `top_level_category -> actual_value`, which means we can do queries like this:
```
SELECT * FROM opportunities GROUP BY custom_fields #>> '{123}' // where 123 is the id for "Integration Type"
```

or,
2. if we wanted to do joins properly, we'd need to add two more tables:
- split the custom field table into one each for the top-level category and the values
- add a junction table with `opportunity_id | custom_field_id | custom_field_parent_id`

This would allow us to do queries like:
```
SELECT COUNT(*) from opportunities as o JOIN custom_fields_relationships as cf 
WHERE o.id = cf.opportunity_id
AND cf.custom_field_parent_id = 123 // the id for the top-level category we are interested in, e.g. "integration type"
GROUP BY cf.custom_field_id
```

My preference is for the first option as it seems simpler and more resilient to Copper domain changes.  What do you think?

packages/pipeline/test/entities/copper_test.ts
xianny: Yep, we only use the first element. It's an array because the fixtures data is also used in the parsers test. I deliberately coupled the fixtures for both set of tests so that the tests also fail if the parser output stops being congruent with the schema. Does that make sense?

packages/pipeline/test/parsers/copper/index_test.ts
xianny: I didn't know that! Thanks for the heads up. I think changing the parsed data to typescript will resolve this too.

packages/pipeline/test/fixtures/copper/api_v1_activity_types.parsed.json
xianny: I think that's a good idea. I tried it and already found some problems with the parser saving `null` instead of `undefined`. 👍

packages/pipeline/src/parsers/copper/index.ts
xianny: thanks, on second thought this *is* confusing. the API response consists of a top level field and (potentially) some nested fields. I added a comment and simplified the code somewhat.

xianny:
Thanks for the detailed review and great suggestions @albrow ! I made a lot of the changes suggested. Left unresolved a few items for discussion. We're getting there!

packages/pipeline/src/parsers/copper/index.ts
xianny: I implemented option #1 in https://github.com/0xProject/0x-monorepo/pull/1472/commits/cfd7600b766e7c13b6974d5f7af02d61c35073c4 because it was relatively straightforward. Take a look and let me know what you think. :)

packages/pipeline/src/parsers/copper/index.ts
albrow: In general, types belong with the data that they describe. The data here is coming from the __data_sources/__ directory. In the current implementation, __data_sources/copper/index.ts__ contains the lines:

```
import {
    CopperActivityTypeCategory,
    CopperActivityTypeResponse,
    CopperCustomFieldResponse,
    CopperSearchResponse,
} from '../../parsers/copper';
```

You're importing types from another file in order to describe data that originates in this file, which seems wrong to me. You wouldn't make a __linked_list.ts__ file implementing a `LinkedList` and then put the types in a different file. I don't really see how this case is any different.

> but there was a lot of referencing between files when writing the parsing logic

I don't see how the current implementation reduces cross-file referencing in any significant way.  Could you elaborate?

packages/pipeline/src/parsers/copper/index.ts
albrow: We talked a bit on Slack. I'm okay with the current solution (option 1).

packages/pipeline/test/entities/copper_test.ts
xianny: I was mainly concerned about it taking too long, but I just switched it and testing all entities takes 13.25s vs 11.85s. Seems alright to me, wdyt?

packages/pipeline/src/parsers/copper/index.ts
xianny: Yeah, I agree that conventionally data types should belong with the data where it originates. I'd argue that in this case, the logical responsibility for a) knowing what the data looks like and b) how to transform it both lie in the parser file. The data_sources file is only responsible for creating well-formed requests and managing pagination and other request logic. It doesn't have to know anything about the shape and types of data it's returning. In addition the practical benefit of having the typedef in the parser file where we are operating on it, is that we can write the parsing logic while having the data types right there in the same file to edit or reference. If we put the typedefs in the data_sources file, then we're cross-referencing /data_sources, /entities, and the /parser file. 

I do see the drawback of losing some breadcrumbs, e.g. someone who's only looking at the parser file can't easily follow the typedefs to the /data_sources file and see where the actual data is coming from. 

There's pros and cons each way -- happy to defer on this one if you think it's still better to put the typedefs in the /data_sources file.

albrow: 🤷‍♂️Not a hill I'm willing to die on so I guess just leave it.

xianny:
Tested on QA [here](https://github.com/0xProject/0x-pipeline-orchestration/pull/2), merging.

PR #1474, opened 2018-12-24
Implement tslint enum-naming to enforce PascalCase on enum members

xianny:
## Description

Implements tslint rule to enforce PascalCase for all enum members. Also implements a testing harness.

See updated CHANGELOGS for list of exported enums affected.

## Testing instructions

run `yarn clean && yarn build && yarn test`
## Types of changes

 

## Checklist:


packages/tslint-config/tslint.json
```diff
@@ -25,6 +25,7 @@
         "curly": true,
         "custom-no-magic-numbers": [true, 0, 1, 2, 3, -1],
         "encoding": true,
+        "enum-snake-case": false,
```
albrow: Did you mean to set this to `true` to enable the linter or are you planning to do that in a separate PR?

packages/tslint-config/rules/enumSnakeCaseRule.ts
```diff
@@ -0,0 +1,43 @@
...
+        let fix = toSnakeCase(key).toUpperCase();
+        if (value !== undefined) {
+            fix += ` = ${value}`;
+        }
+        return new Lint.Replacement(node.getStart(ctx.sourceFile), node.getWidth(ctx.sourceFile), fix);
+    }
+}
+
+function isSnakeCase(s: string): boolean {
+    const regex = /^[A-Z\d]+(_{1}[A-Z\d]+)*$/g;
```
albrow: nit: You could define `regex` outside of the function scope so it can be more efficiently re-used.

packages/tslint-config/rules/enumSnakeCaseRule.ts
```diff
@@ -0,0 +1,43 @@
...
+            return ctx.addFailureAtNode(node, Rule.FAILURE_STRING, getFix(node));
+        }
+        // Continue recursion into the AST by calling function `cb` for every child of the current node.
+        return ts.forEachChild(node, cb);
+    }
+
+    function getFix(node: ts.Node): Lint.Replacement {
+        const [key, value] = node
+            .getText(ctx.sourceFile)
+            .split('=')
```
albrow: Tthis fails on some edge cases where `value` also contains one or more `=`.

For example:

```ts
export enum Operators {
     ASSIGN = '=',
     EQUALS = '==',
     TRIPLE_EQUALS = '===',
}
```

You can use this test to confirm:

```ts
    it('testing fixer with equals sign in value example', () => {
        const src = `enum Operators { Assign = '=', Equals = '==', TripleEquals = '===' }`;
        const expected = `enum Operators { ASSIGN = '=', EQUALS = '==', TRIPLE_EQUALS = '===' }`;
        const actual = getFixedResult(src, rule);
        const result = helper(src, rule);
        assert.equal(result.errorCount, 3); // tslint:disable-line:custom-no-magic-numbers
        assert.equal(actual, expected);
    });
```

The TypeScript parser has likely already sorted this out when constructing the AST. Is there a way for us to take advantage of the existing AST structure instead of trying to do a bit of manual parsing?

packages/tslint-config/tslint.json
xianny: I meant to leave it as false for this commit. I initially had it as true and started renaming everything, then had second thoughts because it seemed like it could be a big breaking case for small practical benefit. Do you think I should just go ahead and rename everything?

LogvinovLeon: I will do that in a separate PR. Maybe package by package.

packages/tslint-config/rules/enumSnakeCaseRule.ts
LogvinovLeon: What's the point of `{1}`. Can't you just skip it?

```suggestion
    const capitalAlphaNum = '[A-Z\d]+';
    const regex = new Regex('^${capitalAlphaNum}(_${capitalAlphaNum})*$');
```

Feel free to ignore. I'm actually not sure if this is more readable. Regexes with template strings look weird.

LogvinovLeon:
Thanks for writing our first linter rule tests. You're the boss. Approved (assuming @albrow feedback is fixed).

packages/tslint-config/tslint.json
fabioberger: I would just do it. We need to bite the bullet and I don't expect there to be a TON on places where we haven't been following this rule, right?

xianny:
Hey @fabioberger @LogvinovLeon , @albrow and I actually discussed on Slack whether we should turn this rule on or not and felt like we should punt on this. We were concerned that renaming a bunch of enums would potentially break backwards compatibility on any external packages using exported members, and it's a pretty big cost just for naming style. Also, there are a few style guides that actually recommend using PascalCase for enums:
screen shot 2019-01-07 at 9 49 47 am

Looking at the linter results, most of our enums are already in PascalCase. Maybe I could rewrite the rule to enforce PascalCase? What was the original reasoning for changing to SCREAMING_SNAKE_CASE?

fabioberger:
Hm, I would like us to enforce some sort of convention here. It'll be less headache for everyone involved long-term. Let's discuss.

xianny:
I've updated to PascalCase as discussed. I think these are the only linter errors but I'm not sure if lerna is outputting lint errors for all files.

screen shot 2019-01-07 at 7 56 27 pm
@LogvinovLeon can you take a look? It's things like `ERC20` vs `Erc20` etc. My opinion is that we should allow acronyms, but it's impossible to distinguish them from ALLCAPS so we either ignore ALLCAPS or manually disable the rule.

LogvinovLeon:
What about listing known acronyms. As a config for the rule.
```json
{
    "no-enum-snake-case": [true, "ZRX", "ERC20", "ERC721"],
}
```
@xianny

packages/instant/src/types.ts
```diff
@@ -95,7 +95,7 @@ export enum Network {
 
 export enum ZeroExInstantError {
     AssetMetaDataNotAvailable = 'ASSET_META_DATA_NOT_AVAILABLE',
-    InsufficientETH = 'INSUFFICIENT_ETH',
+    InsufficientEth = 'INSUFFICIENT_ETH',
```
steveklebanoff: Can we use `// tslint:disable-line:enum-naming` here, as we use it above to keep ETH as `ETH` instead of `Eth` ? 

Related: I like the @LogvinovLeon 's idea of having a config for known/whitelisted acronyms, but understand that it could be a follow-up

PR #1475, opened 2018-12-24
Use Compressed Calldata in Contract Wrappers

hysz:
## Description

The contract templates now use the optimized ABI encoder to compress all calldata.

We wanted to do this without changing the interface of the templates at all, so a few minor changes were made to the encoder:
1. Enforce lowercase addresses
   (this also fixes a bug in current return value decoding, where nested addresses would not always be decoded)
2. The types `uint8` and `int8` are decoded as `number` instead of `BigNumber`
3. Convenience functions to get decoded data as an array
4. Convenience function to create an encoder from a signature.
    Ex, signatures like `'string'` and `'(string,uint256)'` and `'(foo string,bar uint256)'` can now all be used to instantiate an optimized encoder.

In a subsequent PR we'll add the ability for templates to decode calldata.

## Checklist:


contracts/protocol/test/utils/exchange_wrapper.ts
```diff
@@ -275,6 +278,19 @@ export class ExchangeWrapper {
         );
         return data;
     }
+    public abiDecodeFillOrder(data: string): AbiDecodedFillOrderData {
+        // Lookup fillOrder ABI in exchange abi
+        const fillOrderAbi = _.find(this._exchange.abi, (value: AbiDefinition) => {
```
LogvinovLeon: Maybe:
```suggestion
        const fillOrderAbi = _.find(this._exchange.abi, { name: 'fillOrder' }) as MethodAbi;
```

We don't usually use that form but I feel like here it's more readable. It's not type safe, but the current solution is not either.

packages/utils/src/abi_encoder/abstract_data_types/data_type.ts
```diff
@@ -51,8 +51,26 @@ export abstract class DataType {
         return value;
     }
 
+    public decodeAsArray(returndata: string, rules?: DecodingRules): any {
```
LogvinovLeon:
```suggestion
    public decodeAsArray(returndata: string, rules?: DecodingRules): any[] {
```

contracts/protocol/test/utils/types.ts
```diff
@@ -0,0 +1,8 @@
+import { SignedOrder } from '@0x/types';
+import { BigNumber } from '@0x/utils';
+
+export interface AbiDecodedFillOrderData {
+    order: SignedOrder;
+    akerAssetFillAmount: BigNumber;
```
fabioberger:
```suggestion
    makerAssetFillAmount: BigNumber;
```

packages/base-contract/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { abiUtils, BigNumber } from '@0x/utils';
```
fabioberger: Were we able to remove abiUtils and ethers.js entirely from our codebase?

packages/utils/src/abi_encoder/abstract_data_types/types/set.ts
```diff
@@ -134,31 +134,26 @@ export abstract class AbstractSetDataType extends DataType {
...
-            if (!(key in childMap)) {
+        _.forEach(this._memberIndexByName, (memberIndex: number, memberName: string) => {
+            if (!(memberName in obj)) {
                 throw new Error(
-                    `Could not assign tuple to object: unrecognized key '${key}' in object ${this.getDataItem().name}`,
+                    `Could not assign tuple to object: missing key '${memberName}' in object ${JSON.stringify(obj)}`,
                 );
             }
-            const memberBlock = this._members[this._memberIndexByName[key]].generateCalldataBlock(value, block);
+            const memberValue: any = (obj as { [key: string]: any })[memberName];
```
fabioberger:
```suggestion
            const memberValue: any = (obj as ObjectMap<any>)[memberName];
```

packages/utils/src/abi_encoder/evm_data_types/method.ts
```diff
@@ -44,7 +44,19 @@ export class MethodDataType extends AbstractSetDataType {
         return returnValues;
     }
 
-    public getSignature(): string {
+    public decodeReturnValuesAsArray(returndata: string, rules?: DecodingRules): any {
```
fabioberger:
```suggestion
    public decodeReturnValuesAsArray(returnData: string, rules?: DecodingRules): any {
```

Always camel--case variable names

packages/utils/src/abi_encoder/evm_data_types/pointer.ts
```diff
@@ -11,7 +11,11 @@ export class PointerDataType extends AbstractPointerDataType {
         super(dataItem, dataTypeFactory, destDataType, parentDataType);
     }
 
-    public getSignature(): string {
-        return this._destination.getSignature();
+    public getSignatureType(): string {
+        return this._destination.getSignature(false);
+    }
+
+    public getSignature(detailed?: boolean): string {
```
fabioberger:
```suggestion
    public getSignature(isDetailed?: boolean): string {
```

Stick to boolean naming conventions.

packages/utils/src/abi_encoder/utils/signatureParser.ts
```diff
@@ -0,0 +1,101 @@
+import { DataItem } from 'ethereum-types';
```
fabioberger: Why is this file name not camel-case? It should be.

packages/utils/src/abi_encoder/evm_data_types/pointer.ts
LogvinovLeon: Let's investigate why was this not caught by linter rule

contracts/protocol/test/utils/fill_order_combinatorial_utils.ts
```diff
@@ -613,13 +613,13 @@ export class FillOrderCombinatorialUtils {
...
         const params = orderUtils.createFill(signedOrder, takerAssetFillAmount);
-        const expectedAbiEncodedData = this.exchangeWrapper.abiEncodeFillOrder(signedOrder, { takerAssetFillAmount });
-        const libsAbiEncodedData = await this.testLibsContract.publicAbiEncodeFillOrder.callAsync(
+        const abiDataEncodedByContract = await this.testLibsContract.publicAbiEncodeFillOrder.callAsync(
             params.order,
             params.takerAssetFillAmount,
             params.signature,
         );
-        expect(libsAbiEncodedData).to.be.equal(expectedAbiEncodedData, 'ABIEncodedFillOrderData');
+        const paramsDecodeddByClient = this.exchangeWrapper.abiDecodeFillOrder(abiDataEncodedByContract);
```
abandeali1: `paramsDecodedByClient`

packages/utils/src/abi_encoder/evm_data_types/method.ts
abandeali1: Shouldn't the return type here be `any[]`?

packages/utils/src/abi_encoder/utils/signatureParser.ts
abandeali1: You mean snake-case right?

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -1,1877 +0,0 @@
-// tslint:disable:no-consecutive-blank-lines ordered-imports align trailing-comma whitespace class-name
```
abandeali1: Why are all of the generated wrappers deleted rather than replaced?

packages/utils/src/abi_encoder/utils/signatureParser.ts
LogvinovLeon: You mean snake case?

packages/utils/src/abi_encoder/evm_data_types/pointer.ts
fabioberger: I suspect it's because it's a param name and not a variable. We should improve our linter rule.

packages/base-contract/src/index.ts
hysz: `abiUtils` is still used to rename overloaded methods and assert correct encoding.

`ethers` is used by `deployAsync`.
There is a task to remove this from `deployAsync` after which we can remove ethers.js.

contracts/protocol/test/utils/types.ts
```diff
@@ -0,0 +1,8 @@
+import { SignedOrder } from '@0x/types';
+import { BigNumber } from '@0x/utils';
+
+export interface AbiDecodedFillOrderData {
+    order: SignedOrder;
+    makerAssetFillAmount: BigNumber;
```
abandeali1: `takerAssetFillAmount`

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -105,15 +105,7 @@ export class AssetProxyOwnerContract extends BaseContract {
...
-            const functionSignature = 'owners(uint256)';
-            const inputAbi = self._lookupAbi(functionSignature).inputs;
-            [index_0
-        ] = BaseContract._formatABIDataItemList(inputAbi, [index_0
-        ], BaseContract._bigNumberToString.bind(self));
-            BaseContract.strictArgumentEncodingCheck(inputAbi, [index_0
-        ]);
-            const ethersFunction = self._lookupEthersInterface(functionSignature).functions.owners;
-            const encodedData = ethersFunction.encode([index_0
+            const encodedData = self._strictEncodeArguments('owners(uint256)', [index_0
```
abandeali1: Doesn't particularly matter since these are auto-generated, but I'm curious why the formatting is so strange (looks like it was this way before too).

packages/utils/src/abi_encoder/evm_data_types/method.ts
```diff
@@ -44,7 +44,19 @@ export class MethodDataType extends AbstractSetDataType {
...
     }
 
-    public getSignature(): string {
+    public decodeReturnValuesAsArray(returnData: string, rules?: DecodingRules): any[] {
+        const returnValues = this.decodeReturnValues(returnData, rules);
+        const returnValuesAsArray = _.isObject(returnValues) ? _.values(returnValues) : [returnValues];
+        return returnValuesAsArray;
+    }
+
+    public decodeReturnValuesAsArrayOrNull(returnData: string, rules?: DecodingRules): any {
```
abandeali1: This should also return `any[]`.

packages/utils/src/abi_encoder/evm_data_types/method.ts
```diff
@@ -44,7 +44,19 @@ export class MethodDataType extends AbstractSetDataType {
         return returnValues;
     }
 
-    public getSignature(): string {
+    public decodeReturnValuesAsArray(returnData: string, rules?: DecodingRules): any[] {
```
abandeali1: For all of these functions with an `any` return type (or variant), should the signatures include a generic type? E.g:

```
public decodeReturnValuesAsArray<T>(returnData: string, rules?: DecodingRules): T[]
```

Thoughts @fabioberger @LogvinovLeon @hysz ?

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
hysz: That is a good question. I always figured it had something to do with Handlebars. @LogvinovLeon any ideas?

packages/utils/src/abi_encoder/evm_data_types/method.ts
```diff
@@ -44,7 +44,20 @@ export class MethodDataType extends AbstractSetDataType {
         return returnValues;
     }
 
-    public getSignature(): string {
+    public strictDecodeReturnValue<T>(returndata: string, rules?: DecodingRules): T {
```
abandeali1: Is there a point in having a non-strict version of this still? I feel like we could probably add generic types to the regular `decodeReturnValues`. 

Also this doesn't have to be in this PR if it's a lot of changes, but we should add generic types to the regular `decode` function as well.

hysz: I've created a task to use generics for all decoding. Shouldn't take long but I think it makes sense to have in a separate PR as well.

PR #1480, opened 2019-01-02
[instant] No more laggy input

fragosti:
## Description

I started by profiling some keystrokes in Instant: 

screen shot 2019-01-02 at 3 48 05 pm

The main thing to notice is that there are 3 main render flame-sub-graphs. The input being re-rendered (good), the order details being re-rendered (OK, since that actually displays the amount), and the token selector (bad, takes the longest, and isn't even displayed).

This PR does ~3 things, but the first two aren't that impactful.
1. Convert all `Component` instances to `PureComponent`. There really isn't a downside to doing this, and there is a perf upside.
2. Give all stateless components the static `displayName` attribute. This helps debugging tools know what component they are. 
3. Unmount the token selector when the sliding panel animation finishes. Now that 3rd flame subgraph is gone (sorry, used a different profiler at the end of the branch so image is slightly different).

screen shot 2019-01-02 at 7 25 08 pm

## Testing instructions

http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com

## Checklist:

### DIFF
--- a/packages/instant/src/components/amount_placeholder.tsx
+++ b/packages/instant/src/components/amount_placeholder.tsx
@@ -30,3 +30,5 @@ export const AmountPlaceholder: React.StatelessComponent<AmountPlaceholderProps>
         return <PlainPlaceholder color={props.color} />;
     }
 };
+
+AmountPlaceholder.displayName = 'AmountPlaceholder';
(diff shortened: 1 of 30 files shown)


steveklebanoff:
Nice detective work!

This looks good to me, although I do have two comments:

- Why do we need to unmount the token selector when the animation is done? Shouldn't it being a PureComponent be enough for it not to re-render when an input is changed?

- The selector feels better on my iPhone 6, but still can feel laggy if entering in and removing numbers quickly.  I looked a bit and don't see any other obvious performance improvements we could make -- I changed a few bigger components to `PureComponent` instead of `StatelessComponent` and didn't see a big change.  I also removed the account heartbeat, and that didn't seem to help either

PR #1481, opened 2019-01-02
[instant] Dropdown hover 10% of primary color

steveklebanoff:
## Description

Makes the dropdown color the primary color at 10% opacity

## Checklist:


fragosti:
fwiw I opted not to do this because it the token selector dropdown is grey and I thought it would be good to be consistent. But if this looks good, ship it!

PR #1482, opened 2019-01-02
[monorepo-scripts] Automatically alert new release to Discord

steveklebanoff:
## Description

Alerts discord channel when new package is published.

⚠️ *AFTER MERGE*: update publish doc to mention new required ENV variable ⚠️ 

I think it'd be good to use a webhook URL for a private channel to test it out for one or two publishes, and then move it to announce in a public channel (existing or new)

Looks like this:

image

## Checklist:


packages/monorepo-scripts/src/utils/github_release_utils.ts
```diff
@@ -12,7 +12,10 @@ import { utils } from './utils';
 
 const publishReleaseAsync = promisify(publishRelease);
 // tslint:disable-next-line:completed-docs
-export async function publishReleaseNotesAsync(packagesToPublish: Package[], isDryRun: boolean): Promise<void> {
+export async function publishReleaseNotesAsync(
+    packagesToPublish: Package[],
+    isDryRun: boolean,
+): Promise<string | undefined> {
```
BMillman19: Looks like we are returning `aggregateNotes` which is initialized as an empty string, is it necessary to expand the return type here to cover `undefined`?

packages/monorepo-scripts/src/utils/discord.ts
```diff
@@ -0,0 +1,27 @@
+import { fetchAsync } from '@0x/utils';
+
+import { constants } from '../constants';
+
+import { utils } from './utils';
+
+export const alertDiscord = async (releaseNotes: string): Promise<boolean> => {
```
BMillman19: following our conventions, this should be `alertDiscordAsync`. We do have a custom linter rule for enforce this suffix for method names but I don't think it applies to consts

packages/monorepo-scripts/src/utils/discord.ts
```diff
@@ -0,0 +1,27 @@
...
+    const webhookUrl = constants.discordAlertWebhookUrl;
+    if (!webhookUrl) {
+        utils.log('Not alerting to discord because webhook url not set');
+        return false;
```
BMillman19: should we throw these as errors and just have a return type of `void`?

packages/monorepo-scripts/src/utils/github_release_utils.ts
steveklebanoff: See line 76, if `isDryRun` there's a blank `return;` which returns undefined

packages/monorepo-scripts/src/utils/discord.ts
steveklebanoff: ugh, keep forgetting this convention :(

packages/monorepo-scripts/src/utils/discord.ts
steveklebanoff: Yes -- that seems slightly better than the current approach

steveklebanoff:
@BMillman19 updated!

packages/monorepo-scripts/src/constants.ts
```diff
@@ -5,5 +5,7 @@ export const constants = {
     stagingWebsite: 'http://staging-0xproject.s3-website-us-east-1.amazonaws.com',
     lernaExecutable: path.join('node_modules', '@0x-lerna-fork', 'lerna', 'cli.js'),
     githubPersonalAccessToken: process.env.GITHUB_PERSONAL_ACCESS_TOKEN_0X_JS,
+    discordAlertWebhookUrl: process.env.DISCORD_GITHUB_RELEASE_WEBHOOK_URL,
```
fabioberger: Does this need to be an ENV var? Why not just hard-code it here? Don't think it will be changing frequently.

packages/monorepo-scripts/src/publish.ts
```diff
@@ -84,7 +85,16 @@ async function confirmAsync(message: string): Promise<void> {
...
     const isDryRun = configs.IS_LOCAL_PUBLISH;
-    await publishReleaseNotesAsync(updatedPublicPackages, isDryRun);
+    const releaseNotes = await publishReleaseNotesAsync(updatedPublicPackages, isDryRun);
+    utils.log('Published release notes');
+
+    if (!isDryRun && releaseNotes) {
+        try {
+            await alertDiscordAsync(releaseNotes);
+        } catch (e) {
+            utils.log("Couldn't alert discord, error: ", e.message, '. Please alert manually.');
```
fabioberger: Let's add a note about how to publish still succeeded so people don't freak out at getting an error message.

packages/monorepo-scripts/src/utils/discord.ts
```diff
@@ -0,0 +1,26 @@
+import { fetchAsync } from '@0x/utils';
+
+import { constants } from '../constants';
+
+import { utils } from './utils';
+
+export const alertDiscordAsync = async (releaseNotes: string): Promise<void> => {
+    const webhookUrl = constants.discordAlertWebhookUrl;
+    if (!webhookUrl) {
```
fabioberger:
```suggestion
    if (webhookUrl === undefined) {
```

packages/monorepo-scripts/src/constants.ts
steveklebanoff: This will be a URL that will allow the posting to the #announcements ( or other channel ) under the name of something like "0x Release Bot".  We shouldn't expose this URL publicly, because malicious folks could use it to spam.

steveklebanoff:
@fabioberger or @BMillman19 this is updated - can I get another set of :eyes: or an approval when ya'll have time?

fabioberger:
Excited for this!

PR #1484, opened 2019-01-04
[sra-spec] Add title tag

steveklebanoff:
## Description

Add title tag to OpenSwagger to ReDoc page.  This was annoying me when I couldn't find the Standard Relayer API in my browser tabs :) 

## Checklist:

### DIFF
--- a/packages/sra-spec/public/index.html
+++ b/packages/sra-spec/public/index.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <title>ReDoc</title>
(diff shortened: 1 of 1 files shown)


packages/sra-spec/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
```
fragosti: Since you didn't change the contents of the package (ie. the API) , I don't think you need to add to the change log.

steveklebanoff: Got it -- removed this

PR #1485, opened 2019-01-04
Python contract demo, with lots of refactoring

feuGeneA:
## Description

Implemented a demonstration call to `Exchange.getOrderInfo()`, in `python-packages/contract_demo/test/test_exchange.py`.  ([Asana task](https://app.asana.com/0/930708194965189/949023754081596/f))

And, did quite a bit of refactoring, including:

* Changing the data types of some fields in the `Order` class so that it will be compatible with web3, allowing you to pass an instance of the class as the argument to `getOrderInfo()`.
*  Some utilities to convert between order representations.  One representation is for compatibility with the JSON schema validation, and the other is for compatibility with web3.  ([Asana task](https://app.asana.com/0/930708194965189/949023754081597/f))
*  Getting all of the contract addresses into python (only had Exchange before), and into their own package (`0x-contract-addresses`).  ([Asana task](https://app.asana.com/0/930708194965189/900917330466661/f))
*  Putting a `0x-contract-artifacts` package in place.  ([Asana task](https://app.asana.com/0/930708194965189/900917330466661/f))
*  Adding some top-level scripts to be able to execute commands (install, test, lint, etc) against all packages simultaneously.

Also note that I had to duplicate both the JSON schemas and the contract artifacts that are stored in the TypeScript packages.  I tried to use symlinks, and tried a few other hacky things, but I couldn't get around this duplication.  To make myself feel better about it, I added an additional "linter," to both `contract-addresses` and `contract-artifacts`, that's just a diff between what's in the python package and what's in the TypeScript package, so that builds will fail if they end up diverging for some reason.

Rendered docs:
*  [`0x-contract-addresses`](http://0x-contract-addresses-py.s3-website-us-east-1.amazonaws.com/)
*  [`0x-contract-artifacts`](http://0x-contract-artifacts-py.s3-website-us-east-1.amazonaws.com/)
*  [`0x-contract-demo`](http://0x-contract-demo-py.s3-website-us-east-1.amazonaws.com/)

## Testing instructions

It's all in CI!

## Types of changes

## Checklist:


LogvinovLeon:
I think I was requested to review that because I'm a CODEOWNER of `.circle.yml`. That file changes look good. Didn't look too closely into other changes. It's a shame we can't re-use schemas. I will maybe give it one more try. I like the solution with a linter command, but it will be brilliant if we can figure it out. Otherwise we'll have that problem in any language in the future. Maybe we can download the npm package in the python folder and unpack it and include files from there. Maybe we can put schemas outside of TS packages and reference it upwards.

python-packages/order_utils/src/zero_ex/json_schemas/schemas/zero_ex_transaction_schema.json
```diff
@@ -0,0 +1,10 @@
+{
+    "id": "/zeroExTransactionSchema",
+    "properties": {
+        "data": { "$ref": "/hexSchema" },
+        "signerAddress": { "$ref": "/addressSchema" },
+        "salt": { "$ref": "/wholeNumberSchema" }
+    },
+    "required": ["data", "salt", "signerAddress"],
+    "type": "object"
+}
```
fabioberger: @feuGeneA did you simply copy over all schemas?

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -107,47 +82,154 @@ class SignatureType(Enum):
...
+     'takerFee': 0}
+    """  # noqa: E501 (line too long)
+    assert_valid(jsdict, "/orderSchema")
+
+    order = cast(Order, copy(jsdict))
+
+    def strip_0x(hex_string: str) -> str:
+        if hex_string.startswith("0x"):
+            return hex_string[2:]
+        return hex_string
```
fabioberger: Why have this be a nested method? I think we will need this in other places. Can we move it to `utils`?

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
```diff
@@ -167,14 +249,15 @@ def generate_order_hash_hex(order: Order) -> str:
     ...         'takerFee': "0",
     ...         'expirationTimeSeconds': "12345",
     ...         'salt': "12345",
-    ...         'makerAssetData': "0x0000000000000000000000000000000000000000",
-    ...         'takerAssetData': "0x0000000000000000000000000000000000000000",
-    ...         'exchangeAddress': "0x0000000000000000000000000000000000000000",
+    ...         'makerAssetData': (0).to_bytes(1, byteorder='big') * 20,
+    ...         'takerAssetData': (0).to_bytes(1, byteorder='big') * 20,
```
fabioberger: Can we return `(0).to_bytes(1, byteorder='big')` from a helper method too? Seems useful elsewhere.

python-packages/contract_demo/test/test_exchange.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # false positive from pylint: disable=no-member
+    exchange: datatypes.Contract = web3_instance.eth.contract(
+        address=to_checksum_address(contract_address),
+        abi=zero_ex.contract_artifacts.abi_by_name("Exchange"),
+    )
+
+    order_info = OrderInfo(*exchange.call().getOrderInfo(order))
+
+    assert isinstance(order_info.order_status, int)
```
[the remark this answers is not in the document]
feuGeneA: No.  This is a python built-in.

python-packages/order_utils/src/zero_ex/json_schemas/schemas/zero_ex_transaction_schema.json
feuGeneA: Yup.

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
feuGeneA: Did it because I default to as-local-as-possible, unless a specific, known re-use is intended.  But I see your point, and, checking further, I see that there's an `eth_utils.remove_0x_prefix()` method that I can use instead of re-inventing the wheel, so I'll do that.

python-packages/order_utils/src/zero_ex/order_utils/__init__.py
feuGeneA: For all of the cases in this PR, I'd prefer to keep it this way, because these are "docstrings", example snippets that, in my opinion, should be as self-contained as possible, ideally not relying on any outside context at all.

python-packages/order_utils/src/zero_ex/json_schemas/schemas/zero_ex_transaction_schema.json
feuGeneA: Note that the diff command in the linter run will ensure that there are no differences between the copies.

feuGeneA:
@LogvinovLeon 

> Otherwise we'll have that problem in any language in the future.

I think the problem is specific to Python, in particular [this bug with its packaging tools not following symlinks](https://github.com/pypa/setuptools/issues/415).

> Maybe we can download the npm package in the python folder and unpack it and include files from there.

That would require using a custom docker image in Circle CI, since their standard Python image doesn't include npm.  Wouldn't be too much work, but I think there's an advantage to keeping things simple enough to use their regularly-maintained image.

> Maybe we can put schemas outside of TS packages and reference it upwards.

Don't think that would work for this case, as it would still be just a relative path from the python package source, which would still need a symlink.  Maybe I'm missing something though...

feuGeneA:
@fabioberger I addressed all your requests.

PR #1489, opened 2019-01-08
[website] Audit and fix 0x.org migration URLs

fragosti:
## Description

This branch goes along with some EC2 load balancer fiddling to make sure the endpoints have the correct SSL certificate. 

The backup geth nodes aren't working anyway because of CORS, so I've gone ahead and changed the configs for those already. 

For website-api.0x.org, I'll have to update those configs when I deploy.

## Checklist:

### DIFF
--- a/packages/website/ts/utils/configs.ts
+++ b/packages/website/ts/utils/configs.ts
@@ -39,8 +39,8 @@ export const configs = {
     ] as OutdatedWrappedEtherByNetworkId[],
     // The order matters. We first try first node and only then fall back to others.
     PUBLIC_NODE_URLS_BY_NETWORK_ID: {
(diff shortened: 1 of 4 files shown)


packages/website/ts/utils/configs.ts
```diff
@@ -5,8 +5,8 @@ const INFURA_API_KEY = 'T5WSC8cautR4KXyYgsRs';
 
 export const configs = {
     AMOUNT_DISPLAY_PRECSION: 5,
-    BACKEND_BASE_PROD_URL: 'https://website-api.0xproject.com',
-    BACKEND_BASE_STAGING_URL: 'https://staging-website-api.0xproject.com',
+    BACKEND_BASE_PROD_URL: 'https://website-api.0x.org',
```
BMillman19: We're using this URL in `modal_contact.tsx`

PR #1494, opened 2019-01-08
Dockerize OrderWatcher WS Server

fabioberger:
## Description

This PR:

- Adds the `server` entry-point that starts a OrderWatcher WS server. This is the entry-point used by Dockerfile.
- A Dockerfile config that can be used to build a docker container that runs the server.
- A few missing assertions to the json-schemas package.
- Fixes the OW WS server tests so that they are guaranteed to be independent from one another.

## Testing instructions

- Run the test suite
- Build the docker container by running `docker build -t order-watcher .` from the `order-watcher` package dir. Then run `docker run -ti -p 9090:8080 -e IS_VERBOSE=true order-watcher` to spin it up on localhost port 9090. Lastly, run the existing test-suite, simply changing port 8080 to 9090 throughout.

## Checklist:


packages/json-schemas/src/schema_validator.ts
```diff
@@ -8,12 +8,18 @@ import { schemas } from './schemas';
  */
 export class SchemaValidator {
     private readonly _validator: Validator;
+    private static _assertSchemaNotUndefined(schema: Schema): void {
```
LogvinovLeon:
```suggestion
    private static _assertSchemaDefined(schema: Schema): void {
```

Issue #1486, opened 2019-01-04
Consider using prettier-solidity


fvictorio:
Hi, I'm a contributor in the [solidity plugin for prettier](https://github.com/prettier-solidity/prettier-plugin-solidity). The project is in a very early stage, but I think it's close to a beta. Given that, it would be very helpful to have some projects like this give it a try.

Since you already use prettier, and since it seems that tooling is important to you (as authors of `sol-cover` and others), maybe you are interested in giving it a try? If that's the case, I could make a PR with it (after solving some pending issues we have).

Thanks!

LogvinovLeon:
We have actually tried it before but there were some issue we've run into (already reported on your repo). We'd be happy to give it a second try when those are solved.

PR #1495, opened 2019-01-09
Add the solidity prettier plugin

fvictorio:
See #1486.

This PR adds the solidity plugin for prettier and applies it to some of the contracts. Since there are several differences between the plugin's output and what there was already here, I only updated the contracts under `contracts/examples`; otherwise the number of changes would've been huge.

The plugin is still in development, so any feedback is welcome!

### DIFF
--- a/contracts/.prettierrc
+++ b/contracts/.prettierrc
@@ -1,0 +1,6 @@
+{
+    "tabWidth": 4,
+    "printWidth": 80,
+    "trailingComma": all,
+    "singleQuote": true
+}
--- a/contracts/examples/contracts/ExchangeWrapper/ExchangeWrapper.sol
+++ b/contracts/examples/contracts/ExchangeWrapper/ExchangeWrapper.sol
@@ -22,16 +22,12 @@ pragma experimental ABIEncoderV2;
 import "@0x/contracts-interfaces/contracts/protocol/Exchange/IExchange.sol";
 import "@0x/contracts-libs/contracts/libs/LibOrder.sol";
 
-
 contract ExchangeWrapper {
-
     // Exchange contract.
     // solhint-disable-next-line var-name-mixedcase
     IExchange internal EXCHANGE;
 
-    constructor (address _exchange)
-        public
-    {
+    constructor(address _exchange) public {
         EXCHANGE = IExchange(_exchange);
     }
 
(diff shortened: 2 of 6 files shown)


abandeali1:
Thanks for the PR @fvictorio . I ran this locally on the `protocol` directory to get a better sense of how this handles different cases, and here are my initial thoughts:

- Contract inheritance doesn't seem to adhere to the specified print width. For example, the first line of the Exchange contract looks like this after running prettier:

```
contract Exchange is MixinExchangeCore, MixinMatchOrders, MixinSignatureValidator, MixinTransactions, MixinAssetProxyDispatcher, MixinWrapperFunctions {
```

I would much prefer these to all be on separate lines for readability purposes.

- In a similar vein, I like modifiers to be on separate lines. This isn't as readable as I'd like (smaller print width makes this a little better, but not perfect):

```
    function fillOrder(
        Order memory order,
        uint256 takerAssetFillAmount,
        bytes memory signature
    ) public nonReentrant returns (FillResults memory fillResults) {
```

I recognize that these are more of a preference than actual issues, but it would be great if these things were somewhat configurable. 

I also noticed that the `"trailingComma": all` and `"singleQuote": true` options in `.prettierrc` don't work, though this doesn't affect our repo as we would disable them anyways.

fvictorio:
Hey @abandeali1, sorry for not responding before, I was on vacation. Thanks a lot for the comments.

> * Contract inheritance doesn't seem to adhere to the specified print width.

This sounds like a bug, I'll look into it.

> * In a similar vein, I like modifiers to be on separate lines.

This is a little tricky. Right now what we are doing is to split the parameters if the line is too long. If, after that, the line with the modifiers is *still* too long, then it gets split too. I would prefer that, if the parameters are split, then the modifiers are split too, but I couldn't find a way to do it (yet).

That being said, I noticed that this repo seem to *always* split the modifiers, no matter what the length is. That would be even easier, but it's probably too weird and (in my opinion) different from what prettier core usually does.

> I also noticed that the `"trailingComma": all` and `"singleQuote": true` options in `.prettierrc` don't work

I'm not sure if Solidity allows trailing commas. Does it? The single quote being ignored is a problem for sure.

Thanks again for the comments, I will open the proper issues.

PR #1506, opened 2019-01-10
 [instant] Implement logging and friendly wallet name for Opera

fragosti:
## Description

Now that Instant works in Opera I added it as a provider type and gave it a friendly name. This also means we'll see a readable name in Heap.

## Checklist:

### DIFF
--- a/packages/instant/src/util/env.ts
+++ b/packages/instant/src/util/env.ts
@@ -42,18 +42,21 @@ export const envUtil = {
         }
     },
     getProviderType(provider: Provider): ProviderType | undefined {
+        const anyProvider = provider as any;
         if (provider.constructor.name === 'EthereumProvider') {
             return ProviderType.Mist;
-        } else if ((provider as any).isTrust) {
+        } else if (anyProvider.isTrust) {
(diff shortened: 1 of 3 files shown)


packages/instant/src/util/env.ts
```diff
@@ -42,18 +42,21 @@ export const envUtil = {
...
+        } else if (anyProvider.isParity) {
             return ProviderType.Parity;
-        } else if ((provider as any).isMetaMask) {
+        } else if (anyProvider.isMetaMask) {
             return ProviderType.MetaMask;
         } else if (!_.isUndefined(_.get(window, 'SOFA'))) {
             return ProviderType.CoinbaseWallet;
         } else if (!_.isUndefined(_.get(window, '__CIPHER__'))) {
             return ProviderType.Cipher;
+        } else if (envUtil.getBrowser() === Browser.Opera && !anyProvider.isMetaMask) {
```
BMillman19: what happens if we're on opera on desktop w/o metamask installed?

fragosti: We don't show anything. This method never gets called.

PR #1508, opened 2019-01-10
Output Chrome 41 compatible version of 0x instant

justinbarry:
This is important for SEO as Googlebot is still using Chrome 41 to crawl
thick clients.

## Description

Certain dependencies of this project ship code with ES6 syntax. Webpack doesn't care, sadly Googlebot does:

https://developers.google.com/search/docs/guides/rendering

## Testing instructions

Load in Chrome v41 using browserstack or similar service.

## Types of changes

## Checklist:


packages/instant/package.json
```diff
@@ -78,6 +80,12 @@
         "@types/redux": "^3.6.0",
         "@types/styled-components": "4.0.1",
         "awesome-typescript-loader": "^5.2.1",
+        "babel-loader": "^7.1.5",
+        "babel-preset-env": "^1.7.0",
+        "babel-plugin-transform-runtime": "^6.23.0",
+        "babel-plugin-transform-es3-member-expression-literals": "^6.22.0",
```
fragosti: Is it necessary to transform to es3, or is this what you would use even if you're just transforming to es5?

justinbarry: Thanks @fragosti!

Those are transitive dependencies a babel config in one of the npm modules requires. If you remove those and build you will see what I mean.

I would *love* to find a way around this.

fragosti:
@justinbarry looks good! if you could CI passing I think we are good to merge.

PR #1512, opened 2019-01-11
[instant] [asset-buyer] Helper methods for checking liquidity

steveklebanoff:
## Description

Adds `getLiquidityForAssetDataAsync` function to `AssetBuyer` class.

Adds following helper functions to `zeroExInstant`:
- `assetDataForERC20TokenAddress`
- `hasMetaDataForAssetData`
- `hasLiquidityForAssetDataAsync`

Adds mocking abilities to AssetBuyer using `typemoq`

⚠️ After Merge ⚠️ 

- Update wiki docs to include these new functions
- Respond to integrator who requested this type of functionality

## Testing instructions

See demo at http://0x-instant-dogfood.s3-website-us-east-1.amazonaws.com/liquidity_demo.html (can view source for code)

## Checklist:


packages/asset-buyer/src/asset_buyer.ts
```diff
@@ -177,6 +179,40 @@ export class AssetBuyer {
...
+        const assetPairs = await this.orderProvider.getAvailableMakerAssetDatasAsync(assetData);
+        const etherTokenAssetData = this._getEtherTokenAssetDataOrThrow();
+        if (!assetPairs.includes(etherTokenAssetData)) {
+            return {
+                tokensAvailableInBaseUnits: 0,
+                ethValueAvailableInWei: 0,
+            };
+        }
+
+        const ordersAndFillableAmounts = await this.getOrdersAndFillableAmountsAsync(
```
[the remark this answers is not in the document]
fragosti: Is it a lot of work? Would be nice to keep that interface consistent but don't feel strongly about it.

packages/asset-buyer/src/types.ts
```diff
@@ -121,3 +126,11 @@ export interface OrdersAndFillableAmounts {
     orders: SignedOrder[];
     remainingFillableMakerAssetAmounts: BigNumber[];
 }
+
+/**
+ * Represents available liquidity for a given assetData
+ */
+export interface LiquidityForAssetData {
+    tokensAvailableInBaseUnits: number;
+    ethValueAvailableInWei: number;
```
fragosti: `amountAvailableInWei`, `valueAvailableInWei` ?

packages/asset-buyer/src/utils/calculate_liquidity.ts
```diff
@@ -0,0 +1,34 @@
...
+            tokensAvailableInBaseUnits: new BigNumber(0),
+            ethValueAvailableInWei: new BigNumber(0),
+        },
+    );
+
+    // Turn into regular numbers
+    return {
+        tokensAvailableInBaseUnits: liquidityInBigNumbers.tokensAvailableInBaseUnits.toNumber(),
+        ethValueAvailableInWei: liquidityInBigNumbers.ethValueAvailableInWei.toNumber(),
+    };
```
fragosti: hmmm would've been nice to stick to `BigNumber`, but yea the === 0 case is not great. 

```
new BigNumber(2) > 1 
true
new BigNumber(1) > 1 
false
new BigNumber(0) > 1 
false
new BigNumber(0) === 0 
false
new BigNumber(0) == 0 
true
```
Maybe keep asset_buyer using `BigNumber` and convert it to number at the Instant level?

packages/instant/src/index.umd.ts
```diff
@@ -122,6 +128,42 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
     window.onpopstate = onPopStateHandler;
 };
 
+export const assetDataForERC20TokenAddress = (tokenAddress: string): string => {
```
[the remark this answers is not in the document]
fragosti: I think they're fine here :)

packages/asset-buyer/src/utils/calculate_liquidity.ts
steveklebanoff: I opted in for using regular numbers here, because I didn't want the integrator to have to introduce the BigNumber dependency for checking liquidity

packages/asset-buyer/src/types.ts
steveklebanoff: Is "wei" an appropriate term for an erc20 token which could have a decimal amount that is different than 18?  I prefer the current naming, but could be swayed if you feel strongly or if there is an established pattern here

packages/asset-buyer/src/asset_buyer.ts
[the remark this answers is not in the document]
steveklebanoff: Not a huge amount of work, but for the sake of time & priority would prefer to leave as-is

packages/asset-buyer/src/utils/calculate_liquidity.ts
steveklebanoff: Ah, good point!  Here's a few of the ideas we discussed:

**Return an instance of BigNumber**. Export zeroExInstant.BigNumber on the window.  To ensure that integrators don't mistake this for a regular number, prepend "bigNumber" onto the return value key names

_Pros:_
- Guarantees precision
- We return BigNumbers throughout the rest of the code-base, so this is likely what a 0x-er would expect

_Cons_
- Forces us to always ship a version of BigNumber with our bundle, increasing the minimum bundle size in a future where we code-split
- If they also package their own version of BigNumber, BigNumber code could be shipped twice the browser
- Could get into weird scenarios if they use a static method (i.e. `BigNumber.maximum`) using a different version of BigNumber

**Return a string representation of the number**:  Have `asset-buyer` return a BigNumber, but call `.toString()` on it in the helper method on the `zeroExInstant` window object.  Prepend the word "string" onto the return value keys so folks do not mistake it for a number. 
 This would allow integrator to use whatever number library they'd like to parse the value.  They could opt-in to use something like [big.js](https://github.com/MikeMcl/big.js/) instead of BigNumber if they were concerned about bundle size (this also would require us updating the check liquidity method to not rely on asset-buyer and BigNumber)

**Rip out the liquidity # response all together**: Just return a boolean representing if the token has liquidity.  This would handle the "Should I show the 'Buy' button for a specific erc20 token?" question, and would allow the integrator to query a list of relayers and choose the first one that has liquidity.  This _does not_ handle the situation in which there more than one relayer supports a token and an integrator wants to choose the relayer with the most liquidity

With instant being served by a cdn which always hosts the latest version of the code, we don't want to later change this interface and break integrations, so it seems important to get this right.

steveklebanoff:
@BMillman19 @fragosti Updated, and also deployed updated demo code on dogfood

packages/asset-buyer/src/asset_buyer.ts
```diff
@@ -177,6 +179,40 @@ export class AssetBuyer {
...
+     *
+     * @return  An object that conforms to LiquidityForAssetData that satisfies the request. See type definition for more information.
+     */
+    public async getLiquidityForAssetDataAsync(
+        assetData: string,
+        options: Partial<LiquidityRequestOpts> = {},
+    ): Promise<LiquidityForAssetData> {
+        const shouldForceOrderRefresh =
+            options.shouldForceOrderRefresh !== undefined ? options.shouldForceOrderRefresh : false;
+        assert.isString('assetData', assetData);
```
BMillman19: We're asserting string here vs hex string which we are doing in other places, but actually it looks like `asset_data_utils.ts` has a new method `validateAssetDataOrThrow` which will actually check the shape of the string and make sure it is valid in the context of erc20, erc721, etc, as well as produce a more helpful error message

packages/asset-buyer/src/types.ts
```diff
@@ -117,7 +124,19 @@ export enum AssetBuyerError {
...
 export interface OrdersAndFillableAmounts {
     orders: SignedOrder[];
     remainingFillableMakerAssetAmounts: BigNumber[];
 }
+
+/**
+ * Represents available liquidity for a given assetData
+ */
+export interface LiquidityForAssetData {
+    tokensAvailableInBaseUnits: BigNumber;
```
BMillman19: did we decide if we are gonna prepend bigNumber here? we don't have a precedent for it but it might be helpful

packages/instant/src/index.umd.ts
```diff
@@ -122,6 +128,43 @@ export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_Z
...
+    if (provider !== undefined) {
+        assert.isWeb3Provider('provider', provider);
+    }
+
+    const bestProvider: Provider = provider || providerFactory.getFallbackNoSigningProvider(networkId);
+
+    const assetBuyerOptions = { networkId };
+
+    const assetBuyer =
+        typeof orderSource === 'string'
```
BMillman19: hmm we should use `_.isString(orderSource` here because it is consistent with how we are type checking in `asset_buyer_factory.ts`, or change it there

packages/asset-buyer/src/types.ts
steveklebanoff: per convo, every # in AssetBuyer is already a BigNumber so no on this, only wanted to append bigNumber to the var name if it was exposed on the global `zeroExInstant` object

PR #1513, opened 2019-01-11
[website] Fix multiple react dom versions and layout issue

fragosti:
## Description

The website is broken. The error:
```
Uncaught Error: Unable to find node on an unmounted component.
    at invariant (react-dom.development.js:55)
    at findCurrentFiberUsingSlowPath (react-dom.development.js:4158)
    at findCurrentHostFiber (react-dom.development.js:4266)
    at findHostInstance (react-dom.development.js:17676)
    at Object.findDOMNode (react-dom.development.js:18145)
    at StyledComponent.componentDidMount (styled-components.browser.esm.js:1918)
    at commitLifeCycles (react-dom.development.js:14362)
    at commitAllLifeCycles (react-dom.development.js:15463)
    at HTMLUnknownElement.callCallback (react-dom.development.js:100)
    at HTMLUnknownElement.t._rollbar_wrapped.t._rollbar_wrapped (rollbar.umd.min.js:1)
```

Some digging led me to believe that this was because of conflicting versions of React DOM being included in the bundle. Upgrading packages with older versions of React fixed the problem. However, it seemed that as a result `yarn` was not installing certain packages, so I nuked yarn.lock and re-installed everything, which fixed this problem as well. The only build-related problem that remained after this was incompatible types with two packages that are meant to play nice together (redux, redux-devtools). **Still looking into this**.

Other changes:
I saw some layout issues in Portal, and discovered that the imported `basscss.css` file now included `var(--custom-var);` values, that were trumping the usual `x rem;` values. I am not sure why this happened, but I figured it was more consistent to use a `<link>` to include basscss anyway. That is how we do it for other css files, and website is not meant to be exported and included in another package.

I also saw a console warning about multiple versions of `styled-components` being present, and went ahead and made sure we were consistent on `styled-components` as well.

## Checklist:


fragosti:
http://staging-0xproject.s3-website-us-east-1.amazonaws.com/

PR #1529, opened 2019-01-18
Use fetchAsync instead of axios

feuGeneA:
## Checklist:

### DIFF
--- a/packages/pipeline/src/utils/index.ts
+++ b/packages/pipeline/src/utils/index.ts
@@ -1,4 +1,4 @@
-import { BigNumber } from '@0x/utils';
+import { BigNumber, fetchAsync } from '@0x/utils';
 export * from './transformers';
 export * from './constants';
 
@@ -50,4 +50,15 @@ export function handleError(e: any): void {
         console.error('(No stack trace)');
     }
     process.exit(1);
+}
+
+/**
+ * Does fetchAsync(), and checks the status code, throwing if it doesn't indicate success.
+ */
+export async function fetchSuccessfullyOrThrowAsync(url: string): Promise<any> {
+    const response = await fetchAsync(url);
+    if (!response.ok) {
+        throw new Error(`Unsuccessful HTTP status code (${response.status}): ${response.statusText}`);
+    }
+    return response.json();
 }
--- a/packages/pipeline/test/data_sources/bloxy/index_test.ts
+++ b/packages/pipeline/test/data_sources/bloxy/index_test.ts
@@ -1,0 +1,23 @@
+import * as chai from 'chai';
+import 'mocha';
+
+import { BloxySource, BloxyTrade } from '../../../src/data_sources/bloxy';
+import { chaiSetup } from '../../utils/chai_setup';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+describe('Bloxy data source', () => {
+    it('should return some trades', async () => {
+        const apiKey = process.env.BLOXY_API_KEY;
+        if (apiKey === undefined) {
+            throw new Error('Missing required env var: BLOXY_API_KEY');
+        }
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const bloxySource = new BloxySource(apiKey /*, 1 trades per query, 0 offset buffer*/);
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const thirtyOneMinutesAgo = Date.now(); // - 1 * 60 * 1000;
+        const rawTrades: BloxyTrade[] = await bloxySource.getDexTradesAsync(thirtyOneMinutesAgo);
+        expect(rawTrades.length).to.be.greaterThan(0);
+    });
+});
--- a/packages/pipeline/test/data_sources/relayer-registry/index_test.ts
+++ b/packages/pipeline/test/data_sources/relayer-registry/index_test.ts
@@ -1,0 +1,24 @@
+import * as chai from 'chai';
+import 'mocha';
+
+import { getRelayerInfoAsync, RelayerResponses } from '../../../src/data_sources/relayer-registry';
+import { chaiSetup } from '../../utils/chai_setup';
+
+const RELAYER_REGISTRY_URL =
+    'https://raw.githubusercontent.com/0xProject/0x-relayer-registry/4701c85677d161ea729a466aebbc1826c6aa2c0b/relayers.json';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+describe('Relayer registry data source', () => {
+    it('should contain Radar Relay', async () => {
+        const relayers: RelayerResponses = await getRelayerInfoAsync(RELAYER_REGISTRY_URL);
+
+        const radarUuid = 'e8d27d8d-ddf6-48b1-9663-60b0a3ddc716';
+
+        expect(radarUuid in relayers).to.be.true();
+
+        const radarEntry = relayers[radarUuid];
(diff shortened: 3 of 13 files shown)


packages/pipeline/test/data_sources/bloxy/index_test.ts
```diff
@@ -0,0 +1,23 @@
...
+const expect = chai.expect;
+
+describe('Bloxy data source', () => {
+    it('should return some trades', async () => {
+        const apiKey = process.env.BLOXY_API_KEY;
+        if (apiKey === undefined) {
+            throw new Error('Missing required env var: BLOXY_API_KEY');
+        }
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const bloxySource = new BloxySource(apiKey /*, 1 trades per query, 0 offset buffer*/);
```
xianny: What's going on with the commented out args here?

packages/pipeline/test/data_sources/bloxy/index_test.ts
```diff
@@ -0,0 +1,23 @@
...
+        // tslint:disable-next-line:custom-no-magic-numbers
+        const thirtyOneMinutesAgo = Date.now(); // - 1 * 60 * 1000;
```
xianny: ditto

packages/pipeline/test/data_sources/relayer-registry/index_test.ts
```diff
@@ -0,0 +1,24 @@
+import * as chai from 'chai';
+import 'mocha';
+
+import { getRelayerInfoAsync, RelayerResponses } from '../../../src/data_sources/relayer-registry';
+import { chaiSetup } from '../../utils/chai_setup';
+
+const RELAYER_REGISTRY_URL =
```
xianny: Hmm... this unit test requires an internet connection? Can we save the json as a fixture and mock the endpoint instead?

packages/pipeline/test/data_sources/trusted_tokens/index_test.ts
```diff
@@ -0,0 +1,46 @@
...
+import {
+    getMetamaskTrustedTokensAsync,
+    getZeroExTrustedTokensAsync,
+    MetamaskTrustedTokens,
+    ZeroExTrustedTokens,
+} from '../../../src/data_sources/trusted_tokens';
+import { chaiSetup } from '../../utils/chai_setup';
+
+const METAMASK_TRUSTED_TOKENS_URL =
+    'https://raw.githubusercontent.com/MetaMask/eth-contract-metadata/d45916c533116510cc8e9e048a8b5fc3732a6b6d/contract-map.json';
```
xianny: same as above for "live" unit test

packages/pipeline/test/parsers/token_metadata/index_test.ts
```diff
@@ -0,0 +1,51 @@
...
+import { chaiSetup } from '../../utils/chai_setup';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+const ZEROEX_CONTRACT_ADDRESS = '0xE41d2489571d322189246DaFA5ebDe1F4699F498';
+
+const ZRX_DECIMALS = 18;
+
+const metamaskTrustedToken: MetamaskTrustedTokenMeta = {
```
xianny: I think we should use actual API responses e.g. in json to test the parsers. Also helps us check if the json conforms to the response types that we defined. I know some of the other parser tests use JS objects, but I think the test is more useful if it uses the raw response. I'm up for further discussion if we don't agree yet.

packages/pipeline/test/data_sources/relayer-registry/index_test.ts
albrow: ➕ IMO we shouldn't send any requests to third party servers as part of our normal test suite. The reason to avoid this is that third party APIs can break (some of the specific ones we're using break more frequently than I would like), resulting in intermittent and confusing errors for other developers on the team.

packages/pipeline/src/utils/index.ts
```diff
@@ -51,3 +51,14 @@ export function handleError(e: any): void {
     }
     process.exit(1);
 }
+
+/**
+ * Does fetchAsync(), and checks the status code, throwing if it doesn't indicate success.
```
albrow: Optional: Let's clarify that "success" in this context means a status code between 200 and 299.

https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

packages/pipeline/test/parsers/token_metadata/index_test.ts
albrow: ➕ I agree with this.

packages/pipeline/test/data_sources/trusted_tokens/index_test.ts
```diff
@@ -0,0 +1,46 @@
...
+        expect(zrxToken.name).to.equal('0x');
+        expect(zrxToken.symbol).to.equal('ZRX');
+        expect(zrxToken.decimals).to.equal(ZRX_DECIMALS);
+        expect(zrxToken.erc20).to.be.true();
+    });
+    it('from 0x should include ZRX', async () => {
+        const tokens: ZeroExTrustedTokens = await getZeroExTrustedTokensAsync(ZEROEX_TRUSTED_TOKENS_URL);
+        const zrxToken = tokens.find(token => token.address === ZEROEX_CONTRACT_ADDRESS.toLowerCase());
+        expect(zrxToken).to.not.be.undefined();
+        if (zrxToken === undefined) {
```
albrow: If we have to implement this manual check anyways to satisfy the compiler, could we simplify the code by just skipping the Mocha version?

E.g.:

```diff
-        expect(zrxToken).to.not.be.undefined();
-        if (zrxToken === undefined) {
-            // without this check tsc complains that later statements could have zrxToken undefined
-            throw new Error('wtf');
-        }
+ if (zrxToken === undefined) {
+     throw new Error('zrxToken should not be undefined');
+ }
```

albrow: Actually there's an even simpler version if we're confident this is a false positive in the type-checker. Just add an `as ZeroExTrustedTokenMeta` type assertion:

```diff
-         const zrxToken = tokens.find(token => token.address === ZEROEX_CONTRACT_ADDRESS.toLowerCase());
+         const zrxToken = tokens.find(
+             token => token.address === ZEROEX_CONTRACT_ADDRESS.toLowerCase(),
+         ) as ZeroExTrustedTokenMeta;
          expect(zrxToken).to.not.be.undefined();
-         if (zrxToken === undefined) {
-             // without this check tsc complains that later statements could have zrxToken undefined
-             throw new Error('wtf');
-         }
```

packages/pipeline/test/parsers/token_metadata/index_test.ts
```diff
@@ -0,0 +1,51 @@
...
+    symbol: 'ZRX',
+    decimals: ZRX_DECIMALS,
+};
+
+describe('Token metadata parser', () => {
+    it('should parse token metadata from Metamask', () => {
+        const tokenMetadatas: TokenMetadata[] = parseMetamaskTrustedTokens({
+            [ZEROEX_CONTRACT_ADDRESS]: metamaskTrustedToken,
+        });
+        expect(tokenMetadatas[0].address).to.equal(ZEROEX_CONTRACT_ADDRESS);
```
albrow: Optional: This might be easier to read if we combine these smaller assertions into a single assertion: `expect(actual).to.deep.equal(expected);`.

PR #1537, opened 2019-01-22
[fix] allow devnet rpc to listen on any vhost

nitper:
## Description

[this only affects the devnet docker image]

geth's default value for `--rpcvhosts` is `localhost`, which does not work when trying to use the devnet in certain environments such as `docker-compose` or `gitlab-ci`

See [geth's CLI wiki](https://github.com/ethereum/go-ethereum/wiki/Command-Line-Options) for the official documentation on `--rpcvhosts`.
## Testing instructions

Build the docker image locally, and then use/connect to it in a `docker-compose` environment where the vhost of the devnet is *not* localhost.

## Types of changes

## Checklist:

### DIFF
--- a/packages/devnet/run.sh
+++ b/packages/devnet/run.sh
@@ -15,6 +15,7 @@ mkdir -p /var/log
     --rpc \
     --rpcaddr '0.0.0.0' \
     --rpcport 8501 \
+    --rpcvhosts '*' \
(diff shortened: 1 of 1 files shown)


nitper:
@albrow any chance you can push an updated image to [docker hub](https://hub.docker.com/r/0xorg/devnet)?

PR #1538, opened 2019-01-22
[feature] [package/0x-devnet] Add docker-compose basic config

BenSchZA:
Add docker-compose.yml config for 0x-devnet package Docker image. This can be expanded upon in future if necessary.

I don't think any testing is necessary for this change, it's more to make the Docker configuration self-commenting for those familiar with the docker-compose tool.

### DIFF
--- a/packages/devnet/docker-compose.yml
+++ b/packages/devnet/docker-compose.yml
@@ -1,0 +1,8 @@
+version: '3'
+services:
+  devnet:
+    image: 0x-devnet:latest
+    build:
+      context: .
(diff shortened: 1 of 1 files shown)


albrow:
Thanks for the PR @BenSchZA 😄 

I don't think this is super necessary but it's such a small change that we might as well merge it in case some people prefer using docker-compose over the plain old docker command. As you sort of hinted at, I think docker-compose might be more compelling in the future if we ever add more configuration options or expand functionality.

PR #1543, opened 2019-01-23
[asset-buyer] Fix test (assetData string)

steveklebanoff:
## Description

Per suggestion in PR ( https://github.com/0xProject/0x-monorepo/pull/1512#discussion_r248506087 ), I called `validateAssetDataOrThrow` to validate an assetData parameter, instead of just doing a `assert.isString('assetData', assetData);`.

But -- `validateAssetDataOrThrow` expects the input to be a string -- it calls `.slice` on it.  The test I had was sending in `false` and the code was throwing when it tried to call `.split`.  

So, now we ensure it's a string before we call the `validateAssetDataOrThrow`

## Checklist:


aniel2019:
# #1543

PR #1545, opened 2019-01-24
TEC MVP

abandeali1:
## Description

This PR implements a the MVP TEC contract with tests. This contract can probably still be heavily optimized.

## Checklist:


contracts/tec/contracts/src/libs/LibEIP712Domain.sol
```diff
@@ -0,0 +1,87 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.2;
+
+
+contract LibEIP712Domain {
```
fabioberger: Why use `EIP712` here? TEC's are the ones signing these orders, and they won't be doing it through a signer UI.

contracts/tec/contracts/src/mixins/MSignatureValidator.sol
```diff
@@ -0,0 +1,39 @@
...
+
+
+contract MSignatureValidator {
+
+    // Allowed signature types.
+    enum SignatureType {
+        Illegal,         // 0x00, default value
+        EIP712,          // 0x01
+        EthSign,         // 0x02
+        NSignatureTypes  // 0x03, number of signature types. Always leave at end.
```
fabioberger: Why "always leave at end"?

contracts/tec/contracts/src/libs/LibEIP712Domain.sol
abandeali1: True... but what's the downside of using EIP712 here? I feel like if we want to push for it as a standard then we should use it for all of our hashing.

contracts/tec/contracts/src/mixins/MSignatureValidator.sol
abandeali1: Kinda just a sanity check to prevent human error when writing this. It's never used, but we have this in the main Exchange contract as well.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
```diff
@@ -0,0 +1,199 @@
...
+        bytes[] memory approvalSignatures
+    )
+        public
+        view
+    {
+        // Get the orders from the the Exchange calldata in the 0x transaction
+        LibOrder.Order[] memory orders = decodeFillDataOrders(transaction.data);
+
+        // No approval is required for non-fill methods
+        if (orders.length > 0) {
```
fabioberger: This will catch `batchCancelOrders` requests. We need to special-case this somehow.

fabioberger: Nvm, misread.

contracts/tec/contracts/src/libs/LibEIP712Domain.sol
fabioberger: Fair. Let's keep it.

contracts/exchange-libs/contracts/src/LibAssetProxyErrors.sol
```diff
@@ -17,22 +17,22 @@
...
-    string constant TARGET_ALREADY_AUTHORIZED = "TARGET_ALREADY_AUTHORIZED";                // Target address must not already be authorized.
-    string constant INDEX_OUT_OF_BOUNDS = "INDEX_OUT_OF_BOUNDS";                            // Specified array index is out of bounds.
-    string constant AUTHORIZED_ADDRESS_MISMATCH = "AUTHORIZED_ADDRESS_MISMATCH";            // Address at index does not match given target address.
+    string constant internal SENDER_NOT_AUTHORIZED = "SENDER_NOT_AUTHORIZED";                        // Sender not authorized to call this method.
+    string constant internal TARGET_NOT_AUTHORIZED = "TARGET_NOT_AUTHORIZED";                        // Target address not authorized to call this method.
+    string constant internal TARGET_ALREADY_AUTHORIZED = "TARGET_ALREADY_AUTHORIZED";                // Target address must not already be authorized.
+    string constant internal INDEX_OUT_OF_BOUNDS = "INDEX_OUT_OF_BOUNDS";                            // Specified array index is out of bounds.
+    string constant internal AUTHORIZED_ADDRESS_MISMATCH = "AUTHORIZED_ADDRESS_MISMATCH";            // Address at index does not match given target address.
```
fabioberger: These aren't used in this PR. Why were they made internal?

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
```diff
@@ -0,0 +1,200 @@
...
+        // Hash 0x transaction
+        bytes32 transactionHash = getTransactionHash(transaction);
+
+        // Create empty list of approval signers
+        address[] memory approvalSignerAddresses = new address[](0);
+
+        uint256 signaturesLength = approvalSignatures.length;
+        for (uint256 i = 0; i < signaturesLength; i++) {
+            // Create approval message
+            uint256 currentApprovalExpirationTimeseconds = approvalExpirationTimeSeconds[i];
```
fabioberger:
```suggestion
            uint256 currentApprovalExpirationTimeSeconds = approvalExpirationTimeSeconds[i];
```

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
```diff
@@ -0,0 +1,200 @@
...
+            address approvalSignerAddress = getSignerAddress(approvalHash, approvalSignatures[i]);
+    
+            // Add approval signer to list of signers
+            approvalSignerAddresses = approvalSignerAddresses.append(approvalSignerAddress);
+        }
+        
+        uint256 ordersLength = orders.length;
+        for (uint256 i = 0; i < ordersLength; i++) {
+            // Do not check approval if the order's senderAddress is null
+            if (orders[i].senderAddress == address(0)) {
```
fabioberger: Shouldn't this check if the `senderAddress` _is_ the TEC contract? There are three cases:
1. SenderAddress is `null` -> fill however you want
2. SenderAddress is TECExtension -> Must fill through TEC contract
3. SenderAddress is otherExtension -> Which might or might not fill all orders via the TEC extension contract

We can't enforce 3) will always send orders through the TEC contract, so can we even support that use-case?

contracts/tec/package.json
```diff
@@ -0,0 +1,90 @@
...
+        "@0x/contracts-erc721": "1.0.2",
+        "@0x/contracts-utils": "^2.0.4",
+        "@0x/order-utils": "^6.1.0",
+        "@0x/types": "^2.0.2",
+        "@0x/typescript-typings": "^4.0.0",
+        "@0x/utils": "^4.1.0",
+        "@0x/web3-wrapper": "^5.0.0",
+        "ethereumjs-util": "^5.1.1",
+        "ethereum-types": "^2.0.0",
+        "lodash": "^4.17.11"
```
fabioberger: Does this package really need all these deps?

contracts/exchange-libs/contracts/src/LibAssetProxyErrors.sol
abandeali1: I believe this was done in #1604. Basically it was annoying that all of these were part of the contract's ABI.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
abandeali1: Basically, this contract lets all of the standard 0x transaction failures fall back on the Exchange contract. Any `senderAddress` that isn't null or the TEC contract address is guaranteed to fail anyways, and it increases gas costs to check if it is the address of the TEC contract.

contracts/tec/package.json
abandeali1: I'll double check, might not need a few.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
```diff
@@ -0,0 +1,200 @@
...
+        view
+    {
+        // Hash 0x transaction
+        bytes32 transactionHash = getTransactionHash(transaction);
+
+        // Create empty list of approval signers
+        address[] memory approvalSignerAddresses = new address[](0);
+
+        uint256 signaturesLength = approvalSignatures.length;
+        for (uint256 i = 0; i < signaturesLength; i++) {
```
hysz: Should we assert that `approvalSignatures.length == approvalExpirationTimeSeconds.length` before iterating in lockstep? Should error out anyway, but I think it'll throw and waste the gas + won't get a helpful error message.

contracts/tec/contracts/src/TEC.sol
```diff
@@ -0,0 +1,39 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity 0.5.3;
```
hysz: Should this be pegged? I think all the others are `^0.5.3`

contracts/tec/contracts/src/libs/LibZeroExTransaction.sol
```diff
@@ -0,0 +1,91 @@
...
+pragma solidity ^0.5.3;
+
+import "./LibEIP712Domain.sol";
+
+
+contract LibZeroExTransaction is
+    LibEIP712Domain
+{
+    // Hash for the EIP712 0x transaction schema
+    bytes32 constant internal EIP712_ZEROEX_TRANSACTION_SCHEMA_HASH = keccak256(abi.encodePacked(
```
hysz: We could save a few hundred gas per call by hardcoding this; the hash (counter-intuitively) gets executed on each call - even though it’s `constant`.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
```diff
@@ -0,0 +1,200 @@
...
+            (LibOrder.Order memory leftOrder, LibOrder.Order memory rightOrder) = abi.decode(
+                data.slice(4, data.length),
+                (LibOrder.Order, LibOrder.Order)
+            );
+
+            // Create array of orders
+            orders = new LibOrder.Order[](2);
+            orders[0] = leftOrder;
+            orders[1] = rightOrder;
+        }
```
hysz: Should we handle the case of an unrecognized selector? Handling this case explicitly would future-proof this code, by ensuring it can’t call functions that it doesn’t know how to handle.

contracts/tec/contracts/src/MixinSignatureValidator.sol
```diff
@@ -0,0 +1,107 @@
...
+
+contract MixinSignatureValidator is
+    MSignatureValidator
+{
+    using LibBytes for bytes;
+
+    /// @dev Recovers the address of a signer given a hash and signature.
+    /// @param hash Any 32 byte hash.
+    /// @param signature Proof that the hash has been signed by signer.
+    function getSignerAddress(bytes32 hash, bytes memory signature)
```
hysz: Is it possible to share the `getSignerAddress` used by the Exchange? Other extensions could also make use of it as a generic lib.

contracts/tec/test/mixins.ts
```diff
@@ -0,0 +1,714 @@
...
+                        [approval.signature],
+                        { from: approvalSignerAddress2 },
+                    ),
+                    RevertReason.InvalidSender,
+                );
+            });
+        }
+    });
+    describe('Batch order approvals', () => {
+        for (const fnName of [
```
hysz: Nice 😎

contracts/tec/contracts/src/TEC.sol
abandeali1: All of the other top level contracts are pegged, but dependencies are not because they may be used elsewhere.

contracts/tec/contracts/src/libs/LibZeroExTransaction.sol
abandeali1: Added a task to do this later, there are definitely a good amount of optimizations to be done.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
abandeali1: As is, an unknown selector will return an empty `Order` array. If that isn't a valid Exchange selector, then the call will fail when it hits the Exchange. I think this is the simplest way of allowing all non-fill functions to be automatically approved, rather than hard coding in the allowed functions.

The default behavior of this contract is to not duplicate any validation that already occurs in the Exchange contract in order to reduce cost/complexity.

contracts/tec/contracts/src/MixinSignatureValidator.sol
abandeali1: The one on the Exchange contract returns a `bool` rather than an `address` unfortunately.

contracts/tec/contracts/src/MixinTECApprovalVerifier.sol
hysz: Yeah so the scenario I was thinking of is - suppose this contract were used with a future version of the exchange, which includes a new set of functions that must be validated by a TEC. These new functions would be automatically approved, when they should be validated. 

While we specify the exchange in the constructor, there's no hard link between this version of the TEC and a compatible version of the exchange.

One way to mitigate this is to have a whitelist of known functions ([this](https://github.com/0xProject/0x-monorepo/blob/development/contracts/extensions/contracts/src/BalanceThresholdFilter/MixinBalanceThresholdFilterCore.sol#L94) list should suffice). Alternatively, we could include a call in the constructor to assert that `Exchange.VERSION` is `2.0.0`, although this would be less robust.

contracts/tec/contracts/src/MixinSignatureValidator.sol
hysz: Ah okay, maybe for v3 we could refactor this. I'll make a note of it.

PR #1553, opened 2019-01-27
[WIP] added USDC to 0x Portal wallet

dave4506:
## Description

Added USDC as a default asset in the wallet experience.

TODO: add USDC into the token registry backend server. (Temporarily added the USDC asset metadata in `backend_client.ts` 

## Checklist:


packages/website/ts/components/wallet/wallet.tsx
```diff
@@ -339,13 +339,15 @@ export class Wallet extends React.Component<WalletProps, WalletState> {
             tokenState.price,
             !tokenState.isLoaded,
         );
+        // console.log("primaryText", primaryText, "secondaryText", secondaryText)
```
fragosti: nit: should remove.

packages/website/ts/utils/backend_client.ts
```diff
@@ -45,6 +45,13 @@ export const backendClient = {
     },
     async getTokenInfosAsync(): Promise<WebsiteBackendTokenInfo[]> {
         const result = await fetchUtils.requestAsync(utils.getBackendBaseUrl(), TOKENS_ENDPOINT);
+        // TEMPORARY: Until backend is updated with USDC, this appends the metadata for USDC
+        result.push({
+            address: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48',
+            decimals: 6,
+            name: 'US Dollar Coin',
+            symbol: 'USDC',
+        });
```
fragosti: Does this need to be in here to work, or can we not have this and just add a row to the DB?

dave4506: This is a temporary fix, its better for it to be in the DB, once its in the DB this can be removed

dave4506: Which specific codebase/where is this DB hosted?

PR #1554, opened 2019-01-28
[WIP] Create OrderHelper extension to expose getOrderHash

xianny:
## Description

Creates OrderHelper extension smart contract to expose `getOrderHash` as a public method.

This is my first Solidity contrib so please be careful :)

### Questions
1. There's 2.0 and 3.0 in `CHANGELOG.json`, both without timestamps. Which should I add this to?
2. Is this the right place for it? Already discussed with @fabioberger but still open to feedback

## Testing instructions

Unit test with `yarn test`

## Types of changes

## Checklist:

### DIFF
--- a/contracts/extensions/contracts/src/OrderHelper/OrderHelper.sol
+++ b/contracts/extensions/contracts/src/OrderHelper/OrderHelper.sol
@@ -1,0 +1,38 @@
+/*
+
+  Copyright 2019 ZeroEx Intl.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity 0.4.24;
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
+
+contract OrderHelper is
+    LibOrder
+{
+    // @dev Calculates Keccak-256 hash of the order.
+    /// @param order The order structure.
+    /// @return Keccak-256 EIP712 hash of the order.
+    function publicGetOrderHash(Order order)
+        public
+        view
+        returns (bytes32 orderHash)
+    {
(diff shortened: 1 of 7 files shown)


contracts/extensions/contracts/src/OrderHelper/OrderHelper.sol
```diff
@@ -0,0 +1,39 @@
...
+{
+    // @dev Calculates Keccak-256 hash of the order.
+    /// @param order The order structure.
+    /// @return Keccak-256 EIP712 hash of the order.
+    function publicGetOrderHash(Order order)
+        public
+        view
+        returns (bytes32 orderHash)
+    {
+        orderHash = hashEIP712Message(hashOrder(order));
```
abandeali1: This will actually return the incorrect value because `LibOrder` inherits from `LibEIP712`, which uses the address of the deployed contract in the hash calculations. Since the address of this contract is different than the address of the Exchange, the hashes will be different.

contracts/extensions/test/order_helper.ts
```diff
@@ -0,0 +1,66 @@
...
+        await blockchainLifecycle.revertAsync();
+    });
+    before(async () => {
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        const makerAddress = accounts[0];
+        contract = await OrderHelperContract.deployFrom0xArtifactAsync(artifacts.OrderHelper, provider, txDefaults);
+
+        const defaultOrderParams = {
+            ...constants.STATIC_ORDER_PARAMS,
+            exchangeAddress: contract.address,
```
abandeali1: Tests pass because of this line, but in production this contract's address will be different than the Exchange contract.

fabioberger:
@abandeali1 If you look at the V3.0 scoping doc, you will notice that `getOrderInfo` will be removed. Currently `getOrderInfo` returns `orderHash`, `orderStatus` and `orderTakerAssetFilledAmount`. Once we have better revert reasons, there will be no need for `orderStatus`. `orderTakerAssetFilledAmount` can be fetched with a "call" to the mapping. We could add another helper method here that fetches both the `orderHash` and the `filledTakerAssetAmount`, which could be helpful, but one doesn't always want both of those together.

xianny:
@fabioberger @abandeali1 how about we expose `getOrderInfo` in `MixinExchangeCore`? It seems the purpose of `getOrderHash` is to help interact with `Exchange` so we want to make sure it's always the same hashing method. I think the cleanest way is to point to the same implementation instead of re-implementing separately. 

From just reading code it seems like `MixinExchangeCore` would be the right place to put it. WDYT?

xianny:
This is super stale now, my bad. @abandeali1 , should I still update and ask for re-review?

xianny:
discussed on Slack, won't merge this.

Issue #1531, opened 2019-01-18
sol-trace error a bit confusing


hoytech:
Basically I was running sol-coverage and I messed up and was deploying a different bytecode than was compiled with sol-compiler. I saw some errors like `Unknown contract creation transaction` and `Transaction to an unknown address: ...` which of course I just ignored.

Took me a while to figure out why it couldn't match the bytecode because those messages above took place after the contracts were successfully loaded by the `SolCompilerArtifactAdapter` (or so I thought).

Anyway I would've figured this out way faster if `getContractDataIfExists` warned that it couldn't find a bytecode.

hoytech:
After thinking about this one a bit more, I don't think there's really anything you can do, it was my screw-up. Maybe changing those "Transaction to an unknown address" to say *why* the address is unknown? Ideally it would say "sol-coverage: Was unable to find matching bytecode for contract creation 0xXYZ, ignoring..." or something.

LogvinovLeon:
Agree, I would make an error message better and more visible. Ideally, it should never happen when you develop unless you call some contracts you don't have the artifacts for (which is a marginal use-case).

PR #1558, opened 2019-01-30
Improve error messages when unable to find bytecode

LogvinovLeon:
## Description

Before:
`Unknown contract creation transaction`
After:
screen shot 2019-01-30 at 2 00 50 pm
Before:
`Transaction to an unknown address: 0x1dc4c1cefef38a777b15aa20260a54e584b16c48`
After:
screen shot 2019-01-30 at 2 00 02 pm

Fixes https://github.com/0xProject/0x-monorepo/issues/1531

## Types of changes

## Checklist:


packages/sol-trace/src/revert_trace_subprovider.ts
```diff
@@ -71,9 +72,21 @@ export class RevertTraceSubprovider extends TraceCollectionSubprovider {
             const bytecode = await this._web3Wrapper.getContractCodeAsync(evmCallStackEntry.address);
             const contractData = utils.getContractDataIfExists(this._contractsData, bytecode);
             if (_.isUndefined(contractData)) {
+                const shortenHex = (hex: string) => {
+                    /**
+                     * Length chooses so that both error messages are of the same length
```
fabioberger:
```suggestion
                     * Length choosen so that both error messages are of the same length
```

PR #1559, opened 2019-01-30
[pipeline] Fix missing DEX Trades from Bloxy

fragosti:
## Description

This removes the assumption from our data pipeline that trades (Fills in our case) have unique transaction hashes. Removing this constraint meant that we needed a new way to uniquely identify a DEX trade. Thankfully Bloxy was willing to (quickly) expose a `trade_index` field through the API, and guarantee that `(trade_index, tx_hash)` pairs are unique. I wrote a migration to add `trade_index` as a primary column to DEX trades, so the composite primary key is now `(source_url, tx_hash, trade_index)`. It'll be useful to have `source_url` in there if we ever want to integrate a source that isn't Bloxy and still easily guarantee row uniqueness. 

Before:
screen shot 2019-01-29 at 1 55 20 pm

After:
screen shot 2019-01-30 at 10 25 01 am

## Checklist:

### DIFF
--- a/packages/pipeline/migrations/1548809952793-AllowDuplicateTxHashesInDexTrades.ts
+++ b/packages/pipeline/migrations/1548809952793-AllowDuplicateTxHashesInDexTrades.ts
@@ -1,0 +1,27 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const DEX_TRADES_TABLE_NAME = 'raw.dex_trades';
+
+export class AllowDuplicateTxHashesInDexTrades1548809952793 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        const dexTradesTable = await queryRunner.getTable(DEX_TRADES_TABLE_NAME);
+        if (dexTradesTable) {
+            // Composite key goes from (source_url, tx_hash) to (trade_index, source_url, tx_hash)
+            await queryRunner.addColumn(
+                DEX_TRADES_TABLE_NAME,
+                new TableColumn({
+                    name: 'trade_index',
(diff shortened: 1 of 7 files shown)


packages/pipeline/migrations/1548809952793-AllowDuplicateTxHashesInDexTrades.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class AllowDuplicateTxHashesInDexTrades1548809952793 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        const dexTradesTable = await queryRunner.getTable(DEX_TRADES_TABLE_NAME);
+        if (dexTradesTable) {
+            // Composite key goes from (source_url, tx_hash) to (trade_index, source_url, tx_hash)
+            await queryRunner.addColumn(
+                DEX_TRADES_TABLE_NAME,
+                new TableColumn({
+                    name: 'trade_index',
```
xianny: I think we should prefix this as `bloxy_trade_index` instead.

Also worried about what happens if we want to use other sources that don't have a trade index. Maybe we should start thinking of this table as containing bloxy info only. The primary key constraint means this column can't be null.

fragosti: I feel like a `trade_index`-like thing will be relevant for every source. We just need (source_url, tx_hash, trade_index) to be unique. If we start getting trades from other sources, we will also need `trade_index` for uniqueness.

fragosti:
Cool this worked, I tried it on QA. Going to merge.

PR #1569, opened 2019-01-31
0x Transaction Data Decoder

hysz:
## Description

A convenient way to decode transaction data for 0x contracts.

Usage:
```
# Decode calldata using contract wrappers
const decodedCalldata = contractWrappers.getAbiDecoder().decodeCalldataOrThrow(calldata);

# Custom decoder
const customDecoder = new AbiDecoder([]);
customDecoder.addABI(myCustomAbi_1, myContractName_1);
customDecoder.addABI(myCustomAbi_N, myContractName_N);
const decodedCalldata = customDecoder.decodeCalldataOrThrow(calldata);
```

Todo: 
Rebased on top of #1592 after its been merged, then create changelogs.

## Checklist:


packages/contract-wrappers/test/zeroex_transaction_decoder_test.ts
```diff
@@ -0,0 +1,190 @@
...
+import { chaiSetup } from './utils/chai_setup';
+import { migrateOnceAsync } from './utils/migrate';
+import { provider, web3Wrapper } from './utils/web3_wrapper';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
+
+describe.only('ZeroExTransactionDecoder', () => {
```
dekz: only

packages/contract-wrappers/src/utils/zeroex_transaction_decoder.ts
```diff
@@ -0,0 +1,88 @@
...
+     * @return singleton instance.
+     */
+    private static _getInstance(): ZeroExTransactionDecoder {
+        if (!ZeroExTransactionDecoder._instance) {
+            ZeroExTransactionDecoder._instance = new ZeroExTransactionDecoder();
+        }
+        return ZeroExTransactionDecoder._instance;
+    }
+    /**
+     * Adds all known contract ABI's defined by the @0x/Artifacts package, along with known 0x
```
dekz: `@0x/contract-artifacts`

packages/contract-wrappers/src/utils/zeroex_transaction_decoder.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import {
+    DeployedContractInfo,
+    DeployedContractInfoByName,
+    TransactionData,
+    TransactionDecoder,
+    TransactionProperties,
+} from '@0x/utils';
+
+export class ZeroExTransactionDecoder extends TransactionDecoder {
```
dekz: I think this should also be exposed by default on the contractWrappers instance. 

I think the name is a little confusing as well.  It doesn't decode transactions but abi encoded data.
`ZeroExCalldataDecoder`? What are your thoughts on this?

packages/utils/src/abi_decoder.ts
```diff
@@ -37,10 +58,10 @@ export class AbiDecoder {
     public tryToDecodeLogOrNoop<ArgsType extends DecodedLogArgs>(log: LogEntry): LogWithDecodedArgs<ArgsType> | RawLog {
         const methodId = log.topics[0];
         const numIndexedArgs = log.topics.length - 1;
-        if (_.isUndefined(this._methodIds[methodId]) || _.isUndefined(this._methodIds[methodId][numIndexedArgs])) {
+        if (_.isUndefined(this._eventIds[methodId]) || _.isUndefined(this._eventIds[methodId][numIndexedArgs])) {
             return log;
         }
-        const event = this._methodIds[methodId][numIndexedArgs];
+        const event = this._eventIds[methodId][numIndexedArgs];
         const ethersInterface = new ethers.utils.Interface([event]);
```
dekz: Should we remove the `ethers` dependency and use `AbiDecoder` instead? Maybe a TODO for another PR?

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+        };
+        return decodedCalldata;
+    }
+    /**
+     * Adds a set of ABI definitions, after which transaction data targeting these ABI's can be decoded.
+     * Additional properties can be included to disambiguate similar ABI's. For example, if two functions
+     * have the same signature but different parameter names, then their ABI definitions can be disambiguated
+     * by specifying a contract name.
+     * @param abiDefinitions ABI definitions for a given contract.
+     * @param contractName Name of contract that encapsulates the ABI definitions (optional).
```
dekz: Maybe a note that this is only useful for call data decoding (not logs)

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+        const functionArguments = functionInfo.abiEncoder.decode(calldata);
+        const decodedCalldata = {
+            functionName,
+            functionSignature,
+            functionArguments,
+        };
+        return decodedCalldata;
+    }
+    /**
+     * Adds a set of ABI definitions, after which transaction data targeting these ABI's can be decoded.
```
dekz:
```suggestion
     * Adds a set of ABI definitions, after which transaction data and logs targeting these ABI's can be decoded.
```

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+     * @return Decoded transaction data. Includes: function name and signature, along with the decoded arguments.
      */
-    public addABI(abiArray: AbiDefinition[]): void {
+    public tryDecodeCalldata(calldata: string, contractName?: string): TransactionData {
+        const functionSelector = AbiDecoder._getFunctionSelector(calldata);
+        const candidateFunctionInfos = this._functionInfoBySelector[functionSelector];
+        if (_.isUndefined(candidateFunctionInfos)) {
+            throw new Error(`No functions registered for selector '${functionSelector}'`);
+        }
+        const functionInfo = _.find(candidateFunctionInfos, txDecoder => {
```
LogvinovLeon: If we have two candidates. One with undefined contractName and one with defined matching contract name - it makes sense to match against the second one. This code will match against the rightmost one.

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+     * have the same signature but different parameter names, then their ABI definitions can be disambiguated
+     * by specifying a contract name.
+     * @param abiDefinitions ABI definitions for a given contract.
+     * @param contractName Name of contract that encapsulates the ABI definitions (optional).
+     */
+    public addABI(abiArray: AbiDefinition[], contractName?: string): void {
         if (_.isUndefined(abiArray)) {
             return;
         }
         const ethersInterface = new ethers.utils.Interface(abiArray);
```
LogvinovLeon: Should we drop ethers here?

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
-                const eventAbi = abi as EventAbi;
-                const topic = ethersInterface.events[eventAbi.name].topic;
-                const numIndexedArgs = _.reduce(eventAbi.inputs, (sum, input) => (input.indexed ? sum + 1 : sum), 0);
-                this._methodIds[topic] = {
-                    ...this._methodIds[topic],
-                    [numIndexedArgs]: eventAbi,
-                };
+            switch (abi.type) {
+                case AbiType.Event:
+                    // tslint:disable no-unnecessary-type-assertion
```
LogvinovLeon: Please use `// tslint:disable-line` or `// tslint:disable-next-line`.

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+
+                default:
+                    // ignore other types
+                    break;
             }
         });
     }
+    private _addEventABI(abi: EventAbi, ethersInterface: ethers.utils.Interface): void {
+        // tslint:disable-next-line:no-unnecessary-type-assertion
+        const eventAbi = abi as EventAbi;
```
LogvinovLeon: It seems like here we don't actually need that type assertion

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
+        const topic = ethersInterface.events[eventAbi.name].topic;
+        const numIndexedArgs = _.reduce(eventAbi.inputs, (sum, input) => (input.indexed ? sum + 1 : sum), 0);
```
LogvinovLeon:
```suggestion
        const numIndexedArgs = _.filter(eventAbi.inputs, input => input.indexed).length;
```
Seems simpler

packages/utils/src/address_utils.ts
```diff
@@ -43,4 +46,25 @@ export const addressUtils = {
     padZeros(address: string): string {
         return addHexPrefix(_.padStart(stripHexPrefix(address), ADDRESS_LENGTH, '0'));
     },
+    /**
+     * Generates a pseudo-random 256-bit salt.
+     * The salt can be included in a 0x order, ensuring that the order generates a unique orderHash
+     * and will not collide with other outstanding orders that are identical in all other parameters.
+     * @return  A pseudo-random 256-bit number that can be used as a salt.
+     */
+    generatePseudoRandomSalt(): BigNumber {
```
LogvinovLeon: We have exactly that code in order utils. Can we dedupe?

packages/utils/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface FunctionInfo {
+    functionSignature: string;
+    contractName?: string;
+    contractAddress?: string;
+    networkId?: number;
+    abiEncoder?: AbiEncoder.Method;
+}
+
+export interface FunctionInfoBySelector {
```
LogvinovLeon: `SelectorToFunctionInfo`

packages/0x.js/src/index.ts
```diff
@@ -68,7 +68,7 @@ export {
     MetamaskSubprovider,
 } from '@0x/subproviders';
 
-export { AbiDecoder } from '@0x/utils';
+export { AbiDecoder, TransactionData } from '@0x/utils';
```
fabioberger: @hysz kinda tangential to your PR but mind also exporting `transactionHashUtils` from `0x.js`? We forgot to do so.

packages/contract-wrappers/test/calldata_decoder_test.ts
```diff
@@ -0,0 +1,123 @@
...
+            makerAssetData: assetDataUtils.encodeERC20AssetData(defaultERC20MakerAssetAddress),
+            makerAssetAmount: new BigNumber(10),
+            takerAddress: '0x0000000000000000000000000000000000000000',
+            takerAssetData: assetDataUtils.encodeERC20AssetData(defaultERC20MakerAssetAddress),
+            takerAssetAmount: new BigNumber(1),
+            feeRecipientAddress,
+            makerFee: new BigNumber(0),
+            takerFee: new BigNumber(0),
+            senderAddress: '0x0000000000000000000000000000000000000000',
+            expirationTimeSeconds: new BigNumber(1549498915),
```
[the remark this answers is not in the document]
fabioberger: At the very least, add a comment about it.

packages/utils/src/abi_decoder.ts
```diff
@@ -6,28 +6,49 @@ import {
...
+import { FunctionInfoBySelector, TransactionData } from './types';
 
 /**
  * AbiDecoder allows you to decode event logs given a set of supplied contract ABI's. It takes the contract's event
  * signature from the ABI and attempts to decode the logs using it.
  */
 export class AbiDecoder {
-    private readonly _methodIds: { [signatureHash: string]: { [numIndexedArgs: number]: EventAbi } } = {};
+    private readonly _eventIds: { [signatureHash: string]: { [numIndexedArgs: number]: EventAbi } } = {};
+    private readonly _functionInfoBySelector: FunctionInfoBySelector = {};
```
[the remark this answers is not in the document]
fabioberger: ☝️  This is a new preferred convention. To use `to` instead of `by`, since the order in the name will then match the map order (i.e key -> value)

packages/utils/src/abi_decoder.ts
fabioberger: Let's make this a TODO for another PR. I'm blocked by getting this merged.

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
     /**
-     * Add additional ABI definitions to the AbiDecoder
-     * @param abiArray An array of ABI definitions to add to the AbiDecoder
+     * Decodes transaction data for a known ABI.
+     * @param calldata hex-encoded transaction data.
+     * @param contractName used to disambiguate similar ABI's (optional).
+     * @return Decoded transaction data. Includes: function name and signature, along with the decoded arguments.
      */
-    public addABI(abiArray: AbiDefinition[]): void {
+    public tryDecodeCalldata(calldata: string, contractName?: string): TransactionData {
```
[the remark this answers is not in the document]
fabioberger: I'd vote for renaming it `decodeCallDataOrThrow` and keeping the behavior the same.

packages/utils/src/abi_decoder.ts
fabioberger: `tslint:disable-next-line:no-unnecessary-type-assertion`

packages/utils/src/abi_decoder.ts
```diff
@@ -89,25 +110,94 @@ export class AbiDecoder {
...
-                const numIndexedArgs = _.reduce(eventAbi.inputs, (sum, input) => (input.indexed ? sum + 1 : sum), 0);
-                this._methodIds[topic] = {
-                    ...this._methodIds[topic],
-                    [numIndexedArgs]: eventAbi,
-                };
+            switch (abi.type) {
+                case AbiType.Event:
+                    // tslint:disable no-unnecessary-type-assertion
+                    this._addEventABI(abi as EventAbi, ethersInterface);
+                    // tslint:enable no-unnecessary-type-assertion
```
fabioberger: Can get rid of this.

packages/contract-wrappers/test/calldata_decoder_test.ts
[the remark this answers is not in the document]
hysz: Values are random. Added a note.

packages/utils/src/abi_decoder.ts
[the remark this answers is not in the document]
hysz: The other method in this class is `tryToDecodeLogOrNoop` so I was aiming to match the local naming convention. I'm fine with updating to `decodeCallDataOrThrow`.

packages/utils/src/abi_decoder.ts
hysz: Yeah there's some additional work needed to replace interfaces. We have a task in the protocol backlog to finish replacing ethers throughout the monorepo.

packages/utils/src/abi_decoder.ts
hysz: See above. This will be in a subsequent PR.

packages/utils/src/abi_decoder.ts
hysz: Much nicer 🙏

packages/utils/src/abi_decoder.ts
hysz: Not sure I follow. At present, if `contractName` is specified by the caller, then this will return the candidate that was registered to this name. If no `contractName` is specified by the caller then the it will return the first instance.

PR #1575, opened 2019-02-01
Contracts generator

LogvinovLeon:
## Description

This PR has quite a bit of stuff.
It's motivation was to reduce the boilerplate in smart contracts packages and make it easier to refactor them and to work with a lot of small packages.

Before that PR if you wanted to move smart contract `A` from `packageA` to `packageB` your course of action would be:
* Remove `A` from `packageA/package.json` `config.abis`
* Add `A` to `packageB/package.json` `config.abis`
* Remove `A` from `packageA/wrappers.ts`
* Add `A` to `packageB/wrappers.ts`
* Remove `A` from `packageA/artifacts.ts`
* Add `A` to `packageB/artifacts.ts`
* Remove `A` from `packageA/tsconfig.json` `files`
* Add `A` to `packageB/tsconfig.json` `files`
* Remove `A` from `packageA/compiler.json` `contracts`
* Add `A` to `packageB/compiler.json` `contracts`
* Finally move the smart contract itself and update imports.

Now you can just move the contract, update compiler.json's and run `contracts:gen` in both packages.
You can also run `contracts:gen` globally which will regenerate all. it's more convenient.

So we now have this new tool `contracts-gen`. it's a dependency of each smart contracts package and helps automate routine tasks and generate boilerplate.
Please refer to it's README for more details.

One more thing. Our thinking about smart contracts artifacts changed. They are no longer considered path of public interface for other smart contracts packages. Now each package needs to specify it's dependencies in `compiler.json`. Smart contracts packages depend on each others code. Not on the artifacts compiled with a specific version of solidity.

Because of that you now have access to all the artifacts you need in your package. You no longer need to do:
```
{
...artifacts,
...erc20Artifacts,
...exchangeArtifacts,
}
```
It all now lives in your local artifacts directory.

Artifacts still will get published for external UIs and stuff but smart contracts packages should ignore them and only import the source code.

One more thing. Because now `compiler.json` includes both local contracts and dependencies - we've started referencing everything by relative path, not name. You can still put a name there and run `contracts:gen` after. It will resolve your name, regenerate boilerplate and sort the list. Sol-compiler name resolver is considered deprecated and will be removed from sol-compiler.

Overall - read through the PR. Specifically the `contracts:gen` part.
Obligatory for the Protocol team. And hit me up with any questions. Nothing set in stone, but hope you like the direction.

There is still one WIP to this PR.

Sol-compiler has `contractName` everywhere. Now in all of those places, it should be `contractNameOrContractPath`. I'm not sure if I should remove the referencing by name functionality now. It will be much cleaner, but slightly less beginner-friendly. I can create something like `PathResolver.getAll()`.

At the same time - I'm slightly against that. It's normal that you should specify the entry point for compilation and I'm hesitant that the use-case - each file is the entry point has a future except for one-file projects.

## Types of changes

## Checklist:


contracts/asset-proxy/package.json
```diff
@@ -29,10 +29,12 @@
         "profiler:report:html": "istanbul report html && open coverage/index.html",
         "coverage:report:lcov": "istanbul report lcov",
         "test:circleci": "yarn test",
+        "contracts:gen": "contracts-gen",
         "lint-contracts": "solhint -c ../.solhint.json contracts/**/**/**/**/*.sol"
     },
     "config": {
-        "abis": "generated-artifacts/@(ERC20Proxy|ERC721Proxy|IAssetData|IAssetProxy|IAuthorizable|MixinAuthorizable|MultiAssetProxy).json"
+        "abis": "./generated-artifacts/@(DummyERC20Token|DummyERC721Receiver|DummyERC721Token|DummyMultipleReturnERC20Token|DummyNoReturnERC20Token|ERC20Proxy|ERC721Proxy|IAssetData|IAssetProxy|IAuthorizable|MixinAuthorizable|MultiAssetProxy).json",
+        "abis:comment": "This list is auto-generated by contracts-gen. Don't edit manually."
```
fabioberger: Let's put this _above_ the "abis" entry. Comments are always _above_ the line they comment on in our code. Less likely someone doesn't see it.

contracts/asset-proxy/src/artifacts.ts
```diff
@@ -0,0 +1,29 @@
+// This file is auto-generated by contracts-gen. Don't edit manually.
```
fabioberger:
```suggestion
/*
 * -----------------------------------------------------------------------------
 * Warning: This file is auto-generated by `contracts-gen`. Don't edit manually.
 * -----------------------------------------------------------------------------
 */
```

Can we do this so that it's physically bigger and less likely someone doesn't see it?

contracts/asset-proxy/test/proxies.ts
```diff
@@ -30,6 +18,12 @@ import * as _ from 'lodash';
 
 import {
     artifacts,
+    DummyERC20TokenContract,
+    DummyERC20TokenTransferEventArgs,
+    DummyERC721ReceiverContract,
+    DummyERC721TokenContract,
+    DummyMultipleReturnERC20TokenContract,
+    DummyNoReturnERC20TokenContract,
```
fabioberger: Can we not import from `../src`? Can we import from the exact file? It makes it harder to grep for when a directory is renamed...

packages/contracts-gen/README.md
```diff
@@ -0,0 +1,76 @@
...
+contracts-gen
+```
+
+You should run this tool after each time you move your contracts around to regenerate boilerplate code and configs.
+
+## What can it generate
+
+This tool does the following:
+
+-   Reads your `compiler.json`. Specifically list of smart contracts.
```
fabioberger:
```suggestion
-   Reads your `compiler.json`. Specifically the list of smart contracts.
```

packages/contracts-gen/README.md
```diff
@@ -0,0 +1,76 @@
...
+-   Creates `wrapper.ts` file which exports all contract wrappers.
+-   Creates `artifacts.ts` file which exports all contract artifacts.
+-   Generates list of JSON artifact files in `tsconfig.json`
+-   Generates a glob for abi-gen in `package.json`
+
+On top of that - if your `compiler.json` has contracts referenced just by name - it will resolve the name to relative path and put it there.
+It also sorts all the lists in it's output leading to smaller and cleaner diffs.
```
fabioberger:
```suggestion
It also sorts the list of smart contracs leading to smaller and cleaner diffs.
```

packages/contracts-gen/src/contracts-gen.ts
```diff
@@ -0,0 +1,171 @@
...
+        const importPath = path.join('..', artifactsDir, `${contractName}.json`);
+        return `import * as ${contractName} from '${importPath}';`;
+    });
+    const sortedImports = _.sortBy(imports);
+    const artifacts = _.map(contracts, contract => {
+        const contractName = path.basename(contract, SOLIDITY_EXTENSION);
+        if (contractName === 'ZRXToken') {
+            // HACK(albrow): "as any" hack still required here because ZRXToken does not
+            // conform to the v2 artifact type.
+            return `${contractName}: (${contractName} as any) as ContractArtifact,`;
```
fabioberger: Is this HACK still required? My understanding is we are regenerating all artifacts so the ZRXToken would conform to the new artifact format.

packages/contracts-gen/src/index.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * This module is a CLI tool. As soon as you run it - it starts doing stuff.
+ * At the same time - our installation tests assume that you can import package without causing side effects.
+ * That's why our main entry point it empty. No side effects. But out secondary entry poing - contracts-gen.ts is a CLI tool and starts running as soon as you import/run it.
```
fabioberger:
```suggestion
 * That's why our main entry point it empty. No side effects. But our secondary entry point - contracts-gen.ts is a CLI tool and starts running as soon as you import/run it.
```

packages/monorepo-scripts/package.json
```diff
@@ -8,6 +8,9 @@
     "description": "Helper scripts for the monorepo",
     "main": "lib/index.js",
     "types": "lib/index.d.ts",
+    "bin": {
+        "contracts-gen": "bin/contracts-gen.js"
+    },
```
fabioberger: Remove this and deps added since you moved these out to a separate package.

packages/monorepo-scripts/src/deps_versions.ts
```diff
@@ -21,11 +21,10 @@ const PACKAGE_JSON_GLOB = '../*/package.json';
 
 // tslint:disable:no-unused-variable
 function getDependencies(path: string): Dependencies {
-    const file = fs.readFileSync(path).toString();
-    const parsed = JSON.parse(file);
+    const packageJSON = utils.readJSONFile<PackageJSON>(path);
```
fabioberger: Super cool 🤓

contracts/asset-proxy/package.json
LogvinovLeon: I'm affraid I can't do that with a current architecture. When I contruct an object - the order doesn't matter.
If I put it like that:
```
packageJSON = {
        ...packageJSON,
        config: {
            ...packageJSON.config,
            'abis:comment': AUTO_GENERATED_BANNER_FOR_LISTS,
            abis: `${artifactsDir}/@(${sortedContractNames.join('|')}).json`,
        },
    };
```
Or like that:
```
packageJSON = {
        ...packageJSON,
        config: {
            ...packageJSON.config,
            abis: `${artifactsDir}/@(${sortedContractNames.join('|')}).json`,
            'abis:comment': AUTO_GENERATED_BANNER_FOR_LISTS,
        },
    };
```
The result is the same. The order of the keys in an object in JS is not defined and I can't influence it.

contracts/asset-proxy/test/proxies.ts
LogvinovLeon: Discussed offline

packages/contracts-gen/README.md
LogvinovLeon: Which list? It sorts multiple lists of different nature. One is the list of smart contract names. One is the list of imports for wrappers. One is the list of imports for artifacts. One is the artifacts JSON files list. One is the list of object keys for artifacts.

packages/contracts-gen/src/contracts-gen.ts
LogvinovLeon: We're using the old version of Solidity and it still does not confirm.

LogvinovLeon: ABI types changed between solidity versions

packages/contracts-gen/README.md
fabioberger: Ah, didn't catch that. Ignore.

PR #1580, opened 2019-02-04
WIP: Fix preSigned signature bug in SignatureUtils

fabioberger:
## Description

Fixes a bug causing pre-signed orders to failed signature validation in our libraries. The bug originated from us passing the `signerAddress` instead of the `exchangeAddress` into the `ExchangeContract` constructor.

## Checklist:

### DIFF
--- a/packages/order-utils/test/signature_utils_test.ts
+++ b/packages/order-utils/test/signature_utils_test.ts
@@ -99,6 +99,17 @@ describe('Signature utils', () => {
             );
             expect(isValidSignatureLocal).to.be.true();
         });
+
+        it.only('should return false if entry not found in `preSigned` mapping', async () => {
+            const preSignedSignature = '0x06';
+            const isValidSignatureLocal = await signatureUtils.isValidSignatureAsync(
+                provider,
+                dataHex,
+                preSignedSignature,
+                address,
+            );
+            expect(isValidSignatureLocal).to.be.true();
+        });
     });
     describe('#isValidECSignature', () => {
         const signature = {
--- a/packages/order-utils/package.json
+++ b/packages/order-utils/package.json
@@ -57,6 +57,7 @@
         "@0x/assert": "^1.0.23",
         "@0x/base-contract": "^3.0.13",
         "@0x/contract-artifacts": "^1.3.0",
+        "@0x/contract-addresses": "^2.2.0",
         "@0x/json-schemas": "^2.1.7",
         "@0x/types": "^1.5.2",
         "@0x/typescript-typings": "^3.0.8",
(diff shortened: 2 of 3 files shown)


packages/0x.js/package.json
```diff
@@ -24,7 +24,7 @@
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
         "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
         "clean": "shx rm -rf _bundles lib test_temp src/generated_contract_wrappers generated_docs",
-        "build:umd:prod": "NODE_ENV=production webpack",
+        "build:umd:prod": "NODE_ENV=production node --max_old_space_size=8192 ../../node_modules/.bin/webpack --mode production",
```
fabioberger: Fixes unrelated "Heap out of memory" error.

packages/order-utils/CHANGELOG.json
```diff
@@ -5,6 +5,10 @@
             {
                 "note": "Upgrade the bignumber.js to v8.0.2",
                 "pr": 1517
+            },
+            {
+                "note": "Fix preSigned isSignatureValidAsync check",
```
LogvinovLeon:
```suggestion
                "note": "Fix preSigned `isSignatureValidAsync` check",
```

packages/order-utils/test/signature_utils_test.ts
```diff
@@ -99,6 +99,17 @@ describe('Signature utils', () => {
             );
             expect(isValidSignatureLocal).to.be.true();
         });
+
```
LogvinovLeon: Maybe also add the test for a happy path?

packages/contract-wrappers/test/exchange_wrapper_test.ts
```diff
@@ -368,6 +368,23 @@ describe('ExchangeWrapper', () => {
...
+            const preSignedSignature = '0x06';
+            const isValidSignature = await contractWrappers.exchange.isValidSignatureAsync(
+                hash,
+                signerAddress,
+                preSignedSignature,
+            );
+            expect(isValidSignature).to.be.true();
+
+            // Test our TS implementation of signature validation
+            const isValidSignatureInTs = await signatureUtils.isValidSignatureAsync(
```
LogvinovLeon: If our TS implementation doesn't live in that package - why do we test it here?

packages/order-utils/test/signature_utils_test.ts
fabioberger: See other file.

packages/contract-wrappers/test/exchange_wrapper_test.ts
fabioberger: Because we don't do any "integration tests" in OrderUtils and for this, it needs to be an integration test. Agree it's weird, but this package is gone after "better revert reasons" merges.

PR #1581, opened 2019-02-04
Big number input to Instant coercion to consistent version

dave4506:
## Description

Added util function to coerce strings or big numbers to the version used by asset-buyer. 

Working on adding test cases and looking for feedback in the general design. 

## Checklist:

-   [ x ] Prefix PR title with `[WIP]` if necessary.


packages/instant/src/util/maybe_big_number.ts
```diff
@@ -16,6 +16,17 @@ export const maybeBigNumberUtil = {
 
         return validBigNumber.isNaN() ? undefined : validBigNumber;
     },
+    // converts a BigNumber or String to the BigNumber used by 0x libraries
+    bigNumberOrStringToMaybeBigNumber: (value: any): Maybe<BigNumber> => {
```
fragosti: I would just call this `toMaybeBigNumber`

packages/instant/src/util/signed_order_coercion.ts
```diff
@@ -0,0 +1,26 @@
+import { SignedOrder } from '@0x/types';
```
fragosti: I would structure this similar to https://github.com/0xProject/0x-monorepo/blob/development/packages/order-utils/src/parsing_utils.ts#L4

So have a general function that tries to convert chosen keys to BigNumber and then define the fields in a separate function. This logic does belong here though as it is pretty specific to instant.

packages/instant/src/index.umd.ts
```diff
@@ -93,7 +94,12 @@ export interface ZeroExInstantConfig extends ZeroExInstantOverlayProps {
 }
 
 export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_ZERO_EX_CONTAINER_SELECTOR) => {
+    if (_.isArray(config.orderSource)) {
+        config.orderSource = orderCoercionUtil.coerceOrderArrayFieldsToBigNumber(config.orderSource);
```
fragosti: Can you change this to not modify the original `config` object?

PR #1586, opened 2019-02-05
Add Heap and Google analytics to Dev Tools Pages

fabioberger:
## Description

See title.

## Checklist:


packages/dev-tools-pages/assets/index.html
```diff
@@ -1,14 +1,66 @@
...
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+        <link rel="preload" href="/fonts/MaisonNeue-Book-subset.woff2" as="font" type="font/woff2" crossorigin />
+        <link rel="preload" href="/fonts/MaisonNeue-Bold-subset.woff2" as="font" type="font/woff2" crossorigin />
+        <title><%= htmlWebpackPlugin.options.title %></title>
+    </head>
+    <body>
+        <!-- Heap SDK -->
+        <script type="text/javascript">
```
steveklebanoff: Might be cleaner to do these analytic embeds in a separate JS file. This could be a small performance boost in the case of repeat visits, although might be slower if the user just visits the site once.  Don't feel too strongly about this though.

packages/dev-tools-pages/assets/index.html
```diff
@@ -1,14 +1,66 @@
...
+                                'track',
+                                'unsetEventProperty',
+                            ],
+                            c = 0;
+                        c < p.length;
+                        c++
+                    )
+                        heap[p[c]] = o(p[c]);
+                });
+            heap.load('<%= htmlWebpackPlugin.options.heapAnalyticsId %>');
```
steveklebanoff: Would be a bit safer to wrap the heap and google embed codes in functions, and only execute if `htmlWebpackPlugin.options.heapAnalyticsId` or `htmlWebpackPlugin.options.googleAnalyticsId` are actually defined

packages/dev-tools-pages/assets/index.html
```diff
@@ -1,14 +1,66 @@
...
+            src="https://www.googletagmanager.com/gtag/js?id=<%= htmlWebpackPlugin.options.googleAnalyticsId %>"
+        ></script>
+        <script>
+            window.dataLayer = window.dataLayer || [];
+            function gtag() {
+                dataLayer.push(arguments);
+            }
+            gtag('js', new Date());
+
+            gtag('config', '<%= htmlWebpackPlugin.options.googleAnalyticsId %>');
```
steveklebanoff: I'd prefer to only render google analytics & heap during a production build, looks like right now we always render in development?

packages/dev-tools-pages/assets/index.html
```diff
@@ -1,14 +1,66 @@
...
+                                'unsetEventProperty',
+                            ],
+                            c = 0;
+                        c < p.length;
+                        c++
+                    )
+                        heap[p[c]] = o(p[c]);
+                });
+            heap.load('<%= htmlWebpackPlugin.options.heapAnalyticsId %>');
+        </script>
```
steveklebanoff: You may not need this, but for instant we call `addEventProperties` with the `GIT_SHA` so we know what revision of the code the user is on

packages/dev-tools-pages/assets/index.html
fabioberger: I'd rather it failed loudly then silently. If they aren't defined, they should be added.

packages/dev-tools-pages/assets/index.html
fabioberger: True. My feeling is that these single landing pages are more of a single visit type deal.

steveklebanoff:
Looks good to me! Only question left is: are the yarn.lock changes intentional?

fabioberger:
Re: yarn.lock, not intentional, probably left over from someone else's PR

PR #1593, opened 2019-02-06
Add migration to fix exchange events primary keys

feuGeneA:
## Description

It was noticed that the `pull_exchange_events.ts` script is regularly failing with the message `error: duplicate key value violates unique constraint "PK_79d3b54710830426f998cd6263b"`, repeated for each of the tables into which it inserts.  This change is to fix that error.

A new migration script will add `transaction_hash` to (or remove it from, for revert) the set of columns that comprise the primary key.

## Testing instructions

With local DB running as described in README,
```
yarn build
yarn migrate:run
psql --username=postgres --password --host=localhost --command="\d raw.exchange_cancel_events"
```
entering `postgres` for the password, and visually confirm that the `PRIMARY KEY` line includes `transaction_hash`.
Repeat `psql` command for other tables, replacing `exchange_cancel_events` with `exchange_cancel_up_to_events`, and subsequently with `exchange_fill_events`, confirming the `PRIMARY KEY` for each.

To test revert:
```
yarn migrate:revert
```
Then run the same `psql` commands, confirming that `PRIMARY KEY` does *not* contain `transaction_hash`.

## Types of changes

## Checklist:

### DIFF
--- a/packages/pipeline/migrations/1549479172800-AddTxHashToExchangeEventPrimaryKey.ts
+++ b/packages/pipeline/migrations/1549479172800-AddTxHashToExchangeEventPrimaryKey.ts
@@ -1,0 +1,35 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+const tableNames = ['exchange_cancel_events', 'exchange_cancel_up_to_events', 'exchange_fill_events'];
+
+const oldPrimaryColumns = ['contract_address', 'log_index', 'block_number'];
+
+const newPrimaryColumns = ['transaction_hash'];
+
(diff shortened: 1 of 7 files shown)


albrow:
Just an update for anyone reading along:

> It was noticed that the `pull_exchange_events.ts` script is regularly failing with the message `error: duplicate key value violates unique constraint "PK_79d3b54710830426f998cd6263b"`, repeated for each of the tables into which it inserts. This change is to fix that error.

I talked to Gene about this separately and we're on the same page now. The error in the logs about a primary key constraint violation is an expected side effect of the hack we had to introduce to work around a bug in TypeORM's `save` method. The real problem that this PR addresses is separate: the primary key we were using for events was flawed and we needed to add `transaction_hash` in order for the primary key to truly denote unique events.

PR #1601, opened 2019-02-08
Slippage calculations from Ethereum DEX Prices Service (EDPS)

askeluv:
## Description

Slippage calculations per (DEX, token, $USD depth) using [Ethereum DEX Pricing Service](https://github.com/grahamperich/ethereum-dex-prices-service). In particular, we calculate the bid-ask spread at different depths denoted in $USD.

## Testing instructions

After setting up the pipeline dev environment per the README:

```
yarn build
yarn migrate:run
yarn test
yarn test:db
```

## Types of changes

## Checklist:

-   [x] Fetch Ethereum DEX Prices Service data
-   [x] Fetch current token USD price
-   [x] Calculate slippage
-   [x] Save slippage records to database

### Optional / for discussion:
-   [x] Replace EDPS endpoint with our own
~Fetch list of tokens from database~


askeluv:
Should be ready for review @fragosti @albrow @fabioberger 
I'm not sure why some of those tests fail - would be great to get your input.

feuGeneA:
@askeluv the `test-publish` test is notoriously failure prone, and `submit-coverage` will fail unless you [do this](https://github.com/0xProject/0x-monorepo/blob/development/CONTRIBUTING.md#fix-submit-coverage-ci-failure) (which you may or may not want to do; the tradeoff is that `test-publish` becomes less likely to pass).

However, the `static-tests` failure seems to actually be your fault, as prettier is flagging your source files as not pretty enough. :stuck_out_tongue:

packages/pipeline/src/entities/slippage.ts
```diff
@@ -0,0 +1,25 @@
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+import { BigNumber } from '@0x/utils';
+
+import { bigNumberTransformer, numberToBigIntTransformer } from '../utils';
+
+@Entity({ name: 'slippage', schema: 'raw' })
+export class Slippage {
+    @PrimaryColumn({ name: 'observed_timestamp', type: 'bigint', transformer: numberToBigIntTransformer})
```
feuGeneA: The set of primary columns in this file is different from the set declared in the migration script.  Change one or the other, so that they're consistent.

fragosti:
For the prettier errors you can run `yarn prettier` at the monorepo root.

packages/pipeline/src/data_sources/slippage/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    USD: number;
+}
+
+// tslint:disable:prefer-function-over-method
+// ^ Keep consistency with other sources and help logical organization
+export class EdpsSource {
+    /**
+     * Call Ethereum DEX Price Service API.
+     */
+    public async getEdpsAsync(direction: string, symbol: string, amount: number): Promise<Map<string, EdpsExchange>> {
```
fragosti: So this is actually the first use of es6 `Map` in the entire codebase! While I don't think we have particularly strong reasons for not using `Map`, we do always just use simple Objects and it might be worth switching this just to stay consistent. We do have an `ObjectMap` type (`import { ObjectMap } from '@0x/types';`) that we use pretty often, that is just like your `EdpsWrapper` type, but with some generics.

https://medium.com/front-end-weekly/es6-map-vs-object-what-and-when-b80621932373

packages/pipeline/src/data_sources/slippage/index.ts
```diff
@@ -0,0 +1,55 @@
+import { fetchAsync } from '@0x/utils';
+
+const EDPS_BASE_URL = 'https://ethereum-dex-prices-service.production.airswap.io';
+const PRICE_BASE_URL = 'https://min-api.cryptocompare.com/data/price?tsyms=USD';
```
fragosti: I feel like this logic could just go into `ohlcv_external/crypto_compare.ts`? and then you can import that code into here.

packages/pipeline/src/parsers/slippage/index.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param usdAmount Amount to buy/sell in USD.
+ * @param exchange Exchange where we are calculating slippage.
+ * @param buyEdps Ethereum DEX price service object for buy side.
+ * @param sellEdps Ethereum DEX price service object for sell side.
+ *
+ */
+
+export function calculateSlippage(usdAmount: number,
+                                  exchange: string,
+                                  buyEdps: Map<string, EdpsExchange>,
```
fragosti: Same `Map` comment.

packages/pipeline/src/scripts/pull_slippage.ts
```diff
@@ -0,0 +1,49 @@
...
+import { handleError } from '../utils';
+
+// Number of orders to save at once.
+const BATCH_SAVE_SIZE = 1000;
+
+// USD amounts for slippage depths
+// tslint:disable-next-line:custom-no-magic-numbers
+const USD_AMOUNTS = [10, 100, 1000, 10000];
+
+// TODO: fetch from database
```
fragosti: We can do this as a followup if you like. If we keep it as a manual list, I think it should contain all the MM program tokens as well. So it's missing LOOM, DNT, CVC

packages/pipeline/src/scripts/pull_slippage.ts
```diff
@@ -0,0 +1,49 @@
...
+
+let connection: Connection;
+
+(async () => {
+    connection = await createConnection(ormConfig as ConnectionOptions);
+    const edpsSource = new EdpsSource();
+
+    logUtils.log('Fetching slippage records');
+    const nestedSlippages: Slippage[][][] = await Promise.all(TOKENS.map(async symbol => {
+        const usdPrice = await edpsSource.getUsdPriceAsync(symbol);
```
fragosti: Should `usdPrice` be the price from the cryptocompare API, or the price of some small amount from the endpoint?

packages/pipeline/src/scripts/pull_slippage.ts
```diff
@@ -0,0 +1,49 @@
...
+            const amount = usdAmount / usdPrice;
+            const buyEdps = await edpsSource.getEdpsAsync('buy', symbol, amount);
+            const sellEdps = await edpsSource.getEdpsAsync('sell', symbol, amount);
+            return Array.from(buyEdps.keys()).map(exchange => {
+                const slippage: Slippage = calculateSlippage(usdAmount, exchange, buyEdps, sellEdps);
+                return slippage;
+            });
+        }));
+    }));
+    const slippagesWithEmptyRecords = nestedSlippages
```
fragosti: What is this doing? is this a `flatten`? https://ramdajs.com/docs/#flatten

packages/pipeline/src/data_sources/slippage/index.ts
fragosti: Oh I was wrong about this. We have used `Map` before, but never returned one from a public interface.

feuGeneA:
I'm having some reservations about the overall structure of the code.

This is the first time we're putting code into this package that isn't just scraping raw data and filling up analogous tables.  That is, this is the first time (in this package) that we're synthesizing multiple data sources into one table.

At the very least, I think we should rename the table and not keep it in `raw.*`.  More generally, though, I think we should consider separating out the scraping from the synthesis, probably keeping scraping in `data_sources`, and maybe putting synthesis into its own folder (`src/transformers`?) or just into `src/scripts`.  Whatever we do, I think we should make sure there's a clear distinction between extraction scripts and transformation scripts.  Does that make sense?

Whatever it is we decide to do for this, I'd like to see the SQL queries from `0x-pipeline-orchestration/airflow/dags/temp_concepts` move over into the `pipeline` package, just so that everything's in one place.  I'm not saying that needs to happen in this PR, just elaborating on a potential vision.

packages/pipeline/src/scripts/pull_slippage.ts
askeluv: Yes, indeed! This is just my awkward Typescript/Javascript - will change to `flatten`

packages/pipeline/src/scripts/pull_slippage.ts
askeluv: Yeah, good question. I went for CryptoCompare mostly as a pragmatic solution, because there are a few issues with using some small amount from the EDPS endpoint:

The main one is that the EDPS endpoint returns ETH prices, so unless we want to denote slippage amounts in ETH, we would still need to look up the ETH price (e.g. via CryptoCompare).

Also, prices from the EDPS are per exchange, so we would have to aggregate them to get 1 price. Definitely doable, but not sure it's worth the cost. Should we do average? Weighted average? Median? Etc.

Finally, a minor point: if we look up price via EDPS, that "small amount" (e.g. a fixed 0.01) will have *some* amount of slippage already built into it (e.g. 0.01 MKR is $5.1 while 0.01 MANA is ~$0.00035). This is not a big problem, but worth mentioning.

packages/pipeline/src/data_sources/slippage/index.ts
```diff
@@ -0,0 +1,55 @@
+import { fetchAsync } from '@0x/utils';
```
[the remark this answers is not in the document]
askeluv: Yes, good call. I'll change it.

packages/pipeline/src/data_sources/slippage/index.ts
askeluv: Great input - thanks, will change it

packages/pipeline/src/scripts/pull_slippage.ts
askeluv: Yeah, if you don't mind, it would be great to ship a first version so that we start capturing data. Will add those 3 tokens.

askeluv:
@feuGeneA Definitely. So you're addressing two things here: transformations + multiple data sources.

On transformations: agree that it's generally cleaner to separate extraction from transformation. I took the liberty of using the `parser` concept as `transformer`, but this might be a stretch.

The easiest option is probably to move the resulting table out of `.raw.*` as you suggest, and then to move `calculateSlippage` out of `src/parsers` and into a new `src/transformers`. Or we could rename `parsers` to `transformers`. This new group would then correspond to "transform" in extract-transform-load.

Another option which requires more work might be to make the `transform` operations SQL queries executed in Airflow DAGs. Specifically there are two transformations happening:
1. Join of buy/sell side so we get them in 1 record -> slightly awkward to do in SQL as we need to fix the timestamps for easy joining, but doable
2. Calculate slippage -> easy to do in SQL
As with all Airflow DAGs, we also have to consider frequency (daily, hourly etc) + data dependencies.

On multiple data sources: it's basically only data captured from EDPS that gets stored. The CryptoCompare endpoint is only used to prepare the query towards EDPS (that is, we need a USD price for each token in our API requests). If we solve the transformation issue, I personally don't think multiple data sources is an issue here - but would be great to get your feedback.

packages/pipeline/src/scripts/pull_slippage.ts
fragosti: Yea all good points. This might even be better. Slippage from global market price seems more useful / more important than slippage from local orderbook market price.

askeluv:
I _think_ the 2 remaining failed tests are not due to the code, but not 100% sure. As recommended by @feuGeneA I read [this on failing submit-coverage](https://github.com/0xProject/0x-monorepo/blob/development/CONTRIBUTING.md#fix-submit-coverage-ci-failure), but the instructions mention I should do something _before_ creating the PR, and I suspect I should not be setting a new COVERALLS_REPO_TOKEN in CircleCI now (?).

Also, [Forkdelta was recently added to ethereum-dex-prices-service](https://github.com/grahamperich/ethereum-dex-prices-service/pull/10), which started generating timeout errors. Unfortunately those errors break the entire request for a specific token/amount. So I set up a (temporary) separate endpoint where Forkdelta has been removed (see `EDPS_BASE_URL`). We can replace my endpoint with something hosted by 0x as soon as we're ready to deploy.

fragosti:
@askeluv can we use the package directly?

askeluv:
@fragosti probably. How might we do that in a reasonable way?

fabioberger:
`Submit-coverage` always fails for external contributors.

fragosti:
@askeluv oh the package is not on NPM. I think this is fine! We can always to this as a follow-up. Lets ship this !

feuGeneA:
> @feuGeneA Definitely. So you're addressing two things here: transformations + multiple data sources.
> 
> On transformations: agree that it's generally cleaner to separate extraction from transformation. I took the liberty of using the `parser` concept as `transformer`, but this might be a stretch.
> 
> The easiest option is probably to move the resulting table out of `raw.*` as you suggest, and then to move `calculateSlippage` out of `src/parsers` and into a new `src/transformers`. Or we could rename `parsers` to `transformers`. This new group would then correspond to "transform" in extract-transform-load.
> 
> Another option which requires more work might be to make the `transform` operations SQL queries executed in Airflow DAGs. Specifically there are two transformations happening:
> 
> 1. Join of buy/sell side so we get them in 1 record -> slightly awkward to do in SQL as we need to fix the timestamps for easy joining, but doable
> 2. Calculate slippage -> easy to do in SQL
>    As with all Airflow DAGs, we also have to consider frequency (daily, hourly etc) + data dependencies.
> 
> On multiple data sources: it's basically only data captured from EDPS that gets stored. The CryptoCompare endpoint is only used to prepare the query towards EDPS (that is, we need a USD price for each token in our API requests). If we solve the transformation issue, I personally don't think multiple data sources is an issue here - but would be great to get your feedback.

Regarding transformations, let's go with your "easiest option" for now (just moving `calculateSlippage` to new `src/transformers`, not renaming `parsers`), so we can ship this sooner, and let's keep a mind for iterating on this as we do future changes.

Regarding multiple data sources, I think it's fine to leave it the way it is.

askeluv:
> @askeluv oh the package is not on NPM. I think this is fine! We can always to this as a follow-up. Lets ship this !

Will leave this as a follow-up. We can revisit once they make an NPM module - it's listed as one of two things they're going to do next.

askeluv:
> I think our convention is to checkout the branch on the QA ec2 instance and deploy to make sure things work as expected before merging into development and deploying to prod.

Sounds good! I don't have AWS access yet - should someone else in the team do the checkout in QA?

Also @feuGeneA made a good point about the `raw` schema not being suitable for the table. We discussed this a bit in parallel.
In brief we agreed to keep it in raw for now, but we should review the overall data architecture later. The 3 layers of raw -> concepts -> analytics make sense but we might want to (at least) rename. E.g. "raw" is not necessarily raw - it's just that this is where data lands first when it enters our data platform. So renaming to "staging" or "inbox" might make sense.

packages/pipeline/src/scripts/pull_slippage.ts
askeluv: Quick note on this - I tried using Ramda but it didn't really flatten as expected. Maybe because the nested arrays contained objects (?). I've left it as is for now.

fragosti: Does unnest work?

PR #1604, opened 2019-02-08
Update exchange-libs and utils Solidity versions + bugfix

### LINKED ISSUE #1528: Function slice() and sliceDestructive() in LibBytes.sol cut off last byte of bytes array.
Functions `slice()` and `sliceDestructive()` in [LibBytes.sol](https://github.com/0xProject/0x-monorepo/blob/b5fd3c72a08aaa6957917d74c333387a16edf66b/contracts/utils/contracts/utils/LibBytes/LibBytes.sol#L168) do not allow to include last byte of a bytes array. 

If bytes array `data` is of length 100 and you want to get the last 50 bytes, you can't use `data.slice(49, 99)` since the `to` is exclusive and the last byte (at index 99) will therefore be excluded. You would need to put `data.slice(49, 100)`, but this will fail because of the `require(to < b.length)` and length is 100.  

This behaviour is also present in `sliceDestructive()`.

Here's the `slice()` code for simplicity:

```solidity
   /// @dev Returns a slices from a byte array.
    /// @param b The byte array to take a slice from.
    /// @param from The starting index for the slice (inclusive).
    /// @param to The final index for the slice (exclusive).
    /// @return result The slice containing bytes at indices [from, to)
    function slice(
        bytes memory b,
        uint256 from,
        uint256 to
    )
        internal
        pure
        returns (bytes memory result)
    {
        require(
            from <= to,
            "FROM_LESS_THAN_TO_REQUIRED"
        );
        require(
            to < b.length,
            "TO_LESS_THAN_LENGTH_REQUIRED"
        );
        
        // Create a new bytes structure and copy contents
        result = new bytes(to - from);
        memCopy(
            result.contentAddress(),
            b.contentAddress() + from,
            result.length
        );
        return result;
    }
```

Simple fix could be to change `require( to < b.length)` to `require(to <= b.length)` although I haven't see if it changes how tests passed.

abandeali1:
## Description

- Updates contract versions to 0.5.3 in `exchange-libs` and `utils` (prerequisite for TEC contracts)
- Fixes #1528 

## Checklist:


contracts/utils/CHANGELOG.json
```diff
@@ -1,4 +1,17 @@
 [
+    {
+        "version": "2.0.5",
+        "changes": [
+            {
+                "note": "Fix bug in `LibBytes.slice` and `LibBytes.sliceDestructive`",
```
hysz: Will this break any existing code in other contract packages?

hysz: * nvm - doesn't look like `slice` / `sliceDestructive` are being used elsewhere.

contracts/utils/test/lib_bytes.ts
```diff
@@ -870,5 +870,85 @@ describe('LibBytes', () => {
         describe('copies forward within one word and one byte overlap', () =>
             test([[0, 0, 1, 'one byte'], [0, 10, 11, 'eleven bytes'], [0, 15, 16, 'sixteen bytes']]));
     });
```
hysz: 👍

contracts/utils/contracts/src/LibBytes.sol
```diff
@@ -213,7 +213,7 @@ library LibBytes {
             "FROM_LESS_THAN_TO_REQUIRED"
         );
         require(
-            to < b.length,
+            to <= b.length,
             "TO_LESS_THAN_LENGTH_REQUIRED"
```
hysz: `TO_LESS_OR_EQUAL_TO_LENGTH_REQUIRED` ?

contracts/utils/contracts/src/LibBytes.sol
```diff
@@ -179,7 +179,7 @@ library LibBytes {
             "FROM_LESS_THAN_TO_REQUIRED"
         );
         require(
-            to < b.length,
+            to <= b.length,
```
hysz: Since `from <= to` and `to <= b.length` we can now encounter the scenario where `from == to == b.length`. We could add an extra assertion that `from < b.length` since it's intended to be a valid index, although looking at the implementations I don't think it should overwrite any unintended memory. Could still be worthwhile to include tests for this scenario though.

abandeali1: This edge case does still modify memory with `sliceDestructive`, although maybe that is to be expected. It shouldn't matter for `slice`, but I'll add the necessary tests.

PR #1608, opened 2019-02-11
Sol resolver fixes v3

LogvinovLeon:
## Description

This PR is a try to make the version locking work.

In particular:
* `NPMResolver` now has more rigorous checks. It checks that all the resolved dependencies are listed in package.json and the versions match the range required.
* `ReentrantERC20Token` was moved to the exchange package

## Types of changes

## Checklist:


package.json
```diff
@@ -43,6 +43,8 @@
         "generate_doc": "node ./packages/monorepo-scripts/lib/doc_generate_and_upload.js",
         "test:generate_docs:circleci": "for i in ${npm_package_config_packagesWithDocPages}; do yarn generate_doc --package $i --shouldUpload false --isStaging true || break -1; done;",
         "bundlewatch": "bundlewatch",
+        "comment:postinstall": "HACK: For some reason `yarn` is not setting up symlinks properly for some modules. We temporarily set them manually. Remove this after V3 refactor is complete.",
+        "postinstall": "rm -rf `pwd`/contracts/exchange-libs/node_modules/@0x; mkdir `pwd`/contracts/exchange-libs/node_modules/@0x; ln -s `pwd`/contracts/utils `pwd`/contracts/exchange-libs/node_modules/@0x/contracts-utils",
```
abandeali1: Man this is ugly. Will it always be these specific packages that are the issue? Is there really no other way to resolve this issue?

packages/sol-compiler/src/compiler.ts
```diff
@@ -113,7 +113,7 @@ export class Compiler {
         this._nameResolver = new NameResolver(this._contractsDir);
         const resolver = new FallthroughResolver();
         resolver.appendResolver(new URLResolver());
-        resolver.appendResolver(new NPMResolver(this._contractsDir));
+        resolver.appendResolver(new NPMResolver(process.cwd()));
```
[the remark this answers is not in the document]
LogvinovLeon: Yes, but we also had that before. Sol-compiler should be run from within your project directory

package.json
LogvinovLeon: It is ugly, but we haven't found any better way around it. We can remove that hack and try to debug yarn. We've done the same for v2.

packages/sol-compiler/src/compiler.ts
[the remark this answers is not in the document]
LogvinovLeon: We do `path.resolve` for the `contractsDir` above

packages/sol-resolver/src/resolvers/npm_resolver.ts
```diff
@@ -1,42 +1,86 @@
...
+        const packageJSONPath = path.join(packagePath, 'package.json');
+        const packageJSON = JSON.parse(fs.readFileSync(packageJSONPath).toString());
+        const packageVersion = packageJSON.version;
+        return packageVersion;
+    }
     constructor(packagePath: string) {
         super();
         this._packagePath = packagePath;
+        const packageJSONPath = path.join(this._packagePath, 'package.json');
+        const packageJSON = JSON.parse(fs.readFileSync(packageJSONPath).toString());
```
fabioberger: Let's dedup this code with the above. Perhaps a helper method that returns the `package.json` for a path?

packages/sol-resolver/src/resolvers/npm_resolver.ts
```diff
@@ -1,42 +1,86 @@
...
-                    : path.join(packageScopeIfExists, packageName);
-                const lookupPath = path.join(currentPath, 'node_modules', packagePath, pathWithinPackage);
-                if (fs.existsSync(lookupPath) && fs.lstatSync(lookupPath).isFile()) {
-                    const fileContent = fs.readFileSync(lookupPath).toString();
-                    return { source: fileContent, path: importPath, absolutePath: lookupPath };
-                }
-                currentPath = path.dirname(currentPath);
-            }
+        if (importPath.startsWith('/')) {
+            return undefined;
```
fabioberger: Can you explain this conditional? Why is this special-cased?

packages/sol-resolver/src/resolvers/npm_resolver.ts
```diff
@@ -1,42 +1,86 @@
...
+import * as semver from 'semver';
 
 import { ContractSource } from '../types';
 
 import { Resolver } from './resolver';
 
 export class NPMResolver extends Resolver {
     private readonly _packagePath: string;
+    private readonly _dependencies: { [dependencyName: string]: string };
+    private static _lookupPath(startingPath: string, lookupTarget: string): string | undefined {
```
fabioberger:
```suggestion
    private static _lookupPathIfExists(startingPath: string, lookupTarget: string): string | undefined {
```

packages/sol-resolver/src/resolvers/npm_resolver.ts
```diff
@@ -1,42 +1,86 @@
...
+        const dependencyLookupTarget = path.join('node_modules', dependencyName, internalPath);
+        const dependencyLookupResult = NPMResolver._lookupPath(this._packagePath, dependencyLookupTarget);
+        if (_.isUndefined(dependencyLookupResult)) {
+            return undefined;
+        }
+        if (_.isUndefined(this._dependencies[dependencyName])) {
+            throw new Error(
+                `Lookup for ${dependencyName} started from ${this._packagePath} failed. Not specified in package.json`,
+            );
+        }
```
fabioberger: Move this check up after the call to `NPMResolver._parseDependencyImport`. It does not depend on looking up the path and we should fail as fast as possible.

PR #1612, opened 2019-02-12
[pipeline] Add Radar orders and maker_address column to token_orderbook_snapshots

fragosti:
## Description

This PR does two things:
1. Scrapes Radars API for orderbook data.
2. Adds a `maker_address` column to the `token_orderbook_snapshots` table. In order to do this, I wrote a migration that sets this value to "unknown" for all previous values. I could not set it to null since this column needs to extend the pre-existing composite key and that guarantees uniqueness across a set of columns. For orderbooks scraped from Radar, this column will always be filled with the maker address, whereas for other sources it will always be "unknown". 

It is also worth noting that we "aggregate" individual orders in orderbooks to create an "orderbook" abstraction that just specifies "price" and "amount" at varying levels for both sides of the orderbook, and omits individual orders entirely. Since we want to compare all orderbooks on a level playing field, and a lot of APIs do not provide orders, this is the best we can do.

screen shot 2019-02-14 at 10 48 40 am

TODO:
- Tests. 

## Checklist:


packages/pipeline/src/parsers/radar_orders/index.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param observedTimestamp Time at which the orders for the market were pulled.
+ * @param source The exchange where these orders are placed. In this case 'radar'.
+ */
+export function parseRadarOrders(
+    radarOrderbook: RadarBook,
+    radarMarket: RadarMarket,
+    observedTimestamp: number,
+    source: string,
+): TokenOrder[] {
+    const transformToGeneric = (radarOrder: RadarSignedOrder) => _toGeneric(radarMarket, radarOrder);
```
xianny: can we call `_toGeneric` directly instead of using this? `radarOrderBook.bids.map(order => _toGeneric(radarMarket, order))`

packages/pipeline/src/parsers/radar_orders/index.ts
```diff
@@ -0,0 +1,91 @@
...
+ * @param radarOrderbook A raw orderbook that we pull from the radar API.
+ * @param radarMarket An object containing market data also directly from the API.
+ * @param observedTimestamp Time at which the orders for the market were pulled.
+ * @param source The exchange where these orders are placed. In this case 'radar'.
+ */
+export function parseRadarOrders(
+    radarOrderbook: RadarBook,
+    radarMarket: RadarMarket,
+    observedTimestamp: number,
+    source: string,
```
xianny: radar results would always have `source = radar` right? We could define RADAR_SOURCE here instead so the other two files don't need to handle any concept of source.

packages/pipeline/src/parsers/radar_orders/index.ts
```diff
@@ -0,0 +1,91 @@
...
+    radarMarket: RadarMarket,
+    observedTimestamp: number,
+    orderType: OrderType,
+    source: string,
+    radarOrder: [string, BigNumber],
+): TokenOrder {
+    const tokenOrder = new TokenOrder();
+    const price = new BigNumber(radarOrder[0]);
+    const amount = radarOrder[1];
+    const splitId = radarMarket.id.split('-');
```
xianny: radarMarket.id is only type-guaranteed to be a string. can we handle potentially malformed strings here? Even just log an error and skip. Would help with debugging.

packages/pipeline/src/scripts/pull_radar_orderbook_snapshots.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Retrieve orderbook from radar API for a given market. Parse orders and insert
+ * them into our database.
+ * @param radarSource Data source which can query radar API.
+ * @param market Object from radar API containing market data.
+ */
+async function getAndSaveMarketOrderbookAsync(radarSource: RadarSource, market: RadarMarket): Promise<void> {
+    const orderBook = await radarSource.getMarketOrderbookAsync(market.id);
+    const observedTimestamp = Date.now();
```
xianny: My instinct is that we should use the same `observedTimestamp` for all markets, so we can use it as a task identifier.

packages/pipeline/src/data_sources/radar/index.ts
```diff
@@ -0,0 +1,34 @@
...
+        return markets;
+    }
+
+    /**
+     * Retrieve orderbook from Radar API for a given market.
+     * @param marketId String identifying the market we want data for. Eg. 'REP/AUG'
+     */
+    public async getMarketOrderbookAsync(marketId: string): Promise<RadarBook> {
+        logUtils.log(`${marketId}: Retrieving orderbook.`);
+        const marketOrderbookUrl = `${ACTIVE_MARKETS_URL}/${marketId}/book?perPage=${MAX_PER_PAGE}`;
```
xianny: does this mean we only get the last 1000 items? will we be missing data?

packages/pipeline/src/scripts/pull_radar_orderbook_snapshots.ts
fragosti: It's too late for this as all other DEX scraping jobs do this -- which also makes sense because orderbooks change pretty quickly in some cases.

packages/pipeline/src/data_sources/radar/index.ts
fragosti: I can change it to 10000

packages/pipeline/src/parsers/radar_orders/index.ts
fragosti: This is how every other parser does it, but change change those too

packages/pipeline/src/parsers/radar_orders/index.ts
fragosti: We log the market IDs elsewhere, so we'd be able to debug (and a stack trace). There are a lot of places where we could be defensive but I think we just need to assume things are formatted correctly.

packages/pipeline/src/parsers/radar_orders/index.ts
fragosti: Actually looking into it again, I think I prefer the current format.

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
```diff
@@ -0,0 +1,36 @@
+import { MigrationInterface, QueryRunner, TableColumn } from 'typeorm';
+
+const TOKEN_ORDERBOOK_SNAPSHOT_TABLE = 'raw.token_orderbook_snapshots';
+const NEW_COLUMN_NAME = 'maker_address';
+
+export class TokenOrderBookSnapshotsAddMakerAddress1550163069315 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<any> {
+        const snapshotTable = await queryRunner.getTable(TOKEN_ORDERBOOK_SNAPSHOT_TABLE);
+        if (snapshotTable) {
```
feuGeneA: Why this `if`?  Shouldn't it be presumed that when this migration is run the table *will* exist?  (This isn't necessarily a critique; it's just as much for my own edification.)  In any case, if the `if` really is useful, then maybe add an error/warning/info message in an `else` block?

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+            // backfill null values
+            await queryRunner.query(`
+                UPDATE ${TOKEN_ORDERBOOK_SNAPSHOT_TABLE}
+                SET ${NEW_COLUMN_NAME}='unknown'
+                WHERE ${NEW_COLUMN_NAME} is NULL;
+            `);
+            await queryRunner.query(`
+                ALTER TABLE ${TOKEN_ORDERBOOK_SNAPSHOT_TABLE}
+                    DROP CONSTRAINT "token_orderbook_snapshots_pkey1",
+                    ADD PRIMARY KEY (observed_timestamp, source, order_type, price, base_asset_symbol, quote_asset_symbol, maker_address);
```
feuGeneA: Did you know that there's an `updatePrimaryKeys()` method provided in typeorm?  Not sure it's functionally any different than what you're doing here, but thought you might like to know.  See how I used it in https://github.com/0xProject/0x-monorepo/blob/76d8a991c0fe5b9fc1371b85b851179d8d8daa01/packages/pipeline/migrations/1549499426238-AddTxHashToERC20ApprovalEventPrimaryKey.ts#L20

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
```diff
@@ -0,0 +1,36 @@
...
+            `);
+        }
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<any> {
+        const snapshotTable = await queryRunner.getTable(TOKEN_ORDERBOOK_SNAPSHOT_TABLE);
+        if (snapshotTable) {
+            await queryRunner.dropColumn(snapshotTable, NEW_COLUMN_NAME);
```
feuGeneA: Do you not need to update the primary key constraint back to the set that excludes your new column?  Or does `dropColumn()` do that implicitly for you?

packages/pipeline/src/parsers/radar_orders/index.ts
```diff
@@ -0,0 +1,121 @@
...
+import { OrderType } from '../../types';
+
+export interface AggregateOrdersByMaker {
+    makerAddress: string;
+    price: string;
+    amount: BigNumber;
+}
+
+/**
+ * Marque function of this file.
```
feuGeneA: Nit: should be spelled `marquee`. :P

packages/pipeline/src/parsers/radar_orders/index.ts
feuGeneA: I must agree with Xianny that this feels weird to always pass `radar` for `source` when the function isn't used for anything else besides Radar.

packages/pipeline/src/parsers/radar_orders/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    tokenOrder.quoteAssetSymbol = splitId[1];
+    tokenOrder.quoteAssetAddress = radarMarket.quoteTokenAddress || null;
+    tokenOrder.quoteVolume = price.times(amount);
+
+    tokenOrder.makerAddress = aggregateOrder.makerAddress;
+    return tokenOrder;
+}
+
+function _toGeneric(radarMarket: RadarMarket, radarOrder: RadarSignedOrder): GenericRawOrder | undefined {
```
feuGeneA: why the `_` prefix?  Is it to denote "private"?  Doesn't the lack of an `export` keyword already do that?  (Again, questioning just as much for my own edification as for the sake of critiquing.)

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
fragosti: The `if` is there because typescript requires it. Adding throws.

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
fragosti: Yea this didn't work for some reason.

packages/pipeline/migrations/1550163069315-TokenOrderBookSnapshotsAddMakerAddress.ts
fragosti: Yep!

packages/pipeline/src/parsers/radar_orders/index.ts
fragosti: hah they all have this problem. fixing them.

packages/pipeline/src/parsers/radar_orders/index.ts
fragosti: I've seen this convention used elsewhere in the code. I think sometimes we export these private methods just to test them, but still don't consider them part of the public interface.

PR #1622, opened 2019-02-19
[Governance] Ledger address selector ZRX balance

dekz:
Fix bug in ledger address selector where provider does not exist. This can happen in browsers without metamask or a injected provider

## Checklist:

### DIFF
--- a/packages/website/ts/pages/governance/address_table.tsx
+++ b/packages/website/ts/pages/governance/address_table.tsx
@@ -70,10 +70,10 @@ export class AddressTable extends React.Component<AddressTableProps, AddressTabl
 const Wrapper = styled.div<{ marginBottom?: string }>`
     background - color: #fff;
     border -radius;: 4;px;
-    Margin-bottom;: $;{props => props.marginBottom || '25px';}
-    Padding: 10;px; 30;px;
-    Height: 230;px;
-    Overflow - y;: auto;
+    margin-bottom;: $;{props => props.marginBottom || '25px';}
+    padding: 10;px; 30;px;
+    height: 230;px;
+    overflow - y;: auto;
 `;
 
 const Table = styled.table`;
--- a/packages/website/ts/pages/governance/connect_form.tsx
+++ b/packages/website/ts/pages/governance/connect_form.tsx
@@ -220,20 +220,14 @@ export class ConnectForm extends React.Component<Props, State> {
     }
     public async getZrxBalanceAsync(owner: string): Promise<BigNumber> {
         utils.assert(!_.isUndefined(this._contractWrappers), 'ContractWrappers must be instantiated.');
-        const injectedProvider = await this._getInjectedProviderIfExistsAsync();
-
-        if (!_.isUndefined(injectedProvider)) {
-            const contractAddresses = getContractAddressesForNetworkOrThrow(this.networkId);
-            const tokenAddress: string = contractAddresses.zrxToken;
-            try {
-                const amount = await this._contractWrappers.erc20Token.getBalanceAsync(tokenAddress, owner);
-                return amount;
-            } catch (error) {
-                return ZERO;
-            }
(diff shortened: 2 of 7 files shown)


packages/website/ts/pages/governance/address_table.tsx
```diff
@@ -68,15 +68,15 @@ export class AddressTable extends React.Component<AddressTableProps, AddressTabl
     }
 }
 const Wrapper = styled.div<{ marginBottom?: string }>`
-    background - color: #fff;
```
BMillman19: hmm, anywhere else this could have happened?

packages/website/ts/pages/governance/connect_form.tsx
```diff
@@ -220,20 +220,14 @@ export class ConnectForm extends React.Component<Props, State> {
...
-            try {
-                const amount = await this._contractWrappers.erc20Token.getBalanceAsync(tokenAddress, owner);
-                return amount;
-            } catch (error) {
-                return ZERO;
-            }
+        const contractAddresses = getContractAddressesForNetworkOrThrow(this.networkId);
+        const tokenAddress: string = contractAddresses.zrxToken;
+        try {
+            const amount = await this._contractWrappers.erc20Token.getBalanceAsync(tokenAddress, owner);
```
BMillman19: should we explicitly check for undefined here in an if statement above? it was not clear to me that this try/catch would also be catching the case where we are trying to grab `erc20Token` from an `undefined` `this._contractWrappers`

packages/website/ts/pages/governance/ledger_sign_note.tsx
```diff
@@ -0,0 +1,50 @@
+import * as _ from 'lodash';
+import * as React from 'react';
+import styled from 'styled-components';
+
+import { colors } from 'ts/style/colors';
+
+interface LedgerSignNoteProps {
+    text?: string;
```
BMillman19: do we want this to be optional?

packages/website/ts/pages/governance/vote_form.tsx
```diff
@@ -193,15 +197,19 @@ export class VoteForm extends React.Component<Props> {
         e.preventDefault();
 
         const { zeip, votePreference, comment } = this.state;
-        const { currentBalance, selectedAddress } = this.props;
+        const { currentBalance, selectedAddress, isLedger } = this.props;
         const makerAddress = selectedAddress;
+
+        if (isLedger) {
```
BMillman19: I'm wondering if there is any case where this setState could potentially be ordered after the `setState` that happens in `_handleError`

packages/website/ts/pages/governance/ledger_sign_note.tsx
sjelfull: I did it like that because of also passing them to the styled-component, which only uses `isVisible` - but is there a better pattern?

Is this better?
```
interface WrapperProps {
	isVisible: boolean;
}
interface LedgerSignNoteProps extends WrapperProps {
	text: string;
}
```

packages/website/ts/pages/governance/vote_form.tsx
sjelfull: I wasn't too worried about it, but thought of it. Do I have to use a callback to ensure it runs before continuing?

packages/website/ts/pages/governance/connect_form.tsx
dekz: As discussed we have an assert above this. This is to catch any network errors or anything else that could go wrong with an initialised contractWrappers.

packages/website/ts/pages/governance/ledger_sign_note.tsx
BMillman19: hmm yea i think that is a better pattern, but i'm fine leaving it either way

PR #1625, opened 2019-02-19
Sol-compiler offline mode

LogvinovLeon:
## Description

While offline @abandeali1 complained that sol-compiler doesn't have an offline mode. It now does.

## Testing instructions

Disable internet. Compile. it fails. Flip the flag to true. Compile. Voila!

## Types of changes

## Checklist:


packages/sol-compiler/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "3.1.0",
+        "changes": [
+            {
+                "note": "Add `isOfflineMode` flag to sol-compiler`",
+                "pr": "TODO"
```
fabioberger:
```suggestion
                "pr": "1625"
```

PR #1626, opened 2019-02-19
[WIP] Fix bug with document titles + add rich metadata for documents

dave4506:
Removed react-document-title and created a component to use react-helmet. Issue appears to be resolved.

Also added document title for '/credits' and '/market-making'. 

**TODO**
add meta tags for each page's description, keyword, and title
## Checklist:


fragosti:
Looks good! Will take another look once that TODO is done.

packages/website/ts/components/documentTitle.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
```
fragosti: This file needs to be snake_case. This convention was consistent before but the b&b guys did not follow it for the 0x.org redesign.

PR #1627, opened 2019-02-20
Support EIP1193 provider interface

fabioberger:
## Description

This PR adds support for EIP 1193 & Web3.js@0.38 providers and transforms supplied providers into a consistent internal provider interface.

## Checklist:


packages/assert/src/index.ts
```diff
@@ -59,6 +59,7 @@ export const assert = {
     isBoolean(variableName: string, value: boolean): void {
         assert.assert(_.isBoolean(value), assert.typeAssertionMessage(variableName, 'boolean', value));
     },
+    // DEPRECATED: Please use providerUtils.standardizeOrThrow() instead
```
LogvinovLeon: This comment will live here forever and no one will ever see it. Can we use `logUtils.warn` instead?

packages/assert/test/assert_test.ts
```diff
@@ -154,18 +154,6 @@ describe('Assertions', () => {
             invalidInputs.forEach(input => expect(assert.isBoolean.bind(assert, variableName, input)).to.throw());
         });
     });
-    describe('#isWeb3Provider', () => {
```
LogvinovLeon: Let's live it here before we remove the code.

packages/ethereum-types/src/index.ts
```diff
@@ -6,10 +6,74 @@ export type JSONRPCErrorCallback = (err: Error | null, result?: JSONRPCResponseP
...
 
+/**
+ * The interface for the provider used internally by 0x libraries
+ * Any property we use from any SupportedProvider should we explicitly
+ * add here
+ */
+export interface ZeroExProvider {
+    isZeroExProvider?: boolean;
+    isMetaMask?: boolean;
+    isParity?: boolean;
```
dekz: Can we have a TODO to turn this remove this and use a getProviderType utility function which returns an enum. Feels a bit weird to just have these two (not Mist/Cipher/CBW etc)

packages/migrations/src/migration.ts
```diff
@@ -16,7 +17,7 @@ import { erc20TokenInfo, erc721TokenInfo } from './utils/token_info';
  * @param txDefaults Default transaction values to use when deploying contracts (e.g., specify the desired contract creator with the `from` parameter).
  * @returns The addresses of the contracts that were deployed.
  */
-export async function runMigrationsAsync(provider: Provider, txDefaults: TxData): Promise<ContractAddresses> {
+export async function runMigrationsAsync(provider: Web3ProviderEngine, txDefaults: TxData): Promise<ContractAddresses> {
```
dekz: We use web3 provider engine internally but this is also a public package. I don't think we need to require this to be a Web3ProviderEngine do we?

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -171,15 +168,22 @@ export class Web3Wrapper {
...
      */
-    public getProvider(): Provider {
+    public getProvider(): SupportedProvider {
+        return this._supportedProvider;
+    }
+    /**
+     * Retrieve the Web3 provider
+     * @return  Web3 provider instance
+     */
+    public getZeroExProvider(): ZeroExProvider {
```
dekz: How often do we use this? Seems only to be used in one test

packages/ethereum-types/src/index.ts
fabioberger: Can't easily do Cipher, CBW checks in Node.js. Added TODO.

packages/web3-wrapper/src/web3_wrapper.ts
fabioberger: Removed.

PR #1628, opened 2019-02-21
Fix inconsistently named constraint in migration

fragosti:
## Description

In https://github.com/0xProject/0x-monorepo/pull/1612 I broke `test-pipeline`.

This is because when writing the migrations, I looked at the name for the PRIMARY KEY constraint on `raw.token_orderbook_snapshots` on the QA and prod instances, both called `token_orderbook_snapshots_pkey1`. I do not know how those constraints became named that way, but when you run the migration code on a fresh DB instance that constraint ends up being named `token_orderbook_snapshots_pkey`. 

I think the solution here is to stick with `token_orderbook_snapshots_pkey` and manually rename the name of the constraint on QA and prod for consistency.

## Checklist:


fragosti:
thanks for the feedback @feuGeneA !

PR #1629, opened 2019-02-21
Pull exchange contract transactions from Etherscan

askeluv:
## Description

Pulling exchange contract transactions from Etherscan - that is, transactions for this address: 0x4f833a24e1f95d70f028921e27040ca56e09ab0b

We can only get 10,000 transactions at a time with the Etherscan API, but we get the oldest transactions first and the next run starts where the former left off. This means that the script will automatically backfill over time. Currently there are ~134k transactions for the exchange contract, so it'll take 14 runs to backfill. With a DAG frequency of 1 run per 30 mins, it will catch up with the current block in 7 hours.

## Testing instructions

```
yarn build
yarn migrate:run
yarn test
yarn test:db
node lib/src/scripts/pull_exchange_transactions_from_etherscan.js
```

## Types of changes

## Checklist

-   [x]  Fetch transactions for 0x4f833a24e1f95d70f028921e27040ca56e09ab0b
-   [x]  Save transactions records to database
-   [x]  Add tests to cover changes as needed.


packages/pipeline/src/scripts/pull_exchange_transactions_from_etherscan.ts
```diff
@@ -0,0 +1,70 @@
...
+
+// API key to use if environment variable has not been set
+const FALLBACK_API_KEY = 'YourApiKeyToken';
+
+let connection: Connection;
+
+(async () => {
+    let apiKey = process.env.ETHERSCAN_API_KEY;
+    if (apiKey === undefined) {
+        logUtils.log(`Missing env var: ETHERSCAN_API_KEY - using default API key: ${FALLBACK_API_KEY}`);
```
xianny: I think if api key is undefined we should just throw an exception, so that we see the failure on airflow.

packages/pipeline/src/entities/etherscan_transaction.ts
```diff
@@ -0,0 +1,61 @@
+import { BigNumber } from '@0x/utils';
+import { Column, Entity, PrimaryColumn } from 'typeorm';
+
+import { bigNumberTransformer, numberToBigIntTransformer } from '../utils';
+
+@Entity({ name: 'etherscan_transactions', schema: 'raw' })
+export class EtherscanTransaction {
+    @PrimaryColumn({ name: 'hash' })
```
xianny: are we sure this column is a good uniqueness identifier? I'm not sure myself and I know that we've had issues with transaction uniqueness from other sources

packages/pipeline/test/parsers/etherscan/index_test.ts
```diff
@@ -0,0 +1,65 @@
...
+import { chaiSetup } from '../../utils/chai_setup';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+// tslint:disable:custom-no-magic-numbers
+describe('etherscan_transactions', () => {
+    describe('parseEtherscanTransactions', () => {
+        it('converts etherscanTransactions to EtherscanTransaction entities', () => {
+            const response: EtherscanTransactionResponse[] = [
```
xianny: can we use an actual API response for the test? Then it will also test if our defined type for the API response works as expected. 

There is an example of how we do this in the Crypto Compare scripts.

I think it would be an improvement, but I don't feel strongly on this.

packages/pipeline/src/entities/etherscan_transaction.ts
askeluv: At the transaction level the hash should be unique. I did a quick query now on the BigQuery Ethereum dataset and this query returned 0 rows:
```select tx.hash, count(*)
from `bigquery-public-data.crypto_ethereum.transactions` as tx
group by 1
having count(*) > 1;```

File: packages/pipeline/test/parsers/etherscan/index_test.ts (hunk 3 above)
Comment (askeluv) [reply packages/pipeline/test/parsers/etherscan/index_test.ts]: I've updated it now to use `EtherscanResponse` - was this what you had in mind?

Comment (xianny) [reply packages/pipeline/test/parsers/etherscan/index_test.ts]: Sorry, I pointed you to the wrong example before. It's in [Copper scripts](https://github.com/0xProject/0x-monorepo/blob/development/packages/pipeline/test/parsers/copper/index_test.ts). I actually meant using the raw  response from the API. I think it's a slightly more useful test than predefining a typescript object.

The type conversion should automatically catch any errors, but the approach with raw JSON fixtures means if anything breaks, we can quickly check for API compatibility by doing a `curl` and dumping the results in the fixtures files.

Up to you if you think it's worth implementing ^^

Comment (askeluv) [reply packages/pipeline/test/parsers/etherscan/index_test.ts]: Ah got it! Thanks - I'll update :)

Comment (askeluv) [reply packages/pipeline/test/parsers/etherscan/index_test.ts]: This was a really good call - it helped me surface a small bug. Thanks @xianny ! 🙏

File: packages/pipeline/test/fixtures/etherscan/api_v1_accounts_transactions.ts [hunk 4]
```diff
@@ -0,0 +1,30 @@
+import { BigNumber } from '@0x/utils';
+
+import { EtherscanTransaction } from '../../../src/entities';
+
+// To re-create the JSON file from the API (e.g. if the API output schema changes), run the below command:
```
xianny: nice 👍

xianny:
looks good to me! if it runs on QA and the data looks good, go ahead and merge :)

PR #1633, opened 2019-02-22
[pipeline] Add concepts schema and radar orderbook USD price concept

fragosti:
## Description

This PR:
1. Introduces the `concepts` schema in migrations, which are meant to be computed tables.
2. Creates the first `concept` which is the the radar order book with USD prices (and maker addresses).

It would be nice to include more data, ie. include data from all order books but only for `('BAT', 'DAI', 'FUN', 'MANA', 'OMG', 'REP', 'TUSD', 'ZRX', 'MKR', 'BNB', 'USDC', 'LOOM', 'DNT', 'CVC')`, but that query currently takes longer than 30 minutes. We need to address https://app.asana.com/0/682899751319128/1110513763783834 before we can use the rest of the data from `token_orderbook_snapshots`.

## Testing instructions

Added a test to run as part of `yarn test:db`

## Checklist:


packages/pipeline/src/concepts/queries/radar_orderbook_usd_prices.ts
```diff
@@ -0,0 +1,51 @@
...
+export async function createTableAsync(connection: Connection): Promise<void> {
+    await connection.query(
+        `
+create table concepts.radar_orderbook_usd_prices as
+with usd_prices as (
+    select * from raw.ohlcv_external
+    where to_symbol = 'USD' and end_time >= date_part('epoch', now() - interval '2 month')*1000
+), radar_orderbook as (
+    select * from raw.token_orderbook_snapshots
+    where source = 'radar' and observed_timestamp  > date_part('epoch', now() - interval '2 month')*1000
```
xianny: first query is `>=` and this one is `>`, is that intentional?

packages/pipeline/src/concepts/queries/radar_orderbook_usd_prices.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Async function that creates the concepts.radar_orderbook_usd_prices table
+ * @param connection connection to the db instance
+ */
+export async function createTableAsync(connection: Connection): Promise<void> {
+    await connection.query(
+        `
+create table concepts.radar_orderbook_usd_prices as
+with usd_prices as (
```
xianny: where do the tables/views `usd_prices1` and `usd_prices2` come from?

packages/pipeline/test/concepts/radar_orderbook_usd_prices_test.ts
```diff
@@ -0,0 +1,15 @@
+import 'mocha';
+
+import { createTableAsync, deleteTableAsync } from '../../src/concepts/queries/radar_orderbook_usd_prices';
+import { createDbConnectionOnceAsync } from '../db_setup';
+import { chaiSetup } from '../utils/chai_setup';
+
+chaiSetup.configure();
+
+describe('radar_orderbook_usd_prices', () => {
+    it('Does not crash when running on current schema', async () => {
```
xianny: 😂

packages/pipeline/src/concepts/queries/radar_orderbook_usd_prices.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Async function that removes the concepts.radar_orderbook_usd_prices table if it exists
+ * @param connection connection to the db instance
+ */
+export async function deleteTableAsync(connection: Connection): Promise<void> {
+    await connection.query(`drop table if exists concepts.radar_orderbook_usd_prices`);
+}
+
+/**
+ * Async function that creates the concepts.radar_orderbook_usd_prices table
```
xianny: Can we expand the docstring to describe what this query does and what the combined data means? Thinking about our 6-months-later-selves. Especially the `join... on` clauses. I had to read pretty close and I'm still not 100% sure :)

packages/pipeline/src/concepts/queries/radar_orderbook_usd_prices.ts
fragosti: woops!

packages/pipeline/src/concepts/queries/radar_orderbook_usd_prices.ts
fragosti: We left join on `usd_prices` twice. `usd_prices1` is from the first join, `usd_prices2` is from the second. We join twice to check the USD price of the token on both sides of the trade. The way the logic is written, if there is a match (meaning we have the info) for the first token (the buy token), we do not join on that row again. If there is no match, we try the second token (the sell token) to see if we know its price at that time. If there is no match, we still have a row for that trade since it is a left join, but do price information (`null`).

xianny: Ahhh I see it now. Missed that the first time. Thank you for explaining!

askeluv:
I'm probably missing something, but if this is mainly SQL operations, couldn't this just take place in Airflow as Postgres operators? That is, SQL queries are added here: https://github.com/0xProject/0x-pipeline-orchestration/tree/master/airflow/dags/temp_concepts and then added to a DAG?

fragosti:
@askeluv we want to move all of those over to this main repo. It is a bit more overhead, but I do like the idea of testing the queries such that you know they're even possible on the existing underlying schema. Currently all testing happens in the `pipeline` package.

albrow:
@askeluv to add to @fragosti's response. Including SQL queries in this repo also gives us the benefit of version control and some semblance of compile-time type safety (in some cases; especially if using TypeORM query builders or other utility functions).

packages/pipeline/src/scripts/update_radar_orderbook_usd_prices.ts
```diff
@@ -0,0 +1,83 @@
+import { logUtils } from '@0x/utils';
+import { Connection, ConnectionOptions, createConnection } from 'typeorm';
+
+import * as ormConfig from '../ormconfig';
+import { handleError, isCli } from '../utils';
+
+if (isCli()) {
```
albrow: I'm okay with this for now, but long-term we should implement proper logging with the ability to turn it on or off (and maybe even set different log levels).

xianny: is this different from `if (require.main === module)`? If not, would prefer to do it [the standard way](https://nodejs.org/api/modules.html#modules_accessing_the_main_module).

fragosti: I got this from https://stackoverflow.com/questions/6398196/node-js-detect-if-called-through-require-or-directly-by-command-line

What made me decide to use this is a comment that implied it works in more situations, but down to change it for more clarity.

PR #1636, opened 2019-02-22
Remove yarn install command from circleci config

BMillman19:
## Description

An attempt to fix the intermittent `Yarn: Permission Denied` errors we see on CircleCI

https://discuss.circleci.com/t/yarn-permission-denied/20160

The fix I decided to apply was to remove the part of our circleCI script that installs yarn and instead opts for using the version of yarn that ships with the node image. I found that that version of yarn that ships with the image that we use is yarn v1.5.1. Despite the low version CI seemed to execute fine (besides the SharedArrayBuffer error we've been seeing)

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,9 +12,6 @@ jobs:
             - checkout
             - run: echo 'export PATH=$HOME/CIRCLE_PROJECT_REPONAME/node_modules/.bin:$PATH' >> $BASH_ENV
             - run:
-                  name: install-yarn
-                  command: sudo npm install --global yarn@1.9.4
-            - run:
                   name: yarn
                   command: yarn --frozen-lockfile --ignore-engines install || yarn --frozen-lockfile --ignore-engines install
             - setup_remote_docker


LogvinovLeon:
Pipeline tests failed. I've rerun them. Let's see if it helps

albrow:
I'm not sure if this is a good idea. It seems important to use the same version of Yarn on CI that we use locally.

BMillman19:
@albrow hmm, seems like the other solutions are already what we're doing? besides the `yarnpkg` one

albrow:
There are a few workarounds proposed in that thread that AFAIK we haven't tried yet.

1. Run `sudo u+x /usr/local/bin/yarn` after installing yarn to manually force certain permissions.
2. Run `sudo npm install -g yarn && yarn install` in a single step instead of installing yarn separately in an `install-yarn` step.

albrow:
Superseded by https://github.com/0xProject/0x-monorepo/pull/1755, which fixes the issue while still allowing us to install a specific version of Yarn.

PR #1647, opened 2019-02-26
Sol-profiler ganache

LogvinovLeon:
## Description

This PR adds a workaround for [ganache issue](https://github.com/trufflesuite/ganache-core/issues/277). This means that now all the dev tools should be working with Ganache as well as Geth.

## Testing instructions

Try running metacoin profiling with Ganache.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-tracing-utils/src/utils.ts
+++ b/packages/sol-tracing-utils/src/utils.ts
@@ -85,34 +85,66 @@ export const utils = {
         if (_.isEmpty(structLogs)) {
             return structLogs;
         }
+        const reduceDepthBy1 = (structLog: StructLog) => ({
+            ...structLog,
+            depth: structLog.depth - 1,
+        });
+        let normalizedStructLogs = structLogs;
+        // HACK(leo): Geth traces sometimes returns those gas costs incorrectly as very big numbers so we manually fix them.
+        const normalizeStaticCallCost = (structLog: StructLog) => (
+            (structLog.op === OpCode.StaticCall) ? {
+                ...structLog,
+                gasCost: STATICCALL_GAS_COST,
+            } : structLog
+        );
+        // HACK(leo): Geth traces sometimes returns those gas costs incorrectly as very big numbers so we manually fix them.
+        const normalizeCallCost = (structLog: StructLog, idx: number) => {
+            if (structLog.op === OpCode.Call) {
+                const HEX_BASE = 16;
+                const callAddress = parseInt(structLog.stack[0], HEX_BASE);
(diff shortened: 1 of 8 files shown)


packages/sol-tracing-utils/src/utils.ts
```diff
@@ -85,34 +85,66 @@ export const utils = {
...
-        if (structLogs[0].depth === 1) {
-            // Geth uses 1-indexed depth counter whilst ganache starts from 0
-            const newStructLogs = _.map(structLogs, (structLog: StructLog, idx: number) => {
-                const newStructLog = {
+        const reduceDepthBy1 = (structLog: StructLog) => ({
+            ...structLog,
+            depth: structLog.depth - 1,
+        });
+        let normalizedStructLogs = structLogs;
+        // HACK(leo): Geth traces sometimes returns those gas costs incorrectly as very big numbers so we manually fix them.
```
fabioberger: Is there an issue you can link to?

packages/sol-tracing-utils/src/utils.ts
```diff
@@ -85,34 +85,66 @@ export const utils = {
...
+        const normalizeStaticCallCost = (structLog: StructLog) => (
+            (structLog.op === OpCode.StaticCall) ? {
+                ...structLog,
+                gasCost: STATICCALL_GAS_COST,
+            } : structLog
+        );
+        // HACK(leo): Geth traces sometimes returns those gas costs incorrectly as very big numbers so we manually fix them.
+        const normalizeCallCost = (structLog: StructLog, idx: number) => {
```
fabioberger:
```suggestion
        const normalizeCallCost = (structLog: StructLog, index: number) => {
```

packages/sol-tracing-utils/src/utils.ts
```diff
@@ -85,34 +85,66 @@ export const utils = {
...
+            if (structLog.op === OpCode.Call) {
+                const HEX_BASE = 16;
+                const callAddress = parseInt(structLog.stack[0], HEX_BASE);
```
fabioberger: What if `stack` is empty? If that could happen, can we handle that case?

LogvinovLeon: Stack will never be empty during the call OpCode because it pops elements of the stack

packages/sol-tracing-utils/src/utils.ts
LogvinovLeon: No. They don't reply anyway ;) ask @albrow XD

Issue #1645, opened 2019-02-26
Cannot read property undefined in sol-compiler


PaulRBerg:
## Expected Behavior

Contracts should be compiled correctly when the `compilerSettings` are set in `compiler.json`.

## Current Behavior

The following error is returned:

> TypeError: Cannot read property 'Numbers' of undefined

## Possible Solution

Well, if I remove the `compilerSettings` from `compiler.json` or the config file altogether, everything works fine. Perhaps tweaking the way the settings are interpreted would solve this issue.

## Steps to Reproduce

```
1. git clone git@github.com:PaulRBerg/sol-compiler-bug.git
2. cd sol-compiler-bug
3. npx @0x/sol-compiler
```

## Context

This has originally occurred in the [AZTEC][aztec-monorepo] repo, in the "protocol" package on the "feat-coverage" branch, but the behaviour is exactly the same in the [bug repo][bug-repo] I created.

## Your Environment

| Tool | Version |
| :--------: | :--------: |
| @0x/sol-compiler | 3.1.0 |
| npm | 6.7.0 |
| node | 10.15.0 |

[aztec-monorepo]: https://github.com/AztecProtocol/AZTEC/tree/feat-coverage
[bug-repo]: https://github.com/PaulRBerg/sol-compiler-bug/tree/bug-cyclical

LogvinovLeon:
Thanks for a repro. I'm able to repro. Looking into it

LogvinovLeon:
Ok. So there are two things here:
1. You should use `Numbers.sol` instead of `contracts/Numbers.sol` in compiler.json
2. The bug on our end: https://github.com/0xProject/0x-monorepo/pull/1652

PR #1652, opened 2019-02-26
Fix a bug when combining sol-compiler settings from different sources

LogvinovLeon:
## Description

If the user specifies some compiler settings in `compier.json` but doesn't specify the outputSelection - solidity produces 0 output.

This PR changes the way we merge the default configs in, so it should work.

FIxes https://github.com/0xProject/0x-monorepo/issues/1645

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-compiler/CHANGELOG.json
+++ b/packages/sol-compiler/CHANGELOG.json
@@ -1,5 +1,13 @@
 [
     {
+        "version": "3.1.1",
+        "changes": [
+            {
+                "note": "Fix a bug when combining compilerSettings from different sources",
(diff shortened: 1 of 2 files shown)


packages/sol-compiler/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "3.1.1",
+        "changes": [
+            {
+                "note": "Fix a bug when combining compilerSettings from different sources",
+            }
```
fabioberger: PR number

Issue #1650, opened 2019-02-26
[sol-cov] Coverage is not working for versions ^0.5.1 of solidity


PaulRBerg:
## Expected Behavior

The coverage report should be generated for versions 0.5.1 and above of solidity.

## Current Behavior

The coverage report is blank:

![@0x/sol-coverage][bug-img-url]

If you bump the solidity version to 0.5.2, 0.5.3 or 0.5.4 (state-of-the-art as of 26th Feb 2019), nothing changes. If you revert to 0.5.0 in `truffle-config` and the contract's pragmas, everything works okay.

## Steps to Reproduce

```
1. git clone git@github.com:PaulRBerg/sol-coverage-bug.git
2. cd sol-coverage-bug
3. npm install
4. npm run coverage
```

## Context

This has originally occurred in the [AZTEC][aztec-monorepo] repo, in the "protocol" package on the "feat-coverage" branch, but the behaviour is exactly the same in the [bug repo][bug-repo] I created.

## Your Environment

| Tool | Version |
| :--------: | :--------: |
| @0x/sol-coverage | 2.0.4 |
| solidity | 0.5.1 |
| npm | 6.7.0 |
| node | 10.15.0 |

[aztec-monorepo]: https://github.com/AztecProtocol/AZTEC/tree/feat-coverage
[bug-img-url]: https://i.imgur.com/cJ7wLp1.png
[bug-repo]: https://github.com/PaulRBerg/sol-coverage-bug

LogvinovLeon:
I've cloned your repo. Ran coverage. Saw a bunch of warnings from sol-coverage not being able to match bytecode.
<img width="908" alt="screen shot 2019-02-26 at 3 32 45 pm" src="https://user-images.githubusercontent.com/6204356/53454370-c9eec400-39db-11e9-920c-8d22a5a0b76a.png">

Started looking into that:
* Compiler contracts with truffle using `truffle compile`.
* Compiler contracts with remix.
* Compiled contracts with sol-compiler.

It looks like truffle produces a different bytecode.
I don't know where does it come from, but:
Sol-compiler and remix produce that:
[...]

PR #1654, opened 2019-02-27
Fix a bug when TruffleArtifactAdapter was incorrectly parsing the truffle config of the newest version

LogvinovLeon:
## Description

Truffle recently released a major new version which changed the format of `truffle-config.json`.
This PR adds support for that new format.
Fixes: https://github.com/0xProject/0x-monorepo/issues/1650

## Types of changes

## Checklist:


packages/sol-tracing-utils/src/artifact_adapters/truffle_artifact_adapter.ts
```diff
@@ -70,6 +79,16 @@ export class TruffleArtifactAdapter extends AbstractArtifactAdapter {
             );
         }
     }
+    private _getTruffleSolcSettings(): Partial<solc.CompilerSettings> {
+        const truffleConfig = this._getTruffleConfig();
```
fabioberger: Please add add a comment here that this logic is here to support two versions of truffle's config file. Otherwise the next dev will be scratching his head.

PR #1658, opened 2019-02-28
Added bottleneck for EDPS to deal with rate-limiting

askeluv:
## Description

We weren't getting all the data we wanted when using ethereum-dex-prices-service (EDPS). The problem was rate limiting (primarily from Radar Relay). This PR adds a bottleneck so we do 1 request per second.

## Testing instructions

```
yarn build
yarn test
```

## Types of changes

* Bug fix (non-breaking change which fixes an issue) -> we weren't getting all the data we wanted

### DIFF
--- a/packages/pipeline/src/data_sources/dex_prices/index.ts
+++ b/packages/pipeline/src/data_sources/dex_prices/index.ts
@@ -1,4 +1,4 @@
-import { fetchAsync, logUtils } from '@0x/utils';
+import { fetchAsync } from '@0x/utils';
 import Bottleneck from 'bottleneck';
 
 const ONE_SECOND = 1000;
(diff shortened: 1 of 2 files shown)


packages/pipeline/src/data_sources/dex_prices/index.ts
```diff
@@ -21,12 +23,21 @@ export interface EdpsExchange {
...
+        });
+    }
+
     /**
      * Call Ethereum DEX Price Service API.
      */
     public async getEdpsAsync(direction: string, symbol: string, amount: number): Promise<EdpsWrapper> {
         const edpsUrl = `${EDPS_BASE_URL}/${direction}?amount=${amount}&symbol=${symbol}&decimals=`;
-        const resp = await fetchAsync(edpsUrl);
+        logUtils.log(`${direction} ${amount} ${symbol}`);
```
feuGeneA: Did you mean to leave this in?

askeluv: No 🙈- nicely spotted. Thanks! Fixed now.

PR #1659, opened 2019-02-28
Adapt to Bloxy API change

feuGeneA:
## Description

We were using the `days` parameter, but that's no longer available in
the API, so our fetching of previous days (before "today") was broken.

This change gets rid of the concept of "days", and uses the API's new
`from_date` and `till_date` parameters to fetch previous days' trades.

The change to the query in `getLastSeenTimestampAsync()` was necessary
because it was returning a string (despite the `as Array<{ tx_timestamp:
number }>` cast), which was later causing problems trying to pass that
value into `new Date()`.  It worked before because we were doing some
math operations on it (multiplying it by some numbers), so coercion was
saving us.  With the change from a raw query to a `typeorm` call, the
`numberToBigIntTransformer` specified in the `DexTrade` entity is now
doing the proper type conversion for us.

The new `MAX_DAYS` constant in `pull_competing_dex_trades.ts` is
necessary in order to avoid trying to pull data all the way back to 1969
:D, which induces the API to return an HTTP 503.

## Testing instructions

```
PKG=@0x/pipeline yarn build
node packages/pipeline/lib/src/scripts/pull_competing_dex_trades.js
```
If your database is empty, the script will take a little while.

## Types of changes

### DIFF
--- a/packages/pipeline/src/scripts/pull_competing_dex_trades.ts
+++ b/packages/pipeline/src/scripts/pull_competing_dex_trades.ts
@@ -12,6 +12,11 @@ import { handleError } from '../utils';
 // Number of trades to save at once.
 const BATCH_SAVE_SIZE = 1000;
 
+// maximum number of days into the past for which trades will be pulled. this
+// is arbitrary and can be tweaked, but deep history takes a significant amount
+// of time to pull.
+const MAX_DAYS = 30;
+
 let connection: Connection;
 
 (async () => {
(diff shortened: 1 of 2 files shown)


packages/pipeline/src/scripts/pull_competing_dex_trades.ts
```diff
@@ -39,14 +44,29 @@ async function getAndSaveTradesAsync(): Promise<void> {
 }
 
 async function getLastSeenTimestampAsync(tradesRepository: Repository<DexTrade>): Promise<number> {
+    const hoursPerDay = 24;
+    const minutesPerHour = 60;
+    const secondsPerMinute = 60;
+    const millisecondsPerSecond = 1000;
+    const millisecondsPerDay = millisecondsPerSecond * secondsPerMinute * minutesPerHour * hoursPerDay;
```
albrow: IMO we don't need to be quite so verbose here in declaring the different units.

```
const millisecondsPerDay = 1000 * 60 * 60 * 24;
```

Seems clear enough. But this is also subjective.

packages/pipeline/src/scripts/pull_competing_dex_trades.ts
```diff
@@ -39,14 +44,29 @@ async function getAndSaveTradesAsync(): Promise<void> {
...
+
+    const oldestTimestampWilling = Date.now() - MAX_DAYS * millisecondsPerDay;
+
     if ((await tradesRepository.count()) === 0) {
-        return 0;
+        return oldestTimestampWilling;
```
xianny: does `oldestTimestampWilling` have a special meaning here? why not `oldestFetchableTimestamp`?

feuGeneA: Because it's not the oldest one that can be fetched; in theory we can fetch arbitrarily old data.  I said "willing" because it's the oldest timestamp we're willing to go back to, per the MAX_DAYS constant defined above.

PR #1657, opened 2019-02-27
erc1155 minimal implementation

hysz:
## Description

This was created for testing the ERC1155 proxy. 
- The implementation is derived from the [ERC-1155 Reference Implementation](https://github.com/enjin/erc-1155).
- Changes to reference implementation include: bug fixes / tests / style
- Supports both fungible & non-fungible tokens

## Checklist:

-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #1661, opened 2019-02-28
ERC1155 Asset Proxy

hysz:
## Description

Asset Proxy for transferring ERC1155 assets. ERC1155 allows a set of fungible / non-fungible assets to exist under the same contract. The proxy uses ERC1155's `safeBatchTransferFrom` to transfer a set of ERC1155 assets between accounts. See #1657 for a minimal ERC1155 implementation.

Note - integration tests with the Exchange and MultiAsset Proxy in a separate PR.

## Checklist:


contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
...
+                // 4 (function selector)
+                // + assetDataOffset
+                let assetDataLength := calldataload(add(4, assetDataOffset))
+
+                // This corresponds to the beginning of the Data Area for Table #3.
+                // Computed by:
+                // 4 (function selector)
+                // + assetDataOffset
+                // + 32 (length of assetData)
+                calldatacopy(32, add(36, assetDataOffset), assetDataLength)
```
merklejerk: You're probably right, but why not `calldatacopy(68, add(36, assetDataOffset), sub(assetDataLength, 36))`?

hysz: Good question - this line copies the entire assetData into memory at offset 32. Memory allocation looks like this:
```
32    :    assetProxyId 
36    :    address of ERC1155 contract
68    :    offset to tokenIds
...
```

This is later transformed into:
```
00    :    batchTransferFrom selector
04    :    from address
36    :    to address
68    :    offset to tokenIds
...
```

If we replaced this line with `calldatacopy(68, add(36, assetDataOffset), sub(assetDataLength, 36))` then we would have the following allocation:
```
68    :    assetProxyId 
72    :    address of ERC1155 contract
104   :    offset to tokenIds
...
```
And it would truncate 36 bytes at the tail end of the asset data.

Note that `assetData` is a byte array so it is abi encoded as:
```
0   : length
32  : contents
```

merklejerk: Thanks! Super appreciate the walkthrough.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
...
+        //
+        //
+        // Asset data is encoded as follows:
+        //                      -- TABLE #2 --
+        // | Area     | Offset      | Length  | Contents                            |
+        // |----------|-------------|---------|-------------------------------------|
+        // | Header   | 0           | 4       | assetProxyId                        |
+        // | Params   |             | 4 * 32  | function parameters:                |
+        // |          | 4           |         |   1. address of ERC1155 contract    |
+        // |          | 36          |         |   2. offset to tokenIds (*)         |
```
abandeali1: I feel like these param names should be `ids`, `values`, and `data` to be consistent with the naming of the actual ERC1155 transfer params.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
abandeali1: I personally like annotating `calldatacopy` like we do in the MAP here: https://github.com/0xProject/0x-monorepo/blob/development/contracts/asset-proxy/contracts/src/MultiAssetProxy.sol#L161

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
...
+                for { let tokenValueOffset := tokenValuesBegin }
+                    lt(tokenValueOffset, tokenValuesEnd)
+                    { tokenValueOffset := add(tokenValueOffset, 32) }
+                {
+                    // Load token value and generate scaled value
+                    let tokenValue := mload(tokenValueOffset)
+                    let scaledTokenValue := mul(tokenValue, scaleAmount)
+                    
+                    // Check if scaled value is zero
+                    if iszero(scaledTokenValue) {
```
abandeali1: We don't actually check if the `amount` is 0 in any of the other proxies. What is the reason for doing so here, and can this ever be potentially limiting?

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
...
+                        mstore(32, 0x0000002000000000000000000000000000000000000000000000000000000000)
+                        mstore(64, 0x000000235452414e534645525f475245415445525f5448414e5f5a45524f5f52)
+                        mstore(96, 0x4551554952454400000000000000000000000000000000000000000000000000)
+                        mstore(128, 0)
+                        revert(0, 132)
+                    }
+
+                    // Check for multiplication overflow
+                    let expectedTokenValue := div(scaledTokenValue, scaleAmount)
+                    if iszero(eq(expectedTokenValue, tokenValue)) {
```
abandeali1: Just a note that if we do remove the last check, we'll still need to check if `scaleAmount` is 0 before doing this division.

packages/types/CHANGELOG.json
```diff
@@ -1,4 +1,22 @@
 [
+    {
+        "version": "2.2.3",
+        "changes": [
+            {
+                "note": "Added `ERC1155AssetData`, `ERC1155AssetDataNoProxyId`, and `ERC1155AssetDataAbi`",
```
abandeali1: This change can be merged with the previous entry (not yet published).

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity 0.4.24;
```
abandeali1: We should test gas differences with a newer version of Solidity too.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
abandeali1: Meh, actually we aren't really consistent in the other proxies either.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -1,5 +1,8 @@
 import {
     AssetProxyId,
+    ERC1155AssetData,
```
abandeali1: You'll need to reexport these in the `index.ts` of this package and 0x.js (see doc-gen failing tests).

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
hysz: I like this +1.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
hysz: Yeah it's only ~100 gas cheaper if we use solidity 0.5.5.

Additionally, I believe all the proxies have to use the same version since they're in the same package, ya? If this is the case and the savings become more substantial it could be worthwhile to split proxies into standalone packages.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
hysz: A check has been added above, right after we load `amount`.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,271 @@
...
+                // Record the address of the destination erc1155 asset for later.
+                let assetAddress := and(
+                    mload(36),
+                    0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff
+                )
+
+                ////////// STEP 2/4 //////////
+                // load amount and ensure it's non-zero (we divide by amount below)
+                let amount := calldataload(100)
+                if iszero(amount) {
```
abandeali1: This is how we handle this case in the MAP:

```
// Revert if `amount` != 0 and multiplication resulted in an overflow 
if iszero(or(
    iszero(amount),
    eq(div(totalAmount, amount), amountsElement)
)) {
    // Revert with `Error("UINT256_OVERFLOW")`
    mstore(0, 0x08c379a000000000000000000000000000000000000000000000000000000000)
    mstore(32, 0x0000002000000000000000000000000000000000000000000000000000000000)
    mstore(64, 0x0000001055494e543235365f4f564552464c4f57000000000000000000000000)
    mstore(96, 0)
    revert(0, 100)
}
```

So we don't actually revert if amount is 0, instead we don't check for an overflow. I'd prefer switching to this for the sake of consistency.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,271 @@
...
+                ////////// STEP 3/4 //////////
+                // Store the safeBatchTransferFrom function selector,
+                // and copy `from`/`to` fields from Table #1 to Table #3.
+
+                // The function selector is computed using:
+                // bytes4(keccak256("safeBatchTransferFrom(address,address,uint256[],uint256[],bytes)"))
+                mstore(0, 0x2eb2c2d600000000000000000000000000000000000000000000000000000000)
+
+                // Copy `from` and `to` fields from Table #1 to Table #3
+                calldatacopy(4, 36, 64)
```
abandeali1: Minor nit: please annotate this too.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
hysz: Ah, okay. That makes sense. So we revert only when `amount` is non-zero and there is an overflow. 
```
amount=0 | no overflow | amount=0 OR no overflow | iszero(amount=0 OR no overflow)
   1           1               1                       0
   1           0               1                       0
   0           1               1                       0
   0           0               0                       1
```
And then when `amount=0` the `totalAmount` is also zero.

PR #1663, opened 2019-03-01
Sol profiler improvements and bug fixes

LogvinovLeon:
## Description

The last sol-profiler PR fixes some stuff and breaked some other stuff. It was caused by me testing it only against Geth and truffle post 5.0. This PR fixes those issues, adds performance optimizations etc.

After this PR sol-compiler works for the DEX contract.
It's also much faster now.

One of the big changes in that PR is fixing the stack ordering. Now both Ganache and Geth stack are ordered upside-down. We also use constants defined in one place for stack param offsets instead of hardcoding 1s and 2. all over the codebase.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sol-profiler/src/profiler_subprovider.ts
+++ b/packages/sol-profiler/src/profiler_subprovider.ts
@@ -126,7 +126,7 @@ export const profilerHandler: SingleFileSubtraceHandler = (
     const statementToGasConsumed: { [statementId: string]: number } = {};
     const statementIds = _.keys(profilerEntriesDescription.statementMap);
     // `interestingStructLogs` are those that map back to source ranges within the current file.
-    // It also doesn't include one that map back to nowhere
+    // It also doesn't include any that cannot be mapped back
     // This is a perf optimization reducing the work done in the loop over `statementIds`.
     // TODO(logvinov) Optimize the loop below.
     const interestingStructLogs = _.filter(subtrace, structLog => {
--- a/packages/sol-profiler/src/profiler_subprovider.ts
+++ b/packages/sol-profiler/src/profiler_subprovider.ts
@@ -65,7 +65,7 @@ export class ProfilerSubprovider extends TraceInfoSubprovider {
         const transactionBaseCost = BASE_COST;
         let totalCost = callDataCost + opcodesCost + BASE_COST;
         logUtils.header('Final breakdown', '-');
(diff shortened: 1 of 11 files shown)


packages/sol-profiler/src/profiler_subprovider.ts
```diff
@@ -125,17 +125,22 @@ export const profilerHandler: SingleFileSubtraceHandler = (
     const profilerEntriesDescription = collectCoverageEntries(contractData.sourceCodes[fileIndex]);
     const statementToGasConsumed: { [statementId: string]: number } = {};
     const statementIds = _.keys(profilerEntriesDescription.statementMap);
+    // `interestingStructLogs` are those that map back to source wanges within the current file.
```
fabioberger:
```suggestion
    // `interestingStructLogs` are those that map back to source ranges within the current file.
```

packages/sol-profiler/src/profiler_subprovider.ts
```diff
@@ -125,17 +125,22 @@ export const profilerHandler: SingleFileSubtraceHandler = (
...
+    // It also doesn't include one that map back to nowhere
```
fabioberger:
```suggestion
    // It also doesn't include any that cannot be mapped back
```

packages/sol-profiler/src/profiler_subprovider.ts
```diff
@@ -125,17 +125,22 @@ export const profilerHandler: SingleFileSubtraceHandler = (
...
+    // This is a perf optimization reducing the work done in the loop over `statementIds`.
+    // TODO(logvinov) Optimize the loop below.
```
fabioberger:
```suggestion
    // TODO(logvinov): Optimize the loop below.
```

packages/sol-tracing-utils/CHANGELOG.json
```diff
@@ -1,4 +1,17 @@
 [
+    {
+        "version": "6.0.7",
+        "changes": [
+            {
+                "note": "Fix a bug when `TruffleArtifactAdapter` wasn't correctly parsing solc config in the pre 5.0 version of trufle",
```
fabioberger:
```suggestion
                "note": "Fix a bug when `TruffleArtifactAdapter` wasn't correctly parsing solc config in pre-5.0 versions of Truffle",
```

PR #1673, opened 2019-03-06
 Exchange & Proxy Integration Tests - ERC1155

hysz:
## Description

Integration tests for the ERC1155 proxy with the Mutli-Asset Proxy and Exchange.

## Checklist:


contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -0,0 +1,268 @@
+/*
```
fabioberger: How this already been reviewed elsewhere? Are we duplicating Solidity source code within the monorepo?

packages/types/CHANGELOG.json
```diff
@@ -1,4 +1,22 @@
...
+    },
+    {
+        "version": "2.2.2",
+        "changes": [
+            {
+                "note": "Added ERC1155 revert reasons",
+                "pr": 1657
+            }
+        ]
+    },
```
fabioberger: These should be one CHANGELOG entry.

fabioberger:
Eventually we'll also need to add ERC1155 support to OrderWatcher.

contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
hysz: Yup! Been reviewed in #1661. Just had to do a cheeky rebase.

contracts/exchange/test/core.ts
```diff
@@ -1040,6 +1086,435 @@ describe('Exchange core', () => {
...
+                // takerAddress / takerToken[1]
+                expectedInitialBalances[10].minus(totalTakerValuesTransferred[1]),
+                // takerAddress / takerToken[2]
+                expectedInitialBalances[11].minus(totalTakerValuesTransferred[2]),
+            ];
+            await erc1155Wrapper.assertBalancesAsync(tokenHolders, tokensToTransfer, expectedFinalBalances);
+        });
+        it('should allow an order exchanging erc1155 assets to be partially filled', async () => {
+            // NOTICE:
+            // As-per the eip1155 standard, there is no way to distinguish between a fungible or non-fungible erc1155 assets.
```
abandeali1: Shouldn't a partial transfer of a non-fungible still fail with `TRANSFER_FAILED`? Whatever the behavior is, we should add a test for it.

hysz: We chatted about this offline.

Partial fills on non-fungible assets should be prohibited by the maker by setting the amount to 1. The proxy cannot distinguish between fungible and non-fungible assets, so it cannot force partial fills to fail in this scenario. If the amount is _not_ set to 1 and an asset is non-fungible then the behavior depends on the implementation of the ERC1155 contract.

PR #1678, opened 2019-03-06
Fix contracts after Solidity 0.5.5 update

hysz:
## Description

Contracts aren't building. This (hopefully) fixes it.

## Checklist:


hysz:
@merklejerk yeah I just added it to all of the packages for consistency. Also so they won't break when we do upgrade.

PR #1680, opened 2019-03-07
Enable cryptokitties data from nonfungible.com

feuGeneA:
## Description

Recall that nonfungible.com sent us an initial data dump so that we wouldn't have to pull *every*thing from the API.  And recall that we're storing that initial dump on S3.  Finally, recall that the cryptokitties data was particularly voluminous (~1GB), which doesn't easily transfer to and from S3.

These changes provide a CLI script to partition that data (or any such data from nonfungible.com) into a set of smaller files; and a scraping script that seamlessly recombines those partitions for loading into the database.

## Testing instructions

After building the `pipeline` component and setting `ZEROEX_DATA_PIPELINE_DB_URL` appropriately,
`node --max-old-space-size=8192 lib/src/scripts/pull_nonfungible_dot_com_trades.js`

## Types of changes

## Checklist:

### DIFF
--- a/packages/pipeline/src/scripts/partition_nonfungible_dot_com_dump.ts
+++ b/packages/pipeline/src/scripts/partition_nonfungible_dot_com_dump.ts
@@ -33,14 +33,15 @@ import {
 (() => {
     const publisher = process.argv[2];
 
-    const sourceJson: NonfungibleDotComHistoryResponse = JSON.parse(
-        readFileSync(`sales_summary_${publisher}.json`).toString(),
-    );
+    const inputFilename = `sales_summary_${publisher}.json`;
+    logUtils.log(`Reading input file ${inputFilename}`);
+    const sourceJson: NonfungibleDotComHistoryResponse = JSON.parse(readFileSync(inputFilename).toString());
 
-    const chunks: NonfungibleDotComTradeResponse[][] = splitEvery(S3_CHUNK_SIZES[publisher], sourceJson.data);
+    const chunkSize = S3_CHUNK_SIZES[publisher];
+    logUtils.log(`Splitting data into chunks of ${chunkSize} trades each`);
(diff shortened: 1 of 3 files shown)


feuGeneA:
Noticed 3 new NFTs that are now supported on nonfungible.com and added those in here too.

packages/pipeline/src/scripts/partition_nonfungible_dot_com_dump.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Writes to disk a set of files named `sales_summary_${publisher}${N}.json`.
+ *
+ * Probably need to use `node` with --max-old-space-size=1024 or maybe
+ * even more.
+ */
+
+import { readFileSync, writeFileSync } from 'fs';
+
+import { chunk } from 'lodash';
```
xianny: Can we replace with [Ramda](https://ramdajs.com/docs/#splitEvery) so that we don't need to add lodash to a dep to this whole package?

feuGeneA:
@xianny thanks for all the great feedback!!

I'm with you that we can break out a dump parsing/loading module if/when we have to modify this again.  Backfilling gaps would be the most likely reason to do that, in my opinion.  As for idempotent data jobs, I'm not sure that will make sense, given that the nonfungible.com API doesn't offer a way to query for specific dates, and given that Airflow only supports automatic backfilling for date-partitioned jobs.

xianny:
Ugh, that makes it harder.

One thing we could do is have the `pull_nonfungibles...` script take a starting block number as a parameter. Then we could have another script (could be as simple as a lamda firing off a cloudwatch event) trigger it while passing in the start block number. The trigger script would be the one controlled from airflow. 

Cross that bridge when we come to it :)

PR #1682, opened 2019-03-07
Upgrade contracts to Solidity 0.5.5

abandeali1:
## Description

- Upgrades most contracts to Solidity 0.5.5 (except those that will not be redeployed)
- Makes the minimum necessary changes to get the contracts to compile and pass tests with the new Solidity version
- Uses new optimizer settings (saves ~3.5k gas / fill!)
- Updates `ganache-cli` and `solcjs` versions
- No longer re-export external contract dependencies within each package (this was causing incompatibility between packages)
- Note: this still compiles with `evmVersion: 'byzantium'`. It appears that compiling with 0.5.5 introduces some new opcodes that cause the contracts to not deploy. Updating `ganache-cli` did not fix the issue (even though it supports Constantinople). This should be fixed in a separate PR.

## Checklist:


merklejerk:
`exchange-forwarder` is still at 0.4.24. Is this intentional?

abandeali1:
@merklejerk that is intentional, but we'll make the change when the Forwarder is ready to be updated.

PR #1688, opened 2019-03-09
[WIP] Rich Revert Reasons PoC (Exchange only for now)

merklejerk:
## Description

This is a proof-of-concept migration from regular string-based revert reasons to much more informative, ABI-encoded revert reasons. This is based around discussions of [EIP 838](https://github.com/ethereum/EIPs/issues/838).

**NOTE** I've only migrated the 0.4.x version of the `exchange` contracts because ATM other contract packages in `development` use a conflicting 0.5.x version of Solidity. Once PR #1682 for modernizing all the legacy contracts gets merged, I can start migrating en masse. But before then, we should have a discussion about this approach, because the implementation has some pain points and touches several packages.

#### Some Background

Current Solidity revert strings are already returned as calldata ABI-encoded bytes with a method signature of `Error(string)`, but you can actually return any arbitrary bytes with an assembly call to `revert()`. 

Revert data can only be reliably recovered during an `eth_call`, because transactions only produce a hash and receipt, with a single status bit indicating success or failure. But a significant issue we've run into is just detecting a `revert` during an `eth_call`, particularly with Geth. A Geth node will simply return the revert bytes, with no signal that the call reverted. Parity supposedly returns a signal, but, as we don't currently test against Parity, I cannot personally confirm this. The current hack (already implemented in `/packages/base-contract` is to basically check if the returned data starts with the `bytes4` selector of the standard Error signature.

#### Where We Are Now

There have been discussions around various schemes of returning rich revert data. The EVM does not seem to sanitize what gets returned via `revert()`, so many options are on the table. Ultimately, for this PoC, we've settled on simply ABI-encoding against custom method signatures for specific Exchange errors. One common example would be `OrderStatusError(string,uint8)`, which encodes an order's hash and current status. This is thrown whenever the order is in an incompatible state during an operation. 

While this approach is relatively simple and wire-compact, it does require prior knowledge of all revert signatures to semi-confidently discern which return values are true reverts and which are just arbitrary bytes. It is also not backwards compatible-- clients that only look for `Error(string)` signatures will likely either fail at the decoding step or miss these reverts entirely. 

Significant additions have been made to the `order-utils`, `test-utils`, and `abi-gen-templates` packages to help intercept, decode, construct, and test against these new revert formats. Small modifications were also made to other sundry packages so they could anticipate the new reverts and pass their tests. 

For the Solidity implementation, please take a look at [/contracts/exchange/.../MRichErrors.sol](/merklejerk/0x-monorepo/blob/feature/contracts/eip-838-rich-reverts-poc/contracts/exchange/contracts/src/mixins/MRichErrors.sol). Usage is peppered among the various [Exchange contracts](/merklejerk/0x-monorepo/blob/feature/contracts/eip-838-rich-reverts-poc/contracts/exchange/contracts/src) (everywhere there used to be a `require()` or `revert()`).

The Typescript implementation currently resides at [`/packages/order-utils/.../rich_reverts.ts](https://github.com/merklejerk/0x-monorepo/blob/feature/contracts/eip-838-rich-reverts-poc/packages/order-utils/src/rich_reverts.ts). Some [tests](https://github.com/merklejerk/0x-monorepo/blob/feature/contracts/eip-838-rich-reverts-poc/packages/order-utils/test/rich_reverts_test.ts) also exist to demonstrate usage (though more could be used).

There is also a [chai extension](https://github.com/merklejerk/0x-monorepo/blob/b1127c5cde4a42f60e7c5a24aee1358d2048f417/packages/contract-wrappers/test/utils/chai_setup.ts#L17) implemented in `contract-wrappers` that demonstrates integration with chai. This will probably be moved to `test-utils` in future commits because that makes more sense.

## Testing instructions

Just run the regular monorepo test suite since this PR spans many packages.

## Types of changes

## Checklist:


packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+    BadSignature,
+    FailedExecution,
+}
+
+const ADDRESS_SIZE = 20;
+
+/**
+ * Base type for rich revert reasons.
+ */
+export abstract class RichRevertReason {
```
abandeali1: Does it make sense for this to be in the `order-utils` package? `contract-wrappers` feels more natural to me since we're decoding our contact's reverts specifically.

hysz:
> it does require prior knowledge of all revert signatures to semi-confidently discern which return values are true reverts and which are just arbitrary bytes
 
I wonder if it would make sense to use a generic signature, like `ZeroExError(bytes)`, where the input `bytes` is a nested encoding of the error message. Now we have a single 4-byte selector to infer an error state.

Example:
```
errorMessage = abiEncode(OrderStatusError(string,uint8))`
zeroExErrorMessage = abiEncode(ZeroExError(errorMessage))
```

If this could be beneficial, it's worth noting that it shouldn't be too much additional cost. We just prepend the selector, a generic offset (`0x..20`) and the length of `errorMessage` (which we should already know).

contracts/exchange/contracts/src/mixins/MRichErrors.sol
```diff
@@ -0,0 +1,321 @@
...
+    // solhint-enable func-name-mixedcase
+
+    /// @dev Reverts an encoded rich revert reason `errorData`.
+    /// Use any of the *Error functions defined in this contract to
+    /// generate a rich revert reason.
+    /// Evaluating the testg condition in an if block then calling this
+    /// may be preferable (gas-wise) to rrequire()
+    /// since you can avoid the needless abi encoding if the test passes.
+    /// @param errorData ABI encoded error data.
+    function rrevert(bytes memory errorData)
```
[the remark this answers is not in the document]
merklejerk: Definitely. Will move them once all contracts agree on solidity version.

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -377,27 +378,21 @@ contract MixinExchangeCore is
     {
         // Revert if fill amount is invalid
         // TODO: reconsider necessity for v2.1
-        require(
-            takerAssetFillAmount != 0,
-            "INVALID_TAKER_AMOUNT"
-        );
-        
+        if (takerAssetFillAmount == 0)
+            rrevert(FillError(orderInfo.orderHash, FillErrorCodes.INVALID_TAKER_AMOUNT));
```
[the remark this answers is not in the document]
merklejerk: Because it would look like `rrequire(takerAssetFillAmount != 0, FillError(...))`.  It's a nicer experience, but it means the revert reason has to be ABI encoded before the condition is checked, prematurely wasting a lot of gas.

merklejerk:
@hysz I like this. Much less chance of collisions with the selector bytes.

contracts/test-utils/CHANGELOG.json
```diff
@@ -1,4 +1,18 @@
 [
+    {
+        "version": "3.1.0",
+        "changes": [
+            {
+                "note": "Create RichRevertReason classes"
```
LogvinovLeon: We sometimes use markdown backticks on identifiers here so that it looks nice on github and discord later.

contracts/test-utils/CHANGELOG.json
```diff
@@ -1,4 +1,18 @@
...
+            },
+            {
+                "note": "Update assertions to support RichRevertReason objects"
```
LogvinovLeon:
```suggestion
                "note": "Update `assertions` to support `RichRevertReason` objects"
```
Like that. Just a nit

packages/abi-gen-templates/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "timestamp": 1551130135,
```
LogvinovLeon:
```suggestion
```

packages/abi-gen-templates/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
...
+        "version": "2.1.0",
+        "changes": [
+            {
+                "note": "Add 'returnData' field to thrown error when failing to decode call return values."
```
LogvinovLeon:
```suggestion
                "note": "Add `returnData` field to thrown error when failing to decode call return values."
```

packages/ethereum-types/src/index.ts
```diff
@@ -116,6 +116,13 @@ export interface EventParameter extends DataItem {
     indexed: boolean;
 }
 
+export interface RichRevertAbi {
+    // Should be set to `'error'`.
+    type: string;
```
LogvinovLeon:
```suggestion
    type: 'error';
```

packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    public toString(): string {
+        const values = _.omitBy(this.values, v => _.isNil(v));
+        const inner = _.isEmpty(values) ? '' : inspect(values);
+        return `${this.constructor.name}(${inner})`;
+    }
+}
+
+export class StandardError extends RichRevertReason {
```
LogvinovLeon: Ok. Now it makes sense why it's a class

packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+    new FillError(),
+    new EpochOrderError(),
+    new AssetProxyExistsError(),
+    new AssetProxyDispatchError(),
+    new NegativeSpreadError(),
+    new TransactionExecutionError(),
+    new IncompleteFillError(),
+];
+
+const RICH_REVERT_LUT: ObjectMap<RichRevertReason> = _.zipObject(
```
LogvinovLeon: What is `LUT`?

packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+    _.map(RICH_REVERT_REGISTRY, r => r.selector),
+    RICH_REVERT_REGISTRY,
+);
+
+const DECODER_CACHE: ObjectMap<Decoder> = _.zipObject(
```
LogvinovLeon: Our naming convention for Objects is stuffToStuff1ToStuff2 etc...
For example `selectorToDecoder`

packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+    _.map(RICH_REVERT_REGISTRY, r => r.selector),
+    _.map(RICH_REVERT_REGISTRY, r => createDecoder(r.abi)),
+);
+
+function checkArgEquality(type: string, a: ArgTypes, b: ArgTypes): boolean {
```
LogvinovLeon: Nit: Maybe name those params `lhs` and `rhs`?left hand side

packages/order-utils/src/rich_reverts.ts
```diff
@@ -0,0 +1,376 @@
...
+
+function createDecoder(abi: RichRevertAbi): (hex: string) => ValueMap {
+    const encoder = AbiEncoder.createMethod(abi.name, abi.arguments || []);
+    return (hex: string): ValueMap => {
+        // tslint:disable-next-line
+        return encoder.decode(hex) as ValueMap;
+    };
+}
+
+function declarationToAbi(decl: string): RichRevertAbi {
```
LogvinovLeon: Ethers.js should have a function to parse the ABI string. Check my prev PR

packages/order-utils/test/rich_reverts_test.ts
```diff
@@ -0,0 +1,111 @@
...
+import * as RichReverts from '../src/rich_reverts';
+
+import { chaiSetup } from './utils/chai_setup';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const assert = chai.assert;
+
+// tslint:disable:custom-no-magic-numbers
+describe.only('RichRevertReasons', () => {
```
LogvinovLeon:
```suggestion
describe('RichRevertReasons', () => {
```

packages/types/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "2.3.0",
+        "changes": [
+            {
+                "note": "Move OrderStatus enum from /contracts/test-utils into here"
+            }
+        ]
+    },
```
LogvinovLeon: Merge with prev entry

contracts/exchange/contracts/src/MixinExchangeCore.sol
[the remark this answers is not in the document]
merklejerk: Maybe. Will check on remix.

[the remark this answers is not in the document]
merklejerk: Yup, `condition || rrevert(Error(stuff));` works. So that's an option if people want a more compact alternative.

contracts/exchange/contracts/src/mixins/MRichErrors.sol
```diff
@@ -0,0 +1,321 @@
...
+    /// @param errorData ABI encoded error data.
+    function rrequire(
+        bool success,
+        bytes memory errorData
+    )
+        internal
+        pure
+    {
+        assembly {
+            if iszero(and(success, 0x00000000000000000000000000000000000000000000000000000000000000FF)) {
```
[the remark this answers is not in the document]
merklejerk: This is my concern: https://solidity.readthedocs.io/en/v0.5.5/security-considerations.html#minor-details

I tried it out in remix and function arguments don't get bitmasked before hitting the body, so the dirty higher bits passed into `rrequire()` persist.

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -415,12 +410,11 @@ contract MixinExchangeCore is
...
-        require(
-            safeMul(makerAssetFilledAmount, order.takerAssetAmount)
-            <= 
-            safeMul(order.makerAssetAmount, takerAssetFilledAmount),
-            "INVALID_FILL_PRICE"
-        );
+        if (safeMul(makerAssetFilledAmount, order.takerAssetAmount)
+                >
+                safeMul(order.makerAssetAmount, takerAssetFilledAmount)) {
+            rrevert(FillError(orderInfo.orderHash, FillErrorCodes.INVALID_FILL_PRICE));
```
fabioberger: Technically we'd need `takerAssetFilledAmount` and `makerAssetFilledAmount` here since without those values, one couldn't figure out what went wrong. However, this is a cautionary assertion that should never hit, so maybe we can pass on this one.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -139,7 +141,7 @@ contract MixinMatchOrders is
         LibOrder.Order memory rightOrder
     )
         internal
-        pure
+        view
```
fabioberger: Why was this changed?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -244,32 +249,26 @@ contract MixinSignatureValidator is
...
+            if eq(success, 1) {
                 // Signature is valid if call did not revert and returned true
                 isValid := mload(cdStart)
             }
         }
-        return isValid;
+        if (success)
+            return isValid;
+        rrevert(
+            SignatureError(hash, SignatureErrorCodes.WALLET_ERROR));
```
fabioberger: This feels a bit unorthodox to me. Can we do:

```
if (!success) {
            rrevert(SignatureError(hash, SignatureErrorCodes.WALLET_ERROR));
}
return isValid;
```

Another nit: it would be nice if boolean naming conventions followed the rest of our codebase. Perhaps `didSucceed`?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -294,31 +293,25 @@ contract MixinSignatureValidator is
...
+            if eq(success, 1) {
                 // Signature is valid if call did not revert and returned true
                 isValid := mload(cdStart)
             }
         }
-        return isValid;
+        if (success)
+            return isValid;
+        rrevert(
+            SignatureError(hash, SignatureErrorCodes.VALIDATOR_ERROR));
```
fabioberger: Ditto the above.

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -61,33 +61,27 @@ contract MixinTransactions is
         ));
 
         // Validate transaction has not been executed
-        require(
-            !transactions[transactionHash],
-            "INVALID_TX_HASH"
-        );
+        if (transactions[transactionHash])
+            rrevert(TransactionExecutionError(TransactionExecutionErrorCodes.ALREADY_EXECUTED));
```
fabioberger: Let's return the `transactionHash` as well.

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -61,33 +61,27 @@ contract MixinTransactions is
...
-                isValidSignature(
+            if (!isValidSignature(
                     transactionHash,
                     signerAddress,
-                    signature
-                ),
-                "INVALID_TX_SIGNATURE"
-            );
+                    signature)) {
+                rrevert(TransactionExecutionError(TransactionExecutionErrorCodes.BAD_SIGNATURE));
```
fabioberger: Ditto here re: `transactionHash`

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -61,33 +61,27 @@ contract MixinTransactions is
...
         }
 
         // Execute transaction
         transactions[transactionHash] = true;
-        require(
-            address(this).delegatecall(data),
-            "FAILED_EXECUTION"
-        );
+        if (!address(this).delegatecall(data))
+            rrevert(TransactionExecutionError(TransactionExecutionErrorCodes.FAILED_EXECUTION));
```
fabioberger: Ditto here too.

contracts/test-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/contracts-test-utils",
-    "version": "3.0.8",
+    "version": "3.1.0",
```
fabioberger: Not needed.

contracts/test-utils/src/assertions.ts
```diff
@@ -159,10 +187,76 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
...
     return expect(p).to.be.rejectedWith(reason);
 }
 
+/**
+ * A hacky way of expecting a rich revert reason to resolve.
+ * This is because geth does not signal when a revert happens during an
+ * eth_call so we're left to guess whether the returned data is an encoded
+ * rich revert reason.
+ * @param p A Promise resulting from a contract call.
+ * @param reason a rich revert reason
```
fabioberger:
```suggestion
 * @param reason A rich revert reason
```

Nit

contracts/test-utils/src/assertions.ts
```diff
@@ -159,10 +187,76 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
...
+ * @returns a promise which will reject `p` does not resolve to a rich revert reason.
```
fabioberger:
```suggestion
 * @returns a promise which will reject if `p` does not resolve to a rich revert reason.
```

contracts/test-utils/src/assertions.ts
```diff
@@ -159,10 +187,76 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
...
+ */
+async function _expectContractCallRichRevertAsync<T>(p: Promise<T>, reason: RichRevertReason): Promise<void> {
+    let decoded: RichRevertReason | undefined;
```
fabioberger: We have a variable naming convention that if a variable might hold some value or `undefined`, we suffix it's name with `IfExists`, i.e., `decodedValueIfExists`. We also generally prefer verbose, readable variable names over concise, potentially confounding names. Reading code is hard enough as it is 🍅

contracts/test-utils/src/assertions.ts
```diff
@@ -159,10 +187,76 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
...
+            decoded = new StandardError(err.message);
+        } else if (err.results) {
+            // Ganache threw a revert error.
+            const info = err.results[_.keys(err.results)[0]];
+            if (info.error === 'revert' && info.return) {
+                decoded = _tryToDecodeRichRevertReason(info.return);
+            }
+        }
+    }
+    if (decoded) {
```
fabioberger: We try to be a bit tighter on our existence checks. This will not pass if `decoded` is set to `false`, `undefined` or `null`. Since we've decided to never use `null` in our codebase, we prefer to check for `undefined` explicitly:

```
if (decoded !== undefined) {
```

This catches unexpected values sooner rather then later.

contracts/test-utils/src/assertions.ts
```diff
@@ -159,10 +187,76 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
...
+ * Try to guess if the data returned by a call is a rich revert reason.
+ * Returns an instance if it is, undefined if it isn't.
+ * @param data Any data returned by a contract call.
+ */
+function _tryToDecodeRichRevertReason(data: any): RichRevertReason | undefined {
+    // Here we try to guess if the return data is a rich revert.
+    if (typeof data === 'string' && data.startsWith('0x')) {
+        // Ensure its size is a multiple of 32 bytes + 4 byte selector.
+        const payloadLength = data.length - 2;
+        if (payloadLength >= 4 * 2 && (payloadLength - 4 * 2) % (32 * 2) === 0) {
```
fabioberger: We could get rid of the comment above if we wrote this as:

```
const BYTE_SELECTOR_LENGTH = 4 * 2;
const THIRTY_TWO_BYTES = 32 * 2;
if (payloadLength >= BYTE_SELECTOR_LENGTH && (payloadLength - BYTE_SELECTOR_LENGTH) % THIRTY_TWO_BYTES === 0) {
```

packages/abi-gen-templates/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/abi-gen-templates",
-    "version": "2.0.1",
+    "version": "2.0.2",
```
fabioberger: Not necessary. Remove.

packages/contract-wrappers/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/contract-wrappers",
-    "version": "8.0.3",
+    "version": "8.1.0",
```
fabioberger: Revert.

packages/contract-wrappers/test/utils/chai_setup.ts
```diff
@@ -9,5 +10,42 @@ export const chaiSetup = {
         chai.use(ChaiBigNumber());
         chai.use(dirtyChai);
         chai.use(chaiAsPromised);
+        chai.use(richRevertEquality);
     },
 };
+
+// A chai utlity to match rich reverts.
```
fabioberger:
```suggestion
// A chai utility to match rich reverts.
```

packages/ethereum-types/CHANGELOG.json
```diff
@@ -1,4 +1,15 @@
 [
+    {
+        "version": "2.2.0",
+        "changes": [
+            {
+                "note": "Create RichRevertAbi interface and add it to AbiDefinition types"
+            },
+            {
+                "note": "Add RichRevertAbi to AbiDefinition type union"
+            }
```
fabioberger: Add PR numbers.

packages/ethereum-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "ethereum-types",
-    "version": "2.1.0",
+    "version": "2.2.0",
```
fabioberger: Revert.

packages/order-utils/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "7.1.0",
+        "changes": [
+            {
+                "note": "Add RichRevertReason classes and tests."
+            }
```
fabioberger: PR #

packages/order-utils/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/order-utils",
-    "version": "7.0.2",
+    "version": "7.1.0",
```
fabioberger: Revert. You get the drill ;)

packages/order-utils/src/index.ts
```diff
@@ -72,3 +72,5 @@ export {
     FeeOrdersAndRemainingFeeAmount,
     OrdersAndRemainingFillAmount,
 } from './types';
+
+export * from './rich_reverts';
```
fabioberger: I'd rather we explicitly list the items we want exported. Someone might add an export to that file and be surprise to find that it's been auto-magically exported by the package.

packages/order-utils/src/rich_reverts.ts
fabioberger: Please use readable variable names over conciseness.

packages/order-utils/test/rich_reverts_test.ts
```diff
@@ -0,0 +1,111 @@
...
+    type ValueTypes = number | string | boolean | BigNumber | undefined;
+
+    describe('decoding', () => {
+        interface DecoderTestData {
+            cls: { name: string };
```
fabioberger: `cls`?

packages/types/CHANGELOG.json
fabioberger: And PR # :)

PR #1695, opened 2019-03-14
Reduce excess block polling when using Web3ProviderEngine

dekz:
## Description

Targeting external facing to reduce the spread of this hack. Internal tools/testing will be remain polling.

Web3ProviderEngine will create an `EthBlockTracker` [in the constructor](https://github.com/MetaMask/provider-engine/blob/master/index.js#L26-L30). When calling `.start()` this will begin to poll the blockchain for new blocks and keep a copy internally.

When left running in the background, this can consume a large amount of data in comparison to normal activities. 

The below screenshot shows data usage of one of our internal apps before and after disabling `EthBlockTracker`.

Screen Shot 2019-03-14 at 1 05 04 pm

Disabling the EthBlockTracker is simple yet is a hack as we reach into the provider and circumvent the starting of this process.

```javascript
(providerEngine as any)._ready.go();
```

Some additional Subproviders in the web3-provider-engine package depend on the existence of this provider. Simply disabling this may have unintended consequences for an app.

These providers make use of the `currentBlock` provided by `EthBlockTracker`:
* FilterSubprovider (eth_newFilter, eth_getFilterChanges, eth_getFilterLogs, eth_newBlockFilter)
* VMSubprovider (eth_call, eth_estimateGas)
* CacheSubprovider

If the app has a simple use case (RPCSubprovider and a SignerSubprovider) it can safely start in a way without the block polling. If the app has its own internal block tracking (ethereumjs-blockstream), and uses this for managing logs, it should be safe to start in this way.

### Targets
- [x] 0x Instant - Simply polls for the success of a transaction, doesn't depend on fill events
- [x] Portal/Website - Uses fill events for a history 
- [x] OrderWatcher - Uses ethereumjs-blockstream to track blocks and events

### Instant
When Metamask is not detected we construct a ProviderEngine. After around 30 seconds we've used 788KB from polling blocks.

Screen Shot 2019-03-14 at 2 11 39 pm

Another issue is we don't stop after the modal is closed. So if a user has opened Instant and closed it will continue polling in the background.

## Checklist:


fabioberger:
@dekz thoughts on making this change to all other

packages/instant/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "4.1.0",
+        "changes": [
+            {
+                "note": "Run Web3ProviderEngine without excess block polling",
+                "pr": 1695
+            }
+        ]
+    },
```
fabioberger: @dekz please merge this with the previous, unpublished CHANGELOG entry.

dekz: Instant hasn't had a published timestamp since verison 1.0.1. What's the way to determine the current unpublished CHANGELOG?

dekz:
@fabioberger I think I will make the change across the codebase. Now that the hack is contained to `providerUtils` I feel less dirty about spreading it everywhere.

charlesmarino:
Alt Text

contracts/utils/CHANGELOG.json
```diff
@@ -17,7 +17,11 @@
             {
                 "note": "Added Address.sol with test for whether or not an address is a contract",
                 "pr": 1657
-            }
+            },
+            {
+                "note": "Run Web3ProviderEngine without excess block polling",
+                "pr": 1695
+
```
fabioberger: AKH! Curly bracket missing

packages/utils/src/provider_utils.ts
```diff
@@ -8,6 +8,17 @@ import {
...
 export const providerUtils = {
+    /**
+     * Starts the Web3ProviderEngine without excess block polling
+     * @param providerEngine The Web3ProviderEngine
+     */
+    startProviderEngine(providerEngine: any): void {
+        if (providerEngine.start === undefined) {
+            throw new Error(`Invalid Web3ProviderEngine`);
+        }
+        providerEngine._ready.go();
```
fabioberger: Let's add a HACK comment here explaining why this is necessary.

PR #1699, opened 2019-03-14
Cheaper Mutex/Reentrancy Lock

merklejerk:
## Description

Replaces our current contract reentrancy guard with [a cheaper version](https://github.com/OpenZeppelin/openzeppelin-solidity/blob/master/contracts/utils/ReentrancyGuard.sol) that only does one SSTORE per call.

Our current implementation would have played well with [EIP-1087](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1087.md)/[EIP-1283](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1283.md), but those have been nixed for Constantinople.

#### Technical Details and Testing Impact 
This new implementation "lazily," detects reentrancy, only reverting when the penultimate reentrant call returns. 

This meant reentrancy tests that relied on immediate detection of the reentrancy had to be reworked to perform valid (non-failing) calls, because a failing call would revert the second `reentrancyGuard` state change, so no reentrancy would be detected. Reentrancy tests written for new `nonReentrant` functions need to take this into account.

## Types of changes

## Checklist:

### DIFF
--- a/contracts/exchange/contracts/test/ReentrantERC20Token.sol
+++ b/contracts/exchange/contracts/test/ReentrantERC20Token.sol
@@ -37,9 +37,6 @@ contract ReentrantERC20Token is
 
     uint8 internal constant BATCH_SIZE = 3;
 
-    // solhint-disable-next-line var-name-mixedcase
-    IExchange internal EXCHANGE;
-
     // All of these functions are potentially vulnerable to reentrancy
     // We do not test any "noThrow" functions because `fillOrderNoThrow` makes a delegatecall to `fillOrder`
     enum ExchangeFunction {
@@ -57,20 +54,20 @@ contract ReentrantERC20Token is
         NONE
     }
 
+    // solhint-disable-next-line
+    IExchange internal exchange;
     uint8 internal currentFunctionId = 0;
 
     constructor (address _exchange)
         public
     {
-        EXCHANGE = IExchange(_exchange);
+        exchange = IExchange(_exchange);
     }
 
     /// @dev Set the exchange function to reenter.
     /// @param _currentFunctionId A number that corresponds to an entry in the
     ///        ExchangeFunction enum
-    function setReentrantFunction(
-        uint8 _currentFunctionId
-    )
+    function setReentrantFunction(uint8 _currentFunctionId)
         external
     {
         currentFunctionId = _currentFunctionId;
@@ -78,13 +75,13 @@ contract ReentrantERC20Token is
 
     /// @dev A version of `transferFrom` that attempts to reenter the Exchange contract
     ///      once.
-    /// @param _from The address of the sender
-    /// @param _to The address of the recipient
-    /// @param _value The amount of token to be transferred
+    /// @param from The address of the sender
+    /// @param to The address of the recipient
+    /// @param value The amount of token to be transferred
     function transferFrom(
-        address _from,
-        address _to,
-        uint256 _value
+        address from,
+        address to,
+        uint256 value
     )
         external
         returns (bool)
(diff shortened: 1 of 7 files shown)


contracts/utils/contracts/src/ReentrancyGuard.sol
```diff
@@ -21,25 +21,23 @@ pragma solidity ^0.5.5;
...
+    // Starts at 1 and increases whenever a nonReentrant function is called.
+    uint256 private reentrancyGuardCounter = 1;
 
-    /// @dev Functions with this modifer cannot be reentered. The mutex will be locked
-    ///      before function execution and unlocked after.
+    /// @dev Functions with this modifer cannot be reentered.
     modifier nonReentrant() {
-        // Ensure mutex is unlocked
+        // Increment the counter.
+        reentrancyGuardCounter += 1;
```
abandeali1: We should be able to save 1 SLOAD here with:

```
uint256 localCounter = reentrancyGuardCounter + 1;
reentrancyGuardCounter = localCounter;
```

merklejerk: Led me down a rabbit hole with remix and found `uint256 localCounter = ++reentrancyGuardCounter;` is actually the cheapest variant on 0.5.6.

hysz: There may be a way to golf this down even further with some assembly :). But I feel like the savings would be pretty minimal.

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -83,106 +89,213 @@ contract ReentrantERC20Token is
...
+        pure
+        returns (uint256 total)
+    {
+        total = 0;
+        for (uint8 i = 0; i < orders.length; i++) {
+            total += orders[i].takerAssetAmount;
+        }
+    }
+
+    function _getRandomAddress() internal view returns (address)
```
abandeali1: Styling is a bit inconsistent for this function.

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -83,106 +89,213 @@ contract ReentrantERC20Token is
...
+    {
+        return address(
+            bytes20(
+                keccak256(
+                    abi.encodePacked(
+                        blockhash(block.number-1), now)
```
abandeali1: Add some spaces here:

```
blockhash(block.number - 1), now)
```

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -83,106 +89,213 @@ contract ReentrantERC20Token is
...
+    {
+        signatures = new bytes[](count);
+        for (uint i = 0; i < count; i++) {
+            signatures[i] = new bytes(66);
+            signatures[i][65] = bytes1(uint8(0x4));
+        }
+    }
+
+    /// @dev Create asset data that points to this ERC20 contract.
+    function _createAssetData() internal view returns (bytes memory assetData){
```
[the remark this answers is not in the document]
abandeali1: Styling is a bit inconsistent here too.

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -83,106 +89,213 @@ contract ReentrantERC20Token is
...
+        assetData = new bytes(36);
+        assembly {
+            mstore(assetData, 36)
+            mstore(add(assetData, 32), 0xf47261b000000000000000000000000000000000000000000000000000000000)
+            mstore(add(assetData, 36), address)
```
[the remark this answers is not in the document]
merklejerk: I think because `address` is an opcode itself, it never has dirty higher bits.

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -83,106 +89,213 @@ contract ReentrantERC20Token is
...
+
+    function _getTakerFillAmounts(
+        LibOrder.Order[] memory orders
+    )
+        internal
+        pure
+        returns (uint256[] memory amounts)
+    {
+        amounts = new uint256[](orders.length);
+        for (uint8 i = 0; i < orders.length; i++) {
```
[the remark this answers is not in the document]
abandeali1: Another minor gas golf: set `orders.length` to a variable to avoid extra mloads.

PR #1702, opened 2019-03-15
Changed EDPS endpoint to a 0x AWS one

askeluv:
## Description

We were previously using an endpoint for Ethereum DEX Prices Service which 0x did not control. This PR changes the endpoint to a service which is hosted in the 0x AWS environment.

In addition, we've increased timeout to 30 seconds + made the bottleneck 2x more aggressive (now max 30 requests per minute) so we don't hit rate-limits for the DEX APIs.

## Testing instructions

```
yarn build
yarn test
```

## Types of changes

* Fix (non-breaking change which fixes an issue)


feuGeneA:
Approval contingent on clean `static-tests` run of course.

PR #1703, opened 2019-03-15
Open lead modals via link

sjelfull:
## Description

If you pass `?modal=contact` to any of the pages containing a contact modal, it will open it on mount.

This also fixes a few SVG console errors.


chriskalani:
Maybe if it could just be `#contact` it would look better?

Also, I noticed 2 small things;

1. The page behind the modal shouldn’t scroll when the modal is open
2. Some modals have the “x” to close and some don’t. I think you added it for the vote stuff but it is missing from other modals.

PR #1704, opened 2019-03-15
Refactor contract function call to support web3.py v5.0.0. (backwards compatible with v4)

chowryan:
## Testing instructions

Verify that the package works with both web3.py v5 and v4

## Types of changes

Refactor to new method of calling contract functions.

## Checklist:


chowryan:
did a search for .call in python-package directory and only found 2 other places.

PR #1712, opened 2019-03-19
Fix LibAddressArray.indexOf() and add LibAddressArray tests.

merklejerk:
## Description

`/contracts/utils/.../LibAddressArray` was missing out on slowing down our test pipeline without unit tests of its own. This PR rights that wrong. 

It even exposed a bug where `indexOf()` was returning the wrong index. Fortunately, `indexOf()` is not used anywhere yet. This PR also fixes that bug. 

## Types of changes

## Checklist:


contracts/utils/contracts/test/TestLibAddressArray.sol
```diff
@@ -0,0 +1,114 @@
...
+        )
+    {
+        // Create a copy of the array.
+        result = new address[](addressArray.length);
+        assembly {
+            oldArrayMemStart := result
+            let length := mload(addressArray)
+            for { let i := 0 } lt(i, length) { i := add(i, 1) } {
+                mstore(add(result, mul(add(i, 1), 32)),
+                    mload(add(addressArray, mul(add(i, 1), 32))))
```
hysz: Indenting

contracts/utils/contracts/test/TestLibAddressArray.sol
```diff
@@ -0,0 +1,114 @@
...
+    )
+        public
+        pure
+        returns (
+            address[] memory result,
+            uint256 oldArrayMemStart,
+            uint256 newArrayMemStart
+        )
+    {
+        // Create a copy of the array.
```
hysz: We actually have a [memcpy](https://github.com/0xProject/0x-monorepo/blob/e8be5a0a8f641fe6473070bfa51646c74c18431f/contracts/utils/contracts/src/LibBytes.sol#L60) that could be useful here

contracts/utils/contracts/test/TestLibAddressArray.sol
```diff
@@ -0,0 +1,114 @@
...
+    ///      then performs the append.
+    ///      This tests the behavior of the address array being reallocated if
+    ///      the memory immediately after the old array is claimed.
+    /// @param addressArray Array of addresses.
+    /// @param freeMemOffset Number of (signed) bytes to offset the free memory pointer (0x40).
+    /// @param addressToAppend Address to append.
+    /// @return The new address array.
+    /// @return The memory address of the old address array.
+    /// @return The memory address of the new address array.
+    function testAppendRealloc(
```
hysz: Nice +1

contracts/utils/test/lib_address_array.ts
```diff
@@ -0,0 +1,156 @@
...
+            const expected = [...arr, addr];
+            const result = await lib.publicAppend.callAsync(arr, addr);
+            expect(result).to.deep.equal(expected);
+        });
+
+        it('should revert if the free memory pointer was moved to before the end of the array', async () => {
+            const arr = _.times(3, () => addressUtils.generatePseudoRandomAddress());
+            const addr = addressUtils.generatePseudoRandomAddress();
+            return expectContractCallFailedAsync(
+                lib.testAppendRealloc.callAsync(arr, new BigNumber(-1), addr),
```
hysz: Nit - `const freeMemOffset = new BigNumber(-1);`

contracts/utils/test/lib_address_array.ts
```diff
@@ -0,0 +1,156 @@
...
+            const arr = _.times(3, () => addressUtils.generatePseudoRandomAddress());
+            const addr = addressUtils.generatePseudoRandomAddress();
+            const expected = [...arr, addr];
+            const [result, oldArrayMemStart, newArrayMemStart] = await lib.testAppendRealloc.callAsync(
+                arr,
+                new BigNumber(1),
+                addr,
+            );
+            expect(result).to.deep.equal(expected);
+            expect(newArrayMemStart).bignumber.to.not.be.equal(oldArrayMemStart);
```
hysz: To ensure `append` is not overwriting other areas of memory, it could be good to explicitly test for the expected value here. The new array should be placed at the end of memory.

contracts/utils/contracts/test/TestLibAddressArray.sol
hysz: Actually, do we need to make a copy?

merklejerk: Wasn't sure at the time so I just did it to be safe. I checked it out in remix just now and it seems that no, we don't.

PR #1714, opened 2019-03-19
Simulate Maker transfer in order validation

dekz:
## Description

Simulate a one sided maker transfer in order validation.

This will trigger the case where the maker has the sufficient balance and allowance, but the asset is non-tradeable. For example in a Paused contract or when the maker has not performed some action to unlock their account.

Pulled from a number of SRA endpoints and walked through the orderbooks with this change. This has flagged a few orders which are currently sitting on orderbooks which are invalid. I.e there are a few tokens where the Maker is "frozen" on live orderbooks.

By default the `order.takerAddress` is used to simulate the receiver from maker. With Open orderbooks this results in the `NULL_ADDRESS` This also has a few false positives where some ERC20 tokens cannot be transferred to the `NULL_ADDRESS`, e.g Bloom token. 

In this case the developer is in the best position to determine an alternative address to use. The receiver address during the simulation can be set by specifying `simulationTakerAddress` in `ValidateOrderFillableOpts`. Whilst 0x could assign a non `NULL_ADDRESS` default, this is unlikely to be fool proof across ALL token standards and networks.

## Testing instructions

Pulled from a number of SRA endpoints and walked through the orderbooks with this change. 

```typescript
import { HttpClient, SignedOrder } from '@0x/connect';
import { ContractWrappers } from '@0x/contract-wrappers';
import { RPCSubprovider, Web3ProviderEngine } from '@0x/subproviders';
import { providerUtils } from '@0x/utils';

const uri = 'https://api.radarrelay.com/0x/v2/';

const client = new HttpClient(uri);
(async () => {
    const response = await client.getAssetPairsAsync();
    const orders: SignedOrder[] = [];
    for (const record of response.records) {
        const { assetDataA, assetDataB } = record;
        const orderBook = await client.getOrderbookAsync({
            baseAssetData: assetDataA.assetData,
            quoteAssetData: assetDataB.assetData,
        });
        for (const order of orderBook.asks.records) {
            orders.push(order.order);
        }
        for (const order of orderBook.bids.records) {
            orders.push(order.order);
        }
    }
    const rpcSubprovider = new RPCSubprovider('https://mainnet.infura.io');
    const provider = new Web3ProviderEngine();
    provider.addProvider(rpcSubprovider);
    providerUtils.startProviderEngine(provider);
    const contractWrappers = new ContractWrappers(provider, { networkId: 1 });
    for (const order of orders) {
        try {
            await contractWrappers.exchange.validateOrderFillableOrThrowAsync(order, {
                validateRemainingOrderAmountIsFillable: false,
            });
        } catch (e) {
            console.log(e.message);
            console.log(order);
        }
    }
    console.log(orders.length);
})();

```

## Checklist:


contracts/erc20/contracts/test/UntransferrableDummyERC20Token.sol
```diff
@@ -0,0 +1,63 @@
...
+pragma solidity ^0.5.5;
+
+import "./DummyERC20Token.sol";
+
+
+// solhint-disable no-empty-blocks
+contract UntransferrableDummyERC20Token is
+    DummyERC20Token
+{
+    bool internal _paused;
```
fabioberger: This is unused. Remove.

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1139,7 +1141,8 @@ export class ExchangeWrapper extends ContractWrapper {
      * Validate if the supplied order is fillable, and throw if it isn't
      * @param signedOrder SignedOrder of interest
      * @param opts ValidateOrderFillableOpts options (e.g expectedFillTakerTokenAmount.
-     * If it isn't supplied, we check if the order is fillable for a non-zero amount)
+     * If it isn't supplied, we check if the order is fillable for the remaining amount.
+     * To check if the order is fillable for any amount set validateRemainingOrderAmountIsFillable to false.)
```
fabioberger:
```suggestion
     * To check if the order is fillable for a non-zero amount, set `validateRemainingOrderAmountIsFillable` to false.)
```

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1156,15 +1159,92 @@ export class ExchangeWrapper extends ContractWrapper {
...
+        );
+        if (opts.expectedFillTakerTokenAmount) {
+            // If the caller has specified a taker fill amount, we use this for all validation
+            fillableTakerAssetAmount = opts.expectedFillTakerTokenAmount;
+        } else if (shouldValidateRemainingOrderAmountIsFillable) {
+            // Historically if a fill amount was not specified we would default to the amount
+            // left on the order.
+            fillableTakerAssetAmount = signedOrder.takerAssetAmount.minus(filledTakerTokenAmount);
+        } else {
+            const makerAssetBalance = await balanceAllowanceStore.getBalanceAsync(
```
fabioberger: Let's move this logic into an `order_utils` method so that others can compute this for other purposes (e.g Coordinator server needs it too).

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1156,15 +1159,92 @@ export class ExchangeWrapper extends ContractWrapper {
...
             exchangeTradeSimulator,
             signedOrder,
             this.getZRXAssetData(),
-            expectedFillTakerTokenAmountIfExists,
+            fillableTakerAssetAmount,
+        );
+        await this.validateMakerTransferThrowIfInvalidAsync(signedOrder, fillableTakerAssetAmount);
+    }
+    /**
+     * Validate the transfer from the Maker to the Taker. This is simulated on chain
```
fabioberger:
```suggestion
     * Validate the transfer from the Maker to the Taker. This is simulated on-chain
```

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1156,15 +1159,92 @@ export class ExchangeWrapper extends ContractWrapper {
...
+     * via an eth_call. If this call fails the asset is unlikely to be transferrable.
```
fabioberger:
```suggestion
     * via an eth_call. If this call fails, the asset is currently nontransferable.
```

packages/contract-wrappers/src/types.ts
```diff
@@ -122,11 +122,14 @@ export interface ContractWrappersConfig {
 /**
  * expectedFillTakerTokenAmount: If specified, the validation method will ensure that the
  * supplied order maker has a sufficient allowance/balance to fill this amount of the order's
- * takerTokenAmount. If not specified, the validation method ensures that the maker has a sufficient
- * allowance/balance to fill the entire remaining order amount.
+ * takerTokenAmount.
+ * validateRemainingOrderAmountIsFillable: The validation method ensures that the maker has a sufficient
+ * allowance/balance to fill the entire remaining order amount. This is the default. If neither options are
+ * specified, the balances and allowances are checked to determine the order is fillable by any amount.
```
fabioberger:
```suggestion
 * specified, the balances and allowances are checked to determine the order is fillable for a non-zero amount. We call such orders "partially fillable orders".
```

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
dekz: refactored to use `orderStateUtils.getOpenOrderRelevantStateAsync`

packages/contract-wrappers/src/contract_wrappers/exchange_wrapper.ts
```diff
@@ -1155,16 +1159,74 @@ export class ExchangeWrapper extends ContractWrapper {
...
+
+        let fillableTakerAssetAmount;
+        const shouldValidateRemainingOrderAmountIsFillable = _.isUndefined(opts.validateRemainingOrderAmountIsFillable)
+            ? true
+            : opts.validateRemainingOrderAmountIsFillable;
+        if (opts.expectedFillTakerTokenAmount) {
+            // If the caller has specified a taker fill amount, we use this for all validation
+            fillableTakerAssetAmount = opts.expectedFillTakerTokenAmount;
+        } else if (shouldValidateRemainingOrderAmountIsFillable) {
+            // Historically if a fill amount was not specified we would default to the amount
```
fabioberger: Not just historically right? It still has this behavior correct?

Issue #1694, opened 2019-03-14
0x Instant Continues To Hit SRA Endpoint After Being Closed


pointtoken:
If you open up a web console and watch network traffic on any implementation of 0x Instant, the browser starts hammering the SRA server endpoint, even after the 0x instant UI is closed, pinging the orderbook every 5 seconds. What is going on here?

You can witness this on the MyCrypto implementation: https://www.mycrypto.com/swap/0x

fragosti:
Hi @pointtoken thanks for raising the issue. Instant pinging for the orderbook every 5 seconds is definitely expected. It does so to keep the quote price up to date. However, it should not be doing that on close and we will address this in the next release.

pointtoken:
Thanks @fragosti  -- how does one know when instant releases happen to the CDN?

fragosti:
We try to announce releases like that in our Discord. We will have a fix by end of next week if not earlier.

BMillman19:
@pointtoken we just deployed a new version!

PR #1722, opened 2019-03-21
Unmount Instant when closing

steveklebanoff:
## Description

See https://github.com/0xProject/0x-monorepo/issues/1694

Unmounts component when Instant is closed. 

Also, provides public `unmount` function for users who render Instant "inline".  For example, MyCrypto  can call `zeroExInstant.unmount` from a `componentWillUnmount` [here](https://github.com/MyCryptoHQ/MyCrypto/blob/eb59b9942491b1a0826395a4f988a1e08ae11256/common/v2/features/BuyAndExchange/ZeroEx/ZeroEx.tsx#L17).

### Repro
- Checkout master
- Select token and type amount
- Close instant
- Notice instant is still polling SRA endpoint

### Test fix
- Checkout this branch
- Notice that bug is fixed: when Instant is closed, polling stops.

## Checklist:


BMillman19:
LGTM, some but of weirdness with the `yarn.lock` changes, where did those come from?

steveklebanoff:
@fragosti thanks for the idempotent addition and pushing this through 🙇

Issue #1727, opened 2019-03-24
Clients generated from SRA spec can't validate numeric inputs.


feuGeneA:
## Expected Behavior

Generating an SRA client from the SRA spec should produce a client that can properly validate its inputs.

## Current Behavior

Generating an SRA client from the SRA spec produces a client that has too many backslashes in its regular expressions to properly validate its inputs.  Specifically, `wholeNumberSchema` (`^\\d+$`) and `numberSchema` (`^\\d+(\\.\\d+)?$`) don't work because the regular expressions have double-backslashes where they should only be single-.

## Possible Solution

Perhaps change @0x/sra-spec to convert double-backslashes to single ones, in JSON schema regular expressions, before injecting them into the API spec.

Alternatively, perhaps remove the extra backslashes from the JSON schemas themselves, though I'm not sure that makes sense, since apparently the implementations of JSON schema validation (npmjs' `jsonschema` and PyPI's `jsonschema) handle this just fine.

## Steps to Reproduce (for bugs)

1.  Generate a client from our SRA spec.  (To see how the Python client was generated, see https://github.com/0xProject/0x-monorepo/blob/development/python-packages/sra_client/generate.sh)
2.  Try to use the client in a way that will validate a whole number.

## Context
This issue arose when @michaelhly was trying to write example usage code for our Python SRA client.

He said:

> there is a bug in the SRA client schema checker
> it is complaining about the fee being '0'
> the setter's regular expression has an extra backslash where it should not
> ```
>     @taker_fee.setter
>     def taker_fee(self, taker_fee):
>         """Sets the taker_fee of this OrderSchema.
> 
> 
>         :param taker_fee: The taker_fee of this OrderSchema.  # noqa: E501
>         :type: str
>         """
>         if taker_fee is None:
>             raise ValueError(
>                 "Invalid value for `taker_fee`, must not be `None`"
>             )  # noqa: E501
>         if taker_fee is not None and not re.search(
>             r"^\\d+$", taker_fee
>         ):  # noqa: E501
>             raise ValueError(
>                 r"Invalid value for `taker_fee`, must be a follow pattern or equal to `/^\\d+$/`"
>             )  # noqa: E501
> 
>         self._taker_fee = taker_fee
> ```
> 
> r"^\\d+$" should be r"^\d+$"
> and this is true for all the validation checkers in the sra_client

But the Python SRA client is all generated code.  That regular expression wasn't written by hand in there; the code generator just copied it from the SRA spec.

[...]

PR #1731, opened 2019-03-25
Fix regex validation on numeric inputs

michaelhly:
## Description

Fix issue described in https://github.com/0xProject/0x-monorepo/issues/1727. 

## Testing instructions

Run `python test`

## Types of changes

 Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/python-packages/sra_client/sra_client/models/order_schema.py
+++ b/python-packages/sra_client/sra_client/models/order_schema.py
@@ -188,7 +188,7 @@ class OrderSchema(object):
             r"^\d+$", maker_fee
         ):  # noqa: E501
             raise ValueError(
-                r"Invalid value for `maker_fee`, must be a follow pattern or equal to `/^\\d+$/`"
+                r"Invalid value for `maker_fee`, must be a follow pattern or equal to `/^\d+$/`"
             )  # noqa: E501
 
         self._maker_fee = maker_fee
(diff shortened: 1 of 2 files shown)


python-packages/sra_client/sra_client/models/order_schema.py
```diff
@@ -401,7 +401,7 @@ def salt(self, salt):
             raise ValueError(
                 "Invalid value for `salt`, must not be `None`"
             )  # noqa: E501
-        if salt is not None and not re.search(r"^\\d+$", salt):  # noqa: E501
+        if salt is not None and not re.search(r"^\d+$", salt):  # noqa: E501
             raise ValueError(
                 r"Invalid value for `salt`, must be a follow pattern or equal to `/^\\d+$/`"
```
feuGeneA: Should update these error messages too.  I see there are quite a few of them, so change them all, not just this one.

feuGeneA:
To clarify for anyone following along, these changes here only fix the problem described in #1727 for *this client*. Any other clients that get generated from the SRA spec will also face the problem described in #1727, even after this PR is merged.

python-packages/sra_client/sra_client/models/order_schema.py
michaelhly: https://github.com/0xProject/0x-monorepo/pull/1731/commits/e066548581514e7a0d7f4667fd081b8239d741aa

PR #1736, opened 2019-03-26
[3.0] Only test contract-related packages in circleci.

dorothy-zbornak:
## Description
The `3.0` branch is contract-focused right now and will introduce breaking changes to a lot of ancillary packages (e.g., `@0x/instant`). The current circleci suite of tests all depend on `build:ci:no_website`, which is bound to fail in `3.0`, which means circleci would give us no meaningful signals on this branch. This updates the circleci config to only build and test contract-related packages.

Note: `test-contracts-geth` still fails, just like in development, but it's still included because it should start working once we update our geth fork.

This also fixes a minor merge artifact from the the `development` rebase.
 
## Types of changes

## Checklist:

### DIFF
--- a/contracts/utils/CHANGELOG.json
+++ b/contracts/utils/CHANGELOG.json
@@ -34,17 +34,12 @@
                 "pr": 1657
             },
             {
-<<<<<<< HEAD
                 "note": "Add unit tests for `LibAddressArray`",
                 "pr": 1712
             },
             {
                 "note": "Fix `LibAddressArray.indexOf` returning incorrect index.",
                 "pr": 1712
-=======
-                "note": "Change ReentrancyGuard implementation to cheaper one",
-                "pr": 1699
->>>>>>> Update PRs in changelogs.
             }
         ],
         "timestamp": 1553091633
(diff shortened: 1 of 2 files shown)


contracts/utils/CHANGELOG.json
```diff
@@ -34,17 +34,12 @@
...
                 "note": "Add unit tests for `LibAddressArray`",
                 "pr": 1712
             },
             {
                 "note": "Fix `LibAddressArray.indexOf` returning incorrect index.",
                 "pr": 1712
-=======
-                "note": "Change ReentrancyGuard implementation to cheaper one",
-                "pr": 1699
->>>>>>> Update PRs in changelogs.
```
abandeali1: Looks like there are still some merge conflicts to resolve here, otherwise looks good to me.

dorothy-zbornak: I think these are artifacts actually removed? Unless you're referring to something else in the file.

abandeali1: Ah whoops, misread this.

PR #1737, opened 2019-03-26
[instant] Add infura project id via env variables when bundling instant

BMillman19:
## Description

Adds two new env variables to `.env_example`. `INSTANT_INFURA_PROJECT_ID_PRODUCTION` and `INSTANT_INFURA_PROJECT_ID_DEVELOPMENT`

## Testing instructions

Try bundling instant with a local `.env` file

## Types of changes

## Checklist:


packages/instant/webpack.config.js
```diff
@@ -141,24 +147,32 @@ const generateConfig = (dischargeTarget, heapConfigOptions, rollbarConfigOptions
...
-                        options: {
-                            "plugins": ["transform-runtime"],
-                            'presets': [
-                                ['env', {
-                                    'targets': {
-                                        "chrome": 41
-                                    },
-                                }],
-                            ],
+                    use: [
```
BMillman19: my prettier did this, I guess CI is not looking at `.js` files

packages/instant/webpack.config.js
```diff
@@ -89,11 +89,17 @@ const generateConfig = (dischargeTarget, heapConfigOptions, rollbarConfigOptions
...
+            ? process.env.INSTANT_INFURA_PROJECT_ID_PRODUCTION
+            : process.env.INSTANT_INFURA_PROJECT_ID_DEVELOPMENT;
+
     const envVars = {
         GIT_SHA: JSON.stringify(GIT_SHA),
         NPM_PACKAGE_VERSION: JSON.stringify(process.env.npm_package_version),
         ROLLBAR_ENABLED: rollbarEnabled,
-        HEAP_ENABLED: heapEnabled
+        HEAP_ENABLED: heapEnabled,
+        INSTANT_INFURA_PROJECT_ID: JSON.stringify(infuraProjectId),
```
PirosB3: is `infuraProjectId` a string? if so, why do we want to serialize it to JSON?

packages/instant/src/constants.ts
```diff
@@ -53,8 +53,8 @@ export const META_MASK_CHROME_STORE_URL =
 export const META_MASK_OPERA_STORE_URL = 'https://addons.opera.com/en/extensions/details/metamask/';
```
fragosti: Unrelated to this change -- `HOST_DOMAINS_EXTERNAL` seems to be missing 0x.org which is not good :P

BMillman19: Good catch, will update in this PR

packages/instant/webpack.config.js
BMillman19: `infuraProjectId` is a number. This stringify call is effectively converting from a number to a string. This seems to be the convention we're following in this config file. I'm happy to change it up but we should prob be consistent either way. @fragosti thoughts?

fragosti: This is a fun little `DefinePlugin` gotcha
https://stackoverflow.com/questions/50454341/why-json-stringifyproduction

PR #1742, opened 2019-03-27
[3.0] Add chain ID to EIP 712 domains

dorothy-zbornak:
## Description

This adds a `chainId` to the domain separator for the Coordinator and Exchange (as per the [EIP-712 spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-domainseparator)). For reference, the chain ID is the unique numerical ID of the network the contracts are deployed on.

There are some notable consequences that come with this change, which is why this PR spans so many packages:
- Both the Coordinator and Exchange need to be deployed with an additional `chainId` constructor parameter.
- The domain separator schema for both the Coordinator and the Exchange now looks like `        EIP712Domain(string name,string version,uint256 chainId,address verifyingContractAddress)`
- In `@0x/types`, `Order`, `ZeroExTransaction`, and signed variants have had their structures changed:
   - `exchangeAddress` has been removed.
   -  A required `domain` field has been added. This field is of the existing type `EIP712DomainWithDefaultSchema`, which has been augmented with a required `chainId` field.
- `@0x/utils.providerUtils` now has a `getChainIdAsync()` function for retrieving the current chain ID given a provider. For testing, you can either use this function or just `@0x/order-utils.constants.TESTPRC_NETWORK_ID`.
- The constructor for `@0x/test-utils.TransactionFactory` now requires a `chainId` constructor parameter.
- In `@0x/order-utils.OrderFactory`, `createOrder()` and `createOrderFromPartial()` now require `chainId` fields/parameters.

I've also taken the liberty of breaking up `@0x/contracts-exchange-libs/contracts/src/LibEIP712.sol` into reusable components (`LibEIP712`, `LibEIP712ExchangeDomain`, and `LibEIP712ExchangeDomainConstants`) so there is less code duplication between the Coordinator and Exchange.

## Types of changes

## Checklist:


dorothy-zbornak:
> Looks good! Might not be necessary but it could be worthwhile to add explicit tests for known Chain ID's / expected behavior when an unrecognized/malformed Chain ID is used client-side.

@hysz Wondering if the new tests I just added to [contracts/exchange-libs/test/src/exchange_libs.ts](../blob/feature/contracts/3.0/add-chainId-to-eip-712-domain/contracts/exchange-libs/test/exchange_libs.ts#L144) cover this requirement. Is it sufficient to prove that `LibOrder` and `LibEIP712` produce different order hashes and domain separators given a different `chainId`? It implies that any logic depending on a hash produced by `LibEIP712.sol` (such as signature validation) would fail if the `chainId` does not match.

contracts/coordinator/contracts/test/TestLibs.sol
```diff
@@ -0,0 +1,64 @@
...
+pragma experimental "ABIEncoderV2";
+
+import "@0x/contracts-exchange-libs/contracts/src/LibEIP712.sol";
+import "../src/libs/LibConstants.sol";
+import "../src/libs/LibCoordinatorApproval.sol";
+import "../src/libs/LibZeroExTransaction.sol";
+
+
+// solhint-disable no-empty-blocks
+contract TestLibs is
```
abandeali1: Why was this file added back? `getCoordinatorApprovalHash` and `getTransactionHash` were already made public.

contracts/coordinator/contracts/test/TestMixins.sol
```diff
@@ -0,0 +1,40 @@
...
+pragma solidity ^0.5.5;
+pragma experimental "ABIEncoderV2";
+
+import "@0x/contracts-exchange-libs/contracts/src/LibEIP712.sol";
+import "../src/libs/LibConstants.sol";
+import "../src/MixinSignatureValidator.sol";
+import "../src/MixinCoordinatorApprovalVerifier.sol";
+
+
+// solhint-disable no-empty-blocks
```
abandeali1: This file was previously removed as well, I don't believe it is currently used anywhere.

contracts/coordinator/contracts/src/libs/LibEIP712Domain.sol
```diff
@@ -18,61 +18,43 @@
...
-            uint256(address(this))
-        ));
-
-        EIP712_EXCHANGE_DOMAIN_HASH = keccak256(abi.encodePacked(
-            EIP712_DOMAIN_SEPARATOR_SCHEMA_HASH,
-            keccak256(bytes(EIP712_EXCHANGE_DOMAIN_NAME)),
-            keccak256(bytes(EIP712_EXCHANGE_DOMAIN_VERSION)),
-            uint256(address(EXCHANGE))
-        ));
+        EIP712_COORDINATOR_DOMAIN_HASH = hashEIP712Domain(
```
abandeali1: Shouldn't these 2 calculations include the `chainId` in the hash?

abandeali1: Ah I see. Feels a bit inconsistent that `verifyingContractAddress` is passed into `hashEIP712Domain` and `chainId` is read from state. I'd probably just add `chainId` as a param too.

contracts/exchange-libs/contracts/src/LibEIP712.sol
```diff
@@ -21,54 +21,65 @@ pragma solidity ^0.5.5;
...
-
-    // EIP712 Domain Version value
-    string constant internal EIP712_DOMAIN_VERSION = "2";
-
     // Hash of the EIP712 Domain Separator Schema
     bytes32 constant internal EIP712_DOMAIN_SEPARATOR_SCHEMA_HASH = keccak256(abi.encodePacked(
         "EIP712Domain(",
         "string name,",
         "string version,",
+        "uint256 chainId,",
```
abandeali1: Nit: should this be `chainId` or `networkId`? Feels like we should just stick to one, and `networkId` is already used in the standard RPC methods.

packages/types/src/index.ts
```diff
@@ -18,6 +18,7 @@ export interface Order {
     takerAssetData: string;
     salt: BigNumber;
     exchangeAddress: string;
+    chainId: number;
```
abandeali1: Rather than just adding the `chainId`, I think we should take this opportunity to add an extra `eip712Domain` field that looks like:

```
{
    verifyingContract: ...,
    chainId: ...,
    name: ...,
    version: ...
}
```

Alternatively, we can just make it a `eip712DomainHash` field and do all of the calculations beforehand. This would consume less bandwidth but make it a bit harder to develop with.

contracts/exchange-libs/contracts/src/LibEIP712.sol
```diff
@@ -21,54 +21,65 @@ pragma solidity ^0.5.5;
...
         "address verifyingContract",
```
abandeali1: Let's also take this opportunity to rename `verifyingContract` to `verifyingContractAddress` for consistency.

contracts/coordinator/contracts/test/TestLibs.sol
dorothy-zbornak: Hmm, must have been a rebase artifact.

contracts/exchange-libs/contracts/src/LibEIP712.sol
dorothy-zbornak: You sure? Seems like that [breaks spec](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-domainseparator).

abandeali1: The spec is purely an example, we can technically add any fields that we want. I just don't like that we're calling this field 2 different things throughout the codebase.

contracts/exchange-libs/contracts/src/LibEIP712.sol
abandeali1: Discussed offline, these are actually 2 different things.

packages/types/src/index.ts
dorothy-zbornak: I went with `domain` rather than `eip712Domain` for the new field in the `Order` and `ZeroExTransaction` types for brevity. Let me know if this is fine or you'd rather be more explicit.

contracts/coordinator/test/coordinator.ts
```diff
@@ -92,24 +95,28 @@ describe('Coordinator tests', () => {
...
+                chainId,
+            },
         };
         const makerPrivateKey = devConstants.TESTRPC_PRIVATE_KEYS[accounts.indexOf(makerAddress)];
         const takerPrivateKey = devConstants.TESTRPC_PRIVATE_KEYS[accounts.indexOf(takerAddress)];
         const feeRecipientPrivateKey = devConstants.TESTRPC_PRIVATE_KEYS[accounts.indexOf(feeRecipientAddress)];
         orderFactory = new OrderFactory(makerPrivateKey, defaultOrderParams);
-        makerTransactionFactory = new TransactionFactory(makerPrivateKey, exchange.address);
-        takerTransactionFactory = new TransactionFactory(takerPrivateKey, exchange.address);
+        makerTransactionFactory = new TransactionFactory(makerPrivateKey, exchange.address, chainId);
```
abandeali1: Not a blocker, but this should probably be changes later to just take a `domain` parameter.

PR #1749, opened 2019-04-01
don't publish tsconfig.json

NoahZinsmeister:
As far as I'm aware, it's bad practice to push the tsconfig.json file to npm.

The reason I noticed was because of a lint warning - "File '../../tsconfig' does not exist." - in VSCode.


dekz:
Hey @NoahZinsmeister. Thanks for opening this issue. This will actually be fixed in #1747

PR #1755, opened 2019-04-03
Fix yarn permission denied errors in CircleCI

albrow:
## Description

I may have found a different way to fix the "Permission denied" errors in CircleCI that still allows us to install a specific version of yarn. Planning to re-run the tests a few times to make sure it works.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,8 +12,13 @@ jobs:
             - checkout
             - run: echo 'export PATH=$HOME/CIRCLE_PROJECT_REPONAME/node_modules/.bin:$PATH' >> $BASH_ENV
             - run:
+                  # HACK(albrow): Without this, yarn commands will sometimes
+                  # fail with a "permission denied" error.
+                  name: Set npm path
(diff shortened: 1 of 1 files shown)


albrow:
With this change in place, CircleCI successfully proceeded past the `yarn install` step of the `build` task 9 times in a row. You can see the full history here: https://circleci.com/gh/0xProject/workflows/0x-monorepo/tree/fix%2Fcircleci-yarn-permission-denied. (For the first 8 runs I canceled the CI run after `yarn install` was done).

I'm pretty confident that this fixes the issue.

LogvinovLeon:
I'm curious about where did you find that solution and why does it work? Supper happy it's fixed though.

albrow:
@LogvinovLeon I don't remember exactly where I found that solution, but I [used it in 0x-mesh](https://github.com/0xProject/0x-mesh/commit/725bfa258d8938fc363e1860b289de2c42648729) when we were running into a similar problem there. We actually don't need this hack in 0x-mesh anymore because we changed the way we install Node.js (using nvm instead of the version provided by the image). I realized it might be useful for the monorepo, so I gave it a shot.

PR #1757, opened 2019-04-04
Python doc polish

feuGeneA:
## Description

After the recent bounty to fully document all of our Python offerings with code examples, I polished things up, mostly copy editing (incl. code examples) for conciseness and improved flow.

The rendered docs are accessible at the links below.  I suggest not reviewing the "Improve documentation" commit (which, btw, encompasses the bulk of the changes in this PR), and instead just review the rendered docs.

* http://0x-contract-addresses-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-sra-client-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-contract-artifacts-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-json-schemas-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-contract-wrappers-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-order-utils-staging-py.s3-website-us-east-1.amazonaws.com/
* http://0x-middlewares-staging-py.s3-website-us-east-1.amazonaws.com/

Do note that virtually all of this documentation is rendered from "docstings" in the source code.  (In Python, a docstring is simply a string literal that's ignored at runtime but that documents a module, method, etc, depending on its context.)  The rendering is done by [Sphinx](http://www.sphinx-doc.org/en/master/), which to my understanding is powerfully customizable, so in theory we could make the rendering look however we wanted.  However, I don't know much about this tool, as I have barely scratched the surface of its usage.

Also note that all of the code examples are "doctests."  (In Python, doctests are code snippets that can be exercised in an automated way.)  When we run our unit tests (in CI and locally), we also test our doctests.  So, when CI is clean, you know those examples are working.

## Types of changes

## Future work

Outside the scope of this PR...

### ~~Ensuring ease of conversion between Python object and JSON~~

Edit: I knocked this out on Future Focused Friday.

~~In scrubbing this code with such scrutiny, it has come to my attention that there's one remaining thing that's still not as easy as it should be: moving order objects between Web3.py-compatibility and Relayer-compatibility.~~

~~To submit an order to our contracts via web3.py the order object must have Python-native data types (asset data as `bytes`, numbers as `int`s, etc), but when working with a Relayer the order must be in JSON (asset data as hex strings, numbers as strings, etc).  We have provided convenience methods for converting between the two types, but we should write some more tests/examples/documentation around that.~~

~~Specifically, I think we should modify the SRA Client examples to declare the order as a web3.py-compatible object (currently it starts out as JSON), then convert it to JSON and submit it to the relayer (launch kit), then have a separate taker account get the order and fill it (which will require converting it back from JSON to a web3.py-compatible object).  Doing this may well reveal worthwhile changes to our Python API.~~

### Integrating docs with 0x.org

The URLs for these docs are ugly.  You see those staging URLs above, but they're not really any different from the "real" doc locations.  (Just take out the `-staging` part, and you have it.)  We should consider altering the rendering to fit our design motifs (see discussion of Sphinx above), and we should host them at friendlier URLs.  If desirable, we may be able to get Sphinx to render everything into one giant doc, rather than one per package.

## Checklist
* [x] Ensure `pip install` instructions are consistent across the docs (some were missing it).
* [x] Correct SRA client documentation describing Network ID and pagination parameters as floats.
* [x] Clean up unmatched parenthesis in doc of return value of `*_with_http_info()` methods in SRA client.


PirosB3:
Eugene! this documentation is really amazing! thanks for putting this together. I took a look at all the rendered documentation and I want to share my comments. Most of these comments are broad across all the documentation:

1. Many times you mention names of products or networks within the Ethereum ecosystem (for example: Infura, Ganache). While 99% of Ethereum developers will know what you are speaking about, the 1% would probably appreciate linking these terms to the appropriate documentation.

2. All libraries have a namespace of `zero_ex`. What's interesting is that, if I install one of these libraries (`python -m pip install 0x-sra-client`) I don't get the`zero_ex` namespace
```
>>> from zero_ex.sra_client import ApiClient, Configuration, DefaultApi
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'zero_ex'
>>> from sra_client import ApiClient, Configuration, DefaultApi
```
Are we sure the `zero_ex` namespace is always installed?

3. When you mention cancels having to deal with the blockchain directly (without going through the SRA) maybe it would be nice to mention the TEC and how this can benefit cancels in the future? just a thought.

4. In a good part of the documentation, arguments to functions that are considered integers (ex. `network_id`) are mentioned as floats in the function signatures. An example is `get_order_config()`. Why is this?

5. Aside from the SRA and `order-wrapper`, I don't see many of these packages having a traditional "pip install" page. Putting myself in the shoes of a end-user: how do I actually install these libraries?

feuGeneA:
@PirosB3 thanks for the great feedback!

> 2. All libraries have a namespace of `zero_ex`. What's interesting is that, if I install one of these libraries (`python -m pip install 0x-sra-client`) I don't get the`zero_ex` namespace
> 
> ```
> >>> from zero_ex.sra_client import ApiClient, Configuration, DefaultApi
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> ModuleNotFoundError: No module named 'zero_ex'
> >>> from sra_client import ApiClient, Configuration, DefaultApi
> ```
> Are we sure the `zero_ex` namespace is always installed?

`0x-sra-client` is a special case here, as it formerly was not in the `zero_ex` namespace, and in this PR I've corrected that.  Further, the packages that were recently introduced by the bounty work (`0x-contract-wrappers`, and `0x-middlwares`) have not yet been published, so at this point you can only install them from the repo (eg `python-packages/contract_wrappers$ pip install -e .`).  If you experience this problem with any of the already-published packages (that aren't `sra_client`) please let me know.

> 2. In a good part of the documentation, arguments to functions that are considered integers (ex. `network_id`) are mentioned as floats in the function signatures. An example is `get_order_config()`. Why is this?

Yeah, I noticed that myself and wasn't sure whether it was worth addressing.  This bulk of the SRA client was generated from [the SRA spec](http://sra-spec.s3-website-us-east-1.amazonaws.com/api.json) (see `generate.sh` in this component), and that's how it came out.  Looking closer, tt's because there are lots of places in the spec where we use `"type":"number"` where we really should be using `"type":"integer"`.  So, technically it's a bug in the spec, which I've just  raised (#1762), but I'll correct it here.

> 3. Aside from the SRA and `order-wrapper`, I don't see many of these packages having a traditional "pip install" page. Putting myself in the shoes of a end-user: how do I actually install these libraries?

Good call, I'll add `pip install` instructions where they're missing.  I hadn't put them in my own docs, but the bounty guys did, so that's why they're inconsistent.

feuGeneA:
I spent my Future Focused Friday knocking off that "Future Work" item in my PR description about ease of JSON conversion.  The SRA demo now goes from native Python, converted to JSON for submission to relayer, and then back from relayer JSON to native Python, demonstrating both a fill and a cancel.

PirosB3:
Hi @feuGeneA thanks for the great update! Have you re-rendered documentation since you made those changes we discussed? can I go ahead with another review?

PR #1760, opened 2019-04-05
[WIP][FFF] Cut the package fat! Part 2: Ethers and Abi-gen-wrappers

dekz:
### Optimised ABI gen templates
* Moved shared behaviour to base functions in the base-contract.
* Moved the function signature to a property. Now accessible by other utilities

```
# Before
115K  exchange.ts
 60K  asset_proxy_owner.ts
 39K  dummy_erc721_token.ts
 32K  dummy_erc20_token.ts
 28K  multi_asset_proxy.ts
 27K  forwarder.ts

# After
 87K  exchange.ts
 45K  asset_proxy_owner.ts
 29K  dummy_erc721_token.ts
 24K  dummy_erc20_token.ts
 22K  forwarder.ts
```

#### Webpack bundle analyzer
**Before**
Screen Shot 2019-04-05 at 11 37 58 am

**After**
Screen Shot 2019-04-05 at 11 38 11 am

#### File size
index.js
**Before**: 2.55MB
**After**: 1.8MB

instant.js
**Before**: 2.9MB
**After**: 2.3MB

## TODO
- [ ] Support event decoding in AbiEncoder @hysz 
- [ ] Support constructor transaction encoding in AbiEncoder @dekz 

## Thoughts
- [ ] Sideloading of JSON artifacts. Exchange duplicates tuples 16 times, whilst gzip compression optimises this it's still a chunky artifact. If we sideloaded tuples (that is define them once and reference) we could save one filesize
- [ ] Bytecode doesn't need to exist for most use cases. Migrations need this for deployment but not required at runtime. These are difficult to compress as well. Consider creating a `contract-artifacts-lite` or `contract-artifacts-dev`?
- [ ] 0x.js index.js and index.min.js are the same file size

### @0x/order-utils
7.1.1 Minified
7.2.0 Minified

### 0x.js
6.0.5 Minified
6.0.6 Minified

### @0x/contract-artifacts
1.4.0 Minified
1.5.0 Minified

### @0x/abi-gen-wrappers
4.1.0 Minified
4.2.0 Minified

### @0x/web3-wrapper
6.0.4 Minified
6.0.5 Minified

### @0x/order-watcher
4.0.5 Minified 
4.0.6 Minified

### DIFF
--- a/packages/abi-gen-wrappers/src/generated-wrappers/coordinator.ts
+++ b/packages/abi-gen-wrappers/src/generated-wrappers/coordinator.ts
@@ -6,8 +6,7 @@ import { BlockParam, BlockParamLiteral, CallData, ContractAbi, ContractArtifact,
 import { BigNumber, classUtils, logUtils, providerUtils } from '@0x/utils';
 import { SimpleContractArtifact } from '@0x/types';
 import { Web3Wrapper } from '@0x/web3-wrapper';
-import * as ethers from 'ethers';
-import * as _ from 'lodash';
+import { isUndefined } from 'lodash';
 // tslint:enable:no-unused-variable
 
 
@@ -16,6 +15,7 @@ import * as _ from 'lodash';
 // tslint:disable-next-line:class-name
 export class CoordinatorContract extends BaseContract {
     public getSignerAddress = {
+        functionSignature: 'getSignerAddress(bytes32,bytes)',
         async callAsync(
             hash: string,
             signature: string,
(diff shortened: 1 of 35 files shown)


packages/abi-gen-wrappers/src/generated-wrappers/coordinator.ts
```diff
@@ -288,9 +229,8 @@ export class CoordinatorContract extends BaseContract {
 ],
             BaseContract._bigNumberToString,
         );
-        const iface = new ethers.utils.Interface(abi);
-        const deployInfo = iface.deployFunction;
-        const txData = deployInfo.encode(bytecode, [_exchange
+        const encoder = new AbiEncoder.Constructor(constructorAbi);
```
hysz: Can we make this work with `AbiEncoder.create` as well?

packages/utils/test/abi_encoder/abi_samples/constructor_abis.ts
```diff
@@ -0,0 +1,24 @@
+import { ConstructorAbi } from 'ethereum-types';
+
+export const simpleAbi: ConstructorAbi = {
+    inputs: [
+        {
+            name: 'greg',
```
hysz: Ah good, so I'm not the only one doing this :')

packages/utils/test/abi_encoder/constructor_test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { AbiEncoder, BigNumber } from '../../src/';
+import { chaiSetup } from '../utils/chai_setup';
+
+import * as AbiSamples from './abi_samples/constructor_abis';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+describe.only('ABI Encoder: Constructor Encoding', () => {
```
hysz: Reminder to remove this before merging

packages/utils/src/abi_encoder/evm_data_types/constructor.ts
```diff
@@ -0,0 +1,27 @@
+import { ConstructorAbi, DataItem, MethodAbi } from 'ethereum-types';
+
+import { DataType } from '../abstract_data_types/data_type';
+import { DataTypeFactory } from '../abstract_data_types/interfaces';
+import { DecodingRules, EncodingRules } from '../utils/rules';
+
+import { TupleDataType } from './tuple';
+
+export class ConstructorDataType {
```
hysz: Could be good to inherit from `DataType` here. Both for consistency w other data types and so it can be used with `AbiEncoder.create()`

packages/utils/test/abi_encoder/abi_samples/constructor_abis.ts
dekz: I just copied yours :)

PR #1761, opened 2019-04-05
[3.0] Exchange Rich Reverts (once more, with feeling!)

dorothy-zbornak:
## Description
sorpresa

This is a complete overhaul and more holistic implementation of my initial proof-of-concept #1688. Kindly refer to that PR for background on the issue, encoding details, and to mock my rookie naiveté.

### Summary

This PR introduces the tools for raising, consuming, and testing rich revert errors from smart contracts. It introduces a new idiom for testing contract reverts (rich or plain) via Chai extensions. It also upgrades all errors thrown by the Exchange contract into said rich reverts.

### Notable Changes
- #### Smart Contracts
    - Rich revert solidity [library](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/contracts/utils/contracts/src/RichErrors.sol) in `contracts/utils` which exposes a `rrevert()` function.
    - Upgrade all boring string reverts in the [Exchange contracts](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/contracts/exchange/contracts/src) to awesome 
    [rich reverts](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/contracts/exchange/contracts/src/mixins/MExchangeRichErrors.sol).
- #### Tooling
    - New [`RevertError`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/utils/src/revert_error.ts) base class in `@0x/utils`:
        - Encapsulates *both* a plain revert ([`StringRevertError`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/utils/src/revert_error.ts#L254)) and rich reverts. 
        - Flexible equality testing between two instances (only mutually defined parameters are tested).
        - Pluggable registry of known types/signatures.
        - Decoding of ABI-encoded revert errors into a concrete instance.
        - Encoding of instances to ABI-encoded revert errors, which is useful in testing of [`TransactionExecutionError`s](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/contracts/exchange/contracts/src/mixins/MExchangeRichErrors.sol#L206) which carry an ABI-encoded payload of the original error. 
         - Extends `Error` for backwards compatiblity with existing tests. 
    - Concrete `RevertError` types for the Exchange in [`@0x/order-utils.ExchangeRevertErrors`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/order-utils/src/exchange_revert_errors.ts).
    - [Robust Chai extensions](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/dev-utils/src/chai_revert_error.ts) in `@0x/dev-utils`:
        - Overrides `equal()` to and `become()` to support comparisons between `RevertError` types.
        - Exposes a new method/predicate, `revertWith()`, for testing *both* transaction and call promises that should revert. It's also backwards compatible with regular string reverts, such as those in `@0x/types.RevertReason`.
        - *Should* work against ganache and geth (geth is iffy until we get our test node back up :woman_shrugging:).
        - Supplants the clunky `expectContractCallFailedAsync()` and `expectTransactionFailedAsync()` functions from `@0x/contracts-test-utils` with a nice Chai expression: `return expect(tx).to.revertWith(...)`.
    - `@0x/contracts-test-utils` and others now inherit their `chaiSetup` from [`@0x/dev-utils.chaiSetup`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/dev-utils/src/chai_setup.ts) so the new Chai extensions are available just about everywhere.  
    - Augments [`@0x/base-contract`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/base-contract/src/index.ts#L81) to automatically decode and throw `RevertError`s.

## Testing instructions

Affected contract tests have been updated/converted to handle rich reverts conform to the new testing idiom. Specifically, these packages are:
- `@0x/contracts-exchange`
- `@0x/contracts-extensions`

Tests for `RevertError`s and the new Chai extensions have been added to:
- [`@0x/utils`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/utils/test/revert_error_test.ts)
- [`@0x/dev-utils`](../blob/feature/contracts/exchange/3.0/rich-revert-reasons/packages/dev-utils/test/chai_test.ts)

## Types of changes

## Checklist:


packages/base-contract/src/index.ts
```diff
@@ -11,7 +11,6 @@ import {
     TxData,
     TxDataPayable,
 } from 'ethereum-types';
-import * as ethers from 'ethers';
```
fabioberger: Woot! Thanks for doing that! 🙏

packages/dev-utils/src/chai_revert_error.ts
```diff
@@ -0,0 +1,198 @@
...
+    condition: boolean,
+    failMessage?: string,
+    negatedFailMessage?: string,
+    expected?: any,
+    actual?: any,
+) => void;
+
+interface GanacheTransactionRevertResult {
+    error: 'revert';
+    program_counter: number;
```
fabioberger: Our convention is to always use camel-case. Is this needed to satisfy a chain interface?

fabioberger:
Approved the Typescript code only.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
```diff
@@ -142,31 +158,69 @@ contract MixinAssetProxyDispatcher is
...
-                    sub(cdEnd, cdStart),    // length of input  
+                    sub(cdEnd, cdStart),    // length of input
                     cdStart,                // write output over input
-                    512                     // reserve 512 bytes for output
+                    288                     // reserve 288 bytes for output
                 )
-                if iszero(success) {
-                    revert(cdStart, returndatasize())
+
+                if iszero(didSucceed) { // Call reverted.
```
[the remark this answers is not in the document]
dorothy-zbornak: If only solidity assembly made sense.

packages/dev-utils/src/chai_revert_error.ts
dorothy-zbornak: Yup. Those are the properties on a Ganache revert, so it's out of our control.

contracts/utils/contracts/src/RichErrors.sol
```diff
@@ -0,0 +1,53 @@
...
+{
+    // solhint-disable func-name-mixedcase
+    function StandardError(
+        string memory message
+    )
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodePacked(
```
abandeali1: Can we use `abi.encodeWithSelector` here instead?

contracts/utils/contracts/src/RichErrors.sol
```diff
@@ -0,0 +1,53 @@
...
+pragma solidity ^0.5.5;
+
+import "./mixins/MRichErrors.sol";
+
+
+contract RichErrors is
+    MRichErrors
+{
+    // solhint-disable func-name-mixedcase
+    function StandardError(
```
abandeali1: Can we add a natspec comment for this, given that it isn't used anywhere at the moment?

contracts/exchange/contracts/src/mixins/MExchangeRichErrors.sol
```diff
@@ -0,0 +1,210 @@
...
+pragma solidity ^0.5.5;
+
+import "@0x/contracts-utils/contracts/src/mixins/MRichErrors.sol";
+import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
+
+
+contract MExchangeRichErrors is
+    MRichErrors
+{
+    enum FillErrorCodes {
```
hysz: It could be beneficial to use a single enum so that errors have a unique identifier across the whole system.

hysz: If this id were in at the same offset in each message (like the first param) then it would also be easy to identify the error without decoding it, if that would be useful.

hysz: Mm although I suppose the combination of selector + error code is unique ya?

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -377,27 +381,25 @@ contract MixinExchangeCore is
     {
         // Revert if fill amount is invalid
         // TODO: reconsider necessity for v2.1
-        require(
-            takerAssetFillAmount != 0,
-            "INVALID_TAKER_AMOUNT"
-        );
-        
+        if (takerAssetFillAmount == 0) {
```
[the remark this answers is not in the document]
hysz: I'm cool with this so long as we have sufficient test vectors for each scenario that should never fail.

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -46,44 +48,51 @@ contract MixinTransactions is
...
-                isValidSignature(
+            if (!isValidSignature(
                     transactionHash,
                     signerAddress,
-                    signature
-                ),
-                "INVALID_TX_SIGNATURE"
-            );
+                    signature)) {
+                rrevert(TransactionError(
```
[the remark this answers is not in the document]
dorothy-zbornak: The validation function could just work on a copy of the signature. That's probably cleaner from a design standpoint since it isn't obvious that the validation function has side effects in the first place. If you agree with this approach, I'll make the change in #1774 after the rebase since that PR is all up in the validation function's business.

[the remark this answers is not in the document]
abandeali1: We don't copy the signature for efficiency reasons (and there isn't any other place in the contracts where it would even matter that the original signature is being mutated). A couple of other options:

- Copy the `signatureType` byte and concat it with the signature after validating (only when we need to reuse the signature after validating). Pretty dirty, but more will probably be the most efficient in the long run.
- In the validation function, use the last byte of the signature as the `signatureType` but do _not_ pop it off. If we do need to pass the signature into an external function call (only the `Wallet` or `Validator` types), then we copy the signature and pop off the last byte from the copy. This seems like probably the cleanest approach.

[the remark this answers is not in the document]
dorothy-zbornak: Second sounds great.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
```diff
@@ -142,31 +158,69 @@ contract MixinAssetProxyDispatcher is
...
+                    // --------------------------------------------------------------------------------
+                    // | Selector        | 0       | 4        | bytes4 function selector (0x08c379a0) |
+                    // | Params          |         |          |                                       |
+                    // |                 | 4       | 32       | uint256 offset to data (o)            |
+                    // | Data            |         |          |                                       |
+                    // |                 | 4 + o   | 32       | uint256 length of data (n             |
+                    // |                 | 32 + o  | n        | Left-aligned message bytes            |
+
+                    // Make sure the return value is long enough.
+                    if gt(returndatasize(), 67) {
```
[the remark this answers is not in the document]
dorothy-zbornak: We can. I just figured it would be a nicer, more language agnostic experience if there were less places to decode nested function signatures, because the decoding tooling wouldn't have to be as robust. `TransactionExecutionError` already does this very thing, but that's because the caught revert can be literally anything. Here we know it's always a string. 

Is the objection because of gas?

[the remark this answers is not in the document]
dorothy-zbornak: Actually, since the asset proxies always revert a string, a lot of the checks in the assembly are unecessary. So we're just left with:
```solidity
cdStart := add(cdStart, 4) 
revertMessage := add(cdStart, mload(cdStart))
```

contracts/exchange/contracts/src/mixins/MExchangeRichErrors.sol
dorothy-zbornak: So I think it's probably not necessary to have globally unique error codes because, as you said, the selector is already pretty unique. But, where error codes *are* used, the code ought to be the first parameter. This might be the more language agnostic play since a lazy client can just read the first 36 bytes (selector + code) to get a basic understanding of the error.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
```diff
@@ -120,19 +134,21 @@ contract MixinAssetProxyDispatcher is
                 /////// Setup State ///////
                 // `cdStart` is the start of the calldata for `assetProxy.transferFrom` (equal to free memory ptr).
                 let cdStart := mload(64)
+                // We reserve 288 bytes from `cdStart` because we'll reuse it later for return data.
+                mstore(64, add(cdStart, 288))
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, this was pretty dumb of me to do, since I don't even free the memory in the event the call *does* succeed. :trollface:

contracts/exchange/contracts/src/MixinTransactions.sol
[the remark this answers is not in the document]
dorothy-zbornak: [I went ahead and did it in this PR](https://github.com/0xProject/0x-monorepo/blob/9c104f4e88b8d7dc69fc1d461a414a79865c2de2/contracts/exchange/contracts/src/MixinSignatureValidator.sol#L305).

For `SignatureType.Wallet` and `SignatureType.Validator` I took a hacky approach by just resizing the signature before encoding the call data then restoring it immediately afterwards:

```solidity
// Shave the validator address and signature type from the signature.
assembly {
    mstore(signature, sub(signatureLength, 21))
}
// Encode the call data.
bytes memory callData = abi.encodeWithSelector(
    IValidator(validatorAddress).isValidSignature.selector,
    hash,
    signerAddress,
    signature
);
// Restore the full signature.
assembly {
    mstore(signature, signatureLength)
}
// Static call the verification function.
(bool didSucceed, bytes memory returnData) = validatorAddress.staticcall(callData);
```
I think this is safe, but let me know if you see any issues.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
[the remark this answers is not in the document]
dorothy-zbornak: Here's what [the new version](../blob/9c104f4e88b8d7dc69fc1d461a414a79865c2de2/contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol#L198) looks like:

```solidity
// Get the size of the string data/payload.
let revertDataSize := sub(returndatasize(), 36)
// We need to move the free memory pointer because we
// still have solidity logic that executes after this assembly.
mstore(64, add(cdStart, revertDataSize))
// Copy the revert string data.
returndatacopy(cdStart, 36, revertDataSize)
revertMessage := cdStart
```

contracts/exchange/contracts/src/MixinTransactions.sol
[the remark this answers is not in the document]
abandeali1: Nice, I like this approach.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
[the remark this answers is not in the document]
abandeali1: Omitting the first 36 bytes works if we assume that the revert data is always a string, but what if we move to rich revert reasons for asset proxies too? In that case, it's probably better to just include all of the return data and decode client side based off of the function selector.

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
```diff
@@ -142,31 +155,64 @@ contract MixinAssetProxyDispatcher is
...
+                    // --------------------------------------------------------------------------------
+                    // | Selector        | 0       | 4        | bytes4 function selector (0x08c379a0) |
+                    // | Params          |         |          |                                       |
+                    // |                 | 4       | 32       | uint256 offset to data (always 32)    |
+                    // | Data            |         |          |                                       |
+                    // |                 | 36      | 32       | uint256 length of data (n)            |
+                    // |                 | 68      | n        | Left-aligned message bytes            |
+
+                    // Get the size of the string data/payload.
+                    let revertDataSize := sub(returndatasize(), 36)
```
abandeali1: Omitting the first 36 bytes works if we assume that the revert data is always a string, but what if we move to rich revert reasons for asset proxies too? In that case, it's probably better to just include all of the return data and decode client side based off of the function selector.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -97,19 +102,27 @@ contract MixinSignatureValidator is
...
+                hash,
+                signerAddress,
+                signature
+            ));
+        }
 
-        // Pop last byte off of signature byte array.
-        uint8 signatureTypeRaw = uint8(signature.popLastByte());
+        // Read the last byte off of signature byte array.
+        uint8 signatureTypeRaw = uint8(signature[signature.length-1]);
```
abandeali1:
```suggestion
        uint8 signatureTypeRaw = uint8(signature[signature.length - 1]);
```

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -236,90 +257,93 @@ contract MixinSignatureValidator is
...
-            }
-            case 1 {
-                // Signature is valid if call did not revert and returned true
-                isValid := mload(cdStart)
-            }
+            mstore(signature, signatureLength)
         }
-        return isValid;
+        // Static call the verification function.
+        (bool didSucceed, bytes memory returnData) = walletAddress.staticcall(callData);
```
abandeali1: I believe that you can just do `IValidator.isValidSignature(...)` and it will compile with the `staticcall` opcode in Solidity ^0.5.0.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -236,90 +257,93 @@ contract MixinSignatureValidator is
...
     )
-        internal
+        private
         view
         returns (bool isValid)
     {
+        uint256 signatureLength = signature.length;
+        // Shave the signature type off the signature.
+        assembly {
+            mstore(signature, sub(signatureLength, 1))
```
abandeali1: I like this approach.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: Yeah, but don't we want to catch the revert?

dorothy-zbornak: We can return the revert data, too.

dorothy-zbornak: Implemented [here](../blob/dfb43ddc18ef10e9080a579e185e8ee7b8edfb33/contracts/exchange/contracts/src/MixinSignatureValidator.sol#L286) and [here](../blob/dfb43ddc18ef10e9080a579e185e8ee7b8edfb33/contracts/exchange/contracts/src/MixinSignatureValidator.sol#L344).

contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol
dorothy-zbornak: Implemented [here](../blob/dfb43ddc18ef10e9080a579e185e8ee7b8edfb33/contracts/exchange/contracts/src/MixinAssetProxyDispatcher.sol#L182).

PR #1766, opened 2019-04-05
[WIP] Replace lodash with built-ins where possible to reduce bundle size

xianny:
## Description

* Replaces `_.isUndefined` and `_.isNull` with the built-in equivalents. 
* Adds a tslint rule and auto-fix going forward

[_.isUndefined(obj)](https://lodash.com/docs/4.17.11#isUndefined) replaced with `obj === undefined`. This is the same as the lodash implementation [source](https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L12199-L12201)

[_.isNull(obj)](https://lodash.com/docs/4.17.11#isNull) replaced with `obj === null`. Also same as the lodash implementation [source](https://github.com/lodash/lodash/blob/4.17.11/lodash.js#L11950-L11952)

## Testing instructions

You can run `cd packages/tslint-config && yarn build && yarn test` to test the auto-fixer
Double-check that monorepo has no instances of the replaced methods
All tests pass on CI

## Types of changes

## Checklist:


packages/sol-tracing-utils/test/sol_compiler_artifact_adapter_test.ts
```diff
@@ -16,6 +16,7 @@ describe('SolCompilerArtifactAdapter', () => {
             const contractsData = await zeroExArtifactsAdapter.collectContractsDataAsync();
             _.forEach(contractsData, contractData => {
                 expect(contractData).to.have.keys([
+                    'name',
```
abandeali1: Was this failing before?

xianny: Yeah, it was failing locally for me. Haven't seen it fail on CI though because I fixed it before pushing. I thought it might have been different version of mocha, but even after blitzing node_modules and reinstalling everything it still failed.

PR #1773, opened 2019-04-10
[WIP] Add SignatureType.OrderValidator support to Exchange

merklejerk:
## Description
This adds support for a new order signature scheme to the Exchange: `SignatureType.OrderValidator`.

This new signature is similar to the existing `SignatureType.Validator`, but instead of passing an `bytes32 orderHash`to the validator's `isValidOrderSignature()`, we pass an `Order order` struct to the validator's `isValidOrder()`.

Here is a sample validator contract for use with this signature type:

```solidity
pragma solidity ^0.5.5;

contract OrderValidator {
    
    // 0x transaction structure
    struct Order {
        address makerAddress;
        address takerAddress;
        address feeRecipientAddress;
        address senderAddress;
        uint256 makerAssetAmount;
        uint256 takerAssetAmount;
        uint256 makerFee;
        uint256 takerFee;
        uint256 salt;
        bytes makerAssetData;
        bytes takerAssetData;
    }

    /// @dev Validate an order AND signature.
    /// @param order The full order.
    /// @param signer The signer (maker) of the order.
    /// @param signature Signature data for the order.
    /// @return `true` if valid.
    function isValidOrder(
        Order calldata order, 
        address signer, 
        bytes calldata signature
    )
        external 
        view
        returns (bool isValid)
    {
        // Validate the order
        // ...
        return true;
    }
}
```

#### Other notes:

- Just like `SignatureType.Validator`, the validator contract is called via `staticall()` and will fail if the validator contract attempts to update state.
- Validators must be registered for a maker ahead of time via `setSignatureValidatorApproval()`.
- Once registered, a validator may use *both* `SignatureType.OrderValidator` and `SignatureType.Validator` schemes.
- The public function `isValidSignature()` has been supplanted by [`isValidHashSignature()`](../blob/feature/3.0/exchange/order-validator-signature/contracts/src/interfaces/ISignatureValidator.sol#L52) and
  [`isValidOrderSignature()`](../blob/feature/3.0/exchange/order-validator-signature/contracts/src/interfaces/ISignatureValidator.sol#L66).
- Transactions use `isValidHashSignature()`, so they do not support `SignatureType.OrderValidator` (which wouldn't make sense anyway).

## Types of changes

## Checklist:

### DIFF
--- a/contracts/exchange/CHANGELOG.json
+++ b/contracts/exchange/CHANGELOG.json
@@ -17,6 +17,10 @@
             {
                 "note": "Refactor example contracts that use `executeTransaction`",
                 "pr": 1753
+            },
+            {
+                "note": "Add support for `SignatureType.OrderValidator` for orders",
+                "pr": 0
             }
         ]
     },
--- a/contracts/exchange/contracts/examples/Validator.sol
+++ b/contracts/exchange/contracts/examples/Validator.sol
@@ -17,15 +17,17 @@
 */
 
 pragma solidity ^0.5.5;
+pragma experimental ABIEncoderV2;
 
+import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
 import "../src/interfaces/IValidator.sol";
 
 
-contract Validator is 
+contract Validator is
     IValidator
 {
 
-    // The single valid signer for this wallet.
+    // The single valid signer for this validator.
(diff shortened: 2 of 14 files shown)


dorothy-zbornak:
Woops, wrong account. :innocent:

PR #1774, opened 2019-04-10
Add OrderValidator and WalletOrderValidator signature strategies to Exchange

dorothy-zbornak:
## Description
This adds support for two new order signature strategies to the Exchange: `SignatureType.OrderValidator` and `SignatureType.WalletOrderValidator`.

The new signature callbacks are similar to the existing `Validator` and `Wallet` strategies, but instead of passing just a `bytes32 orderHash` to the validator's `isValidSignature()`, we also pass an `Order order` struct to the validating contract's `isValidOrderSignature()` function (see [IWallet](../blob/feature/3.0/exchange/order-validator-signature/contracts/exchange/contracts/src/interfaces/IWallet.sol#L44) and [IOrderValidator](../blob/feature/3.0/exchange/order-validator-signature/contracts/exchange/contracts/src/interfaces/IOrderValidator.sol#L32)).

Here is a sample validator contract for use with *either* `SignatureType.OrderValidator` or `SignatureType.WalletOrderValidator` (they have identical function signatures):

```solidity
pragma solidity ^0.5.5;

contract OrderValidator {
    
    // 0x order structure
    struct Order {
        address makerAddress;
        address takerAddress;
        address feeRecipientAddress;
        address senderAddress;
        uint256 makerAssetAmount;
        uint256 takerAssetAmount;
        uint256 makerFee;
        uint256 takerFee;
        uint256 salt;
        bytes makerAssetData;
        bytes takerAssetData;
    }

    /// @dev Validate an order AND signature.
    /// @param order The full order.
    /// @param orderHash The order hash.
    /// @param signature Signature data for the order.
    /// @return `true` if valid.
    function isValidOrderSignature(
        Order calldata order,
        bytes32 orderHash,
        bytes calldata signature
    )
        external 
        view
        returns (bool isValid)
    {
        // Validate the order
        // ...
        return true;
    }
}
```

#### Other notes:

- Just like `SignatureType.Validator` and `SignatureType.Wallet`, the validator contract is called via `staticall()` and will fail if the validator contract attempts to update state.
- `OrderValidator`s must be registered for a maker ahead of time via `setOrderValidatorApproval()`, which is distinct from `setSignatureValidatorApproval()` used by `Validator`.
- The public function `isValidSignature()` has been supplanted by [`isValidHashSignature()`](../blob/feature/3.0/exchange/order-validator-signature/contracts/exchange/contracts/src/interfaces/ISignatureValidator.sol#L52) and
  [`isValidOrderSignature()`](../blob/feature/3.0/exchange/order-validator-signature/contracts/exchange/contracts/src/interfaces/ISignatureValidator.sol#L66).
- Transactions use `isValidHashSignature()`, so they do not support the new signature strategies (which wouldn't make sense anyway).

## Types of changes

## Checklist:


contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -112,7 +203,7 @@ contract MixinSignatureValidator is
         }
 
         // Read the last byte off of signature byte array.
-        uint8 signatureTypeRaw = uint8(signature[signature.length - 1]);
+        uint8 signatureTypeRaw = uint8(signature[signature.length-1]);
```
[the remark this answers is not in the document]
dorothy-zbornak: Woops. Rebase artifact.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -348,4 +334,167 @@ contract MixinSignatureValidator is
...
+    /// @dev Verifies order AND signature via Validator contract.
+    ///      If used with an order, the maker of the order can still be an EOA.
+    /// @param order The order.
+    /// @param orderHash The order hash.
+    /// @param signerAddress Address that should have signed the given hash.
+    /// @param signature Proof that the hash has been signed by signer.
+    /// @return True if the address recovered from the provided signature matches the input signer address.
+    function validateOrderValidatorSignature(
+        Order memory order,
+        bytes32 orderHash,
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, `signerAddress` can definitely go.

I see this as potentially supplanting the  `SignatureType.Validator` scheme, so it's likely validators will want the `orderHash` anyway.

dorothy-zbornak:
> At some point we should also add the `Wallet` equivalent of this signature type, but that can be in a separate PR.

I can do that in this one. The work is fairly similar. Stay tuned. :tv:

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -348,4 +334,167 @@ contract MixinSignatureValidator is
...
+        // | Offset   | Length | Contents                        |
+        // | 0x00     | x      | Signature to validate           |
+        // | 0x00 + x | 20     | Address of validator contract   |
+        // | 0x14 + x | 1      | Signature type is always "\x07" |
+
+        uint256 signatureLength = signature.length;
+        // Read the validator address from the signature.
+        address validatorAddress = signature.readAddress(signatureLength - 21);
+        // Ensure signer has approved validator.
+        if (!allowedValidators[signerAddress][validatorAddress]) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Fixed in new version.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -88,19 +95,138 @@ contract MixinSignatureValidator is
...
+    /// @param order The order.
+    /// @param signerAddress Address that should have signed the given order.
+    /// @param signature Proof that the order has been signed by signer.
+    /// @return True if the signature is valid for the given order and signer.
+    function isValidOrderSignature(
+        Order memory order,
+        address signerAddress,
+        bytes memory signature
+    )
+        public
```
abandeali1: This is never actually used internally, right? We should make it `external` if not.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -88,19 +95,138 @@ contract MixinSignatureValidator is
...
+            signerAddress,
+            signature
+        );
+    }
+
     /// @dev Verifies that a hash has been signed by the given signer.
-    /// @param hash Any 32 byte hash.
-    /// @param signerAddress Address that should have signed the given hash.
+    /// @param hash Any 32-byte hash.
+    /// @param signerAddress Address that should have signed the.Signat given hash.
```
abandeali1: Typo

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -88,19 +95,138 @@ contract MixinSignatureValidator is
...
+    function isValidHashSignature(
         bytes32 hash,
         address signerAddress,
         bytes memory signature
     )
         public
         view
         returns (bool isValid)
+    {
+        SignatureType signatureType = readValidSignatureType(hash, signerAddress, signature);
```
abandeali1: Nit: I'd default to putting any function calls with >2 arguments onto multiple lines.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -88,19 +95,138 @@ contract MixinSignatureValidator is
...
+        Order memory order,
+        bytes32 orderHash,
+        address signerAddress,
+        bytes memory signature
+    )
+        internal
+        view
+        returns (bool isValid)
+    {
+        SignatureType signatureType = readValidSignatureType(orderHash, signerAddress, signature);
```
abandeali1: Same as above.

PR #1775, opened 2019-04-11
WIP: Update devnet to post-constantinople version

albrow:
## Description

This PR updates devnet to use our updated fork of Geth, which has support for post-Constantinople features.

Still WIP only because I haven't published the latest Docker image and re-enabled Geth tests in CI yet.

## Checklist:

### DIFF
--- a/contracts/test-utils/src/assertions.ts
+++ b/contracts/test-utils/src/assertions.ts
@@ -42,7 +42,7 @@ async function _getInsufficientFundsErrorMessageAsync(): Promise<string> {
 }
 
 async function _getTransactionFailedErrorMessageAsync(): Promise<string> {
-    return _getGanacheOrGethErrorAsync('revert', 'always failing transaction');
+    return _getGanacheOrGethErrorAsync('revert', 'Transaction failed');
 }
 
 async function _getContractCallFailedErrorMessageAsync(): Promise<string> {
@@ -65,7 +65,7 @@ export async function getInvalidOpcodeErrorMessageForCallAsync(): Promise<string
  * @returns the expected error message.
  */
 export async function getRevertReasonOrErrorMessageForSendTransactionAsync(reason: RevertReason): Promise<string> {
-    return _getGanacheOrGethErrorAsync(reason, 'always failing transaction');
+    return _getGanacheOrGethErrorAsync(reason, 'Transaction failed');
 }
 
 /**
--- a/contracts/test-utils/src/web3_wrapper.ts
+++ b/contracts/test-utils/src/web3_wrapper.ts
@@ -40,7 +40,7 @@ export const txDefaults = testProvider === ProviderType.Ganache ? ganacheTxDefau
 const gethConfigs = {
     shouldUseInProcessGanache: false,
     rpcUrl: 'http://localhost:8501',
-    shouldUseFakeGasEstimate: false,
+    shouldUseFakeGasEstimate: true,
 };
 const ganacheConfigs = {
     shouldUseInProcessGanache: true,
--- a/packages/dev-utils/src/blockchain_lifecycle.ts
+++ b/packages/dev-utils/src/blockchain_lifecycle.ts
@@ -1,4 +1,3 @@
-import { logUtils } from '@0x/utils';
 import { NodeType, Web3Wrapper } from '@0x/web3-wrapper';
 import * as _ from 'lodash';
 
--- a/packages/dev-utils/src/blockchain_lifecycle.ts
+++ b/packages/dev-utils/src/blockchain_lifecycle.ts
@@ -2,13 +2,6 @@ import { logUtils } from '@0x/utils';
 import { NodeType, Web3Wrapper } from '@0x/web3-wrapper';
 import * as _ from 'lodash';
 
-// HACK(albrow): 🐉 We have to do this so that debug.setHead works correctly.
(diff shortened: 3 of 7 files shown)


contracts/test-utils/src/assertions.ts
```diff
@@ -42,7 +42,7 @@ async function _getInsufficientFundsErrorMessageAsync(): Promise<string> {
 }
 
 async function _getTransactionFailedErrorMessageAsync(): Promise<string> {
-    return _getGanacheOrGethErrorAsync('revert', 'always failing transaction');
+    return _getGanacheOrGethErrorAsync('revert', 'Transaction failed');
```
albrow: The latest version of Geth changed this error message, so I had to update our assertions.

contracts/test-utils/src/web3_wrapper.ts
```diff
@@ -40,7 +40,7 @@ export const txDefaults = testProvider === ProviderType.Ganache ? ganacheTxDefau
 const gethConfigs = {
     shouldUseInProcessGanache: false,
     rpcUrl: 'http://localhost:8501',
-    shouldUseFakeGasEstimate: false,
+    shouldUseFakeGasEstimate: true,
```
albrow: For some reason, the gas estimator in the latest version of Geth cannot estimate gas correctly when deploying our smart contracts. We've had some issues in the past but on the latest version it basically never works correctly. As a workaround we decided to disable it and just use the fake gas estimator.

packages/dev-utils/src/blockchain_lifecycle.ts
```diff
@@ -2,13 +2,6 @@ import { logUtils } from '@0x/utils';
 import { NodeType, Web3Wrapper } from '@0x/web3-wrapper';
 import * as _ from 'lodash';
 
-// HACK(albrow): 🐉 We have to do this so that debug.setHead works correctly.
```
albrow: I finally figured out why we needed this hack in the first place. The __genesis.json__ file allows you to configure the block number for certain hard forks (e.g. the DAO hard fork or the Byzantium update). We had some of these configured to block number 1, 2, 3, and 4. I updated the __genesis.json__ file to set all these block numbers to 0 so we no longer need to mine dummy blocks at the start of every test run.

albrow:
Exchange core tests are failing:

```

 |   16 passing (5s)
 |   1 failing
 | 
 |   1) Exchange core
 |        Testing exchange of ERC20 tokens with no return values
 |          should transfer the correct amounts when makerAssetAmount === takerAssetAmount:
 | 
 |       AssertionError: expected '0' to equal '-100000000000000000000'
 |       + expected - actual
 | 
 |       -0
 |       +-100000000000000000000
```

I can't reproduce this locally. Might take some time to figure out.

packages/devnet/run.sh
```diff
@@ -5,6 +5,7 @@ mkdir -p /var/log
 
 # Start Geth and direct output to stdout
 /geth \
+    --allow-insecure-unlock \
```
albrow: Due to a new security feature added in the latest version of Geth, you can't unlock accounts and enable the RPC API without this flag.

albrow:
Just to update you on the current state of our Geth fork, there are 4 commits on top of upstream Geth. One of the commits adds support for the new `debug_increaseTime` method (cherry-picked from https://github.com/ethereum/go-ethereum/pull/16834). The other three commits are hotfixes/workarounds to address some bugs in the current upstream version:


```
commit ee91ac42c646631c642f2e987dd9b3cc2689dfbf (HEAD -> 0x-testing-constantinople, 0x/0x-testing-constantinople)
Author: Alex Browne <stephenalexbrowne@gmail.com>
Date:   Fri Apr 12 14:59:52 2019 -0700

    Flush tx pool in debug_setHead

    Without flushing the tx pool, old transactions linger and are sometimes
    included in future blocks. This can mess up our tests, which assume the
    state changes in each test case are independent of all others and are
    completely reverted when you call setHead.

commit e9c114691e10a8d92ce556fb090ab13796024138
Author: Alex Browne <stephenalexbrowne@gmail.com>
Date:   Tue Jan 29 15:04:25 2019 -0800

    common, consensus, miner: introduce debug_increaseTime RPC method

commit 4b97eed19d55d0ec694cb4c08c1511c423c8dabd (fix-clique-set-head)
Author: Alex Browne <stephenalexbrowne@gmail.com>
Date:   Thu Apr 11 14:09:48 2019 -0700

    Add workaround for issue #19186

    See https://github.com/ethereum/go-ethereum/issues/19186.

    The root cause of this issue is a race condition in rpc/handler.go which
    causes eth_calls to return null (encoded in hex as "0x0"). The
    workaround, which was proposed by a community member on that issue page,
    is simply to change the order of two lines of code.

commit 3c6221e8d7bec445367da27db51f74572217e7c3
Author: Alex Browne <stephenalexbrowne@gmail.com>
Date:   Thu Apr 11 14:07:26 2019 -0700

    Add workaround for issue #19346

    See https://github.com/ethereum/go-ethereum/issues/19346.

    The root cause is a nonce check that prevents the miner from adding
    any new transactions to the block after debug_setHead is called due to
    the nonce being higher than expected. The workaround is simply to remove
    this nonce check.

```

albrow:
See https://github.com/0xProject/0x-monorepo/issues/1831#issuecomment-495340127. We're not going to try to maintain our own set of extra features on Geth moving forward.

PR #1780, opened 2019-04-17
0x/instant add Rinkeby and Ropsten to Network enum

ajm13:
## Description

0x/instant added Ropsten and Rinkeby to...
- Network enum
- AssetDataByNetwork interface
- ETHEREUM_NODE_URL_BY_NETWORK infura URLs

## Testing instructions

Render zeroExInstant with `networkId: 3` in config, and with MetaMask network set to mainnet, error message should show "Wrong network detected. Try switching to Ropsten."

## Types of changes

## Checklist:

I'm not sure which of the above are needed, since it is such a small fix.


fragosti:
@ajm13 why do you need this change? I can merge this but it may be a while before we release a new instant version.

PR #1783, opened 2019-04-22
Deprecate Pipelines package by removing README

MaCaki:
## Description

All the code in this package has been migrated into an internal repo `0x-pipeline-orchestration` where it will be maintained and refactored. 
We're not removing the actual code right now, but will do so in a subsequent PR.

## Testing instructions

N/A

## Types of changes

Deprecation via removal of the README.

## Checklist:


xianny:
I updated this to remove `pipeline` everywhere.

PR #1790, opened 2019-04-24
[3.0] Add LibExchangeRichErrorDecoder to exchange package

dorothy-zbornak:
### Summary 
This PR adds a [`LibExchangeRichErrorDecoder`](../tree/feature/3.0/contracts/exchange/rich-revert-decoder/contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol) contract to the `@0x/contracts-exchange` package. This contract exposes decoding (more like decomposing) functionality for ABI-encoded rich revert reasons (#1761).

### Notes
- `Exchange` does not use this new contract (but it could). It's designed to be inherited by a TBD standalone contract, that will likely live in separate package.
- `exchange-libs` is *not* this contract's home for two reasons:
    - This is more DRY, as it can inherit types and selectors from `exchange`'s `MExchangeRichErrorTypes`. An alternative would be to move `MExchangeRichErrorTypes` into `exchange-libs`, but that seems like overkill since no other packages use that mixin.
    - Looking ahead to when we upgrade other contracts to rich reverts, it wouldn't make sense to place their decoder libraries in `exchange-libs` because they aren't really Exchange related. Rich reverts are defined and used within their contract package, so they might as well host their own decoder libraries. 
    
## Testing instructions

You can find the new tests for this contract in [`contracts/exchange/test/src/lib_exchange_rich_error_decoder.ts`](../tree/feature/3.0/contracts/exchange/rich-revert-decoder/contracts/exchange/test/lib_exchange_rich_error_decoder.ts).

## Types of changes

## Checklist:


contracts/exchange/contracts/src/MixinExchangeRichErrors.sol
```diff
@@ -124,17 +124,17 @@ contract MixinExchangeRichErrors is
     }
 
     function OrderStatusError(
-        LibOrder.OrderStatus orderStatus,
-        bytes32 orderHash
+        bytes32 orderHash,
+        LibOrder.OrderStatus orderStatus
```
dorothy-zbornak: Swapped the order of these for consistency with other errors.

contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol
```diff
@@ -0,0 +1,443 @@
...
+        returns (
+            bytes32 orderHash
+        )
+    {
+        _assertSelectorBytes(encoded, INCOMPLETE_FILL_ERROR_SELECTOR);
+        orderHash = _readErrorParameterAsBytes32(encoded, 0);
+    }
+
+    /// @dev Revert if the leading 4 bytes of `encoded` is not `selector`.
+    function _assertSelectorBytes(bytes memory encoded, bytes4 selector)
```
hysz: I like the `_` prefix for private members, although I don't believe it's used elsewhere in our codebase. Down to standardize this if everyone else is onboard, in which case we should add it to the internal style guidelines doc.

contracts/exchange/contracts/src/mixins/MExchangeRichErrorTypes.sol
```diff
@@ -0,0 +1,109 @@
...
+        INVALID_ASSET_DATA_LENGTH,
+        UNKNOWN_ASSET_PROXY
+    }
+
+    enum TransactionErrorCodes {
+        NO_REENTRANCY,
+        ALREADY_EXECUTED
+    }
+
+    bytes4 internal constant SIGNATURE_ERROR_SELECTOR =
```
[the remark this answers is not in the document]
dorothy-zbornak: Oh, wow, I had no idea. That makes no sense to me. Definitely worth changing.

contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol
dorothy-zbornak: I think @abandeali1 is planning on switching the whole codebase over to use the `_` prefix for internal/privates. I can switch to the current convention if this is no longer on the agenda.

contracts/exchange/contracts/src/mixins/MExchangeRichErrorTypes.sol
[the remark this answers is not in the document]
abandeali1: @hysz does the new `constantOptimizer` optimizer flag change this? I haven't really looked into it, but I'm not sure what else it would do (this is already set to `true` in our `compiler.json`).

contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol
abandeali1: Yep, planning on doing a sweep to add the `_` to all private/internal functions at some point.

contracts/exchange/contracts/src/mixins/MExchangeRichErrorTypes.sol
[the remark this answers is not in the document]
dorothy-zbornak: Fixed in 8243a6f

PR #1792, opened 2019-04-25
Add coordinator wrapper

xianny:
## Description

Adds a wrapper for handling Coordinator transactions

## Testing instructions

`cd packages/contract-wrappers && yarn build && yarn test`

Unit tests cover fills, soft cancels, hard cancels, and handling server errors.

## Types of changes

## Checklist:


packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        };
+    }
+
+    private async _requestServerResponseAsync(
+        transaction: SignedZeroExTransaction,
+        senderAddress: string,
+        feeRecipientAddress: string,
+    ): Promise<CoordinatorServerApprovalResponse | CoordinatorServerCancellationResponse> {
+        assert.doesConformToSchema('transaction', transaction, schemas.zeroExTransactionSchema);
+        assert.isETHAddressHex('feeRecipientAddress', feeRecipientAddress);
```
fabioberger: These assertions are not needed. Typescript does a good enough job with verifying these types at run-time. Remove.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+
+        const coordinatorOperatorEndpoint = await this._registryInstance.getCoordinatorEndpoint.callAsync(
+            feeRecipientAddress,
+        );
+        if (coordinatorOperatorEndpoint === undefined) {
+            throw new Error(
+                `Could not find endpoint for coordinator operator with { coordinatorAddress: ${
+                    this.address
+                }, registryAddress: ${this.registryAddress}, operatorAddress: ${feeRecipientAddress}}`,
```
fabioberger: This error message feels a bit confusing to me. How about:

```
throw new Error(`No Coordinator server endpoint found in Coordinator registry for feeRecipientAddress: ${feeRecipientAddress}. Registry contract address: ${this.registryAddress}`);
```

packages/contract-wrappers/package.json
```diff
@@ -39,6 +39,7 @@
     },
     "devDependencies": {
         "@0x/contracts-test-utils": "^3.1.2",
+        "@0x/coordinator-server": "^0.0.6",
```
dekz: It looks like this is bringing in some old dependencies of contract-wrappers, might result in some strange behaviour.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+
+    private async _requestServerResponseAsync(
+        transaction: SignedZeroExTransaction,
+        senderAddress: string,
+        feeRecipientAddress: string,
+    ): Promise<CoordinatorServerApprovalResponse | CoordinatorServerCancellationResponse> {
+        assert.doesConformToSchema('transaction', transaction, schemas.zeroExTransactionSchema);
+        assert.isETHAddressHex('feeRecipientAddress', feeRecipientAddress);
+
+        const coordinatorOperatorEndpoint = await this._registryInstance.getCoordinatorEndpoint.callAsync(
```
fabioberger: I feel strongly we should do this step _before_ we ask the user to _sign_ the 0x transaction. If the endpoint doesn't exist in the registry, the signing process is a waste of time. Can we move this up?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        orders: SignedOrder[],
+        orderTransactionOpts: OrderTransactionOpts = { shouldValidate: true },
+    ): Promise<string> {
+        assert.doesConformToSchema('orders', orders, schemas.ordersSchema);
+        const makerAddress = getMakerAddressOrThrow(orders);
+        assert.doesConformToSchema('orderTransactionOpts', orderTransactionOpts, orderTxOptsSchema, [txOptsSchema]);
+        await assert.isSenderAddressAsync('makerAddress', makerAddress, this._web3Wrapper);
+
+        const data = this._transactionEncoder.batchCancelOrdersTx(orders);
+        const transaction = await this._toSignedZeroExTransactionAsync(data, makerAddress);
```
dekz: Do we need to do it this way for hard cancels? The resulting UX would be 2 MM prompts wouldn't it?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+            throw new Error(
+                `Could not find endpoint for coordinator operator with { coordinatorAddress: ${
+                    this.address
+                }, registryAddress: ${this.registryAddress}, operatorAddress: ${feeRecipientAddress}}`,
+            );
+        }
+
+        const requestPayload = {
+            signedTransaction: transaction,
+            txOrigin: senderAddress,
```
fabioberger: `txOrigin` should default  to `takerAddress`. `txOrigin` specifies who can ultimately submit the Ethereum transaction to the network.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+                body: JSON.stringify(requestPayload),
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json; charset=utf-8',
+                },
+            },
+        );
+
+        if (response.status !== HTTP_OK) {
+            throw new Error(`${response.status}: ${JSON.stringify(await response.json())}`); // todo (xianny)
```
fabioberger: Is this error message going to improved still?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        // const data = this._transactionEncoder.cancelOrdersUpToTx(targetOrderEpoch);
+        // const transaction = await this._toSignedZeroExTransactionAsync(data, senderAddress);
+        // return this._sendTransactionAsync(
+        //     transaction,
+        //     senderAddress,
+        //     transaction.signature,
+        //     [],
+        //     [],
+        //     orderTransactionOpts,
+        // );
```
fabioberger: Is this outstanding?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        );
+    }
+
+    public async getSignerAddressAsync(hash: string, signature: string): Promise<string> {
+        assert.isHexString('hash', hash);
+        assert.isHexString('signature', signature);
+        return this._contractInstance.getSignerAddress.callAsync(hash, signature);
+    }
+
+    private async _sendTransactionAsync(
```
fabioberger:
```suggestion
    private async _submitCoordinatorTransactionAsync(
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+            this.exchangeAddress,
+            this._web3Wrapper.getProvider(),
+            this._web3Wrapper.getContractDefaults(),
+        );
+
+        this._transactionEncoder = new TransactionEncoder(this._exchangeInstance);
+    }
+
+    /**
+     * Fills a signed order with an amount denominated in baseUnits of the taker asset.
```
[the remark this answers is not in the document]
fabioberger: Let's put a more detailed description of what these methods do on all the subsequent methods as well.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+     * @return  Transaction hash.
+     */
+    @decorators.asyncZeroExErrorHandler
+    public async batchFillOrdersAsync(
+        signedOrders: SignedOrder[],
+        takerAssetFillAmounts: BigNumber[],
+        takerAddress: string,
+        orderTransactionOpts: OrderTransactionOpts = { shouldValidate: true },
+    ): Promise<string> {
+        assert.doesConformToSchema('signedOrders', signedOrders, schemas.signedOrdersSchema);
```
fabioberger: Add assertions for all the params of this method.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        for (const takerAssetFillAmount of takerAssetFillAmounts) {
+            assert.isBigNumber('takerAssetFillAmount', takerAssetFillAmount);
+        }
+        const feeRecipientAddress = getFeeRecipientOrThrow(signedOrders);
```
fabioberger: There can be more then one `feeRecipientAddress` in the supplied orders. If there are multiple coordinator server involved in a batch fill, the client must send a request to every coordinator server and must retrieve a signature from each. Sorry if this was not clear.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+     * the orders are both being filled at the right order's price point.
+     * The profit made by the left order goes to the taker (whoever matched the two orders).
+     * @param leftSignedOrder  First order to match.
+     * @param rightSignedOrder Second order to match.
+     * @param takerAddress     The address that sends the transaction and gets the spread.
+     * @param orderTransactionOpts Optional arguments this method accepts.
+     * @return Transaction hash.
+     */
+    @decorators.asyncZeroExErrorHandler
+    public async matchOrdersAsync(
```
fabioberger: Let's remove this method. It's not currently supported by the Coordinator server implementation.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        const transaction = await this._toSignedZeroExTransactionAsync(data, makerAddress);
+        return (await this._requestServerResponseAsync(
+            transaction,
+            makerAddress,
+            feeRecipientAddress,
+        )) as CoordinatorServerCancellationResponse;
+    }
+
+    /**
+     * Hard cancellation on Exchange contract. Cancels a single order.
```
fabioberger:
```suggestion
     * Hard cancellation cancels an order on-chain and involves an Ethereum transaction.
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+            order.makerAddress,
+            transaction.signature,
+            [],
+            [],
+            orderTransactionOpts,
+        );
+    }
+
+    /**
+     * Hard cancellation on Exchange contract.
```
fabioberger:
```suggestion
     * Batch hard cancellation on Exchange contract.
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        //     transaction,
+        //     senderAddress,
+        //     transaction.signature,
+        //     [],
+        //     [],
+        //     orderTransactionOpts,
+        // );
+    }
+
+    public async assertValidCoordinatorApprovalsAsync(
```
fabioberger: If this is a public method, it needs doc comments.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        return this._contractInstance.assertValidCoordinatorApprovals.callAsync(
+            transaction,
+            txOrigin,
+            signature,
+            approvalExpirationTimeSeconds,
+            approvalSignatures,
+        );
+    }
+
+    public async getSignerAddressAsync(hash: string, signature: string): Promise<string> {
```
fabioberger: Ditto re: doc comments.

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -0,0 +1,359 @@
...
+        app = await getAppAsync(
+            {
+                [config.networkId]: provider,
+            },
+            coordinatorServerConfigs,
+        );
+
+        await app.listen(coordinatorServerConfigs.HTTP_PORT, () => {
+            // tslint:disable-line:await-promise
+            console.log(`Coordinator SERVER API (HTTP) listening on port ${coordinatorServerConfigs.HTTP_PORT}!`); // tslint:disable-line:no-console
```
fabioberger: Let's use the `logUtils.log` method here instead of `console.log`.

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -0,0 +1,359 @@
...
+        };
+        app = await getAppAsync(
+            {
+                [config.networkId]: provider,
+            },
+            coordinatorServerConfigs,
+        );
+
+        await app.listen(coordinatorServerConfigs.HTTP_PORT, () => {
+            // tslint:disable-line:await-promise
```
fabioberger: This doesn't seem necessary.

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -0,0 +1,359 @@
...
+            console.log(`Coordinator SERVER API (HTTP) listening on port ${coordinatorServerConfigs.HTTP_PORT}!`); // tslint:disable-line:no-console
+        });
+
+        // setup coordinator registry
+        coordinatorRegistryInstance = new CoordinatorRegistryContract(
+            CoordinatorRegistry.compilerOutput.abi,
+            contractAddresses.coordinatorRegistry,
+            provider,
+            { gas: 6000000 },
```
fabioberger: Is this really needed?

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -0,0 +1,359 @@
...
+                feeRecipientAddress,
+            );
+            expect(recordedCoordinatorEndpoint).to.be.equal(coordinatorEndpoint);
+        });
+        it('should have coordinator server endpoint which responds to pings', async () => {
+            const result = await fetchAsync(`${coordinatorEndpoint}/v1/ping`);
+            expect(result.status).to.be.equal(200);
+            expect(await result.text()).to.be.equal('pong');
+        });
+    });
```
fabioberger: Let's add some failure cases. Some that come to mind are:

- `feeRecipientAddress` is not in registry.
- `coordinatorEndpoint` in registry is incorrect. 

An additional edge-case:
- Batch fill with 2 orders each specifying a different coordinator server

packages/contract-wrappers/test/coordinator_wrapper_test.ts
xianny: oops, that was leftover from when I was using the ganache snapshot in docker  instead of in-process ganache. removing.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: this sounds way better, thanks!

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+        }
+
+        return (await response.json()) as CoordinatorServerApprovalResponse | CoordinatorServerCancellationResponse;
+    }
+
+    private async _handleFillAsync(
+        data: string,
+        takerAddress: string,
+        feeRecipientAddress: string,
+        orderTransactionOpts: OrderTransactionOpts = { shouldValidate: true },
```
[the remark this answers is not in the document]
xianny: I don't have a strong opinion on this, but in every other place where it's accepted as an argument, this default is provided.

I'll defer to you on it if you still think we should remove it?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: Yeah. Not sure why it doesnt work. It looks like the Coordinator contract should just be forwarding the transaction to the Exchange contract. Cancellations skip approval verification.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: Yep, agree. I'll fix it

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: This is passed in from any calling methods but the variable is named `txOrigin` for consistency with the request payload schema that the server expects. When called for a fill approval it's passed as the takerAddress, and when called for a soft cancellation it's the makerAddress. Is that right?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: yep!

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: Not sure I understand - is that because the Exchange method also does a MM prompt? 

Looks like the coordinator contract under the hood skips everything for cancels and forwards directly to the Exchange contract so we could call the exchange method directly.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,672 @@
...
+    private async _requestServerResponseAsync(
+        transaction: SignedZeroExTransaction,
+        senderAddress: string,
+        feeRecipientAddress: string,
+    ): Promise<CoordinatorServerApprovalResponse | CoordinatorServerCancellationResponse> {
+        assert.doesConformToSchema('transaction', transaction, schemas.zeroExTransactionSchema);
+        assert.isETHAddressHex('feeRecipientAddress', feeRecipientAddress);
+
+        const coordinatorOperatorEndpoint = await this._registryInstance.getCoordinatorEndpoint.callAsync(
+            feeRecipientAddress,
```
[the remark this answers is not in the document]
xianny: reimplementing

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: updated it, what do you think?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
dekz: Ignore this my brain wasn't working!

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
dekz: It must go through the Coordinator correct? Otherwise the `cancelOrdersUpTo` will cancel orders for tuple
`(makerAddress, NULL_ADDRESS) with salt > targetOrderEpoch`. Rather than `(makerAddress, coordinatorAddress) with salt > targetOrderEpoch `

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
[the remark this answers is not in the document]
dekz: I think it is fine to default in the `public` method. 

My take on having it defaulted in a `private` method is that a future developer could make a change and forget to pass the option from the user down into the `private` method resulting in unwanted behaviour. E.g it is specified as `false` in `hardCancelOrdersUpToAsync(...., false)` but we forget to pass that on and then becomes `true` in `_handleFillAsync` from the default. My thought is this makes it safer.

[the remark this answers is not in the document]
xianny: makes sense to me!

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: Ahhh ok, I see what I was doing wrong. the test was retrieving orders for  (makerAddress, NULL_ADDRESS) instead of (makerAddress, coordinatorAddress). fixed it :)

xianny:
I'd still like to write more tests for error cases to check the format of the errors, but apart from that I've implemented all the changes brought up. 

I changed everything to return `Promise<txhash | CoordinatorError>` at first, then changed it back to throw errors. But had to stringify `CoordinatorError` objects. Feeling ambivalent about this. If we return instead of throwing, it breaks the resolve/reject interface for Promises. Is there a way to throw an error with an object instead of just a string?

xianny:
tried to rebase on development and something funky happened. Managed to reset it but that's why there are weird events in the log.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            feeRecipientsToOrders[feeRecipient] = feeRecipientsToOrders[feeRecipient].concat(order);
+        }
+
+        const serverEndpointsToFeeRecipients: { [feeRecipient: string]: string[] } = {};
+        for (const feeRecipient of Object.keys(feeRecipientsToOrders)) {
+            const endpoint = await this._getServerEndpointOrThrowAsync(feeRecipient);
+            if (serverEndpointsToFeeRecipients[endpoint] === undefined) {
+                serverEndpointsToFeeRecipients[endpoint] = [];
+            }
+            serverEndpointsToFeeRecipients[endpoint] = serverEndpointsToFeeRecipients[endpoint].concat(feeRecipient);
```
fabioberger: `push`

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+
+        // make server requests
+        let numErrors = 0;
+        let errorResponses: CoordinatorServerResponse[] = [];
+        let approvalResponses: CoordinatorServerResponse[] = [];
+        const transaction = await this._generateSignedZeroExTransactionAsync(data, takerAddress);
+        for (const endpoint of Object.keys(serverEndpointsToFeeRecipients)) {
+            const response = await this._executeServerRequestAsync(transaction, takerAddress, endpoint);
+            if (response.isError) {
+                errorResponses = errorResponses.concat(response);
```
fabioberger: `push` everywhere.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+                .reduce((previous, current) => {
+                    previous.signatures = previous.signatures.concat(current.signatures);
+                    previous.expirationTimeSeconds = previous.expirationTimeSeconds.concat(
+                        current.expirationTimeSeconds,
+                    );
+                    return previous;
+                });
+
+            // submit transaction with approvals
+            return this._submitCoordinatorTransactionAsync(
```
fabioberger: Let's assign and return so that it's clearer what is getting returned.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+                allApprovals.signatures,
+                orderTransactionOpts,
+            );
+        } else {
+            // format errors and approvals
+            // concatenate approvals
+            const approvedOrders = approvalResponses
+                .map(resp => {
+                    const endpoint = resp.coordinatorOperator;
+                    const feeRecipients: string[] = serverEndpointsToFeeRecipients[endpoint];
```
fabioberger: Is this type declaration `string[]` necessary?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+                return {
+                    ...resp,
+                    orders: _orders,
+                };
+            });
+
+            // return errors and approvals
+            throw new CoordinatorServerError(
+                CoordinatorServerErrorMsg.CancellationFailed,
+                [],
```
fabioberger: For readability, our convention is to declare a variable with a descriptive name rather then passing in values directly. Which is more readable to you:

```
const orders = getOrders(2, [], false);
```

or

```
const limit = 2;
const excludedFields = [];
const isFilled = false;
const orders = getOrders(limit, excludedFields, isFilled);
```

?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+                successResponses,
+                errorsWithOrders,
+            );
+        }
+    }
+
+    /**
+     * Cancels an order on-chain and involves an Ethereum transaction.
```
fabioberger:
```suggestion
     * Cancels an order on-chain by submitting an Ethereum transaction.
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+        await assert.isSenderAddressAsync('makerAddress', order.makerAddress, this._web3Wrapper);
+
+        const data = this._transactionEncoder.cancelOrderTx(order);
+        const transaction = await this._generateSignedZeroExTransactionAsync(data, order.makerAddress);
+        return this._submitCoordinatorTransactionAsync(
+            transaction,
+            order.makerAddress,
+            transaction.signature,
+            [],
+            [],
```
fabioberger: ditto above re: passing named variables.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            orderTransactionOpts,
+        );
+    }
+
+    /**
+     * Batch version of hardCancelOrderAsync. Cancels orders on-chain and involves an Ethereum transaction.
```
fabioberger:
```suggestion
     * Batch version of hardCancelOrderAsync. Cancels orders on-chain by submitting an Ethereum transaction.
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            makerAddress,
+            transaction.signature,
+            [],
+            [],
+            orderTransactionOpts,
+        );
+    }
+
+    /**
+     * Cancels orders on-chain and involves an Ethereum transaction.
```
fabioberger:
```suggestion
     * Cancels orders on-chain by submitting an Ethereum transaction.
```

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+        orderTransactionOpts: OrderTransactionOpts = { shouldValidate: true },
+    ): Promise<string> {
+        assert.doesConformToSchema('signedOrder', signedOrder, schemas.signedOrderSchema);
+        assert.isValidBaseUnitAmount('takerAssetFillAmount', takerAssetFillAmount);
+        assert.isETHAddressHex('takerAddress', takerAddress);
+        assert.doesConformToSchema('orderTransactionOpts', orderTransactionOpts, orderTxOptsSchema, [txOptsSchema]);
+        await assert.isSenderAddressAsync('takerAddress', takerAddress, this._web3Wrapper);
+
+        const data = this._transactionEncoder.fillOrderTx(signedOrder, takerAssetFillAmount);
+        return this._handleFillsAsync(data, takerAddress, [signedOrder], orderTransactionOpts);
```
fabioberger: Assign then return.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+        }
+
+        const serverEndpointsToFeeRecipients: { [feeRecipient: string]: string[] } = {};
+        for (const feeRecipient of Object.keys(feeRecipientsToOrders)) {
+            const endpoint = await this._getServerEndpointOrThrowAsync(feeRecipient);
+            if (serverEndpointsToFeeRecipients[endpoint] === undefined) {
+                serverEndpointsToFeeRecipients[endpoint] = [];
+            }
+            serverEndpointsToFeeRecipients[endpoint].push(feeRecipient);
+        }
```
fabioberger: This logic is duplicated in at least two locations. Let's functionally decompose and make it DRY.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            },
+        });
+
+        const isError = response.status !== HTTP_OK;
+
+        let json;
+        try {
+            json = await response.json();
+        } catch (e) {
+            // ignore
```
fabioberger: Hm, this feel dangerous to me. When would this fail?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            error: isError ? json : undefined,
+            request: requestPayload,
+            coordinatorOperator: endpoint,
+        };
+
+        return result;
+    }
+
+    private async _submitCoordinatorTransactionAsync(
+        transaction: { salt: BigNumber; signerAddress: string; data: string },
```
fabioberger: Let's not in-line type declarations. This might by one of the reasons why your CI doc generation test is failing.

packages/contract-wrappers/src/types.ts
```diff
@@ -225,3 +225,61 @@ export interface DutchAuctionData {
...
+}
+
+export interface CoordinatorServerCancellationResponse {
+    outstandingFillSignatures: [
+        {
+            orderHash: string;
+            approvalSignatures: string[];
+            expirationTimeSeconds: BigNumber;
+            takerAssetFillAmount: BigNumber;
+        }
```
fabioberger: Make this a separate type declaration. Our docs cannot handle infinitely nested type definitions yet

packages/contract-wrappers/src/types.ts
```diff
@@ -225,3 +225,61 @@ export interface DutchAuctionData {
...
+}
+
+export interface CoordinatorServerResponse {
+    isError: boolean;
+    status: number;
+    body?: CoordinatorServerCancellationResponse | CoordinatorServerApprovalRawResponse;
+    error?: any;
+    request: {
+        signedTransaction: SignedZeroExTransaction;
+        txOrigin: string;
```
fabioberger: Ditto above here.

packages/contract-wrappers/src/types.ts
```diff
@@ -225,3 +225,61 @@ export interface DutchAuctionData {
...
+    };
+    coordinatorOperator: string;
+    orders?: Array<SignedOrder | Order>;
+}
+
+export class CoordinatorServerError extends Error {
```
fabioberger: I wouldn't consider classes as types (as understood in Typescript). Let's move this to it's own file.

packages/contract-wrappers/src/types.ts
```diff
@@ -225,3 +225,61 @@ export interface DutchAuctionData {
...
+        }
+    ];
+    cancellationSignatures: string[];
+}
+
+export interface CoordinatorServerResponse {
+    isError: boolean;
+    status: number;
+    body?: CoordinatorServerCancellationResponse | CoordinatorServerApprovalRawResponse;
+    error?: any;
```
fabioberger: Why `any`?

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -0,0 +1,689 @@
...
+                    takerAssetFillAmounts,
+                    takerAddress,
+                );
+
+                await web3Wrapper.awaitTransactionSuccessAsync(txHash, constants.AWAIT_TRANSACTION_MINED_MS);
+            });
+            // coordinator-server, as currently implemented, shares a singleton database connection across
+            // all instantiations. Making the request to two different mocked server endpoints still hits the
+            // same database and fails because of the uniqueness constraint on transactions in the database.
+            it.skip('should fill a batch of orders with different feeRecipientAddresses with different coordinator servers', async () => {
```
fabioberger: Will this still get fixed before we merge?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+import { TransactionEncoder } from '../utils/transaction_encoder';
+
+import { ContractWrapper } from './contract_wrapper';
+
+const HTTP_OK = 200;
+
+/**
+ * This class includes all the functionality related to filling or cancelling orders through
+ * the 0x V2 Coordinator smart contract.
+ * All logic related to filling orders is implemented in the private method `_handleFillsAsync`.
```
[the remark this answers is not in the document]
xianny: I'll remove it :)

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+        const feeRecipientsToOrders: { [feeRecipient: string]: SignedOrder[] } = {};
+        for (const order of signedOrders) {
+            const feeRecipient = order.feeRecipientAddress;
+            if (feeRecipientsToOrders[feeRecipient] === undefined) {
+                feeRecipientsToOrders[feeRecipient] = [] as SignedOrder[];
+            }
+            feeRecipientsToOrders[feeRecipient] = feeRecipientsToOrders[feeRecipient].concat(order);
+        }
+
+        const serverEndpointsToFeeRecipients: { [feeRecipient: string]: string[] } = {};
```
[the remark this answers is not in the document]
xianny: it's `endpoint -> feeRecipients`, the typedef was wrong. Fixing!

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,857 @@
...
+            };
+        }
+    }
+
+    private async _getServerEndpointOrThrowAsync(feeRecipientAddress: string): Promise<string> {
+        const coordinatorOperatorEndpoint = await this._registryInstance.getCoordinatorEndpoint.callAsync(
+            feeRecipientAddress,
+        );
+        if (coordinatorOperatorEndpoint === '') {
+            throw new Error(
```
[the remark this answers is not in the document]
xianny: If `senderAddress = coordinator address` then it *has* to have a valid fee recipient address right? where "valid" means registered in the coordinator registry. That's also how Coordinator.sol works. 

I'll add a check to filter for orders that only have coordinator as the sender address, and throw if these coordinator orders don't have a valid fee recipient?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: nope, removing

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: this would fail if the server doesn't return a json response. e.g. something unexpected happens on server and it returns 500. I've had this happen a few times in testing, usually an uncaught exception from typeorm. 

This method basically tries to wrap server responses in a predictable format. The `json` field is either useful details or defaults to undefined, but `isError` and `status` are always informative.

packages/contract-wrappers/src/types.ts
xianny: Hmm. on second thought, this could be `ValidationError[]` with the type imported from the server. I guess the server will never return any json that isn't a success response known error type.

packages/contract-wrappers/test/coordinator_wrapper_test.ts
xianny: No, this involves fixing the server side of things. I still think it's out of scope like we discussed. But will make a task to fix it after merging.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: makes sense! will do

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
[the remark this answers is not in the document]
fabioberger: Yep, I like that idea. Then users can batch fill OO and Coordinator orders.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
fabioberger: If it only happens when a non-200 is returned, I'd rather we do:

```
const json = isError ? undefined : await response.json();
```

That way, if we are expecting a body, and `.json()` throws, it won't be swallowed. Generally very cautious of swallowing thrown errors since a lot of bugs go undiscovered in this way.

packages/contract-wrappers/src/types.ts
xianny: Looking closer at the server response types, looks like the error response isn't exactly the formally defined type either. 

Response: https://github.com/0xProject/0x-coordinator-server/blob/master/ts/src/middleware/error_handling.ts#L34-L47

Exported type: https://github.com/0xProject/0x-coordinator-server/blob/master/ts/src/errors.ts#L19

Leaning towards leaving it as `any`, at least for now. This error field is never surfaced on its own, it's just for additional details from the server. The consumer always has access to other informative fields.

The data in this field will either be undefined or some kind of validation error so would have to be surfaced to end user anyway right? As in, even if we strongly type this field, it's not likely that the developer would be able to do automated error correction?

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: I simplified this part but left the rest separate (handling server responses) because slightly different things need to happen for fills vs cancels.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
xianny: we actually expect a body for 200 and 400 but that's easy to express. Changing it!

packages/order-utils/src/index.ts
```diff
@@ -68,7 +68,7 @@ export {
 } from '@0x/types';
 
 export {
-    OrderError,
+    TypedDataError as OrderError,
```
xianny: @fabioberger should I export this as `TypedData` and add it to CHANGELOG?

fabioberger: Yes, let's make this change.

packages/order-utils/src/signature_utils.ts
```diff
@@ -278,7 +286,50 @@ export const signatureUtils = {
...
             if ((provider as any).isMetaMask) {
-                throw new Error(OrderError.InvalidMetamaskSigner);
+                throw new Error(TypedDataError.InvalidMetamaskSigner);
+            } else {
+                throw err;
+            }
+        }
+    },
+    /**
+     * Signs an order using `eth_signTypedData` and returns a SignedOrder.
```
fabioberger: Update this comment to `transaction`

packages/order-utils/src/signature_utils.ts
```diff
@@ -278,7 +286,50 @@ export const signatureUtils = {
...
+     * @param   supportedProvider      Web3 provider to use for all JSON RPC requests
+     * @param   transaction The ZeroEx Transaction to sign.
+     * @param   signerAddress   The hex encoded Ethereum address you wish to sign it with. This address
+     *          must be available via the supplied Provider.
+     * @return  A SignedOrder containing the order and Elliptic curve signature with Signature Type.
+     */
+    async ecSignTypedDataTransactionAsync(
```
fabioberger: Let's also add a CHANGELOG entry for `order-utils` with a minor version bump detailing the addition of this method.

packages/contract-wrappers/src/contract_wrappers/coordinator_wrapper.ts
```diff
@@ -0,0 +1,855 @@
...
+} from '../utils/coordinator_server_types';
+import { decorators } from '../utils/decorators';
+import { TransactionEncoder } from '../utils/transaction_encoder';
+
+import { ContractWrapper } from './contract_wrapper';
+/**
+ * This class includes all the functionality related to filling or cancelling orders through
+ * the 0x V2 Coordinator extension contract.
+ */
+export class CoordinatorWrapper extends ContractWrapper {
```
fabioberger: Add a CHANGELOG entry to `contract-wrappers` detailing the addition of this wrapper.

packages/order-utils/src/signature_utils.ts
xianny: shouldn't it be major version bump because we also change `OrderError` to `TypedDataError`?

packages/contract-wrappers/src/utils/coordinator_server_types.ts
```diff
@@ -0,0 +1,61 @@
...
+import { BigNumber } from '@0x/utils';
+
+export interface CoordinatorServerApprovalResponse {
+    signatures: string[];
+    expirationTimeSeconds: BigNumber[];
+}
+export interface CoordinatorServerApprovalRawResponse {
+    signatures: string[];
+    expirationTimeSeconds: BigNumber;
+}
```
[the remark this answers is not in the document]
xianny: the Coordinator contract expects signatures and expiration times to be a same length array. It iterates through signatures and compare to expiration times at the same index to decide whether the signature is still valid or not.

packages/order-utils/src/signature_utils.ts
xianny: ooops. thanks for catching

packages/contract-wrappers/src/utils/coordinator_server_types.ts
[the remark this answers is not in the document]
fabioberger: Ah yep, makes sense.

xianny:
`static-tests-python` is failing everywhere else too. I'm going to call it unrelated and merge this. 🎉

PR #1802, opened 2019-05-03
Transaction decoder for Coordinator

feuGeneA:
## Description

Add a transaction decoder library, for use in decoding events from the Coordinator.  [Asana task](https://app.asana.com/0/711604532004534/1115993806396758/f)

"Spec" (informal conversation) from @abandeali1 :
> for the transaction parser, the discussion started because of this https://github.com/0xProject/0x-protocol-specification/blob/master/v2/coordinator-specification.md#fill_request_accepted
> having the `functionName` and the `orders` felt redundant because all of that information is already technically included in `signedTransaction.data`
> but we decided to leave those fields in for now since it's not super easy to decode the data
> so basically want a function that takes some Exchange calldata as an input and returns its decoded parameters
> the output probably looks something like: `{ functionName: ..., orders: [...], takerAssetFillAmounts: [...], signatures: [...] }`
> where for single fill functions, those fields would just be arrays of length 1
> I think it's a bit tricky to figure out the exactly right schema for the return data, since all of these functions take slightly different params

- [x]    function batchCancelOrders
- [x]    function batchFillOrders
- [x]    function batchFillOrdersNoThrow
- [x]    function batchFillOrKillOrders
- [x]    function cancelOrder
- [x]    function fillOrder
- [x]    function fillOrderNoThrow
- [x]    function fillOrKillOrder
- [x]    function marketBuyOrders
- [x]    function marketBuyOrdersNoThrow
- [x]    function marketSellOrders
- [x]    function marketSellOrdersNoThrow
~- [ ]    function cancelOrdersUpTo~
~- [ ]    function executeTransaction~
- [x]    function matchOrders
- [x]    add devdoc comments
- [ ]    update changelog

## Testing instructions

`yarn test` (integrated with CI)

## Types of changes


contracts/exchange-libs/contracts/src/LibTransactionDecoder.sol
```diff
@@ -0,0 +1,162 @@
...
+            uint256[] memory takerAssetFillAmounts,
+            bytes[] memory signatures
+        )
+    {
+        bytes4 functionSelector = signedTransactionData.readBytes4(0);
+
+        if (functionSelector == BATCH_CANCEL_ORDERS_SELECTOR) {
+            functionName = "batchCancelOrders";
+        } else if (functionSelector == BATCH_FILL_ORDERS_SELECTOR) {
+            functionName = "batchFillOrders";
```
feuGeneA: I wonder if this giant if/else shouldn't be wrapped in a function in `LibExchangeSelectors`.

feuGeneA:
Amir's initial guess at the output (the `{functionName: ..., orders: [...], takerAssetFillAmounts: [...], signatures: [...]}` in the quote in the original comment), has held up will for 12 of the 15 functions.  However, those last 3 are significantly different from the others...

contracts/exchange-libs/contracts/src/LibTransactionDecoder.sol
```diff
@@ -0,0 +1,191 @@
...
+            takerAssetFillAmounts[1] = rightOrder.takerAssetAmount;
+
+            signatures = new bytes[](2);
+            signatures[0] = leftSignature;
+            signatures[1] = rightSignature;
+        }
+    }
+
+    function _makeReturnValuesForSingleOrderFill(bytes memory signedTransactionData)
+        internal
```
dorothy-zbornak: Maybe these should be `private` since I don't foresee anyone ever using them.

contracts/exchange-libs/test/lib_transaction_decoder.ts
```diff
@@ -0,0 +1,234 @@
...
+    // code that generated them.  Running that code depends on the following:
+    //
+    //     import { artifacts as exchangeArtifacts, ExchangeContract } from '@0x/contracts-exchange';
+    //     import { getContractAddressesForNetworkOrThrow, NetworkId } from '@0x/contract-addresses';
+    //     const exchangeContract = new ExchangeContract(
+    //         exchangeArtifacts.Exchange.compilerOutput.abi,
+    //         getContractAddressesForNetworkOrThrow(NetworkId.Ganache).exchange,
+    //         provider,
+    //     );
+    //
```
dorothy-zbornak: :+1:

contracts/exchange-libs/contracts/src/LibTransactionDecoder.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    /// @dev Decodes the call data for an Exchange contract method call.
+    /// @param signedTransactionData ABI-encoded calldata for an Exchange
+    ///     contract method call.
+    /// @return The name of the function called, and the parameters it was
+    ///     given.  For single-order fills and cancels, the arrays will have
+    ///     just one element.
+    function decodeTransaction(bytes memory signedTransactionData)
+        public
+        view
```
dorothy-zbornak: Can't this be `pure`?

dorothy-zbornak: Can't this be `pure`?

feuGeneA: I swore I tried it as `pure` and hit a problem because of accessing the selector constants, but I just changed it to `pure` locally and it built and tested just fine.  I'll change it.

contracts/exchange-libs/contracts/src/LibTransactionDecoder.sol
```diff
@@ -0,0 +1,191 @@
...
+contract LibTransactionDecoder is LibExchangeSelectors {
+    using LibBytes for bytes;
+
+    /// @dev Decodes the call data for an Exchange contract method call.
+    /// @param signedTransactionData ABI-encoded calldata for an Exchange
+    ///     contract method call.
+    /// @return The name of the function called, and the parameters it was
+    ///     given.  For single-order fills and cancels, the arrays will have
+    ///     just one element.
+    function decodeTransaction(bytes memory signedTransactionData)
```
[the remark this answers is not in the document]
abandeali1: The function name should also probably be a bit more specific. I like `decodeZeroExTransactionData`.

contracts/exchange-libs/contracts/src/LibTransactionDecoder.sol
```diff
@@ -0,0 +1,191 @@
...
+            orders[1] = rightOrder;
+
+            takerAssetFillAmounts = new uint256[](2);
+            takerAssetFillAmounts[0] = leftOrder.takerAssetAmount;
+            takerAssetFillAmounts[1] = rightOrder.takerAssetAmount;
+
+            signatures = new bytes[](2);
+            signatures[0] = leftSignature;
+            signatures[1] = rightSignature;
+        }
```
hysz: Nit - could be good to add an `else` here as well that reverts. There are so many selectors now; this way we'll know if one is ever unhandled.

feuGeneA: I declined to catch unknown selectors in this `if` because I already set up such an else in the previous `if` in this function.

PR #1803, opened 2019-05-04
Added performedCallback attribute for Redux state, and added a hook f…

PirosB3:
## Description

0x Instant currently does not have a "success callback" feature that allows developers to listen to. This callback can be used by application developers to create side effects of successfully filling an order.

Example:
A web application has an online store in DAI or USDC. Upon checkout, the application prompts the customer to buy some DAI using Instant (since their balance is too low). Once the checkout is complete, the web application can automatically close the Instant window and proceed with the checkout.

## Changes

In the current 0x Instant, the web-app has no control over if the rendering of 0x Instant widget resulted in a trade.
In the new 0x Instant (after this fix), the web-app can be called when a successful 0x trade happens (the event also contains the transaction hash) and take control of the application.

## Testing instructions

Tested manually in the developer page - I noticed Instant does not have tests :(

## Types of changes
- adds a new `onSuccess()` callback hook that can be passed when `render()` is called


packages/instant/src/redux/reducer.ts
```diff
@@ -180,13 +182,14 @@ export const createReducer = (initialState: State) => {
                 }
                 return state;
             case ActionTypes.SetBuyOrderStateSuccess:
-                const successTxHash = action.data;
+                const successTxHash = action.data.txHash;
```
fragosti: Was this not correct??

PirosB3: it was, but I refatcored `data` to be an object now

PirosB3:
@fragosti I'm not sure how you would implement this, compared to what I have done. Maybe we can speak about this as soon as the priority becomes current?

PirosB3:
@fragosti this is ready to be reviewed again :)

packages/instant/src/containers/selected_asset_buy_order_state_buttons.ts
```diff
@@ -107,7 +108,26 @@ const mapDispatchToProps = (
     },
 });
 
-export const SelectedAssetBuyOrderStateButtons: React.ComponentClass<SelectedAssetBuyOrderStateButtons> = connect(
+const mergeProps = (
+    connectedState: ConnectedState,
+    connectedDispatch: ConnectedDispatch,
+    ownProps: SelectedAssetBuyOrderStateButtons,
```
fragosti: Is this the correct type for `ownProps`?

PirosB3: So it looks like `SelectedAssetBuyOrderStateButtons` is being used as "props" for both `mapStateToProps` and `mapDispatchToProps`.

`SelectedAssetBuyOrderStateButtons` is defined as
```
export interface SelectedAssetBuyOrderStateButtons {}
```

PirosB3: There also exists a suspicious `BuyOrderStateButtonProps` (without the `Selected` prefix) that contains a lot of properties. This interface is present in `buy_order_state_button.ts` and it's never imported/used within `selected_asset_buy_order_state_buttons.ts` so I assume it's different. Can you confirm?

fragosti: Oh that makes sense. All good.

PR #1804, opened 2019-05-06
fix.docs(README.md): wrong package name

yxliang01:


abandeali1:
Thank you for the fix!

PR #1810, opened 2019-05-13
Fix bug where remainingTakerAssetAmount is not converted to BigNumber

dekz:
## Description

Asset buyer checks for `remainingTakerAssetAmount` MetaData in the APIOrder from the SRA endpoint. This value is not explicitly part of the spec and as such it is not converted from string to BigNumber in `@0x/connect`. Using this value as a string results in errors when calculating the maker asset amount.

## Checklist:

### DIFF
--- a/packages/asset-buyer/src/order_providers/standard_relayer_api_order_provider.ts
+++ b/packages/asset-buyer/src/order_providers/standard_relayer_api_order_provider.ts
@@ -1,6 +1,7 @@
 import { HttpClient } from '@0x/connect';
-import { orderCalculationUtils, orderParsingUtils } from '@0x/order-utils';
+import { orderCalculationUtils } from '@0x/order-utils';
 import { APIOrder, AssetPairsResponse, OrderbookResponse } from '@0x/types';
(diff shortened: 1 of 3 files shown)


packages/asset-buyer/src/order_providers/standard_relayer_api_order_provider.ts
```diff
@@ -25,12 +26,12 @@ export class StandardRelayerAPIOrderProvider implements OrderProvider {
...
             const { order, metaData } = apiOrder;
-            // calculate remainingFillableMakerAssetAmount from api metadata, else assume order is completely fillable
-            const remainingFillableTakerAssetAmount = _.get(
-                metaData,
-                'remainingTakerAssetAmount',
-                order.takerAssetAmount,
-            );
+            // The contents of metaData is not explicity defined in the spec
+            // We check for remainingTakerAssetAmount as a string and use this value if populated
+            const metaDataRemainingTakerAssetAmount = _.get(metaData, 'remainingTakerAssetAmount');
```
steveklebanoff: Minor but would prefer typing this `as string | undefined`

PR #1812, opened 2019-05-13
Update coordinator contracts

xianny:
## Description

Updates Coordinator bytecode so that approval signature verification doesn't fail. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-addresses/CHANGELOG.json
+++ b/packages/contract-addresses/CHANGELOG.json
@@ -3,7 +3,8 @@
         "version": "2.3.2",
         "changes": [
             {
-                "note": "Redeploy Coordinator contract"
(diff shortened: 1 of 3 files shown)


packages/contract-addresses/CHANGELOG.json
```diff
@@ -1,4 +1,12 @@
 [
+    {
+        "version": "2.3.2",
+        "changes": [
+            {
+                "note": "Redeploy Coordinator contract"
+            }
```
fabioberger: PR number?

xianny: adding!

is patch version bump OK or should we use minor?

PR #1816, opened 2019-05-14
Added Alex Kroeger and removed Leo from team page :(

chriskalani:
For Monday morning:
- Added Alex Kroeger
- Removed Leo 😭


chriskalani:
But what if he comes back!?!? Yeah, I can

PR #1817, opened 2019-05-15
fallback to eth_sign if eth_signTypedData fails

xianny:
Lots of wallets/web3providers still don't support the new signing methods.

## Description

Add a new method `ecSignTransactionAsync` that will first attempt `eth_signTypedData ` and fall back to `eth_sign` if it's not available. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/order-utils/src/signature_utils.ts
+++ b/packages/order-utils/src/signature_utils.ts
@@ -20,6 +20,7 @@ import * as _ from 'lodash';
 import { assert } from './assert';
 import { eip712Utils } from './eip712_utils';
 import { orderHashUtils } from './order_hash';
+import { transactionHashUtils } from './transaction_hash';
 import { TypedDataError } from './types';
 import { utils } from './utils';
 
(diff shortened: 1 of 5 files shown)


packages/order-utils/src/signature_utils.ts
```diff
@@ -292,6 +293,46 @@ export const signatureUtils = {
...
+        } catch (err) {
+            // HACK: We are unable to handle specific errors thrown since provider is not an object
+            //       under our control. It could be Metamask Web3, Ethers, or any general RPC provider.
+            //       We check for a user denying the signature request in a way that supports Metamask and
+            //       Coinbase Wallet. Unfortunately for signers with a different error message,
+            //       they will receive two signature requests.
+            if (err.message.includes('User denied message signature')) {
+                throw err;
+            }
+            const orderHash = transactionHashUtils.getTransactionHashHex(transaction);
```
fabioberger:
```suggestion
            const transactionHash = transactionHashUtils.getTransactionHashHex(transaction);
```

PR #1818, opened 2019-05-16
[WIP] Remove Taker validation from order validations

dekz:
## Description

This PR removes the taker side validation when validating an order. This is a non-obvious change so I'll do my best to convince you it is not pragmatic.

Firstly, open orderbook use `address(0)` and this Taker side validation was always skipped.

In the event a `takerAddress` is specified it is only required to have funds before `fillOrder` which can be satisfied by filling an order prior and becoming funded. This case is often validated by the `eth_call` prior to sending the transaction

When using `matchOrders` the `takerAddress` is not required to be funded for the `takerAsset`, only for the fee (if specified). Not all matching relayers use `senderAddress` and it is not required in for example the DutchAuction contract.

These above cases detail that the taker validation is often either skipped, or not useful for matching relayers.

## Checklist:

### DIFF
--- a/packages/order-utils/test/exchange_transfer_simulator_test.ts
+++ b/packages/order-utils/test/exchange_transfer_simulator_test.ts
@@ -9,7 +9,7 @@ import { assetDataUtils } from '../src/asset_data_utils';
 import { constants } from '../src/constants';
 import { ExchangeTransferSimulator } from '../src/exchange_transfer_simulator';
 import { BalanceAndProxyAllowanceLazyStore } from '../src/store/balance_and_proxy_allowance_lazy_store';
-import { TransferType } from '../src/types';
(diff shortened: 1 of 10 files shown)


packages/order-utils/test/exchange_transfer_simulator_test.ts
```diff
@@ -86,18 +86,6 @@ describe('ExchangeTransferSimulator', async () => {
...
         });
-        it("throws if the user doesn't have enough allowance", async () => {
-            return expect(
-                exchangeTransferSimulator.transferFromAsync(
-                    exampleAssetData,
-                    sender,
-                    recipient,
-                    transferAmount,
-                    TradeSide.Taker,
-                    TransferType.Trade,
```
fabioberger: Why was this removed instead of switched to maker like the other tests?

PR #1819, opened 2019-05-16
Arbitrary Fee Tokens

dorothy-zbornak:
## Goodbye, ZRX fees!

wru mcafee

*Who knew so much of our code relied on ZRX fees being baked in?*

This PR implements [ZEIP-28](https://github.com/0xProject/zeips/issues/28), which does away with ZRX fees and instead allows for arbitrary maker and taker fee tokens.

### The Highlights
- The `Order` type now has two new fields: `makerFeeAssetData` and `takerFeeAssetData`.
- Works as expected with both `fillOrder()` and `matchOrders()`.
- Because we swap assets before paying maker/taker fees, both the maker and taker can actually cover their fees with the tokens they just bought in the same transaction!
-  This even works with ERC721, ERC1155 (fungible and non fungible), and MultiAsset fee tokens. Madness!

### Bonus Material
- The order of transfers in `fillOrder()` has been change to taker -> maker first, instead of maker -> taker first. This is necessary for something like [ZEIP-38](https://github.com/0xProject/ZEIPs/issues/38).
- ERC1155 and MultiAsset assets have been well incorporated into the `fillOrder()` and `matchOrders()` tests. 
- Enhanced `base-contract` and `abi-gen-templates` so that any `Error`s thrown by a `callAsync()` and `sendTransactionAsync()` (ganache only) will be automatically decoded into a `RevertError`, if possible. This means you no longer just see an opaque `VM exception: revert` when contract calls unexpectedly explode.
-  Many (but not all) package tools used in testing the V3 contracts have also been updated as collateral damage.
- A couple tests and testing tools have been completely or nearly rewritten (such as [`fill_order_combinatorial_utils`](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange/test/utils/fill_order_combinatorial_utils.ts) and [`match_order_tester`](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange/test/utils/match_order_tester.ts)) because they weren't robust enough to handle all the edge cases for arbitrary fees, as well as the upcoming changes to `matchOrders()` [fee splitting](https://github.com/0xProject/ZEIPs/issues/40).
- Added a [package script](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange-libs/scripts/generate-exchange-selectors.js) to `exchange-libs` to automatically generate [`LibExchangeSelectors.sol`](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange-libs/contracts/src/LibExchangeSelectors.sol) because updating that by hand is a bunch of nopes. Just run `yarn generate-exchange-selectors` from the package directory.
- `Web3Wrapper.toBaseUnitAmount()` can now accept a `number` in addition to a `BigNumber`, which makes a couple tests a lot more compact. 

### Issues/Caveats
- Our `Fill` event grew by two new parameters, which was just enough to break the stack for `emit`, so I had to resort to an [ugly, low-level solution](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange/contracts/src/MixinExchangeCore.sol#L277). 
- `extensions` and `exchange-forwarder` packages have been disabled (hence why code coverage plummeted) from the pipeline. Those packages will probably need major rework beyond the scope of this PR.
- I decoupled the `fillOrder()` combinatorial tests from `OrderStateUtils` and `OrderValidationUtils`. The states that were being tested didn't really warrant them because the combinatorial tests do not generate consecutive fills. We were actually missing out on testing some reverting states because `OrderValidationUtils` was preventing us from using fill amounts that would fail.

## Testing Instructions

You can find the most relevant tests in:
- [`fill_order.ts`](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange/test/fill_order.ts)
- [`match_orders.ts`](../tree/feature/contracts/3.0/arbitrary-fees/contracts/exchange/test/match_orders.ts#L1263)

## Types of changes

Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


.circleci/config.yml
```diff
@@ -84,6 +84,25 @@ jobs:
...
+            - run: yarn wsrun test:circleci @0x/contracts-multisig
+            - run: yarn wsrun test:circleci @0x/contracts-utils
+            - run: yarn wsrun test:circleci @0x/contracts-exchange-libs
+            - run: yarn wsrun test:circleci @0x/contracts-erc20
+            - run: yarn wsrun test:circleci @0x/contracts-erc721
+            - run: yarn wsrun test:circleci @0x/contracts-erc1155
+            # - run: yarn wsrun test:circleci @0x/contracts-extensions
+            - run: yarn wsrun test:circleci @0x/contracts-asset-proxy
+            - run: yarn wsrun test:circleci @0x/contracts-exchange
+            # - run: yarn wsrun test:circleci @0x/contracts-exchange-forwarder
```
fabioberger: Please add `# TODO()` commands next to these explaining why they are commented out (for posterity).

contracts/coordinator/test/mixins.ts
```diff
@@ -68,6 +68,8 @@ describe('Mixins tests', () => {
             takerAssetAmount: constants.ZERO_AMOUNT,
             makerFee: constants.ZERO_AMOUNT,
             takerFee: constants.ZERO_AMOUNT,
+            makerFeeAssetData: constants.NULL_BYTES,
+            takerFeeAssetData: constants.NULL_BYTES,
```
fabioberger: Q: Are there any constraints on what assetData can be used for fees? Could one use `MultiAsset` for example?

contracts/exchange-libs/scripts/generate-exchange-selectors.js
```diff
@@ -0,0 +1,110 @@
+'use strict'
```
fabioberger: Why is this in JS? Scripts can also be written in TS.

contracts/exchange-libs/tslint.json
```diff
@@ -2,5 +2,8 @@
     "extends": ["@0x/tslint-config"],
     "rules": {
         "custom-no-magic-numbers": false
+    },
+    "linterOptions": {
+        "exclude": ["./scripts/**.js"]
```
fabioberger: This is one good reason not to write it in JS. It now has no linter coverage.

packages/base-contract/src/index.ts
```diff
@@ -100,17 +108,32 @@ export class BaseContract {
...
+            // Re-cast StringRevertErrors as plain Errors for backwards-compatibility.
+            if (revertError instanceof StringRevertError) {
+                throw new Error(revertError.values.message as string);
+            }
+        } catch (err) {
+            // Can't decode it.
+            return;
+        }
+        throw revertError;
+    }
```
fabioberger: Woot! Thanks for doing this!

fabioberger: Can you also use it in `awaitTransactionSuccessAsync` within the `tx.handlebars` abi-gen template?

contracts/exchange-libs/scripts/generate-exchange-selectors.js
dorothy-zbornak: My logic was that it's awkward to build a package in order to run a script that generates code for the package that has to be rebuilt again. Thinking about it, maybe a better place for the TS version of this script is `monorepo-scripts`?

contracts/coordinator/test/mixins.ts
dorothy-zbornak: Yup, should work just fine. I don't think I have any explicit tests for MAPs, but I'll look into adding some. MAPs are kind of under-represented across all our tests.

packages/base-contract/src/index.ts
dorothy-zbornak: Yeah, I thought about this, but it would only work against ganache, so it would embed extra logic that's only meaningful in tests. Is there a way to generate/flag "debug" versions of contract wrappers?

contracts/exchange/test/match_orders.ts
```diff
@@ -1270,37 +1228,261 @@ describe('matchOrders', () => {
...
+                    ERC721D: erc721Token.address,
+                };
+            });
+
+            for (const combo of feeAssetCombinations) {
+                const description = combo.description || JSON.stringify(combo);
+                it(description, async () => {
+                    // Create orders to match. For ERC20s, there will be a spread.
+                    const leftMakerAssetAmount = combo.leftMaker.startsWith('ERC20')
+                        ? Web3Wrapper.toBaseUnitAmount(15, 18)
```
hysz: Is there anything special about these values or were they just picked at random?

contracts/exchange/contracts/src/MixinWrapperFunctions.sol
```diff
@@ -74,7 +72,11 @@ contract MixinWrapperFunctions is
         returns (FillResults memory fillResults)
     {
         // ABI encode calldata for `fillOrder`
-        bytes memory fillOrderCalldata = _abiEncodeFillOrder(
+        bytes memory fillOrderCalldata = abi.encodeWithSelector(
+            // solhint-disable-next-line max-line-length
+            // bytes4(keccak256("fillOrder((address,address,address,address,uint256,uint256,uint256,uint256,uint256,uint256,bytes,bytes,bytes,bytes),uint256,bytes)"))
+            // = 0x9b44d556
+            0x9b44d556,
```
abandeali1: Same as above - might want to use the constant from `LibExchangeSelectors` here.

contracts/exchange/test/core.ts
```diff
@@ -402,6 +405,17 @@ describe('Exchange core', () => {
                 (transferLogs[1] as LogWithDecodedArgs<DummyERC20TokenTransferEventArgs>).args._value,
             ).to.be.bignumber.equal(signedOrder.takerFee);
         });
+
+        it('should throw if order is expired', async () => {
```
abandeali1: Any reason this was added? I think it's pretty well tested in `fill_order.ts`. Maybe some of these tests should be restructured at some point so that this is more obvious.

contracts/exchange/test/match_orders.ts
```diff
@@ -1270,37 +1228,261 @@ describe('matchOrders', () => {
...
+            ];
+
+            let erc721TokenIdsByOwner: ERC721TokenIdsByOwner;
+            let nameToERC20Tokens: { [name: string]: string };
+            let nameToERC721Tokens: { [name: string]: string };
+
+            function getAssetData(tokenType: TokenType, ownerAddress: string): string {
+                const encodeERC20AssetData = assetDataUtils.encodeERC20AssetData;
+                const encodeERC721AssetData = assetDataUtils.encodeERC721AssetData;
+                if (tokenType in nameToERC20Tokens) {
```
abandeali1: Nit: the convention we use throughout the rest of the codebase is `if (nameToERC20Tokens[tokenType] !== undefined)`

contracts/exchange/test/match_orders.ts
```diff
@@ -1270,37 +1228,261 @@ describe('matchOrders', () => {
...
+                    const tokenAddress = nameToERC20Tokens[tokenType];
+                    return encodeERC20AssetData(tokenAddress);
+                }
+                if (tokenType in nameToERC721Tokens) {
```
abandeali1: Same as above.

contracts/exchange/test/match_orders.ts
dorothy-zbornak: The values were taken from the existing ERC20 tests. I was wondering this too.

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -524,14 +603,14 @@ contract MixinExchangeCore is
         );
         _dispatchTransferFrom(
             orderHash,
-            zrxAssetData,
+            order.makerFeeAssetData,
```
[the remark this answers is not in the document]
dorothy-zbornak: Makes sense.

contracts/exchange/test/core.ts
dorothy-zbornak: Oh, I disabled the expired orders in the combinatorial tests to make up for fee tokens exponentially introducing more tests. I didn't see a benefit in testing expiration combinations since that logic throws well ahead of any accounting. So I added this test to make up for it.

contracts/exchange-libs/package.json
```diff
@@ -31,10 +31,11 @@
         "coverage:report:lcov": "istanbul report lcov",
         "test:circleci": "yarn test",
         "contracts:gen": "contracts-gen",
-        "lint-contracts": "solhint -c ../.solhint.json contracts/**/**/**/**/*.sol"
+        "lint-contracts": "solhint -c ../.solhint.json contracts/**/**/**/**/*.sol",
+        "generate-exchange-selectors": "node scripts/generate-exchange-selectors.js ../../exchange/generated-artifacts/Exchange.json ./contracts/src/LibExchangeSelectors.sol"
```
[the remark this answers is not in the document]
dorothy-zbornak: Hmm, maybe it's not a good idea because it means you couldn't build `exchange-libs` without first building `exchange`...

dorothy-zbornak:
## Round 2!
For convenience, [here is the diff against the original PR](../compare/d5f2003..84faad3).

### Summary
- `sendTransactionAsync()` and `awaitTransactionSuccessAsync()` methods on contracts now also decode and throw rich reverts. This only works on ganache, though.
- The [script to generate `LibExchangeSelectors.sol`](../blob/feature/contracts/3.0/arbitrary-fees/contracts/exchange-libs/scripts/generate-exchange-selectors.ts) has been converted to typescript. It also now includes event selectors.
- `EIP712_ORDER_SCHEMA_HASH` in `LibOrder` is now public.
- Added ERC1155 and MultiAsset proxy support to the `@0x/order-utils` `ExchangeTransferSimulator` class.
- Added `setProxyAllowanceForAllAsync()` to `ERC1155ProxyWrapper` for testing purposes.
- Added `mintKnownFungibleTokensAsync()`, `isNonFungibleItemAsync()`, `isFungibleItemAsync()`, `getOwnerOfAsync()`, `getBalanceAsync()` to `Erc1155Wrapper`, also for testing purposes.
- Added `MultiAssetProxy`, `ERC1155Fungible`, and `ERC1155NonFungible` assets tests to the `fillOrder()` tests (*).
- Added `MultiAssetProxy` and `ERC1155` assets to the `matchOrders()` tests.
- `fillOrder()` and `matchOrders()` will avoid making a redundant transfer if maker/feeRecipient and takerAsset/makerFeeAsset are the same.

#### * fillOrder() combinatorial tests are disabled by default.
Adding new asset fields and asset types exponentially increased the number of combinatorial tests in `exchange/test/fill_order.ts`. They take around **20 minutes** to run locally (:sob:), so they are skipped by default (and on CI).

You can enable them by setting the environment variable `TEST_ALL=1`. So doing `TEST_ALL=1 yarn test` will run the full suite, which you should do if you are working on anything that is a component of the Exchange contracts.

albrow:
> Adding new asset fields and asset types exponentially increased the number of combinatorial tests in `exchange/test/fill_order.ts`.

I've been toying with the idea of using an embedded EVM implementation (i.e. one written in TypeScript/JavaScript) for combinatorial tests to drastically improve performance. @xianny I know you've been working with `ethereumjs-vm`. Do you think this idea is feasible?

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -282,22 +283,82 @@ contract MixinMatchOrders is
...
+        address rightFeeRecipientAddress = rightOrder.feeRecipientAddress;
+
+        // Settle left order taker asset and maker fees.
+        if (
+            leftOrder.makerAddress == leftFeeRecipientAddress &&
+            leftOrder.makerFeeAssetData.equals(leftOrder.takerAssetData)
+        ) {
+            // Maker is fee recipient and the maker fee asset is the same as the taker asset.
+            // We can transfer the taker asset and maker fees in one go.
+            _dispatchTransferFrom(
```
hysz: If I'm reading this line correctly, the Right Maker is paying their fee to the Left Fee Recipient. But it should go to the Right Fee Recipient.

This is the intended logic:
```
left.makerFeePaid → paid by Left Maker to Left Fee Recipient
left.takerFeePaid → paid by Taker to to Left Fee Recipient

right.makerFeePaid → paid by Right Maker to Right Fee Recipient
right.takerFeePaid → paid by Taker to Right Fee Recipient
```

dorothy-zbornak: You're right, that is exactly what it's doing. I admit I have to work this out in my head every time I see it, but I think this behavior is correct.

We are basically collapsing the two transfers in the following `else` clause because the the `leftMaker == leftFeeRecipient` and the `leftMakerFeeAssetData == leftTakerAssetData`. So `rightMaker` can just pay the combined `leftTakerAsset` and `leftMakerFee` to `leftMaker`/`leftFeeRecipient` since it's all the same asset going to the same place.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -282,22 +283,82 @@ contract MixinMatchOrders is
...
+            rightOrder.makerAddress == rightFeeRecipientAddress &&
+            rightOrder.makerFeeAssetData.equals(rightOrder.takerAssetData)
+        ) {
+            // Maker is fee recipient and the maker fee asset is the same as the taker asset.
+            // We can transfer the taker asset and maker fees in one go.
+            _dispatchTransferFrom(
+                leftOrderHash,
+                leftOrder.makerAssetData,
+                leftOrder.makerAddress,
+                rightOrder.makerAddress,
```
dorothy-zbornak:
```suggestion
                rightFeeRecipientAddress,
```

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -273,20 +276,121 @@ contract MixinExchangeCore is
...
+        // | 352                    | C1      | makerAssetData contents        |
+        // | 352 + C1               | 32      | takerAssetData length          |
+        // | 384 + C1               | C2      | takerAssetData contents        |
+        // | 384 + C1 + C2          | 32      | makerFeeAssetData length       |
+        // | 416 + C1 + C2          | C3      | makerFeeAssetData contents     |
+        // | 416 + C1 + C2          | 32      | takerFeeAssetData length       |
+        // | 448 + C1 + C2 + C3     | C4      | takerFeeAssetData contents     |
+        // |-------------------------------------------------------------------|
+        // | Total Length: 448 + C1 + C2 + C3                                  |
+        bytes memory logData = new bytes(
```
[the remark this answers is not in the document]
dorothy-zbornak: I will check, but we don't really have an alternative aside from pure assembly.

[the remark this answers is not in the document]
dorothy-zbornak: It costs 4.8% more gas to emit the hard way. Let's add it to the backlog if we want to further optimize it out.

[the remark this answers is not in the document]
abandeali1: I think the alternative would be cutting fields from the event... there aren't any obvious choices though.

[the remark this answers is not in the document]
dorothy-zbornak: @hysz Figured out that if you rearranged the `Fill` event args, you can fit an `emit` on the stack!

contracts/exchange/contracts/src/MixinMatchOrders.sol
dorothy-zbornak: Spoke with @hysz and he pointed out there were some other fundamental errors in this settlement optimization (for one, takers were actually paying more than their fair share). This affected both `matchOrders()` and `fillOrders()`. The combinatorial tests also weren't covering these scenarios so they slipped under the radar. Totes my bad!

I just now updated things so the only optimizations we do now are avoiding maker fee transfers if `maker == feeRecipient` and taker fee transfers if `taker == feeRecipient`. There are now also explicit tests to cover all these (and the previous) scenarios.

dorothy-zbornak: Combinatorial tests now take even longer though. :tired_face:
I am definitely adding these to demo sprint.

PR #1821, opened 2019-05-17
Created package for stake-based liquidity incentives

hysz:
## Description

This is the foundation for the staking package. Reviewers, please ensure the package is setup correctly (ex, properly added to all monorepo config files)

## Checklist:


contracts/staking/compiler.json
```diff
@@ -0,0 +1,21 @@
+{
+    "artifactsDir": "generated-artifacts",
+    "contractsDir": "contracts",
+    "useDockerisedSolc": false,
+    "compilerSettings": {
+        "evmVersion": "constantinople",
+        "optimizer": { "enabled": true, "runs": 1000000 },
```
abandeali1: I believe we have an updated version of `compiler.json` with more specific optimizer settings.

contracts/staking/package.json
```diff
@@ -0,0 +1,83 @@
...
+        "make-promises-safe": "^1.1.0",
+        "mocha": "^4.1.0",
+        "npm-run-all": "^4.1.2",
+        "shx": "^0.2.2",
+        "solhint": "^1.4.1",
+        "tslint": "5.11.0",
+        "typescript": "3.0.1"
+    },
+    "dependencies": {
+        "@0x/base-contract": "^5.1.0",
```
abandeali1: Most of these deps should probably be cleared out and added as needed. Otherwise we might end up with some unused dependencies. If you know you'll be using something 100%, feel free to leave it in.

contracts/staking/compiler.json
hysz: Good eye, I've updated the `erc1155` package's compiler.json as well 🙏

PR #1825, opened 2019-05-21
ERC1155 Asset Proxy: Store scaled values at end of ERC1155 calldata

hysz:
## Description

Suppose the `tokensToTransfer` and `valuesToTransfer` are identical; their offsets in the ABI-encoded asset data may be the same. E.g. token IDs [1, 2] and values [1, 2]. Suppose we then scale by a factor of 2, then we expect to trade token IDs [1, 2] and values [2, 4]. This update ensures that scaling the values does not simultaneously scale the token IDs.

In this PR:
1. The ERC1155AssetProxy now stores `scaledValues` at the end of the input ERC1155 calldata.
2. The ERC1155 now has a `createWithType` function to create a token with a specific type identifier.
3. A comprehensive test ensuring the successful transfer of value when token ids and values are abi encoded to same entry in calldata.
4. Added new tests for different lengths of receiver callback data, since we are now appending `scaledValues` after `data`.

## Checklist:

### DIFF
--- a/contracts/asset-proxy/test/erc1155_proxy.ts
+++ b/contracts/asset-proxy/test/erc1155_proxy.ts
@@ -20,6 +20,7 @@ import { AssetProxyId, RevertReason } from '@0x/types';
 import { BigNumber } from '@0x/utils';
 import * as chai from 'chai';
 import { LogWithDecodedArgs } from 'ethereum-types';
+import * as ethUtil from 'ethereumjs-util';
 import * as _ from 'lodash';
 
 import { ERC1155ProxyWrapper, ERC721ProxyContract } from '../src';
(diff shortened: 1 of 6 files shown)


contracts/erc1155/contracts/src/ERC1155Mintable.sol
```diff
@@ -63,6 +63,34 @@ contract ERC1155Mintable is
...
     }
 
+    /// @dev creates a new token
+    /// @param type_ of token
+    /// @param uri URI of token
+    /// @param isNF is non-fungible token
+    function createWithType(
+        uint256 type_,
+        string calldata uri,
+        bool isNF
```
dorothy-zbornak: Don't see this parameter being used. Placeholder?

contracts/asset-proxy/test/erc1155_proxy.ts
```diff
@@ -519,6 +680,114 @@ describe('ERC1155Proxy', () => {
...
+            const spenderInitialBalance = new BigNumber(4);
+            const receiverInitialBalance = new BigNumber(0);
+            const tokenUri = '';
+            const tokenIsNft = false;
+            for (const tokenToCreate of tokensToCreate) {
+                // create token
+                await web3Wrapper.awaitTransactionSuccessAsync(
+                    await erc1155Wrapper
+                        .getContract()
+                        .createWithType.sendTransactionAsync(tokenToCreate, tokenUri, tokenIsNft, {
```
dorothy-zbornak: You should be able to use `createWithType.awaitTransactionSuccessAsync()` here.

contracts/asset-proxy/test/erc1155_proxy.ts
```diff
@@ -519,6 +680,114 @@ describe('ERC1155Proxy', () => {
...
+                            from: owner,
+                        }),
+                    constants.AWAIT_TRANSACTION_MINED_MS,
+                );
+                // mint balance for spender
+                await web3Wrapper.awaitTransactionSuccessAsync(
+                    await erc1155Wrapper
+                        .getContract()
+                        .mintFungible.sendTransactionAsync(tokenToCreate, [spender], [spenderInitialBalance], {
```
dorothy-zbornak: ditto

contracts/erc1155/contracts/src/ERC1155Mintable.sol
hysz: Ah, yeah. Taken from regular `create` but isn't needed. I'll take it out.

hysz:
This PR has been superseded by #1837

PR #1826, opened 2019-05-21
ERC1155 Proxy: Improved Readability

hysz:
## Description

In this PR:
1. Load from calldata into memory only the components of asset data that will remain unchanged
2. Read ERC1155 asset address from calldata, rather than memory.
3. The offset to asset data includes 4 bytes for the selector, simplifying some calculations and making it consistent with other offsets.

## Checklist:


hysz:
This PR has been superseded by #1837

PR #1828, opened 2019-05-21
Move ganache subprovider

dekz:
## Description

This has been bugging me for a while. `GanacheSubprovider` gets brought in via `@0x/subproviders` and comes with some chunky dependencies (ganache-core). It bloats up docker images. Whilst it is possible to tree-shake this out, it also comes with `scrypt` which adds to breakages on different systems (linux and windows).

This PR moves the dependency to lower level (dev-utils). Any other project who might want to use it can bring in `@0x/dev-utils`. 

## Testing instructions
```
yarn add @0x/subproviders
```
### Before
Screen Shot 2019-07-18 at 2 35 30 pm

### After

Screen Shot 2019-07-18 at 2 36 35 pm

## Checklist:

### DIFF
--- a/packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
+++ b/packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
@@ -1,0 +1,23 @@
+import { JSONRPCRequestPayload } from 'ethereum-types';
+
+import { Callback, ErrorCallback } from './types';
+
+import { Subprovider } from './subprovider';
+
+/**
+ * This is duplicated from Subproviders
+ */
+export class EmptyWalletSubprovider extends Subprovider {
+    // tslint:disable-next-line:prefer-function-over-method async-suffix
+    public async handleRequest(payload: JSONRPCRequestPayload, next: Callback, end: ErrorCallback): Promise<void> {
+        switch (payload.method) {
+            case 'eth_accounts':
+                end(null, []);
+                return;
+
+            default:
+                next();
+                return;
+        }
+    }
+}
--- a/packages/dev-utils/src/subproviders/subprovider.ts
+++ b/packages/dev-utils/src/subproviders/subprovider.ts
@@ -1,11 +1,10 @@
 import { promisify } from '@0x/utils';
 import { JSONRPCRequestPayload, JSONRPCResponsePayload } from 'ethereum-types';
-import Web3ProviderEngine = require('web3-provider-engine');
 
-import { Callback, ErrorCallback, JSONRPCRequestPayloadWithMethod } from '../types';
+import { Callback, ErrorCallback, JSONRPCRequestPayloadWithMethod } from './types';
 /**
- * A altered version of the base class Subprovider found in [web3-provider-engine](https://github.com/MetaMask/provider-engine).
- * This one has an async/await `emitPayloadAsync` and also defined types.
+ * Use the subprovider in @0x/subproviders. This is internal for dev-utils only.
+ * Using any below avoiding types due to cyclic dependencies
  */
 export abstract class Subprovider {
     // tslint:disable-next-line:underscore-private-and-protected
@@ -65,7 +64,7 @@ export abstract class Subprovider {
      * directly.
      * @param engine The ProviderEngine this subprovider is added to
      */
-    public setEngine(engine: Web3ProviderEngine): void {
+    public setEngine(engine: any): void {
         this.engine = engine;
     }
 }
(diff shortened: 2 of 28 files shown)


packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,23 @@
+import { JSONRPCRequestPayload } from 'ethereum-types';
+
```
PirosB3: many spaces here? should we compact the imports?

packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Callback, ErrorCallback } from './types';
+
+import { Subprovider } from './subprovider';
+
+/**
+ * This is duplicated from Subproviders
```
PirosB3: why not import it from `subproviders`? that should be fine, right?

packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export class EmptyWalletSubprovider extends Subprovider {
+    // tslint:disable-next-line:prefer-function-over-method async-suffix
+    public async handleRequest(payload: JSONRPCRequestPayload, next: Callback, end: ErrorCallback): Promise<void> {
```
PirosB3: why `async`?

packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
```diff
@@ -0,0 +1,23 @@
...
+        switch (payload.method) {
+            case 'eth_accounts':
+                end(null, []);
```
PirosB3: wouldn't you fallback to any other subprovider that may have implemented `eth_accounts` in this case?

packages/dev-utils/src/subproviders/subprovider.ts
```diff
@@ -0,0 +1,69 @@
+import { promisify } from '@0x/utils';
+import { JSONRPCRequestPayload, JSONRPCResponsePayload } from 'ethereum-types';
+import Web3ProviderEngine = require('web3-provider-engine');
+
+import { Callback, ErrorCallback } from './types';
+/**
+ * Use the subprovider in @0x/subproviders. This is internal for dev-utils only.
+ * Using any below avoiding types due to cyclic dependencies
```
PirosB3: why are there cycling dependencies? doesn't `@0x/dev-utils` import `@0x/subproviders`, but not vice-versa?

packages/dev-utils/src/subproviders/subprovider.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+export abstract class Subprovider {
+    // tslint:disable-next-line:underscore-private-and-protected
+    private engine!: Web3ProviderEngine;
+    protected static _createFinalPayload(payload: Partial<JSONRPCRequestPayload>): Partial<JSONRPCRequestPayload> {
+        const finalPayload = {
+            // defaults
+            id: Subprovider._getRandomId(),
+            jsonrpc: '2.0',
+            params: [],
```
PirosB3: wouldn't you want to use the params that came in from `payload` here?

packages/subproviders/test/utils/ganache_subprovider.ts
```diff
@@ -15,4 +15,4 @@ export const ganacheSubprovider = new GanacheSubprovider({
     port: configs.port,
     networkId: configs.networkId,
     mnemonic: configs.mnemonic,
-});
```
PirosB3: why `as any`?

packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
dekz: cyclic dependency :(

packages/dev-utils/src/subproviders/subprovider.ts
dekz: All of our packages use dev-utils to setup ganache provider for testing.

packages/dev-utils/src/subproviders/empty_wallet_subprovider.ts
dekz: You only want to use this in testing or when a proper Wallet provider is not available, this will prevent it from reaching the RPC node as it is handled sooner.

INFURA for example will throw on a `eth_accounts`, but you might construct EmptyWallet + INFURA RPC if the user has no metamask.

packages/dev-utils/src/subproviders/subprovider.ts
dekz: It gets overwritten on line 19 with the exploded `...payload`, but this syntax has a default of [] if not provided.

packages/subproviders/test/utils/ganache_subprovider.ts
dekz: Typescript isn't determining that these are the same types since Subprovider is redeclared in dev-utils. I'll add a comment about this.

PR #1830, opened 2019-05-23
subproviders: Fix Trezor subprovider inefficiency

fabioberger:
## Description

Cache the derived key info since it does not change and currently we request it from the Trezor device for each request.

Credit goes to @NoahZinsmeister who suggested this optimization: https://github.com/NoahZinsmeister/0x-monorepo/commit/d77ea80509a365acf941136c9ecc92be7bb2c7d1

## Checklist:

### DIFF
--- a/packages/subproviders/src/subproviders/trezor.ts
+++ b/packages/subproviders/src/subproviders/trezor.ts
@@ -29,6 +29,7 @@ export class TrezorSubprovider extends BaseWalletSubprovider {
     private readonly _trezorConnectClientApi: any;
     private readonly _networkId: number;
     private readonly _addressSearchLimit: number;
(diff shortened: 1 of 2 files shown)


packages/subproviders/src/subproviders/trezor.ts
```diff
@@ -29,6 +29,7 @@ export class TrezorSubprovider extends BaseWalletSubprovider {
     private readonly _trezorConnectClientApi: any;
     private readonly _networkId: number;
     private readonly _addressSearchLimit: number;
+    private _initialDerivedKeyInfo: any;
```
dekz:
```suggestion
    private _initialDerivedKeyInfo: DerivedHDKeyInfo | undefined;
```

PR #1832, opened 2019-05-24
Add transaction expirations and execution logs

abandeali1:
## Description

- Adds a `TransactionExecution` event that logs the `transactionHash`
- Adds an `expirationTimeSeconds` field to transactions
- Note that the `coordinator` package still builds and passes tests, but the approvals `expirationTimeSeconds` is now probably redundant. This will be changes in a later PR after the 3.0 contracts are complete.

## Checklist:

### DIFF
--- a/contracts/exchange/test/transactions.ts
+++ b/contracts/exchange/test/transactions.ts
@@ -1057,6 +1057,7 @@ describe('Exchange transactions', () => {
                     await exchangeWrapperContract.cancelOrdersUpTo.awaitTransactionSuccessAsync(
                         targetOrderEpoch,
                         cancelTransaction.salt,
+                        cancelTransaction.expirationTimeSeconds,
                         cancelTransaction.signature,
                         { from: makerAddress },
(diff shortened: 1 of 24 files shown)


contracts/exchange/test/transactions.ts
```diff
@@ -157,6 +159,44 @@ describe('Exchange transactions', () => {
...
+                const transactionExecutionLogs = transactionReceipt.logs.filter(
+                    log =>
+                        (log as LogWithDecodedArgs<ExchangeTransactionExecutionEventArgs>).event ===
+                        'TransactionExecution',
+                );
+                expect(transactionExecutionLogs.length).to.eq(1);
+                const executionLogArgs = (transactionExecutionLogs[0] as LogWithDecodedArgs<
+                    ExchangeTransactionExecutionEventArgs
+                >).args;
+                expect(transactionHashUtils.getTransactionHashHex(transaction)).to.equal(
```
merklejerk: Nit: Assertion order is backwards: should be `actual.equal(expected)`. Also appears in other places in this file.

PR #1834, opened 2019-05-26
Return array of FillResults in batch fill methods

abandeali1:
## Description

- Return an array of `FillResults` for the batch fill variants
- Adds return value tests for all wrapper functions
- Removes `formatters` from the `test-utils` package and no longer passes in any null `assetData` fields into any function calls. Note that we had a task to remove the calldata optimizations in `matchOrders` and the `marketFill` functions. For example, in `matchOrders` we have this rather than checking equality of the `assetData` fields:

```
rightOrder.makerAssetData = leftOrder.takerAssetData;
rightOrder.takerAssetData = leftOrder.makerAssetData;
```

I decided to leave these as is but clarify things a bit more in the comments, since this is actually cheaper than checking equality and still plays nicely with deduped calldata.

## Checklist:


contracts/exchange/contracts/src/interfaces/IWrapperFunctions.sol
```diff
@@ -55,41 +55,41 @@ contract IWrapperFunctions {
...
         LibOrder.Order[] memory orders,
         uint256[] memory takerAssetFillAmounts,
         bytes[] memory signatures
     )
         public
-        returns (LibFillResults.FillResults memory totalFillResults);
+        returns (LibFillResults.FillResults[] memory fillResults);
 
     /// @dev Fills an order with specified parameters and ECDSA signature.
     ///      Returns false if the transaction would otherwise revert.
```
merklejerk: I'm seeing a lot of these `@dev` comments for `*NoThrow` functions saying the function `"Returns false if the transaction would otherwise revert."`, but I don't see anywhere where it ever does (not that it would make sense to anyway).

PR #1835, opened 2019-05-27
Update OrderValidator and create DevUtils contract

abandeali1:
## Description

- Renames `OrderValidator` to `OrderValidationUtils` for clarity
- Adds signature checking to `OrderValidationUtils`
- Refactors `OrderValidationUtils` to use logic from `LibAssetData`
- Cosmetic changes to `LibAssetData`
- Creates a top level `DevUtils` contract that will inherit all `pure` and `view` functions that are not in the core contracts

TODOs for a later PR:
- Many of the `DevUtils` tests should be moved to the `asset-proxy` package
- Add a `remainingFillableAmount` to `getOrderAndTraderInfo`
- Move `DevUtils` out into its own package

## Checklist:


contracts/extensions/contracts/src/DevUtils/OrderValidationUtils.sol
```diff
@@ -0,0 +1,173 @@
...
+        )
+    {
+        orderInfo = EXCHANGE.getOrderInfo(order);
+        isValidSignature = EXCHANGE.isValidSignature(
+            orderInfo.orderHash,
+            order.makerAddress,
+            signature
+        );
+        traderInfo = getTraderInfo(order, takerAddress);
+        return (orderInfo, traderInfo, isValidSignature);
```
fabioberger: @abandeali1 instead of returning `isValidSignature`, can't we set the `orderInfo.OrderStatus` to `SignatureInvalid`? If the signature is invalid, that should take precedence over any other status (e.g., `Fillable` doesn't make sense as an OrderStatus if the signature is invalid).

contracts/extensions/contracts/src/DevUtils/OrderValidationUtils.sol
```diff
@@ -0,0 +1,173 @@
...
+        returns (TraderInfo[] memory)
+    {
+        uint256 ordersLength = orders.length;
+        TraderInfo[] memory tradersInfo = new TraderInfo[](ordersLength);
+        for (uint256 i = 0; i != ordersLength; i++) {
+            tradersInfo[i] = getTraderInfo(orders[i], takerAddresses[i]);
+        }
+        return tradersInfo;
+    }
+}
```
fabioberger: All I want is a 

```
getOrdersInfo(
        LibOrder.Order[] memory orders,
        bytes[] memory signatures,
        address[] memory takerAddresses
)
```

Method. I don't care about `TradersInfo` once I get back the `remainingFillableAmount`. I guess this refactor can happen in the PR where you add `remainingFillableAmount`.

contracts/extensions/contracts/src/DevUtils/OrderValidationUtils.sol
dorothy-zbornak: I think the argument against this would be that `SignatureInvalid` isn't really the state of the order, but the state of the call. The order could be valid in the future if  the correct signature is provided, or may be invalid for other reasons. If you don't think that distinction is likely to be useful, we could collapse these two fields into a single `TradeState` enum.

contracts/asset-proxy/contracts/src/libs/LibAssetData.sol
```diff
@@ -108,64 +115,74 @@ library LibAssetData {
...
+            allowance = IERC20Token(tokenAddress).allowance(owner, spender);
         } else if (proxyId == ERC721_PROXY_ID) {
             (, address tokenAddress, uint256 tokenId) = decodeERC721AssetData(assetData);
             IERC721Token token = IERC721Token(tokenAddress);
-            if (spender == token.getApproved(tokenId) || token.isApprovedForAll(owner, spender)) {
-                return 1;
-            } else {
-                return 0;
-            }
+            allowance = (token.getApproved(tokenId) != address(0) || token.isApprovedForAll(owner, spender)) ? 1 : 0;
```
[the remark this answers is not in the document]
abandeali1: Whoops, not sure how that happened.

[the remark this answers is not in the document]
abandeali1: Side note, I think I'm going to make all unlimited allowances return 2^256 - 1, because this reflects the actual max amounts of units that can be transferred in a single order. `getApproved(tokenId) == spender` would just return an allowance of 1 in this case.

contracts/extensions/contracts/src/DevUtils/OrderValidationUtils.sol
abandeali1: I think these should definitely be distinct. As is, you can pass in a valid order with an invalid signature to check the state of an order if you don't care about validating the signature. Also, an order can still be `Fillable` without a signature if the order has already been partially filled.

PR #1829, opened 2019-05-22
wip: create `callPure` template

xianny:
## Description

Replaces eth_call with pure-evm (javascript EVM implementation) for pure function calls in generated contract-wrappers

## Types of changes

* New feature (non-breaking change which adds functionality)

## Checklist:

-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #1842, opened 2019-05-31
Add workflow for copying compiled artifacts to 0x/contract-artifacts

xianny:
## Description

Currently we have to copy compiled artifacts manually into `@0x/contract-artifacts`. This is tedious and error-prone.

I added a `contracts:postcompile` step to copy the compiled artifacts into the `packages/contract-artifacts/artifacts`. This will only copy files that are exported from the `@0x/contract-artifacts` package.

I also added a linter script to remove the unneeded properties from the compiler output, so that we don't unnecessarily bloat bundle size. 

The workflow is as follows:

* Run `yarn contracts:compile` as usual. This compiles and then copies the files to contract-artifacts. There will be a prompt to lint `contract-artifacts`.
* run `yarn lint` inside `contract-artifacts`, then run `yarn test` to confirm it works.

Question: Should we add the lint step to the compile+postcompile step? 

### Part II
This PR also updates the artifacts to include `deployedBytecode` which is needed for the multilang project. 

## Testing instructions

Follow above workflow.

## Types of changes

## Checklist:


albrow:
Let's make sure that `@0x/contract-artifacts` only contains artifacts that match up with the currently deployed smart contracts. If we change any smart contracts in preparation for v3 we should not update `@0x/contract-artifacts` until those new contracts are fully deployed.

albrow:
This PR does change a lot of the artifacts, but for the most part they are just formatting changes that remove excess whitespace. However, there are changes to the `"bytecode"` and `"deployedBytecode"`. I'm not familiar with exactly what these are used for. Can someone confirm that these changes are okay and aren't going to break anything?

packages/contract-artifacts/artifacts/ZRXToken.json
```diff
@@ -7,221 +7,150 @@
...
-                    }
+                    { "indexed": true, "name": "_owner", "type": "address" },
+                    { "indexed": true, "name": "_spender", "type": "address" },
+                    { "indexed": false, "name": "_value", "type": "uint256" }
                 ],
                 "name": "Approval",
                 "type": "event"
             }
         ],
+        "devdoc": {
```
dekz: Hmm where did these devdocs come from? I don't think we want these in the artifacts

packages/contract-artifacts/artifacts/ZRXToken.json
```diff
@@ -7,221 +7,150 @@
...
+            },
+            "gasEstimates": {
```
dekz: We should get rid of these extra fields. Maybe cleaner to have an allowed list and everything else is banned.

packages/monorepo-scripts/src/copy_contract_artifacts.ts
```diff
@@ -0,0 +1,41 @@
+const pkgNames = process.argv.slice(2);
+import * as artifacts from '@0x/contract-artifacts';
+import * as fs from 'fs';
+import * as path from 'path';
+/*
+ * @param dir        the directory name under 0x-monorepo/contracts, e.g. 'exchange', 'extensions'
+ * @param artifact   the artifact name without extension, e.g. DutchAuction, Exchange
+ */
+function shouldSkip(dir: string, artifact: string): boolean {
+    return (dir === 'extensions' && artifact === 'Exchange') || (dir === 'extensions' && artifact === 'WETH9');
```
dekz: Why do we skip?

packages/sol-compiler/src/compiler.ts
```diff
@@ -55,7 +55,7 @@ const DEFAULT_COMPILER_SETTINGS: solc.CompilerSettings = {
     },
     outputSelection: {
         [ALL_FILES_IDENTIFIER]: {
-            [ALL_CONTRACTS_IDENTIFIER]: ['abi', 'evm.bytecode.object'],
+            [ALL_CONTRACTS_IDENTIFIER]: ['abi', 'evm.bytecode.object', 'evm.deployedBytecode.object'],
```
dekz: Think we need to be careful here. This increases the bundle size of all packages. 

Deployed Bytecode is the bytecode + constructor arguments right? What are the constructor arguments given an artifact and are they applicable over all networks? Can this be created easily from the `evm.bytecode.object` rather than stored (in some cases as a duplicate)?

fabioberger:
I agree with @albrow. Copying of the new artifacts into `@0x/contract-artifacts` should **not** happen everytime the contracts are re-built. There is a deliberate separation between contracts and our tooling so that we only copy the artifacts over before a new protocol upgrade. Up until that point, contract changes **should not** affect our tooling. Because of this, I suggest we simply make the copy helper script a separate command that does **not** run as part of the contract build step. Make sense?

packages/contract-artifacts/src/lint_properties.ts
```diff
@@ -0,0 +1,61 @@
+import * as fs from 'fs';
+import * as path from 'path';
+
+import * as artifacts from './index';
+
+export const RequiredProperties: string[] = [
+    'compilerOutput.evm.bytecode.object',
+    'compilerOutput.evm.deployedBytecode.object',
```
fabioberger: @xianny I thought you said that `deployedBytecode` was not ultimately necessary for EVM execution? Did this change?

packages/monorepo-scripts/src/copy_contract_artifacts.ts
fabioberger: If there is a good reason, please add it to the source code to help the next reader understand.

packages/monorepo-scripts/src/copy_contract_artifacts.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+const artifactsToPublish = Object.keys(artifacts);
+
+const contractsPath = path.join(__dirname, '../../../contracts');
+const contractsDirs = pkgNames.map(pkg => pkg.split('/contracts-')[1]);
```
fabioberger: There is an implicit assumption that all packages under the `contracts` dir will always start with `contracts-`. I think we should make this assumption explicit and throw a helpful error message if we find this not to be the case.

packages/monorepo-scripts/src/copy_contract_artifacts.ts
```diff
@@ -0,0 +1,41 @@
...
+            return acc;
+        }
+        const compiledArtifacts: string[] = fs
+            .readdirSync(artifactsDir)
+            .filter(artifact => {
+                const artifactWithoutExt = artifact.split('.')[0];
+                return artifactsToPublish.includes(artifactWithoutExt) && !shouldSkip(dir, artifactWithoutExt);
+            })
+            .map(artifact => path.join(artifactsDir, artifact));
+        return acc.concat(compiledArtifacts);
```
fabioberger: Let's use `push` here with a deconstructor instead of `concat`.

packages/monorepo-scripts/src/copy_contract_artifacts.ts
```diff
@@ -0,0 +1,41 @@
...
+    },
+    [] as string[],
+);
+
+for (const _path of compiledArtifactPaths) {
+    const fileName = _path.split('/').slice(-1)[0];
+    const targetPath = path.join(__dirname, '../../contract-artifacts/artifacts', fileName);
+    fs.copyFileSync(_path, targetPath);
+    console.log(`Copied from ${_path} to ${targetPath}`); // tslint:disable-line:no-console
```
fabioberger: Use `logUtils` instead of `no-console`

packages/sol-compiler/src/compiler.ts
xianny: Looks like `bytecode` is `deployedBytecode` with the constructor and initialiser code prepended. The prefix bytes are variable. From examining the compiler output, the pattern is:

`0x{...variable length...}6000396000f3{fe/00}{deployedBytecode}`

We could extract with this pattern, but it seems kind of hacky. What is the significance of the `6000396000f3` sequence and can we count on it being the same across contract upgrades?

The alternative is to use the `getContractCode` eth call once and cache the bytecode in the instantiated wrapper. Cleaner but adds an RPC call. If we are hoping to replace utility libs (e.g order-utils/getOrderHash and similar) entirely with generated wrappers, this adds an additional layer of complexity that isn't strictly necessary. 

@fabioberger what do you think of either of these options?

packages/contract-artifacts/src/lint_properties.ts
xianny: Yup, we need `deployedBytecode` for EVM execution.

xianny:
@albrow @fabioberger, gotcha on updating `contract-artifacts` being a separate deliberate step. 

Can we guarantee/assume that any contract development happens in a separate branch and is published immediately after merging? Otherwise, if there are incremental un-deployed contract changes, how do we generate artifacts to use in monorepo development?

packages/contract-artifacts/artifacts/ZRXToken.json
xianny: They were auto-generated. Sounds like I should assume that anything not currently in there isn't wanted?

fabioberger: Yes.

packages/sol-compiler/src/compiler.ts
fabioberger: I'm not sure tbh. @dekz or @albrow  do you have a preference?

My vote would probably be for trying to reliably de-dup `bytecode` and `deployedBytecode`. Is there any kind of fixed delimiter between it and whats prepended?

fabioberger:
Thanks @xianny.

Yeah, that's really annoying. We didn't do a good job of having the protocol team work on V3 in a separate branch from the get-go. Typically that should happen moving forward. Right @abandeali1? 

That being said, updating the artifacts in our tooling should still remain a separate deliberate step.

packages/sol-compiler/src/compiler.ts
dekz: Yeah think my understanding was backwards.

`bytecode` is a subset of `deployedBytecode`. The extra data would be the constructor, plus any internal function calls only called by the constructor.

We should be able to store them in parts, combine for the deployment (since we need constructor execution) and use `deployedByteCode` when running things in our little evm.

Though not sure how this goes with general ecosystem. Haven't checked our format versus the default or truffle format, are already in a different output format?

We know when to split since it is a subset. 

I.e here's WETH `deployedBytecode` and `bytecode`
```
wo.split(wd.substr(2))
[ '0x60c0604052600d60808190527f577261707065642045746865720000000000000000000000000000000000000060a090815261003e91600091906100a3565b506040805180820190915260048082527f57455448000000000000000000000000000000000000000000000000000000006020909201918252610083916001916100a3565b506002805460ff1916601217905534801561009d57600080fd5b5061013e565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106100e457805160ff1916838001178555610111565b82800160010185558215610111579182015b828111156101115782518255916020019190600101906100f6565b5061011d929150610121565b5090565b61013b91905b8082111561011d5760008155600101610127565b90565b61074c8061014d6000396000f300',
  '' ]
```

So the first chunk is the constructor+init functions.

xianny: From examining the compiler output, it looks like `deployedBytecode` is the subset of `bytecode`. I wrote a little script to output where `deployedBytecode` starts for all the artifacts we have right now:

```
Found at 1570 in AssetProxyOwner.json
Found at 1764 in Coordinator.json
Found at 66 in CoordinatorRegistry.json
Found at 992 in DummyERC20Token.json
Found at 924 in DummyERC721Token.json
Found at 296 in DutchAuction.json
Found at 76 in ERC20Proxy.json
Found at 66 in ERC20Token.json
Found at 76 in ERC721Proxy.json
Found at 66 in ERC721Token.json
Found at 1918 in Exchange.json
Found at 3526 in Forwarder.json
Found at 0 in IAssetProxy.json
Found at 0 in IValidator.json
Found at 0 in IWallet.json
Found at 76 in MultiAssetProxy.json
Found at 1046 in OrderValidator.json
Found at 668 in WETH9.json
Found at 150 in ZRXToken.json
```

Here are the `bytecode` prefixes:

```
0x60806040523480156200001157600080fd5b50604051620024473803806200244783398101604090815281516020830151918301516060840151918401805190949390930192909190600090819086908590859083908390869082603282118015906200006c5750818111155b80156200007857508015155b80156200008457508115155b15156200009057600080fd5b600092505b84518310156200016857600260008685815181101515620000b257fe5b6020908102909101810151600160a060020a031682528101919091526040016000205460ff161580156200010857508483815181101515620000f057fe5b90602001906020020151600160a060020a0316600014155b15156200011457600080fd5b60016002600087868151811015156200012957fe5b602090810291909101810151600160a060020a03168252810191909152604001600020805460ff19169115159190911790556001929092019162000095565b84516200017d9060039060208801906200026c565b50505060049190915550506006555060009250505b845182101562000260578482815181101515620001ab57fe5b602090810290910101519050600160a060020a03811615156200022f57604080517f08c379a000000000000000000000000000000000000000000000000000000000815260206004820152601360248201527f494e56414c49445f41535345545f50524f585900000000000000000000000000604482015290519081900360640190fd5b600160a060020a0381166000908152600860205260409020805460ff19166001908117909155919091019062000192565b50505050505062000300565b828054828255906000526020600020908101928215620002c4579160200282015b82811115620002c45782518254600160a060020a031916600160a060020a039091161782556020909201916001909101906200028d565b50620002d2929150620002d6565b5090565b620002fd91905b80821115620002d2578054600160a060020a0319168155600101620002dd565b90565b61213780620003106000396000f300
0x60806040523480156200001157600080fd5b506040516200237438038062002374833981016040819052620000349162000253565b600080546001600160a01b0319166001600160a01b0383161790556040516200006090602001620002a0565b60408051601f1981840301815282825280516020918201208383018352601784527f30782050726f746f636f6c20436f6f7264696e61746f720000000000000000009382019390935281518083018352600581527f312e302e300000000000000000000000000000000000000000000000000000009082015290516200012e92917f626d101e477fd17dd52afb3f9ad9eb016bf60f6e377877f34e8f3ea84c930236917f06c015bd22b4c69690933c1058878ebdfef31f9aaae40bbe86d8a09fe1b2972c9130910162000285565b60408051601f19818403018152908290528051602091820120600155620001569101620002a0565b60408051601f1981840301815282825280516020918201208383018352600b84527f30782050726f746f636f6c0000000000000000000000000000000000000000009382019390935281518083018352600181527f32000000000000000000000000000000000000000000000000000000000000009082015260005491516200023193927ff0f24618f4c4be1e62e026fb039a20ef96f4495294817d1027ffaa6d1f70e61e927fad7c5bef027816a800da1736444fb58a807ef4c9603b7848673f7e3a68eb14a5926001600160a01b03909216910162000285565b60408051601f1981840301815291905280516020909101206002555062000361565b6000602082840312156200026657600080fd5b81516001600160a01b03811681146200027e57600080fd5b9392505050565b93845260208401929092526040830152606082015260800190565b7f454950373132446f6d61696e280000000000000000000000000000000000000081527f737472696e67206e616d652c0000000000000000000000000000000000000000600d8201527f737472696e672076657273696f6e2c000000000000000000000000000000000060198201527f6164647265737320766572696679696e67436f6e74726163740000000000000060288201527f2900000000000000000000000000000000000000000000000000000000000000604182015260420190565b61200380620003716000396000f3fe
0x608060405234801561001057600080fd5b506103e9806100206000396000f3fe
0x60806040523480156200001157600080fd5b506040516200105a3803806200105a833981810160405260808110156200003757600080fd5b8101908080516401000000008111156200005057600080fd5b820160208101848111156200006457600080fd5b81516401000000008111828201871017156200007f57600080fd5b505092919060200180516401000000008111156200009c57600080fd5b82016020810184811115620000b057600080fd5b8151640100000000811182820187101715620000cb57600080fd5b5050602080830151604090930151600080546001600160a01b03191633179055865192955092935062000104916004918701906200013a565b5082516200011a9060059060208601906200013a565b506006919091553360009081526001602052604090205550620001df9050565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106200017d57805160ff1916838001178555620001ad565b82800160010185558215620001ad579182015b82811115620001ad57825182559160200191906001019062000190565b50620001bb929150620001bf565b5090565b620001dc91905b80821115620001bb5760008155600101620001c6565b90565b610e6b80620001ef6000396000f3fe
0x60806040523480156200001157600080fd5b506040516200188138038062001881833981810160405260408110156200003757600080fd5b8101908080516401000000008111156200005057600080fd5b820160208101848111156200006457600080fd5b81516401000000008111828201871017156200007f57600080fd5b505092919060200180516401000000008111156200009c57600080fd5b82016020810184811115620000b057600080fd5b8151640100000000811182820187101715620000cb57600080fd5b5050600080546001600160a01b031916331790558451909350620000f9925060059150602085019062000118565b5080516200010f90600690602084019062000118565b505050620001bd565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106200015b57805160ff19168380011785556200018b565b828001600101855582156200018b579182015b828111156200018b5782518255916020019190600101906200016e565b50620001999291506200019d565b5090565b620001ba91905b80821115620001995760008155600101620001a4565b90565b6116b480620001cd6000396000f3fe
0x608060405234801561001057600080fd5b506040516110b83803806110b883398101604081905261002f91610054565b600080546001600160a01b0319166001600160a01b0392909216919091179055610084565b60006020828403121561006657600080fd5b81516001600160a01b038116811461007d57600080fd5b9392505050565b611025806100936000396000f3fe
0x6080604052600080546001600160a01b03191633179055610f42806100256000396000f3fe
0x608060405234801561001057600080fd5b506106bf806100206000396000f3fe
0x6080604052600080546001600160a01b03191633179055610fb8806100256000396000f3fe
0x608060405234801561001057600080fd5b50610ec2806100206000396000f3fe
0x60806040526000805460ff191690553480156200001b57600080fd5b5060405162005560380380620055608339810160408190526200003e916200020f565b80518190620000559060019060208401906200016a565b50506040516020016200006890620002ed565b60408051601f1981840301815282825280516020918201208383018352600b84527f30782050726f746f636f6c0000000000000000000000000000000000000000009382019390935281518083018352600181527f32000000000000000000000000000000000000000000000000000000000000009082015290516200013692917ff0f24618f4c4be1e62e026fb039a20ef96f4495294817d1027ffaa6d1f70e61e917fad7c5bef027816a800da1736444fb58a807ef4c9603b7848673f7e3a68eb14a591309101620002d2565b60408051601f19818403018152919052805160209091012060025550600380546001600160a01b03191633179055620003ae565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f10620001ad57805160ff1916838001178555620001dd565b82800160010185558215620001dd579182015b82811115620001dd578251825591602001919060010190620001c0565b50620001eb929150620001ef565b5090565b6200020c91905b80821115620001eb5760008155600101620001f6565b90565b6000602080838503121562000222578182fd5b82516001600160401b038082111562000239578384fd5b81850186601f8201126200024b578485fd5b80519250818311156200025c578485fd5b604051601f8401601f19168101850183811182821017156200027c578687fd5b604052838152818401850188101562000293578586fd5b8592505b83831015620002b6578183018501518184018601529184019162000297565b83831115620002c757858585830101525b979650505050505050565b93845260208401929092526040830152606082015260800190565b7f454950373132446f6d61696e280000000000000000000000000000000000000081527f737472696e67206e616d652c0000000000000000000000000000000000000000600d8201527f737472696e672076657273696f6e2c000000000000000000000000000000000060198201527f6164647265737320766572696679696e67436f6e74726163740000000000000060288201527f2900000000000000000000000000000000000000000000000000000000000000604182015260420190565b6151a280620003be6000396000f3fe
0x60806040523480156200001157600080fd5b50604051620029cd380380620029cd833981016040819052620000349162000524565b60008054336001600160a01b031991821617909155600180549091166001600160a01b038516179055815183908390839062000078906004906020850190620003a8565b5080516200008e906005906020840190620003a8565b506000620000ac6010836200035360201b620017531790919060201c565b90506000620000cb6010856200035360201b620017531790919060201c565b600280546001600160a01b03199081166001600160a01b039586161790915560038054909116918416919091179055506001546040516000955091169250636070410891506200011b90620005cf565b6040519081900381207fffffffff0000000000000000000000000000000000000000000000000000000060e084901b1682526200015b9160040162000611565b60206040518083038186803b1580156200017457600080fd5b505afa15801562000189573d6000803e3d6000fd5b505050506040513d601f19601f82011682018060405250620001af9190810190620004f2565b90506001600160a01b038116620001fd576040517f08c379a0000000000000000000000000000000000000000000000000000000008152600401620001f4906200069b565b60405180910390fd5b6002546040517f095ea7b30000000000000000000000000000000000000000000000000000000081526001600160a01b039091169063095ea7b3906200024c90849060001990600401620005f8565b602060405180830381600087803b1580156200026757600080fd5b505af11580156200027c573d6000803e3d6000fd5b505050506040513d601f19601f82011682018060405250620002a29190810190620005ab565b506003546040517f095ea7b30000000000000000000000000000000000000000000000000000000081526001600160a01b039091169063095ea7b390620002f290849060001990600401620005f8565b602060405180830381600087803b1580156200030d57600080fd5b505af115801562000322573d6000803e3d6000fd5b505050506040513d601f19601f82011682018060405250620003489190810190620005ab565b5050505050620006d2565b6000816014018351101562000396576040517f08c379a0000000000000000000000000000000000000000000000000000000008152600401620001f4906200063e565b5001601401516001600160a01b031690565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f10620003eb57805160ff19168380011785556200041b565b828001600101855582156200041b579182015b828111156200041b578251825591602001919060010190620003fe565b50620004299291506200042d565b5090565b6200044a91905b8082111562000429576000815560010162000434565b90565b600082601f8301126200045e578081fd5b81516001600160401b038082111562000475578283fd5b602060405181601f19601f860116820101818110848211171562000497578586fd5b6040528381529350838584018201871015620004b257600080fd5b600092505b83831015620004d65785830182015181840183015291810191620004b7565b83831115620004e85760008285830101525b5050505092915050565b6000602082840312156200050557600080fd5b81516001600160a01b03811681146200051d57600080fd5b9392505050565b60008060006060848603121562000539578182fd5b83516001600160a01b038116811462000550578283fd5b60208501519093506001600160401b03808211156200056d578384fd5b6200057b878388016200044d565b935060408601519150808211156200059257600080fd5b50620005a1868287016200044d565b9150509250925092565b600060208284031215620005be57600080fd5b815180151581146200051d57600080fd5b7f4552433230546f6b656e28616464726573732900000000000000000000000000815260130190565b6001600160a01b03929092168252602082015260400190565b7fffffffff0000000000000000000000000000000000000000000000000000000091909116815260200190565b60208082526026908201527f475245415445525f4f525f455155414c5f544f5f32305f4c454e4754485f524560408201527f5155495245440000000000000000000000000000000000000000000000000000606082015260800190565b60208082526018908201527f554e524547495354455245445f41535345545f50524f58590000000000000000604082015260600190565b6122eb80620006e26000396000f3fe



0x6080604052600080546001600160a01b03191633179055611510806100256000396000f3fe
0x60806040523480156200001157600080fd5b506040516200199338038062001993833981016040819052620000349162000112565b600080546001600160a01b0319166001600160a01b0384161790558051620000649060019060208401906200006d565b505050620001fa565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f10620000b057805160ff1916838001178555620000e0565b82800160010185558215620000e0579182015b82811115620000e0578251825591602001919060010190620000c3565b50620000ee929150620000f2565b5090565b6200010f91905b80821115620000ee5760008155600101620000f9565b90565b600080604083850312156200012657600080fd5b82516001600160a01b03811681146200013e57600080fd5b602084810151919350906001600160401b03808211156200015d578384fd5b81860187601f8201126200016f578485fd5b805192508183111562000180578485fd5b604051601f8401601f1916810185018381118282101715620001a0578687fd5b6040528381528184018501891015620001b7578586fd5b8592505b83831015620001da5781830185015181840186015291840191620001bb565b83831115620001eb57858585830101525b80955050505050509250929050565b611789806200020a6000396000f3fe
0x60c0604052600d60808190527f577261707065642045746865720000000000000000000000000000000000000060a090815261003e91600091906100a3565b506040805180820190915260048082527f57455448000000000000000000000000000000000000000000000000000000006020909201918252610083916001916100a3565b506002805460ff1916601217905534801561009d57600080fd5b5061013e565b828054600181600116156101000203166002900490600052602060002090601f016020900481019282601f106100e457805160ff1916838001178555610111565b82800160010185558215610111579182015b828111156101115782518255916020019190600101906100f6565b5061011d929150610121565b5090565b61013b91905b8082111561011d5760008155600101610127565b90565b61074c8061014d6000396000f300
0x60606040526b033b2e3c9fd0803ce8000000600355341561001c57fe5b5b600354600160a060020a0333166000908152602081905260409020555b5b61078d8061004a6000396000f300
```

They start and end in the same way but apart from that I don't see any similarities. Does anyone understand what the significance of `6000396000f3` is and if we can depend on it as a fixed delimiter? If we can, that would be the easiest solution. 

Alternative #1: we could omit the `bytecode` field and output the artifact with a custom field `bytecodePrefix` that can be used for cases where we do need the `bytecode` and not `deployedBytecode`:

```
...
evm : {
    deployedBytecode: {
        object: '....';
    },
    bytecodePrefix: '...',
}
```

Or if we want to preserve backwards compatibility:
```
...
evm : {
    bytecode: {
        object: '....';
    },
    bytecodeOnlyPrefix: '...',
}
```

It would be pretty simple to do this step as part of the post-generation linting where we remove unwanted fields. Downside is that we would have to require this custom field to initialize wrappers, so external devs couldn't use `abi-gen` to create wrappers for non-0x contracts. I'm hesitant to deviate from the standard `ethereum-types` depending on how predominantly they are used.

Alternative #2: If we are absolutely against keeping both fields, and don't want to create a dependency on a non-standard type, I'm leaning towards using `eth_getCode` once to get the deployed bytecode.

fabioberger: Honestly don't want us to modify anything under the `evm` key since this is part of the Solidity compiler standard output BUT we could just have `deployedBytecode` there and move the `bytecodePrefix` key outside of the `evm` key. 

But have we found a reliable way of teasing out the prefix from the rest? One crazy idea is having the compiler return both the `bytecode` & `deployedBytecode` and in the "post generation linting" step use a [largest common substring](http://cache.mifio.com/javascript002.html) function to tease out the chunk they have in common, and then store the start and end indexes that delimit this shared chunk (that also happens to correspond to the `deployedBytecode`). That way, we only store `bytecode` but can easily pull out `deployedBytecode` from it. Thoughts?

xianny: @fabioberger I think that's a good idea! Only question is where to store it. We already have some `SimpleContractArtifact` types. It looks like they were created just to eliminate `sourceMap` field from the bytecode output. What do you think of nesting it under there? 

We could even still accept `ContractArtifact | SimpleContractArtifact` for deploys, and fallback to an eth_call to get deployed bytecode if it isn't supplied in the standard artifact. `deployedBytecode` is a field in the `ContractArtifact` type but it looks like we don't assert types here. 

```
 public static async deployFrom0xArtifactAsync(
        artifact: ContractArtifact | SimpleContractArtifact,
...
```

sol compiler output types:
```
export interface ContractArtifact extends ContractVersionData {
    schemaVersion: string;
    contractName: string;
    networks: ContractNetworks;
}

export interface ContractVersionData {
    compiler: CompilerOpts;
    sources: {
        [sourceName: string]: {
            id: number;
        };
    };
    sourceCodes: {
        [sourceName: string]: string;
    };
    sourceTreeHashHex: string;
    compilerOutput: StandardContractOutput;
}

export interface StandardContractOutput {
    abi: ContractAbi;
    evm: EvmOutput;
    devdoc?: DevdocOutput;
}

export interface EvmOutput {
    bytecode: EvmBytecodeOutput;
    deployedBytecode: EvmBytecodeOutput;
}

export interface EvmBytecodeOutput {
    object: string;
    sourceMap: string;
}

```


0x simple types:
```
export interface SimpleContractArtifact {
    schemaVersion: string;
    contractName: string;
    compilerOutput: SimpleStandardContractOutput;
    networks: ContractNetworks;
}

export interface SimpleStandardContractOutput {
    abi: ContractAbi;
    evm: SimpleEvmOutput;
    devdoc?: DevdocOutput;
}

export interface SimpleEvmOutput {
    bytecode: SimpleEvmBytecodeOutput;
    deployedBytecode: SimpleEvmDeployedBytecodeIndices; // add
}

export interface SimpleEvmBytecodeOutput {
    object: string;
}

/* add */
export interface SimpleEvmDeployedBytecodeIndices {
    start: number;
    end: number;
}
```

xianny: discussed on Slack, we will add a `metadata` field to `SimpleContractArtifact`. Will do that in https://github.com/0xProject/0x-monorepo/pull/1829

xianny:
@fabioberger @dekz ready for re-review. Much simpler changes now.

packages/contract-artifacts/package.json
```diff
@@ -10,13 +10,19 @@
...
+        "artifacts_update": "yarn artifacts_copy && yarn artifacts_transform",
         "build": "yarn tsc -b",
         "build:ci": "yarn build",
         "test": "yarn run_mocha",
         "test:circleci": "yarn test",
         "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --bail --exit",
         "clean": "shx rm -rf lib"
     },
+    "config": {
+        "contractsPackages": "@0x/contracts-asset-proxy @0x/contracts-erc20 @0x/contracts-erc721 @0x/contracts-erc1155 @0x/contracts-exchange @0x/contracts-exchange-forwarder @0x/contracts-exchange-libs @0x/contracts-extensions @0x/contracts-multisig @0x/contracts-test-utils @0x/contracts-utils @0x/contracts-coordinator"
```
fabioberger: In the script, let's actually read it from the root `package.json` instead of duplicating this here.

packages/contract-artifacts/src/properties.ts
```diff
@@ -0,0 +1,22 @@
+export const RequiredProperties: string[] = [
+    'schemaVersion',
+    'contractName',
+    'compilerOutput.evm.bytecode.object',
+    'compilerOutput.abi',
+];
+
+export const ForbiddenProperties: string[] = [
```
fabioberger: Our convention is all uppercase, snake-case for constants.

packages/contract-artifacts/test/contract_artifacts_test.ts
```diff
@@ -29,4 +24,19 @@ describe('Contract Artifacts', () => {
         }
         expect(forbiddenPropertiesByArtifact).to.eql({});
     });
+    it('should include all required attributes', () => {
+        const missingRequiredPropertiesByArtifact: { [name: string]: string[] } = {};
```
fabioberger: Use the type: `ObjectMap<string>` here.

packages/contract-artifacts/test/contract_artifacts_test.ts
```diff
@@ -29,4 +24,19 @@ describe('Contract Artifacts', () => {
...
+        for (const [artifactName, artifact] of Object.entries(artifacts)) {
+            for (const requiredProperty of RequiredProperties) {
+                const requiredValue = get(artifact, requiredProperty);
```
fabioberger: `get` is not type-safe I believe. Can we avoid using it?

packages/contract-artifacts/package.json
```diff
@@ -10,13 +10,19 @@
         "test": "test"
     },
     "scripts": {
+        "artifacts_copy": "node ../monorepo-scripts/lib/contract_artifacts_copy.js ${npm_package_config_contractsPackages}",
```
fabioberger: Let's move this into the `contract-artifacts` package.

packages/abi-gen-wrappers/package.json
```diff
@@ -13,10 +13,11 @@
         "build": "yarn pre_build && tsc -b",
         "build:ci": "yarn build",
         "lint": "tslint --format stylish --project .",
+        "artifacts_transform": "node ../monorepo-scripts/lib/contract_artifacts_transform.js ../contract-artifacts/artifacts && prettier --write ../contract-artifacts/artifacts/*.json",
```
fabioberger: Let's remove this from here and also move this script into the `contract-artifacts` package.

packages/contract-artifacts/src/properties.ts
```diff
@@ -0,0 +1,22 @@
+export const RequiredProperties: string[] = [
```
fabioberger: Let's stop exporting this.

packages/monorepo-scripts/src/contract_artifacts_copy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import * as fs from 'fs';
+import * as path from 'path';
+
+import { utils } from './utils/utils';
+
+const pkgNames = process.argv.slice(2);
+const contractsDirs = [];
+for (const pkgName of pkgNames) {
+    if (pkgName.indexOf('/contracts-') === -1) {
```
fabioberger: Can we use `includes`? Or `startsWith` here?

packages/monorepo-scripts/src/contract_artifacts_copy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+for (const _path of allArtifactPaths) {
+    const fileName = _path.split('/').slice(-1)[0];
+    const targetPath = path.join(__dirname, '../../contract-artifacts/artifacts', fileName);
+    const targetPathPython = path.join(__dirname, '../../../python-packages/contract_artifacts/src/zero_ex/contract_artifacts/artifacts', fileName);
+    fs.copyFileSync(_path, targetPath);
+    fs.copyFileSync(_path, targetPathPython);
+    utils.log(`Copied from ${_path} to ${targetPath}`);
+}
+utils.log(`Finished copying contract-artifacts. Remember to lint artifacts before using abi-gen.`);
```
fabioberger:
```suggestion
utils.log(`Finished copying contract-artifacts.`);
```

The transform happens automatically now right?

packages/monorepo-scripts/src/contract_artifacts_copy.ts
```diff
@@ -0,0 +1,54 @@
...
+
+/*
+ * @param dir        the directory name under 0x-monorepo/contracts, e.g. 'exchange', 'extensions'
+ * @param artifact   the artifact name without extension, e.g. DutchAuction, Exchange
+ */
+function shouldSkip(dir: string, artifact: string): boolean {
+    return (
+        (dir === 'extensions' && artifact === 'Exchange') || // duplicate artifact
+        (dir === 'extensions' && artifact === 'WETH9') // duplicate artifact
+    );
```
fabioberger: Either remove these and let the artifact get over-written, or leave a comment as to _why_ we skip these.

packages/monorepo-scripts/src/contract_artifacts_transform.ts
```diff
@@ -0,0 +1,72 @@
+import { logUtils } from '@0x/utils';
+import * as fs from 'fs';
+
+// xianny: hack, copied from contract-artifacts to avoid changing contract-artifacts exports
```
fabioberger:
```suggestion
// HACK(xianny): copied from contract-artifacts to avoid changing contract-artifacts exports
```

fabioberger: This can be removed once this file is in `contract-artifacts` 🍾

packages/monorepo-scripts/src/contract_artifacts_transform.ts
```diff
@@ -0,0 +1,72 @@
...
+    for (const filePath of filePaths) {
+        const artifact = JSON.parse(fs.readFileSync(filePath).toString()) as { [name: string]: any };
+        for (const property of ForbiddenProperties) {
+            deleteNestedProperty(artifact, property);
+        }
+        fs.writeFileSync(filePath.replace(inputDir, outputDir), JSON.stringify(artifact));
+    }
+}
+
+function deleteNestedProperty(obj: any, propPath: string): void {
```
fabioberger: Can we move this to a `@0x/utils`? Feel like this would be useful elsewhere.

packages/monorepo-scripts/src/contract_artifacts_copy.ts
xianny: Not necessarily. The yarn shortcut runs `copy && transform`, but this script only does `copy`, so I think we should keep this log line in case the script is used alone.

The resulting output looks like this:
```
[17:24:58]contract-artifacts (feature/prepublish-contract-artifacts)$ yarn artifacts_update
yarn run v1.9.4
$ yarn artifacts_copy && yarn artifacts_transform
$ node lib/src/copy.js ${npm_package_config_contractsPackages}
Finished copying contract-artifacts. Remember to transform artifacts before publishing or using abi-gen
$ node lib/src/transform.js ./artifacts && prettier --write ./artifacts/*.json &&  cp -r ./artifacts/ ../../python-packages/contract_artifacts/src/zero_ex/contract_artifacts/artifacts/
Deleting forbidden properties from artifacts in ./artifacts. Output to ./artifacts
artifacts/AssetProxyOwner.json 117ms
artifacts/Coordinator.json 37ms
artifacts/CoordinatorRegistry.json 22ms
artifacts/DummyERC20Token.json 27ms
artifacts/DummyERC721Token.json 37ms
artifacts/DutchAuction.json 22ms
artifacts/ERC20Proxy.json 27ms
artifacts/ERC20Token.json 25ms
artifacts/ERC721Proxy.json 23ms
artifacts/ERC721Token.json 19ms
artifacts/Exchange.json 74ms
artifacts/Forwarder.json 20ms
artifacts/IAssetProxy.json 9ms
artifacts/IValidator.json 8ms
artifacts/IWallet.json 7ms
artifacts/MultiAssetProxy.json 12ms
artifacts/OrderValidator.json 21ms
artifacts/WETH9.json 14ms
artifacts/ZRXToken.json 13ms
✨  Done in 2.54s.
[17:25:12]contract-artifacts (feature/prepublish-contract-artifacts)$ 

```

packages/monorepo-scripts/src/contract_artifacts_copy.ts
xianny: removing.

packages/contract-artifacts/package.json
xianny: To get this to build, I had to change `rootDir: ./` to `rootDir: ../..` in the `tsconfig.json`. Waiting for CI to see if it causes any issues.

xianny: This wound up breaking relative links by introducing nesting to the build output. I'm not sure it's worth it to introduce this weirdness here. 

Build output:
```
lib
├── package.json
└── packages
    └── contract-artifacts
        ├── artifacts
        │   ├── AssetProxyOwner.json
        │   ├── Coordinator.json
        │   ├── CoordinatorRegistry.json
        │   ├── DummyERC20Token.json
        │   ├── DummyERC721Token.json
        │   ├── DutchAuction.json
        │   ├── ERC20Proxy.json
        │   ├── ERC20Token.json
        │   ├── ERC721Proxy.json
        │   ├── ERC721Token.json
        │   ├── Exchange.json
        │   ├── Forwarder.json
        │   ├── IAssetProxy.json
        │   ├── IValidator.json
        │   ├── IWallet.json
        │   ├── MultiAssetProxy.json
        │   ├── OrderValidator.json
        │   ├── WETH9.json
        │   └── ZRXToken.json
        ├── src
        │   ├── copy.d.ts
        │   ├── copy.d.ts.map
        │   ├── copy.js
        │   ├── copy.js.map
        │   ├── index.d.ts
        │   ├── index.d.ts.map
        │   ├── index.js
        │   ├── index.js.map
        │   ├── transform.d.ts
        │   ├── transform.d.ts.map
        │   ├── transform.js
        │   └── transform.js.map
        └── test
            ├── contract_artifacts_test.d.ts
            ├── contract_artifacts_test.d.ts.map
            ├── contract_artifacts_test.js
            └── contract_artifacts_test.js.map
```

fabioberger: Let's use `fs.readFile`

packages/monorepo-scripts/src/contract_artifacts_copy.ts
fabioberger: Ok

packages/contract-artifacts/src/transform.ts
```diff
@@ -0,0 +1,50 @@
+import { deleteNestedProperty, logUtils } from '@0x/utils';
+import * as fs from 'fs';
+
+export const REQUIRE_PROPERTIES: string[] = [
```
fabioberger:
```suggestion
export const REQUIRED_PROPERTIES: string[] = [
```

PR #1848, opened 2019-06-04
Refactor OrderValidationUtils and create contracts-dev-utils package

abandeali1:
## Description

- Refactors `OrderValidationUtils` completely. The primary function is `getOrderRelevantState`, which returns the `fillableTakerAssetAmount` rather than a `TraderInfo` instance.
- Moves `DevUtils` into its own package.
- Moves `LibTransactionDecoder` into the `contracts/dev-utils` package.
- Refactors tests for the above 3 contracts.

## Checklist:


contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+        // Get the transferable amount of the `makerAsset`
+        uint256 transferableMakerAssetAmount = getTransferableAssetAmount(order.makerAssetData, makerAddress);
+
+        // Assign to stack variables to reduce redundant mloads
+        uint256 takerAssetAmount = order.takerAssetAmount;
+        uint256 makerFee = order.makerFee;
+    
+        // Get the amount of `takerAsset` that is purchasable given the transferability of `makerAsset` and `makerFeeAsset`
+        uint256 purchasableTakerAssetAmount;
+        if (order.makerAssetData.equals(ZRX_ASSET_DATA)) {
```
fabioberger: Why are we still special-casing ZRX here? I thought the fee can be paid in any token now?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+            // If `makerAsset` equals `makerFeeAsset`, the % that can be filled is
+            // transferableMakerAssetAmount / (makerAssetAmount + makerFee)
+            purchasableTakerAssetAmount = getPartialAmountFloor(
+                transferableMakerAssetAmount,
+                safeAdd(order.makerAssetAmount, makerFee),
+                takerAssetAmount
+            );
+        } else {
+            // Get the transferable amount of the `makerFeeAsset`
+            uint256 transferableMakerFeeAssetAmount = getTransferableAssetAmount(ZRX_ASSET_DATA, makerAddress);
```
fabioberger: Ditto here.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+        // `fillableTakerAssetAmount` is the lower of the order's remaining `takerAssetAmount` and the `purchasableTakerAssetAmount`
+        fillableTakerAssetAmount = min256(
+            safeSub(takerAssetAmount, orderInfo.orderTakerAssetFilledAmount),
+            purchasableTakerAssetAmount
+        );
+
+        return (orderInfo, fillableTakerAssetAmount, isValidSignature);
+    }
+
+    /// @dev Fetches information for all passed in orders and the makers/takers of each order.
```
fabioberger:
```suggestion
 /// @dev Fetches all order-relevant information needed to validate if the supplied orders are fillable.
```

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+    /// @param orders Array of order specifications.
+    /// @param signatures Proofs that orders have been created by makers.
```
fabioberger: This sounds a bit awkward. Perhaps just `Order signatures`? Or `Signatures corresponding to the supplied orders`?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+            (ordersInfo[i], fillableTakerAssetAmounts[i], isValidSignature[i]) = getOrderRelevantState(
+                orders[i],
+                signatures[i]
+            );
+        }
+
+        return (ordersInfo, fillableTakerAssetAmounts, isValidSignature);
+    }
+
+    /// @dev Gets the address of the AssetProxy that corresponds to the given assetData.
```
fabioberger:
```suggestion
    /// @dev Gets the AssetProxy contract address that corresponds to the supplied assetData.
```

contracts/dev-utils/test/order_validation_utils.ts
```diff
@@ -0,0 +1,393 @@
...
+        });
+        it('should return isValidSignature=false when the signature is invalid', async () => {
+            const invalidSignature = '0x01';
+            const [, , isValidSignature] = await devUtils.getOrderRelevantState.callAsync(
+                signedOrder,
+                invalidSignature,
+            );
+            expect(isValidSignature).to.equal(false);
+        });
+        it('should return a fillableTakerAssetAmount of 0 when no balances or allowances are insufficient', async () => {
```
fabioberger:
```suggestion
        it('should return a fillableTakerAssetAmount of 0 when no balances or allowances are sufficient', async () => {
```

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+    {
+        EXCHANGE = IExchange(_exchange);
+        ZRX_ASSET_DATA = _zrxAssetData;
+        ERC20_PROXY_ADDRESS = EXCHANGE.getAssetProxy(ERC20_PROXY_ID);
+    }
+
+    /// @dev Fetches information for order and maker/taker of order.
+    /// @param order The order structure.
+    /// @param signature Proof that order has been created by maker.
+    /// @return OrderInfo, the remaining amount fillable by the taker, and validity of signature for given order.
```
fabioberger: A general thought about Solidity function comments. These comments are going to become **the** canonical docs we will be showing them to **all** developers browsing our docs. 

Screen Shot 2019-06-04 at 10 45 08

And so it's really important that they be fully-fleshed out and robust. Until now, they have been more minimal with the expectation that those reading them will also read the source-code, but this won't be the case for the majority of the readers. Therefore, I would appreciate it if you guys could spend more time making the function comments more robust and over-explain things. 

E.g.:

```
 /// @dev Fetches all order-relevant information that is needed to validate that the supplied 0x order is
///  fillable and for how much it can be filled.
/// @param order The order structure.
/// @param signature Order signature corresponding to the supplied order. If the signature does not need to be
/// validated, `0x01` can be supplied for the signature to always return `false`.
/// @return OrderInfo, the remaining amount fillable by the taker, and validity of signature for the given order. 
/// If the `takerAssetData` encodes for multiple assets, `fillableTakerAssetAmount` will represent a "scaled"
/// amount, meaning it must be multiplied by all the individual asset amounts within the multiAsset to get the final
/// amount of each asset that can still be filled.
```

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+            LibOrder.OrderInfo memory orderInfo,
+            uint256 fillableTakerAssetAmount,
+            bool isValidSignature
+        )
+    {
+        // Get info specific to order
+        orderInfo = EXCHANGE.getOrderInfo(order);
+
+        // Validate the maker's signature
+        // If the signature does not need to be validated, `0x01` can be supplied for the signature to always return `false`.
```
fabioberger: Move to function comment since then it'll be picked up by the auto-generated docs.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -0,0 +1,190 @@
...
+    {
+        if (assetData.equals(ZRX_ASSET_DATA)) {
+            return ERC20_PROXY_ADDRESS;
+        }
+        bytes4 assetProxyId = assetData.readBytes4(0);
+        assetProxyAddress = EXCHANGE.getAssetProxy(assetProxyId);
+        return assetProxyAddress;
+    }
+
+    /// @dev Gets the amount of an asset transferable by the owner.
```
fabioberger: Let's mention how multiAsset is handled here. It's important for devs to understand.

contracts/dev-utils/test/order_validation_utils.ts
```diff
@@ -0,0 +1,393 @@
...
+            await erc20Token.approve.awaitTransactionSuccessAsync(erc20Proxy.address, signedOrder.makerAssetAmount, {
+                from: makerAddress,
+            });
+            const [, fillableTakerAssetAmount] = await devUtils.getOrderRelevantState.callAsync(
+                signedOrder,
+                signedOrder.signature,
+            );
+            expect(fillableTakerAssetAmount).to.bignumber.equal(signedOrder.takerAssetAmount);
+        });
+        it('should return a fillableTakerAssetAmount when the remaining takerAssetAmount is less than the transferable amount', async () => {
```
fabioberger: Perhaps we should add a test for an order involving a multiAsset?

packages/abi-gen-wrappers/src/generated-wrappers/coordinator.ts
```diff
@@ -56,6 +56,62 @@ export class CoordinatorContract extends BaseContract {
             return result;
         },
     };
+    public getTransactionHash = {
```
fabioberger: What are these changes?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
abandeali1: This is still in the `development` branch so that you can start using it with 2.0 immediately, but it's written in a way that makes it very easy to update to 3.0.

packages/abi-gen-wrappers/src/generated-wrappers/coordinator.ts
xianny: Capturing from Slack: I updated the `Coordinator.json` artifact in [this PR](https://github.com/0xProject/0x-monorepo/pull/1792/files#diff-220fa7bd7fd736b670332bab49179793) because it was outdated from `Coordinator.sol`, but didn't know about updating abi-gen-wrappers.

xianny: Note that `contract-artifacts/CHANGELOG.json` should probably be updated with these changes.

PR #1852, opened 2019-06-05
Fixes false positives in `expectTransactionFailedAsync`

hysz:
## Description

The function `expectTransactionFailedAsync(transactionReceipt, expectedRevertReason)` returns a false positive when the expected revert reason is a substring of the exception message. 

For example, if the expected revert reason is `INVALID_DATA` and the actual reason is `INVALID_DATA_OFFSET` then there will be a false positive.

To go one step further, the full exception message is `VM Exception while processing transaction: revert INVALID_DATA_OFFSET`. Any substring within that message will also return a false positive, for example, `processing trans`.

## Checklist:

### DIFF
--- a/contracts/test-utils/src/assertions.ts
+++ b/contracts/test-utils/src/assertions.ts
@@ -104,7 +104,8 @@ export async function expectTransactionFailedAsync(p: sendTransactionResult, rea
     }
     switch (nodeType) {
         case NodeType.Ganache:
-            return expect(p).to.be.rejectedWith(reason);
+            const rejectionMessageRegex = new RegExp(`^VM Exception while processing transaction: revert ${reason}$`);
+            return expect(p).to.be.rejectedWith(rejectionMessageRegex);
         case NodeType.Geth:
             logUtils.warn(
                 'WARNING: Geth does not support revert reasons for sendTransaction. This test will pass if the transaction fails for any reason.',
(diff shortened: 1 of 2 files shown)


contracts/test-utils/src/assertions.ts
```diff
@@ -160,7 +161,8 @@ export async function expectTransactionFailedWithoutReasonAsync(p: sendTransacti
  * otherwise resolve with no value.
  */
 export async function expectContractCallFailedAsync<T>(p: Promise<T>, reason: RevertReason): Promise<void> {
-    return expect(p).to.be.rejectedWith(reason);
+    const rejectionMessageRegex = new RegExp(`^VM Exception while processing transaction: revert ${reason}$`);
```
albrow: I don't think this will work in Geth. We have disabled Geth tests for now (see https://github.com/0xProject/0x-monorepo/pull/1754 and https://github.com/0xProject/0x-monorepo/pull/1775). Going forward, instead of running all of our unit tests through Geth and relying on features that don't exist upstream, we want to have integration tests that operate at a higher level and don't rely on any extra features.

Anyways, I'm okay with this for now but if we write integration tests for Geth in the future we'll have to update it.

albrow:
> To go one step further, the full exception message is VM Exception while processing transaction: revert INVALID_DATA_OFFSET. Any substring within that message will also return a false positive, for example, processing trans.

Oof. If this is true we should report the issue to Mocha/Chai. I can't imagine this was the intended behavior.

hysz:
> Oof. If this is true we should report the issue to Mocha/Chai. I can't imagine this was the intended behavior.

@albrow I had the same thought and checked out their source code ([here](https://github.com/domenic/chai-as-promised/blob/master/lib/chai-as-promised.js#L135)). Funny enough it does appear to be intentional ¯\\\_(ツ)_/¯...

PR #1854, opened 2019-06-05
Remove mixins/ directories from contracts/

jalextowle:
## Description

This change removes the `mixins/` directories from the `contracts/` directory. This change required many parts of the inheritance hierarchy to change. Most of the changes to the inheritance hierarchy occurred in the `contracts/exchange/contracts/src` directory, where `MixinExchangeRichErrors.sol` became a base contract for all of the other `Mixin` files in the directory. 

Another change that is introduced in this pull request is the addition of a `lint:contracts` command in the `package.json` file. This command allows the `contracts/` directory to be linted without linting the rest of the repository. The linting is also done in stages rather than in parallel to make linting easier on machines with weak specs. 

## Testing instructions

1. `yarn build:contracts`
2. `yarn test:contracts`
3. `yarn lint:contracts`

## Types of changes

## Checklist:


contracts/coordinator/contracts/src/MixinSignatureValidator.sol
```diff
@@ -19,14 +19,28 @@
...
 
 
 contract MixinSignatureValidator is
-    MSignatureValidator
+    ISignatureValidator
 {
     using LibBytes for bytes;
 
+   // Allowed signature types.
+    enum SignatureType {
```
hysz: I'd probably also put Enum's in the interface.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -267,6 +268,48 @@ contract MixinMatchOrders is
         return matchedFillResults;
     }
 
+    function _dispatchTransferFrom(
```
hysz: How are these declarations used?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -210,6 +231,11 @@ contract MixinSignatureValidator is
         );
     }
 
+    function _getCurrentContextAddress()
```
hysz: Same comment as above - how is this used?

contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol
```diff
@@ -18,13 +18,91 @@
...
+import "../MixinSignatureValidator.sol";
+import "../MixinTransactions.sol";
 
 
-contract LibExchangeRichErrorDecoder is
-    MExchangeRichErrorTypes
-{
+contract LibExchangeRichErrorDecoder {
+    // bytes4(keccak256("SignatureError(uint8,bytes32,address,bytes)"))
+    bytes4 internal constant SIGNATURE_ERROR_SELECTOR =
```
hysz: See comment above related to these selectors.

contracts/exchange/contracts/src/MixinExchangeRichErrors.sol
```diff
@@ -20,13 +20,113 @@ pragma solidity ^0.5.9;
...
+        INAPPROPRIATE_SIGNATURE_TYPE
+    }
+
+    enum TransactionErrorCodes {
+        NO_REENTRANCY,
+        ALREADY_EXECUTED,
+        EXPIRED
+    }
+
+    // bytes4(keccak256("SignatureError(uint8,bytes32,address,bytes)"))
```
hysz: It could be better to put these in a standalone contract. Especially since they're repeated below in `LibExchangeRichErrorDecoder.sol`.

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: This (as well as all other abstract functions that I added) are needed to specify the functions in the Mixin. In this case, `_dispatchTransferFrom` and `_getCurrentContextAddress` are used in the private function at the end of the file. 

We could eliminate some of these functions by creating special purpose interfaces for them. I didn't do this initially because it felt like doing this was sort of against the spirit of these changes. This said, I'm happy to try and come up with some interfaces that we could use like that.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
abandeali1: I think it probably makes sense to just inherit `MixinTransactions` whenever they are used.

contracts/exchange/contracts/src/MixinWrapperFunctions.sol
```diff
@@ -451,4 +452,17 @@ contract MixinWrapperFunctions is
         }
         return fillResults;
     }
+
+    function _fillOrder(
```
abandeali1: This contract should probably just inherit `MixinExchangeCore` as well.

contracts/exchange/contracts/src/interfaces/IDispatcher.sol
```diff
@@ -0,0 +1,13 @@
+pragma solidity ^0.5.9;
+
+
+contract IDispatcher {
```
abandeali1: Why do we need this? Interfaces should only contain external/public functions.

contracts/exchange/contracts/src/interfaces/IExchangeCore.sol
```diff
@@ -21,10 +21,45 @@ pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
 import "@0x/contracts-exchange-libs/contracts/src/LibFillResults.sol";
+import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
```
abandeali1: Why is this duplicated?

contracts/exchange/contracts/src/interfaces/IExchangeCore.sol
```diff
@@ -57,4 +92,14 @@ contract IExchangeCore {
         public
         view
         returns (LibOrder.OrderInfo memory orderInfo);
+
+    function _isValidOrderWithHashSignature(
```
abandeali1: Why does this internal function need to be in the interface?

contracts/exchange/contracts/src/interfaces/IExchangeRichErrors.sol
```diff
@@ -1,30 +1,12 @@
...
-
-*/
-
 pragma solidity ^0.5.9;
 
-import "@0x/contracts-utils/contracts/src/mixins/MRichErrorTypes.sol";
-import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
 
+contract IExchangeRichErrors {
+    enum AssetProxyDispatchErrorCodes {
```
abandeali1: Styling nit: Add a newline after `contract IExchangeRichErrors {` (I've seen this in a few other places as well).

contracts/exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol
```diff
@@ -34,14 +35,14 @@ contract LibExchangeRichErrorDecoder is
         public
         pure
         returns (
-            SignatureErrorCodes errorCode,
+            IExchangeRichErrors.SignatureErrorCodes errorCode,
```
abandeali1: Why this change?

contracts/exchange/contracts/src/interfaces/IFillAssertions.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
+import "@0x/contracts-exchange-libs/contracts/src/LibFillResults.sol";
+
+
+contract IFillAssertions {
```
abandeali1: I think I'd prefer removing this and `IGetContext` in favor of just inheriting `MixinExchangeCore`. Down the line the actual implementations for these functions can potentially live in a separate mixin. I'd leave it as is for now, since I think we'll be modifying the structure of these contracts a good amount soon.

contracts/exchange/contracts/src/MixinMatchOrders.sol
hysz: Ohh, I see what's going on. So declarations in `MFoobar` would need to be replaced by their respective implementation, defined in `MixinFoobar`. In this case, `MixinMatchOrders` would now inherit from `MixinAssetProxyDispatcher` rather than `MAssetProxyDispatcher`.

contracts/exchange/contracts/src/interfaces/IExchangeRichErrors.sol
jalextowle: Do we need to put a new line after every contract declaration, or only when the declaration fits on a singe line?

Ex. Would the following need a new line after `{`?

```
contract A is 
    B,
    C
{
    enum D { }
}

File: contracts/utils/contracts/src/RichErrors.sol [hunk 13]
```diff
@@ -18,12 +18,12 @@
 
 pragma solidity ^0.5.9;
 
-import "./mixins/MRichErrors.sol";
 
+contract RichErrors {
+    // bytes4(keccak256("Error(string)"))
```
abandeali1: Nit: add newline before this comment

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -54,20 +53,20 @@ contract MixinTransactions is
     }
 
     /// @dev Executes a batch of Exchange method calls in the context of signer(s).
-    /// @param transactions Array of 0x transactions containing salt, signerAddress, and data.
+    /// @param zeroxTransactions Array of 0x transactions containing salt, signerAddress, and data.
```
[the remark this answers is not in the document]
jalextowle: It was shadowing the mapping, so I changed it to avoid the compiler warning. I can switch it back if you'd rather.

contracts/exchange/contracts/src/interfaces/IExchangeRichErrors.sol
abandeali1: Only after the declaration fits on a single line (so you wouldn't need one in your example).

contracts/exchange/contracts/src/interfaces/IExchangeCore.sol
abandeali1: Hmm, looks like the extra import is still here.

PR #1855, opened 2019-06-05
Refactor LibAssetData

abandeali1:
## Description

- Changes `getAllowance` in `LibAssetData` to `getAssetProxyAllowance`. The previous implementation did not work correctly with `MultiAssetProxy` data, since multiple spenders are required for a single `assetData`. Now the proxy addresses are determined when the contract is deployed, and you are no longer able to pass in a `spender`.
- Since `LibAssetData` now requires the `Exchange` address to be passed into the constructor, it has been also moved into the `dev-utils` package.
- `getBalance` and `getAssetProxyAllowance` no longer revert under any circumstances (and will return 0 instead).

## Checklist:


contracts/dev-utils/CHANGELOG.json
```diff
@@ -5,6 +5,22 @@
             {
                 "note": "Create dev-utils package",
                 "pr": 1848
+            },
+            {
+                "note": "Add `LibssetData` and `LibTransactionDecoder` contracts",
```
fabioberger:
```suggestion
                "note": "Add `LibAssetData` and `LibTransactionDecoder` contracts",
```

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -33,23 +33,26 @@ contract OrderValidationUtils is
...
-    /// @param signature Proof that order has been created by maker.
-    /// @return OrderInfo, the remaining amount fillable by the taker, and validity of signature for given order.
+    /// @param signature Signature provided by maker that proves the order's authenticity.
+    /// `0x01` can always be provided if the signature does not need to be validated.
+    /// @return orderInfo The hash, status, and `takerAssetAmount` already filled for the given order.
+    /// @return fillableTakerAssetAmount The amount of the order's `takerAssetAmount` that is fillable given all on-chain state.
+    /// If the `takerAssetData` encodes data for multiple assets, `fillableTakerAssetAmount` will represent a "scaled"
+    /// amount, meaning it must be multiplied by all the individual asset amounts within the `takerAssetData` to get the final
+    /// amount of each asset that can be filled.
+    /// @return isValidSignature The validity of the provided signature.
```
fabioberger: My gut is multiple `@return` won't work. Can we either remove or find out?

abandeali1: It didn't work well, removed.

contracts/dev-utils/contracts/src/LibAssetData.sol
```diff
@@ -0,0 +1,518 @@
...
+            // Get array of values and array of assetDatas
+            (, uint256[] memory amounts, bytes[] memory nestedAssetData) = decodeMultiAssetData(assetData);
+
+            uint256 length = nestedAssetData.length;
+            for (uint256 i = 0; i != length; i++) {
+                // Query allowance of individual assetData
+                uint256 totalAllowance = getAssetProxyAllowance(ownerAddress, nestedAssetData[i]);
+
+                // Scale total allowance down by corresponding value in assetData
+                uint256 scaledAllowance = totalAllowance / amounts[i];
```
hysz: I wonder if we should special case this for `isApprovedForAll`. Suppose the nested assets are all ERC1155 with `isApprovedForAll=true`. A client may expect the returned allowance to be `_MAX_UINT256`.

abandeali1: It's an interesting idea, but it does introduce another edge case where `order.takerAssetAmount * allowance` could overflow. Might still make this change at some point, but I think I'll punt on it for now.

PR #1859, opened 2019-06-06
[WIP] Asset-buyer 2.0 Exchange SwapQuote Consumer

dave4506:
## Checklist:

### DIFF
--- a/packages/asset-buyer/src/quote_consumers/exchange_swap_quote_consumer.ts
+++ b/packages/asset-buyer/src/quote_consumers/exchange_swap_quote_consumer.ts
@@ -56,5 +56,5 @@ export class ExchangeSwapQuoteConsumer implements SwapQuoteConsumer {
 
     public executeSwapQuoteOrThrowAsync = async (quote: SwapQuote, opts: Partial<SwapQuoteExecutionOpts>): Promise<string> => {
 
-    }
+    };
 }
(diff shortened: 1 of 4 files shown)


packages/asset-buyer/src/quote_consumers/exchange_swap_quote_consumer.ts
```diff
@@ -0,0 +1,127 @@
...
+        const calldataHexString = abiEncoder.encode(args);
+        return {
+            calldataHexString,
+            methodAbi,
+            to,
+            ethAmount,
+        };
+    }
+
+    public async getSmartContractParamsOrThrowAsync(quote: SwapQuote, opts: Partial<SwapQuoteGetOutputOpts>): Promise<SmartContractParamsInfo<ExchangeMarketBuySmartContractParams>> {
```
fragosti: I don't think you're using the `opts` here.

packages/asset-buyer/src/quote_consumers/exchange_swap_quote_consumer.ts
```diff
@@ -0,0 +1,127 @@
...
+            params,
+            to: this._contractWrappers.exchange.address,
+            methodAbi,
+        };
+    }
+
+    public async executeSwapQuoteOrThrowAsync(quote: SwapQuote, opts: Partial<SwapQuoteExecutionOpts>): Promise<string> {
+        assert.isValidSwapQuote('quote', quote);
+
+        const { takerAddress, gasLimit, gasPrice } = opts;
```
fragosti: No default values here?

dave4506: These values are all optional so there is no need for a default value obj.

packages/asset-buyer/src/quote_consumers/exchange_swap_quote_consumer.ts
dave4506: Later, when supporting marketSell, the opts will probably be used.

PR #1868, opened 2019-06-17
Add transfer simulator and reorder Exchange transfers

abandeali1:
## Description

- Adds `MixinTransferSimulator`, which performs a series of transfers and then always reverts. Example usage can be found in `OrderTransferSimulationUtils`.
- Adds `OrderTransferSimulationUtils`, which allows batch order transfer simulations.
- Reorders transfers in `fillOrder` and `matchOrders` such that the maker can spend the takerAsset  and such that both the maker/taker can spend received assets on fees.
-Removes some redundant checks done before transfers. `dispatchTransferFrom` will already not perform a transfer if `from == to`.

## Checklist:


contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -0,0 +1,152 @@
...
+            (, bytes32 failedTransferIndex,) = decodeAssetProxyDispatchError(returnData);
+            return OrderTransferResults(uint8(uint256(failedTransferIndex)));
+        } else if (selector == ASSET_PROXY_TRANSFER_ERROR_SELECTOR) {
+            // Decode AssetProxyTransferError and return index of failed transfer
+            (bytes32 failedTransferIndex, ,) = decodeAssetProxyTransferError(returnData);
+            return OrderTransferResults(uint8(uint256(failedTransferIndex)));
+        } else if (keccak256(returnData) == _TRANSFERS_SUCCESSFUL_RESULT_HASH) {
+            // All transfers were successful
+            return OrderTransferResults.TransfersSuccessful;
+        }
```
hysz: Nit - it could be good to add a catch-all `else` here to show that we've considered all cases + for future-proofing.

contracts/exchange/contracts/src/MixinTransferSimulator.sol
```diff
@@ -0,0 +1,59 @@
...
+        bytes[] memory assetData,
+        address[] memory fromAddresses,
+        address[] memory toAddresses,
+        uint256[] memory amounts
+    )
+        public
+    {
+        uint256 length = assetData.length;
+        for (uint256 i = 0; i != length; i++) {
+            _dispatchTransferFrom(
```
hysz: Somewhere it may be helpful to note that passing in the index as the order hash is not simply filler, but necessary for `getSimulatedOrderTransferResults` to retrieve the index of the failed transfer.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -336,39 +335,34 @@ contract MixinMatchOrders is
         ) {
             // Fee recipients and taker fee assets are identical, so we can
             // transfer them in one go.
-            if (takerAddress != leftFeeRecipientAddress) {
```
hysz: +1 RIP

contracts/exchange/contracts/src/MixinTransferSimulator.sol
```diff
@@ -0,0 +1,59 @@
...
+
+
+contract MixinTransferSimulator is
+    MixinAssetProxyDispatcher
+{
+    /// @dev This function may be used to simulate any amount of transfers
+    /// As they would occur through the Exchange contract. Note that this function
+    /// will always revert, even if all transfers are successful. However, it may
+    /// be used with eth_call or with a try/catch pattern in order to simulate
+    /// the results of the transfers.
```
hysz: This is an interesting design pattern. Why is this better than reverting only when there is an error?

dorothy-zbornak: I think we would need a way to detect we're in an `eth_call` context otherwise.

hysz: Ah, that's right. Okay 👍

abandeali1: Because it's a public function with access to assets, so we want to ensure that it always reverts state when called :) The purpose of this is to allow validation of arbitrary transfers within a single RPC call (people have been experiencing rate limiting issues + this will speed up validation a good amount).

PR #1881, opened 2019-06-21
Add dummy contract to test abi-gen-wrappers

### LINKED ISSUE #1880: Add dummy contract to test abi-gen-wrappers
We have been working on deprecating `@0x/contract-wrappers` by replacing it with code auto-generated by the `abi-gen` tool. To increase efficiency, we are trying to replace calls to pure Solidity functions with simulated EVM computations (#1872). As part of this push, it would be useful to have a dummy contract with different kinds of test functions so we can debug EVM evaluations in isolated unit tests. 

## Expected behaviour
A dummy contract implementing different kinds of pure Solidity functions, and corresponding unit tests. The intended workflow is:
1. Make desired changes to `abi-gen` or `abi-gen-templates`
2. Build the dummy contract package and generate wrappers for it
3. Run the unit tests to ensure the wrappers work as expected

cc @fabioberger

xianny:
## Description

Adds a dummy contract to use for testing pure functions in abi-gen-wrappers. Resolves #1880.

## Testing instructions

Unit tests for the package are included. 

## Types of changes

## Checklist:


xianny:
Thanks @abandeali1 , good idea. I ran into some awkwardness with a circular build dependency and had to resolve it with a second `tsconfig.json`. Kind of weird - what do you think?

I'm also getting some mysterious failures with chai BigNumber now. Running CI to see if it happens there too.

.gitignore
```diff
@@ -83,6 +83,7 @@ packages/react-docs/example/public/bundle*
 packages/testnet-faucets/server/
 
 # generated contract artifacts/
+contracts/abi-gen-test/generated-artifacts/
```
fabioberger: Is this still needed?

.gitignore
```diff
@@ -95,11 +96,13 @@ contracts/erc1155/generated-artifacts/
 contracts/extensions/generated-artifacts/
 contracts/exchange-forwarder/generated-artifacts/
 contracts/dev-utils/generated-artifacts/
+/packages/abi-gen/test/generate_test/generated-artifacts
 packages/sol-tracing-utils/test/fixtures/artifacts/
 packages/metacoin/artifacts/
 
 # generated contract wrappers
 packages/abi-gen-wrappers/wrappers
+contracts/abi-gen-test/generated-wrappers/
```
fabioberger: Same here.

packages/abi-gen/package.json
```diff
@@ -8,16 +8,33 @@
...
+        "fix": "tslint --fix --format stylish --project . --exclude ./test/generate_test/generated-wrappers/**/* --exclude ./test/generate_test/generated-artifacts/**/* --exclude **/lib/**/* && yarn lint-contracts",
+        "clean": "shx rm -rf lib test/generate_test/generated-artifacts test/generate_test/generated-wrappers",
         "build": "tsc -b",
+        "pre_test": "run-s compile generate_contract_wrappers build:test",
+        "build:test": "tsc --project test/generate_test/tsconfig.json ",
         "build:ci": "yarn build",
-        "test": "yarn run_mocha",
-        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js --bail --exit",
+        "test": "yarn pre_test && yarn run_mocha",
+        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js test/generate_test/lib/**/*_test.js --timeout 100000 --bail --exit",
```
fabioberger: Should dir `generate_test` actually be named `generated_test`?

packages/abi-gen/package.json
```diff
@@ -8,16 +8,33 @@
...
         "test:circleci": "yarn test:coverage",
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
-        "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info"
+        "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
+        "rebuild_and_test": "run-s build test",
+        "test:profiler": "SOLIDITY_PROFILER=true run-s build run_mocha profiler:report:html",
+        "test:trace": "SOLIDITY_REVERT_TRACE=true run-s build run_mocha",
+        "compile": "sol-compiler",
+        "watch": "sol-compiler -w",
```
fabioberger: These commands should be namespaced. These commands would also make sense in the context of the Typescript code in this repo. `compile:sol` and `watch:sol`

packages/abi-gen/test/generate_test/contracts/LibDummy.sol
```diff
@@ -0,0 +1,42 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.5;
+
+
+library LibDummy {
```
fabioberger: Was this pulled from OpenZepplin?

packages/abi-gen/test/generate_test/contracts/AbiGenDummy.sol
```diff
@@ -0,0 +1,79 @@
...
+        return 1 + x;
+    }
+
+    function pureFunctionWithConstant ()
+        public
+        pure
+        returns (uint256 someConstant)
+    {
+        return SOME_CONSTANT;
+    }
```
fabioberger: Let's add a case where we call `ecrecover` since we know that failed in the WASM EVM implementation.

fabioberger: Any other edge-cases you encountered while implementing your EVM PR?

packages/abi-gen/test/generate_test/test/abi_gen_dummy_test.ts
```diff
@@ -0,0 +1,130 @@
...
+        expect(result).bignumber.to.equal(new BigNumber(2));
+    });
+
+    it('should call a library function referencing a constant', async () => {
+        const result = await libDummy.publicAddConstant.callAsync(new BigNumber(1));
+        expect(result).bignumber.to.equal(new BigNumber(1235));
+    });
+});
+
+// HACK (xianny): copied from @0x/contracts-test-utils to avoid circular dependency
```
fabioberger:
```suggestion
// HACK(xianny): copied from @0x/contracts-test-utils to avoid circular dependency
```

fabioberger: Is this causing a circular dep event when imported as a `devDependency`?

.gitignore
xianny: nope, removing. thanks for catching :)

packages/abi-gen/test/generate_test/contracts/LibDummy.sol
xianny: Nope!

Just took a quick look at their repos.  Is there something you have in mind that we could use here?

packages/abi-gen/test/generate_test/test/abi_gen_dummy_test.ts
xianny: Yeah, setting it to `devDependency` instead of `dependency` doesn't seem to make a difference :/ 

```
ERROR: Dependency cycle detected:
   @0x/abi-gen-wrappers <- @0x/abi-gen <- @0x/contracts-test-utils <- @0x/order-utils <- @0x/abi-gen-wrappers 
```

I removed `abi-gen` completely from `contracts-test-utils`. This is with `abi-gen` listed as `devDependency` in `abi-gen-wrappers`.

packages/abi-gen/test/generated_test/contracts/AbiGenDummy.sol
```diff
@@ -0,0 +1,90 @@
...
+        returns (uint256 someConstant)
+    {
+        return SOME_CONSTANT;
+    }
+
+    function simpleRevert ()
+        public
+        pure
+    {
+        revert("VALIDATOR_ERROR");
```
feuGeneA: I suggest using a different revert message for each revert, so that if one fails it's obvious which one you should go look at.  (This comment applies to all the other reverts too, not just this one.)

packages/abi-gen/test/generated_test/test/abi_gen_dummy_test.ts
```diff
@@ -0,0 +1,149 @@
...
+        providerUtils.startProviderEngine(provider);
+        abiGenDummy = await AbiGenDummyContract.deployFrom0xArtifactAsync(artifacts.AbiGenDummy, provider, txDefaults);
+    });
+    after(async () => {
+        provider.stop();
+    });
+    beforeEach(async () => {
+        await blockchainLifecycle.startAsync();
+    });
+    afterEach(async () => {
```
feuGeneA: I suggest doing the `blockchainLifecycle.startAsync()` and `...revertAsync()` in `before()` and `after()` rather than `...each()`.  No need to undeploy and redeploy the contract for every test case, since you're not mutating contract state with your test cases (right?).  Won't make THAT much of a difference, but our tests need every little speed-up they can get.

packages/abi-gen/package.json
```diff
@@ -8,16 +8,33 @@
     "main": "lib/src/index.js",
     "types": "lib/src/index.d.ts",
     "scripts": {
-        "lint": "tslint --format stylish --project .",
-        "fix": "tslint --fix --format stylish --project .",
-        "clean": "shx rm -rf lib",
+        "lint": "tslint --format stylish --project . --exclude ./test/generated_test/generated-wrappers/**/* --exclude ./test/generated_test/generated-artifacts/**/* --exclude **/lib/**/* && yarn lint-contracts",
+        "fix": "tslint --fix --format stylish --project . --exclude ./test/generated_test/generated-wrappers/**/* --exclude ./test/generated_test/generated-artifacts/**/* --exclude **/lib/**/* && yarn lint-contracts",
```
feuGeneA: I suggest moving these exclusions to `abi-gen/tslint.json`.

packages/abi-gen/package.json
```diff
@@ -8,16 +8,33 @@
...
+        "clean": "shx rm -rf lib test/generated_test/generated-artifacts test/generated_test/generated-wrappers",
```
feuGeneA:
```suggestion
        "clean": "shx rm -rf lib test/generated_test/generated-*",
```

feuGeneA:
I suggest that you check in the generated wrappers, in a separate "known-good" location (separate from the folder where the test run will generate them), and have `yarn test` include doing a diff of the known-good vs the freshly-generated.  This will allow future reviewers of handlebars template changes to easily see the resulting changes in generated output.  (And if a template change doesn't result in 
a change to the test fixture (dummy) wrapper, then that change should be considered uncovered by tests!) See [my comment on the corresponding Issue for more background](https://github.com/0xProject/0x-monorepo/issues/1880#issuecomment-505972087).

packages/abi-gen/test/generate_test/contracts/AbiGenDummy.sol
xianny: None that I recall right now :)

packages/abi-gen/test/generate_test/contracts/LibDummy.sol
fabioberger: Hm, can't seem to find the one I meant. Asked @abandeali1 about it.

fabioberger:
> I suggest that you check in the generated wrappers, in a separate "known-good" location (separate from the folder where the test run will generate them), and have `yarn test` include doing a diff of the known-good vs the freshly-generated. This will allow future reviewers of handlebars template changes to easily see the resulting changes in generated output. (And if a template change doesn't result in
> a change to the test fixture (dummy) wrapper, then that change should be considered uncovered by tests!) See [my comment on the corresponding Issue for more background](https://github.com/0xProject/0x-monorepo/issues/1880#issuecomment-505972087).

That's an interesting idea @feuGeneA. When and how would these "known-good" contracts get updated?

xianny:
hey @fabioberger , we continued the discussion [here](https://github.com/0xProject/0x-monorepo/issues/1880#issuecomment-506058361)

Devs should update the "known-good" wrappers at the same time they make changes to `abi-gen` or `abi-gen-templates` that affect the generated-wrappers. This forces devs to check in the changes and review generated diffs at that time. `yarn test` will fail if the "known-good" wrappers aren't up to date.

Issue #1882, opened 2019-06-21
Include ABI in generated wrappers


xianny:
## Problem

Currently, when instantiating a wrapper from `abi-gen-wrappers`, developers have to import the correct contract artifact and pass its ABI as an argument to the constructor. This introduces potential for error if the developer passes in an ABI that is different from the one used to generate the wrapper, and is also unnecessarily tedious. 

## Proposed solution

When generating the wrapper file, hardcode the ABI as part of the wrapper class. This reduces errors as the generated methods of the class will always match the hardcoded ABI.

PR #1883, opened 2019-06-21
[Multilang] hardcode ABI into abi-gen-wrappers

xianny:
## Description

Fixes: #1882

This PR also re-exports `@0x/contract-addresses` from `@0x/abi-gen-wrappers` for convenience. Developers will be able to instantiate wrappers from `@0x/abi-gen-wrappers` without needing other imports. 

**\*\*BREAKING CHANGE\*\*** 
**wrappers will no longer accept ABIs as a constructor argument!**

## Testing instructions

Everything should build and pass unit tests as before.

## Types of changes

## Checklist:


packages/abi-gen-templates/contract.handlebars
```diff
@@ -105,14 +105,23 @@ export class {{contractName}}Contract extends BaseContract {
...
+    constructor(address: string, supportedProvider: SupportedProvider, txDefaults?: Partial<TxData>) {
+        super('{{contractName}}', ABI(), address, supportedProvider, txDefaults);
+        classUtils.bindAll(this, ['_abiEncoderByFunctionSignature', 'address', '_web3Wrapper']);
     }
-} // tslint:disable:max-file-line-count
+} 
+
+// Returns the ABI that was used to generate this wrapper
+// Hardcoded, do not edit!
+function ABI(): ContractAbi {
```
xianny: Should we make this a class method instead?

fabioberger:
@xianny if you start your PR description with `Fixes: #1882` merging this PR will automatically close the issue.

contracts/coordinator/package.json
```diff
@@ -71,7 +71,7 @@
         "@0x/base-contract": "^5.1.0",
         "@0x/contracts-asset-proxy": "^2.1.5",
         "@0x/contracts-erc20": "^2.2.5",
-        "@0x/contracts-exchange": "1.0.2",
+        "@0x/contracts-exchange": "^2.1.5",
```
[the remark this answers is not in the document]
xianny: Every time we run `contracts:build` from monorepo root, all the generated wrappers in all of the contracts packages are re-generated and pick up the changes made in `abi-gen`/`abi-gen-templates`/`abi-gen-wrappers`. Even though these generated wrappers aren't committed, they are exported and other contracts packages depend on them.

`contracts-coordinator` was pinned to an old version of `contracts-exchange` which was cached in `node_modules` and not picking up any of the `abi-gen` changes. The pinning seemed unintentional because `contracts-exchange` is only imported in test files in `contracts-coordinator`, so I updated it.

packages/abi-gen-templates/contract.handlebars
```diff
@@ -105,14 +105,23 @@ export class {{contractName}}Contract extends BaseContract {
...
+
+    // HACK (xianny): handlebars cannot directly pass an array
```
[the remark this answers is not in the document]
xianny: I haven't really been able to find documentation explicitly saying this. But handlebars documentation lacks any examples of returning anything except strings in helpers, and there are a lot of StackOverflow/reddit/etc threads about people trying to return complex objects and failing. I can make this comment a bit more descriptive though.

packages/abi-gen-templates/contract.handlebars
```diff
@@ -105,14 +105,21 @@ export class {{contractName}}Contract extends BaseContract {
...
         logUtils.log(`{{contractName}} successfully deployed at ${txReceipt.contractAddress}`);
-        const contractInstance = new {{contractName}}Contract(abi, txReceipt.contractAddress as string, provider, txDefaults);
+        const contractInstance = new {{contractName}}Contract(txReceipt.contractAddress as string, provider, txDefaults);
         contractInstance.constructorArgs = [{{> params inputs=ctor.inputs}}];
         return contractInstance;
     }
-    constructor(abi: ContractAbi, address: string, supportedProvider: SupportedProvider, txDefaults?: Partial<TxData>) {
-        super('{{contractName}}', abi, address, supportedProvider, txDefaults);
-        classUtils.bindAll(this, ['_abiEncoderByFunctionSignature', 'address', 'abi', '_web3Wrapper']);
+    // Returns the contract ABI
```
fabioberger: This should be a properly formatted javadoc-style comment. `@return ...`

packages/abi-gen-templates/contract.handlebars
```diff
@@ -105,14 +105,21 @@ export class {{contractName}}Contract extends BaseContract {
...
+    public static ABI(): ContractAbi {
+        // HACK (xianny): handlebars cannot directly pass an array of objects unless the template statically enumerates all the properties of all the objects
```
fabioberger: Adding a `// Source: ` to stackoverflow or reddit works too

packages/abi-gen/src/index.ts
```diff
@@ -143,6 +143,7 @@ for (const abiFileName of abiFileNames) {
     const contextData = {
         contractName: namedContent.name,
         ctor,
+        ABI: JSON.stringify(ABI),
         methods: methodsData,
         events: eventAbis,
```
fabioberger: @xianny how come we're able to pass in the `eventAbis` directly but not the ABI's?

xianny: We pass in the `eventAbis` (and `methodAbis`) and then iterate through them to pull out certain named properties, e.g. `name`. We can do that too with the contract ABI. What we can't do is reference the whole object as we iterate. If we iterate through an array of primitives, we can print each primitive with `{{this}}`. But if we iterate through an array of objects and try to reference `{{this}}`, we'll get `Object object`. 

Screen Shot 2019-06-28 at 11 20 37 AM
Screen Shot 2019-06-28 at 11 21 24 AM

I haven't been able to find a single place that explicitly says you can't do this, but after combing through the docs and experimenting with different kinds of handlebars block helpers, I think it's going to be a waste of time to continue.

xianny: Not sure what to do about the comment. I could make a stack overflow post with more details and link to it from the comment?

xianny: The workaround would be to hardcode all the object properties. I guess that's fine - ABI typedef is pretty stable.

packages/abi-gen-wrappers/src/generated-wrappers/coordinator_registry.ts
```diff
@@ -219,32 +198,42 @@ export class CoordinatorRegistryContract extends BaseContract {
         ]);
         const provider = providerUtils.standardizeOrThrow(supportedProvider);
         const constructorAbi = BaseContract._lookupConstructorAbi(abi);
-        [] = BaseContract._formatABIDataItemList(
-            constructorAbi.inputs,
-            [],
-            BaseContract._bigNumberToString,
-        );
+        [] = BaseContract._formatABIDataItemList(constructorAbi.inputs, [], BaseContract._bigNumberToString);
```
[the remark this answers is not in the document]
xianny: we don't run `prettier` on `abi-gen-wrappers` generated wrappers. I think we should, as long as we are comparing diffs on them.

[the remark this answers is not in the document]
feuGeneA: totally agree. i've already got Python set up to run `black` (akin to `prettier`) on wrappers.  granted, that's because i'm depending on `black` to align indents and do other whitespace changes necessary to satisfy the Python interpreter, but it sure does make for nicely legible code in the end!

packages/abi-gen/src/index.ts
fabioberger: Ok! Np, then, was just curious but it makes sense that we're using the entire ABI differently.

packages/abi-gen-templates/contract.handlebars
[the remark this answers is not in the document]
fabioberger: Ok, np then.

packages/abi-gen-wrappers/package.json
```diff
@@ -15,6 +15,7 @@
         "lint": "tslint --format stylish --project .",
         "fix": "tslint --fix --format stylish --project .",
         "pre_build": "yarn generate_contract_wrappers",
+        "prettier": "prettier --write src/**/* --config ../../.prettierrc",
```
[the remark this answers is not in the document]
xianny: 🗣 discussed and agreed to do more general overhaul of abi-gen build process and file structure. Won't address this here.

packages/abi-gen-wrappers/src/generated-wrappers/coordinator_registry.ts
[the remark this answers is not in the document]
feuGeneA: Resolved by creating issue https://github.com/0xProject/0x-monorepo/issues/1917

packages/abi-gen-wrappers/package.json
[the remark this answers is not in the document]
feuGeneA: cf. https://github.com/0xProject/0x-monorepo/issues/1917 & https://github.com/0xProject/0x-monorepo/issues/1925

packages/abi-gen/test/generated-test/generated-artifacts/AbiGenDummy.json
```diff
@@ -0,0 +1,175 @@
+{
+	"schemaVersion": "2.0.0",
+	"contractName": "AbiGenDummy",
+	"compilerOutput": {
```
feuGeneA: Did you check in the `generated-artifacts` folder on accident?  This change seems unrelated to the commit title ("prettier"), and I thought that the current setup has these files being re-generated on every test run (though btw I have doubts about the appropriateness of that approach...).

xianny: gah. Yes these shouldn't be checked in. Thanks for the catch!

PR #1885, opened 2019-06-22
Always validate callback signatures & add support for EIP-1271 signatures

dorothy-zbornak:
## Description

This PR addresses ZEIP [#33](https://github.com/0xProject/ZEIPs/issues/33).

cat

### Contract Signature Types Always Checked
We now *always* check any signature type that relies on a contract. Prior to this, we only checked the signature callback on the first fill. Now this occurs on every subsequent fill. The signature types affected are:
- `Wallet`
- `Validator`
- `EIP1271Wallet` (*new: see below*)

### Consolidated Signature Types
- A new signature type, `EIP1271Wallet`, has been added for wallets that implement the [EIP-1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md) standard, which take an arbitrary `bytes data` field along with a signature.
- `WalletOrderValidator` has been removed since `EIP1271Wallet` is a much more flexible alternative now.
- There is now only a single `Validator` signature type for signatures that specify an external validator contract.
- All these signature types also work with `ZeroExTransaction`s, in addition to `Order`s!

### Interface for `EIP1271Wallet` and `Validator` contracts:
Both of these signature types follow the EIP-1271 pattern and require a contract that exposes the following callback:
```solidity
function isValidSignature(
    // ABI-encoded data associated with the signature.
    bytes calldata data,
    // arbitrary signature bytes passed in top-level call.
    bytes calldata signature
)
    external
    view
    // Should return 0x20c13b0b if the signature is valid.
    returns (bytes4 magicValue);
```

#### EIP-1271 `data` Encoding
There are two ways the `data` parameter for `EIP1271Wallet` and `Validator` signatures can be encoded, depending on what type of data is signed:
- For Orders: `abi.encode(Order order, bytes32 orderHash)`
- For ZeroExTransactions: `abi.encode(ZeroExTransaction transaction, bytes32 transactionHash)`

### Bonus Material
- I consolidated all the various wallet/validator test contracts into a single configurable one and rewrote most of the tests in `signature_validator.ts` to match.
- I killed a handful of signature tests in in `core.ts` because I felt like they were already pretty well covered in `signature_validator.ts`. 
- A `validatorAddress` field has been added to `SignatureValidatorError`.
- Promises returned by `awaitTransactionSuccessAsync()` are actually compatible with the base `Promise` type now (not just `PromiseLike`).
- `assertValidFill()` and `calculateMatchedFillResults()` functions on the Exchange have been made public.

## Testing instructions

`yarn test:contracts` and cross your fingers.

## Types of changes

## Checklist:


contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -305,6 +350,53 @@ contract MixinSignatureValidator is
...
+        uint256 signatureLength = signature.length;
+        // Shave the signature type off the signature.
+        assembly {
+            mstore(signature, sub(signatureLength, 1))
+        }
+        // Encode the call data.
+        bytes memory data = new bytes(32);
+        data.writeBytes32(0, hash);
+        bytes memory callData = abi.encodeWithSelector(
+            IEIP1271Wallet(walletAddress).isValidSignature.selector,
```
abandeali1: Can we hard code these selectors somewhere? (or at least check what the gas differences are).

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -305,6 +350,53 @@ contract MixinSignatureValidator is
...
+        );
+        // Restore the full signature.
+        assembly {
+            mstore(signature, signatureLength)
+        }
+        // Static call the verification function.
+        (bool didSucceed, bytes memory returnData) = walletAddress.staticcall(callData);
+        // Return data should be the `EIP1271_MAGIC_VALUE`.
+        if (didSucceed && returnData.length == 32) {
+            return bytes4(returnData.readBytes32(0)) == EIP1271_MAGIC_VALUE;
```
abandeali1: Why not use `readBytes4` here?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -305,6 +350,53 @@ contract MixinSignatureValidator is
         ));
     }
 
+    /// @dev Verifies signature using logic defined by an EIP1271 Wallet contract.
+    /// @param hash Any 32 byte hash.
+    /// @param walletAddress Address that should have signed the given hash
+    ///                      and defines its own signature verification method.
+    /// @param signature Proof that the hash has been signed by signer.
+    /// @return True if the signature is validated by the Walidator.
+    function _validateHashWithEIP1271Wallet(
```
abandeali1: This signature type doesn't feel necessary to me. The hash can always be derived from the data, so I guess this is just a cheaper alternative?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -404,7 +500,55 @@ contract MixinSignatureValidator is
...
+    }
+
+    /// @dev Verifies order AND signature via an EIP1271 Wallet contract.
+    /// @param order The order.
+    /// @param orderHash The order hash.
+    /// @param walletAddress Address that should have signed the given hash
+    ///                      and defines its own order/signature verification method.
+    /// @param signature Proof that the order has been signed by signer.
+    /// @return True if order and signature are validated by the Wallet.
+    function _validateOrderWithEIP1271Wallet(
```
abandeali1: I wonder if we should consolidate this signature type with the `OrderWallet` type, since they pretty much serve the same purpose. The `OrderWallet` receiver takes an `orderHash` argument in addition to what we are doing here, which doesn't feel that necessary. That could also be encoded in the data if we want to keep it in.

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -0,0 +1,296 @@
...
+
+
+// solhint-disable no-unused-vars
+contract TestValidatorWallet {
+    using LibBytes for bytes;
+
+    // Revert reason for `Revert` `ValidatorAction`.
+    string constant public REVERT_REASON = "you shall not pass";
+    // Magic bytes returned by EIP1271 wallets on success.
+    bytes4 constant public EIP1271_MAGIC_VALUE = 0x20c13b0b;
```
abandeali1: I'm seeing this constant in a lot of places. Is there a single place it would make sense to keep this and inherit from?

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -0,0 +1,296 @@
...
+        ValidateSignature,
+        NTypes
+    }
+
+    /// @dev The Exchange contract.
+    ISimplifiedExchange internal _exchange;
+    /// @dev Internal state to modify.
+    uint256 internal _state = 1;
+    /// @dev What action to execute when a hash is validated .
+    mapping (bytes32=>ValidatorAction) internal _hashActions;
```
abandeali1: Nit: `mapping (bytes32 => ValidatorAction)`

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -0,0 +1,296 @@
...
+    /// @dev Allowed signers for hash signature types.
+    mapping (bytes32=>address) internal _validSignerForHash;
```
abandeali1: `mapping (bytes32 => address)`

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -0,0 +1,296 @@
...
+        } else if (action == ValidatorAction.Accept) {
+            magicValue = EIP1271_MAGIC_VALUE;
+        } else if (action == ValidatorAction.Revert) {
+            revert(REVERT_REASON);
+        } else if (action == ValidatorAction.UpdateState) {
+            _updateState();
+        } else { // action == ValidatorAction.ValidateSignature
+            if (data.length != 32) {
+                // `data` is an abi-encoded Order.
+                LibOrder.Order memory order = _getOrderFromEIP1271Data(data);
```
abandeali1: Why not use `abi.decode(data, (LibOrder.Order))` here?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: Just to be extra sure they meant to return the magic bytes and not something that just coincidentally starts with the magic bytes, I suppose.

contracts/exchange/contracts/test/TestValidatorWallet.sol
dorothy-zbornak: Ran into stack issues. Stuff like this might make `EIP1271OrderWallet` not a super attractive alternative to `OrderWallet`, and why I wouldn't be in favor of `EIP1271OrderWallet` supplanting `OrderWallet`. Thoughts?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: Yeah. I was mainly aiming for parity with existing signature types. I would be in favor of removing this signature type, but my concern is what if we want to use EIP-1271 with `ZeroExTransaction`s, which currently only pass a hash?

contracts/exchange/contracts/test/TestValidatorWallet.sol
dorothy-zbornak: Oh, it works with `abi.decode(data, (LibOrder.Order))`. I didn't know you could do that. I was doing it the long way before (explicitly passing in all the fields).  :trollface:

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: I am down for this, now that I see that decoding is nbd.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -305,6 +350,53 @@ contract MixinSignatureValidator is
...
+        bytes32 hash,
+        address walletAddress,
+        bytes memory signature
+    )
+        private
+        view
+        returns (bool isValid)
+    {
+        uint256 signatureLength = signature.length;
+        // Shave the signature type off the signature.
```
hysz: I believe we can use `LibBytes.popLastByte` here.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -305,6 +350,53 @@ contract MixinSignatureValidator is
...
+        }
+        // Encode the call data.
+        bytes memory data = new bytes(32);
+        data.writeBytes32(0, hash);
+        bytes memory callData = abi.encodeWithSelector(
+            IEIP1271Wallet(walletAddress).isValidSignature.selector,
+            data,
+            signature
+        );
+        // Restore the full signature.
```
hysz: I'm concerned that we may extend/refactor this someday, and it's not necessarily the case that this byte will still be in-place in memory. We could probably copy the logic from `LibAddressArray.append` to write a proper  `LibBytes.append`. Although if that's too expensive, perhaps just tag this as a hack and note not to write to memory.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -404,7 +500,55 @@ contract MixinSignatureValidator is
...
+        bytes32 orderHash,
+        address walletAddress,
+        bytes memory signature
+    )
+        private
+        view
+        returns (bool isValid)
+    {
+        uint256 signatureLength = signature.length;
+        // Shave the signature type off the signature.
```
hysz: Same notes as above for this function

contracts/exchange/test/signature_validator.ts
```diff
@@ -255,185 +224,255 @@ describe('MixinSignatureValidator', () => {
...
-                ethUtil.toBuffer(ecSignature.v),
-                ecSignature.r,
-                ecSignature.s,
-                ethUtil.toBuffer(`0x${SignatureType.EthSign}`),
+                crypto.randomBytes(SIGNATURE_LENGTH),
+                ethUtil.toBuffer([SignatureType.EthSign]),
             ]);
             const signatureHex = ethUtil.bufferToHex(signature);
             // Validate signature.
             // This will fail because `signerAddress` signed the message, but we're passing in `notSignerAddress`
```
jalextowle: Small nit: This comment should probably be updated to reflect that `signerAddress` is being passed in, and that the signature is invalid because it's a set of random bytes.
```suggestion
             // This will fail because the `signature` is a sequence of random bytes, so the `signedAddress` 
             // will not be recovered from the signature and the order hash.
```

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: Doh, you guys are totes right. Put the cast on the wrong side.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: Yeah, `append` would be cleaner but feels like overkill for this, especially now that we verify certain signatures on every fill. The rest of `MixinSignatureValidator` has been previously rewritten to be non-destructive to the signature (mostly so we can revert with the full signature), but I will underscore the necessity for this pattern in the comments.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: If Remix is to be believed, on `0.5.10`, this way is weirdly cheaper *both* with and without optimizations by 28 and 8 gas, respectively.  :woman_shrugging:

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -177,6 +171,66 @@ contract MixinExchangeCore is
...
+    /// @param takerAssetFilledAmount Amount of takerAsset that will be filled.
+    /// @param makerAssetFilledAmount Amount of makerAsset that will be transfered.
+    function assertValidFill(
+        Order memory order,
+        OrderInfo memory orderInfo,
+        uint256 takerAssetFillAmount,  // TODO: use FillResults
+        uint256 takerAssetFilledAmount,
+        uint256 makerAssetFilledAmount
+    )
+        public
```
abandeali1: Is this useful as a public function? Feels like it might not be, since you need to know the results fo the fill before calling it.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -183,32 +214,168 @@ contract MixinSignatureValidator is
...
             signature
         );
-        if (signatureType == SignatureType.OrderValidator) {
-            // The entire order is verified by validator contract.
-            isValid = _validateOrderWithValidator(
-                order,
+        if (signatureType == SignatureType.Validator) {
+            // The entire order is verified by a validator contract.
+            isValid = _validateBytesWithValidator(
+                abi.encode(order, orderHash),
```
abandeali1: I'm wondering if it still makes sense to pass the `orderHash` in here. Would you blindly trust that the hash corresponds to the data when verifying the signature? I feel like if you _only_ want the hash, you'd probably just use the regular `Wallet` type. If you want both the hash and the data, you probably won't trust the hash without verifying.

abandeali1: I also feel like the `Validator` type isn't very useful with just a hash, generally speaking.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -183,32 +214,168 @@ contract MixinSignatureValidator is
...
                 signature
             );
-            return isValid;
-        } else if (signatureType == SignatureType.WalletOrderValidator) {
+        } else if (signatureType == SignatureType.EIP1271Wallet) {
             // The entire order is verified by a wallet contract.
-            isValid = _validateOrderWithWallet(
-                order,
+            isValid = _validateBytesWithWallet(
+                abi.encode(order, orderHash),
```
abandeali1: Same as above.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
abandeali1: Bizarre... going to add a task to test this throughout the codebase. It's nice not having to maintain all of these contracts that just contain constants.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -399,29 +520,29 @@ contract MixinSignatureValidator is
         }
         // Static call the verification function.
         (bool didSucceed, bytes memory returnData) = walletAddress.staticcall(callData);
-        // Return data should be a single bool.
-        if (didSucceed && returnData.length == 32) {
-            return returnData.readUint256(0) == 1;
+        // Return data should be the `EIP1271_MAGIC_VALUE`.
+        if (didSucceed && returnData.length <= 32) {
```
abandeali1: Shouldn't `returnData.length` always be 4?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
dorothy-zbornak: My arguments for including the hash would be that the hash is a very useful, canonical ID, which is not cheap to compute. For example, a validator could use it to immediately look up the `filled` state of the order.

Also, wouldn't just gating your callback such that `msg.sender == exchange` be sufficient to trust that the hash provided is authentic?

contracts/exchange/contracts/src/MixinExchangeCore.sol
dorothy-zbornak: I agree. There's a lot of burden to get to the point to where you could even make this call. I think a higher level function would be more useful (maybe there's one already exposed). Unless @fabioberger has any objections, I'm making this internal again.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
abandeali1: I suppose it can't hurt. Curious to see how much it gets used.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -123,42 +142,54 @@ contract MixinSignatureValidator is
...
         bytes memory signature
     )
         public
         view
         returns (bool isValid)
     {
-        SignatureType signatureType = _readValidSignatureType(
-            hash,
+        bytes32 transactionHash = getTransactionHash(transaction);
+        return _isValidTransactionWithHashSignature(
```
jalextowle: You can remove the `isValid` name for the return value.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -123,42 +142,54 @@ contract MixinSignatureValidator is
...
+    /// @return needsRegularValidation True if the signature should be validated
+    ///                                for every action.
+    function doesSignatureRequireRegularValidation(
+        bytes32 hash,
+        address signerAddress,
+        bytes memory signature
+    )
+        public
+        pure
+        returns (bool needsRegularValidation)
```
jalextowle: Same here. Is having named return values a part of our style guide?

abandeali1: For the most part, yes. I believe it saves a stack variable slot, which is pretty valuable in some cases.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -20,24 +20,26 @@ pragma solidity ^0.5.9;
...
-import "./MixinTransactions.sol";
 import "./MixinExchangeRichErrors.sol";
+import "./MixinTransactions.sol";
 
 
-contract MixinSignatureValidator is 
-    MixinExchangeRichErrors, 
-    ReentrancyGuard, 
-    LibOrder, 
+contract MixinSignatureValidator is
```
hysz: This is out of scope for this PR, but I think it could be good to abstract a lot of this logic into a generic library like `libSignatureValidator` that can be used across our packages.

PR #1900, opened 2019-06-26
Batch Order Matching

jalextowle:
## Description
This PR partially address [ZEIP 41](https://github.com/0xProject/ZEIPs/issues/41).

image

### Batch Order Matching
These changes add a `batchMatchOrders` function, which allows several orders to be matched in a single transaction. This implementation can easily be modified to also meet the requirements for `batchMatchOrdersWithMaximalFill` as soon as [ZEIP 40](https://github.com/0xProject/ZEIPs/issues/40) is implemented.

### Simulating Batch Order Matching
A `batchMatchOrdersAndAssertEffectsAsync` function was added to facilitate testing the new `batchMatchOrders` function, and in the process I was able to refactor some of the `match_order_tester.ts` file to be more general. 

## Testing instructions

## Types of changes

## Checklist:


contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -33,6 +33,101 @@ contract MixinMatchOrders is
...
+            );
+
+            // Add the matchResults and the profit made during the match to the
+            // batchMatchedFillResults for this batch.
+            batchMatchedFillResults.left[i] = matchResults.left;
+            batchMatchedFillResults.right[i] = matchResults.right;
+            batchMatchedFillResults.profitInLeftMakerAsset +=
+                matchResults.left.makerFeePaid;
+            batchMatchedFillResults.profitInRightMakerAsset +=
+                matchResults.right.makerFeePaid;
```
dorothy-zbornak: The fees aren't the profit. The profit is the spread between the two orders, the given to the taker (matcher). Currently we only support a spread in the left maker asset (that's what ZEIP-40 would address). So I think you want something like:
```solidity
batchMatchedFillResults.profitInLeftMakerAsset += matchResults.leftMakerAssetSpreadAmount;
batchMatchedFillResults.profitInRightMakerAsset += 0;
```

Should probably use `_safeAdd()` for these, as well.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +279,125 @@ contract MixinMatchOrders is
...
+    /// @return A boolean reflecting whether or not the order was fully filled
+    function _isFilled(
+        LibOrder.Order memory order,
+        LibFillResults.FillResults memory fillResults
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        if (fillResults.takerAssetFilledAmount >= order.takerAssetAmount) {
```
dorothy-zbornak: So this is the one mixin I haven't really touched, so maybe I'm wrong, but this doesn't feel right. I think `fillResults.takerAssetFilledAmount` is how much this order was filled in *this* fill, not total. I think you might actually need to keep a running tally of the fill amounts for left and right orders in your batch matching function.

jalextowle:
Due to the fact that ZEIP 40 addresses several of the issues in this ZEIP, more tests for `batchMatchOrders` will be added in that PR.

contracts/exchange/test/utils/exchange_wrapper.ts
```diff
@@ -266,6 +266,22 @@ export class ExchangeWrapper {
...
+        const txHash = await this._exchange.batchMatchOrders.sendTransactionAsync(
+            params.leftOrders,
+            params.rightOrders,
+            params.leftSignatures,
+            params.rightSignatures,
+            { from },
+        );
+        const tx = await this._logDecoder.getTxWithDecodedLogsAsync(txHash);
+        return tx;
+    }
```
dorothy-zbornak: We should probably add a `getMatchOrdersResultsAsync()` and `getBatchMatchOrdersResultsAsync()` (where we do an `eth_call` to get the return value) to this class and validate the return values also in the tests.

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: After some digging, I think that you're right about this one. I'll update!

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +285,121 @@ contract MixinMatchOrders is
...
 
+    /// @dev Determines whether or not an order has been fully filled and returns the result.
+    /// @param order The order that should be checked.
+    /// @return A boolean reflecting whether or not the order was fully filled
+    function _isFilled(LibOrder.Order memory order)
+        internal
+        view
+        returns (bool)
+    {
+        LibOrder.OrderInfo memory orderInfo = getOrderInfo(order);
```
dorothy-zbornak: Could we move this logic into `batchMatchOrders()` so we only call `getOrderInfo()` once per order? `getOrderInfo()` computes the hash and does a couple SLOADs each time you call it, so it would be nice to minimize it.

You just need to update `OrderInfo.orderTakerAssetFilledAmount` with each match operation and use that to check if its been fully filled.

contracts/exchange/test/match_orders.ts
```diff
@@ -1849,5 +1850,235 @@ describe('matchOrders', () => {
...
+            const rightOrders = [
+                await orderFactoryRight.newSignedOrderAsync({
+                    makerAddress: makerAddressRight,
+                    makerAssetAmount: Web3Wrapper.toBaseUnitAmount(1, 0),
+                    takerAssetAmount: Web3Wrapper.toBaseUnitAmount(2, 0),
+                    feeRecipientAddress: feeRecipientAddressRight,
+                }),
+            ];
+            const tx = exchangeWrapper.batchMatchOrdersAsync(leftOrders, rightOrders, takerAddress);
+            return expect(tx).to.be.rejected();
```
dorothy-zbornak: Can we use `to.revertWith()` here with a specific rich revert error type?

contracts/exchange/test/match_orders.ts
```diff
@@ -1849,5 +1850,235 @@ describe('matchOrders', () => {
...
+                await orderFactoryLeft.newSignedOrderAsync({
+                    makerAddress: makerAddressLeft,
+                    makerAssetAmount: Web3Wrapper.toBaseUnitAmount(1, 0),
+                    takerAssetAmount: Web3Wrapper.toBaseUnitAmount(2, 0),
+                    feeRecipientAddress: feeRecipientAddressLeft,
+                }),
+            ];
+            const rightOrders: SignedOrder[] = [];
+            const tx = exchangeWrapper.batchMatchOrdersAsync(leftOrders, rightOrders, takerAddress);
+            return expect(tx).to.be.rejected();
```
dorothy-zbornak: Ditto.

contracts/exchange/test/match_orders.ts
```diff
@@ -1849,5 +1850,235 @@ describe('matchOrders', () => {
...
+        });
```
dorothy-zbornak: Would also be wise to add tests that hit the other two revert errors.

contracts/exchange/test/match_orders.ts
```diff
@@ -1849,5 +1850,235 @@ describe('matchOrders', () => {
...
+                    rightOrders,
+                    leftOrdersTakerAssetFilledAmounts: [ZERO],
+                    rightOrdersTakerAssetFilledAmounts: [ZERO, ZERO],
+                },
+                takerAddress,
+                [[0, 0], [0, 1]],
+                expectedTransferAmounts,
+            );
+        });
+    });
```
dorothy-zbornak: One test I would like to see is the case where there are unbalanced left and right orders. Basically, there would be one extra order on either side that shouldn't be filled at all because the other side has been fully consumed. Do any of these do that?

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -33,6 +33,110 @@ contract MixinMatchOrders is
...
+
+            // Add the matchResults and the profit made during the match to the
+            // batchMatchedFillResults for this batch.
+            batchMatchedFillResults.left[matchCount] = matchResults.left;
+            batchMatchedFillResults.right[matchCount] = matchResults.right;
+            batchMatchedFillResults.profitInLeftMakerAsset = _safeAdd(
+                batchMatchedFillResults.profitInLeftMakerAsset,
+                matchResults.leftMakerAssetSpreadAmount
+            );
+            // batchMatchedFillResults.profitInRightMakerAsset += 0; // Placeholder for ZEIP 40
```
dorothy-zbornak: I know I told you to just comment this out before, but seeing as the ZEIP-40 PR is up now, could you retrofit `rightMakerAssetSpreadAmount` into `MatchedFillResults` in this PR and flesh out the logic for this line? I don't want us to forget about this line later.

contracts/exchange/test/utils/match_order_tester.ts
```diff
@@ -126,6 +148,59 @@ export class MatchOrderTester {
...
+        expect(matchPairs.length).to.be.eq(expectedTransferAmounts.length);
+        expect(orders.leftOrders.length).to.be.eq(orders.leftOrdersTakerAssetFilledAmounts.length);
+        expect(orders.rightOrders.length).to.be.eq(orders.rightOrdersTakerAssetFilledAmounts.length);
+        // Ensure that the exchange is in the expected state.
+        await assertBatchOrderStatesAsync(orders, this.exchangeWrapper);
+        // Get the token balances before executing `batchMatchOrders()`.
+        const _initialTokenBalances = initialTokenBalances
+            ? initialTokenBalances
+            : await this._initialTokenBalancesPromise;
+        // Execute `batchMatchOrders()`
```
dorothy-zbornak: Right before this, can we do an `exchangeWrapper.getBatchMatchOrdersResultsAsync()` and assert the results against the simulation? I don't see where we are checking the return value of `batchMatchOrders()` anywhere.

Would also be nice to have this done for the regular `matchOrders()` tests, as well.

contracts/exchange/test/utils/match_order_tester.ts
```diff
@@ -634,4 +910,61 @@ function encodeTokenBalances(obj: any): any {
...
+        for (const contract of _.keys(totalBalances.erc721[owner])) {
+            totalBalances.erc721[owner][contract] = _.zipWith(
+                totalBalances.erc721[owner][contract],
+                balances.erc721[owner][contract],
+                initialBalances.erc721[owner][contract],
+                (a: BigNumber, b: BigNumber, c: BigNumber) => a.plus(b.minus(c)),
+            );
+        }
+    }
+    // TODO(jalextowle): Implement the same as the above for ERC1155
```
dorothy-zbornak: Do we have any `matchOrders` coverage for 1155s? Let's add an asana task for this.

contracts/exchange/test/match_orders.ts
jalextowle: Yes. Basically the first orders fully fill themselves, and the last order shouldn't be touched. There is a test for this for both sides.

contracts/exchange/test/utils/match_order_tester.ts
jalextowle: I'll check and get back to you.

PR #1904, opened 2019-06-27
Adding dYdX to Explore page

chriskalani:
Screen Shot 2019-06-26 at 5 11 20 PM


packages/website/ts/types.ts
```diff
@@ -255,6 +255,7 @@ export interface ExploreProject {
     url: string;
     keywords: string[];
     instant?: ExploreProjectInstantMetadata;
+    logo_max_width?: string;
```
fragosti: camelCase -> `logoMaxWidth`

fragosti: Oh realized we're doing this snake case stuff elsewhere... shouldn't be that way :p

AntonioJuliano:
😄

packages/website/ts/types.ts
dave4506: I think the snake case is on me, I was confused it was snake case as well; dunno what the heck I was thinking when I wrote it.

PR #1878, opened 2019-06-19
First auto-gen'd Python contract wrapper: ERC20

feuGeneA:
## Description

Fixes #1887.  Fixes #1899.

Change `abi-gen` to support generating Python, including a new handlebars template to produce Python contract wrappers, and use it to replace the existing Python implementation of the ERC20 contract wrapper.  Existing tests/examples should all pass with minimal changes (ideally none).

## Checklist:

-   [x] Modify `abi-gen` interface to allow specification of language, and draft Python Handlebars templates.
-   [x] Conform to linter `black`.
-   [x] Conform to linter `pycodestyle`.
-   [x] Conform to linter `pydocstyle`.
-   [x] Conform to linter `mypy`.
-   [x] Conform to linter `bandit`.
-   [x] Conform to linter `pylint`. (Make method names `snake_case` (not camel), and remove unused import.)
-   [x] Generate event wrappers.
-   [x] Have wrapper store the ABI used by the template, rather than referring to the one in `zero_ex.contract_artifacts` (which won't even be available for arbitrary (non-0x) contracts).
-   [x] Have constructor accept contract address as arg, remove arg from methods.
-   [x] Figure out why my `abi-gen` changes have caused changes to generated _TypeScript_ code, and either fix it so they don't change or commit the changed versions.
-   [x] Add new entries to the relevant CHANGELOG.jsons.
-   [x] Update READMEs with info about how, when and where code generation output flows through/to the different components.
-   [x] Move Python stuff out of `abi-gen-wrappers` and into its own folder under `0x-monorepo/packages`.
-   [x] Add some tests of Python generation to `abi-gen`, given the new framework merged in by #1881 
-   [x] Fix newly-written but failing test: return type missing on some generated methods.
-   [x] Fix newly-written but failing test: `BaseContract` and `TxParams` classes should be imported from canonical package, not relative path. relative path is working for wrapping OUR contracts, but would not be usable in a more general context.
-   [x] In template, initialize base class per best practices.  Source: [comment below](https://github.com/0xProject/0x-monorepo/pull/1878#discussion_r299248641).

PR #1907, opened 2019-06-28
Add devdoc to all contract artifacts

### LINKED ISSUE #1899: ethereum-types:  The Devdoc output doesn't cover possible valid output.
## Expected Behavior

https://github.com/0xProject/0x-monorepo/blob/e20fa3a1bf3b801ef1748792d983fb79fccb1265/packages/ethereum-types/src/index.ts#L614-L626

Should cover all possible valid compiled output.

## Current Behavior
Typings cause issues when sol-compiler includes devdoc output.

## Possible Solution

The custom types we used to override
```
interface KosuDevDocOutput {
    title?: string;
    author?: string;
    methods?: {
        [signature: string]: {
            details?: string;
            params?: {
                [name: string]: string;
            };
            return?: string;
        };
    };
}
```

## Steps to Reproduce (for bugs)

```
1. Add devdocs to compiler.json outputSelection
1. Compile a contract missing some part of the devdocs required by the typing. 
1. Attempt to compile the typescript.
```

## Context
We are attempting to use Solidity devdocs to generate some documentation.  The types are requiring us to override artifacts.ts generated with contracts-gen

## Your Environment

| Package | Version |
| ------: | :------ |

| Network |
| ------- |
| NAME    |

### LINKED ISSUE #1906: Add devdoc to all contract artifacts
## Expected Behavior

We want to add devdocs to generated wrappers for #1903 (and also for the Python generated wrappers). To do this, we need to first add devdocs to the contract artifacts. 

This change should be careful not to introduce bytecode changes.

xianny:
Resolves #1906.  Resolves #1899.

## Description

* Add devdocs to all contract-artifacts
* Update contract-artifacts scripts to require devdoc
* Does NOT introduce bytecode changes. I noticed that some of the `*.sol` files in `development` branch generate different bytecode from artifacts in `development` branch, and manually removed the changes.
* Fixes a small bug in the contract-artifacts `copy` script where it was outputting to the wrong path

@feuGeneA I cherry-picked your commit from #1878 to hopefully minimise merge conflicts!

## Testing instructions

* Ensure no bytecode changes
* Ensure all artifacts have devdocs
* Run unit tests

## Types of changes

## Checklist:


packages/contract-artifacts/artifacts/OrderValidator.json
```diff
@@ -231,6 +231,7 @@
                 "type": "constructor"
             }
         ],
+        "devdoc": { "methods": { } },
```
feuGeneA: Gasp! :)

packages/metacoin/compiler.json
```diff
@@ -6,6 +6,7 @@
             "*": {
                 "*": [
                     "abi",
+                    "devdoc",
```
feuGeneA: This file changed, but there is no corresponding change to any artifacts anywhere.  Seems like there should be either both or neither.

packages/contract-artifacts/artifacts/OrderValidator.json
xianny: lol! I think WETH9 also doesn't have any devdocs. `OrderValidator` needs to be removed anyway. I'll do that in a separate PR.

packages/metacoin/compiler.json
xianny: This is an example contract. I'll add an example devdoc to one of the methods. But it's not going to be re-exported in `contract-artifacts`. I think we should keep this to show that we include devdoc as a best practice, since we expect `abi-gen` to depend on devdocs.

feuGeneA: Makes sense, thanks.  (Micro nit: `abi-gen` should _support_ devdocs, but not _depend on_ them.  Some users will have _no_ devdocs, and that's okay.)

PR #1913, opened 2019-07-01
New Order Matching Strategy

jalextowle:
## Description

These changes implements the functionality of [ZEIP 40](https://github.com/0xProject/ZEIPs/issues/40) and the remaining functionality from [ZEIP 41](https://github.com/0xProject/ZEIPs/issues/41).

image

### New Order Matching Strategy
The bulk of the changes in this pull request are related to the new order matching strategy that ZEIP 40 introduces. 

The new order matching strategy comes with a new function that is used to expose the functionality of the new order matching strategy:

```
    function matchOrdersWithMaximalFill(
        LibOrder.Order memory leftOrder,
        LibOrder.Order memory rightOrder,
        bytes memory leftSignature,
        bytes memory rightSignature
    )
        public
        nonReentrant
        returns (LibFillResults.MatchedFillResults memory matchedFillResults);
```

The new order matching strategy allows profit to be paid out of the `left maker asset`, the `right maker asset`, or both. There are three cases in the order matching strategy:
- `leftTakerAssetAmountRemaining` > `rightMakerAssetAmountRemaining`, which indicates that the right order can be fully filled but the left order cannot. In this case, the profit will be paid out in the `left maker asset`, and the profit will be calculated as the spread in the `left maker asset`.
- `rightTakerAssetAmountRemaining` > `rightMakerAssetAmountRemaining`, which indicates that the right order can be fully filled but the left order cannot. In this case, the profit will be paid out in the `right maker asset`, and the profit will be calculated as the spread in the `right maker asset`.
- `leftTakerAssetAmountRemaining` > `rightMakerAssetAmountRemaining && rightTakerAssetAmountRemaining` <= `rightMakerAssetAmountRemaining`, which indicates that both orders can be fully filled. In this case, the profit will be paid out in both maker assets, and the profit will be calculated as the spread in the `left maker asset` and the spread in the `right maker asset`.

### Batch Orders with Maximal Fill
A reentrant version of the function is available through the internal function `_matchOrders` if its argument `withMaximalFill` is true, which allowed a `batchMatchOrdersWithMaximalFill` function to be created using almost identical logic to `batchMatchOrders`.

### Split up TestExchangeInternals
We have now reached the point where `TestExchangeInternals` will not deploy with all of the internal functions defined. To facilitate the process of breaking of the tests, I created a `TestExchangeMath` contract that only inherits the math functions of the exchange. Over time, the creation of `Test` contracts that only inherit from one or several mixins may become necessary if deployment costs continue to increase.

## Testing instructions

```yarn build:contracts && yarn test:contracts```

## Types of changes

## Checklist:

### CODE REVIEW

**Comment by jalextowle** [comment]
I'm working on the `out of gas` issue. The tests that pertain to `matchOrdersWithMaximalFill` are passing, so this shouldn't affect the review very much.

File: contracts/exchange/test/utils/exchange_wrapper.ts [hunk 1]
```diff
@@ -282,6 +381,52 @@ export class ExchangeWrapper {
         const tx = await this._logDecoder.getTxWithDecodedLogsAsync(txHash);
         return tx;
     }
+    public async getMatchOrdersResultsAsync(
```
dorothy-zbornak: :raised_hands:

contracts/test-utils/src/types.ts
```diff
@@ -143,3 +150,16 @@ export interface FillResults {
     makerFeePaid: BigNumber;
     takerFeePaid: BigNumber;
 }
+
+export interface MatchedFillResults {
+    left: FillResults;
+    right: FillResults;
+    leftMakerAssetSpreadAmount: BigNumber;
```
dorothy-zbornak: Missing `rightMakerAssetSpreadAmount`?

contracts/exchange/test/utils/match_order_tester.ts
```diff
@@ -108,37 +129,117 @@ export class MatchOrderTester {
...
+        expect(matchPairs.length).to.be.eq(expectedTransferAmounts.length);
+        expect(orders.leftOrders.length).to.be.eq(orders.leftOrdersTakerAssetFilledAmounts.length);
+        expect(orders.rightOrders.length).to.be.eq(orders.rightOrdersTakerAssetFilledAmounts.length);
+        // Ensure that the exchange is in the expected state.
+        await assertBatchOrderStatesAsync(orders, this.exchangeWrapper);
+        // Get the token balances before executing `batchMatchOrders()`.
+        const _initialTokenBalances = initialTokenBalances
+            ? initialTokenBalances
+            : await this._initialTokenBalancesPromise;
+        // Execute `batchMatchOrders()`
```
dorothy-zbornak: Around here we need to call `getBatchMatchOrders...ResultsAsync()` and assert the results against our simulation.

Ditto for the `matchOrdersAndAssertEffectsAsync()` function.

contracts/exchange/test/match_orders.ts
```diff
@@ -1849,5 +1882,860 @@ describe('matchOrders', () => {
...
+                    leftOrdersTakerAssetFilledAmounts: [ZERO],
+                    rightOrdersTakerAssetFilledAmounts: [ZERO],
+                },
+                takerAddress,
+                [[0, 0]],
+                expectedTransferAmounts,
+                true,
+            );
+        });
+        it('should correctly match one left order to two right orders, where the last should not be touched', async () => {
```
dorothy-zbornak: :pray:

contracts/test-utils/src/types.ts
jalextowle: Oops, I'll update!

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -52,85 +117,182 @@ contract MixinMatchOrders is
...
+        }
+        if (rightOrders.length != rightSignatures.length) {
+            _rrevert(BatchMatchOrdersError(BatchMatchOrdersErrorCodes.INCOMPATIBLE_RIGHT_ORDERS));
+        }
+
+        // Without simulating all of the order matching, this program cannot know how many
+        // matches there will be. To ensure that batchMatchedFillResults has enough memory
+        // allocated for the left and the right side, we will allocate enough space for the
+        // maximum amount of matches (the maximum of the left and the right sides).
+        uint256 maxLength = _max256(leftOrders.length, rightOrders.length);
```
hysz: On each iteration at least one order will get filled. I believe the following scenario is possible and would result in a greater number of matches.
```
leftOrders = [orderA, orderB]
rightOrders = [orderC, orderD]

# iteration 1
orderC fills orderA
# iteration 2
orderB fills orderC
# iteration 3
orderB fills orderD
```

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -166,47 +331,126 @@ contract MixinMatchOrders is
...
-            matchedFillResults.left.takerAssetFilledAmount = matchedFillResults.right.makerAssetFilledAmount;
-            // Round down to ensure the maker's exchange rate does not exceed the price specified by the order.
-            // We favor the maker when the exchange rate must be rounded.
-            matchedFillResults.left.makerAssetFilledAmount = _safeGetPartialAmountFloor(
-                leftOrder.makerAssetAmount,
-                leftOrder.takerAssetAmount,
-                matchedFillResults.left.takerAssetFilledAmount
-            );
+        // Maximally fill the orders and pay out profits to the matcher in one or both of the maker assets.
+        if (withMaximalFill) {
```
hysz: Nit - this `if` statement is pretty lengthy. It could help with readability to break up it up into two helper functions.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -52,85 +117,182 @@ contract MixinMatchOrders is
...
-            takerAddress,
-            rightOrderInfo.orderHash,
-            rightOrderInfo.orderTakerAssetFilledAmount,
-            matchedFillResults.right
-        );
+            // If the rightOrder is filled, update the rightIdx, rightOrder, and rightSignature,
+            // or break out of the loop if there are no more rightOrders to match.
+            if (rightOrderInfo.orderTakerAssetFilledAmount >= rightOrder.takerAssetAmount) {
+                if (++rightIdx == rightOrders.length) {
+                    matchCount++;
```
hysz: Won't this result in a redundant increment of `matchCount` when both orders are fully filled?

jalextowle: Good point! I'll update

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: You're right. I should probably set `maxLength` to `leftOrders.length + rightOrders.length`.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -166,47 +331,126 @@ contract MixinMatchOrders is
...
+            if (leftTakerAssetAmountRemaining > rightMakerAssetAmountRemaining) {
+                // Case 1: Right order is fully filled with the profit paid in the left makerAsset
+                matchedFillResults.right.makerAssetFilledAmount = rightMakerAssetAmountRemaining;
+                matchedFillResults.right.takerAssetFilledAmount = rightTakerAssetAmountRemaining;
+                matchedFillResults.left.takerAssetFilledAmount = rightMakerAssetAmountRemaining;
+                // Round down to ensure the left maker's exchange rate does not exceed the price specified by the order.
+                // We favor the left maker when the exchange rate must be rounded and the profit is being paid in the
+                // left maker asset.
+                matchedFillResults.left.makerAssetFilledAmount = _safeGetPartialAmountFloor(
+                    leftMakerAssetAmountRemaining,
```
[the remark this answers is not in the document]
jalextowle: Since we use the latter in the bottom else, I'm going to switch it for now. Since the latter is the price that the maker actually selected, I think it makes more sense to use that for approximations.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -166,47 +331,126 @@ contract MixinMatchOrders is
...
+                leftSpread = true;
+            } else if (rightTakerAssetAmountRemaining > leftMakerAssetAmountRemaining) {
+                // Case 2: Left order is fully filled with the profit paid in the right makerAsset.
+                matchedFillResults.left.makerAssetFilledAmount = leftMakerAssetAmountRemaining;
+                matchedFillResults.left.takerAssetFilledAmount = leftTakerAssetAmountRemaining;
+                // Round down to ensure the right maker's exchange rate does not exceed the price specified by the order.
+                // We favor the right maker when the exchange rate must be rounded and the profit is being paid in the
+                // right maker asset.
+                matchedFillResults.right.makerAssetFilledAmount = _safeGetPartialAmountFloor(
+                    rightMakerAssetAmountRemaining,
```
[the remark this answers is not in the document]
jalextowle: Same as above.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +355,237 @@ contract MixinMatchOrders is
...
+                matchResults.left.takerAssetFilledAmount
+            );
+            rightOrderInfo.orderTakerAssetFilledAmount = _safeAdd(
+                rightOrderInfo.orderTakerAssetFilledAmount,
+                matchResults.right.takerAssetFilledAmount
+            );
+
+            // Add the matchResults and the profit made during the match to the
+            // batchMatchedFillResults for this batch.
+            batchMatchedFillResults.left[matchCount] = matchResults.left;
```
abandeali1: These results correspond to the index of each match rather than the index of each order. I feel like the latter would be more useful, since you can quickly check the results to see how much of each order has been filled (i.e `matchedFillResults.left[3]` would tell you how much of `leftOrders[3]` has been filled). It would probably require us to track a `leftFillResults` and `rightFillResults`, much in the same way we do for `orderInfo` in this implementation.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +355,237 @@ contract MixinMatchOrders is
...
+        }
+        if (rightOrders.length != rightSignatures.length) {
+            _rrevert(BatchMatchOrdersError(BatchMatchOrdersErrorCodes.INCOMPATIBLE_RIGHT_ORDERS));
+        }
+
+        // Without simulating all of the order matching, this program cannot know how many
+        // matches there will be. To ensure that batchMatchedFillResults has enough memory
+        // allocated for the left and the right side, we will allocate enough space for the
+        // maximum amount of matches (the left side length added to the right side length).
+        uint256 maxLength = leftOrders.length + rightOrders.length;
```
dorothy-zbornak: Seems like the upper bound should be `left + right - 1`?

jalextowle: You're correct. 

Proof: In the worst case scenario, every order except for one must be maximally filled, every match in the batch except for the last match (the last match doesn't matter) only maximally fills one order, and every order can be filled. One order is maximally filled during each match by the definition of our order matching strategy, so in the worst case `leftOrders.length + rightOrders.length - 1`.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -166,47 +331,126 @@ contract MixinMatchOrders is
...
+            // If the left order is fully filled, profit will be paid out in the left maker asset. If the right order is fully filled,
+            // the profit will be out in the right maker asset.
+            //
+            // There are three cases to consider:
+            // Case 1.
+            //   If the left maker can buy more than the right maker can sell, then only the right order is fully filled.
+            // Case 2.
+            //   If the right maker can buy more than the left maker can sell, then only the right order is fully filled.
+            // Case 3.
+            //   If the right maker can sell the max of what the left maker can buy and the left maker can sell the max of
```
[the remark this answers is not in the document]
jalextowle: Yes, that is correct.  In theory, the else would work, but it will do redundant computation and may have different precision. I'll update.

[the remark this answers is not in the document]
jalextowle: Re code duplication, I agree, and I'll probably address this using @hysz's above suggestion regarding helper functions.

contracts/exchange/contracts/src/MixinExchangeRichErrors.sol
```diff
@@ -28,6 +28,17 @@ contract MixinExchangeRichErrors is
     IExchangeRichErrors
 {
     // solhint-disable func-name-mixedcase
+    function BatchMatchOrdersError(BatchMatchOrdersErrorCodes errorCode)
+        internal
+        pure
+        returns (bytes memory)
+    {
+        abi.encodeWithSelector(
```
abandeali1: Can we add an explicit return here? I didn't even know you could implicitly return unnamed variables.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +258,424 @@ contract MixinMatchOrders is
...
+    ///      The profit made by the left order goes to the taker (who matched the two orders). This
+    ///      function is needed to allow for reentrant order matching (used by `batchMatchOrders` and
+    ///      `batchMatchOrdersWithMaximalFill`).
+    /// @param leftOrder First order to match.
+    /// @param rightOrder Second order to match.
+    /// @param leftSignature Proof that order was created by the left maker.
+    /// @param rightSignature Proof that order was created by the right maker.
+    /// @param shouldMaximallyFillOrders Indicates whether or not the maximal fill matching strategy should be used
+    /// @return matchedFillResults Amounts filled and fees paid by maker and taker of matched orders.
+    function _matchOrders(
```
abandeali1: Nit 1: why is this `private` while `_batchMatchOrders` is `internal`? Probably make that `private` as well.

contracts/exchange/contracts/src/MixinExchangeRichErrors.sol
jalextowle: I didn't either... That's pretty weird. I'll update!

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -262,6 +258,424 @@ contract MixinMatchOrders is
...
+            matchedFillResults.left.takerAssetFilledAmount = leftTakerAssetAmountRemaining;
+            matchedFillResults.right.makerAssetFilledAmount = leftTakerAssetAmountRemaining;
+            // Round up to ensure the maker's exchange rate does not exceed the price specified by the order.
+            // We favor the maker when the exchange rate must be rounded.
+            matchedFillResults.right.takerAssetFilledAmount = _safeGetPartialAmountCeil(
+                rightOrder.takerAssetAmount,
+                rightOrder.makerAssetAmount,
+                leftTakerAssetAmountRemaining // matchedFillResults.right.makerAssetFilledAmount
+            );
+        }
```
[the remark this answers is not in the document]
jalextowle: Nice catch!

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,446 @@ contract MixinMatchOrders is
...
+        //
+        // There are three cases to consider:
+        // Case 1.
+        //   If the left maker can buy more than or equal to the right maker can sell, then only the right order is fully filled.
+        // Case 2.
+        //   If the right maker can buy more than or equal to the left maker can sell, then only the right order is fully filled.
+        // Case 3.
+        //   If the right maker can sell the max of what the left maker can buy and the left maker can sell the max of
+        //   what the right maker can buy, then both orders are fully filled.
+        if (leftTakerAssetAmountRemaining >= rightMakerAssetAmountRemaining) {
```
hysz: I believe we want to make these `>` opposed to `>=`, since both orders will be fully filled in the `==` case. This would be aligned with the regular `matchOrders` and would eliminate unnecessary rounding in the `==` case.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,446 @@ contract MixinMatchOrders is
...
+    {
+        // Calculate fill results for maker and taker assets: at least one order will be fully filled.
+        // The maximum amount the left maker can buy is `leftTakerAssetAmountRemaining`
+        // The maximum amount the right maker can sell is `rightMakerAssetAmountRemaining`
+        // We have two distinct cases for calculating the fill results:
+        // Case 1.
+        //   If the left maker can buy more than the right maker can sell, then only the right order is fully filled.
+        //   If the left maker can buy exactly what the right maker can sell, then both orders are fully filled.
+        // Case 2.
+        //   If the left maker cannot buy more than the right maker can sell, then only the left order is fully filled.
```
hysz: Please add a comment to describe Case 3 as well.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,446 @@ contract MixinMatchOrders is
...
+        if (leftTakerAssetAmountRemaining > rightMakerAssetAmountRemaining) {
```
hysz: Readability nit:
```
if (leftTakerAssetAmountRemaining > rightMakerAssetAmountRemaining) {..}
else if (leftTakerAssetAmountRemaining < rightMakerAssetAmountRemaining) {..}
else {..} // leftTakerAssetAmountRemaining == rightMakerAssetAmountRemaining
```

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,446 @@ contract MixinMatchOrders is
...
+            // Indicate that the profit should be set to the spread denominated in the right maker asset.
+            doesRightMakerAssetProfitExist = true;
+        } else {
+            // Case 3: The right and left orders are fully filled
+            matchedFillResults.right.makerAssetFilledAmount = rightMakerAssetAmountRemaining;
+            matchedFillResults.right.takerAssetFilledAmount = rightTakerAssetAmountRemaining;
+            matchedFillResults.left.makerAssetFilledAmount = leftMakerAssetAmountRemaining;
+            matchedFillResults.left.takerAssetFilledAmount = leftTakerAssetAmountRemaining;
+            // Indicate that the profit should be set to the spread denominated in the left and the right maker assets.
+            doesLeftMakerAssetProfitExist = true;
```
hysz: If we do make the other statements `>` then this would become:
```
doesLeftMakerAssetProfitExist = rightTakerAssetAmountRemaining < leftMakerAssetAmountRemaining;
doesRightMakerAssetProfitExist = leftTakerAssetAmountRemaining < rightMakerAssetAmountRemaining;
```

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,446 @@ contract MixinMatchOrders is
...
+            matchedFillResults.right.makerAssetFilledAmount = _safeGetPartialAmountFloor(
+                rightOrder.makerAssetAmount,
+                rightOrder.takerAssetAmount,
+                leftMakerAssetAmountRemaining
+            );
+            matchedFillResults.right.takerAssetFilledAmount = leftMakerAssetAmountRemaining;
+            // Indicate that the profit should be set to the spread denominated in the right maker asset.
+            doesRightMakerAssetProfitExist = true;
+        } else {
+            // Case 3: The right and left orders are fully filled
```
hysz: To use the same DRY pattern as above, this could be moved into a function called `_calculateCompleteFillBoth`

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: Agreed, it'll be more accurate.

contracts/exchange/test/match_orders.ts
```diff
@@ -694,39 +709,7 @@ describe('matchOrders', () => {
                 },
                 takerAddress,
                 expectedTransferAmounts,
-            );
-        });
-
-        it('should transfer the correct amounts when orders completely fill each other and taker doesnt take a profit', async () => {
```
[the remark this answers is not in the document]
jalextowle: I guess I accidentally removed it. I'm putting it back in

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,473 @@ contract MixinMatchOrders is
...
+                // Update the batched fill results once the leftIdx is updated.
+                batchMatchedFillResults.left[leftIdx++] = leftFillResults;
+                // Clear the intermediate fill results value.
+                leftFillResults = LibFillResults.FillResults(0, 0, 0, 0);
+
+                // If all of the left orders have been filled, break out of the loop.
+                // Otherwise, update the current right order.
+                if (leftIdx == leftOrders.length) {
+                    // Update the right batched fill results
+                    batchMatchedFillResults.right[rightIdx] = rightFillResults;
```
hysz: We may want to add a note to the definition of `BatchMatchedFillResults` that there will be no fill result for orders that were not matched. Right now we label the fields as "Fill results for left/right orders", which is sort of ambiguous. A client may expect to receive fill results for all orders they pass in.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -263,6 +259,473 @@ contract MixinMatchOrders is
...
+                // Update the batched fill results once the rightIdx is updated.
+                batchMatchedFillResults.right[rightIdx++] = rightFillResults;
+                // Clear the intermediate fill results value.
+                rightFillResults = LibFillResults.FillResults(0, 0, 0, 0);
+
+                // If all of the right orders have been filled, break out of the loop.
+                // Otherwise, update the current right order.
+                if (rightIdx == rightOrders.length) {
+                    // Update the left batched fill results
+                    batchMatchedFillResults.left[leftIdx] = leftFillResults;
```
hysz: Is this necessary if the left order was also filled during this iteration? I believe this would write an empty fill result for the next iteration's `left` order. If this is the case, then we should ensure there is a test case for this scenario.

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: Amir and I talked about this earlier. I don’t know that we came to a conclusion on it. I’ll talk to him in the morning and see what he thinks. I’m happy to add a comment for now, and depending on what he says I’ll either create an asana task or update the PR.

contracts/exchange/contracts/src/MixinMatchOrders.sol
jalextowle: Good point. I actually don’t know if this is a problem though since in this case the next left order wouldn’t have been filled anyways. It’s redundant, but it doesn’t seem to change any behavior since we process the orders in order and the last order will never be filled in this case. I actually think it may be cleaner to keep it the same, but I’m interested to hear your thoughts.

contracts/exchange/contracts/src/MixinMatchOrders.sol
hysz: Sounds good. For this PR, we should at least ensure there is consistent behavior across match scenarios. So we should either (i) return empty fill results for all unmatched orders, or (ii) return no empty fill results for any unmatched orders. I personally think it's okay to only return a fill result for orders that are matched.

Ex of current inconsistent behavior:
```
-- Scenario 1 --
leftOrders = [order_a, order_b]
rightOrders = [order_c]
* order_a and order_c are fully matched*
leftResult = [a_fill_result, (empty)]
rightResult = [c_fill_result]

-- Scenario 2 --
leftOrders = [order_a]
rightOrders = [order_c, order_d]
* order_a and order_c are fully matched*
leftResult = [a_fill_result]
rightResult = [c_fill_result]
```

contracts/exchange/contracts/src/MixinMatchOrders.sol
hysz: Please see comment above 🙏

contracts/exchange/contracts/src/MixinMatchOrders.sol
hysz: Note:
Chatted with Alex offline. The array initialization was updated earlier in the PR, so this is just an issue with redundant writes. We may want to refactor this section so that we only write the left/right result once, but in terms of correct output this looks good to me 👍 .

PR #1915, opened 2019-07-02
[WIP] Asset buyer 2.0 Dynamic swap quote consumer

dave4506:
## Description

Created a new consumer that will use the 'best' underlying consumer (Exchange, Forwarder, in the future Coordinator) to provide output or execute the swap quote. 

If the order is forwarder valid, the consumer will look at eth and weth balances as well to choose which consumer is best. 

## Checklist:

### DIFF
--- a/packages/asset-buyer/src/quote_consumers/dynamic_swap_quote_consumer.ts
+++ b/packages/asset-buyer/src/quote_consumers/dynamic_swap_quote_consumer.ts
@@ -24,12 +24,15 @@ import { swapQuoteConsumerUtils } from '../utils/swap_quote_consumer_utils';
 import { ExchangeSwapQuoteConsumer } from './exchange_swap_quote_consumer';
 import { ForwarderSwapQuoteConsumer } from './forwarder_swap_quote_consumer';
 
-type SmartContractParams = ExchangeMarketBuySmartContractParams | ExchangeMarketSellSmartContractParams | ForwarderMarketBuySmartContractParams | ForwarderMarketSellSmartContractParams;
+type SmartContractParams =
+    | ExchangeMarketBuySmartContractParams
+    | ExchangeMarketSellSmartContractParams
+    | ForwarderMarketBuySmartContractParams
+    | ForwarderMarketSellSmartContractParams;
 
 type ValidSwapQuoteConsumer = ExchangeSwapQuoteConsumer | ForwarderSwapQuoteConsumer;
 
-export class DynamicSwapQuoteConsumer
-    implements SwapQuoteConsumer<SmartContractParams> {
+export class DynamicSwapQuoteConsumer implements SwapQuoteConsumer<SmartContractParams> {
     public readonly provider: ZeroExProvider;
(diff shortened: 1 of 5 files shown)


packages/asset-buyer/src/quote_consumers/dynamic_swap_quote_consumer.ts
```diff
@@ -0,0 +1,117 @@
...
+import { assetDataUtils } from '../utils/asset_data_utils';
+import { swapQuoteConsumerUtils } from '../utils/swap_quote_consumer_utils';
+
+import { ExchangeSwapQuoteConsumer } from './exchange_swap_quote_consumer';
+import { ForwarderSwapQuoteConsumer } from './forwarder_swap_quote_consumer';
+
+type SmartContractParams =
+    | ExchangeMarketBuySmartContractParams
+    | ExchangeMarketSellSmartContractParams
+    | ForwarderMarketBuySmartContractParams
```
fragosti: Can we export these unions from their respective classes and then have the final union here.

packages/asset-buyer/src/quote_consumers/dynamic_swap_quote_consumer.ts
```diff
@@ -0,0 +1,117 @@
...
+            const isEnoughEthAndWethBalance = _.map(takerEthAndWethBalance, (balance: BigNumber) =>
+                balance.isGreaterThanOrEqualTo(ethAmount),
+            );
+            if (isEnoughEthAndWethBalance[1]) {
+                // should be more gas efficient to use exchange consumer, so if possible use it.
+                return this._exchangeConsumer;
+            } else if (isEnoughEthAndWethBalance[0] && !isEnoughEthAndWethBalance[1]) {
+                return this._forwarderConsumer;
+            }
+            // Note: Should we default to a consumer if takerAddress has not enough Eth or Weth?
```
fragosti: Maybe return the exchange consumer? Either way I agree we should return something and let that thing deal with the error. I would replace this comment with a comment that explains the reasoning.

packages/asset-buyer/src/quote_consumers/dynamic_swap_quote_consumer.ts
```diff
@@ -0,0 +1,117 @@
...
+    public async executeSwapQuoteOrThrowAsync(
+        quote: SwapQuote,
+        opts: Partial<DynamicSwapQuoteExecutionOpts>,
+    ): Promise<string> {
+        assert.isValidSwapQuote('quote', quote);
+        const consumer = await this._getConsumerForSwapQuoteAsync(quote, opts);
+        return consumer.executeSwapQuoteOrThrowAsync(quote, opts);
+    }
+
+    private async _getConsumerForSwapQuoteAsync(
```
fragosti: Is it tenable to write a test for this function?

dave4506: making the function public; should make it testable

PR #1919, opened 2019-07-03
Auto-gen Python Exchange wrapper

### LINKED ISSUE #1888: Auto-generate Python wrapper for the Exchange contract
## Expected Behavior

`0x-monorepo/packages/python-contract-wrappers` should generate a wrapper for the `Exchange`contract, and that wrapper should be used as the basis for all of our tests and doc'd code examples.

## Current Behavior

`python-contract-wrappers` is currently generating the wrapper for the `ERC20Token` contract, and that wrapper is successfully serving as the basis for our automated tests and examples.  But it is the only generated wrapper being used in this way at this time

feuGeneA:
Fixes #1888 .

## Description

Include `Exchange` in the list of ABI's that `packages/python-contract-wrapper` is currently generating wrappers for, and change `python-packages/contract_wrappers` to pull in that code (overwriting the existing, manually-written wrapper) before linting and testing itself.  Existing tests/examples should all pass with minimal changes (ideally none).

## Testing instructions

Since the end result is generated code, reviewing that code (and the docs generated from it) is the first step in testing.

I found some CircleCI configuration to persist "Build Artifacts," which accurately describes our generated code.  With this, you can go to a build job's "Artifacts" tab ([here's a link to the "Artifacts" tab on a recent run of `test-python`](https://circleci.com/gh/0xProject/0x-monorepo/79585#artifacts/containers/0)), and drill down through the repo tree to find persisted artifacts.

I'm using this to persist generated Python wrappers, generated test output from `abi-gen`'s command-line tests, and documentation generated from the Python code.  Here are the links to the artifacts from the latest CircleCI run:

- [Generated Python Exchange wrapper](https://80172-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/__init__.py)
- [`contract_wrappers.py` generated documentation](https://80172-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/build/docs/html/index.html) (includes docs for the Exchange wrapper class)
- [Generated Python ERC20Token wrapper](https://80172-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/erc20_token/__init__.py)
---
If you want to actually run the generation process yourself:
```
0x-monorepo/python-packages$ ./pre_install
0x-monorepo/python-packages$ ./install_editable
0x-monorepo/python-packages/contract_wrappers$ PKG=@0x/python-contract-wrappers yarn build && ./setup.py pre_install && ./setup.py lint && ./setup.py test
[edit code]
0x-monorepo/python-packages/contract_wrappers$ PKG=@0x/python-contract-wrappers yarn build && ./setup.py pre_install && ./setup.py lint && ./setup.py test
[repeat]
```

## Types of changes

## Checklist:

-   [x] Rename contract methods parameters in existing exchange contract to match contract method parameter names
-   [x] Rename existing wraper from `exchange_wrapper.py` to just `exchange.py`.
-   [x] Fix faulty encoding of Order struct tuple in generated wrapper methods.  See [CI failure](https://circleci.com/gh/0xProject/0x-monorepo/77740?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).
-   [x] Generate documentation from the generated wrapper and make sure it looks okay.
-   [x] Document `bytes` parameter methods as requiring a UTF-8 encoding.
-   [x] Stop generating `TypedDict`s for structs that are only used as the return value for non-const methods, since there's no generated reference to such a `TypedDict`, because those return values aren't accessible since a non-const method can only return a transaction hash.
-   [x] Document flags used in `abi-gen/diff.sh`.  Also, move that `diff.sh` into `test-cli`, since that's the only place it's ever used, and remove the parameters, since it's always called with the same arguments, and also rename it to `run_diff_test.sh`.
-   [x] In `abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol`, rename `Event` to `AnEvent`.
-   [x] `.gitignore` JSON schema files in python package area, and also generated wrappers in python area.
-   [x] Review old, manually-written wrapper to see if there are other things that need to be ported over.
-   [x] Add JSON schema validation.
-   [x] Rebase on `development` in order to fix test-publish failure.
-   [x] ~Add signature validation.~ Decided not to do this; see [comment below](url).
-   [x] When generated methods are called with `view_only=True` (to do an `eth_call` instead of an `eth_sendTransaction`) the method should return the value returned by the method, not return a transaction hash.
-   [x] Fix bug with wrong return value for some generated methods, shown in [this comment below](https://github.com/0xProject/0x-monorepo/pull/1919#issuecomment-513203461).
-   [ ] Raise follow-on Issue: Add refined types to output values of `get_*_event()` methods.  Currently they're all just `Tuple[AttributeDict]`.
-   [ ] Raise follow-on Issue: Add support for accepting signatures as `str`ings (not just `bytes`).


feuGeneA:
This branch is currently in a good checkpoint state.  The Exchange contract is being generated and run through the tests and doctest examples with every build, and are clean in CircleCI (except for the `test-publish` failure, quoted below).

However, generated wrappers are currently not yet doing any JSON schema validation, nor any signature validation.  There may also be other specially crafted logic in there too, which would need to be ported over, but at this time all of the examples/tests are using this generated wrapper and passing.  These omissions are reflected in the checklist in the PR description.

The generated Exchange wrapper, as of this point in time, can be seen here:  https://gist.github.com/feuGeneA/de3754c87d3732e8a2e2e69e53da78de

The pre-existing, manually-written Exchange wrapper, can be seen in [the commit that deleted it](https://github.com/0xProject/0x-monorepo/pull/1919/commits/56cd84c82a8668fec750a070d5f75136cbd9094c).

For the reviewer's convenience, here is a diff of the two: https://www.diffchecker.com/7tEnPG7v .  Sadly, there is a lot of change in the diff.

`test-publish` is failing with
```
ERROR in package @0x/contracts-exchange:
Command failed: npm install --save @0x/contracts-exchange@2.1.6 --registry=http://localhost:4873/
npm WARN deprecated istanbul@0.4.5: This module is no longer maintained, try this instead:
npm WARN deprecated   npm i nyc
npm WARN deprecated Visit https://istanbul.js.org/integrations for other alternatives.
```
:man_shrugging: a re-run didn't fix it...

packages/abi-gen/src/utils.ts
```diff
@@ -190,4 +183,166 @@ export const utils = {
...
+    /**
+     * simply concatenate all of the names of the components, and convert that
+     * concatenation into PascalCase to conform to Python convention.
+     */
+    makePythonTupleName(tupleComponents: DataItem[]): string {
+        return changeCase.pascal(_.map(tupleComponents, component => component.name).join('_'));
+    },
+    /**
+     * @returns a string that is a Python code snippet that's intended to be
+     * used as the second parameter to a TypedDict() insatnatiation; value
```
fabioberger:
```suggestion
     * used as the second parameter to a TypedDict() instantiation; value
```

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
```diff
@@ -11,6 +19,30 @@
...
+class InnerStructDescription(  # noqa: E501 (line too long)
+    TypedDict
+):
+    """Python representation of a tuple or struct."""
+
+    inner_struct: SomeBytesAnIntegerADynamicArrayOfBytesAString  # noqa: E501 (line too long)
+
+    description: str
+
+class SomeBytesAnIntegerADynamicArrayOfBytesAString(  # noqa: E501 (line too long)
```
fabioberger: Instead of concatenating all the properties as the name here, let's use `typedDict0`, `typedDict1`, etc... which is also what `go-ethereum's` `abigen` tool does. Is it possible to then make named classes in a separate file that inhert from these classes? This way, users will see and be able to use the properly naming classes that we will manually maintain, but if we ever forget to create an alias, it'll default to `typedDictX`?

This is what we do in Mesh: https://github.com/0xProject/0x-mesh/blob/development/ethereum/wrappers/order_validation_utils_aliases.go

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
```diff
@@ -61,22 +94,48 @@ def simple_require(
...
+        return self._invoke_function_call(
+            func=func,
+            tx_params=tx_params,
+            view_only=True
+        )
+
     def ecrecover_fn(
         self,
-        hash: bytes,
+        _hash: bytes,
```
fabioberger: Why was the underscore added in this case?

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
```diff
@@ -88,4 +90,26 @@ contract AbiGenDummy
         return ecrecover(prefixedHash, v, r, s);
     }
 
+    event Event(uint8 param);
```
fabioberger: Let's rename this to `AnEvent`. It's a bit confusing to have it identically named to the keyword.

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
feuGeneA: Because hash is a built-in function in Python, so it shouldn't be used as an identifier name, or else `pylint` will complain about it.  It was added by `abi-gen`'s `utils.toPythonIdentifier()`, which was added in the commit titled `
abi-gen: convert Py method params to snake case`.

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
feuGeneA: The problem is that some sort of state would need to be held, such that when the references to structs are generated they can refer back to the the classes that have already been declared.  This could probably be addressed with a state variable in `abi-gen/index.src` that's captured in the closure for the helpers, but... There's also the complication of, when generating the class declarations, being able to support nested definitions, which would require multiple passes over all of the methods, since references to struct A within struct B depend on having already encountered struct A, which may not always be the case.  There might be a relatively elegant solution to this, but IMO it's overly convoluted,  and the code will be hard to read, review and maintain, at least as compared to the idea of just using a deterministic way of generating the `TypedDict` names.

For the time being, I'd prefer to leave the deterministic name generation method in place, and only bring in the state convolution later, and only if necessary.  If you'd _really_ like to see this happen, I'd vote for leaving it out of this PR and creating a follow-on Issue to track it.

However, I agree that the way it's being done now, with these super long names, is quite ugly, but I chose to do it that way anyways, because it was easy and it works.  Thinking further, though, what do you think about using a nicer but still deterministic approach?  I'm envisioning `TypedDict9fdb6c`, where it's just `TypedDict` concatenated with a truncated hash of the method names.  This would provide a short but unique name which could be determined anywhere (declarations, references, nested declarations, etc) without needing to refer back to any name choices made earlier in the generation process.

packages/abi-gen/diff.sh
```diff
@@ -1,4 +1,4 @@
-diff -r $1 $2
+diff -r -U 5 $1 $2
```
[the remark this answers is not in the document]
feuGeneA: Added this to the checklist.

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
feuGeneA: Added this to the checklist.

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
feuGeneA: I didn't address your comment about type aliases.  I'm confused about how they fit in to the generated code.  Do you expect they'll appear in the generated code at all?  Or are they just for the user of the wrapper to use in constructing arguments and deconstructing return values?

feuGeneA:
I decided not to do signature validation because I don't think it belongs in the generated code, and because as I started to add it, to the `contract_wrappers.exchange.validator` module's `ExchangeValidator` class, I realized that it was non-trivial, though probably relatively easy.  I figure that signature validation is something we want to offer contract functionality for, not a native implementation.  If we do want to have native signature validation, I propose we put raise a follow-on Issue.

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
feuGeneA: I went ahead with my proposal of naming type classes with the hash of their components' names.  I introduced type aliases to the same Python module that contains the Exchange wrapper, seen [here](https://github.com/0xProject/0x-monorepo/blob/51d9b45ec44ca14616f4ac6c3f35c1b999275197/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/types.py#L21-L65).  Unfortunately the Exchange API itself still uses weird names (eg `Tuple0xffffff`) for its parameter specification, but the client can write calls in terms of the type aliases, as shown in the module docstring [here](https://github.com/0xProject/0x-monorepo/blob/51d9b45ec44ca14616f4ac6c3f35c1b999275197/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/__init__.py#L120-L174).

feuGeneA:
I added a thing to the Circle CI config to persist the generated Python wrappers as "Build Artifacts".  You can go to the `test-python` task's "Artifacts" tab (["Artifacts" tab on a recent build](https://circleci.com/gh/0xProject/0x-monorepo/79585#artifacts/containers/0)), and drill down the repo tree to find them.  Here are direct links for a recent enough build: [Exchange wrapper](https://79585-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/__init__.py); [ERC20 token wrapper](https://79585-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/erc20_token/__init__.py).

feuGeneA:
I added some CircleCI configuration to persist "Build Artifacts".  With this, you can go to a build job's "Artifacts" tab ([here's a link to the "Artifacts" tab on a recent run of `test-python`](https://circleci.com/gh/0xProject/0x-monorepo/79585#artifacts/containers/0)), and drill down through the repo tree to find persisted artifacts.

I'm using this to persist generated Python wrappers, generated test output from `abi-gen`'s command-line tests, and documentation generated from the Python code.  Here are some direct links (so you don't have to drill down (and know where to drill down into)) from the most recent build:

- [Exchange wrapper](https://79656-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/__init__.py)
- [ERC20 token wrapper](https://79656-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/erc20_token/__init__.py)
- [Documentation for Python module `zero_ex.contract_wrappers`](https://79656-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/build/docs/html/index.html), which includes the generated docs for the `Exchange` wrapper class, as well as for the `ERC20Token` class.

feuGeneA:
@fabioberger @xianny @PirosB3 this PR is pretty big, but not as big as it looks.

It says I added 2,300 new lines of code, but a lot of that is generated: expected outputs for those new abi-gen CLI tests, pre-compiled contract artifacts for abi-gen test fixtures, etc.

And there are 30 commits, but many of them are very small.  There are maybe 2 or 3 commits that are pretty big and hairy, and maybe 3 or 4 medium sized ones, and the rest are small to tiny.

packages/abi-gen/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "3.1.1",
+        "changes": [
+            {
+                "note": "Python method parameters are now in snake case; Python wrappers now support tuples in method parameters; document Python method's bytes params as requiring UTF-8; generate Python output into a contract-named folder, not a file (eg exchange/__init__.py rather than exchange.py) leaving space for user-defined additions to the same module, such as for custom types, as used by the Exchange wrapper's manually-wriiten type aliases in the contract_wrappers.exchange.types Python module; support for customizable parameter validation for Python wrappers; wrap Python docstrings better, for pydocstyle compliance; lots of fixes to satisfy linters of generated Python code.",
```
fabioberger:
```suggestion
                "note": "Python method parameters are now in snake case; Python wrappers now support tuples in method parameters; document Python method's bytes params as requiring UTF-8; generate Python output into a contract-named folder, not a file (eg exchange/__init__.py rather than exchange.py) leaving space for user-defined additions to the same module, such as for custom types, as used by the Exchange wrapper's manually-written type aliases in the contract_wrappers.exchange.types Python module; support for customizable parameter validation for Python wrappers; wrap Python docstrings better, for pydocstyle compliance; lots of fixes to satisfy linters of generated Python code.",
```

packages/abi-gen/src/index.ts
```diff
@@ -116,12 +126,111 @@ function registerPythonHelpers(): void {
...
+                    `:param tx_hash: hash of transaction emitting ${eventName} event`,
+                    indent,
+                ),
+            ),
+    );
+    Handlebars.registerHelper('tupleDefinitions', (abisJSON: string, options) => {
+        const abis: AbiDefinition[] = JSON.parse(abisJSON);
+        // build an array of objects, each of which has one key, the Python
+        // name of a tuple, with a string value holding the Python
+        // definition of that tuple.  using a key-value object conveniently
```
fabioberger:
```suggestion
        // definition of that tuple. Using a key-value object conveniently
```

packages/abi-gen/src/index.ts
```diff
@@ -116,12 +126,111 @@ function registerPythonHelpers(): void {
...
+        // filters duplicate references to the same tuple.
+        const tupleDefinitions: { [pythonTupleName: string]: string } = {};
```
fabioberger: Nit: Let's use type `ObjectMap<string>` here. It's in `@0x/types`

python-packages/order_utils/CHANGELOG.md
```diff
@@ -10,3 +10,7 @@
 -   Deprecated methods `encode_erc20_asset_data()` and `encode_erc721_asset_data()`, in favor of new methods `encode_erc20()` and `encode_erc721()`. The old methods return a string, which is less than convenient for building orders using the provided `Order` type, which expects asset data to be `bytes`. The new methods return `bytes`.
 -   Expanded documentation.
 -   Stopped using deprecated web3.py interface `contract.call()` in favor of `contract.functions.X.call()`. This provides compatibility with the upcoming 5.x release of web3.py, and it also eliminates some runtime warning messages.
+
+## 3.0.0 - TBD
+
+-   Major breaking changes: removal of definitions for Order, OrderInfo, order_to_jsdict, jsdict_to_order, all of which have been moved to contract_wrappers.exchange.types; removal of signature validation.
```
fabioberger: Nit: Can we invert the ordering of the CHANGELOG so that the latest version is at the top?

.circleci/config.yml
```diff
@@ -234,6 +239,24 @@ jobs:
...
+            - store_artifacts:
+                  path: ~/repo/python-packages/contract_wrappers/build
+            - store_artifacts:
+                  path: ~/repo/python-packages/json_schemas/build
+            - store_artifacts:
+                  path: ~/repo/python-packages/middlewares/build
+            - store_artifacts:
+                  path: ~/repo/python-packages/order_utils/build
+            - store_artifacts:
+                  path: ~/repo/python-packages/sra_client/build
```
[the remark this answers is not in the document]
feuGeneA: It's not for speeding up, it's for persisting output files (ie generated code) produced by the build.  I explained it in [this comment](https://github.com/0xProject/0x-monorepo/pull/1919#issuecomment-512642261).

.gitignore
```diff
@@ -115,6 +115,8 @@ contracts/extensions/generated-wrappers/
 contracts/exchange-forwarder/generated-wrappers/
 contracts/dev-utils/generated-wrappers/
 packages/metacoin/src/contract_wrappers
+python-packages/contract_wrappers/src/zero_ex/contract_wrappers/erc20_token/__init__.py
+python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/__init__.py
```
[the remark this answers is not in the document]
feuGeneA: No, we're not.  TypeScript build runs first, including `packages/python-contract-wrappers`, which emits the initial contract wrappers.  After TypeScript, the `python-packages` build runs, and that includes a new step "pre_install" (for packages where it's appropriate, namely `contract_wrappers`, `json_schemas`, and `contract_artifacts`) whose purpose is to pull over files from the TypeScript build area.  After that the Python package publication process marches on with the folders found in the local source tree (which at that time includes the generated code).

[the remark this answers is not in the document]
feuGeneA: Also note that, as explained in [this comment](https://github.com/0xProject/0x-monorepo/pull/1919#issuecomment-512642261), you can take a look at the generated wrappers via the CircleCI Build Artifacts.

packages/abi-gen-templates/Python/contract.handlebars
```diff
@@ -11,13 +20,55 @@ from zero_ex.contract_wrappers._base_contract_wrapper import BaseContractWrapper
 from zero_ex.contract_wrappers.tx_params import TxParams
 
 
+class {{contractName}}ValidatorBase:
```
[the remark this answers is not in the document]
feuGeneA: Well, I _could_ do that, but...
-   that would mean a not insignificant change to `abi-gen`, to start outputting more than one file, and I'm feeling like I'd rather see this PR merge and follow up later as necessary; but also
-   it's not uncommon in Python land to have multiple classes in a file, _especially_ in an `__init__.py` file, which represents the "root" of the Python module named after its parent folder.  we're now writing the wrapper to `exchange/__init__.py`, not `exchange.py`, and people generally know that if you're looking for the definition of a class, and you don't see a file for it, it must be in `__init__.py`.

Maybe @PirosB3 can weigh in on my assessment of the Pythonic style here...

packages/abi-gen-templates/Python/contract.handlebars
```diff
@@ -11,13 +20,55 @@ from zero_ex.contract_wrappers._base_contract_wrapper import BaseContractWrapper
...
 class {{contractName}}(BaseContractWrapper):
-    """Wrapper class for {{contractName}} Solidity contract."""
+    """Wrapper class for {{contractName}} Solidity contract.{{docBytesIfNecessary ABIString}}"""
 
     def __init__(
         self,
         provider: BaseProvider,
         contract_address: str,
+        validator: {{contractName}}Validator = None,
         private_key: str = None,
```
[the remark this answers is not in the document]
feuGeneA: Nope. Why get rid of it?  You realize it's optional, right?  Used only when you want to sign locally and send raw.

packages/abi-gen/src/index.ts
feuGeneA: Honestly I prefer this way because

1. you can see everything right here (don't have to go look up what an `ObjectMap` is),
2. `ObjectMap` itself is a trivial enough to just inline here without even adding another line, and
3. I think having the index signature name visible here, `pythonTupleName`, really helps improve the readability of this somewhat convoluted code.

python-packages/order_utils/CHANGELOG.md
feuGeneA: https://github.com/0xProject/0x-monorepo/pull/1919/commits/2a5a27f51df1c6ec0ea864e86fa462e3e00a61bf

packages/abi-gen-templates/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.3.1",
+        "changes": [
+            {
+                "note": "Python: fix broken event handling; custom validator class support; python linter fixes; normalize bytes parameters in wrapper methods",
```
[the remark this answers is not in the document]
feuGeneA: https://github.com/0xProject/0x-monorepo/pull/1919/commits/0eb65cc7667ab4e4eebc6e312c0f4da1834f39fc

packages/abi-gen-templates/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "2.3.1",
```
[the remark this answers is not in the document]
feuGeneA: https://github.com/0xProject/0x-monorepo/pull/1919/commits/f204f645815ec701ba4a925e586879bdd5726524

packages/abi-gen-templates/Python/contract.handlebars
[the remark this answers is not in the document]
feuGeneA: Oh you mean it's own _handlebars_ file, I mis-read, sorry.  I think my logic still applies though, in that I think it makes sense to structure the template files analogous to the generated files.  In TypeScript there's a conventional 1:1 mapping between files and classes, but it may not be that way in every target language environment, and it's not in Python.  I think it makes sense to structure the templates analogous to the files they will produce.  (I don't feel _that_ strongly about it, but that's my logic.)

feuGeneA:
Latest Build Artifacts:
- [Generated Python Exchange wrapper](https://79916-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/exchange/__init__.py)
- [`contract_wrappers.py` generated documentation](https://79916-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/build/docs/html/index.html) (includes docs for the Exchange wrapper class)
- [Generated Python ERC20Token wrapper](https://79916-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/erc20_token/__init__.py)

(see [my previous comment](https://github.com/0xProject/0x-monorepo/pull/1919#issuecomment-512642261) for an explanation of these Build Artifacts)

packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy.py
fabioberger: Re: your second paragraph question -- it's only for users to construct the arguments and deconstructing return values.

fabioberger: I'm happy with this solution! Thanks

packages/abi-gen-templates/Python/contract.handlebars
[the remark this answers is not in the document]
fabioberger: Ok, I guess we can leave it. Just feels weird to have 2 ways of doing the same thing (sign stuff), but it is also much easier to pass a PK in then setup a provider with a PK middleware, so there is value here. Let's leave it for now.

fabioberger:
Looking great @feuGeneA! 🎉

feuGeneA:
Bug found in docs:
image
This method should have a return return type of `Union[None, HexBytes, bytes]`, since if it's a call it can return the contract method's underlying return value (which here is `None`) but also if it's a sendTransaction it can return a transaction hash.

It's the wrapper code itself that has the wrong type, not just the doc.

PR #1930, opened 2019-07-09
Move CLI tests for abi-gen

### LINKED ISSUE #1925: Move command-line interface tests to their own folder
## Expected Behavior

Tests of code-generating command-line utilities should house their unit tests separate from their command-line tests.

## Current Behavior

Our tests of the command-line interface to `abi-gen` are currently mixed in with the same folder that contains the unit tests.

## Possible Solution

Move CLI tests from `abi-gen/test` to `abi-gen/test-cli`

While in there, rename things to make them more intuitive.  I suggest:
`abi-gen/test-cli/generated-output` -- freshly written to with each test run
and
`abi-gen/test-cli/known-good-output` -- checked in and compared against for each test run

## Context

Testing the command-line interface to `abi-gen` involves verifying it's file output, and to do that involves having lots of files.  Having all that noise in the filesystem muddies the waters when just trying to work with the unit tests, so the two types of tests should be in separate folders.

Further, we want to actually build and run the generated code, but to do that requires a separate TypeScript configuration than the configuration used to build the main package-level one.  Specifically, the package-level config needs to exclude building the test-generated code, and the test-level config needs to exclude building the package source and unit tests.

### LINKED ISSUE #1926: Tests of generated wrappers are silently failing to even run
## Expected Behavior

Tests of `abi-gen` should include generating TypeScript wrappers and building and running those tests.

## Current Behavior

Tests of `abi-gen` are generating TypeScript wrappers, and are verifying that output against a known-good baseline, but they are not building and executing the wrappers.  The build seemed to be clean, but was emitting the message shown in the last line here:

image

This was happening on the development branch, but the building and testing of those wrappers was commented out in #1878 .

xianny:
## Description

All CLI-related tests now live in `test-cli` folder. There are separate subfolders for `python` and `typescript`, which will hopefully make it easier to add language-specific tests down the road. 

`yarn clean`, `yarn build`, and `yarn test` will now clean, build, and test the abi-gen utility AND the related CLI tests. 

```
packages/abi-gen
├── test
│   └── utils_test.ts
├── test-cli
│   ├── expected-output
│   │   ├── python
│   │   │   ├── abi_gen_dummy.py
│   │   │   ├── lib_dummy.py
│   │   │   └── test_lib_dummy.py
│   │   └── typescript
│   │       ├── abi_gen_dummy.ts
│   │       ├── lib_dummy.ts
│   │       └── test_lib_dummy.ts
│   ├── fixtures
│   │   ├── artifacts
│   │   │   ├── AbiGenDummy.json
│   │   │   ├── LibDummy.json
│   │   │   └── TestLibDummy.json
│   │   └── contracts
│   │       ├── AbiGenDummy.sol
│   │       ├── LibDummy.sol
│   │       └── TestLibDummy.sol
│   ├── output (generated files, gitignored)
│   │   ├── python
│   │   │   ├── abi_gen_dummy.py
│   │   │   ├── lib_dummy.py
│   │   │   └── test_lib_dummy.py
│   │   └── typescript
│   │       ├── abi_gen_dummy.ts
│   │       ├── lib_dummy.ts
│   │       └── test_lib_dummy.ts
│   ├── test_typescript (unit tests for generated typescript)
│   │   ├── lib (tsc build output)
│   │   ├── src
│   │   │   ├── artifacts.ts
│   │   │   ├── index.ts
│   │   │   └── wrappers.ts
│   │   └── test
│   │       └── abi_gen_dummy_test.ts
│   └── tsconfig.json
├── tsconfig.json
└── tslint.json

```
Resolves #1925, Resolves #1926

## Testing instructions

The best way to test it is to check it out and run the workflow yourself. Would appreciate feedback on comprehensibility and ease of use. 

Make sure to clean first. 

```
yarn clean && yarn build && yarn test
```

## Types of changes

## Checklist:


packages/abi-gen/.gitignore
```diff
@@ -0,0 +1,3 @@
+test-cli/artifacts
+test-cli/python/generated-wrappers
+test-cli/typescript/generated-wrappers
```
feuGeneA: Why not put this stuff in the top-level `.gitignore`? (with the `/packages/abi-gen/` prefix of course).

packages/abi-gen/README.md
```diff
@@ -102,3 +102,23 @@ yarn clean
 ```bash
 yarn lint
 ```
+### CLI tests
+
+The files in `test-cli/` are used to test the CLI output against a set of dummy contracts. 
+
+Compile dummy contracts and generate wrappers:
```
feuGeneA: Can we scrub out any new mentions of "dummy"?  Right now it's true that all we have is "a dummy contract," but we're planning on expanding the test cases in there.  If we'd like to have a word/phrase that describes the contract(s) as "not real," might I suggest "test fixture" or simply "fixture."

Going further, and adding on to my top-level comment (trying to choose a folder hierarchy that can be applied to any package providing a command-line utility), I suggest we keep all the contracts and artifacts (and maybe even tsconfig.json?) under a folder called `test-cli/fixtures`.

packages/abi-gen/package.json
```diff
@@ -9,38 +9,40 @@
...
+        "test_cli:prebuild": "yarn compile:sol && yarn generate_contract_wrappers",
+        "test_cli:build": "tsc --project test-cli/tsconfig.json",
+        "test_cli": "yarn test_cli:run_mocha && yarn diff_contract_wrappers",
+        "build:ci": "yarn build && yarn test_cli:prebuild && yarn test_cli:build",
+        "test": "yarn run_mocha",
         "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/*_test.js --timeout 100000 --bail --exit",
-        "run_contract_wrapper_tests": "mocha --require source-map-support/register --require make-promises-safe test/generated-test/output/TypeScript/lib/**/*_test.js --timeout 100000 --bail --exit",
-        "test:circleci": "yarn test:coverage",
+        "test_cli:run_mocha": "mocha --require source-map-support/register --require make-promises-safe test-cli/typescript/lib/**/*_test.js --timeout 100000 --bail --exit",
+        "test:circleci": "yarn test:coverage && yarn test_cli",
```
feuGeneA: why not also integrate with `yarn test`? this looks like a developer trap: someone can easily forget to run `yarn test_cli` locally, and only find out there are problems when CI runs.

yarn.lock
```diff
@@ -677,12 +677,6 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
-"@babel/runtime@^7.3.1":
-  version "7.4.5"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.4.5.tgz#582bb531f5f9dc67d2fcb682979894f75e253f12"
-  dependencies:
```
feuGeneA: You'd think I'd know better by now, but `yarn.lock` changes still intimidate me.  Maybe you can enlighten me here.  Why are you checking this in if you didn't change any dependencies anywhere?

feuGeneA:
If the pre-build step is time consuming, I suggest just getting rid of it and checking in the pre-generated artifacts.  I know I just added it in there :) but it's really not necessary.  We're testing `abi-gen`, not `sol-compiler`, so I think it's safe to just skip ever running the compiler.  Thinking more, I guess we should keep it in the `package.json`, so that it can be run when updating the test fixture contracts, but I guess my suggestion is just to remove it from the typical build-and-test loop.

feuGeneA:
> If the pre-build step is time consuming, I suggest just getting rid of it ...

Just saw that pre-build includes `generate_contract_wrappers`, and wanted to clarify that by "getting rid of pre-build" I just mean getting rid of the `sol-compiler` invocation.

packages/abi-gen/README.md
xianny: tsconfig.json can only reference files in subdirectories so it has to stay in the highest level directory of all files it touches. 

But I agree that we can check in the artifacts and move them to a fixtures folder.

yarn.lock
xianny: I removed `sleep` because it wasn't being used anywhere and ran `yarn install` to update the lockfile. Possibly these are deps of `sleep`?

packages/abi-gen/package.json
xianny: Could do that. Should we also integrate `test_cli:build` into `yarn build`?

packages/abi-gen/.gitignore
xianny: I initially thought it was easier to track it here since these are all specific to this project only. At this point though the `output` vs `expected-output` idea makes a lot of sense, and that would also make sense to add to the top-level gitignore to start a convention of CLI test folder names.

Changing.

packages/abi-gen/package.json
feuGeneA: Yup.

packages/abi-gen/README.md
```diff
@@ -125,4 +125,4 @@ Run unit tests and check diffs of generated wrappers vs known wrappers:
 yarn test_cli
 ```
 
-Known wrappers have been previously committed and are kept in `test-cli/{language}/known-wrappers`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/{language}/generated-wrappers`, and then manually copy it to `known-wrappers`.
+Known wrappers have been previously committed and are kept in `test-cli/{language}/expected-output`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/{language}/output`, and then manually copy it to `expected-output`.
```
feuGeneA: Nit: Change "Known wrappers..." to "Known-good wrappers..."

packages/abi-gen/diff.sh
```diff
@@ -1,5 +1,5 @@
 diff $1 $2
 if [ $? -ne 0 ]; then
-    echo "ERROR: Freshly generated code does not match known-good, commited output.  If you're confident that the freshly generated code should be considered known-good, then copy it into there and commit it."
+    echo "ERROR: Freshly generated code output does not match expected output.  If you're confident that the expected output should be updated, copy it in there and commit it."
```
feuGeneA: Nit: remove word "code".  This same file/message should theoretically be able to be copied without change into another package where we want to establish some tests of a command-line interface.

packages/abi-gen/package.json
```diff
@@ -11,38 +11,37 @@
...
+        "clean": "shx rm -rf lib test-cli/**/output",
+        "build": "tsc -b && yarn generate_contract_wrappers && yarn test_cli:build",
+        "build:ci": "yarn build",
+        "test": "yarn run_mocha && yarn test_cli",
         "test:circleci": "yarn test:coverage && yarn test_cli",
+        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/*_test.js --timeout 100000 --bail --exit",
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
         "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
+        "test_cli": "yarn test_cli:run_mocha && yarn diff_contract_wrappers",
+        "test_cli:clean": "yarn clean:sol && rm -rf test-cli/**/output && rm -rf test-cli/typescript/lib",
```
feuGeneA: Remove `yarn clean:sol` from here.  The way it is right now, running this will result in a git status of missing files, which will be very confusing to someone new to this code.

packages/abi-gen/package.json
```diff
@@ -11,38 +11,37 @@
...
-        "build": "tsc -b",
-        "test_cli:clean": "yarn clean:sol && rm -rf test-cli/**/generated-wrappers && rm -rf test-cli/typescript/lib",
-        "test_cli:prebuild": "yarn compile:sol && yarn generate_contract_wrappers",
-        "test_cli:build": "tsc --project test-cli/tsconfig.json",
-        "test_cli": "yarn test_cli:run_mocha && yarn diff_contract_wrappers",
-        "build:ci": "yarn build && yarn test_cli:prebuild && yarn test_cli:build",
-        "test": "yarn run_mocha",
-        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/*_test.js --timeout 100000 --bail --exit",
-        "test_cli:run_mocha": "mocha --require source-map-support/register --require make-promises-safe test-cli/typescript/lib/**/*_test.js --timeout 100000 --bail --exit",
+        "clean": "shx rm -rf lib test-cli/**/output",
```
feuGeneA: add in `&& yarn test_cli:clean`.  didn't realize it before, but if we integrated with `yarn build` and then we should integrate with `yarn clean` too.

packages/abi-gen/tslint.json
```diff
@@ -1,6 +1,6 @@
 {
     "extends": ["@0x/tslint-config"],
     "linterOptions": {
-        "exclude": ["./test/generated-test/generated-artifacts/**/*", "**/lib/**/*"]
+        "exclude": ["./test-cli/fixtures/**/*", "**/lib/**/*"]
```
feuGeneA: so are we running linters against the generated wrappers? (i don't see them excluded here).  that would be great!!

.prettierignore
```diff
@@ -24,11 +24,11 @@ lib
...
-/packages/abi-gen/test-cli/artifacts/
-/packages/abi-gen/test-cli/typescript/generated-wrappers
-/packages/abi-gen/test-cli/python/generated-wrappers
-/packages/abi-gen/test-cli/typescript/known-wrappers
-/packages/abi-gen/test-cli/python/known-wrappers
+/packages/abi-gen/test-cli/fixtures/artifacts/
+/packages/abi-gen/test-cli/typescript/output
+/packages/abi-gen/test-cli/python/output
+/packages/abi-gen/test-cli/typescript/expected-output
+/packages/abi-gen/test-cli/python/expected-output
```
feuGeneA: I'd like to see `test-cli/output/typescript` rather than `test-cli/typescript/output` (and the analog for the `expected-` and `python` permutations), in order to establish that convention of `test-cli/output`.

packages/abi-gen/tslint.json
xianny: yes we are!

packages/abi-gen/package.json
xianny: good catch

packages/abi-gen/package.json
```diff
@@ -20,21 +20,21 @@
...
+        "gen_typescript": "abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/partials/**/*.handlebars' --output ./test-cli/output/typescript --backend ethers",
+        "gen_python": "abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/Python/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/Python/partials/**/*.handlebars' --output ./test-cli/output/python --language Python",
         "diff_contract_wrappers": "run-p diff_typescript diff_python",
-        "diff_typescript": "./diff.sh ./test-cli/typescript/expected-output ./test-cli/typescript/output",
-        "diff_python": "./diff.sh ./test-cli/python/expected-output ./test-cli/python/output",
+        "diff_typescript": "./diff.sh ./test-cli/expected-output/typescript ./test-cli/output/typescript",
+        "diff_python": "./diff.sh ./test-cli/expected-output/python ./test-cli/output/python",
```
feuGeneA: Not essential, but we had talked about collapsing these 3 `diff` lines to just `"diff_contract_wrappers": "./diff.sh ./test-cli/expected-output ./test-cli/output",`.  Does that still make sense to you?  Or do you think there's a good reason to keep them separate?

xianny: yes that makes sense! I just overlooked it, thanks for catching

PR #1931, opened 2019-07-09
RichErrors in the remainder of utils

jalextowle:
## Description

These changes replace all remaining `require` statements in `contracts/utils/`. Once these changes are merged, there will be no more `revert` or `require` statements in `contracts/utils`!

## Testing instructions

## Types of changes

## Checklist:


contracts/utils/contracts/src/LibAddressArrayRichErrors.sol
```diff
@@ -0,0 +1,46 @@
...
+        uint256 addressArrayEndPtr
+    )
+        internal
+        pure
+    {
+        abi.encodeWithSelector(
+            MISMANAGED_MEMORY_ERROR_SELECTOR,
+            freeMemPtr,
+            addressArrayEndPtr
+        )._rrevert();
```
abandeali1: I don't mind this style of reverting within the error function itself, but why are we using a pattern that is different from all of our other rich reverts? I prefer to be consistent, and it's probably easier to change this than it is to change all of our other rich reverts.

contracts/utils/contracts/src/MixinReentrancyGuardRichErrors.sol
```diff
@@ -0,0 +1,46 @@
...
+        uint256 globalReentrancyCounter,
+        uint256 localReentrancyCounter
+    )
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodeWithSelector(
+            ILLEGAL_REENTRANCY_ERROR_SELECTOR,
+            globalReentrancyCounter,
```
abandeali1: I feel like the counters aren't really useful here, we can just return the hard coded byte array that is equal to `abi.encodePacked(ILLEGAL_REENTRANCY_ERROR_SELECTOR)` (note the selector will change when we remove these params).

contracts/utils/contracts/src/LibRichErrors.sol
```diff
@@ -0,0 +1,58 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+
+library LibRichErrors {
```
abandeali1: Why add this, rather than replacing `RichErrors`?

contracts/utils/contracts/src/MixinReentrancyGuardRichErrors.sol
```diff
@@ -0,0 +1,46 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "./RichErrors.sol";
+
+
+contract MixinReentrancyGuardRichErrors is
```
abandeali1: When do we use the `Mixin` patterns for errors rather than the library pattern?

contracts/utils/test/lib_bytes.ts
```diff
@@ -22,6 +13,8 @@ chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
+const ZERO = new BigNumber(0);
```
abandeali1: `constants.ZERO_AMOUNT` already exists FYI.

contracts/utils/contracts/src/LibAddressArrayRichErrors.sol
jalextowle: The reason for creating the new pattern was to create a nice and reusable pattern for implementing RichErrors in _libraries_. As far as I know (I've greeped in most of the contracts directories for "_rrevert"), we do not use RichErrors in any libraries at the moment, so there wasn't an established style that I saw. 

Now, the reason that we need a new style for libraries is because they cannot inherit from other contracts. This is the pattern that we usually use to get RichErrors, so I needed to use a different mechanism, which was the `using ... for ...` pattern of libraries. I actually think that this is clean enough to consider making it the standard for RichErrors because the library pattern will work on either contracts or libraries.

All of this said, if you know of a way to avoid making a new pattern, I'll happily adopt that

contracts/utils/contracts/src/LibRichErrors.sol
jalextowle: I had to add this so that I could support "library-style inheritance" (not a technical term) via the "using ... for ..." syntax. I tried to increase code reuse by using LibRichErrors to implement all of the functions in RichErrors, but it didn't seem worth it because the functions were the same size.

contracts/utils/contracts/src/MixinReentrancyGuardRichErrors.sol
jalextowle: We use the library pattern when the rich revert occurs in a library like LibBytes. We'll use the `Mixin` pattern for contracts if we decide to stick with it.

contracts/utils/contracts/src/LibBytesRichErrors.sol
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    // bytes4(keccak256("InvalidByteOperationError(uint8,uint256,uint256)"))
+    bytes4 internal constant INVALID_BYTE_OPERATION_ERROR_SELECTOR =
+        0x28006595;
+
+    // solhint-disable func-name-mixedcase
+    function InvalidByteOperationErrorRevert(
+        InvalidByteOperationErrorCodes errorCode,
+        uint256 endpoint,
```
dorothy-zbornak: Is this the right name? I see a couple places where this is actually the start offset. Maybe this should just be a generic `offset`...

contracts/utils/contracts/src/LibRichErrors.sol
dorothy-zbornak: I wonder if we should just have `RichErrors.sol` be an inheritable thin wrapper for this library.

contracts/utils/contracts/test/TestLibBytes.sol
```diff
@@ -22,9 +22,20 @@ import "../src/LibBytes.sol";
...
 
 contract TestLibBytes {
-    
+
     using LibBytes for bytes;
 
+    /// @dev Pops the last byte off of a byte array by modifying its length.
+    /// @param b Byte array that will be modified.
+    /// @return The byte that was popped off.
+    function publicPopLastByteStateful(bytes memory b)
```
dorothy-zbornak: What's the point of this? Relic from your `sendTransactionAsync()` issues?

jalextowle: Yes, this is a relic. I'll remove it.

jalextowle: Because of a bug in the Github UI, this is actually a reply to the above comment.

I think I can remove `RichErrors.sol` now that we are going to switch over to libraries entirely. Do you see anything wrong with that?

contracts/utils/contracts/src/LibBytesRichErrors.sol
jalextowle: I was using `endpoint` in the sense that they are `endpoints` of a line segment. So the start is the beginning endpoint and the end is the ending endpoint. This said, I feel like we've used `offset` in other cases and it's probably less confusing. I'll update

contracts/utils/contracts/test/TestLibBytes.sol
abandeali1: Nope! Let's remove it if we can.

jalextowle:
I had to make a change to make the CI run again (it failed with a permission error), which is why I changed the comment in *MixinMatchOrders*.

PR #1937, opened 2019-07-10
Copy validation methods from ExchangeWrapper to order-utils

### LINKED ISSUE #1873: Move validation methods from Exchange contract wrapper to order-utils
** UPDATED **

Copy the validation methods in exchange contract-wrapper to order-utils. Check if anything similar already exists. These methods were manually added in `@0x/contract-wrappers` in the exchange wrapper, but will not be automatically generated in `abi-gen-wrappers`.  They will eventually be deprecated after v3 launch. 

Methods to copy:
- [ ] `validateFillOrderThrowIfInvalidAsync`
- [ ] `validateMakerTransferThrowIfInvalidAsync`
- [ ]  `validateOrderFillableOrThrowAsync` 

Part of #1911

xianny:
## Description

Validation methods currently in `@0x/contract-wrappers` ExchangeWrapper will not be automatically generated in `@0x/abi-gen-wrappers`. This PR copies the validation behaviour into `@0x/order-utils`. This is a temporary fix and we intend to remove the validation methods after V3 contracts are deployed and provide better revert reasons.

This also changes `ExchangeWrapper.validateMakerTransferThrowIfInvalidAsync ` to call the method defined in `@0x/order-utils`. I kept the existing unit test to ensure functionality did not change, but did not add a unit test to `order-utils` since it would be a more significant change and we have a plan to deprecate the validation methods. 
 
Resolves #1873 

Will need to update dependency in `@0x/contract-wrappers` after the `@0x/order-utils` changes are published. From `"@0x/order-utils": "^8.1.1"` -> `"@0x/order-utils": "^8.2.0"`

## Testing instructions

Unit tests should pass.

## Types of changes

## Checklist:

### DIFF
--- a/packages/order-utils/src/order_validation_utils.ts
+++ b/packages/order-utils/src/order_validation_utils.ts
@@ -1,4 +1,9 @@
-import { ExchangeContract, IAssetProxyContract } from '@0x/abi-gen-wrappers';
+import {
+    ExchangeContract,
+    getContractAddressesForNetworkOrThrow,
+    IAssetProxyContract,
+    NetworkId,
+} from '@0x/abi-gen-wrappers';
 import { ExchangeContractErrs, RevertReason, SignedOrder } from '@0x/types';
 import { BigNumber, providerUtils } from '@0x/utils';
 import { SupportedProvider, ZeroExProvider } from 'ethereum-types';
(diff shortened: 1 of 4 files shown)


packages/order-utils/src/order_validation_utils.ts
```diff
@@ -106,6 +108,51 @@ export class OrderValidationUtils {
...
+     * Validate the transfer from the maker to the taker. This is simulated on-chain
+     * via an eth_call. If this call fails, the asset is currently nontransferable.
+     * @param exchangeTradeEmulator ExchangeTradeEmulator to use
+     * @param signedOrder SignedOrder of interest
+     * @param makerAssetAmount Amount to transfer from the maker
+     * @param takerAddress The address to transfer to, defaults to signedOrder.takerAddress
+     */
+    public static async validateMakerTransferThrowIfInvalidAsync(
+        exchangeTradeEmulator: ExchangeTransferSimulator,
+        exchangeContract: ExchangeContract,
```
fabioberger: Let's not pass the `exchangeContract` in here. That's a bit "heavy-duty". We have the `assetProxy` addresses for each networkId in `@0x/contract-addresses`, let's get them from there instead.

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -106,6 +108,51 @@ export class OrderValidationUtils {
...
+        supportedProvider: SupportedProvider,
+        signedOrder: SignedOrder,
+        makerAssetAmount: BigNumber,
+        takerAddress?: string,
+    ): Promise<void> {
+        const toAddress = takerAddress === undefined ? signedOrder.takerAddress : takerAddress;
+        await exchangeTradeEmulator.transferFromAsync(
```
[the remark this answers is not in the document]
xianny: That's true. I originally added it to be more in line with the other validation methods. I'll remove it

packages/order-utils/src/order_validation_utils.ts
xianny: yeah, I would love to get rid of `exchangeContract`! Do you think we should pass the `networkId` as an argument here? I'm more inclined to passing the `assetProxyAddress` as an arg, since the only use for `networkId` here is to fetch it. Unless there's a better way of getting it in scope?

fabioberger: If we ask them to pass in the `assetProxyAddress`, then they need to decode the `makerAssetData` themselves and figure out which `assetProxyAddress` to pass in. I think `networkId` is more seamless for the developer. We do the magic under-the-hood.

xianny:
Dang. I rebased on development and force pushed to fix `test-python`, but that changed the older commit hash too. I didn't make any changes to the older commit, so it should be OK to review just the new commit.

PR #1940, opened 2019-07-10
[WIP] Renaming asset-buyer to asset-swapper

dave4506:
## Checklist:


dave4506:
We are renaming the package; since the package is still in beta, decided to keep `asset-buyer` for legacy support; once `asset-swapper` becomes prod ready, we will be deprecating `asset-buyer`.

PR #1943, opened 2019-07-11
Copy dutch auction asset data utils to @0x/order-utils

### LINKED ISSUE #1933: Move Dutch Auction util methods to order-utils
Copy util methods from `@0x/contract-wrappers` DutchAuctionWrapper to `order-utils. These won't be automatically generated in `abi-gen-wrappers`.

Methods to move:
- [ ] `decodeDutchAuctionData`
- [ ] `encodeDutchAuctionAssetData`

Part of #1911

xianny:
## Description

Copies to `@0x/order-utils` encoding and decoding methods for Dutch Auction asset data.

Does not remove the methods from `@0x/contract-wrappers` as we will deprecate them wholesale later.

Resolves #1933

## Testing instructions

Unit tests

## Types of changes

## Checklist:


packages/order-utils/test/asset_data_utils_test.ts
```diff
@@ -40,6 +40,15 @@ const KNOWN_MULTI_ASSET_ENCODING = {
...
 };
 
+const KNOWN_DUTCH_AUCTION_ENCODING = {
+    tokenAddress: '0x34d402f14d58e001d8efbe6585051bf9706aa064',
+    assetData: '0xf47261b000000000000000000000000034d402f14d58e001d8efbe6585051bf9706aa064', // ERC20
+    beginTimeSeconds: new BigNumber(1562807905),
+    beginAmount: new BigNumber(5),
+    dutchAuctionAssetData:
+        '0xf47261b000000000000000000000000034d402f14d58e001d8efbe6585051bf9706aa064000000000000000000000000000000000000000000000000000000005d268e610000000000000000000000000000000000000000000000000000000000000005',
```
feuGeneA: Where did this come from?  I looked over in `contract-wrappers` and couldn't find anything analogous.

It would be nice to cite a source in the comments here, either where you copied it from, or how you generated it.  Would be even better to do some round trip encoding/decoding testing.  Up to you how far to take it, given your intended scope with this PR.

packages/order-utils/src/asset_data_utils.ts
```diff
@@ -230,6 +233,59 @@ export const assetDataUtils = {
...
+            dutchAuctionDetailsBuffer,
+        );
+        const beginTimeSeconds = new BigNumber(beginTimeSecondsAsBN.toString());
+        const beginAmount = new BigNumber(beginAmountAsBN.toString());
+        return {
+            assetData,
+            beginTimeSeconds,
+            beginAmount,
+        };
+    },
```
fabioberger: Now that we have the logic in `order-utils`, which is a dependency of `contract-wrappers`, let's have the methods in `contract-wrappers` call these same methods under-the-hood. That way, we still keep it exposed there, but our code remains DRY.

packages/order-utils/test/asset_data_utils_test.ts
xianny: This is from a unit test run using Ganache in `contract-wrappers`. 
 Round trip encoding/decoding sounds useful -- what would it be like in this case? Something other than encoding/decoding the ERC20 asset data + the dutch auction info?

xianny:
Update, I rebased on development and force-pushed to fix `test-python`, that messed up the earlier commit 9e0acd1dbf77a999a5e5dcaa1767b85800c9792f but I didn't make any changes on that commit! It's OK to just review the latest commit.

PR #1949, opened 2019-07-12
Fix case in pre-publish write access checks

dave4506:
## Description

New packages do not exist in registry and therefore would be caught by `checkPublishRequiredSetupAsync` as packages that don't have write access. Added changes for code to not throw an error if packages are not published yet.

### DIFF
--- a/packages/monorepo-scripts/src/prepublish_checks.ts
+++ b/packages/monorepo-scripts/src/prepublish_checks.ts
@@ -148,7 +148,14 @@ async function checkPublishRequiredSetupAsync(updatedPublicPackages: Package[]):
     const writePermissions = Object.keys(pkgPermissions).filter(pkgName => {
         return pkgPermissions[pkgName] === 'read-write';
     });
-    const unwriteablePkgs = updatedPublicPackages.filter(pkg => !writePermissions.includes(pkg.packageJson.name));
+    const unwriteablePkgs = [];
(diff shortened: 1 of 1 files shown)


packages/monorepo-scripts/src/prepublish_checks.ts
```diff
@@ -148,7 +148,15 @@ async function checkPublishRequiredSetupAsync(updatedPublicPackages: Package[]):
     const writePermissions = Object.keys(pkgPermissions).filter(pkgName => {
         return pkgPermissions[pkgName] === 'read-write';
     });
-    const unwriteablePkgs = updatedPublicPackages.filter(pkg => !writePermissions.includes(pkg.packageJson.name));
+    const unwriteablePkgs = [];
+    for (const pkg of updatedPublicPackages) {
+        const isPackagePublished =
+            (await npmUtils.getPackageRegistryJsonIfExistsAsync(pkg.packageJson.name)) !== undefined;
+        const isPackageWritePermissionsGranted = writePermissions.includes(pkg.packageJson.name);
+        if (!isPackageWritePermissionsGranted && isPackagePublished) {
```
fabioberger:
```suggestion
        if (isPackagePublished && !isPackageWritePermissionsGranted) {
```

fabioberger: Tiny nit, sorry!

PR #1961, opened 2019-07-16
Add compiler settings to published artifacts

### LINKED ISSUE #1960: Add compiler settings to published artifacts
## Expected Behavior

Published artifacts right now lack compiler settings. 

Add compiler settings to make it possible to regenerate the artifacts for Etherscan verification.

xianny:
## Description

Resolves #1960

Upon looking closer, the scripts in `development` preserves the `compiler` field already. If we run `yarn artifacts_update` in `@0x/contract-artifacts`, we get the same artifacts currently published but **with the compiler field added**.  It was not included in the initial commit of `@0x/contract-artifacts`. 

This PR updates a test to ensure that publishing v3 artifacts will require the `compiler` field to pass CI. These are the fields that will be required:

Screen Shot 2019-07-15 at 5 57 58 PM

## Testing instructions

Try running `yarn artifacts_update` in `packages/contract-artifacts` and see that the `compiler` field is preserved. 

@fabioberger can you confirm these are the fields we need?

## Types of changes

## Checklist:

### DIFF
--- a/packages/contract-artifacts/test/contract_artifacts_test.ts
+++ b/packages/contract-artifacts/test/contract_artifacts_test.ts
@@ -29,6 +29,10 @@ describe('Contract Artifacts', () => {
         const missingRequiredPropertiesByArtifact: ObjectMap<string[]> = {};
         for (const [artifactName, artifact] of Object.entries(artifacts)) {
             for (const requiredProperty of REQUIRED_PROPERTIES) {
+                // HACK (xianny): Remove after `compiler` field is added in v3.
+                if (requiredProperty === 'compiler' && artifact.schemaVersion === '2.0.0') {
+                    continue;
+                }
                 const requiredValue = get(artifact, requiredProperty);
                 if (requiredValue === undefined || requiredValue === '') {
                     const previousMissing = missingRequiredPropertiesByArtifact[artifactName];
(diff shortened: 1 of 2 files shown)


packages/contract-artifacts/test/contract_artifacts_test.ts
```diff
@@ -29,6 +29,10 @@ describe('Contract Artifacts', () => {
         const missingRequiredPropertiesByArtifact: ObjectMap<string[]> = {};
         for (const [artifactName, artifact] of Object.entries(artifacts)) {
             for (const requiredProperty of REQUIRED_PROPERTIES) {
+                // HACK (xianny): Remove after `compiler` field is added in v3.
```
fabioberger:
```suggestion
                // HACK(xianny): Remove after `compiler` field is added in v3.
```

fabioberger: Can you elaborate on when exactly we will remove this check? Also, `schemaVersion` is the version of the artifact schema. So since we're adding `compiler`, shouldn't be bump the schema version to `2.1.0`?

xianny: discussed 🗣 and agreed that we will keep schemaVersion the same because we haven't changed any compiler output fields. These fields are only changed for publishing on npm so we will bump the npm version when the artifacts are updated next time.

Issue #1945, opened 2019-07-11
subproviders/private_key_wallet.ts: Checksummed addresses can't be used in input to signTransactionAsync


bstchow:
This is a similar issue to #1247.

## Expected Behavior
When calling `signTransactionAsync`, any address inputs should be treated the same regardless of casing.

## Current Behavior
Calling `signTransactionAsync` with any input `txParams` where `txParams.from` is assigned a checksummed address or an address with any casing applied, is rejected.

```
UnhandledPromiseRejectionWarning: Error: Requested to sign transaction with address: 0xf17650d33a3Eeb099b22a7cC53855659B947211D, instantiated with address: 0xf17650d33a3eeb099b22a7cc53855659b947211d
```

## Possible Solution
Downcase addresses before comparisons.

Change line 53 of subproviders/src/subproviders/private_key_wallet.ts:
```
if (txParams.from !== undefined && txParams.from.toLowerCase() !== this._address.toLowerCase()) {
```

## Steps to Reproduce (for bugs)

1. Instantiate a PrivateKeyWallet with any private key
2. Call `signTransactionAsync` with input `txParams` where `txParams.from` is assigned a checksummed address, or an address with any casing applied.

## Your Environment

| Package | Version |
| ----- | ----- |
| `@0x/subproviders` | ^2.1.4 |

| Network |
| ------- |
| rinkeby |

dekz:
@bstchow would be happy to merge in a PR if you'd like to quickly create one to get this into the next release ASAP.

Quick question, how are you getting the checksummed address in the first place? Most of our tooling which fetch addresses return the address in lowercase. This is required for hashing contents where a checksummed address will return a different hash

bstchow:
@dekz I'll try to get a PR up ASAP. As far as context for the checksummed address, I believe it's a result of `private_key_wallet.ts`'s `signTransactionAsync` being called from `signTransactionAsync` in `mnemonic_wallet.ts`

PR #1962, opened 2019-07-16
[Issue #1945] Downcase address comparison in private_key_wallet.ts in subproviders module.

bstchow:
## Description
Downcase the initialized address of the private key wallet subprovider before comparing it with a signed transaction's origin address before comparing the two addresses to check for equality.

## Testing instructions
1) Instantiate a PrivateKeyWallet with any private key
2) Call signTransactionAsync with input txParams where txParams.from is assigned a checksummed address which matches the private key provided the PrivateKeyWallet.
3) The transaction should complete without any errors.

## Types of changes
Fixes defective behavior described in Issue #1945 as per direction from @dekz.

## Checklist:


packages/subproviders/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "4.1.2",
+        "changes": [
+            {
+                "note": "Fix bug in Private Key subprovider causing checksummed tx.origin addresses to be rejected.",
+                "pr": 1962
+            }
+        ],
+        "timestamp": 1563240294
```
dekz: can you delete this line, this is auto created when published.

PR #1963, opened 2019-07-16
Implement ERC1155Proxy and StaticCallProxy in Solidity

abandeali1:
## Description

- Implements the `ERC1155Proxy` and `StaticCallProxy` in Solidity
- Note: `StaticCallProxy` cannot inherit `IAssetProxy` because its `transferFrom` function is marked as `view`.
- Many revert conditions no longer contain reasons, as they are handled within Solidity
- There are no longer any checks for if the length of `assetData` is a multiple of 32. However, any lengths less than required should fail (TODO: add tests for this)
- Found some pretty bizarre behavior is Solidity's `abi.decode`. When multiple offsets point to the same location in calldata (i.e if `values` == `ids`) then the decoded output is also copied to the same location in memory. This means we cannot modify `values` directly, or we may end up modifying `ids` or `data` as well.
- Thoughts on using `slice` vs `sliceDestructive` here? I opted for `sliceDestructive` because `assetData` is not reused, it is cheaper, and the implementation is arguably simpler.

TODO: gas benchmarking

## Checklist:

### DIFF
--- a/contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
+++ b/contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
@@ -18,349 +18,71 @@
 
 pragma solidity ^0.5.9;
 
+import "@0x/contracts-utils/contracts/src/LibBytes.sol";
+import "@0x/contracts-utils/contracts/src/SafeMath.sol";
+import "@0x/contracts-erc1155/contracts/src/interfaces/IERC1155.sol";
 import "./MixinAuthorizable.sol";
+import "./interfaces/IAssetProxy.sol";
 
 
 contract ERC1155Proxy is
-    MixinAuthorizable
+    MixinAuthorizable,
+    SafeMath,
+    IAssetProxy
 {
+    using LibBytes for bytes;
 
     // Id of this proxy.
     bytes4 constant internal PROXY_ID = bytes4(keccak256("ERC1155Assets(address,uint256[],uint256[],bytes)"));
 
-    // solhint-disable-next-line payable-fallback
-    function () 
+    /// @dev Transfers batch of ERC1155 assets. Either succeeds or throws.
+    /// @param assetData Byte array encoded with ERC1155 token address, array of ids, array of values, and callback data.
+    /// @param from Address to transfer assets from.
+    /// @param to Address to transfer assets to.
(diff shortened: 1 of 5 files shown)


contracts/asset-proxy/contracts/src/ERC1155Proxy.sol
```diff
@@ -18,349 +18,71 @@
...
-            revert(0, 0)
+        // Decode params from `assetData`
+        // solhint-disable indent
+        (
+            address erc1155TokenAddress,
+            uint256[] memory ids,
+            uint256[] memory values,
+            bytes memory data
+        ) = abi.decode(
+            assetData.sliceDestructive(4, assetData.length),
```
dorothy-zbornak: I'm surprised this works with `assetData` being `calldata` and not `memory`. Is it cloning `assetData` to memory at the time of this call?

abandeali1: I believe calldata gets copied to memory when you pass it into any internal function (`sliceDestructive` in this case). That might not be implemented for things that require ABIv2 encoding, though (or maybe just structs?).

contracts/asset-proxy/contracts/src/StaticCallProxy.sol
```diff
@@ -18,181 +18,57 @@
...
+            address staticCallTarget,
+            bytes memory staticCallData,
+            bytes32 expectedReturnDataHash
+        ) = abi.decode(
+            assetData.sliceDestructive(4, assetData.length),
+            (address, bytes, bytes32)
+        );
+
+        // Execute staticcall
+        (bool success, bytes memory returnData) = staticCallTarget.staticcall(staticCallData);
```
dorothy-zbornak: Should we be doing an `extcodesize()` check to safeguard people from accidentally calling EOAs? Could be a problem if people expect a `keccak(bytes(0))` result.

At the very least, we should add a test that confirms that calling an EOA results in `keccak(bytes(0))`.

abandeali1: I will add a test for this, but it feels a bit excessive to me to add an `extcodesize` or `extcodehash` check here. You really have to know the interface of and trust the `staticCallTarget` beforehand in order to use this.

PR #1964, opened 2019-07-16
Update MAP + add validation to assetDataUtils

abandeali1:
## Description

This includes the MAP code that is currently deployed to mainnet.

## Checklist:


contracts/asset-proxy/contracts/src/MultiAssetProxy.sol
```diff
@@ -90,7 +90,10 @@ contract MultiAssetProxy is
                 //          offset to assetData.
 
                 // Load offset to `assetData`
-                let assetDataOffset := calldataload(4)
+                let assetDataOffset := add(calldataload(4), 4)
```
jalextowle: This was a good call. It really cleaned up the buffer arithmetic 👍

contracts/asset-proxy/contracts/src/MultiAssetProxy.sol
```diff
@@ -204,15 +228,20 @@ contract MultiAssetProxy is
...
                     // + nestedAssetDataElementOffset
                     // + 32 (nestedAssetDataElement len)
-                    let nestedAssetDataElementContentsStart := add(assetDataOffset, add(nestedAssetDataOffset, add(nestedAssetDataElementOffset, 104)))
+                    let nestedAssetDataElementContentsStart := add(
+                        assetDataOffset,
+                        add(
+                            nestedAssetDataOffset,
+                            add(nestedAssetDataElementOffset, 100)
+                        )
+                    )
```
jalextowle: It seems like there is a gas efficiency reason for not asserting that `nestedAssetDataContentsStart + nestedAssetDataLen <= assetDataEnd`, but I wonder if it would be worth it to include it just to ensure that the nestedAssetData is in 1:1 correspondence with the data that is actually sent.

abandeali1: In general I'm all for extra validation, but I primarily want to get this exact version merged since it is the one deployed to mainnet.

PR #1966, opened 2019-07-17
Added options, features for asset-swapper

dave4506:
## Description

Added the following: 
- Option to disable fee order logic
- Option to default swap quote consumer to a certain `ConsumerType`
- Added a utils library for developers to check allowances for taker
- Optimized `SwapQuoteConsumer` output (deduped data for lower gas consumption)

## Checklist:


packages/asset-swapper/src/types.ts
```diff
@@ -262,6 +272,7 @@ export interface SwapQuoteInfo {
  */
```
fragosti: You forgot to add doc strings.

packages/asset-swapper/src/types.ts
```diff
@@ -198,6 +207,7 @@ export type SwapQuote = MarketBuySwapQuote | MarketSellSwapQuote;
...
```
fragosti: Are these docstrings correct?

packages/asset-swapper/src/utils/swap_quote_utils.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const provider = providerUtils.standardizeOrThrow(supportedProvider);
+        this.provider = provider;
+        this.networkId = networkId;
+        this._contractWrappers = new ContractWrappers(this.provider, {
+            networkId,
+        });
+    }
+
+    public async isTakerAddressAllowanceEnoughForBestAndWorstQuoteInfoAsync(
```
fragosti: Should we remove this class and just put this util on `SwapQuoter`?

packages/asset-swapper/test/swap_quote_calculator_test.ts
```diff
@@ -78,6 +78,7 @@ describe('swapQuoteCalculator', () => {
                         new BigNumber(200),
                         0,
                         false,
+                        false,
```
fragosti: Tests for the true condition would be nice.

dave4506: Yeah, will be adding coverage in a seperate PR

PR #1975, opened 2019-07-20
Use less strict validation in `cancelOrder`, remove `cancelOrderNoThrow`

abandeali1:
## Description

In the interest of simplifying the Exchange contract interface, this PR uses less strict validation in `cancelOrder` and removes `cancelOrderNoThrow`. `cancelOrder` now returns early if the order is not `FILLABLE`, rather than reverting.

## Checklist:

### DIFF
--- a/contracts/exchange/contracts/src/MixinExchangeCore.sol
+++ b/contracts/exchange/contracts/src/MixinExchangeCore.sol
@@ -256,6 +256,11 @@ contract MixinExchangeCore is
         // Validate context
         _assertValidCancel(order, orderInfo);
 
+        // Noop if order is already unfillable
+        if (orderInfo.orderStatus != OrderStatus.FILLABLE) {
+            return;
+        }
+
         // Perform cancel
         _updateCancelledState(order, orderInfo.orderHash);
     }
@@ -466,15 +471,6 @@ contract MixinExchangeCore is
         internal
         view
     {
-        // Ensure order is valid
-        // An order can only be cancelled if its status is FILLABLE.
-        if (orderInfo.orderStatus != uint8(OrderStatus.FILLABLE)) {
-            LibRichErrors._rrevert(LibExchangeRichErrors.OrderStatusError(
-                orderInfo.orderHash,
-                OrderStatus(orderInfo.orderStatus)
-            ));
-        }
-
         // Validate sender is allowed to cancel this order
         if (order.senderAddress != address(0)) {
             if (order.senderAddress != msg.sender) {
--- a/contracts/exchange/contracts/src/MixinExchangeCore.sol
+++ b/contracts/exchange/contracts/src/MixinExchangeCore.sol
@@ -257,7 +257,7 @@ contract MixinExchangeCore is
         _assertValidCancel(order, orderInfo);
 
         // Noop if order is already unfillable
-        if (orderInfo.orderStatus != OrderStatus.FILLABLE) {
+        if (orderInfo.orderStatus != uint8(OrderStatus.FILLABLE)) {
             return;
         }
 
--- a/contracts/exchange/contracts/src/MixinWrapperFunctions.sol
+++ b/contracts/exchange/contracts/src/MixinWrapperFunctions.sol
@@ -371,19 +371,6 @@ contract MixinWrapperFunctions is
         return fillResults;
     }
 
-    /// @dev After calling, the order can not be filled anymore.
-    //       Returns false if the cancelOrder call would otherwise revert.
-    /// @param order Order to cancel. Order must be OrderStatus.FILLABLE.
-    /// @return True if the order was cancelled successfully.
-    function cancelOrderNoThrow(LibOrder.Order memory order)
-        public
-        returns (bool didCancel)
-    {
-        bytes memory cancelOrderCallData = abi.encodeWithSelector(CANCEL_ORDER_SELECTOR, order);
-        (didCancel,) = address(this).delegatecall(cancelOrderCallData);
-        return didCancel;
-    }
-
     /// @dev Executes multiple calls of cancelOrder.
     /// @param orders Array of order specifications.
     function batchCancelOrders(LibOrder.Order[] memory orders)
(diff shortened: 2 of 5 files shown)


contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -466,9 +466,8 @@ contract MixinExchangeCore is
         internal
         view
     {
-        // Ensure order is valid
-        // An order can only be cancelled if its status is FILLABLE.
-        if (orderInfo.orderStatus != uint8(OrderStatus.FILLABLE)) {
+        // Ensure order has not already been cancelled.
+        if (orderInfo.orderStatus == uint8(OrderStatus.CANCELLED)) {
```
hysz: I wonder if we should also revert if the order has been filled 🤔. This could be more aligned with the expectations of a relayer or end-user. Ex, on most exchanges I've traded on, I can't cancel an order once it's filled.

abandeali1: We should either keep `cancelOrderNoThrow` or make cancels deterministic IMO. Otherwise batch cancels become unreliable. Is there any downside to allowing filled orders to be cancelled, other than gas costs?

abandeali1: Also note that `getOrderInfo` will still return `FILLED` for an order that is both filled and cancelled.

contracts/exchange/contracts/src/MixinWrapperFunctions.sol
```diff
@@ -371,19 +371,6 @@ contract MixinWrapperFunctions is
         return fillResults;
     }
 
-    /// @dev After calling, the order can not be filled anymore.
-    //       Returns false if the cancelOrder call would otherwise revert.
-    /// @param order Order to cancel. Order must be OrderStatus.FILLABLE.
-    /// @return True if the order was cancelled successfully.
-    function cancelOrderNoThrow(LibOrder.Order memory order)
```
jalextowle: Why are we getting rid of these? Were they just not used?

contracts/exchange/contracts/src/MixinExchangeCore.sol
jalextowle: It seems like reverting if the order has `expired` would make a lot of sense since a call to `cancel` could take a few blocks to get mined, which would end up costing a lot more gas now.

dorothy-zbornak: On the `FILLED` front, I think that distinction that can be handled by UX.

I think there still is a need for `batchCancelOrdersNoThrow()` because this function throws if the order is already cancelled. How about we just remove this check entirely or do nothing if the order is already cancelled?

contracts/exchange/test/core.ts
```diff
@@ -749,30 +772,20 @@ describe('Exchange core', () => {
...
                 makerAssetAmount: new BigNumber(0),
             });
-            const orderHash = orderHashUtils.getOrderHashHex(signedOrder);
-            const expectedError = new ExchangeRevertErrors.OrderStatusError(
-                orderHash,
-                OrderStatus.InvalidMakerAssetAmount,
-            );
-            const tx = exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
-            return expect(tx).to.revertWith(expectedError);
+            const tx = await exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
```
jalextowle: Will the length of the logs be zero if the tx reverts? If so, we should probably also assert that the tx succeeds.

contracts/exchange/test/core.ts
```diff
@@ -749,30 +772,20 @@ describe('Exchange core', () => {
...
                 takerAssetAmount: new BigNumber(0),
             });
-            const orderHash = orderHashUtils.getOrderHashHex(signedOrder);
-            const expectedError = new ExchangeRevertErrors.OrderStatusError(
-                orderHash,
-                OrderStatus.InvalidTakerAssetAmount,
-            );
-            const tx = exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
-            return expect(tx).to.revertWith(expectedError);
+            const tx = await exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
```
jalextowle: Same as above

contracts/exchange/test/core.ts
```diff
@@ -800,46 +813,19 @@ describe('Exchange core', () => {
...
         });
 
-        it('should revert if already cancelled', async () => {
+        it('should noop if already cancelled', async () => {
             await exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
-            const orderHash = orderHashUtils.getOrderHashHex(signedOrder);
-            const expectedError = new ExchangeRevertErrors.OrderStatusError(orderHash, OrderStatus.Cancelled);
-            const tx = exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
-            return expect(tx).to.revertWith(expectedError);
+            const tx = await exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
```
jalextowle: Same as above

contracts/exchange/test/core.ts
```diff
@@ -800,46 +813,19 @@ describe('Exchange core', () => {
...
-            const expectedError = new LibMathRevertErrors.RoundingError(
-                fillTakerAssetAmount2,
-                new BigNumber(3),
-                new BigNumber(1001),
-            );
-            const tx = exchangeWrapper.fillOrderAsync(signedOrder, takerAddress, {
-                takerAssetFillAmount: fillTakerAssetAmount2,
-            });
-            return expect(tx).to.revertWith(expectedError);
+            const tx = await exchangeWrapper.cancelOrderAsync(signedOrder, makerAddress);
```
jalextowle: Same as above.

contracts/exchange/test/core.ts
abandeali1: The transaction will throw an error if it reverts rather than returning a txReceipt.

jalextowle: `.fulfilled('')`

PR #1987, opened 2019-07-24
Remove Forwarder fee abstraction and revamp tests

mzhu25:
## Description
Previously, fees on Forwarder orders were (a) always denominated in ZRX and (b) represented by a separate array of feeOrders. In v3 we instead use takerFee and takerFeeAssetData, allowing order fees to be denominated in either WETH or the maker asset (percentage fees). This PR updates the exchange-forwarder contracts to implement this change, which also comes with a few notable side-effects:
- The Forwarder contract no longer needs to keep a balance of ZRX, so it also no longer needs to reserve some ETH value to replenish its ZRX balance. 
- As a byproduct of percentage fees, ERC20AssetProxy now needs to be approved to transfer arbitrary maker asset tokens on behalf of the Forwarder. We allow anyone to trigger this approval using approveMakerAssetProxy
- Also as a result of percentage fees, marketBuyOrdersWithEth will potentially spend some portion of makerAssetFillAmount on fees, so the amount transferred to the user will be (makerAssetFillAmount - percentageFeesPaid)
This PR also introduces ForwarderTestFactory, which abstracts away most of the balance-checking logic in the forwarder tests. It's also more conducive for combinatorial testing, although that isn't including in this PR. 

## Testing instructions
`PKG=@0x/contracts-exchange-forwarder yarn test`

## Types of changes

* New feature:
    * ForwarderTestFactory for more concise tests
    * Remove feeOrders abstraction from Forwarder contracts
    * Add logic to instead handle WETH fees and percentage fees on Forwarder orders

## Checklist:


contracts/exchange-forwarder/package.json
```diff
@@ -46,11 +46,19 @@
     },
     "homepage": "https://github.com/0xProject/0x-monorepo/contracts/extensions/README.md",
     "devDependencies": {
+<<<<<<< HEAD
```
jalextowle: This merge conflict should be addressed.

contracts/exchange-forwarder/test/utils/forwarder_test_factory.ts
```diff
@@ -0,0 +1,359 @@
+import { ERC20Wrapper } from '@0x/contracts-asset-proxy';
```
jalextowle: This file is really nice! It cuts down a lot on the logic in the forwarder tests

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -92,45 +92,90 @@ contract MixinExchangeWrapper is
...
-
-            // We assume that asset being bought by taker is the same for each order.
-            // We assume that asset being sold by taker is WETH for each order.
-            orders[i].makerAssetData = makerAssetData;
-            orders[i].takerAssetData = wethAssetData;
 
-            // Calculate the remaining amount of WETH to sell
-            uint256 remainingTakerAssetFillAmount = _safeSub(wethSellAmount, totalFillResults.takerAssetFilledAmount);
+        for (uint256 i = 0; i != ordersLength; i++) {
+            require(
```
jalextowle: We should make this a RichError instead of a normal revert. Let me know if you have questions regarding how to use RichErrors.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -139,31 +184,24 @@ contract MixinExchangeWrapper is
...
+            uint256 makerAssetAcquiredAmount
+        )
     {
-        bytes memory makerAssetData = orders[0].makerAssetData;
-        bytes memory wethAssetData = WETH_ASSET_DATA;
-
         uint256 ordersLength = orders.length;
-        uint256 makerAssetFilledAmount = 0;
         for (uint256 i = 0; i != ordersLength; i++) {
+            require(
```
jalextowle: See above comment ^. You can use the same RichError to replace both of these `require` statements

contracts/exchange-forwarder/contracts/src/MixinAssets.sol
```diff
@@ -50,6 +50,26 @@ contract MixinAssets is
         _transferAssetToSender(assetData, amount);
     }
 
+    /// @dev Approves the respective proxy for a given asset to transfer tokens on the Forwarder contract's behalf.
+    ///      This is necessary because an order fee denominated in the maker asset (i.e. a percentage fee) is sent by the
+    ///      Forwarder contract to the fee recipient.
+    ///      This method needs to be called before forwarding orders of a maker asset that hasn't
+    ///      previously been approved.
+    /// @param assetData Byte array encoded for the respective asset proxy.
+    function approveMakerAssetProxy(bytes calldata assetData)
```
jalextowle: I wonder if it would make sense to have another function that sets a user-specified approval value.

contracts/exchange-forwarder/test/forwarder.ts
```diff
@@ -186,244 +195,89 @@ describe(ContractName.Forwarder, () => {
...
-            expect(newBalances[forwarderContract.address][defaultMakerAssetAddress]).to.be.bignumber.equal(
-                constants.ZERO_AMOUNT,
-            );
-            expect(forwarderEthBalance).to.be.bignumber.equal(constants.ZERO_AMOUNT);
+        it('should fill a single order without a taker fee', async () => {
+            await forwarderTestFactory.marketSellTestAsync([orderWithoutFee], new BigNumber(0.5), erc20Token);
         });
-        it('should fill multiple orders', async () => {
+        it('should fill multiple orders without taker fees', async () => {
             const secondOrderWithoutFee = await orderFactory.newSignedOrderAsync();
```
dorothy-zbornak: I would prefer if these orders had different maker/taker amounts. There are some other places where we generate identical orders, as well.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -92,45 +92,90 @@ contract MixinExchangeWrapper is
...
+                makerAssetAcquiredAmount = _safeAdd(
+                    makerAssetAcquiredAmount,
+                    _safeSub(singleFillResults.makerAssetFilledAmount, singleFillResults.takerFeePaid)
+                );
+
+                wethSpentAmount = _safeAdd(
+                    wethSpentAmount,
+                    singleFillResults.takerAssetFilledAmount
+                );
+            // WETH fee
```
hysz: How do we enforce this assumption? What is the expected behavior if `makerAssetData != takerFeeAssetData && takerFeeAssetData != WETH`.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -172,91 +210,43 @@ contract MixinExchangeWrapper is
...
+            } else {
```
hysz: Same as above.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -172,91 +210,43 @@ contract MixinExchangeWrapper is
...
+                makerAssetAcquiredAmount = _safeAdd(
+                    makerAssetAcquiredAmount,
+                    singleFillResults.makerAssetFilledAmount
+                );
+
+                // WETH is also spent on the taker fee, so we add it here.
+                wethSpentAmount = _safeAdd(
+                    wethSpentAmount,
+                    _safeAdd(singleFillResults.takerAssetFilledAmount, singleFillResults.takerFeePaid)
+                );
```
hysz: Nit - this for loop and the one above are getting quite long with lots of nesting. I think it could improve readability to break these up.

dorothy-zbornak:
We patched the `3.0` branch to to get CI working again. Also, this package is disabled on CI. When you get a chance, plx rebase this branch from `3.0` and make the following change to your `/.circleci/config.yml`. This will give your PR magic CI sauce.

```diff
@@ -90,7 +90,7 @@ jobs:
             - restore_cache:
                   keys:
                       - repo-{{ .Environment.CIRCLE_SHA1 }}
-            - run: yarn wsrun test:circleci @0x/contracts-multisig @0x/contracts-utils @0x/contracts-exchange-libs @0x/contracts-erc20 @0x/contracts-erc721 @0x/contracts-erc1155 @0x/contracts-asset-proxy @0x/contracts-exchange @0x/contracts-coordinator @0x/contracts-dev-utils @0x/contracts-staking
+            - run: yarn wsrun test:circleci @0x/contracts-multisig @0x/contracts-utils @0x/contracts-exchange-libs @0x/contracts-erc20 @0x/contracts-erc721 @0x/contracts-erc1155 @0x/contracts-asset-proxy @0x/contracts-exchange @0x/contracts-coordinator @0x/contracts-dev-utils @0x/contracts-staking @0x/contracts-extensions
             # TODO(dorothy-zbornak): Re-enable after updating this package for 3.0.
             # - run: yarn wsrun test:circleci @0x/contracts-extensions
             # TODO(dorothy-zbornak): Re-enable after updating this package for 3.0.
```

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -92,45 +92,90 @@ contract MixinExchangeWrapper is
         bytes[] memory signatures
     )
         internal
-        returns (FillResults memory totalFillResults)
+        returns (
```
abandeali1: Since these functions no longer require `takerAssetData` to be WETH, let's rip out `weth` from all of our variable and function names. `_marketSellOrders` and `_marketBuyOrders` should be fine.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -92,45 +92,90 @@ contract MixinExchangeWrapper is
...
+            uint256 remainingTakerAssetFillAmount = _safeSub(
+                wethSellAmount,
+                wethSpentAmount
             );
+            FillResults memory singleFillResults;
+
+            // Percentage fee
+            if (orders[i].makerAssetData.equals(orders[i].takerFeeAssetData)) {
+                // Attempt to sell the remaining amount of WETH
+                singleFillResults = _fillOrderNoThrow(
```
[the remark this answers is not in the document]
abandeali1: Discussed offline and I don't think we'll actually be able to ge the desired functionality by using the `Exchange` functions.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -92,45 +92,90 @@ contract MixinExchangeWrapper is
...
-                signatures[i]
+            // The remaining amount of WETH to sell
+            uint256 remainingTakerAssetFillAmount = _safeSub(
+                wethSellAmount,
+                wethSpentAmount
             );
+            FillResults memory singleFillResults;
+
+            // Percentage fee
+            if (orders[i].makerAssetData.equals(orders[i].takerFeeAssetData)) {
```
abandeali1: Shouldnt this check that the order's `takerAssetData` matches the `takerFeeAssetData`, since we don't want to over-sell the `takerAsset`?

abandeali1: I see, this is yet another case. Let's write out all of the different cases and see if we can simplify the logic at all.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
abandeali1: Yeah, I think this logical path should check if `takerAssetData` == `takerFeeAssetData` specifically. We can revert in any other case, as this contract will not be able to pay the fee.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
abandeali1: +1

contracts/exchange-forwarder/contracts/src/MixinForwarderCore.sol
```diff
@@ -47,164 +53,114 @@ contract MixinForwarderCore is
...
+            PERCENTAGE_DENOMINATOR,
+            _safeAdd(PERCENTAGE_DENOMINATOR, feePercentage),
+            msg.value
+        );
+
+        // Spends up to wethSellAmount to fill orders and pay WETH order fees.
+        uint256 wethSpentAmount;
+        uint256 makerAssetAcquiredAmount;
+        (
+            orderFillResults,
```
abandeali1: It looks like we don't actually use `orderFillResults` at all. We can probably simplify the logic in `MixinExchangeWrapper` if we don't even track or return `totalFillResults`.

contracts/exchange-forwarder/contracts/src/interfaces/IForwarderCore.sol
```diff
@@ -25,56 +25,41 @@ import "@0x/contracts-exchange-libs/contracts/src/LibFillResults.sol";
...
         uint256 feePercentage,
         address payable feeRecipient
     )
         public
         payable
-        returns (
-            LibFillResults.FillResults memory orderFillResults,
-            LibFillResults.FillResults memory feeOrderFillResults
-        );
+        returns (LibFillResults.FillResults memory orderFillResults);
```
abandeali1: We should think about what is most useful to return here. It might just be `takerAssetAmountSold`, `makerAssetAmountPurchased`, and `affiliateFeePaid`.

contracts/exchange-forwarder/contracts/src/interfaces/IForwarderCore.sol
```diff
@@ -25,56 +25,41 @@ import "@0x/contracts-exchange-libs/contracts/src/LibFillResults.sol";
...
```
abandeali1: Same as above.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
mzhu25: There's still some contract logic that assumes the taker asset is WETH (e.g. keeping track of `wethSpentAmount`) so I think we can hold off on renaming this stuff until the next PR

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -77,183 +80,243 @@ contract MixinExchangeWrapper is
...
-                signatures[i]
+                signature
+            );
+        // WETH fee
+        } else if (order.takerFeeAssetData.equals(order.takerAssetData)) {
+            // Calculate the remaining amount of takerAsset to sell
+            uint256 remainingTakerAssetFillAmount = LibMath.getPartialAmountCeil(
+                order.takerAssetAmount,
+                order.makerAssetAmount,
+                makerAssetBuyAmount.safeSub(makerAssetAcquiredAmount)
```
abandeali1: This line is reused. Can we set it to `remainingMakerAssetFillAmount` above, similar to what we do in `_marketSellSingleOrder`?

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -77,183 +80,243 @@ contract MixinExchangeWrapper is
...
-            if (makerAssetFilledAmount >= makerAssetFillAmount) {
-                break;
-            }
+            // Attempt to sell the remaining amount of takerAsset
+            singleFillResults = _fillOrderNoThrow(
+                order,
+                remainingTakerAssetFillAmount,
+                signature
+            );
+        } else {
```
abandeali1: Thinking about it more, should there be a third case where `takerFee == 0`? It is possible that an order has null or random `takerFeeAssetData` with a 0 fee, which should still be supported by this contract. That should probably be the first case we check for both `_marketBuySingleOrder` and `_marketSellSingleOrder` since it is the simplest.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -77,183 +80,243 @@ contract MixinExchangeWrapper is
...
-                signatures[i]
+                signatures[i],
+                wethSellAmount,
+                wethSpentAmount
             );
 
-            // Update amounts filled and fees paid by maker and taker
-            _addFillResults(totalFillResults, singleFillResults);
+            // Percentage fee
+            if (orders[i].takerFeeAssetData.equals(orders[i].makerAssetData)) {
```
abandeali1: Rather than repeating these `if...else` blocks, can we put this logic into `_marketSellSingleOrder`? We should already have all of the information we need there. It can probably just return `makerAssetAcquiredAmount` and `wethSpentAmount` rather than `singleFillResults`.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -77,183 +80,243 @@ contract MixinExchangeWrapper is
...
+                signatures[i],
+                makerAssetBuyAmount,
+                makerAssetAcquiredAmount
             );
 
-            // Update amounts filled and fees paid by maker and taker.
-            _addFillResults(totalFillResults, singleFillResult);
-            zrxPurchased = _safeSub(totalFillResults.makerAssetFilledAmount, totalFillResults.takerFeePaid);
+            // Percentage fee
+            if (orders[i].takerFeeAssetData.equals(orders[i].makerAssetData)) {
```
abandeali1: Similar to above, I think this logic can live in `_marketBuySingleOrder` so that we don't have to repeat the `if...else` blocks.

contracts/exchange-forwarder/contracts/src/MixinWeth.sol
```diff
@@ -18,78 +18,83 @@
...
 
     /// @dev Converts message call's ETH value into WETH.
     function _convertEthToWeth()
         internal
     {
-        require(
-            msg.value > 0,
-            "INVALID_MSG_VALUE"
-        );
+        if (msg.value <= 0) {
```
abandeali1: Nit: `msg.value` can't ever be less than 0.

contracts/exchange-forwarder/contracts/src/libs/LibForwarderRichErrors.sol
```diff
@@ -0,0 +1,214 @@
...
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodeWithSelector(
+            UNSUPPORTED_ASSET_PROXY_ERROR_SELECTOR,
+            proxyId
+        );
+    }
+
+    function CompleteFillFailedError()
```
abandeali1: We have a similar error in the `Exchange` contract and recently added `uint256 expectedAssetFillAmount` and `uint256 actualAssetFillAmount` as parameters. Can we make that change here too?

contracts/exchange-forwarder/contracts/src/libs/LibForwarderRichErrors.sol
```diff
@@ -0,0 +1,214 @@
...
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodeWithSelector(
+            FEE_PERCENTAGE_TOO_LARGE_ERROR_SELECTOR,
+            feePercentage
+        );
+    }
+
+    function InsufficientEthForFeeError()
```
abandeali1: Let's add `ethFeeRequired` and `ethAvailable` parameters to this error.

contracts/exchange-forwarder/contracts/src/libs/LibForwarderRichErrors.sol
```diff
@@ -0,0 +1,214 @@
...
+    function TransferFailedError()
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return TRANSFER_FAILED_ERROR;
+    }
+
+    function DefaultFunctionWethContractOnlyError(
+        address callerAddress
```
abandeali1: Nit: we use `senderAddress` everywhere else.

contracts/exchange-forwarder/contracts/src/libs/LibForwarderRichErrors.sol
```diff
@@ -0,0 +1,214 @@
...
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodeWithSelector(
+            DEFAULT_FUNCTION_WETH_CONTRACT_ONLY_ERROR_SELECTOR,
+            callerAddress
+        );
+    }
+
+    function InvalidMsgValueError()
```
abandeali1: Since we're really just checking that `msg.value` isn't 0, can we rename to to `MsgValueCantEqualZeroError` or something similar?

contracts/exchange-forwarder/contracts/src/libs/LibForwarderRichErrors.sol
```diff
@@ -0,0 +1,214 @@
...
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return INVALID_MSG_VALUE_ERROR;
+    }
+
+    function InvalidErc721AmountError(
```
abandeali1: Similar to above, can we rename this to something like `Erc721AmountMustEqualOneError`?

contracts/exchange-forwarder/contracts/src/MixinAssets.sol
```diff
@@ -89,10 +116,9 @@ contract MixinAssets is
             msg.sender,
             amount
         ));
-        require(
-            success,
-            "TRANSFER_FAILED"
-        );
+        if (!success) {
+            LibRichErrors.rrevert(LibForwarderRichErrors.TransferFailedError());
```
abandeali1: We can get the `returnData` from the above call (the second return value). It might be useful to pass that into `TransferFailedError`.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -77,183 +80,243 @@ contract MixinExchangeWrapper is
...
+        }
+    }
+
     /// @dev Synchronously executes multiple calls of fillOrder until total amount of WETH has been sold by taker.
     ///      Returns false if the transaction would otherwise revert.
     /// @param orders Array of order specifications.
     /// @param wethSellAmount Desired amount of WETH to sell.
     /// @param signatures Proofs that orders have been signed by makers.
-    /// @return Amounts filled and fees paid by makers and taker.
+    /// @return Amounts of WETH spent and makerAsset acquired by the taker.
```
dorothy-zbornak: I know we're not very consistent about this, but we should probably start using multiple `@return` directives for functions that have multiple return values.

packages/order-utils/src/forwarder_revert_errors.ts
```diff
@@ -0,0 +1,112 @@
+import { BigNumber, RevertError } from '@0x/utils';
+import * as _ from 'lodash';
```
dorothy-zbornak: Not used.

contracts/exchange-forwarder/test/forwarder.ts
```diff
@@ -262,119 +264,318 @@ describe(ContractName.Forwarder, () => {
...
             });
+            const fillableOrder = await orderFactory.newSignedOrderAsync();
+
+            await forwarderTestFactory.marketSellTestAsync(
+                [unfillableOrder, fillableOrder],
+                new BigNumber(1.5),
+                erc20Token,
+            );
+        });
+        it('should skip over an expired order', async () => {
```
dorothy-zbornak: :+1:

contracts/exchange-forwarder/test/forwarder.ts
```diff
@@ -6,61 +6,52 @@ import {
...
-import { artifacts, ForwarderContract, ForwarderWrapper } from '../src';
+import { artifacts, ForwarderContract, ForwarderTestFactory, ForwarderWrapper } from '../src';
 
 chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 const DECIMALS_DEFAULT = 18;
-const MAX_WETH_FILL_PERCENTAGE = 95;
 
 describe(ContractName.Forwarder, () => {
```
dorothy-zbornak: We should probably upgrade these fixtures to use the new mocha extensions added in #2007 while we're here.

contracts/exchange-forwarder/test/utils/forwarder_test_factory.ts
```diff
@@ -0,0 +1,396 @@
...
+        this._takerAddress = takerAddress;
+        this._orderFeeRecipientAddress = orderFeeRecipientAddress;
+        this._forwarderFeeRecipientAddress = forwarderFeeRecipientAddress;
+        this._wethAddress = wethAddress;
+        this._gasPrice = gasPrice;
+    }
+
+    public async marketBuyTestAsync(
+        orders: SignedOrder[],
+        fractionalNumberOfOrdersToFill: BigNumber,
```
dorothy-zbornak: It kind of feels like a cognitive waste to make this a `BigNumber` since it seems to always be something that fits in a `number`.

PR #1995, opened 2019-07-25
Decode log arguments in `awaitTransactionSuccessAsync`

hysz:
## Description

The `awaitTransactionSuccessAsync` function in generated contract wrappers will decode logs, but requires ABI definitions to map topics to arguments. To do this we need (i) the abi for the wrapped contract and (ii) the abi for any contracts it calls.

This PR introduces a new parameter, `abiDependencies`, to the constructor of generated contract wrappers. This is passed in when we deploy the contract (which instantiates the wrapper):

```
// Old
const exchange = await ExchangeContract.deployFrom0xArtifactAsync(
    artifacts.Exchange,
    provider,
    txDefaults,
    zrxTokenAddress
);

// New
const dependencyArtifacts = {
    ...erc20Artifacts,
    ...erc721Artifacts,
    ...erc1155Artifacts,
};
const exchange = await ExchangeContract.deployFrom0xArtifactAsync(
    artifacts.Exchange,
    provider,
    txDefaults,
    dependencyArtifacts,
    zrxTokenAddress
);
```

## Checklist:


contracts/exchange/test/utils/exchange_wrapper.ts
```diff
@@ -251,15 +222,14 @@ export class ExchangeWrapper {
         from: string,
     ): Promise<TransactionReceiptWithDecodedLogs> {
         const params = orderUtils.createMatchOrders(signedOrderLeft, signedOrderRight);
-        const txHash = await this._exchange.matchOrders.sendTransactionAsync(
+        const txReceipt = await this._exchange.matchOrders.awaitTransactionSuccessAsync(
```
jalextowle: This is very nice 🎉

contracts/utils/contracts/test/TestLogDecodingDownstream.sol
```diff
@@ -0,0 +1,48 @@
...
+contract TestLogDecodingDownstream is
+    ITestLogDecodingDownstream
+{
+
+    /// @dev event with fields different than those in `TestLogDecoding.TestEvent`
+    /// Note: do not include this in the interface
+    /// For testing, we want to emit an event that is
+    /// not known by the calling contract.
+    event TestEvent2(
+        uint256 lorem,
```
jalextowle: I love the `lorem ipsum` 👍

contracts/utils/test/log_decoding.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    describe('Decoding Log Arguments', () => {
+        it('should decode locally emitted event args when no dependencies are passed into wrapper', async () => {
+            const txReceipt = await testLogDecodingDeployedWithoutDependencies.emitEvent.awaitTransactionSuccessAsync();
+            expect(txReceipt.logs.length).to.be.equal(1);
+            // tslint:disable no-unnecessary-type-assertion
+            expect((txReceipt.logs[0] as LogWithDecodedArgs<DecodedLogArgs>).args).to.be.deep.equal(expectedEvent);
+        });
+        it('should not event args when no dependencies are passed into wrapper', async () => {
```
jalextowle: "should not decode"?

contracts/utils/test/log_decoding.ts
```diff
@@ -0,0 +1,76 @@
...
+        );
+    });
+    beforeEach(async () => {
+        await blockchainLifecycle.startAsync();
+    });
+    afterEach(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+
+    describe('Decoding Log Arguments', () => {
```
dorothy-zbornak: It isn't immediately clear from the tests what the behavior is when we do an `emitEvent()` and `emitEventDownstream()` at once, without downstream artifacts. I assume `TestEvent` will turn into a `DecodedLogEntry` and `TestEvent2` will remain a `LogEntry`, but it might be useful to demonstrate it.

hysz: Heh I actually had a test for that and removed it. I'll put it back in. 🙏

packages/abi-gen-wrappers/src/generated-wrappers/erc20_token.ts
```diff
@@ -583,6 +584,7 @@ export class ERC20TokenContract extends BaseContract {
         artifact: ContractArtifact | SimpleContractArtifact,
         supportedProvider: SupportedProvider,
         txDefaults: Partial<TxData>,
+        artifactDependencies: { [contractName: string]: ContractArtifact | SimpleContractArtifact },
```
fabioberger: Shouldn't this be named `contractDependenciesArtifacts`? You are passing in the artifacts of the dependencies of the contract no? `artifactDependencies` made me think these are dependencies of the artifacts.

hysz:
Thanks @fabioberger @xianny, I appreciate the feedback!

To provide more context, the artifacts/abis are dependencies of the wrapper but not necessarily of the underlying contract. Moreover, they’re known by the end user of the wrapper but may not be known to the author of the contract. 

Here are two examples of when these dependencies are needed by the wrapper:

Screen Shot 2019-07-27 at 12 53 18 PM


At present these artifacts are used solely to decode logs, although I used more general naming to try and future proof the parameter. That said, it sounds like this naming is too ambiguous and may not be prescriptive enough. What if we go with something like,`logDecodeDependencies`?

fabioberger:
Thanks for the explanation @hysz. I prefer `logDecodeDependencies` personally. Let's also make sure to add a very good explanation in the doc comment for that method, to further educate the end dev on what it's for.

PR #2007, opened 2019-07-28
Mocha extensions for smart contract tests

dorothy-zbornak:
## Description
This PR introduces various boilerplate reductions for smart contract tests via `@0x/contracts-test-utils`.
- `@0x/contracts-test-utils`
    - Exports a `blockchainTests()` function that:
        - Has similar semantics to `describe()`.
        - Automatically spins up a `BlockchainLifecycle` on first call.
        - Passes an "environment" object into the callback with many common requirements.
        - Supports the following test suite "modifiers":
            - `blockchainTests.skip`: Analogous to `describe.skip`.
            - `blockchainTests.only`: Analogous to `describe.only`.
            - `blockchainTests.optional`: Only runs if the `TEST_ALL` environment variable is set.
            - `blockchainTests.resets`: Automatically calls `BlockchainLifecycle.startAsync()/revertAsync()` before/after every nested test case.
    - Exports a custom `describe()` function that supports the `optional` modifier.
    - Automatically calls `chaiSetup.configure()` on import.
    - Exports the Chai `expect` function, instead of have to import `chai` and doing `const expect = chai.expect`.

#### The Environment Object
The handler for `blockchainTests()` receives a single object with the following interface:
```ts
interface BlockchainTestsEnvironment {
    blockchainLifecycle: BlockchainLifecycle;
    provider: Web3ProviderEngine;
    txDefaults: Partial<TxData>;
    web3Wrapper: Web3Wrapper;
    async getChainIdAsync(): Promise<number>;
    async getAccountAddressesAsync(): Promise<string[]>;
}
```

#### Subtests 
The [tests](../blob/feature/3.0/mocha-blockchain-extensions/contracts/test-utils/test/mocha_blockchain.ts#L78) also feature an example of including separate test files within a test suite. It seems to work just fine if you want to explore this pattern.

## Testing instructions

New tests have been added to `@0x/contracts-test-utils`.

## Types of changes

## Checklist:

### DIFF
--- a/contracts/exchange/test/fill_order.ts
+++ b/contracts/exchange/test/fill_order.ts
@@ -1,8 +1,5 @@
-import { chaiSetup, txDefaults, web3Wrapper } from '@0x/contracts-test-utils';
-import { BlockchainLifecycle } from '@0x/dev-utils';
-import { logUtils } from '@0x/utils';
+import { blockchainTests } from '@0x/contracts-test-utils';
 import * as _ from 'lodash';
-import { env } from 'process';
 
 import {
     AllowanceAmountScenario,
@@ -20,15 +17,6 @@ import {
     FillOrderCombinatorialUtils,
     fillOrderCombinatorialUtilsFactoryAsync,
 } from './utils/fill_order_combinatorial_utils';
-
-chaiSetup.configure();
-const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
-
-let describeOptional = describe;
-if (env.TEST_ALL === undefined || env.TEST_ALL === '0') {
-    describeOptional = describe.skip;
-    logUtils.warn(`Combinatorial tests are disabled. Run with "TEST_ALL=1" env variable to enable.`);
-}
 
 const defaultFillScenario = {
     orderScenario: {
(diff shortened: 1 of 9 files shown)


contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -0,0 +1,187 @@
...
+export type BlockchainSuiteCallback = (this: ISuiteCallbackContext, env: BlockchainTestsEnvironment) => void;
+export type BlockchainContextDefinitionCallback<T> = (description: string, callback: BlockchainSuiteCallback) => T;
+export interface ContextDefinition extends Mocha.IContextDefinition {
+    optional: ContextDefinitionCallback<ISuite | void>;
+}
+
+/**
+ * Interface for `blockchainTests()`.
+ */
+export interface BlockchainContextDefinition {
```
hysz: If the subfields of `resets` will always be the same as the outer fields, it possible to do something like this?
```
export interface BlockchainContextDefinitionPartial {
    only: BlockchainContextDefinitionCallback<ISuite>;
    skip: BlockchainContextDefinitionCallback<void>;
    optional: BlockchainContextDefinitionCallback<ISuite | void>;
    (description: string, callback: BlockchainSuiteCallback): ISuite;
}

export interface BlockchainContextDefinition extends BlockchainContextDefinitionPartial {
    resets: BlockchainContextDefinitionPartial;
}
```

dorothy-zbornak: yeeerp :+1:

contracts/exchange/test/fill_order.ts
```diff
@@ -21,15 +18,6 @@ import {
     fillOrderCombinatorialUtilsFactoryAsync,
 } from './utils/fill_order_combinatorial_utils';
 
-chaiSetup.configure();
```
jalextowle: Goodbye boilerplate 🎉

PR #2005, opened 2019-07-27
Use TypeDoc Markdown docs for @0x/web3-wrapper

fabioberger:
## Description

We would like to move away from using our own custom-built documentation pages. The first step is to generate TypeDoc Markdown documentation pages for non-essential packages. This PR shows what this will look like by doing it for `@0x/web3-wrapper`. 

## Checklist:

-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #2013, opened 2019-07-30
Add devdocs to generated contract wrappers

### LINKED ISSUE #1903: Add devdoc to typescript generated wrappers
Typescript generated wrappers should have devdocs for all generated methods. When we eventually deprecate `contract-wrappers`, we'll need these to generate docpages.

Devdocs may need to be added to compiler output.

xianny:
## Description

Adds devdocs to generated contract-wrappers from Solidity doc comments. Resolves #1903

TODO
- [x] add test in abi-gen for method params
- [x] add test in abi-gen for multiline method desc
- [x] new lines can't be detected in devdocs
- [x] improve method descriptions for standard methods e.g. `callAsync`, `estimateGasAsync`
~- [ ] put generated docs in publish workflow or try [new docs workflow](https://github.com/0xProject/0x-monorepo/pull/2005)~ ON HOLD

## Testing instructions

Run `yarn docs` and look at generated docs `index.html`!

## Checklist:


packages/abi-gen-templates/partials/callAsync.handlebars
```diff
@@ -1,3 +1,10 @@
+/**
+ * Calls the method
```
fabioberger: `Sends a read-only call to the contract method returning the result that would happen were one to submit an Ethereum transaction to this method given the current state of the blockchain. Calls do not cost gas since they don't modify state.`

packages/abi-gen-templates/partials/callAsync.handlebars
```diff
@@ -34,6 +41,11 @@ async callAsync(
     // tslint:enable boolean-naming
     return result;
 },
+
+/**
+ * Returns the ABI encoded transaction data
```
fabioberger: `Returns the ABI encoded transaction data that would be sent in the Ethereum tx calling this method. This encoded tx data can then be sent to a separate signing service or can be used to create a 0x transaction (see protocol spec for more details).`

packages/abi-gen-templates/partials/tx.handlebars
```diff
@@ -1,4 +1,10 @@
 public {{languageSpecificName}} = {
+    /**
+     * Sends the transaction
```
fabioberger: `Sends an Ethereum transaction executing this method with the supplied parameters. This is a read/write Ethereum operation and will cost gas.`

packages/abi-gen-templates/partials/tx.handlebars
```diff
@@ -27,6 +33,13 @@ public {{languageSpecificName}} = {
     const txHash = await self._web3Wrapper.sendTransactionAsync(txDataWithDefaults);
     return txHash;
     },
+    /**
+     * Sends the transaction and wait for it to succeed
```
fabioberger: `Sends an Ethereum transaction and waits until the transaction has been successfully mined without reverting. If the transaction was mined, but reverted, an error is thrown.`

packages/abi-gen-templates/partials/tx.handlebars
```diff
@@ -54,6 +67,12 @@ public {{languageSpecificName}} = {
         })(),
     );
     },
+    /**
+     * Estimate gas to send the transaction
```
fabioberger: `Estimates the gas cost of sending an Ethereum transaction calling a specific method with specific arguments.`

packages/abi-gen-templates/partials/callAsync.handlebars
```diff
@@ -34,6 +41,11 @@ async callAsync(
     // tslint:enable boolean-naming
     return result;
 },
+
+/**
+ * Returns the ABI encoded transaction data needed to send an Ethereum tx calling this method. Before sending the Ethereum tx, this encoded tx data can first be sent to a separate signing service or can be used to create a 0x transaction (see protocol spec for more details).
```
fabioberger:
```suggestion
 * Returns the ABI encoded transaction data needed to send an Ethereum transaction calling this method. Before sending the Ethereum tx, this encoded tx data can first be sent to a separate signing service or can be used to create a 0x transaction (see protocol spec for more details).
```

packages/abi-gen-wrappers/README.md
```diff
@@ -71,3 +71,8 @@ yarn lint
 ```bash
 yarn test
 ```
+
+### Reading Documentation
+
+Documentation for this package is generated by TypeDoc, using the Solidity source code for 0x contracts. Each contract corresponds to one global-level module, which contains relevant enums and interfaces for its events and structs. Most significantly, each module exports a class, `<ContractName>Contract`, e.g. `ExchangeContract`, which implements helper methods for all the functions defined in the corresponding contract.
+A convention to note is that these contract-specific helper methods are defined as _object literals_, which are separated from methods in the generated documentation. Each contract method has a number of sub-methods, e.g. `sendTransactionAsync`, or `estimateGasAsync`, which are documented separately. This is an example of an expected method call signature: `exchangeContractInstance.fillOrder.sendTransactionAsync(...arguments)`.
```
feuGeneA: If you used a newline here to denote a new paragraph, it's not working.  (Github's "rich diff" view is showing that this line and the previous one are rendering as one big paragraph.)

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -699,6 +846,14 @@ export class AssetProxyOwnerContract extends BaseContract {
             const txHash = await self._web3Wrapper.sendTransactionAsync(txDataWithDefaults);
             return txHash;
         },
+        /**
+         * Sends an Ethereum transaction and waits until the transaction has been successfully mined without reverting. If the transaction was mined, but reverted, an error is thrown.
+         * @param assetProxyContract         Address of AssetProxy contract.
+         * @param isRegistered         Status of approval for AssetProxy contract.
+         * @param txData                Additional data for transaction
+         * @param pollingIntervalMs     Interval at which to poll for success
```
feuGeneA: There are many instances of weird misalignment like this.  I'm wondering whether it's not too unconventional to just put a single space between the parameter name and it's description, and not worry about alignment at all.

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -1013,6 +1252,10 @@ export class AssetProxyOwnerContract extends BaseContract {
             const gas = await self._web3Wrapper.estimateGasAsync(txDataWithDefaults);
             return gas;
         },
+        /**
+         * Sends a read-only call to the contract method. Returns the result that would happen if one were to send an Ethereum transaction to this method, given the current state of the blockchain. Calls do not cost gas since they don't modify state.
+         * @param _secondsTimeLocked         Duration needed after a transaction is confirmed and before it becomes executable, in seconds.
```
feuGeneA: Here's a good example of a long parameter description which could be wrapped properly by `cli-format`.

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -1095,6 +1346,10 @@ export class AssetProxyOwnerContract extends BaseContract {
             // tslint:enable boolean-naming
             return result;
         },
+
+        /**
+         * Returns the ABI encoded transaction data needed to send an Ethereum tx calling this method. Before sending the Ethereum tx, this encoded tx data can first be sent to a separate signing service or can be used to create a 0x transaction (see protocol spec for more details).
+         */
```
feuGeneA: Unnecessary newline.  This seems to occur everywhere this comment appears.

packages/abi-gen-wrappers/package.json
```diff
@@ -18,7 +18,8 @@
         "prettier": "prettier --write src/**/* --config ../../.prettierrc",
         "prettier_contract_wrappers": "prettier --write src/generated-wrappers/* --config ../../.prettierrc",
         "clean": "shx rm -rf lib src/generated-wrappers",
-        "generate_contract_wrappers": "abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/partials/**/*.handlebars' --output src/generated-wrappers --backend ethers"
+        "generate_contract_wrappers": "abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/partials/**/*.handlebars' --output src/generated-wrappers --backend ethers",
+        "docs": "typedoc --excludePrivate --excludeExternals --target ES5 --tsconfig typedoc-tsconfig.json --out generated_docs ./src/generated-wrappers/*"
```
feuGeneA: I think we should somehow integrate this new `docs` script with both CI and the local development workflow.  Including it in `yarn build` seems reasonable enough for me, but there may be a better place. (Isn't there a `build-docs` step in CI?)

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
xianny: I don't feel strongly about alignment. Single space looks a bit crowded to me though. Trying 4 spaces

packages/abi-gen-wrappers/package.json
xianny: There _is_ a `build-docs` step in CI, but @fabioberger is working on a totally new docs publishing flow which should be ready soon. We decided _not_ to add this one to the existing publishing flow, so it's in limbo right now until the new flow is ready. For now we only need to make sure that it's valid typedoc.

packages/abi-gen-templates/partials/callAsync.handlebars
```diff
@@ -1,3 +1,10 @@
+/**
+ * Sends a read-only call to the contract method. Returns the result that would happen if one were to send an Ethereum transaction to this method, given the current state of the blockchain. Calls do not cost gas since they don't modify state.
```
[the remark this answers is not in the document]
xianny: thanks, I forgot about `cli-format! will change.

packages/abi-gen-wrappers/package.json
feuGeneA: Okay, that all makes sense, but I still think we should integrate this with CI somehow (`yarn build` is plenty sufficient), just so that this command doesn't go unexercised and break at some point without us knowing.

.circleci/config.yml
```diff
@@ -234,6 +234,8 @@ jobs:
                   key: coverage-python-sra-client-{{ .Environment.CIRCLE_SHA1 }}
                   paths:
                       - ~/repo/python-packages/sra_client/.coverage
+            - store_artifacts:
+                  path: ~/repo/packages/abi-gen-wrappers/generated_docs
```
feuGeneA: This is in the wrong spot.  You've got it persisting artifacts generated as part of `test-python`, but the docs are generated over in `build`, not `test-python`.  This addition needs to move over to be part of `build`.  (Concretely, I don't see any `abi-gen-wrappers` artifacts available in the `test-python` build results.)

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -1292,10 +1343,12 @@ export class AssetProxyOwnerContract extends BaseContract {
...
         },
-
         /**
-         * Returns the ABI encoded transaction data needed to send an Ethereum tx calling this method. Before sending the Ethereum tx, this encoded tx data can first be sent to a separate signing service or can be used to create a 0x transaction (see protocol spec for more details).
-         * @param _secondsTimeLocked         Duration needed after a transaction is confirmed and before it becomes executable, in seconds.
+         * Returns the ABI encoded transaction data needed to send an Ethereum transaction calling this method. Before
+         * sending the Ethereum tx, this encoded tx data can first be sent to a separate signing service or can be used
+         * to create a 0x transaction (see protocol spec for more details).
+         * @param _secondsTimeLocked Duration needed after a transaction is confirmed
+         * and before it becomes executable, in seconds.
```
feuGeneA: My aesthetic eye would like to see wrapped lines (here "and before...") indented 4 spaces, aka a hanging indent.  But if this is convention, and, more importantly, if the generated docs look fine, then whatever.

xianny:
Updated. Build artifacts visible here: https://82906-92181371-gh.circle-artifacts.com/0/home/circleci/repo/packages/abi-gen-wrappers/generated_docs/index.html

PR #2014, opened 2019-07-30
Added Unit Tests to Utils

jalextowle:
## Description

Added unit tests to `contracts/utils/`.

## Testing instructions

## Types of changes

* Testing change

## Checklist:


contracts/utils/contracts/test/TestOwnable.sol
```diff
@@ -0,0 +1,13 @@
...
+import "../src/Ownable.sol";
+
+
+contract TestOwnable is
+    Ownable
+{
+    function externalOnlyOwner()
+        external
+        onlyOwner
+    {} // solhint-disable-line no-empty-blocks
```
hysz: It may be good to return something on success to avoid false positives. Perhaps the function selector, although just returning a bool is probably fine too.

contracts/utils/contracts/test/TestReentrancyGuard.sol
```diff
@@ -0,0 +1,54 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "../src/ReentrancyGuard.sol";
+
+
+contract TestReentrancyGuard is
```
hysz: Could be good to add some function comments here, as it's not immediately obvious how this is to be used.

contracts/utils/contracts/test/TestReentrancyGuard.sol
```diff
@@ -0,0 +1,54 @@
...
+    {
+        if (counter > 0) {
+            counter--;
+            this.guarded(true);
+        } else {
+            counter = 2;
+        }
+    }
+
+    function nonExploitive()
```
hysz: Same as above re return value.

contracts/utils/test/lib_eip712.ts
```diff
@@ -0,0 +1,107 @@
...
+                '',
+                0,
+                constants.NULL_ADDRESS,
+            );
+        });
+    });
+
+    describe('_hashEIP712Message', () => {
+        it('should correctly hash empty input', async () => {
+            /*
```
hysz: Should this be uncommented or removed?

contracts/utils/test/ownable.ts
```diff
@@ -0,0 +1,65 @@
...
+        });
+
+        it('should succeed if sender is owner', async () => {
+            expect(ownable.externalOnlyOwner.callAsync({ from: owner })).to.be.fulfilled;
+        });
+    });
+
+    describe('transferOwnership', () => {
+        it('should not transfer ownership if the specified new owner is the zero address', async () => {
+            expect(ownable.transferOwnership.sendTransactionAsync(constants.NULL_ADDRESS, { from: owner })).to.be.fulfilled;
```
hysz: Hmm not sure how I feel about this failing silently. It's probably outside the scope of this PR, but I've created a task in the protocol backlog to change this.

contracts/utils/contracts/test/TestSafeMath.sol
```diff
@@ -0,0 +1,66 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "../src/SafeMath.sol";
+
+
+contract TestSafeMath is
```
hysz: Nit - even though we don't have an explicit exception, I think it would still be good to test `_safeDiv`. This would assert out assumption that the function throws on div-by-zero, as well as future-proof the expected behavior.

contracts/utils/test/ownable.ts
jalextowle: Yeah, I actually had a similar thought. I'm working on a PR that will disallow 0 in MixinAuthorizable as well, and I can fix it there.

contracts/utils/test/lib_eip712.ts
jalextowle: Removed 😬

contracts/utils/contracts/src/LibAddress.sol
```diff
@@ -1,6 +1,6 @@
 /*
 
-  Copyright 2018 ZeroEx Intl.
+  Copyright 2019 ZeroEx Intl.
```
dorothy-zbornak: Lol, I thought I was the only one doing this.

contracts/utils/test/lib_rich_errors.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    after(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+
+    describe('_rrevert', () => {
+        it('should correctly revert the extra bytes', async () => {
+            return expect(lib.externalRRevert.callAsync(constants.NULL_BYTES)).to.revertWith(constants.NULL_BYTES);
+        });
+    });
```
dorothy-zbornak: Not high stakes, but can we get a test with a payload, like `StandardError`? You can use the `@0x/utils.StringRevertError` type (probably my fault that they're named differently):

```ts
// This should work.
const error = new StringRevertError('foo');
return expect(..externalRRevert.callAsync(error.encode()).to.revertWith(error);
```

contracts/utils/test/ownable.ts
```diff
@@ -0,0 +1,65 @@
...
+    let ownable: TestOwnableContract;
+    let owner: string;
+    let nonOwner: string;
+
+    before(async () => {
+        const accounts = await web3Wrapper.getAvailableAddressesAsync();
+        owner = await accounts[0];
+        nonOwner = await accounts[1];
+        await blockchainLifecycle.startAsync();
+        // Deploy SafeMath from the owner address
```
dorothy-zbornak: Lies!

contracts/utils/test/ownable.ts
```diff
@@ -0,0 +1,65 @@
...
+        it('should throw if sender is not owner', async () => {
+            const expectedError = new OwnableRevertErrors.OnlyOwnerError(
+                nonOwner,
+                owner,
+            );
+            return expect(ownable.externalOnlyOwner.callAsync({ from: nonOwner })).to.revertWith(expectedError);
+        });
+
+        it('should succeed if sender is owner', async () => {
+            expect(ownable.externalOnlyOwner.callAsync({ from: owner })).to.be.fulfilled;
```
dorothy-zbornak: The process will end before this test actually executes.
```suggestion
            return expect(ownable.externalOnlyOwner.callAsync({ from: owner })).to.be.fulfilled('');
```

contracts/utils/test/ownable.ts
dorothy-zbornak:
```suggestion
            await expect(ownable.transferOwnership.sendTransactionAsync(constants.NULL_ADDRESS, { from: owner })).to.be.fulfilled('');
```

contracts/utils/test/ownable.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    describe('transferOwnership', () => {
+        it('should not transfer ownership if the specified new owner is the zero address', async () => {
+            expect(ownable.transferOwnership.sendTransactionAsync(constants.NULL_ADDRESS, { from: owner })).to.be.fulfilled;
+            const updatedOwner = await ownable.owner.callAsync();
+            expect(updatedOwner).to.be.eq(owner);
+        });
+
+        it('should transfer ownership if the specified new owner is not the zero address', async () => {
+            expect(ownable.transferOwnership.sendTransactionAsync(nonOwner, { from: owner })).to.be.fulfilled;
```
dorothy-zbornak:
```suggestion
            await expect(ownable.transferOwnership.sendTransactionAsync(nonOwner, { from: owner })).to.be.fulfilled('');
```

contracts/utils/test/reentrancy_guard.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+
+    describe('nonReentrant', () => {
+        it('should throw if reentrancy occurs', async () => {
+            const expectedError = new ReentrancyGuardRevertErrors.IllegalReentrancyError();
+            return expect(guard.guarded.sendTransactionAsync(true)).to.revertWith(expectedError);
+        });
+
+        it('should succeed if reentrancy does not occur', async () => {
+            expect(guard.guarded.sendTransactionAsync(false)).to.be.fulfilled; // tslint:disable-line:no-unused-expression
```
dorothy-zbornak:
```suggestion
            return expect(guard.guarded.sendTransactionAsync(false)).to.be.fulfilled('');
```

contracts/utils/test/safe_math.ts
```diff
@@ -0,0 +1,146 @@
...
+            const result = await safeMath.externalMaxUint256.callAsync(constants.ZERO_AMOUNT, toBN(13));
+            expect(result).bignumber.to.be.eq(toBN(13));
+        });
+
+        it('should return first argument if it is equal the second', async () => {
+            const result = await safeMath.externalMaxUint256.callAsync(constants.ZERO_AMOUNT, constants.ZERO_AMOUNT);
+            expect(result).bignumber.to.be.eq(constants.ZERO_AMOUNT);
+        });
+
+        it('should return second argument if it is greater than the first', async () => {
```
dorothy-zbornak:
```suggestion
        it('should return second argument if it is less than the first', async () => {
```

packages/utils/src/safe_math_revert_errors.ts
```diff
@@ -10,7 +10,7 @@ export enum SafeMathErrorCodes {
 }
 
 export class SafeMathError extends RevertError {
-    constructor(error?: SafeMathErrorCodes, a?: BigNumber | number | string, b?: BigNumber | number | string) {
+    constructor(error?: SafeMathErrorCodes, a?: BigNumber, b?: BigNumber) {
```
dorothy-zbornak: `string`s are a little sketchy, but what was wrong allowing `number`s?

contracts/utils/contracts/src/LibAddress.sol
jalextowle: Nope lol, I try to do it in every file I touch

packages/utils/src/safe_math_revert_errors.ts
jalextowle: I just did this for consistency. I've noticed that in a lot of places we just accept BigNumbers, which I kind of like. If you feel strongly about this, I can change it back

contracts/utils/test/lib_rich_errors.ts
```diff
@@ -0,0 +1,40 @@
...
+import * as chai from 'chai';
+import * as _ from 'lodash';
+
+import { artifacts, TestLibRichErrorsContract } from '../src';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
+
+describe('LibEIP712', () => {
```
hysz: `LibRichRevert` ?

contracts/utils/test/lib_eip712.ts
```diff
@@ -0,0 +1,87 @@
...
+        await blockchainLifecycle.startAsync();
+        // Deploy SafeMath
+        lib = await TestLibEIP712Contract.deployFrom0xArtifactAsync(artifacts.TestLibEIP712, provider, txDefaults);
+    });
+
+    after(async () => {
+        await blockchainLifecycle.revertAsync();
+    });
+
+    describe('_hashEIP712Domain', async () => {
```
hysz: Should we test hashing non-empty input here and below?

jalextowle: From a code coverage standpoint, I actually don't think that it is important, but the reason that I abstracted the logic into functions is that it makes it really easy to add test cases. I'll add a few.

contracts/utils/test/safe_math.ts
```diff
@@ -0,0 +1,171 @@
...
+            const result = await safeMath.externalSafeDiv.callAsync(toBN(2), toBN(3));
+            expect(result).bignumber.to.be.eq(constants.ZERO_AMOUNT);
+        });
+
+        it('should return zero if first argument is zero', async () => {
+            const result = await safeMath.externalSafeDiv.callAsync(constants.ZERO_AMOUNT, toBN(1));
+            expect(result).bignumber.to.be.eq(constants.ZERO_AMOUNT);
+        });
+
+        it('should return zero if second argument is zero', async () => {
```
abandeali1: Shouldn't this be `it('should revert if...)`?

contracts/utils/test/safe_math.ts
```diff
@@ -0,0 +1,171 @@
...
+            const errMessage = 'VM Exception while processing transaction: invalid opcode';
+            return expect(safeMath.externalSafeDiv.callAsync(toBN(1), constants.ZERO_AMOUNT)).to.be.rejectedWith(
+                errMessage,
+            );
+        });
+    });
+
+    describe('_safeSub', () => {
+        it('should throw if the subtraction underflows', async () => {
```
abandeali1: nit: `should revert`

contracts/utils/test/safe_math.ts
```diff
@@ -0,0 +1,171 @@
...
+        });
+
+        it('should calculate correct value for values that are not equal', async () => {
+            const result = await safeMath.externalSafeSub.callAsync(toBN(15), toBN(13));
+            expect(result).bignumber.to.be.eq(toBN(2));
+        });
+    });
+
+    describe('_safeAdd', () => {
+        it('should throw if the addition overflows', async () => {
```
abandeali1: `should revert`

contracts/utils/test/safe_math.ts
```diff
@@ -0,0 +1,171 @@
...
+import * as chai from 'chai';
+import * as _ from 'lodash';
+
+import { artifacts, TestSafeMathContract } from '../src';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
+
+function toBN(a: number | string): BigNumber {
```
abandeali1: Can we rename this `toBigNumber`? `BN` is the name of another popular BigNumber library.

PR #2017, opened 2019-07-30
Fix broken validator signatures

### LINKED ISSUE #1998: @0x/order-utils: Validator signatures are broken
## Expected Behavior

Given a valid `Validator` type signature, `signatureUtils.isValidValidatorSignatureAsync` should eventually return true.

## Current Behavior
A valid `Validator` type signature is rejected with an error: 
```
Error: Invalid address: 'ead22c4e5cf1a4320ad8'
```

## Possible Solution
Currently there are multiple issues with `Validator` type signatures. The error given above originates from the `signatureUtils.parseValidatorSignature` function where the validatorAddress is extracted from the signature with only 20 characters instead of 40 (probably because addresses are 20 bytes, but the signature is hex encoded and therefore addresses are 40 characters). Also the `0x` prefix is missing from the address. This is line 494 and 495 in `./packages/order-utils/src/signature_utils.ts`.

However even if properly parsed, more problems exist. The `ExchangeContract` and `IValidatorContract` instances are initialized with `signerAddress` instead of the actual address of the exchange and `validatorAddress`, respectively. These are lines 160 and 171 in `./packages/order-utils/src/signature_utils.ts`. 

We are running on a development chain, therefore it would be extra useful if we could somehow inject the actual exchange address instead of taking it from the `getContractAddressesForNetworkOrThrow` as done in `signatureUtils.isValidPresignedSignatureAsync`.

## Steps to Reproduce (for bugs)

1. Given deployed 0x contracts, deploy custom validator contract that always returns true, at some address `validatorAddress`
2. With some address `maker`, call the `setSignatureValidatorApproval` function of the exchange contract.
3. Create an unsigned order with `makerAddress: maker` and hash it 
4. Create a `signature` for the order with some _other_ address `signer`
5. Concatenate the `signature` with the `validatorAddress` and signature type `05` to create the `validatorSignature`
6. Call the `isValidSignatureAsync` with `data` being the `orderHash`, `signature` being the `validatorSignature` and `signerAddress` being `maker`.

## Context
As mentioned, we are running on a local chain, with a deployed 0x exchange contract, trying to allow signatures through a custom validator contract. 

## Your Environment

| Package | Version |
| `@0x/order-utils` | 8.2.3 |

| Network |
| ------- |
| testrpc  |

jangerritharms:
## Description
This pull request fixes #1998, which were a series of bugs in the typescript implementation of the validator signature validation, in `signatureUtils`. In order to properly instantiate exchange contracts in situations where a custom exchange contract is used (like in a private/development chain setting), I have added an optional parameter to the `isValidSignature` function which allows to pass options like the exchange address to those validation functions that require it. The `parseValidatorSignature` function was added to the exported functions for unit testing.

## Testing instructions
Unit tests were written for the `parseValidatorSignature` function, integration tests are WIP. 

## Types of changes

* Bug fix: `parseValidatorSignature` parses signature correctly now 
* Bug fix: `isValidValidatorSignature` instantiates Exchange contract correctly
* Bug fix: `isValidValidatorSignature` instantiates IValidator contract at validator address taken from the `validatorSignature`

* New feature: optionally exchange addresses are passed to the isValidSignature to instantiate custom instances of the exchange contract

## Checklist:


jangerritharms:
I am wondering where to put the integration tests of the TS code because the tests in `@0x/order-utils` are only unit tests. @fabioberger I see that in #1580 you have put those in the tests of `@0x/contract-wrappers` but in my case I would require a Validator to be migrated as well. Should I add that to the tests in `@0x/contract-wrappers`? Alternatively I could test it in the `@0x/contracts-exchange` tests where the validator contract is tested, and check whether the TS implementation returns the same result as the contracts?!

jangerritharms:
I can't figure out why `test-publish` keeps failing. Am I missing any environment variables? I don't think I changed any code related to the errors there. Any ideas?

abandeali1:
@jangerritharms no need to worry about `test-publish`. I believe that always fails for external PRs.

fabioberger:
@jangerritharms I'd put the tests in `@0x/contracts-exchange`. We are about to replace `@0x/contract-wrappers` with `@0x/abi-gen-wrappers` and all the tests there will get removed. As a philosophy we are moving towards auto-generating our contract-wrappers and consolidating all testing of the Smart Contract logic into the contract package itself.

packages/order-utils/src/types.ts
```diff
@@ -25,6 +25,15 @@ export interface CreateOrderOpts {
...
 }
 
+export type SignatureValidationOpts = ValidatorSignatureOpts | PresignedSignatureOpts;
+
+export interface ValidatorSignatureOpts {
+    exchangeAddress: string;
+}
+
+export interface PresignedSignatureOpts {
+    exchangeAddress: string;
```
fabioberger: What other opts are we imagining could be added? Maybe we should just pass the `exchangeAddress` directly as an opt?

packages/order-utils/src/signature_utils.ts
```diff
@@ -144,20 +161,33 @@ export const signatureUtils = {
...
     ): Promise<boolean> {
         const provider = providerUtils.standardizeOrThrow(supportedProvider);
         assert.isHexString('data', data);
         assert.isHexString('signature', signature);
         assert.isETHAddressHex('signerAddress', signerAddress);
-        const validatorSignature = parseValidatorSignature(signature);
-        const exchangeContract = new ExchangeContract(signerAddress, provider);
+
+        let exchangeContract: ExchangeContract;
+        if (exchangeAddress) {
```
fabioberger: Nit: We prefer to be explicit and avoid using "truthiness"
```suggestion
        if (exchangeAddress !== undefined) {
```

packages/order-watcher/src/utils/assert.ts
```diff
@@ -16,12 +16,14 @@ export const assert = {
...
         signature: string,
         signerAddress: string,
+        signatureValidationOpts: SignatureValidationOpts,
     ): Promise<void> {
         const isValid = await signatureUtils.isValidSignatureAsync(
             supportedProvider,
             orderHash,
             signature,
             signerAddress,
+            signatureValidationOpts,
```
fabioberger: Let's add an entry in the `order-watcher` and `0x.js` CHANGELOGS since both of their public interfaces have changed in a non-breaking way. Minor patches.

packages/order-utils/src/types.ts
jangerritharms: Yeah, felt a little unnecessary when I was writing it. Will just be passing the exchangeAddress to the `isValidSignatureAsync` then. Done in 7228cbfe92ff6fee4bc59681fba1789bec85c4f9.

packages/order-utils/src/signature_utils.ts
jangerritharms: Fixed it in 42d5bdd3abd22bea9782606bd3a47e643e5cbeb1.

packages/order-watcher/src/utils/assert.ts
jangerritharms: Updated MINOR version of `0x.js` because of new optional argument in `signatureUtils.isValidSignatureAsync` and PATCH version in `order-watcher` because the public interface did not change (exchange address is taken from the `signedOrder`). Hope you agree?!

fabioberger:
@jangerritharms thanks for addressing my feedback. Looks like `static-tests` is failing because some of the deps in `@0x/contract-wrappers` haven't been updated:

`@0x/subproviders` -> `^5.0.1`
`@0x/dev-utils` -> `^2.3.0`
`@0x/migrations` -> `^4.2.0`
`@0x/fill-scenarios` -> `^3.0.16`

Once those are updated we can get this merged.

jangerritharms:
It seems that those versions are also outdated in the `development` branch. I did not make any changes to the `@0x/contract-wrappers` package. There are also some more conflicts of dependency versions between `@0x/website`, `@0x/instant` and `@0x/dev-tools-page`. Should I update those as part of this pull-request or should that rather be a separate one?

xianny:
@jangerritharms apologies, I broke the `yarn deps_versions:ci` check recently in development. It is fixed now so if you merge from development again the check should pass.

fabioberger:
Once the checks pass, I'll merge :)

jangerritharms:
Cool! Just pulled in latest `development`. Hope it passes 🤞

jangerritharms:
@fabioberger Everything passes except the `test-publish`, [which won't pass for external PRs](https://github.com/0xProject/0x-monorepo/pull/2017#issuecomment-516513428). So as far as I can see we can merge :).

jangerritharms:
@fabioberger Any chance you have the time to merge this or does it require additional changes?

xianny:
hey @jangerritharms thanks for updating the PR! Could you rebuild the failing `test-rest` job on CircleCI? I could not reproduce that failure locally, so it might be a sporadic thing. If it passes, we can merge.

jangerritharms:
I updated with the latest changes again. So the `test-contracts-ganache` step is failing because it's running out of memory. I have tried a few times. Problem is that I don't have a paid plan for CircleCI so it cannot use a `medium+` instance, probably that makes the difference. 😕 Not sure how I can make this pass, don't think it's actually broken.

feuGeneA:
@jangerritharms I'm so sorry you're having trouble with this.

> the `test-contracts-ganache` step is failing because it's running out of memory. I have tried a few times. Problem is that I don't have a paid plan for CircleCI so it cannot use a `medium+` instance, probably that makes the difference.  Not sure how I can make this pass, don't think it's actually broken.

I understand this problem.  I recently changed our CircleCI configuration to give more resources to `test-contracts-ganache`, because that test suite has grown significantly in recent weeks/months, so I know exactly what you're talking about.

I pushed your branch to 0x's git remote, so that CircleCI would build it on 0x's account.  Interestingly, CircleCI was smart enough to update the "checks" on this PR to use the runs from my branch, not yours.  Check the results again.  There is a `revert VALIDATOR_ERROR` test failure in the `MixinSignatureValidator` test in the `@0x/contracts-exchange` package.  Feel free to mention me here and I can do this again; I'm just starting my day and will be watching this PR.

We're exploring alternative CI solutions right now.  Your experience here highlights a problem with our current setup. I'll make sure that this highlight makes it into the list of considerations that guide our actions.

jangerritharms:
Thanks a lot @feuGeneA. It seems that #2078 updated the way invalid signatures are handled by the signature validator (reverting instead of returning false). I adjusted the failing test. Just strange that this wasn't caught in the CI pipeline of #2078, which passed. @feuGeneA Could you run it again if it fails on my CircleCI run?

feuGeneA:
@jangerritharms looking good!  Unfortunately, while I can help with this CI nonsense, I don't have the context needed to give you final approval here.

@fabioberger ? A quick re-review?

fabioberger:
Thanks for the contribution @jangerritharms! 🙏

PR #2018, opened 2019-07-30
Decode Calldata + Return Values in Contract Wrappers

hysz:
## Description

This adds ABI decoding of transaction calldata and return values to contract wrappers. These are particularly useful for meta-transactions and the delegate-call proxy pattern.

1. Adds `getABIDecodedTransactionData` to contract wrappers, which decodes calldata.
2. Adds `getABIDecodedReturnData` to contract wrappers, which decodes return values. 
3. Adds bounds-checking to the 0x abi encoder. Previously we interpreted anything past calldata as a zero; now reading past the end of calldata throws an exception. This is aligned with decoding in solidity and web3.
4. Adds a `strictDecode` to the abi encoder, which decodes to the same format as the function's arguments.

## Checklist:


packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
```diff
@@ -138,4 +138,134 @@ contract AbiGenDummy
...
+        bytes lorem;
+        bytes ipsum;
+        string dolor;
+    }
+
+    /// @dev The fallback function calls into this contract and executes one of the above functions.
+    /// This allows us to test `getABIDecodedTransactionData` and `getABIDecodedReturnData` that is
+    /// include in contract wrappers.
+    // solhint-disable no-complex-fallback
+    function ()
```
jalextowle: Doesn't this function just loop forever? If the calldata hits the fallback function, then the fallback will call back into this contract using the same calldata, which would intuitively also make it hit the fallback function, creating an infinite loop. Please let me know if I'm wrong about this, but I've done a bit of testing locally and it seems to loop forever.

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
```diff
@@ -138,4 +138,134 @@ contract AbiGenDummy
...
+        external
+        payable
+    {
+        address addr = address(this);
+        assembly {
```
jalextowle: Is there a reason why you're using assembly here? It seems like you could use high-level solidity for this.

packages/utils/test/abi_encoder/methods_test.ts
```diff
@@ -10,130 +11,114 @@ chaiSetup.configure();
 const expect = chai.expect;
 
 describe('ABI Encoder: Method Encoding / Decoding', () => {
-    const encodingRules: AbiEncoder.EncodingRules = { shouldOptimize: false }; // optimizer is tested separately.
+    const defaultEncodingRules: AbiEncoder.EncodingRules = { shouldOptimize: false }; // optimizer is tested separately.
+    const defaultDecodingRules: AbiEncoder.DecodingRules = { shouldConvertStructsToObjects: false };
+    const runTest = <T>(
```
jalextowle: This function is 🔥

packages/utils/src/abi_encoder/evm_data_types/method.ts
```diff
@@ -34,6 +34,19 @@ export class MethodDataType extends AbstractSetDataType {
         return value;
     }
 
+    public strictDecode<T>(calldata: string, rules?: DecodingRules): T {
+        const value = super.decode(calldata, rules, this._methodSelector);
```
dorothy-zbornak: I can't find any tests where we check that `decode()` with a mismatched method selector throws. Do we have one?

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
dorothy-zbornak: Wouldn't this never get called if used as intended because the functions it's forwarding to are defined in this contract?

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
```diff
@@ -138,4 +138,134 @@ contract AbiGenDummy
...
+    /// @dev complex input is dynamic and more difficult to decode than simple input.
+    struct ComplexOutput {
+        ComplexInput input;
+        bytes lorem;
+        bytes ipsum;
+        string dolor;
+    }
+
+    /// @dev The fallback function calls into this contract and executes one of the above functions.
+    /// This allows us to test `getABIDecodedTransactionData` and `getABIDecodedReturnData` that is
```
dorothy-zbornak: I'm not really sure how this can be used to test `getABIDecodedTransactionData()`. Where are the tests that use this?

packages/abi-gen-templates/partials/callAsync.handlebars
```diff
@@ -44,3 +44,21 @@ getABIEncodedTransactionData(
     const abiEncodedTransactionData = self._strictEncodeArguments('{{this.functionSignature}}', [{{> normalized_params inputs=inputs}}]);
     return abiEncodedTransactionData;
 },
+getABIDecodedTransactionData(
+    returnData: string
```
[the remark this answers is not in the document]
hysz: Type, thank you 🙏

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
hysz: Ah, yeah you're right this can be removed now that all tests are in the same contract. 🙏

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
hysz: See below 🙏

packages/utils/src/abi_encoder/evm_data_types/method.ts
hysz: Great catch -- test added!

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
```diff
@@ -138,4 +138,90 @@ contract AbiGenDummy
     uint someState;
     function nonPureMethod() public returns(uint) { return someState += 1; }
     function nonPureMethodThatReturnsNothing() public { someState += 1; }
+
```
xianny: We've been keeping all the test cases in one contract to speed up test run times. I'm thinking about how we could keep it somewhat organised, since we're working towards having full coverage of Solidity language features. We don't have this convention yet, but what about prefixing a section of test cases with the methods it's meant to test?

e.g. 
```
// begin tests for `decodeTransactionData`, `decodeReturnData`
...
// end tests for `decodeTransactionData`, `decodeReturnData`
```

packages/abi-gen/test-cli/test_typescript/test/abi_gen_dummy_test.ts
```diff
@@ -28,6 +28,20 @@ const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
 describe('AbiGenDummy Contract', () => {
     let abiGenDummy: AbiGenDummyContract;
+    const runTestAsync = async (contractMethod: any, input: any, output: any) => {
```
xianny: love this setup! 👍

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
jalextowle: @dorothy-zbornak Yeah, I think that is the case. I was just saying that it would loop infinitely if it was called for some reason.

packages/abi-gen/test-cli/fixtures/contracts/AbiGenDummy.sol
dorothy-zbornak: Why not just use inheritance?

hysz: @xianny I've added comments for now for consistency, although I agree w Dorothy - it could be good to split this into multiple contracts alongside the refactor of `abi-gen`.

xianny: Ya, let's keep that in mind. I'm a little concerned that inheritance might complicate testing, especially when we add the EVM simulator. As in we would have to test whatever case we want PLUS inheritance.

PR #2019, opened 2019-07-30
Disallow Zero address in MixinAuthorizable and Ownable

jalextowle:
Note: This PR is based on PR 2014 (because Ownable did not have unit tests previously). 2014 should be merged first.

## Description
- Previously, the zero address could be added to the `authorized` list in Authorizable (a new contract in `contracts/utils/` that is based on `MixinAuthorizable` from `contracts/asset-proxy/`. This PR fixes this issue and adds a unit test to ensure that adding zero will always fail.
- Additionally, Ownable will now revert if the owner attempts to transfer ownership to the zero address.
- Since the asset proxies are stable contracts and are almost never redeployed, an `archive/` directory was added to `contracts/asset-proxy/`, and the `Ownable` contract that was used for the `asset-proxy` deployment was put into the directory. This change sets a precedent for tracking old deployments while keeping them in the monorepo. 

## Testing instructions

## Types of changes

## Checklist:


contracts/utils/contracts/src/MixinAuthorizable.sol
```diff
@@ -0,0 +1,121 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/Ownable.sol";
+import "./interfaces/IAuthorizable.sol";
+
+
+contract MixinAuthorizable is
```
abandeali1: Maybe this should just be `Authorizable` now that this is a standalone util.

contracts/utils/contracts/src/MixinAuthorizable.sol
```diff
@@ -0,0 +1,121 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/Ownable.sol";
```
abandeali1: This can be imported from `./Ownable.sol`.

contracts/utils/contracts/src/MixinAuthorizable.sol
```diff
@@ -0,0 +1,121 @@
...
+
+contract MixinAuthorizable is
+    Ownable,
+    IAuthorizable
+{
+    /// @dev Only authorized addresses can invoke functions with this modifier.
+    modifier onlyAuthorized {
+        require(
+            authorized[msg.sender],
+            "SENDER_NOT_AUTHORIZED"
```
abandeali1: As discussed offline, let's add rich reverts to this contract and move the old `asset-proxy` versions into an `archived` directory in the `asset-proxy` package. The already deployed proxies should inherit the archived versions of this and `Ownable`.

contracts/asset-proxy/contracts/src/MixinAuthorizable.sol
```diff
@@ -18,7 +18,7 @@
 
 pragma solidity ^0.5.9;
 
-import "@0x/contracts-utils/contracts/src/Ownable.sol";
+import "../archive/Ownable.sol";
```
dorothy-zbornak: Hmmm.

I think this file should be archived too. New asset proxies will use `@0x/utils/Authorizable`, right?

contracts/asset-proxy/contracts/src/interfaces/IAuthorizable.sol
```diff
@@ -18,7 +18,7 @@
 
 pragma solidity ^0.5.9;
 
-import "@0x/contracts-utils/contracts/src/interfaces/IOwnable.sol";
+import "../../archive/interfaces/IOwnable.sol";
```
dorothy-zbornak: Hmm.

contracts/utils/test/authorizable.ts
```diff
@@ -0,0 +1,212 @@
...
+            const expectedError = new OwnableRevertErrors.OnlyOwnerError(notOwner, owner);
+            const tx = authorizable.addAuthorizedAddress.sendTransactionAsync(notOwner, { from: notOwner });
+            return expect(tx).to.revertWith(expectedError);
+        });
+
+        it('should allow owner to add an authorized address', async () => {
+            await authorizable.addAuthorizedAddress.awaitTransactionSuccessAsync(
+                address,
+                { from: owner },
+                constants.AWAIT_TRANSACTION_MINED_MS,
```
dorothy-zbornak: Don't think you really need this parameter.

contracts/asset-proxy/contracts/src/MixinAssetProxyDispatcher.sol
```diff
@@ -1,6 +1,6 @@
 /*
```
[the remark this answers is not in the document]
jalextowle: Good point!

contracts/asset-proxy/contracts/src/interfaces/IAuthorizable.sol
jalextowle: There probably wasn't a reason to archive this in retrospect. The reason I did was because the Solidity version is different, but I'll go ahead and remove.

PR #2024, opened 2019-07-31
[WIP] Feature/exposing receipt to generated wrappers

Freydal:
## Description

Adding code to the template and base class to expose the transaction receipt from contract deployment when using the generated contract wrappers.  This is primarily a change to support usage of 0x tooling rather than the 0x contract assets.

## Testing instructions

A freshly deployed contract wrapper instance should now have the `txReceipt` to obtain any useful information.

## Types of changes

## Checklist:

### DIFF
--- a/packages/abi-gen-templates/CHANGELOG.json
+++ b/packages/abi-gen-templates/CHANGELOG.json
@@ -9,6 +9,10 @@
             {
                 "note": "Updated interface to `deployAsync` to include log decode dependencies.",
                 "pr": 1995
+            },
+            {
+                "note": "Updated template to set the `txReceipt` to the contract instance after successful deployment.",
+                "pr": 2024
             }
         ]
     },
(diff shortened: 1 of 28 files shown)


Freydal:
I agree that persisting in the object is not ideal for general usage, but was much less complex for our local solution so I thought I'd start there here.  I tried to find another solution, but blindly cherry-picking a past transaction isn't very easy.

Locally I am also was adding the type to the wrapper directly so that the only custom file we have to maintain is the `contract.handlebars`.  I thought it would be more appropriate in `BaseContract` for a shared solution.

We recently took a pass on outdated dependencies which required us to update the template prompting me to make this contribution.

fabioberger:
Thanks for submitting this @Freydal. Can you elaborate on what you need the `contract creation` tx hash for? 

I have to agree with Gene that I feel most users will just want to contract instance and we'd be complicating the interface by returning both the instance and the transaction hash. Another alternative would be to add a method that returns both.

Freydal:
Currently the main usages are monitoring deployment gas costs in test cases and was storing the transaction hash for possible reference.  I understand that this may need to remain a custom solution.

PR #2026, opened 2019-07-31
Add sandwich video to landing page

fragosti:
## Description

https://dogfood.0xproject.com/

## Checklist:


packages/website/ts/components/sections/landing/hero.tsx
```diff
@@ -13,26 +15,68 @@ import { WebsitePaths } from 'ts/types';
...
+                    isFullWidth={true}
+                    description="0x is an open protocol that enables the peer-to-peer exchange of assets on the Ethereum blockchain."
+                    figure={<LandingAnimation image={<HeroAnimation />} />}
+                    actions={<HeroActions onPlayVideoClick={this._openModalVideo} />}
+                    // announcement={announcement}
+                />
+                <ModalVideo
+                    channel="youtube"
+                    isOpen={isVideoOpen}
+                    videoId="c04eIt3FQ5I"
```
dave4506: should we put the videoId in constants?

packages/website/ts/components/sections/landing/hero.tsx
```diff
@@ -13,26 +15,68 @@ import { WebsitePaths } from 'ts/types';
...
+                    onClose={this._closeModalVideo}
+                    youtube={{
+                        autoplay: 1,
+                        controls: 0,
+                        showinfo: 0,
+                        modestbranding: 1,
```
dave4506: maybe put the video prop settings in constants? fine with either

packages/website/ts/components/sections/landing/hero.tsx
fragosti: Our rule is if it's used twice then it must be a variable, so I think we're good here.

packages/website/ts/components/sections/landing/hero.tsx
fragosti: I think it's good here.

chriskalani:
Looks pretty good! Can we make it bigger though? It's pretty small.

chriskalani:
Also, can we hid everything on top of the video when it first loads? These overlays require the user to mouse over and off to go away.
Screen Shot 2019-07-31 at 8 52 38 PM

PR #2027, opened 2019-08-01
Addition of AssetSwapper Convenience Class to asset-swapper library

dave4506:
## Checklist:


fragosti:
We chatted about this and decided that while having the quote and execution done in one code block is terse, it is a bit confusing from the API perspective and at the end of the day not that useful.

Issue #2012, opened 2019-07-30
Replace diff test in abi-gen with CI check


xianny:
## Current Behavior

`abi-gen` currently has a copy of generated code checked in at `abi-gen/test-cli/expected-output`. The `yarn build` step for this package will generate fresh code at `abi-gen/test-cli/output` (gitignored) and fail a test if it differs from the `expected-output`. Devs have to manually copy `output` to `expected-output` when making any changes that ripple through the generated output.  This results in annoying CI failures unless devs are familiar with the unusual workflow.

## Possible Solution

Get rid of the `expected-output` folder. Remove `output` from gitignore. The build step will continue to update `output` and keep it in sync with any generator changes. Unit tests continue to rely on `output`.

Add a CI step after building that compares freshly generated code to the `output` folder at the current commit. 
```abi-gen$ git diff test-cli/output```
 If there are diffs, it fails. This enforces that changes to generator *and* generated are always in sync, even if devs deviate from the `yarn build` workflow, e.g. by manually generating wrappers.

Devs should manually review changes to `output` in their PRs.

PR #2033, opened 2019-08-03
Generate MD docs for all packages

fabioberger:
## Description

This PR:
- Adds the ability to generate markdown doc pages for our packages. 
- It generates the `reference.md` for the current version of each package that currently has a doc page.
- It adds MD doc generation and S3 uploading to the publish flow.

The Docs 2.0 site will then involve pulling down the latest versions from S3, indexing the MD files on Algolia and bundling them into the site for faster loading times.

What is cool about this approach is that even packages that live _outside_ the monorepo can be searchable/renderable on the Docs 2.0 site (e.g., the TS client for Mesh https://github.com/0xProject/0x-mesh/pull/334). 

## Checklist:


feuGeneA:
I haven't reviewed all of this, but I wanted to discourage you from checking in generated code.  (Deja vu? :P) 

If I understand correctly, the workflow is set up such that someone could make changes to some TypeScript and forget to also check in changes to the generated markdown.  This sounds like a recipe for confusion and headache.  The next person who DOES try to update the markdown will have to sift through a bunch of changes that are reflective of prior work, and any reviewer would have to go digging through commit histories in order to make sense of it all.

To me the best thing to do would be to not check these in at all, and to persist them as CircleCI Build Artifacts.  (See usage of `store_artifacts` in `.circleci/config.yml`.)  These generated markdown files _are_ build artifacts, and are not hand-written, maintained code, so I think this is the _right_ approach for them.

However, I realize there's an argument to be made that persisting as Build Artifacts rather than checked-in code does make it difficult to review changes to the generated markdown.  But I would counter that argument by saying that it's always difficult to review a change in rendering (I've always wondered how a web developer reviews changes to the rendering of HTML/JS code in a PR...) and that as a user of a renderer we should just trust that the renderer is tested properly and not doing anything stupid on our behalf.

If you must insist that we keep checking in these markdown files, I suggest adding a test that will fail if someone changes the TypeScript source without also updating the generated markdown.  This is the approach we plan to take for the `abi-gen/test-cli` output, which is described in #2012 .

fabioberger:
@feuGeneA actually the MD docs will only be generated during the publishing flow and auto-commited to Github along with the updated CHANGELOGS. No one will ever need to review them. We only guarantee that the docs at the publish commit are reflective of the packages at that version, not every intermediate state after each PR. 

Do you now agree with the decision to commit them? It's nice to have the docs in the commit history for every version through time.

feuGeneA:
I'm sorry to be a stickler, but I still disagree that they should be checked in.  I love the idea of having a canonical/authoritative place for generated docs, and of having a versioning system that explicitly associates them with particular releases, but I don't think the source tree is the right place for that.  These are artifacts, not code.

It sounds these markdown files are beyond just "build artifacts", and could accurately be described as "release artifacts."  Could we leverage the GitHub Releases .zip files as the authoritative place to find the generated docs associated with a particular version?

To be clear, I'm not trying to be a purist, and I want to advocate pragmatism.  If what you've got here is the most practical solution, then don't let me be a blocker.  (I for one wouldn't know where to start with GitHub Releases and could see myself taking the tech-debt-incurring shortcut of just checking these in.)  But I _do_ feel pretty strongly that a repo is for source code, not artifacts (whether build, release, or otherwise), and that if we have a good-enough and not-too-onerous mechanism for a particular type of artifact then we should utilize it, and since we _do_ have a mechanism at our disposal for "release artifacts" then we should at least _consider_ using it.

feuGeneA:
One other thought:  If you took the GitHub Releases .zip file approach, maybe then you could skip the S3 step and have the Docs 2.0 site pull the markdown straight from GitHub?  I dunno if that makes sense or not, but I wanted to throw it out there for consideration.

dekz:
What is the benefit of having the mdx files checked into the source code? Is it a requirement for something else or just a nice to have?

They're generated on publish. Then versioned and uploaded to s3. They likely also exist in the published npm module. 

I agree with Gene that we shouldn't have artifacts in the repository which aren't required for a usable module, e.g #2044, generated wrappers.

If they're used for some other piece of functionality, could they be retrieved via the npm package which would contain these files?

packages/asset-buyer/docs/reference.mdx
```diff
@@ -0,0 +1,1352 @@
...
+* [getLiquidityForAssetDataAsync](#getliquidityforassetdataasync)
+* [getOrdersAndFillableAmountsAsync](#getordersandfillableamountsasync)
+* [getAssetBuyerForProvidedOrders](#static-getassetbuyerforprovidedorders)
+* [getAssetBuyerForStandardRelayerAPIUrl](#static-getassetbuyerforstandardrelayerapiurl)
+
+## Constructors
+
+###  constructor
+
+\+ **new AssetBuyer**(`supportedProvider`: `SupportedProvider`, `orderProvider`: [OrderProvider](#class-assetbuyer)*
```
dekz: This seems to be missing a closing parenthesis and looks like its picked up the incorrect type for the returns. 
`[OrderProvider](#class-assetbuyer)`

Renders:
> new AssetBuyer(supportedProvider: SupportedProvider, orderProvider: OrderProvider*

packages/asset-buyer/docs/reference.mdx
```diff
@@ -0,0 +1,1352 @@
...
+
+___
+
+###  message
+
+• **message**: *string*
+
+*Inherited from void*
+
+Defined in /Users/rickmorty/Documents/projects/0x/0x-monorepo/node_modules/typedoc/node_modules/typescript/lib/lib.es5.d.ts:974
```
dekz: We should skip the `Defined in` for these primitives that come from `Error`

packages/asset-buyer/docs/reference.mdx
```diff
@@ -0,0 +1,1352 @@
...
+
+• **orderRefreshIntervalMs**: *number*
+
+*Defined in [types.ts:107](https://github.com/0xProject/0x-monorepo/blob/6dd77d5c8/packages/asset-buyer/src/types.ts#L107)*
+
+<hr />
+
+> # Interface: BuyQuote
+
+assetData: String that represents a specific asset (for more info: https://github.com/0xProject/0x-protocol-specification/blob/master/v2/v2-specification.md).
```
dekz: Going forward how do we get these definitions down to the appropriate properties, do we just describe it above the property itself and it will render nicely?

For now we should respect the new line with a double space. Otherwise they all render as one big line.

fabioberger:
@feuGeneA @dekz S3 will hold MD docs for all 0x packages, across multiple repos, and not _only_ for monorepo packages. This makes it a more sensible place then the release zip. Secondly, the release zip is a very user-unfriendly way of browsing the docs for a package, it is hard to find, requires a download and unzip to read. 

The MD doc files within the monorepo is a nice-to-have. It's not used by anything but is more for someone preferring to use Github to explore our packages instead of our website. It's also about keeping the docs for posterity's sake, our site will only show the docs for the last 10 versions of a package. If you want the docs for an older package, you have to browse for it on Github. 

Packaging the docs into the NPM published code is an alternative approach we could consider but it'll increase the bundle size which is not desirable. We could have people browse our S3 bucket but that won't render the markdown nicely. Although I agree with the premise of the argument, I don't think these should cause anyone any grief given that they are _only_ updated during the publish flow, will never be reviewed by a human and won't cause any merge conflicts.

packages/asset-buyer/docs/reference.mdx
fabioberger: Fixed this issue. Thanks for spotting!

dekz:
Packaging the docs into the NPM also makes it available offline where the other options require an internet connection to read the docs.

I don't think the size increase is going to be large in an node module, since the docs can be compressed well as it is non binary data. Does the same practice hold for other ecosystems though? I.e in go/python deps?

I agree that a .zip file is not a great user experience in getting to the docs.

packages/asset-buyer/docs/reference.mdx
fabioberger: Got rid of these. Thanks.

packages/asset-buyer/docs/reference.mdx
fabioberger: We need to investigate if there is a way to make them render properly. This is the first pass, and we'll keep iterating on it.

packages/asset-buyer/docs/reference.mdx
```diff
@@ -0,0 +1,1300 @@
...
+* [takerAssetData](#takerassetdata)
+* [takerFee](#takerfee)
+
+## Properties
+
+###  exchangeAddress
+
+• **exchangeAddress**: *string*
+
+*Inherited from void*
```
xianny: Hmm do we need these `Inherited from void` lines? They are everywhere, any way to get rid of them?

fabioberger:
Thanks Xianny. Yes, your understand is correct. There is one issue with your proposal, and that is that it would require a user wanting to read the docs to clone our repo, run `yarn install` and then `yarn generate_md_docs`, which isn't as easy as browsing files on Github. 

I really appreciate everyone's commitment to avoiding adding generated artifacts to the repo but in this case, there isn't really a way to get an easy end-developer experience otherwise (minus @dekz's approach of bundling them into NPM packages which comes at the cost of increased bundle sizes).

feuGeneA:
Is increased bundle sizes really an issue? We're talking human-readable text, which gets compressed typically 10:1 or better.  (Not saying it's *not* an issue, just saying, perhaps do a quick measure before deciding it *is*.)

feuGeneA:
> they are _only_ updated during the publish flow, will never be reviewed by a human and won't cause any merge conflicts.

How can you be so sure?  Would you be up for a simple test to enforce it?  Like #2012 .

If really never to be reviewed by a human, please tag them as `generated` in `.gitattributes`?

feuGeneA:
> Would you be up for a simple test to enforce it? Like #2012 .

quick tldr: during CI runs do a `git diff folder-holding-generated-markdown`.  if return code is non-zero, the markdown generation process produced changes that were not checked in, so the build should fail.

xianny:
> There is one issue with your proposal, and that is that it would require a user wanting to read the docs to clone our repo, run yarn install and then yarn generate_md_docs, which isn't as easy as browsing files on Github.

Yeah, that's correct. I feel pretty neutral about having generated docs on GH. My suggestion was if we decide **not** to commit the generated docs, I think it would be easier to prompt users to generate locally from the code they are already looking at, than to have them search for our S3, git release files, etc. I think bundling with npm is the most user-friendly solution but self-generating is easy if we are really really opposed to that. 

I think adding docs to GH does add some overhead, but I also think being able to see diffs between docs could be useful.

fabioberger:
Thanks for all the ideas and discussions. Turns out the MD files can get quite large (1MB uncompressed for `@0x/contract-wrappers` for instance). I've therefore decided to keep the MD files in Github, but to add a CI test to make sure they are never modified in a regular PR. This will avoid all merge-conflicts.

packages/asset-swapper/docs/reference.mdx
```diff
@@ -0,0 +1,2872 @@
...
+
+# Class: ExchangeSwapQuoteConsumer
+
+### Hierarchy
+
+* **ExchangeSwapQuoteConsumer**
+
+### Implements
+
+* [SwapQuoteConsumerBase](#exchangesmartcontractparams)*›
```
dekz: This type anchor link seems off:
`[SwapQuoteConsumerBase](#exchangesmartcontractparams)`

Also a weird hanging `*>` here.

PR #2038, opened 2019-08-07
Migrate to Web3.py v5

### LINKED ISSUE #2029: Update Python dependencies to use mainline Web3.py rather than the 0x fork

feuGeneA:
Fixes #2029 

This PR started as simply a migration, but turned into a bit more.

First of all, the migration exposed a lack of a `clean` command in the `sra_client` package, which was causing headaches and which I've fixed here.  Also, a recent change I made to build components in parallel simply doesn't make sense; they need to be built in dependency order, and I've fixed that here.  And, it turns out that the dependency order stored in the monorepo scripts was mixed up, so I fixed it and also provided a script to automatically produce that dependency order by inspecting the packages' dependencies.

The migration also turned up two problems with the SRA client tests.  One of those problems is that currently Launch Kit Backend does not support EIP-55 checksummed addresses.  This is a problem because Web3.py v5 enforces all addresses coming through its interfaces to be checksummed, which throws a kink in the SRA tests.  I've raised https://github.com/0xProject/0x-launch-kit-backend/issues/73 to track this problem, and @dekz has created a custom build (docker image, used in CI here) that allows my tests to _post_ an order with a checksummed address, but because retrieving an order still conveys a `toLower()`ed address, the SRA example has been temporarily modified to checksum the maker address in the order retrieved from launch kit before it proceeds to fill that order.

The other problem with the SRA tests is with the migration to the new Launch Kit Backend docker image.  Previously these tests were using `0xorg/launch-kit-ci`, but that was a one-off thing created just for CI, back before there was a regularly maintained docker image of Launch Kit.  In this PR I changed the setup to use `0xorg/launch-kit-backend` instead, since it's regularly maintained/updated.  However, there is an outstanding problem with this setup in CircleCI, which I will need to raise with their support team.  If you look at [the commit where I changed docker images](https://github.com/0xProject/0x-monorepo/pull/2038/commits/a851e63bc537c94277a472a0e6f1d806b702da84), you'll see that there are two places the image has changed: in `.circleci/config.yml`, and in `python-packages/sra_client/test/relayer/docker-compose.yml`.  Note that both before and after that commit, those two files have an effectively identical configuration.  However, while the new `docker-compose.yml` file works great for local development, the `.circleci/config.yml` file is simply not working.  The docker image never produces any output in the build run, and the tests time out trying to connect to the relayer endpoint.  For this reason, this PR disables the SRA client tests for CI.  (They remain enabled for local development.)

Finally, I happened to notice some problems with the generated docs for the contract wrappers, which I've also fixed in this PR.  The latest build of the contract wrapper documentation is [here](https://83353-92181371-gh.circle-artifacts.com/0/home/circleci/repo/python-packages/contract_wrappers/build/docs/html/index.html).  The problems were that (a) we weren't generating docs for all of the wrappers (just Exchange and ERC20Token), and (b) the docs weren't including the contract wrappers' member variables, which are the instantiated, named method classes.

-   [x] Migrate to new API.
-   [x] ~Update `launch-kit-ci` docker image with recent changes to JSON schema for `address` data type.  Web3.py 5.x is now handling *all* addresses with checksum formatting, and the current `launch-kit-ci` image is still using the old JSON schema which only allows lower-case hex digits in an address.~ Punted on this, and raised https://github.com/0xProject/0x-launch-kit-backend/issues/73 .


packages/abi-gen/test-cli/expected-output/python/abi_gen_dummy/__init__.py
```diff
@@ -446,8 +446,8 @@ def validate_and_normalize_inputs(self, _hash: bytes, v: int, r: bytes, s: bytes
     def call(self, _hash: bytes, v: int, r: bytes, s: bytes, tx_params: Optional[TxParams] = None) -> str:
         """Execute underlying contract method via eth_call.
 
-        test that devdocs will be generated and     that multiline devdocs will
-        look okay
+        test that devdocs will be generated andthat multiline devdocs will look
```
xianny: looks like we need 1 space here!

python-packages/contract_wrappers/setup.py
```diff
@@ -246,11 +246,9 @@ def run(self):
         "0x-contract-artifacts",
         "0x-json-schemas",
         "0x-order-utils",
-        "0x-web3",
+        "web3",
```
xianny: 👍

python-packages/contract_wrappers/test/test_erc20_wrapper.py
```diff
@@ -5,7 +5,8 @@
 import pytest
 
 from zero_ex.contract_addresses import NETWORK_TO_ADDRESSES, NetworkId
-from zero_ex.contract_wrappers import ERC20Token, TxParams
+from zero_ex.contract_wrappers import TxParams
+from zero_ex.contract_wrappers.erc20_token import ERC20Token
 
 
 MAX_ALLOWANCE = int("{:.0f}".format(Decimal(2) ** 256 - 1))
```
RLlane2126: Duplicate of #

python-packages/contract_wrappers/test/test_erc20_wrapper.py
```diff
@@ -5,7 +5,8 @@
 import pytest
 
 from zero_ex.contract_addresses import NETWORK_TO_ADDRESSES, NetworkId
-from zero_ex.contract_wrappers import ERC20Token, TxParams
+from zero_ex.contract_wrappers import TxParams
+from zero_ex.contract_wrappers.erc20_token import ERC20Token
```
RLlane2126: from zero_ex.contract_wrappers import ERC20Token, TxParams
X2

PR #2042, opened 2019-08-08
Wrapper/Reentrancy unit tests, etc.

dorothy-zbornak:
## Description
tape

This PR introduces a couple things:

### Exchange Wrapper Unit Tests

You'll find many new unit tests for Exchange wrapper functions in [wrapper_unit_tests.ts](../blob/feature/3.0/exchange/wrapper-tests/contracts/exchange/test/wrapper_unit_tests.ts).

This is accomplished by overriding a few Exchange internal functions in [TestWrappersFunctions.sol](../blob/feature/3.0/exchange/wrapper-tests/contracts/exchange/contracts/test/TestWrapperFunctions.sol) so we test **only** the logic of the wrapper functions themselves. 

### Automatic Reentrancy Guard Testing

You'll find some lovely tests in [reentrancy_tests.ts](../blob/feature/3.0/exchange/wrapper-tests/contracts/exchange/test/reentrancy_tests.ts) that automatically :robot: checks if public Exchange functions are secured by the `nonReentrant` modifier. This voodoo is accomplished by the [ReentrancyTester.sol](../blob/feature/3.0/exchange/wrapper-tests/contracts/exchange/contracts/test/ReentrancyTester.sol) contract. 

By default, it will check all mutator public functions that are not whitelisted in [test/utils/constants.ts](../blob/feature/3.0/exchange/wrapper-tests/contracts/exchange/test/utils/constants.ts), and it'll even check that those whitelisted functions are actually reentrant. This makes it much harder to accidentally expose a reentrant public function.

I've also *removed* a lot of the legacy reentrancy tests scattered around the tests because I feel like the new tests cover them. This is might be controversial. :innocent:

### Other goodies:
- `marketBuyOrders` and `marketSellOrders` have been removed and their `noThrow` variants have replaced them.
- Signature validating functions in `MixinSignatureValidator` will now throw an `InvalidSigner` if the `signerAddress == 0`. This is to prevent people from using `0x0` as a `signerAddress` and passing bad signature data to `ecrecover()`, which will happily return `0x0`.
- There is no more assembly in `fillOrderNoThrow` :tada: 
- Updated the legacy `Wallet` signature type (which we still support in V3) to use the magic bytes pattern from v2.1 of the Exchange.
- Added all mutator functions to `constants.ExchangeFunctions` in `@0x/contracts/exchange/test/utils.ts`.
- Blacklisted yet another file from solhint because of `abi.decode` syntax: `contracts/coordinator/contracts/src/MixinCoordinatorApprovalVerifier.sol`. :disappointed: 

### Some Thoughts
I think it might be cleaner to move ALL integration tests out of the exchange package and into its own package (like a `protocol-tests` package?), when we have more time. There's a lot of integration tests that pull in many external contract packages so things are looking a little spaghetti...

## Testing instructions

You know the drill.

## Types of changes

## Checklist:


contracts/exchange-libs/contracts/src/LibExchangeSelectors.sol
```diff
@@ -22,9 +22,6 @@ pragma solidity ^0.5.9;
 contract LibExchangeSelectors {
```
dorothy-zbornak: I think we can probably ditch this file, right?

contracts/exchange/test/utils/constants.ts
```diff
@@ -1,22 +1,14 @@
...
-        'BATCH_MATCH_ORDERS',
-        'BATCH_MATCH_ORDERS_WITH_MAXIMAL_FILL',
-        'CANCEL_ORDER',
-        'BATCH_CANCEL_ORDERS',
-        'CANCEL_ORDERS_UP_TO',
-        'PRE_SIGN',
-        'SET_SIGNATURE_VALIDATOR_APPROVAL',
+    // These are functions not secured by the `nonReentrant`, directly or
+    // indirectly (by calling a function that has the modifier).
+    REENTRANT_FUNCTIONS: [
```
dorothy-zbornak: Here's your reentrancy whitelist.

hysz:
> I think it might be cleaner to move ALL integration tests out of the exchange package and into its own package.

I really like the idea of consolidating integration tests into a single package. We'd get an overview of protocol-wide test coverage without having to sift through all of our packages.

contracts/exchange/contracts/test/ReentrancyTester.sol
```diff
@@ -0,0 +1,130 @@
...
+                isReentrant = true;
+            }
+        }
+    }
+
+    /// @dev Calls a public function to check if it is reentrant.
+    /// Because this function uses the `nonReentrant` modifier, if
+    /// the function being called is also guarded by the `nonReentrant` modifier,
+    /// it will revert when it returns.
+    function testReentrantFunction(bytes calldata fnCallData)
```
hysz: Is this only called by `isReentrant`? If so, why not apply the `nonReentrant` modifier  directly to `isReentrant`?

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -45,6 +45,8 @@ contract TestValidatorWallet is
 {
     using LibBytes for bytes;
 
+    bytes4 private constant LEGACY_WALLET_MAGIC_VALUE = 0xb0671381;
```
hysz: Nit - add comment describing generator (`keccak256(...)`)

contracts/exchange/test/wrapper_unit_tests.ts
```diff
@@ -0,0 +1,1129 @@
+import { ReferenceFunctions as LibReferenceFunctions } from '@0x/contracts-exchange-libs';
```
hysz: Readability lvl 💯

contracts/exchange/contracts/test/ReentrancyTester.sol
dorothy-zbornak: Mostly for UX reasons. Because our reentrancy guard is lazy, only the last non-reentrant call will revert. If I made `isReentrant()` `nonReentrant`, I wouldn't be able to capture and compare the revert data within the contract and provide a nice boolean result.

contracts/exchange-libs/contracts/src/LibExchangeSelectors.sol
dorothy-zbornak: Removed it entirely from the Exchange contracts. Coordinator and DevUtils still use it quite heavily, so I added it to the backlog.

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -158,19 +162,19 @@ contract TestValidatorWallet is
...
     /// @param signature Proof of signing.
-    /// @return isValid `true` if the signature check succeeds.
+    /// @return `LEGACY_WALLET_MAGIC_VALUE` if the signature check succeeds.
     function isValidSignature(
         bytes32 hash,
         bytes memory signature
     )
         public
-        returns (bool isValid)
+        returns (bytes4 magicValue)
```
abandeali1: We should update the `IWallet` contract as well.

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -158,19 +162,19 @@ contract TestValidatorWallet is
...
     {
         ValidatorAction action = _hashActions[hash];
         if (action == ValidatorAction.Reject) {
-            isValid = false;
+            magicValue = 0x0;
```
abandeali1: Let's use `bytes4(0)` here to be more explicit.

contracts/exchange/test/reentrancy_tests.ts
```diff
@@ -0,0 +1,120 @@
+import { blockchainTests, constants, describe, expect, hexRandom } from '@0x/contracts-test-utils';
+import { BigNumber } from '@0x/utils';
+import { DataItem, MethodAbi, TupleDataItem } from 'ethereum-types';
+import * as _ from 'lodash';
+
+import { artifacts, ReentrancyTesterContract } from '../src';
+
+import { constants as TestConstants } from './utils/constants';
+
+blockchainTests.resets('Reentrancy Tests', env => {
```
abandeali1: Nice! Having this all in one place is so much cleaner.

contracts/exchange/test/utils/types.ts
```diff
@@ -8,20 +8,26 @@ export interface AbiDecodedFillOrderData {
...
+    FillOrderNoThrow = 'fillOrderNoThrow',
     MarketBuyOrders = 'marketBuyOrders',
-    MarketBuyOrdersNoThrow = 'marketBuyOrdersNoThrow',
     MarketSellOrders = 'marketSellOrders',
-    MarketSellOrdersNoThrow = 'marketSellOrdersNoThrow',
     MatchOrders = 'matchOrders',
-    CancelOrder = 'cancelOrder',
-    BatchCancelOrders = 'batchCancelOrders',
-    CancelOrdersUpTo = 'cancelOrdersUpTo',
+    MatchOrdersWithMaximalfill = 'matchOrdersWithMaximalFill',
```
abandeali1: Typo: `MatchOrdersWithMaximalFill`

contracts/exchange/test/wrapper_unit_tests.ts
abandeali1: +1

contracts/coordinator/.solhintignore
```diff
@@ -0,0 +1,2 @@
+# solhint can't parse `abi.decode` syntax.
+contracts/src/MixinCoordinatorApprovalVerifier.sol
```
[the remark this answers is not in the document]
dorothy-zbornak: I tried but solhint throws a parser error when it hits that contract. I'm a little confused about why this just started happening and with only certain forms of `abi.decode()`. I don't think it's just my setup because the same thing happened on CI (that's how I first noticed it).

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -210,6 +216,8 @@ contract MixinSignatureValidator is
         view
         returns (bool isValid)
     {
+        _assertValidSigner(signerAddress, orderHash, signature);
```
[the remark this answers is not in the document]
dorothy-zbornak: This is my fault. I did the refactoring on `MixinSignatureValidator`. I agree, it doesn't make sense. I'll make the change in this PR.

contracts/exchange/contracts/test/ReentrancyTester.sol
```diff
@@ -0,0 +1,130 @@
...
+contract ReentrancyTester is
+    Exchange
+{
+    using LibBytes for bytes;
+
+    // solhint-disable no-empty-blocks
+    // solhint-disable no-unused-vars
+    constructor ()
+        public
+        Exchange(0x74657374)
```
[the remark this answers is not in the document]
dorothy-zbornak: Just "test" in hex. I forgot to add the comment.

contracts/exchange/contracts/test/ReentrancyTester.sol
```diff
@@ -0,0 +1,130 @@
...
+    {}
+
+    /// @dev Calls a public function to check if it is reentrant.
+    function isReentrant(bytes calldata fnCallData)
```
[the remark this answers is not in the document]
jalextowle: I think that it would make sense to start compiling a list of functions that need this modifier so that we can manually check during code reviews. That would also make it really easy when we start using static analysis.

contracts/exchange/contracts/test/ReentrancyTester.sol
```diff
@@ -0,0 +1,130 @@
...
+
+    /// @dev Overridden to do nothing.
+    function _executeTransaction(
+        ZeroExTransaction memory transaction,
+        bytes memory signature
+    )
+        internal
+        returns (bytes memory resultData)
+    {
+        resultData = new bytes(0);
```
jalextowle: Do you need to do this? Wouldn't `resultData` already be set to an equivalent pointer?

contracts/exchange/contracts/test/ReentrantERC20Token.sol
```diff
@@ -1,388 +0,0 @@
-/*
```
jalextowle: 🔥 So glad we're getting rid of this contract

contracts/exchange/test/reentrancy_tests.ts
jalextowle: ➕

contracts/exchange/test/signature_validator.ts
```diff
@@ -236,6 +236,18 @@ blockchainTests.resets('MixinSignatureValidator', env => {
             return expect(tx).to.revertWith(expectedError);
         });
 
+        it('should revert when signer is an EOA and SignatureType=Wallet', async () => {
```
jalextowle: Thanks for adding this test case. Another that we should probably add is: `should revert when signer is a contract that returns the empty bytes (or true) and SignatureType=Wallet`

contracts/exchange/contracts/test/ReentrancyTester.sol
dorothy-zbornak: Ah, yeah, good point. I forgot the EVM would just point them to `0x60`.

PR #2043, opened 2019-08-08
MatchOrders Unit Tests

jalextowle:
## Description

This PR adds unit tests for the `MixinMatchOrders` contract. The functions that are tested are:
- `_assertValidMatch`
- `_calculateCompleteFillBoth`
- `_calculateCompleteRightFill`
- `_calculateMatchedFillResults`
- `_calculateMatchedFillResultsWithMaximalFill`
- `_settleMatchedOrders`

## Testing instructions

## Types of changes

* New Tests
* Refactor

## Checklist:


contracts/exchange/test/internal.ts
```diff
@@ -6,12 +6,14 @@ import {
     expect,
     hexRandom,
     LogDecoder,
+    OrderFactory,
```
dorothy-zbornak: Can we do this without `OrderFactory` or any signature validation? This feels a lot less like unit tests if we need to validate signatures as well.

contracts/exchange/test/internal.ts
```diff
@@ -48,16 +50,358 @@ blockchainTests('Exchange core internal functions', env => {
...
+                makerAssetAmount: Web3Wrapper.toBaseUnitAmount(50, 18),
+                takerAssetAmount: Web3Wrapper.toBaseUnitAmount(100, 18),
+            });
+            const rightOrder = await orderFactoryRight.newSignedOrderAsync({
+                makerAssetAmount: Web3Wrapper.toBaseUnitAmount(100, 18),
+                takerAssetAmount: Web3Wrapper.toBaseUnitAmount(50, 18),
+            });
+            return expect(testExchange.assertValidMatch.callAsync(leftOrder, rightOrder)).to.be.fulfilled('');
+        });
+    });
```
dorothy-zbornak: How about tests for the two cases where `_assertValidMatch()` can overflow?

contracts/exchange/test/internal.ts
```diff
@@ -439,5 +2357,240 @@ blockchainTests('Exchange core internal functions', env => {
...
+            expect(logs[6].args.amount).bignumber.to.be.eq(matchedFillResults.right.takerFeePaid);
+
+            // Right taker fee -> right fee recipient
+            expect(logs[7].args.orderHash).to.be.eq(leftOrderHash);
+            expect(logs[7].args.assetData).to.be.eq(leftOrder.takerFeeAssetData);
+            expect(logs[7].args.from).to.be.eq(takerAddress);
+            expect(logs[7].args.to).to.be.eq(leftOrder.feeRecipientAddress);
+            expect(logs[7].args.amount).bignumber.to.be.eq(matchedFillResults.left.takerFeePaid);
+        });
+    });
```
dorothy-zbornak: Would be nice to check the case where `takerFeePaid` can overflow.

contracts/exchange/test/internal.ts
```diff
@@ -439,5 +2357,240 @@ blockchainTests('Exchange core internal functions', env => {
             expect(logs[3].args.amount).to.bignumber.eq(fillResults.makerFeePaid);
         });
     });
+
+    blockchainTests('settleMatchOrders', () => {
+        const getLeftOrder = () => {
```
dorothy-zbornak: Why have a separate left and right order generators? I don't think `_settleMatchOrders()` cares if the orders are complementary.

contracts/exchange/test/internal.ts
```diff
@@ -48,16 +50,358 @@ blockchainTests('Exchange core internal functions', env => {
     let testExchange: TestExchangeInternalsContract;
     let logDecoder: LogDecoder;
     let senderAddress: string;
+    let makerAddressLeft: string;
+    let makerAddressRight: string;
+    let orderFactoryLeft: OrderFactory;
+    let orderFactoryRight: OrderFactory;
 
     before(async () => {
```
dorothy-zbornak: This makes me sad :panda_face: @ how this file ballooned by 6X, and so much of it is visually similar. There must be a way to cut this down.

contracts/exchange/test/internal.ts
```diff
@@ -48,16 +50,358 @@ blockchainTests('Exchange core internal functions', env => {
...
+            return expect(testExchange.assertValidMatch.callAsync(leftOrder, rightOrder)).to.be.fulfilled('');
+        });
+    });
+
+    blockchainTests('calculateCompleteFillBoth', () => {
+        /**
+         * Asserts that the actual result of calling `calculateCompleteFillBoth()` is as expected.
+         */
+        async function assertCalculateCompleteFillBothAsync(
+            expectedMatchedFillResults: MatchedFillResults,
```
dorothy-zbornak: I don't think this pattern of passing in explicit results really works for us, because it is so unmaintainable. I would rather we implement *well-tested* reference functions (see `../src/reference_functions.ts`). This would cut down the line count of this file by A LOT, and we'd only have to update the reference function when the implementation changes. Not to mention that reference functions will be useful when we upgrade some of the older integration tests.

contracts/exchange/test/internal.ts
jalextowle: I think that I can probably remove this. I was just using it for convenience, but I will remove

contracts/exchange/test/internal.ts
jalextowle: I'm glad you mentioned this. I didn't write these intentionally because these cases should (and are) covered in the unit tests for `_safeMul`. This said, if it makes us feel better, I'm happy to add them back in.

contracts/exchange/test/internal.ts
jalextowle: Same as above

contracts/exchange/test/internal.ts
jalextowle: You're right. I did it to make the tests seem more realistic, but I'll change it

contracts/exchange/test/internal.ts
jalextowle: I'll look into it. I think that creating a helper to reduce the size of the `expectedMatchedFillResults` objects could help with this.

contracts/exchange/test/internal.ts
jalextowle: I'm not sure I agree with you on this one. In my opinion, well testing the reference functions would actually mean testing them the same as I test the real functions here. Otherwise, I don't think that the assertions we make are very strong, and it's easy to introduce logical errors into the reference functions. We should discuss this during standup.

contracts/exchange/test/internal.ts
dorothy-zbornak: It's kind of more to protect against regressions, i.e., that SafeMath is still being used if we refactor this function.

jalextowle: That's a fair point. I'll go ahead and change it. This said, I wonder if it makes sense to enforce a linter rule that normal arithmetic can't be used. I think that there is a rule in `solhint` that would enforce this. Do you think that would be sufficient to prevent regressions?

contracts/exchange/test/internal.ts
dorothy-zbornak: Just override `_isValidOrderWithHashSignature()` in the contract to get around it.

contracts/exchange/test/internal.ts
dorothy-zbornak: I can dig it.

contracts/exchange/test/internal.ts
jalextowle: It also might make sense to start splitting `internal` up. I think that reducing the code in this PR is a fair point, but it's going to be pretty big even after some refactoring. I'd even go so far as to say that we could do a unit test file per significant function (like a `calculateFillResults` or `calculateMatchedFillResults`). We could keep internal around just for the smaller or easier to test functions.

contracts/exchange/contracts/test/TestExchangeInternals.sol
```diff
@@ -39,6 +39,16 @@ contract TestExchangeInternals is
         Exchange(chainId)
     {}
 
+    function assertValidMatch(
```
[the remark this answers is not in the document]
abandeali1: Let's actually backlog this until we remove `internal.ts` in a different PR, as discussed offline.

contracts/exchange/test/internal.ts
```diff
@@ -42,15 +43,101 @@ blockchainTests('Exchange core internal functions', env => {
...
     const randomAssetData = () => hexRandom(36);
+    const randomUint256 = () => new BigNumber(hexRandom(constants.WORD_LENGTH));
     let testExchange: TestExchangeInternalsContract;
     let logDecoder: LogDecoder;
     let senderAddress: string;
+    let makerAddressLeft: string;
+    let makerAddressRight: string;
+
+    function createMatchedFillResultsFromFilled(
+        leftMakerAssetFilledAmount: BigNumber,
```
dorothy-zbornak: Same.

contracts/exchange/test/internal.ts
```diff
@@ -60,16 +147,322 @@ blockchainTests('Exchange core internal functions', env => {
...
+            return expect(testExchange.assertValidMatch.callAsync(leftOrder, rightOrder)).to.be.fulfilled('');
+        });
+    });
+
+    blockchainTests('calculateCompleteFillBoth', () => {
+        /**
+         * Asserts that the actual result of calling `calculateCompleteFillBoth()` is as expected.
+         */
+        async function assertCalculateCompleteFillBothAsync(
+            expectedMatchedFillResults: MatchedFillResults,
```
dorothy-zbornak: Same

contracts/exchange/test/internal.ts
```diff
@@ -42,15 +43,101 @@ blockchainTests('Exchange core internal functions', env => {
...
+            right: {
+                ...EMPTY_FILL_RESULTS,
+                makerAssetFilledAmount: rightMakerAssetFilledAmount,
+                takerAssetFilledAmount: rightTakerAssetFilledAmount,
+            },
+        };
+    }
+
+    function createMatchedFillResults(
+        leftMakerAssetFilledAmount: BigNumber,
```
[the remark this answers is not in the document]
jalextowle: Fair point. I'll fix this before merging using a similar pattern to `makeOrder(Partial<Order>)`.

contracts/exchange/test/internal.ts
jalextowle: Forgot about fixing this one, I'm gonna make a note of this and backburner it

PR #2053, opened 2019-08-10
Fixes for broken `tox` runs

feuGeneA:
`tox` is used to test a package in a clean "virtual environment", that is, an environment that is freshly created from packages downloaded from PyPI (and not messed up by locally-installed/manipulated packages in the local Python environment).

I ran `tox` on each of our packages and found some problems.  The most glaring was some missing source files in the published `contract_wrappers.py` package! :grimacing: Due to the urgency, and since it was really just a problem with the packaging process and not really the code itself, I went ahead and re-packaged/re-published it to fix it.  The respective change in this PR is just a version bump.

I also fixed up a few other problems with the `tox` runs in here too.

Now all of the `tox` runs, for all the packages, are running clean.  But we need to have a conversation about how we make this part of some CI pipeline, a conversation which I intend to raise at the multilang sync on Thursday.  (I'm OOO Monday.)  (It's not as simple as just adding a test to the monorepo CI runs...)


feuGeneA:
`static-test` CI failure is affecting all builds right now...

PR #2054, opened 2019-08-10
subproviders: Change MetamaskSubprovider eth_sign to personal_sign

Arctek:
## Description

Fix for MetamaskSubprovider to use personal_sign over manually prefixing eth_sign, so proxied hardware wallets work through MetaMask (Ledger/Trezor).

See: 
https://metamask.github.io/metamask-docs/API_Reference/Signing_Data/Personal_Sign
https://github.com/MetaMask/eth-ledger-bridge-keyring/blob/master/index.js#L192
https://github.com/MetaMask/eth-trezor-keyring/blob/master/index.js#L211
https://github.com/MetaMask/eth-sig-util/blob/master/index.js#L250

## Types of changes

## Checklist:


dekz:
Thanks @Arctek I think we're in a space now where we can consider using `personal_sign` now that EIP712 has landed support in all Metamask installations. 

For history, Metamask has had an incorrect implementation of `eth_sign` for years, claiming it is a dangerous method. The spec was updated years ago as well to add the prefix but the Metamask implementation remained the same. This architecture meant the implementation of hardware wallet support in Metamask never received the `eth_sign` mapping where it should've. Really we've just been in a losing battle trying to get Metamask to change its behaviour here and support proper  `eth_sign` as per the spec.

`personal_sign` has no spec and no EIP and renders as an unverifiable string  when it's a hash. 

If we can guarantee that `personal_sign` maps to hardware wallet `eth_sign` then I think we can merge in this PR. @fabioberger have any thoughts here?

fabioberger:
@Arctek thanks for this contribution. I'm down with making this change. Can you please add a CHANGELOG entry for this change?

Arctek:
Sure just did - though CircleCI is failing, not sure if this is something I can fix.

PR #2055, opened 2019-08-11
Refactor library usage

abandeali1:
## Description

- Refactors `LibMath`, `LibFillResults`, `LibOrder`, `LibZeroExTransaction`, and `LibSafeMath` to be actual libraries. Note that the old `SafeMath` still exists so that contracts that weren't yet refactored don't break. This can be removed in a separate PR.
- Simplifies `LibEIP712ExchangeDomain` to only calculate the domain hash in its constructor and nothing else.
- Moves `calculateFillResults` and `calculateMatchedFillResults` to `LibFillResults`. All functions in `LibFillResults` were also refactored to be pure. *This refactor should probably be one of the focus areas of this review.*
- Removes all selector constants in favor of the `IExchange(address(0)).fillOrder.selector` pattern.
- Refactors `LibEIP712.hashEIP712Domain` to use assembly. This was actually written pre-vuln, but might be a good discussion to have for all of the hashing functions.
- Refactors `contracts-gen` to accept a `compiler.json` with no specified contracts. In this case, boilerplate for all contracts in the package will be generated. The `contracts:gen` command is also added into the `build` command for each package, so it no longer needs to be explicitly called.
- Removes the `_` prefix from internal function names in actual libraries.
- Moves `LibExchangeRichErrors` to the `exchange-libs` package.
- Random cleanup.

## Checklist:


contracts/exchange-libs/contracts/src/LibFillResults.sol
```diff
@@ -45,17 +48,358 @@ contract LibFillResults is
...
+
     /// @dev Adds properties of both FillResults instances.
-    ///      Modifies the first FillResults instance specified.
-    /// @param totalFillResults Fill results instance that will be added onto.
-    /// @param singleFillResults Fill results instance that will be added to totalFillResults.
-    function _addFillResults(FillResults memory totalFillResults, FillResults memory singleFillResults)
+    /// @param fillResults1 The first FillResults.
+    /// @param fillResults2 The second FillResults.
+    /// @return The sum of both fill results.
+    function addFillResults(
```
jalextowle: I'm a fan of this not being a destructive function. I think that we should make this the norm (if it isn't already).

contracts/exchange-libs/contracts/src/LibFillResults.sol
```diff
@@ -45,17 +48,358 @@ contract LibFillResults is
...
+        // Case 1.
+        //   If the left maker can buy more than the right maker can sell, then only the right order is fully filled.
+        //   If the left maker can buy exactly what the right maker can sell, then both orders are fully filled.
+        // Case 2.
+        //   If the left maker cannot buy more than the right maker can sell, then only the left order is fully filled.
+        // Case 3.
+        //   If the left maker can buy exactly as much as the right maker can sell, then both orders are fully filled.
+        if (leftTakerAssetAmountRemaining > rightMakerAssetAmountRemaining) {
+            // Case 1: Right order is fully filled
+            matchedFillResults = _calculateCompleteRightFill(
```
jalextowle: Same as above. Really glad this is non-destructive now.

contracts/exchange-libs/contracts/src/LibFillResults.sol
```diff
@@ -45,17 +48,358 @@ contract LibFillResults is
...
+            matchedFillResults.right.takerAssetFilledAmount = LibMath.safeGetPartialAmountCeil(
+                rightOrder.takerAssetAmount,
+                rightOrder.makerAssetAmount,
+                leftTakerAssetAmountRemaining // matchedFillResults.right.makerAssetFilledAmount
+            );
+        } else { 
+            // leftTakerAssetAmountRemaining == rightMakerAssetAmountRemaining
+            // Case 3: Both orders are fully filled. Technically, this could be captured by the above cases, but
+            //         this calculation will be more precise since it does not include rounding.
+            matchedFillResults = _calculateCompleteFillBoth(
```
jalextowle: Ditto

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -100,23 +101,23 @@ contract OrderValidationUtils is
...
                     takerAssetAmount
                 );
-                uint256 transferableMakerFeeToTakerAmount = _getPartialAmountFloor(
+                uint256 transferableMakerFeeToTakerAmount = LibMath.getPartialAmountFloor(
                     transferableMakerFeeAssetAmount,
                     makerFee,
                     takerAssetAmount
                 );
-                transferableTakerAssetAmount = _min256(transferableMakerToTakerAmount, transferableMakerFeeToTakerAmount);
+                transferableTakerAssetAmount = LibSafeMath.min256(transferableMakerToTakerAmount, transferableMakerFeeToTakerAmount);
```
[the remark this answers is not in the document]
abandeali1: I felt like that was less readable, but I could go either way.

contracts/dev-utils/test/lib_asset_data.ts
```diff
@@ -465,20 +465,6 @@ describe('LibAssetData', () => {
         });
     });
 
-    describe('getERC721TokenOwner', async () => {
```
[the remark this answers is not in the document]
abandeali1: The logic was just moved into the `if...else` block rather than its own function, similar to what we were doing with all of the other similar logic.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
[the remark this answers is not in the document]
jalextowle: I don't have a huge preference either way.

contracts/exchange-libs/test/lib_order.ts
```diff
@@ -0,0 +1,75 @@
...
+    let libOrderContract: TestLibOrderContract;
+    let order: Order;
+    let chainId: number;
+    before(async () => {
+        libOrderContract = await TestLibOrderContract.deployFrom0xArtifactAsync(
+            artifacts.TestLibOrder,
+            env.provider,
+            env.txDefaults,
+        );
+        chainId = await providerUtils.getChainIdAsync(env.provider);
```
dorothy-zbornak:
```suggestion
        chainId = await env.getChainIdAsync();
```
But, also, why even get the env's chainId? Doesn't seem to matter what this value is so long as it isn't `1337`.

contracts/exchange-libs/test/lib_zero_ex_transaction.ts
```diff
@@ -0,0 +1,77 @@
...
+    let libZeroExTransactionContract: TestLibZeroExTransactionContract;
+    let zeroExTransaction: ZeroExTransaction;
+    let chainId: number;
+    before(async () => {
+        libZeroExTransactionContract = await TestLibZeroExTransactionContract.deployFrom0xArtifactAsync(
+            artifacts.TestLibZeroExTransaction,
+            env.provider,
+            env.txDefaults,
+        );
+        chainId = await providerUtils.getChainIdAsync(env.provider);
```
dorothy-zbornak: Again, getting a legitimate `chainId` doesn't seem to matter?

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -207,11 +203,11 @@ contract MixinExchangeCore is
...
 
         // Get amount of takerAsset to fill
-        uint256 remainingTakerAssetAmount = _safeSub(order.takerAssetAmount, orderInfo.orderTakerAssetFilledAmount);
-        uint256 takerAssetFilledAmount = _min256(takerAssetFillAmount, remainingTakerAssetAmount);
+        uint256 remainingTakerAssetAmount = order.takerAssetAmount.safeSub(orderInfo.orderTakerAssetFilledAmount);
+        uint256 takerAssetFilledAmount = LibSafeMath.min256(takerAssetFillAmount, remainingTakerAssetAmount);
 
         // Compute proportional fill amounts
-        fillResults = _calculateFillResults(order, takerAssetFilledAmount);
+        fillResults = LibFillResults.calculateFillResults(order, takerAssetFilledAmount);
```
dorothy-zbornak: :+1:

contracts/utils/contracts/src/LibSafeMath.sol
```diff
@@ -0,0 +1,90 @@
+pragma solidity ^0.5.9;
+
+import "./LibRichErrors.sol";
+import "./LibSafeMathRichErrors.sol";
+
+
+library LibSafeMath {
```
hysz: 🙏

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -21,24 +21,12 @@ pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
 import "@0x/contracts-exchange-libs/contracts/src/LibZeroExTransaction.sol";
+import "@0x/contracts-exchange-libs/contracts/src/LibEIP712ExchangeDomain.sol";
 import "@0x/contracts-erc20/contracts/src/interfaces/IERC20Token.sol";
 import "@0x/contracts-utils/contracts/src/LibBytes.sol";
 import "@0x/contracts-utils/contracts/src/LibEIP1271.sol";
 
 
-interface ISimplifiedExchange {
```
dorothy-zbornak: :pray:

contracts/utils/contracts/src/LibEIP712.sol
```diff
@@ -46,21 +47,44 @@ contract LibEIP712 {
...
+        // Assembly for more efficient computing:
+        // keccak256(abi.encodePacked(
+        //     _EIP712_DOMAIN_SEPARATOR_SCHEMA_HASH,
+        //     keccak256(bytes(name)),
+        //     keccak256(bytes(version)),
+        //     chainId,
+        //     uint256(verifyingContractAddress)
+        // ))
+
+        assembly {
```
hysz: This looks good to me ~ do you know roughly how much gas this saves?

contracts/utils/contracts/test/TestConstants.sol
```diff
@@ -1,57 +0,0 @@
-/*
```
dorothy-zbornak: Didn't even know this contract existed.

packages/contracts-gen/src/contracts-gen.ts
```diff
@@ -50,16 +53,18 @@ function generateCompilerJSONContractsList(
...
-        if (contract.endsWith(SOLIDITY_EXTENSION)) {
-            // If it's already a relative path - NO-OP.
-            return contract;
-        } else {
-            // If it's just a contract name - resolve it and rewrite.
-            return new NameResolver(contractsDir).resolve(contract).path;
-        }
-    });
-    compilerJSON.contracts = _.sortBy(compilerJSON.contracts);
+    if (compilerJSON.contracts !== undefined && compilerJSON.contracts !== ALL_CONTRACTS_IDENTIFIER) {
```
dorothy-zbornak: NOICE! :+1:

contracts/utils/contracts/src/LibEIP712.sol
abandeali1: I don't think it's super significant, maybe 1000 or so. Also note that this function only gets called once in the `Exchange` contract. I mostly did this for consistency pre-vuln since all of the other hashing is done in assembly, but it would probably be a good time to formalize what we want to do with all of the hashing as a whole.

contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -87,24 +87,27 @@ contract LibOrder is
     /// @dev Calculates Keccak-256 hash of the order.
     /// @param order The order structure.
     /// @return Keccak-256 EIP712 hash of the order.
-    function getOrderHash(Order memory order)
-        public
-        view
+    function getOrderHash(Order memory order, bytes32 eip712ExchangeDomainHash)
+        internal
```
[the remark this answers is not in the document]
abandeali1: I discussed this with Fabio and we agreed that we can just put this in the `DevUtils` contract. Do you think that is sufficient?

contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -87,24 +87,27 @@ contract LibOrder is
     /// @dev Calculates Keccak-256 hash of the order.
     /// @param order The order structure.
     /// @return Keccak-256 EIP712 hash of the order.
-    function getOrderHash(Order memory order)
-        public
-        view
+    function getOrderHash(Order memory order, bytes32 eip712ExchangeDomainHash)
```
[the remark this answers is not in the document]
abandeali1: If we go strictly by the EIP712 spec, these should probably be `getTypedDataHash` and `getStructHash`. I think the terminology is pretty confusing though, I wonder if there is a better distinction we can use.

PR #2059, opened 2019-08-12
Fill Order Assertion Wrapper

hysz:
## Description

This is an assertion wrapper for `exchange.fillOrder`. It encapsulates all pre- and post- assertions we want to make when calling `fillOrder`. This includes validating: order state, fill results, balance changes, and events.

This PR includes:
1. Assertion wrapper for `exchange.fillOrder`. This may be used as a template for future assertion wrappers.
3. A set of `Balance Store` classes which can be recycled by other assertion wrappers / test utilities to track/compare/analyze balances throughout a test.
2. A reduction in size by 35% of `test/core.ts`

Notes:
Much of this code was reworked from the contract wrappers and the match order tester. There still exists a lot of redundant test code in those classes, and elsewhere. As we migrate to this pattern we will eventually be able to easily eliminate most of the redundant code (_massive RIP_). We can probably shrink the match order tester down to about 1/4 its current size.

## Checklist:

### DIFF
--- a/contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
+++ b/contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
@@ -1,0 +1,282 @@
+import { artifacts as proxyArtifacts , ERC1155ProxyWrapper, ERC20Wrapper, ERC721Wrapper } from '@0x/contracts-asset-proxy';
+import { artifacts as erc20Artifacts } from '@0x/contracts-erc20';
+import { artifacts as erc721Artifacts } from '@0x/contracts-erc721';
+import {
+    chaiSetup,
+    FillEventArgs,
+    LogDecoder,
+    OrderStatus,
+    orderUtils,
+    Web3ProviderEngine,
+} from '@0x/contracts-test-utils';
+import { orderHashUtils } from '@0x/order-utils';
+import { FillResults, SignedOrder } from '@0x/types';
+import { BigNumber } from '@0x/utils';
+import { Web3Wrapper } from '@0x/web3-wrapper';
+import * as chai from 'chai';
+import { LogWithDecodedArgs, TransactionReceiptWithDecodedLogs, ZeroExProvider } from 'ethereum-types';
+import * as _ from 'lodash';
+
+import { artifacts, ExchangeContract } from '../../src';
+import { BalanceStore } from '../balance_stores/balance_store';
+import { BlockchainBalanceStore } from '../balance_stores/blockchain_balance_store';
+import { LocalBalanceStore } from '../balance_stores/local_balance_store';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+export class FillOrderWrapper {
+    private readonly _exchange: ExchangeContract;
+    private readonly _blockchainBalanceStore: BlockchainBalanceStore;
+    private readonly _web3Wrapper: Web3Wrapper;
+    /**
+     * Simulates matching two orders by transferring amounts defined in
+     * `transferAmounts` and returns the results.
+     * @param orders The orders being matched and their filled states.
+     * @param takerAddress Address of taker (the address who matched the two orders)
+     * @param tokenBalances Current token balances.
+     * @param transferAmounts Amounts to transfer during the simulation.
+     * @return The new account balances and fill events that occurred during the match.
+     */
+    public static simulateFillOrder(
+        signedOrder: SignedOrder,
+        takerAddress: string,
+        opts: { takerAssetFillAmount?: BigNumber } = {},
+        initBalanceStore: BalanceStore,
+        partialFillResults: Partial<FillResults>,
+    ): [FillResults, FillEventArgs, BalanceStore] {
+        const balanceStore = LocalBalanceStore.create(initBalanceStore);
+        const takerAssetFillAmount =
+            opts.takerAssetFillAmount !== undefined ? opts.takerAssetFillAmount : signedOrder.takerAssetAmount;
+        const fillResults = FillOrderWrapper._createFillResultsFromPartial(
+            partialFillResults,
+            signedOrder,
+            takerAssetFillAmount,
+        );
+        const fillEvent = FillOrderWrapper.simulateFillEvent(signedOrder, takerAddress, fillResults);
+        // Taker -> Maker
+        balanceStore.transferAsset(
+            takerAddress,
+            signedOrder.makerAddress,
+            fillResults.takerAssetFilledAmount,
+            signedOrder.takerAssetData,
+        );
+        // Taker -> Fee Recipient
+        balanceStore.transferAsset(
+            takerAddress,
+            signedOrder.feeRecipientAddress,
+            fillResults.takerFeePaid,
+            signedOrder.takerFeeAssetData,
+        );
+        // Maker -> Taker
+        balanceStore.transferAsset(
+            signedOrder.makerAddress,
+            takerAddress,
+            fillResults.makerAssetFilledAmount,
+            signedOrder.makerAssetData,
+        );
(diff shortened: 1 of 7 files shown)


contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -0,0 +1,282 @@
...
+import { LocalBalanceStore } from '../balance_stores/local_balance_store';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+export class FillOrderWrapper {
+    private readonly _exchange: ExchangeContract;
+    private readonly _blockchainBalanceStore: BlockchainBalanceStore;
+    private readonly _web3Wrapper: Web3Wrapper;
+    /**
```
jalextowle: Nit: Should we start adding lines between functions? Personally I'd prefer it.

contracts/exchange/test/balance_stores/blockchain_balance_store.ts
```diff
@@ -0,0 +1,39 @@
+import { ERC1155ProxyWrapper, ERC20Wrapper, ERC721Wrapper } from '@0x/contracts-asset-proxy';
+import * as _ from 'lodash';
+
+import { BalanceStore } from './balance_store';
+
+export class BlockchainBalanceStore extends BalanceStore {
```
jalextowle: This is really nice abstraction 👍

contracts/exchange/test/balance_stores/local_balance_store.ts
```diff
@@ -0,0 +1,106 @@
+import { assetDataUtils } from '@0x/order-utils';
+import { AssetProxyId } from '@0x/types';
+import { BigNumber } from '@0x/utils';
+import * as _ from 'lodash';
+
+import { BalanceStore } from './balance_store';
+
+export class LocalBalanceStore extends BalanceStore {
```
jalextowle: This is also super nice! I'm super happy about the fact that we removed the balance logic from the simulators, and it seems like it would be trivial to use this with `MatchOrders`

contracts/exchange/test/balance_stores/blockchain_balance_store.ts
```diff
@@ -0,0 +1,39 @@
+import { ERC1155ProxyWrapper, ERC20Wrapper, ERC721Wrapper } from '@0x/contracts-asset-proxy';
+import * as _ from 'lodash';
```
dorothy-zbornak: Not used.

contracts/exchange/test/balance_stores/local_balance_store.ts
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Creates a new balance store based on an existing one.
+     * @param balanceStore Existing balance store whose values should be copied.
+     */
+    public static create(sourceBalanceStore?: BalanceStore): LocalBalanceStore {
```
dorothy-zbornak: Why not just have this behavior in the constructor?

contracts/exchange/test/core.ts
```diff
@@ -390,37 +404,7 @@ describe('Exchange core', () => {
         });
 
         it('should not emit transfer events for transfers where from == to', async () => {
-            const txReceipt = await exchangeWrapper.fillOrderAsync(signedOrder, makerAddress);
```
jalextowle: 🙏

contracts/exchange/test/balance_stores/local_balance_store.ts
```diff
@@ -0,0 +1,106 @@
...
+                        } else {
+                            // Transfer a fungible.
+                            const _tokenId = tokenId.toString(10);
+                            fromBalances.fungible[_tokenId] = fromBalances.fungible[_tokenId].minus(tokenAmount);
+                            toBalances.fungible[_tokenId] = toBalances.fungible[_tokenId].plus(tokenAmount);
+                        }
+                    }
+                }
+                // sort NFT's by name
+                toBalances.nonFungible = _.sortBy(toBalances.nonFungible);
```
dorothy-zbornak:
```suggestion
                toBalances.nonFungible.sort();
```

contracts/test-utils/src/types.ts
```diff
@@ -137,3 +137,38 @@ export interface MatchOrder {
     leftSignature: string;
     rightSignature: string;
 }
+
```
dorothy-zbornak: Aren't all these just used by the new classes in the `exchange` package? Why not keep them close? Either that or put the new classes in `test-utils`.

contracts/exchange/test/core.ts
```diff
@@ -56,6 +55,8 @@ chaiSetup.configure();
 const expect = chai.expect;
 const blockchainLifecycle = new BlockchainLifecycle(web3Wrapper);
 
+import { FillOrderWrapper } from './assertion_wrappers/fill_order_wrapper';
+
 // tslint:disable:no-unnecessary-type-assertion
 describe('Exchange core', () => {
```
dorothy-zbornak: Can we upgrade these tests to use the new `blockchainTest()` and import semantics while we're here? Should shave off a whole ~10 lines!

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
hysz: Nice, I'm happy you mentioned this lol. I've been doing this for consistency but I'm more than happy to start adding newlines now. :)

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -0,0 +1,282 @@
...
+        );
+        const fillEvent = FillOrderWrapper.simulateFillEvent(signedOrder, takerAddress, fillResults);
+        // Taker -> Maker
+        balanceStore.transferAsset(
+            takerAddress,
+            signedOrder.makerAddress,
+            fillResults.takerAssetFilledAmount,
+            signedOrder.takerAssetData,
+        );
+        // Taker -> Fee Recipient
```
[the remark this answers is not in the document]
hysz: Mm yeah it shouldn't make a difference but probably good to keep it consistent. Good catch 🙏

contracts/test-utils/src/types.ts
hysz: I'm trying to start aggregating types here, like what we do in `@0x/types. I also agree that the new classes should live here as well, but at present it creates a circular dependency so I put them in `exchange` for now. I'll add a note to the code for future reference.

contracts/exchange/test/balance_stores/local_balance_store.ts
hysz: Hmm I guess it seems awkward to me to include an instance of a class as a parameter to its own constructor. I guess we could parameterize the raw balance struct, although I'm hoping to abstract away the inner mechanics as much as possible.

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -0,0 +1,282 @@
...
+        interface RawFillEventArgs {
+            orderHash: string;
+            makerAddress: string;
+            takerAddress: string;
+            makerAssetFilledAmount: string;
+            takerAssetFilledAmount: string;
+            makerFeePaid: string;
+            takerFeePaid: string;
+        }
+        const actualFills = (_.filter(receipt.logs, ['event', 'Fill']) as any) as Array<
```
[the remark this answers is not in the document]
hysz: Nice 😎

contracts/exchange/test/core.ts
hysz: Darn tootin we can ! #RIP

PR #2060, opened 2019-08-13
Added DevUtils Wrappers and migrations

dekz:
## Description

Added DevUtils Wrappers and migrations

s3://ganache-snapshots.0x.org/0x_ganache_snapshot-4.3.0.zip

| network | address                                                                                                                            |
| ------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| mainnet |  [0x92d9a4d50190ae04e03914db2ee650124af844e6](https://etherscan.io/address/0x92d9a4d50190ae04e03914db2ee650124af844e6#code) |
| ropsten | [0x3e0b46bad8e374e4a110c12b832cb120dbe4a479](https://ropsten.etherscan.io/address/0x3e0b46bad8e374e4a110c12b832cb120dbe4a479#code) |
| rinkeby | [0x2d4a9abda7b8b3605c8dbd34e3550a7467c78287](https://rinkeby.etherscan.io/address/0x2d4a9abda7b8b3605c8dbd34e3550a7467c78287#code) |
| kovan   | [0x1e3616bc5144362f95d72de41874395567697e93](https://kovan.etherscan.io/address/0x1e3616bc5144362f95d72de41874395567697e93#code)   |
| ganache | 0x38ef19fdf8e8415f18c307ed71967e19aac28ba1                                                                                         |

PR #2061, opened 2019-08-13
Add Python wrapper for DevUtils contract

feuGeneA:
For #2060


feuGeneA:
`submit-coverage` failure is just coveralls down for maintenance

PR #2068, opened 2019-08-14
Update contract-wrappers everywhere

xianny:
## Description

This updates all monorepo packages to the new version of contract-wrappers.

Packages affected:
* asset-buyer
* asset-swapper
* 0x.js
* website

It also adds a new convenience method `validateAndSendTransactionAsync` to generated wrappers to replace the `shouldValidate : true` option in the previous version of contract wrappers. 

## Testing instructions

Existing unit tests haven't been changed, so if they pass we should be alright.

## Types of changes

 * Bug fix (non-breaking change which fixes an issue) -->

## Checklist:


contracts/extensions/package.json
```diff
@@ -48,7 +48,6 @@
     "homepage": "https://github.com/0xProject/0x-monorepo/contracts/extensions/README.md",
     "devDependencies": {
         "@0x/abi-gen": "^4.1.0",
-        "@0x/contract-wrappers": "9.1.7",
```
xianny: not used in this package

packages/asset-buyer/package.json
```diff
@@ -38,8 +38,9 @@
     "homepage": "https://github.com/0xProject/0x-monorepo/packages/asset-buyer/README.md",
     "dependencies": {
         "@0x/assert": "^2.1.3",
+        "@0x/asset-swapper": "^1.0.1",
```
fabioberger: How does `asset-buyer` use `asset-swapper`?

packages/contract-wrappers/src/contract_wrappers.ts
```diff
@@ -69,6 +74,8 @@ export class ContractWrappers {
      */
     public coordinator: CoordinatorWrapper;
 
+    public constants: { [key: string]: any } = wrapperConstants;
```
fabioberger: Let's not do this. There was one constant needed in a single file, let's just add the constant to that package instead.

packages/contract-wrappers/src/utils/calldata_optimization_utils.ts
```diff
@@ -0,0 +1,44 @@
+import { SignedOrder } from '@0x/types';
+import * as _ from 'lodash';
+
+const NULL_BYTES = '0x';
+
+export const calldataOptimizationUtils = {
```
fabioberger: This doesn't belong in `contract-wrappers`. Why did it end up in this package?

packages/website/ts/blockchain.ts
```diff
@@ -329,7 +330,7 @@ export class Blockchain {
     public async getUnavailableTakerAmountAsync(orderHash: string): Promise<BigNumber> {
         utils.assert(orderHashUtils.isValidOrderHash(orderHash), 'Must be valid orderHash');
         utils.assert(this._contractWrappers !== undefined, 'ContractWrappers must be instantiated.');
-        const unavailableTakerAmount = await this._contractWrappers.exchange.getFilledTakerAssetAmountAsync(orderHash);
+        const unavailableTakerAmount = await this._contractWrappers.exchange.filled.callAsync(orderHash, {});
```
fabioberger: Why do we need to pass an empty anonymous object as the second parameter?

packages/asset-buyer/package.json
xianny: Oops, this is leftover. Will remove

packages/contract-wrappers/src/utils/calldata_optimization_utils.ts
dekz: It used to live in the there for the Forwarder contract wrapper

packages/website/ts/blockchain.ts
xianny: it's the txData - we can set it to default to an empty object!

packages/contract-wrappers/src/utils/calldata_optimization_utils.ts
fabioberger: Right, but since it's not used by the Forwarder contract wrapper anymore, I'd vote to move it to wherever it's actually being used.

packages/website/ts/blockchain.ts
```diff
@@ -2,11 +2,13 @@ import {
     BlockRange,
```
fabioberger: @xianny did you build and load the website and exercise the functionality changed? We do not have test coverage for the website so CI won't catch if anything broke. You need to do some manual QA.

packages/contract-wrappers/src/utils/calldata_optimization_utils.ts
dekz: @abandeali1 @hysz is this still useful with the optimised ABI encoder?

hysz: @dekz this is no longer necessary as it optimized encoding happens by default.

xianny: removed 🔥

packages/website/ts/blockchain.ts
xianny: good catch @fabioberger. I'm getting @fragosti to test the changed components to Portal on staging. I built and ran the other changes locally and they seem fine.

PR #2069, opened 2019-08-14
MixinTransaction Unit Tests

jalextowle:
## Description

This PR introduces unit tests for the `MixinTransaction` contract.

## Testing instructions

## Types of changes

* Unit Tests

## Checklist:


contracts/exchange/contracts/test/TestTransactions.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    constructor ()
+        public
+        Exchange(1337)
+    {} // solhint-disable-line no-empty-blocks
+
+    // This fallback function will succeed if the bool `shouldSucceedCall` has been set
+    // to true, and will fail otherwise. It will return returndata `fallbackReturnData`
+    // in either case.
+    function ()
```
dorothy-zbornak: Hmm, I'm not sure if this is the clearest way to accomplish these tests and I'm not convinced that they test reentrancy intuitively. There's also some setup you need to do for each test, so I think the contract may be more complicated than it needs to be.

What if, instead, we had a function:

```solidity
function executable(bool shouldSucceed, bytes memory callData, bytes memory returnData)
    public returns (bytes memory)
{
    emit ExecutableCalled(...);
    require(shouldSucceed, 'EXECUTABLE_FAILED');
    if (callData.length != 0) {
        (bool didSucceed, bytes memory callResultData) = address(this).call(callData);
        if (!didSucceed) {
            assembly { revert(add(callResultData, 0x20), mload(callResultData)) }
        }
    }
    return bytes;
}
```

This way:
- The function's intent is clear.
- You get log entries to assert.
- You do not rely on any state to setup.
- You can easily test reentrancy (to any depth) by just encoding another
`executeTransaction()` call inside `callData`.

Does this make sense?

contracts/exchange/contracts/test/TestTransactions.sol
```diff
@@ -0,0 +1,110 @@
...
+        ZeroExTransaction memory,
+        bytes32,
+        address,
+        bytes memory
+    )
+        internal
+        view
+        returns (bool)
+    {
+        return shouldBeValid;
```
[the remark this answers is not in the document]
jalextowle: I think that you're right. I'd have a lot more control if I did it that way.

contracts/exchange/contracts/test/TestTransactions.sol
jalextowle: This makes a lot of sense. I'll work on switching it over.

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+            const transaction2 = {
+                ...EMPTY_ZERO_EX_TRANSACTION,
+                expirationTimeSeconds: new BigNumber(currentTimestamp).plus(10),
+                data: callData,
+                signerAddress: accounts[0],
+                domain,
+            };
+            const transactionHash2 = transactionHashUtils.getTransactionHashHex(transaction2);
+            const abiEncoder = AbiEncoder.createMethod('TransactionError', ['uint8', 'bytes32']);
+            const errorData = abiEncoder.encode([
```
dorothy-zbornak: The `RevertError` type should have an `encode()` method that does this for you.

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+                from: accounts[0],
+            });
+            return expect(tx).to.revertWith(expectedError);
+        });
+
+        it('should revert if reentrancy occurs in the middle of an executeTransaction call and msg.sender != signer for both calls', async () => {
+            const validSignature = randomSignature();
+
+            const currentTimestamp = await getLatestBlockTimestampAsync();
+            const transaction1 = {
```
dorothy-zbornak: Can we call these txs something like `innerTx` and `outerTx`?

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+                ...EMPTY_ZERO_EX_TRANSACTION,
+                expirationTimeSeconds: new BigNumber(currentTimestamp).plus(10),
+                data: getExecutableCallData(true, constants.NULL_BYTES, constants.NULL_BYTES),
+                domain,
+            };
+            const transactionHash = transactionHashUtils.getTransactionHashHex(transaction);
+
+            // Use the transaction in execute transaction.
+            await expect(
+                transactionsContract.executeTransaction.sendTransactionAsync(transaction, validSignature),
```
dorothy-zbornak: This technically doesn't wait for the tx to be mined (even though it's probably all the same on ganache).
```suggestion
                transactionsContract.executeTransaction.awaitTransactionSuccessAsync(transaction, validSignature),
```
There's a couple other spots where this is done, too.

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+            );
+            const tx = transactionsContract.executeTransaction.sendTransactionAsync(transaction, validSignature);
+            return expect(tx).to.revertWith(expectedError);
+        });
+
+        it('should revert if the signer != msg.sender and the signature is not valid', async () => {
+            const currentTimestamp = await getLatestBlockTimestampAsync();
+            const transaction = {
+                ...EMPTY_ZERO_EX_TRANSACTION,
+                expirationTimeSeconds: new BigNumber(currentTimestamp).plus(10),
```
dorothy-zbornak: This is getting a little repetitive. How about having `EMPTY_ZERO_EX_TRANSACTION`  hold some timestamp in the distant future? Like `new BigNumber(Math.floor(_.now() / 1000) + 86400)`.

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+                signerAddress: accounts[1], // This is different than the account that will be used to send.
+                data: callData,
+                domain,
+            };
+            const transactionHash = transactionHashUtils.getTransactionHashHex(transaction);
+
+            // Verify that the returndata of the transaction is 0xDEADBEEF
+            const result = await transactionsContract.executeTransaction.callAsync(transaction, validSignature, {
```
dorothy-zbornak: The `TransactionHelper` class in `@0x/contracts-test-utils` might reduce some of the code in blocks where you do a `callAsync`, `sendTransactionAsync`, and `getTxWithDecodedLogsAsync` since it does all three.

contracts/exchange/contracts/test/TestTransactions.sol
```diff
@@ -0,0 +1,95 @@
...
+    // is only meant to be used during testing.
+    function executable(
+        bool shouldSucceed,
+        bytes memory data,
+        bytes memory returnData
+    )
+        public
+        returns (bytes memory)
+    {
+        emit ExecutableCalled(data, returnData);
```
dorothy-zbornak: I think it might be super cool to log the current context address, too.

contracts/exchange/test/transactions_unit_tests.ts
```diff
@@ -0,0 +1,694 @@
...
+            // Create an abiEncoder for bytes. This will be used to decode the result and encode what
+            // is expected.
+            const abiEncoder = AbiEncoder.create('bytes');
+
+            // Ensure that the result contains the abi-encoded bytes "0xdeadbeef"
+            const encodedDeadbeef = abiEncoder.encode('0xdeadbeef');
+            expect(result.length).to.be.eq(2);
+            expect(
+                result[0] ===
+                    '0x0000000000000000000000000000000000000000000000000000000000000020'.concat(
```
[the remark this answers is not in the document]
jalextowle: I agree. My rationale for not doing this is that a fix to the AbiEncoder will actually allow us to use that for this check. At the moment, I have to do this because `AbiEncoder.create('bytes').decode(result[0])` will not correctly decode the bytes (it will actually just return the 32 byte slot containing the length). I've added an Asana backlog task for this, so it should actually be completed at some point. What do you think of this?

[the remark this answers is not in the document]
dorothy-zbornak: If it's a hassle, nbd.

PR #2072, opened 2019-08-16
Run CI in an environment that includes Python, and test generated code

### LINKED ISSUE #2036: Change CircleCI job `test-rest` to use a container with Python available
This is needed in order to be able to run tests of Python code generated by the command-line tests of `abi-gen`. 

Currently we're generating Python code as part of that command-line test (`abi-gen/test-cli`), but we're not even attempting to parse that generated code, much less test it.  With this change we will be able to write tests that exercise the wrappers generated via `abi-gen/test-cli`.

While we're making that change, we should also enable linting of that generated code.  Specifically, we should be checking it with `mypy` and `pylint` at the very least, but likely also `pycodestyle` and `pydocstyle`, and using `bandit` wouldn't hurt too.  And, before doing any linting, we'll need to use `black` to reformat the code.

feuGeneA:
Despite the small amount of code change here, this is actually a pretty big operational change to CI.

Before, we were running builds and tests using the node.js docker image provided by CircleCI.  But we've outgrown that, as we have a need to run Python to exercise code generated by command-line tests of `abi-gen`.

Now, we're using https://github.com/nikolaik/docker-python-nodejs , which appears to be published by "just some guy," which kinda feels a little crazy to switch to.  But, it's up to date, seemingly quite popular (500+K docker hub pulls), and, most importantly, it works! :grin:

Fixes #2036 .

-   [x] Change CI build and test runs to use a docker image that includes Python.
-   [x] Add a minimum viable test of generated Python code to `abi-gen/test-cli`.
-   [x] Fix approximately 4 code quality problems identified by Pylint.  One is shown in the most recent CI failure.  The other 3 I experienced locally.  (I guess different versions of Pylint between me and the build container?)  I think all 4 are quite minor.


.circleci/config.yml
```diff
@@ -4,18 +4,23 @@ jobs:
...
+            # HACK(feuGeneA): commented out this hack as we're changing
+            # from a circleci-maintained container to a different
+            # container, and this hack may not apply anymore, as
+            # suggested by the non-existance of `/home/circleci/.bashrc`
+            # when running the command below.
+            # - run:
+            #       # HACK(albrow): Without this, yarn commands will sometimes
+            #       # fail with a "permission denied" error.
+            #       name: Set npm path
+            #       command: npm set prefix=/home/circleci/npm && echo 'export PATH=$HOME/circleci/npm/bin:$PATH' >> /home/circleci/.bashrc
```
fabioberger: And? Are we able to remove it?

feuGeneA: I don't think we can say for sure just yet.  It was an intermittent problem.  But I have _not_ experienced that intermittent problem since disabling it, and I've been through probably 6 or 7 builds now.

packages/abi-gen/test-cli/fixtures/pylintrc
```diff
@@ -1,5 +1,5 @@
 [MESSAGES CONTROL]
-disable=C0330,line-too-long,fixme,too-few-public-methods,too-many-ancestors,duplicate-code
+disable=C0330,line-too-long,fixme,too-few-public-methods,too-many-ancestors,duplicate-code,blacklisted-name
```
feuGeneA: Before, it was complaining about names `foo` and `bar` being used.

PR #2075, opened 2019-08-16
Add marketBuy/SellOrdersFillOrKill() to Exchange

dorothy-zbornak:
## Description

This PR introduces two new functions to the Exchange's `MixinWrapperFunctions`, which is the subject of [ZEIP-50](https://github.com/0xProject/ZEIPs/issues/50):
- **`marketSellOrdersFillOrKill()`**
    - After calling `marketSellOrdersNoThrow()`, ensures that `takerAssetFilledAmount > takerAssetFillAmount`.
- **`marketBuyOrdersFillOrKill()`**
    - After calling `marketBuyOrdersNoThrow()`, ensures that `makerAssetFilledAmount > makerAssetFillAmount`.

#### Re-introducing the `noThrow()` suffix
In PR #2042, the `marketBuyOrders()` and `marketSellOrders()` functions were replaced by `marketBuyOrdersNoThrow()` and `marketSellOrdersNoThrow()`, assuming the formers' names.

This was probably a bad idea because the function signatures were identical, meaning it was very possible that a dapp would call those functions expecting the old behavior.

So I've added the `noThrow()` suffixes back to those functions in this PR.

#### `RevertError` types can now take array fields
The new `marketBuy/SellOrdersFillOrKill()` functions needed to throw a rich revert that holds the order hashes of all the orders passed. So now this is a thing.

#### Added a `fillAmount` field to existing `IncompleteFillError` rich revert
Made sense.

#### Use `abi.decode()` in `LibTransactionDecoder`
`LibTransactionDecoder` relied heavily on `LibBytes` for decoding individual error parameters. This worked fine for simple types, but `LibBytes` doesn't support reading a `bytes32[]` (which I needed), so I figured we might as well just be consistent just use `abi.decode()` throughout that contract. 

#### Added an `UNLIMITED_CONTRACT_SIZE` environment variable, `@0x/contracts-dev-utils` tests now run with it.
This environment variable is picked up by `@0x/contracts-test-utils` and creates the default `web3Wrapper` instance with `allowUnlimitedContractSize` enabled if set to `true`.

With the addition of the new methods, the `DevUtils` contract is no longer deployable under normal conditions [because it is > 24KB](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-170.md). The `UNLIMITED_CONTRACT_SIZE` is a temporary fix for our tests, but we need to re-architecture this contract to be smaller/deployable inevitably.

## Testing instructions

You know the drill, so DRILL IT.

## Types of changes

## Checklist:


packages/utils/src/revert_error.ts
```diff
@@ -388,14 +388,16 @@ function declarationToAbi(declaration: string): RevertErrorAbi {
     const [name, args] = m.slice(1);
     const argList: string[] = _.filter(args.split(','));
     const argData: DataItem[] = _.map(argList, (a: string) => {
-        m = /^\s*([_a-z][a-z0-9_]*)\s+([_a-z][a-z0-9_]*)\s*$/i.exec(a);
+        m = /^\s*(([_a-z][a-z0-9_]*)(\[\d*\])*)\s+([_a-z][a-z0-9_]*)\s*$/i.exec(a);
```
jalextowle: Nit: Could you add a comment to this regex? It's getting a bit long, and it is not immediately obvious that this regex just matches an identifier, possibly ending with an array bracket, separated from another identifier by whitespace?

packages/utils/src/revert_error.ts
```diff
@@ -408,10 +410,34 @@ function declarationToAbi(declaration: string): RevertErrorAbi {
 function checkArgEquality(type: string, lhs: ArgTypes, rhs: ArgTypes): boolean {
     if (type === 'address') {
         return normalizeAddress(lhs as string) === normalizeAddress(rhs as string);
-    } else if (type.startsWith('bytes')) {
+    } else if (type === 'bytes' || /^bytes(\d+)$/.test(type)) {
         return normalizeBytes(lhs as string) === normalizeBytes(rhs as string);
     } else if (type === 'string') {
         return lhs === rhs;
+    } else if (/\[\d*\]$/.test(type)) {
```
jalextowle: Very clean! Might be good to add a comment on the larger regexs, but they aren't too bad yet.

contracts/exchange/contracts/src/MixinWrapperFunctions.sol
```diff
@@ -243,13 +242,22 @@ contract MixinWrapperFunctions is
...
+
             // Attempt to sell the remaining amount of takerAsset
             LibFillResults.FillResults memory singleFillResults = fillOrderNoThrow(
                 orders[i],
                 remainingTakerAssetFillAmount,
                 signatures[i]
             );
 
+            // Restore the original `makerAssetData` so we're non-destructive.
+            orders[i].makerAssetData = originalMakerAssetData;
```
abandeali1: Hmm, this change doesn't feel super necessary to me since this isn't intended to be a library or used in external contracts. We also do something similar in `matchOrders`, right? Is there anywhere where we actually reuse the in-memory orders after calling these functions?

packages/utils/test/revert_error_test.ts
```diff
@@ -41,6 +53,50 @@ describe('RevertError', () => {
             const revert2 = new AnyRevertError();
             expect(revert1.equals(revert2)).to.be.true();
         });
+        it('should equate two revert errors with identical array fields', () => {
+            const strings = ['foo', 'bar'];
+            const revert1 = new ArrayRevertError(strings);
+            const revert2 = new ArrayRevertError(strings);
+            expect(revert1.equals(revert2)).to.be.true();
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, there's a semantic difference, but it doesn't really matter in this case. `equals()` will ignore missing fields.

contracts/exchange/contracts/src/MixinWrapperFunctions.sol
dorothy-zbornak: It was to preserve the order hashes in the event of a rich revert, but we won't need it anymore if we aren't returning order hashes.

packages/utils/test/revert_error_test.ts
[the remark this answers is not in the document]
jalextowle: Good to know 👍

contracts/exchange/test/utils/exchange_wrapper.ts
```diff
@@ -143,12 +143,40 @@ export class ExchangeWrapper {
...
+        );
+        const tx = await this._logDecoder.getTxWithDecodedLogsAsync(txHash);
+        return tx;
+    }
+    public async marketSellOrdersFillOrKillAsync(
+        orders: SignedOrder[],
+        from: string,
+        opts: { takerAssetFillAmount: BigNumber; gas?: number },
+    ): Promise<TransactionReceiptWithDecodedLogs> {
+        const txHash = await this._exchange.marketSellOrdersFillOrKill.sendTransactionAsync(
```
hysz: If we use `awaitTransactionSuccessAsync` here then we don't need the log decoder :)

contracts/exchange/test/utils/exchange_wrapper.ts
```diff
@@ -143,12 +143,40 @@ export class ExchangeWrapper {
...
+        );
+        const tx = await this._logDecoder.getTxWithDecodedLogsAsync(txHash);
+        return tx;
+    }
+    public async marketBuyOrdersFillOrKillAsync(
+        orders: SignedOrder[],
+        from: string,
+        opts: { makerAssetFillAmount: BigNumber; gas?: number },
+    ): Promise<TransactionReceiptWithDecodedLogs> {
+        const txHash = await this._exchange.marketBuyOrdersFillOrKill.sendTransactionAsync(
```
hysz: Same note

PR #2077, opened 2019-08-17
Update README.md

feuGeneA:
@abandeali1 encountered problems building `abi-gen`, due to the new Python dependency, and reported that there was no documentation around what was required to build.

> amir [1:42 PM]
> Does `gen_python` need to be a part of the `abi-gen`’s default `build` command? I can’t build the monorepo because of it and there is no documentation on which python/pip versions are needed.
> I think it’s because my latest python/pip versions are mapped to `python3` and `pip3` (edited) 

> amir [1:53 PM]
> It doesn’t pick up my alias for pip=pip3, but it does work when I change pip to pip3 in the `gen_python` command. It feels like we shouldn’t require any python dependencies for totally unrelated tooling. Is there a better place to build the wrappers?

> Gene [2:46 PM]
> this is by design.  i'm open to discussing and tweaking the design, but we did do this (introduce a dependency on python to the normal build pipeline) on purpose.  now that `abi-gen` generates python, it's important that we exercise the code it generates.  (this isn't a failure to build the wrappers for our contracts. it's a failure to build the code generated by the command-line tests of `abi-gen`. so, it's not really unrelated tooling)  we decided that this testing of `abi-gen` should be integrated with the normal build/test pipeline for the component, so that no one forgets/neglects to run it.  in theory we could put it into a separate build/test pipeline, but i'm afraid of the special casing confusing people who are trying to get started working on `abi-gen`.  totally hear you that we need to document what's required, i'll add that to the README.md right now, but really the requirements are just "python 3" (and the `pip` that comes with it).

> amir [3:07 PM]
> Needed to run `pip3 install --upgrade --force pip` to overwrite my pip in `/usr/bin/`

This PR documents the Python dependency, and also spruces up a bunch of other stuff that I noticed was stale.  The Python dependency explanation is explained at the end of the "Contributing" section.


feuGeneA:
> Can we also update the `Usage` section of the top level README since this change affects the entire monorepo? I'm pretty sure the Node version specified there is also out of date.

Great call @abandeali1 , thank you, done.

I wasn't sure what you meant by the Node version comment.  The top-level README suggests node >=6.12, and that matches all of the `package.json` files in the monorepo (including the top-level one).

PR #2078, opened 2019-08-19
Update development to 2.1 isValidSignature with magic salt

dekz:
## Description

Our local development contracts (and therefor ganache snapshots and tools) use the old signatures without magic salt values.

## Checklist:

-   [x] @jalextowle you mentioned wanting to rewrite as solidity?


contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -244,7 +244,17 @@ contract MixinSignatureValidator is
             hash,
             signature
         );
+        bytes32 magicValue = bytes32(bytes4(keccak256("isValidWalletSignature(bytes32,address,bytes)")));
```
abandeali1: Can we add a comment with the `bytes4` constants for both of these signature types?

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -244,7 +244,17 @@ contract MixinSignatureValidator is
...
         assembly {
+            if iszero(extcodesize(walletAddress)) {
```
abandeali1: Let's add a comment that the `extcodesize` checks are redundant, but were added as an extra safety measure.

packages/order-utils/src/signature_utils.ts
```diff
@@ -169,12 +174,16 @@ export const signatureUtils = {
...
 
         const validatorContract = new IValidatorContract(signerAddress, provider);
-        const isValid = await validatorContract.isValidSignature.callAsync(
-            data,
-            signerAddress,
-            validatorSignature.signature,
-        );
-        return isValid;
+        try {
+            const magicValue = await validatorContract.isValidSignature.callAsync(
```
abandeali1: PR #2017 fixed a bug where we weren't checking if the validator contract was approved in the `Exchange` contract. We should really just be calling `Exchange.isValidSignature` here to ensure that is being checked with a single RPC call.

dekz: We should just refactor this to use devutils entirely right? Is that currently checking if the Validator is approved?

dekz: Let's follow up after #2017 in #2083

PR #2079, opened 2019-08-19
abi-gen/test-cli: rm expected-output; use git diff

### LINKED ISSUE #2012: Replace diff test in abi-gen with CI check

feuGeneA:
Fixes #2012 .

Before, any time someone updated the fixture contract (used by `abi-gen/test-cli`), one had to also update the `test-cli/expected-output` folder.  This step served as a check to ensure that known-good code will be present for test runs.

However, this was too onerous.  It required one review by the contributor (before moving from `test-cli/output` to `test-cli/expected-output`), and then another review by the actual reviewer of a PR.  And, it required a manual step, which is too much cognitive overhead for being able to get started contributing to `abi-gen`.

Now, that manual step has been eliminated, but we've retained the strict checking.  Now, `git diff` is used to perform the check, against the working copy.  When a build runs, it will put the `test-cli` output into `test-cli/output`, overwriting what was already there.  If at that point there any changes present (relative to the git branch) then `yarn test` will fail.

In short, now any changes enacting differences in the generated output of `abi-gen` will also need to be accompanied by changes to the `test-cli/output`, or else the build will fail.  And the changes that need to be checked in will be clearly visible to the contributor, due to them being flagged as changed by `git` itself.


packages/abi-gen/package.json
```diff
@@ -38,7 +38,7 @@
         "generate_contract_wrappers": "run-p gen_typescript gen_python",
         "gen_typescript": "abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/partials/**/*.handlebars' --output ./test-cli/output/typescript --backend ethers",
         "gen_python": "pip install black && abi-gen --abis  ${npm_package_config_abis} --template ../../node_modules/@0x/abi-gen-templates/Python/contract.handlebars --partials '../../node_modules/@0x/abi-gen-templates/Python/partials/**/*.handlebars' --output ./test-cli/output/python --language Python",
-        "diff_contract_wrappers": "test-cli/diff.sh ./test-cli/expected-output ./test-cli/output",
+        "diff_contract_wrappers": "cd test-cli/output; git diff --exit-code .",
```
xianny: does this change the current directory after running `yarn diff_contract_wrappers`? That seems undesirable. Would it be possible to do `git diff --exit-code test-cli/output` instead?

feuGeneA: I'm pretty positive that it doesn't make any difference, because I think every invocation of `yarn whatever` starts with its own shell process spawned with the current working directory set to the package directory, and any `cd`ing that happens in that script run is confined to that script run.

But, I really like the suggestion, so I changed it anyways :grin:

xianny: ya you're totally right! It's fine on CI. `yarn test` runs `yarn diff_contract_wrappers` tho so could be confusing for devs.

xianny:
thanks for the edit!

I double checked because I wasn't sure offhand - `output` folder is still in gitignore. We need to remove that if I understand correctly.

 https://github.com/0xProject/0x-monorepo/blob/fix/abi-gen/git-diff-for-test-cli/.gitignore#L137

PR #2081, opened 2019-08-20
CFL page

fragosti:
## Description

Design: https://www.figma.com/file/sqRd1hnxv5fKh5mXjiKZ42/0x.org?node-id=2690%3A11357

Screen Shot 2019-08-19 at 7 26 15 PM

### Deviations from design
- The code editor has 3 tabs. Had to fill in the tabs with code myself so that could use some review.
- Some of the copy was changed. In particular the labels for the metrics were made more specific / changed to be 7 day.
- We just say that the data can be 30 minutes old instead of saying exactly how fresh it is (because the freshness is different for each component which can be confusing).

### TODO
- Improve sample code.
- Refactor after changes in backend have been applied.

## Testing instructions

https://dogfood.0xproject.com/asset-swapper

## Checklist:


packages/website/ts/pages/cfl/cfl_metrics.tsx
```diff
@@ -0,0 +1,197 @@
...
+    private _onTabClick(index: number): void {
+        this.setState({ selectedIndex: index });
+    }
+    private _getSelectedPairData(): CFLMetricsPairData {
+        return this.state.cflMetricsData[this.state.selectedIndex];
+    }
+    private _getLastPrice(): string {
+        const data = this._getSelectedPairData();
+        if (!data.lastTradePrice) {
+            return '—';
```
BMillman19: should we give some indication of loading other than the dashes?

chriskalani: nah

PR #2082, opened 2019-08-20
abi-gen: inline @0x/abi-gen-templates

### LINKED ISSUE #1923: Merge abi-gen-templates into abi-gen
## Expected Behavior

A typical user of `abi-gen` won't want to bring their own templates.  `abi-gen` should use sensible defaults for its `--template` and `--partials` arguments, based on the `--language` specified (`TypeScript` or `Python`).  In order to support such defaults, the template code should live within the `abi-gen` package.

Users should have the _option_ of specifying a template and partials, but they shouldn't be _required_ to.

## Current Behavior

Every invocation of `abi-gen` must specify arguments for its `--template` and `--partials` parameters.  Further, the user of `abi-gen` must "bring their own" templates.  They're probably just using the templates we wrote, but never the less they have to understand where those templates live and how to refer to them.

feuGeneA:
## Description

Fixes #1923.

Inline handlebars templates from @0x/abi-gen-templates, and use those inlines by default, relieving caller of having to specify --template and --partials all the time.

Simplify abi-gen usage in all package.json's, omitting --template and --partials to defer to the defaults.

## Testing instructions

Just build the monorepo, it will get tested.

## Types of changes

## Checklist:

-   [x] Inline @0x/abi-gen-templates into @0x/abi-gen, and use them by default.


packages/abi-gen/package.json
```diff
@@ -19,7 +19,7 @@
         "test:coverage": "nyc npm run test --all && yarn coverage:report:lcov",
         "coverage:report:lcov": "nyc report --reporter=text-lcov > coverage/lcov.info",
         "test_cli": "run-p test_cli:test_typescript diff_contract_wrappers",
-        "test_cli:clean": "rm -rf test-cli/output && rm -rf test-cli/test_typescript/lib",
+        "test_cli:clean": "rm -rf test-cli/test_typescript/lib",
         "test_cli:build": "tsc --project test-cli/tsconfig.json",
```
feuGeneA: This was bothering me because we're cleaning up generated code upon `yarn clean`, which i think is a good idea, but we're also checking in that generated code, so it pollutes the `git status` with missing files.  And I don't think we want to .gitignore those files, right? (If we just ignore them, why even have them checked in?)

xianny: Yeah, good call! This has been bothering me too, thanks for fixing it.

PR #2086, opened 2019-08-21
Remove generated wrappers from contract-wrappers

### LINKED ISSUE #2058: Refactor uses of abi-gen-wrappers
Some packages in the monorepo depend on `abi-gen-wrappers`. Refactor them to use `contract-wrappers` (new version). The latest version of contract-wrappers is abi-generated.

After this is done, we can delete `abi-gen-wrappers`.

Part of #1989

xianny:
## Description

This PR removes generated code from `@0x/contract-wrappers`. Resolves #2058

We initially added generated code to `@0x/contract-wrappers` intending to remove `@0x/abi-gen-wrappers` completely. While attempting to replace dependencies on `@0x/abi-gen-wrappers` with `@0x/contract-wrappers` across all packages, I found a number of circular dependencies, mostly because of unit tests, coordinator wrapper, and some utility functions.

<details>

To deal with this, I started extracting components from `@0x/contract-wrappers` and created two new packages: a tests package and a package for the Coordinator wrapper.
https://github.com/0xProject/0x-monorepo/tree/multilang-2058/remove-abi-gen-wrappers

~If we pursue the direction of stripping `@0x/contract-wrappers` to the minimum where there's not much added to the generated code (thus requiring minimal dependencies), and putting everything we remove into separate modules, I think it's highly likely that we would eventually want to offer a a package that integrates these modules that are intended to be used together. E.g. generated wrappers + manually written coordinator wrapper. **This is what `@0x/contract-wrappers` already offers**, so we could just use it instead of stripping it down only to rewrite it later.~

~I propose the following mental model:~
~- `@0x/abi-gen-wrappers` - individual contract wrappers. low level libs are free to import only the wrappers they need, thus avoiding circular dependencies~
~- `@0x/contract-wrappers` - re-exports the generated wrappers, provides a convenience instantiation class, unit tests, coordinator wrapper, and other stuff that we want to add. We should be careful to only add code that adds new functionality, not just syntactic sugar, otherwise we end up in the same situation as before.~
</details>

For now, we will keep both packages and import the generated code where needed. After V3 release, we will be able to delete much of the handwritten code and can revisit extracting coordinator wrapper into its own package.

## Testing instructions

* Build the monorepo, unit tests should work
* Check that abi-gen-wrappers is updated

## Types of changes

## Checklist:


packages/abi-gen-wrappers/src/generated-wrappers/erc1155_proxy.ts
```diff
@@ -31,100 +31,29 @@ import { assert } from '@0x/assert';
 import * as ethers from 'ethers';
 // tslint:enable:no-unused-variable
 
-export type MultiAssetProxyEventArgs =
-    | MultiAssetProxyAuthorizedAddressAddedEventArgs
-    | MultiAssetProxyAuthorizedAddressRemovedEventArgs
-    | MultiAssetProxyAssetProxyRegisteredEventArgs;
+export type ERC1155ProxyEventArgs =
```
xianny: to be clear, `abi-gen-wrappers` now contains **both** MultiAssetProxyContract and ERC1155ProxyContract. It looks like the file contents are similar enough to confuse git.

packages/abi-gen-wrappers/src/generated-wrappers/static_call_proxy.ts
```diff
@@ -26,30 +26,33 @@ import * as ethers from 'ethers';
 /* istanbul ignore next */
 // tslint:disable:no-parameter-reassignment
 // tslint:disable-next-line:class-name
-export class IValidatorContract extends BaseContract {
```
xianny: `abi-gen-wrappers` contains both IValidatorContract and `StaticCallProxyContract`. git is confusing the diffs.

.gitignore
```diff
@@ -98,7 +98,6 @@ packages/sol-tracing-utils/test/fixtures/artifacts/
 python-packages/contract_artifacts/src/zero_ex/contract_artifacts/artifacts/
 
 # generated contract wrappers
-packages/abi-gen-wrappers/src/generated-wrappers/
```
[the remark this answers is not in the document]
xianny: Yup, we're going to commit these here for now instead of in `contract-wrappers`.

PR #2087, opened 2019-08-21
remove from != to check and update tests

mzhu25:
## Description
Remove `from != to` check from `_dispatchTransferFrom` and update related tests

## Checklist:


mzhu25:
We could also check that the expected Transfer events are emitted, but idk if that's really necessary given what's already checked in `fillOrderAndAssertEffectsAsync`

Issue #1989, opened 2019-07-24
abi-gen-wrappers -> contract-wrappers


xianny:
This issue tracks all the things that need to be done as part of converting @0x/contract-wrappers to auto-generated wrappers. 

- [x] Review https://github.com/0xProject/0x-monorepo/issues/1911 and make a comprehensive list of public interface changes. post to github issue for posterity - DONE in #2040
- [x] Rename abi-gen-wrappers to contract-wrappers. Update README to remove refs to `contract-wrappers` and update description. update changelog to reference github issue of changes. Move over any relevant tests and util libs. 
For now, pin monorepo dependencies on manual version of `contract-wrappers`. The interface has changed a lot and refactoring is complicated in a couple of cases.  DONE #2037
- [x] Publish new version of `contract-wrappers`

AFTER
- [x] Publish documentation (waiting for #2033)
- [x] Upgrade ethereumjs-blockstream to latest version (7.0.0) in @0x/base-contract. The only other packages that use it are @0x/contract-wrappers and @0x/order-watcher (also deprecated)
- [x] Refactor monorepo to use new version of `contract-wrappers` or find workarounds
        * `assetDataUtils.getEtherTokenAssetData` relies on `forwarder.etherTokenAddress` which no longer exists. Affects `@0x/asset-buyer` and `@0x/asset-swapper` - DONE #2068
- [x] ~~Refactor monorepo to replace dependencies on `abi-gen-wrappers` with new `contract-wrappers` #2086~~ 
- [x] ~~Delete `abi-gen-wrappers`~~

PR #2089, opened 2019-08-21
update ethereumjs-blockstream 6.0.0 -> ^7.0.0

xianny:
## Description

Updates ethereumjs-blockstream.  `6.0.0` will be removed after we update coordinator-server deps in https://github.com/0xProject/0x-coordinator-server/issues/14

Relates to #1989

## Checklist:

### DIFF
--- a/packages/base-contract/src/subscription_manager.ts
+++ b/packages/base-contract/src/subscription_manager.ts
@@ -101,17 +101,23 @@ export class SubscriptionManager<ContractEventArgs, ContractEvents extends strin
         const logWithDecodedArgs = abiDecoder.tryToDecodeLogOrNoop(log);
         return logWithDecodedArgs;
     }
-    private _onLogStateChanged<ArgsType extends ContractEventArgs>(isRemoved: boolean, rawLog: RawLogEntry): void {
-        const log: LogEntry = marshaller.unmarshalLog(rawLog);
-        _.forEach(this._filters, (filter: FilterObject, filterToken: string) => {
-            if (filterUtils.matchesFilter(log, filter)) {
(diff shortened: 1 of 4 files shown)


packages/base-contract/src/subscription_manager.ts
```diff
@@ -171,8 +171,8 @@ export class SubscriptionManager<ContractEventArgs, ContractEvents extends strin
         if (this._blockAndLogStreamerIfExists === undefined) {
             throw new Error(SubscriptionErrors.SubscriptionNotFound);
         }
-        this._blockAndLogStreamerIfExists.unsubscribeFromOnLogAdded(this._onLogAddedSubscriptionToken as string);
-        this._blockAndLogStreamerIfExists.unsubscribeFromOnLogRemoved(this._onLogRemovedSubscriptionToken as string);
+        this._blockAndLogStreamerIfExists.unsubscribeFromOnLogsAdded(this._onLogAddedSubscriptionToken as string);
```
dekz: Looks like 7.0.0 now calls back with an [Array of logs](https://github.com/ethereumjs/ethereumjs-blockstream/commit/7ac65170c054db83eac51a731388816879911a41) rather than calling back with a single log. This type casting should result in a test failure.

xianny: nice catch @dekz . Looks like `subscribeToOnLogsAdded/Removed` still returns a string, so the subscriptionTokens are still strings. But the callbacks take an array of logs instead of a single log. I think I fixed it in ec2ca78a4a25ebc885bb5c6d9d1a08999bd25ace

Thanks for calling my attention to this, wouldn't have noticed it otherwise :/

Issue #2041, opened 2019-08-07
general - Unable to use 0x libraries: 'cannot read property slice of undefined'


the-jackalope:
Doing almost anything that makes requests to the network with the latest release of various 0x libraries results in the same error. The easiest to demonstrate is the `contract-wrappers` repo - I expect to retrieve a public address's ZRX token balance on the mainnet with the following code:

```
const Web3 = require('web3');
const provider = new Web3.providers.HttpProvider('http://provider-url-here');

const { ContractWrappers } = require('@0x/contract-wrappers');
const  contractWrappers = new ContractWrappers(provider, { networkId: 1 });

contractWrappers.erc20Token.getBalanceAsync('ZRX_ADDRESS_HERE', 'ANY_ETHEREUM_ADDRESS_HERE').then(console.log).catch(console.error)
```

Which consistently yield the following error:
```
> TypeError: Cannot read property 'slice' of undefined
    at Function.BaseContract._throwIfRevertWithReasonCallResult (/Users/johngcook/Developer/fetch/desktop/node_modules/@0x/base-contract/lib/src/index.js:169:27)
    at ERC20TokenContract.<anonymous> (/Users/johngcook/Developer/fetch/desktop/node_modules/@0x/contract-wrappers/node_modules/@0x/abi-gen-wrappers/lib/src/generated-wrappers/erc20_token.js:428:62)
    at step (/Users/johngcook/Developer/fetch/desktop/node_modules/@0x/contract-wrappers/node_modules/@0x/abi-gen-wrappers/lib/src/generated-wrappers/erc20_token.js:56:23)
    at Object.next (/Users/johngcook/Developer/fetch/desktop/node_modules/@0x/contract-wrappers/node_modules/@0x/abi-gen-wrappers/lib/src/generated-wrappers/erc20_token.js:37:53)
    at fulfilled (/Users/johngcook/Developer/fetch/desktop/node_modules/@0x/contract-wrappers/node_modules/@0x/abi-gen-wrappers/lib/src/generated-wrappers/erc20_token.js:28:58)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
```

Whereas I expect to get the address's token balance. I don't believe it to be an issue with the provider as I can instantiate a Web3 instance with the provider and use it to correctly query the address's ETH balance.

@0x/contract-wrappers: `10.1.0`
web3: `1.0.0-beta.55`

DonGambas:
I'm experiencing the same exact issue.

[...]

PR #2037, opened 2019-08-07
switch @0x/contract-wrappers to generated wrappers

xianny:
## Description

Switch to using generated wrappers in `@0x/contract-wrappers`.

This PR reproduces the factory class `ContractWrappers` to help interface with the individual contracts. 

This PR does *not* refactor the current uses of `contract-wrappers` in the monorepo. Most cases are easy to refactor but there are a couple cases where we will need to implement a workaround. Before publishing the new version of `contract-wrappers`, we can't specify new vs old version for existing monorepo dependencies, so it will be easier to update them all at the same time and after we have published the new version. There are 5 packages that currently use `contract-wrappers`:
- contracts/extensions
- 0x.js
- asset-buyer
- asset-swapper
- website
- order-watcher [DEPRECATED]

This PR does **not** get rid of `abi-gen-wrappers`. We will remove it after monorepo use cases have been refactored out. 

This follow-up work is tracked here: https://github.com/0xProject/0x-monorepo/issues/1989

## Testing instructions

I've kept the existing manually written exchange_wrapper_test, which works. 

## Types of changes

* Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

-   [x] Prefix PR title with `[WIP]` if necessary.
-   [x] Add tests to cover changes as needed.
-   [x] Update documentation as needed.
-   [x] Add new entries to the relevant CHANGELOG.jsons.

PR #2096, opened 2019-08-24
Make it easier to use validateOrderFillableOrThrowAsync

xianny:
## Description

The handwritten `ExchangeWrapper` in `@0x/contract-wrappers < 11.0.0` previously exposed a method `validateOrderFillableOrThrowAsync`. This method simulated filling an order with or without a taker, and covered [this specific edge case where an order trades ZRX *and* has ZRX fees](https://github.com/0xProject/0x-monorepo/pull/2096#discussion_r317339948).

The method `validateOrderFillableOrThrowAsync` is still exposed in `OrderValidationUtils` but without the boilerplate that was previously implemented in ExchangeWrapper. 

This PR implements a new method `simpleValidateOrderFillableOrThrowAsync` in `OrderValidationUtils` that reinstates the interface of the ExchangeWrapper method and performs the same simulation. 

Context: This is newly urgent because the [signatures fix](https://github.com/0xProject/0x-monorepo/pull/2078) is forcing devs to upgrade sooner rather than later. 

> I have one additional question re: the upgrade guide. All of the order validation functions got moved to the order-utils package. It would be helpful to see an example of how to use the order validation utils in the order-utils package in particular how to set up the exchangeTradeEmulator.
https://github.com/0xProject/0x-monorepo/issues/2041#issuecomment-521750217

image (1)

## Types of changes

## Checklist:


packages/order-utils/src/asset_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,101 @@
+import { ERC20ProxyContract, ERC20TokenContract, ERC721ProxyContract, ERC721TokenContract } from '@0x/abi-gen-wrappers';
```
xianny: This file was retrieved from https://github.com/0xProject/0x-monorepo/pull/2037/files#diff-7db46b4c4a4907f01f44e8f0018b4219

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -171,6 +179,81 @@ export class OrderValidationUtils {
...
+            // Default behaviour is to validate the amount left on the order.
+            const filledTakerTokenAmount = await exchange.filled.callAsync(orderHashUtils.getOrderHashHex(signedOrder));
+            fillableTakerAssetAmount = signedOrder.takerAssetAmount.minus(filledTakerTokenAmount);
+        } else {
+            const orderStateUtils = new OrderStateUtils(balanceAllowanceStore, this._orderFilledCancelledFetcher);
+            // Calculate the taker amount fillable given the maker balance and allowance
+            const orderRelevantState = await orderStateUtils.getOpenOrderRelevantStateAsync(signedOrder);
+            fillableTakerAssetAmount = orderRelevantState.remainingFillableTakerAssetAmount;
+        }
+
```
xianny: above boilerplate retrieved from https://github.com/0xProject/0x-monorepo/pull/2037/files#diff-540ab868df58adcdec38223c40efc2e1L1160-L1188

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -225,6 +308,44 @@ export class OrderValidationUtils {
...
+    ): Promise<void> {
+        const balanceAllowanceFetcher = new AssetBalanceAndProxyAllowanceFetcher(
+            new ERC20ProxyContract(erc20ProxyAddress, provider),
+            new ERC721ProxyContract(erc721ProxyAddress, provider),
+            provider,
+            BlockParamLiteral.Latest,
+        );
+        const balanceAllowanceStore = new BalanceAndProxyAllowanceLazyStore(balanceAllowanceFetcher);
+        const exchangeTradeSimulator = new ExchangeTransferSimulator(balanceAllowanceStore);
+
```
xianny: above boilerplate retrieved from https://github.com/0xProject/0x-monorepo/pull/2037/files#diff-540ab868df58adcdec38223c40efc2e1L1236-L1243

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -171,6 +179,81 @@ export class OrderValidationUtils {
...
+     * @param zrxTokenAddress A valid address where ZRXTokenContract is deployed
+     * @param networkId The network for all contracts
+     * @param provider The same provider used to interact with contracts
+     * @param signedOrder SignedOrder of interest
+     * @param opts ValidateOrderFillableOpts options (e.g expectedFillTakerTokenAmount.
+     * If it isn't supplied, we check if the order is fillable for the remaining amount.
+     * To check if the order is fillable for a non-zero amount, set `validateRemainingOrderAmountIsFillable` to false.)
+     */
+    public async simpleValidateOrderFillableOrThrowAsync(
+        erc20ProxyAddress: string,
```
dekz: We should be able to detect these either via an Exchange lookup or from the Exchange address -> contractAddresses -> erc20ProxyAddress.  The exchangeAddress is also on the `signedOrder` so is redundant.

Thoughts on reducing this to just `provider, signedOrder, opts` and looking up internally?

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -225,6 +308,44 @@ export class OrderValidationUtils {
...
+     * calls validateFillOrderThrowIfInvalidAsync under the hood
+     * @param erc20ProxyAddress A valid address where an ERC20ProxyContract is deployed
+     * @param erc721ProxyAddress A valid address where an ERC721ProxyContract is deployed
+     * @param zrxTokenAddress A valid address where ZRXTokenContract is deployed
+     * @param provider The same provider used to interact with contracts
+     * @param signedOrder SignedOrder of interest
+     * @param fillTakerAssetAmount Amount we'd like to fill the order for
+     * @param takerAddress The taker of the order
+     */
+    public async simpleValidateFillOrderThrowIfInvalidAsync(
```
dekz: Wasn't this being replaced with the `callAsync` to `fillOrder`? Don't think we need this.

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -171,6 +179,81 @@ export class OrderValidationUtils {
...
+        assert.doesConformToSchema('opts', opts, validateOrderFillableOptsSchema);
+
+        const exchange = new ExchangeContract(exchangeAddress, provider);
+        const balanceAllowanceFetcher = new AssetBalanceAndProxyAllowanceFetcher(
+            new ERC20ProxyContract(erc20ProxyAddress, provider),
+            new ERC721ProxyContract(erc721ProxyAddress, provider),
+            provider,
+            BlockParamLiteral.Latest,
+        );
+        const balanceAllowanceStore = new BalanceAndProxyAllowanceLazyStore(balanceAllowanceFetcher);
```
dekz: `ExchangeTransferSimulator` is useful for basically one reason, If the user is trading XYZ for ZRX and the order has fees, then the `ExchangeTransferSimulator` can handle this case (as it simulates the transfer of ZRX to maker prior to fee payment). This is an annoying case to handle for and is often overlooked, resulting in invalid orders for completely valid orders if only Balances and Allowances are validated. 

I do not believe DevUtils currently handles this edge case.

Extrapolating to V3. I'm selling a Kitty for 100 WETH, there is a 1 WETH fee. I have 0 WETH. This order is valid as there will be funds for fees during settlement but not prior to settlement.

packages/order-utils/src/asset_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,101 @@
...
+        provider: SupportedProvider,
+        stateLayer: BlockParamLiteral,
+    ) {
+        this._erc20Proxy = erc20Proxy;
+        this._erc721Proxy = erc721Proxy;
+        this._stateLayer = stateLayer;
+        this._provider = provider;
+    }
+    public async getBalanceAsync(assetData: string, userAddress: string): Promise<BigNumber> {
+        const decodedAssetData = assetDataUtils.decodeAssetDataOrThrow(assetData);
```
dekz: We should be able to refactor this to use DevUtils.

```
devUtils.getBalance(owner, assetData)
```

Which handles ERC20, ERC721, ERC1155 and MAP.

packages/order-utils/src/asset_balance_and_proxy_allowance_fetcher.ts
```diff
@@ -0,0 +1,101 @@
...
+                    userAddress,
+                )).dividedToIntegerBy(nestedAmountElement);
+                if (balance === undefined || nestedAssetBalance.isLessThan(balance)) {
+                    balance = nestedAssetBalance;
+                }
+            }
+        }
+        return balance as BigNumber;
+    }
+    public async getProxyAllowanceAsync(assetData: string, userAddress: string): Promise<BigNumber> {
```
dekz: We should be able to refactor this to use DevUtils.

```
devUtils.getAssetProxyAllowance.callAsync(owner, assetData)
```

Which handles ERC20, ERC721, ERC1155 and MAP.

packages/order-utils/src/order_validation_utils.ts
xianny: Removed

packages/order-utils/src/order_validation_utils.ts
```diff
@@ -225,86 +300,4 @@ export class OrderValidationUtils {
             zrxAssetData,
         );
     }
-    /**
-     * Validate a call to FillOrder and throw if it wouldn't succeed
```
xianny: We can now use `exchange.fillOrder.callAsync` instead.

packages/migrations/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "4.3.2",
+        "changes": [
+            {
+                "note": "Removed dependency on @0x/order-utils",
```
xianny: I had to introduce a dependency on `@0x/migrations` to `@0x/order-utils` for tests (which were retrieved from the old exchange wrapper tests). I've run into similar issues with tests introducing circular dependencies for other packages. Would like to discuss the possibility of developing an integration test suite for any tests that run with ganache and the standard 0x migration. This would also potentially speed up our testing time, by doing one migration + snapshots instead of separate migrations in numerous unit test suites. 

Since we intend to remove these methods in a couple months, this seems like the easiest solution for now to unblock this task.

packages/order-utils/test/exchange_transfer_simulator_test.ts
```diff
@@ -42,6 +42,7 @@ describe('ExchangeTransferSimulator', async () => {
             from: devConstants.TESTRPC_FIRST_ADDRESS,
         };
 
+        await blockchainLifecycle.startAsync();
```
xianny: This was affecting the new unit tests in `order_validation_utils_test.ts` that are also blockchain dependent.

albrow:
@fabioberger should probably review also since he has more context for some of these changes. I was mostly focusing on the `contract-addresses` package and related changes.

packages/contract-addresses/src/index.ts
```diff
@@ -131,3 +131,21 @@ export function getContractAddressesForNetworkOrThrow(networkId: NetworkId): Con
...
+ * Uses a given exchange address to look up the network id that the exchange contract is deployed
+ * on. Only works for Ethereum mainnet or a supported testnet. Throws if the exchange address
+ * does not correspond to a known deployed exchange contract.
+ * @param exchangeAddress The exchange address of concern
+ * @returns The network ID on which the exchange contract is deployed
+ */
+export function getNetworkIdByExchangeAddressOrThrow(exchangeAddress: string): NetworkId {
+    for (const networkId of Object.keys(networkToAddresses)) {
+        if (networkToAddresses[networkId as any].exchange === exchangeAddress) {
+            return networkId as any;
```
fabioberger: Why do we need to cast to `any` here? Why not cast to `NetworkId`?

packages/migrations/src/migration.ts
```diff
@@ -1,15 +1,43 @@
...
 import { erc20TokenInfo, erc721TokenInfo } from './utils/token_info';
 
+// HACK (xianny): Copied from @0x/order-utils to get rid of circular dependency
+/**
+ * Encodes an ERC20 token address into a hex encoded assetData string, usable in the makerAssetData or
+ * takerAssetData fields in a 0x order.
+ * @param tokenAddress  The ERC20 token address to encode
+ * @return The hex encoded assetData string
+ */
+function encodeERC20AssetData(tokenAddress: string): string {
```
fabioberger: @xianny we have `DevUtils` contract now in `abi-gen-wrappers` that has this exact method. Why not just use that?

packages/order-utils/src/order_validation_utils.ts
fabioberger: @dekz aren't fees paid in ETH _not_ WETH in V3?

packages/migrations/src/migration.ts
xianny: https://github.com/0xProject/0x-monorepo/pull/2108 isn't merged yet. Without that, we still need to provide the address to a deployed DevUtils Contract, which doesn't exist before migrations run. We need to encode zrx asset data to call `DevUtilsContract.deployFromZeroExArtifact`. 

I can change it after merging #2108

packages/contract-addresses/src/index.ts
xianny: `Object.keys` casts everything to string. I got a `string has no overlap with NetworkId` error. For clarity we can cast `networkId as any as NetworkId` instead

fabioberger: I'd prefer that.

packages/migrations/src/migration.ts
fabioberger: ok!

PR #2097, opened 2019-08-24
Exchange Protocol Fees

jalextowle:
## Description

image

This PR adds protocol fees to the `Exchange` contract. Several things had to happen for this to occur.

#### Refundable
A new contract was added to `@0x:contracts-utils` called `Refundable`. `Refundable` is a contract with no dependencies that introduces two new modifiers and a state variable called `shouldNotRefund`. Both modifiers are designed to respect the `refundFinalBalance` state variable in that they will only refund if `shouldNotRefund` is `false`.
- `disableRefundUntilEnd`: This modifier will set `shouldNotRefund` to true, call the function that it modifies, and then refund all of the ether in the contract to the sender. This will disable all nested functions that use `disableRefundUntilEnd` and `refundFinalBalance`, and this modifier will itself be disabled if `shouldNotRefund` is true.
- `refundFinalBalance`: This modifier will call the function that it is modifying and then refund all of the ether in the contract to the sender. This will be disabled if `shouldNotRefund` is true.

#### Payable Functions
Almost every function in the exchange had to be made `payable`. Some notable exceptions include
- `cancelOrder`
- `batchCancelOrders`
- `cancelOrdersUpTo`

Since these functions are now payable, they all were also modified by `refundFinalBalance` or `disableRefundUntilEnd`. The functions that were given `refundFinalBalance` have no risk of calling a function that also uses `refundFinalBalance` and experiencing strange behavior. Similarly, functions whose behavior would be affected by refunds in nested calls are modified by `disableRefundUntilEnd`.

#### Protocol Fees
Protocol fees were added to the internal functions `_fillOrder` and `_matchOrders`, so every function that calls these functions or has the potential to call them will potentially pay protocol fees. `fillOrder` protocol fees are half of the price of `matchOrders` protocol fees, and these fees will be paid on a per-fill basis. For example, a call to `batchFillOrders` with `k` orders that are filled will pay `k * protocolFeePerFill`, where `protocolFeePerFill = exchange.protocolFeeMultiplier * tx.gasprice`.

New state that was added to the exchange (like `protocolFeeMultiplier` from the above example), including:
- `protocolFeeMultiplier`: This is the number that will be multiplied to `tx.gasprice` to calculate the protocol fee of a single fill.
- `protocolFeeCollector`: This is the contract that will receive `protocolFee` payments.

Both of these state variables are protected by the `onlyOwner` modifier, and will thus not be mutable by normal addresses.

#### Glorious Solidity Tests
One of my favorite parts of writing this PR was discovering that writing tests in Solidity is actually surprisingly nice. The tests that I wrote are in the form:

```
function testSomething(SomethingContract testContract, ...)
    external
{
    testContract.something(...); // Call something with args

    verifySomething(...); // Make assertions about the call to something
}
```

The assertions that are made are generally made by leveraging contract inheritance's ability to override important internal functions. These overridden functions are made to change state in ways that indicate internal behavior.

Both `Protocol Fees` and `Refundable` received good Solidity test treatment that runs quickly and makes strong assertions about their behavior.

## Testing instructions

`yarn build:contracts && yarn test:contracts && yarn lint:contracts` (or `ycall` for those with my `.bash_profile` sourced :))

## Types of changes

## Checklist:


contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -448,5 +562,27 @@ contract MixinExchangeCore is
...
+                _dispatchTransferFrom(
+                    orderHash,
+                    WETH_ASSET_DATA,
+                    takerAddress,
+                    staking,
+                    protocolFee
+                );
+
+                // Attribute the protocol fee to the maker
+                IStaking(staking).recordProtocolFee(order.makerAddress, protocolFee);
```
abandeali1: Thinking about it more, I kind of like having the `weth.transferFrom` call happen within `staking.recordProtocolFee` rather than the `Exchange` contract. This change drastically reduces the amount of trust that the `Staking` contract must have in the `Exchange` contract. It only slightly increases the amount of trust the `ERC20Proxy` has in the `Staking` contract, since it already needs to be authorized to transfer ZRX.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -482,6 +490,41 @@ contract MixinMatchOrders is
             matchedFillResults.profitInRightMakerAsset
         );
 
+        // Pay protocol fees
+        if (staking != address(0)) {
```
abandeali1: Set `staking` to a stack variable.

contracts/exchange/contracts/src/MixinMatchOrders.sol
```diff
@@ -482,6 +490,41 @@ contract MixinMatchOrders is
...
+                _dispatchTransferFrom(
+                    leftOrderHash,
+                    WETH_ASSET_DATA,
+                    takerAddress,
+                    staking,
+                    2 * protocolFee
+                );
+
+                // Attribute the protocol fees to the maker addresses
+                IStaking(staking).batchRecordProtocolFees(makers, fees);
```
abandeali1: Same as above - let's put the WETH `transferFrom` call in `batchRecordProtocolFees`.

abandeali1: I also wonder if we should just remove these batch methods for the sake of simplicity. We're only really saving ~700 gas here, and this won't be used anywhere else in the `Exchange` contract.

contracts/exchange/contracts/src/MixinStakingManager.sol
```diff
@@ -0,0 +1,67 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/Ownable.sol";
+import "./interfaces/IStakingManager.sol";
+
+
+contract MixinStakingManager is
```
abandeali1: Thoughts on making the language in this contract more generic (e.g `MixinProtocolFeeManager`? The logic in the `Exchange` contract is pretty much entirely independent of the `Staking` contract.

contracts/exchange/contracts/src/MixinStakingManager.sol
```diff
@@ -0,0 +1,67 @@
...
+    IStakingManager,
+    Ownable
+{
+    // The protocol fee multiplier -- the owner can update this field.
+    uint256 public protocolFeeMultiplier;
+
+    // The address of the registered staking contract -- the owner can update this field.
+    address public staking;
```
abandeali1: If we make this more generic, this would probably be `protocolFeeReceiver` or `protocolFeeCollector`.

contracts/exchange/contracts/src/MixinStakingManager.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    // The address of the wrapped ether contract -- the owner can update this field.
+    address public weth;
```
abandeali1: We can remove this if the WETH logic is moved into `recordProtocolFee`.

contracts/exchange/contracts/src/MixinStakingManager.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    /// @dev Allows the owner to update the protocol fee multiplier.
+    /// @param updatedProtocolFeeMultiplier The updated protocol fee multiplier.
+    function updateProtocolFeeMultiplier(uint256 updatedProtocolFeeMultiplier)
+        external
+        onlyOwner()
```
abandeali1: Nit: we don't actually invoke modifiers anywhere else (just `onlyOwner` here).

contracts/utils/contracts/src/Refundable.sol
```diff
@@ -19,6 +19,18 @@
...
+contract Refundable {
+
+    // This bool is used by the refund modifier to allow for lazily evaluated refunds.
+    bool internal shouldNotRefund;
+
+    modifier refund {
+        if (shouldNotRefund) {
+            _;
+        } else {
+            shouldNotRefund = true;
```
abandeali1: It looks like calling this modifier twice will actually disable refunds if I'm reading this correctly? How would they be turned back on?

We can also implement this without requiring an SSTORE for the default case. All this really needs to do is:

```
_;
if (!shouldNotRefund) {
    msg.sender.transfer(address(this).balance);
}
```

This only requires adding the cost of an SLOAD to every single function that uses the modifier, with the exception of the batch `noThrow` functions (only 3 functions I believe). For those we will have to set `shouldNotRefund` to `true` at the beginning of the function call and reset it to `false` at the end of the function call. That can actually probably be its own modifier in this contract (maybe call it `disableRefundsUntilEnd` or something similar).

contracts/utils/contracts/src/Refundable.sol
```diff
@@ -19,6 +19,18 @@
...
 
 
-interface IStaking {
-
+contract Refundable {
+
+    // This bool is used by the refund modifier to allow for lazily evaluated refunds.
+    bool internal shouldNotRefund;
+
+    modifier refund {
```
abandeali1: Can we rename this to `refundFinalBalance`? It makes the intent of this modifier much more clear IMO.

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -271,21 +280,126 @@ contract MixinExchangeCore is
...
+        // to the fact that 4 of the fields are of type `bytes`, there is added complexity in that
+        // these fields in the log data are actually byte offsets to the length-prefaced `bytes`
+        // of the field. This means that we also need to account for 4 length slots -- 4 * 32 = 128
+        // more bytes in the log data -- and for the length of the bytes themselves -- total_length =
+        // order.makerAssetData.length + order.takerAssetData.length + order.makerFeeAssetData.length
+        // + order.takerFeeAssetData.length.
+        //
+        // Altogether, the log data will be 384 + 128 + total_length = 512 + total_length bytes long,
+        // which explains the memory allocation of the bytes below:
+        bytes memory logData = new bytes(
```
abandeali1: Do we have a sense of the gas differences using this approach vs logging 2 separate events?

contracts/exchange-libs/contracts/src/LibFillResults.sol
```diff
@@ -78,6 +84,8 @@ library LibFillResults {
             order.takerFee
         );
 
+        fillResults.protocolFeePaid = gasPrice.safeMul(protocolFeeMultiplier);
```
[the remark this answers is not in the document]
jalextowle: As discussed offline, I'll be using the original solution.

contracts/exchange/test/internal.ts
```diff
@@ -195,20 +202,32 @@ blockchainTests('Exchange core internal functions', env => {
...
+            // expect(fillEvent.args.protocolFeePaid).to.bignumber.eq(fillResults.protocolFeePaid);
+            // expect(fillEvent.args.isProtocolFeePaidInEth).to.eq(isProtocolFeePaidInEth);
+            expect(fillEvent.args.makerAssetData).to.eq(order.makerAssetData);
+            expect(fillEvent.args.takerAssetData).to.eq(order.takerAssetData);
+            expect(fillEvent.args.makerFeeAssetData).to.eq(order.makerFeeAssetData);
+            expect(fillEvent.args.takerFeeAssetData).to.eq(order.takerFeeAssetData);
         }
 
-        it('emits a `Fill` event and updates `filled` state correctly', async () => {
+        it('emits a `Fill` event and updates `filled` state correctly if protocol fee is paid in ETH', async () => {
```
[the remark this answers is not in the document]
jalextowle: Now we are 👍

contracts/utils/contracts/test/TestRefundableReceiver.sol
```diff
@@ -0,0 +1,191 @@
...
+            // Ensure that the other contract's balance is equal to zero.
+            require(address(testRefundable).balance == 0, "Incorrect balance for TestRefundable");
+        }
+
+        // Ensure that `shouldNotRefund` in TestRefundable is set to the parameter `shouldNotRefund`
+        // after the call (i.e. the value didn't change during the function call).
+        require(testRefundable.getShouldNotRefund() == shouldNotRefund, "Incorrect shouldNotRefund value");
+
+        // Drain the contract of funds so that subsequent tests don't have to account for leftover ether.
+        msg.sender.transfer(address(this).balance);
```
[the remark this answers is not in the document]
jalextowle: So it does, but I'm making two calls inside a single `it`. I'll have to leave this one in, but I removed my **over-carefullness** in `TestProtocolFeesReceiver`.

contracts/exchange-libs/contracts/src/LibFillResults.sol
```diff
@@ -78,6 +81,9 @@ library LibFillResults {
             order.takerFee
         );
 
+        // Compute the protocol fee for a single fill.
+        fillResults.protocolFeePaid = tx.gasprice.safeMul(protocolFeeMultiplier);
```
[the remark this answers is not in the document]
jalextowle: Nope, I took your suggestion when I changed it back, as a bonus.

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -47,11 +48,24 @@ export class FillOrderWrapper {
         takerAddress: string,
         opts: { takerAssetFillAmount?: BigNumber } = {},
         initBalanceStore: BalanceStore,
+        // stakingOpts: {
```
abandeali1: Uncomment/remove this

contracts/exchange/test/internal.ts
```diff
@@ -169,10 +171,20 @@ blockchainTests('Exchange core internal functions', env => {
...
             const orderHash = randomHash();
-            const fillResults = LibReferenceFunctions.calculateFillResults(order, takerAssetFillAmount);
+            const fillResults = LibReferenceFunctions.calculateFillResults(
+                order,
+                takerAssetFillAmount,
+                protocolFeeMultiplier,
+                gasPrice,
+            );
             const expectedFilledState = orderTakerAssetFilledAmount.plus(takerAssetFillAmount);
+            // const opts = isProtocolFeePaidInEth
```
abandeali1: Uncomment/remove this

contracts/exchange/test/transactions.ts
```diff
@@ -1037,13 +1037,14 @@ blockchainTests.resets('Exchange transactions', env => {
                     const cancelTransaction = await makerTransactionFactory.newSignedTransactionAsync({
                         data: cancelData,
                     });
+
                     await exchangeWrapperContract.cancelOrdersUpTo.awaitTransactionSuccessAsync(
                         targetOrderEpoch,
                         cancelTransaction.salt,
                         cancelTransaction.expirationTimeSeconds,
                         cancelTransaction.signature,
+                        // { from: makerAddress },
```
abandeali1: Uncomment/delete

PR #2104, opened 2019-08-26
Reimplement Makers joining Pools

mzhu25:
## Description
Instead of having makers sign an approval to be added to a staking pool, use an on-chain handshake: 
1. Maker calls `joinStakingPool(poolId)` (reverts if the maker is already in a pool)
2. Pool operator calls `addMakerToStakingPool(poolId, makerAddress)` 
3. If the staking pool is not yet full (`MAX_POOL_SIZE` in MixinDeploymentConstants), the maker gets added to the staking pool

This PR also:
- updates web3Wrapper to initialize 20 test accounts so we can test the behavior when a pool is full
- adds a flag to `createStakingPool` to add the operator as a maker (so they don't have to go through the handshake procedure)

## Checklist:


contracts/staking/test/pools_test.ts
```diff
@@ -47,12 +46,23 @@ blockchainTests('Staking Pool Management', env => {
...
+        it('Should successfully create a pool and add owner as a maker', async () => {
+            // test parameters
+            const operatorAddress = users[0];
+            const operatorShare = 39;
+            const poolOperator = new PoolOperatorActor(operatorAddress, stakingWrapper);
+            // create pool
+            const poolId = await poolOperator.createStakingPoolAsync(operatorShare, true);
+            expect(poolId).to.be.equal(stakingConstants.INITIAL_POOL_ID);
+            // check that the next pool id was incremented
+            const nextPoolId = await stakingWrapper.getNextStakingPoolIdAsync();
```
jalextowle: Let's throw this check into the `PoolOperatorActor` class.

contracts/staking/test/pools_test.ts
```diff
@@ -212,45 +243,69 @@ blockchainTests('Staking Pool Management', env => {
...
             );
-            const tx = stakingWrapper.removeMakerFromStakingPoolAsync(poolId, makerAddress, notOperatorAddress);
+            const tx = stakingWrapper.removeMakerFromStakingPoolAsync(
+                poolId,
+                makerAddress,
+                neitherOperatorNorMakerAddress,
+            );
             await expect(tx).to.revertWith(revertError);
         });
+        it('Should fail to add a maker if the pool is full', async () => {
```
jalextowle: This test case looks great! I'm like the use of `Promise.all` here.

contracts/staking/test/pools_test.ts
```diff
@@ -82,126 +110,129 @@ blockchainTests('Staking Pool Management', env => {
...
             // add makers to pool
-            const makerApprovals = [
-                makers[0].signApprovalForStakingPool(poolId),
-                makers[1].signApprovalForStakingPool(poolId),
-                makers[2].signApprovalForStakingPool(poolId),
-            ];
-            await poolOperator.addMakerToStakingPoolAsync(poolId, makerAddresses[0], makerApprovals[0].signature);
-            await poolOperator.addMakerToStakingPoolAsync(poolId, makerAddresses[1], makerApprovals[1].signature);
-            await poolOperator.addMakerToStakingPoolAsync(poolId, makerAddresses[2], makerApprovals[2].signature);
+            await Promise.all([
```
jalextowle: I wonder if it would be useful to introduce a `MakerMasterActor` that takes has a function called "assignMakerActions" that takes a method name, and then calls this method with the `poolId` and encapsulates some of this logic. Thoughts?

I'm starting to see the `Promise.all` used a fair amount, so it might be good to consider a higher level abstraction.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -122,36 +121,97 @@ contract MixinStakingPool is
...
+            LibRichErrors.rrevert(LibStakingRichErrors.PoolIsFullError(poolId));
+        }
+
+        // Add maker to pool
+        IStructs.MakerPoolJoinStatus memory poolJoinStatus = IStructs.MakerPoolJoinStatus({
+            poolId: poolId,
+            confirmed: true
+        });
+        poolJoinedByMakerAddress[makerAddress] = poolJoinStatus;
         makerAddressesByPoolId[poolId].push(makerAddress);
```
dorothy-zbornak: What do we think about just getting rid of this state variable? `poolJoinedByMakerAddress` already tells us if a maker is part of a pool. I guess being able to query the addresses is helpful for the assertion wrappers, but is that it? You can always derive the list by examining event logs...

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -295,49 +200,62 @@ library LibStakingRichErrors {
...
+            makerAddress
         );
     }
 
-    function MakerAddressAlreadyRegisteredError(
-        address makerAddress
+    function MakerAddressNotRegisteredError(
+        address makerAddress,
+        bytes32 makerPoolId,
+        bytes32 poolId
```
dorothy-zbornak: Do we need this field? The caller already passes in `poolId`, so it's already known.
Maybe we can even consolidate a bunch of these maker errors with an error code pattern if they share similar fields.

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -295,49 +200,62 @@ library LibStakingRichErrors {
...
+            poolId
         );
     }
 
-    function MakerAddressNotRegisteredError(
+    function MakerNotPendingJoinError(
         address makerAddress,
-        bytes32 makerPoolId,
+        bytes32 pendingJoinPoolId,
         bytes32 poolId
```
dorothy-zbornak: Same.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -122,36 +121,97 @@ contract MixinStakingPool is
...
+    function joinStakingPoolAsMaker(
+        bytes32 poolId
+    )
+        external
+    {
+        // Is the maker already in a pool?
+        address makerAddress = msg.sender;
+        if (isMakerAssignedToStakingPool(makerAddress)) {
+            LibRichErrors.rrevert(LibStakingRichErrors.MakerAddressAlreadyRegisteredError(
+                makerAddress
```
dorothy-zbornak: Might be nice to also include the `poolId` it's assigned to.

contracts/staking/test/pools_test.ts
dorothy-zbornak: ya this code is strangely aesthetically pleasing :+1:

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
mzhu25: Hm that's true... We use the length of the array to check the size of the pool, but we can do that with just a counter. @hysz thoughts?

contracts/staking/test/pools_test.ts
mzhu25: Idk, a `MakerMasterActor` doesn't really map to a real-world entity though. But we could introduce `batchAddMakerToStakingPoolAsync` and `batchRemoveMakerFromStakingPoolAsync` to `PoolOperatorActor` though

contracts/staking/test/pools_test.ts
mzhu25: o wait actually the reason this is here is because we're checking the poolId against a string constant, so to do the check in `createStakingPoolAsync` we'd need to increment the pool id in TypeScript, which is probably not worth

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
hysz: Yeah let's do it. Since the list is capped - if we did ever need this data on-chain - we could parameterize it and validate the makers in the same number of `sload` operations.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -235,43 +245,22 @@ contract MixinStakingPool is
...
+            LibRichErrors.rrevert(LibStakingRichErrors.MakerPoolAssignmentError(
+                LibStakingRichErrors.MakerPoolAssignmentErrorCodes.MAKER_ADDRESS_NOT_REGISTERED,
                 makerAddress,
-                makerPoolId,
-                poolId
+                makerPoolId
             ));
         }
 
-        // load list of makers for the input pool.
```
dorothy-zbornak: :tada:

PR #2108, opened 2019-08-28
Use local EVM for pure functions

### LINKED ISSUE #1872: Use EVM to replace eth_call for pure functions
## Expected Behavior

Generated contract wrappers should reroute calls to pure Solidity functions to a native EVM implementation.

## Current Behavior

Generated wrappers make an `eth_call` for every Solidity function called

## Possible Solution

In `abi-gen-templates`, check if `stateMutability === pure` for each function and then reroute to an EVM if so.

xianny:
## Description

Currently, `@0x/contract-wrappers` provides a convenience method `callAsync` which does an `eth_call` for any `pure` or `view` functions. The `pure` functions don't require any blockchain state and can be computed locally. This PR uses `ethereumjs-vm` to instantiate a local EVM, and uses it for `pure` functions instead of performing the `eth_call`. This is an implementation detail and is hidden from end users. The method signature for `callAsync` is the same in either case.

Affected methods:
- 9 methods in DevUtils related to encoding and decoding
- all `getProxyId` methods (5) 

Resolves #1872
## Testing instructions

Pure methods in `abi_gen_dummy.ts` have been updated as part of this change. All unit tests should still pass. 

## Types of changes

## Checklist:


package.json
```diff
@@ -61,11 +61,11 @@
         "files": [
             {
                 "path": "packages/0x.js/_bundles/index.min.js",
-                "maxSize": "800kB"
+                "maxSize": "1200kB"
```
xianny: ouch, @fabioberger what do you think?

fabioberger: Not ideal but worth it.

packages/0x.js/webpack.config.js
```diff
@@ -39,6 +39,9 @@ module.exports = {
             }),
         ],
     },
+    node: {
+        fs: 'empty',
+    },
```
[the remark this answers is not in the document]
fabioberger: Hm, will this bite anyone else in the butt?

packages/base-contract/src/globals.d.ts
```diff
@@ -4,3 +4,5 @@ declare module '*.json' {
     export default json;
     /* tslint:enable */
 }
+
+declare module 'ethereumjs-vm';
```
fabioberger: Can we move this to `typescript-typings` package and write the types for the methods we use of this dep?

packages/base-contract/src/index.ts
```diff
@@ -61,6 +68,8 @@ export class BaseContract {
     public address: string;
     public contractName: string;
     public constructorArgs: any[] = [];
+    private _evmIfExists?: any; // hack (xianny): VM is also used as a namespace
```
fabioberger: Can you elaborate on the need to use `any` here?

packages/instant/webpack.config.js
```diff
@@ -135,6 +135,9 @@ const generateConfig = (dischargeTarget, heapConfigOptions, rollbarConfigOptions
         resolve: {
             extensions: ['.js', '.json', '.ts', '.tsx'],
         },
+        node: {
+            fs: 'empty',
+        },
```
fabioberger: @fragosti @BMillman19 see any issues with doing this?

packages/base-contract/src/index.ts
```diff
@@ -150,6 +159,38 @@ export class BaseContract {
         }
         return rawEncoded;
     }
+    public async evmExecAsync(input: Buffer): Promise<string> {
+        // should only run once, the first time it is called
+        if (this._evmIfExists === undefined) {
+            const vm = new VM();
```
feuGeneA: i wonder if it would be worthwhile to move the one-time code into a method, which could optionally be called by the client in order to quicken the first actual contract call.

package.json
```diff
@@ -61,11 +61,11 @@
...
             },
             {
                 "path": "packages/instant/umd/instant.js",
-                "maxSize": "1000kB"
+                "maxSize": "1300kB"
```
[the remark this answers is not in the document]
xianny: discussed on Slack and decided to go ahead

package.json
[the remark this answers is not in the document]
xianny: The actual bundle size is around 1.1mb. `ethereumjs-vm` seems to add ~270kb.

packages/base-contract/src/index.ts
xianny: I think it'll be alright this way - any functions using the local EVM won't be time sensitive.

package.json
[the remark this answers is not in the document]
xianny: discussed and agreed not ideal but worth it.

contracts/utils/test/lib_address_array.ts
```diff
@@ -1,11 +1,4 @@
-import {
-    addressUtils,
-    chaiSetup,
-    expectContractCallFailedAsync,
-    provider,
-    txDefaults,
-    web3Wrapper,
-} from '@0x/contracts-test-utils';
+import { addressUtils, chaiSetup, provider, txDefaults, web3Wrapper } from '@0x/contracts-test-utils';
```
jalextowle: You might consider using Lawrence's new `blockchainTests` system. To make this change, you would remove any code relating to chai, import `expect` from `contracts-test-utils`, and then change:

```
describe('...', () => {
    before(async () => {
        blockchain_lifecycle.startAsync();
    });
    after(async () => {
       blockchain_lifecycle.revertAsync();
    });
});
```

to

```
blockchainTests.resets('...', env => // This env variable will have provider, txDefaults, and web3Wrapper
});
```

Essentially, this change makes things a little more succinct, matches the current style of tests used by the protocol team, and allows the `BlockchainLifecycle` dependency to be removed.

contracts/utils/test/lib_bytes.ts
```diff
@@ -1,12 +1,4 @@
-import {
-    chaiSetup,
-    constants,
-    expectContractCallFailedAsync,
-    provider,
-    txDefaults,
-    typeEncodingUtils,
-    web3Wrapper,
-} from '@0x/contracts-test-utils';
+import { chaiSetup, constants, provider, txDefaults, typeEncodingUtils, web3Wrapper } from '@0x/contracts-test-utils';
```
jalextowle: Same as above (it's optional, but very nice).

packages/base-contract/src/index.ts
```diff
@@ -150,6 +159,38 @@ export class BaseContract {
         }
         return rawEncoded;
     }
+    public async evmExecAsync(input: Buffer): Promise<string> {
```
jalextowle: This looks great! One thought that I have is that it might occasionally be useful to specify the bytecode that is being used. For example, some of our tests will test that the `codesize` of the contract is within the limits of what the evm will allow (`codesize < 24576`). In such a case, we could memoize the bytecode that is being used in the contract and avoid using `getContractCodeAsync`.

packages/abi-gen/templates/TypeScript/partials/method_call_pure.handlebars
```diff
@@ -0,0 +1,37 @@
...
+        schemas.jsNumber,
+    ]);
+    if (defaultBlock !== undefined) {
+        assert.isBlockParam('defaultBlock', defaultBlock);
+    }
+    const self = this as any as {{contractName}}Contract;
+    const encodedData = self._strictEncodeArguments('{{this.functionSignature}}', [{{> normalized_params inputs=inputs}}]);
+    const encodedDataBytes = Buffer.from(encodedData.substr(2), 'hex');
+    
+    const rawCallResult = await self.evmExecAsync(encodedDataBytes);
```
jalextowle: 🎉

packages/instant/webpack.config.js
xianny: chatted on Slack, seems like it will be OK. 

for context,  this tells webpack to stub `fs` with an empty module.

packages/base-contract/src/index.ts
xianny: The type should be `VM`, but `ethereumjs-vm` defines it as a namespace and the ts compiler doesn't allow it to be used as a type. I tried importing it all different ways and couldn't figure it out :/

packages/base-contract/src/index.ts
xianny: Yeah, it would be ideal if we could pass bytecode at instantiation. I tried this earlier, but found that it created too many disruptive changes. Also, the deployed bytecode and deploy*able* bytecode is different; we currently store only deploy**able** bytecode in `@0x/contract-artifacts`, and we need deploy**ed** bytecode for running calls on contract functions, so we would have to store both. @dekz  and @fabioberger felt strongly that would bloat the `@0x/contract-artifacts` package too much, so we opted for always calling `getContractCodeAsync` instead.

packages/base-contract/src/index.ts
xianny: Aha! there was a leftover module declaration in `base-contract`. removing it fixed it.

xianny:
Ready for re-review. I rebased off `development` and force-pushed the rebase. New changes are https://github.com/0xProject/0x-monorepo/pull/2108/commits/62d7cd559c89f717bb7a0e8b144ce4560d2aeee3 and https://github.com/0xProject/0x-monorepo/pull/2108/commits/5ead509172e4438b5a860f039ee475403205dea5

contracts/utils/test/lib_address_array.ts
xianny: discussed on Slack. this is only in the `3.0` branch right now.

packages/0x.js/webpack.config.js
fabioberger: I'm still worried about this hack. Does this mean anyone using `0x.js` in a webpack project will need to do this too to get it to work? If so, please add browser-specific instructions to the README explaining the need for this step.

xianny: I created a test project importing 0x.js and tried bundling with webpack. Sadly yes and there doesn't seem to be any workaround  as long as we are using `ethereumjs-vm` as a dependency. I'll add instructions to the README.

PR #2112, opened 2019-08-29
Delete redundant functions

abandeali1:
## Description

- _Most_ of this PR just deletes redundant functions. However, there are a few subtle changes in functionality (explained below)
- There were previously many public getter functions in the Staking contract that called other public getter functions in the Vault contracts. These functions were removed so that you now need to call the Vaults directly if you want to query this information.
- The Vault contracts now receive the `stakingProxyAddress` in their constructors. This address was previously upgradable by the `owner`. However, the StakingProxy is very unlikely to be replaced in the near term, since just the Staking logic contract can be upgraded in most cases. This change removes some flexibility but reduces the attack surface and trust requirements by a lot IMO. 

## Checklist:


contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -443,4 +447,19 @@ library LibStakingRichErrors {
             poolId
         );
     }
+
+    function ZrxWithdrawalFailedError(
```
jalextowle: I wonder if it would make sense to include an enum that gives info about why a withdrawal failed in this error. For example, if the `WETH` can't be unwrapped for some reason, it might be useful to include that in this error message.

On the other hand, if there aren't several cases in which withdrawing can fail that aren't like this scenario (which is pretty unlikely), then it's fine to skip this.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -232,7 +220,8 @@ contract MixinExchangeFees is
             ));
         }
         if (totalRewardsPaid > 0) {
-            _depositIntoStakingPoolRewardVault(totalRewardsPaid);
+            address payable rewardVaultAddress = address(uint160(address(rewardVault)));
```
jalextowle: Are you doing this to cast `address(rewardVault)` to `address payable`?

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewardVault.sol
```diff
@@ -52,85 +49,4 @@ contract MixinStakingPoolRewardVault is
     {
         return address(rewardVault);
     }
-
```
jalextowle: Big RIP!

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
abandeali1: It's just a `ZRX.transfer` call, so it can really only fail if the balance is insufficient. It actually _should_ never fail, but it felt dirty not checking the return value.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
abandeali1: Yep, seems this is the only way =/

hysz: Yeah I couldn't find a better way to do it, either.

PR #2113, opened 2019-08-29
Move pool operator logic to vault and allow operators to decrease operatorShare

mzhu25:
## Description
Moves the pool operator getter and related modifiers from `MixinStakingPool` to `MixinStakingPoolRewardVault`, moving the state from `MixinStorage` to `StakingPoolRewardVault`. 
This PR also adds the functionality for operators to decrease their share of the rewards in the pool. 
As a byproduct of these changes, `IStructs.Pool` is ded but the `Balance` struct is renamed to `Pool` because it now has a few fields unrelated to balances.

## Testing instructions
Probably want to add some integration tests involving `decreaseStakingPoolOperatorShare` in a future PR, for now I think unit tests will suffice

## Types of changes

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/interfaces/IStructs.sol
+++ b/contracts/staking/contracts/src/interfaces/IStructs.sol
@@ -21,24 +21,6 @@ pragma solidity ^0.5.9;
 
 interface IStructs {
 
-    /// @dev Allowed signature types.
-    enum SignatureType {
-        Illegal,            // 0x00, default value
-        Invalid,            // 0x01
-        EIP712,             // 0x02
-        EthSign,            // 0x03
-        Wallet,             // 0x04
-        NSignatureTypes     // 0x05, number of signature types. Always leave at end.
-    }
-
-    /// @dev Status for Staking Pools (see MixinStakingPool).
-    /// @param operatorAddress Address of pool operator.
-    /// @param operatorShare Portion of pool rewards owned by operator, in ppm.
-    struct Pool {
-        address payable operatorAddress;
-        uint32 operatorShare;
-    }
-
     /// @dev Status for a pool that actively traded during the current epoch.
     /// (see MixinExchangeFees).
     /// @param poolId Unique Id of staking pool.
--- a/contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
+++ b/contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
@@ -84,9 +84,9 @@ library LibStakingRichErrors {
     bytes4 internal constant AMOUNT_EXCEEDS_BALANCE_OF_POOL_ERROR_SELECTOR =
         0x4c5c09dd;
 
-    // bytes4(keccak256("InvalidPoolOperatorShareError(bytes32,uint32)"))
-    bytes4 internal constant INVALID_POOL_OPERATOR_SHARE_ERROR_SELECTOR =
-        0x70f55b5a;
+    // bytes4(keccak256("OperatorShareError(uint8,bytes32,uint32)"))
+    bytes4 internal constant OPERATOR_SHARE_ERROR_SELECTOR =
+        0x22df9597;
 
     // bytes4(keccak256("PoolAlreadyExistsError(bytes32)"))
     bytes4 internal constant POOL_ALREADY_EXISTS_ERROR_SELECTOR =
@@ -112,13 +112,6 @@ library LibStakingRichErrors {
     bytes internal constant PROXY_DESTINATION_CANNOT_BE_NIL =
         hex"01ecebea";
 
-    enum MakerPoolAssignmentErrorCodes {
-        MAKER_ADDRESS_ALREADY_REGISTERED,
-        MAKER_ADDRESS_NOT_REGISTERED,
-        MAKER_ADDRESS_NOT_PENDING_ADD,
-        POOL_IS_FULL
-    }
-
     // bytes4(keccak256("InvalidProtocolFeePaymentError(uint8,uint256,uint256)"))
     bytes4 internal constant INVALID_PROTOCOL_FEE_PAYMENT_ERROR_SELECTOR =
         0xefd6cb33;
@@ -126,6 +119,18 @@ library LibStakingRichErrors {
     // bytes4(keccak256("InvalidWethAssetDataError()"))
     bytes internal constant INVALID_WETH_ASSET_DATA_ERROR =
         hex"24bf322c";
+
+    enum MakerPoolAssignmentErrorCodes {
+        MAKER_ADDRESS_ALREADY_REGISTERED,
+        MAKER_ADDRESS_NOT_REGISTERED,
+        MAKER_ADDRESS_NOT_PENDING_ADD,
+        POOL_IS_FULL
+    }
+
+    enum OperatorShareErrorCodes {
+        OPERATOR_SHARE_MUST_BE_BETWEEN_0_AND_100,
+        CAN_ONLY_DECREASE_OPERATOR_SHARE
+    }
 
     // solhint-disable func-name-mixedcase
     function MiscalculatedRewardsError(
@@ -320,18 +325,20 @@ library LibStakingRichErrors {
         );
     }
 
-    function InvalidPoolOperatorShareError(
+    function OperatorShareError(
+        OperatorShareErrorCodes errorCode,
         bytes32 poolId,
-        uint32 poolOperatorShare
+        uint32 operatorShare
     )
         internal
         pure
         returns (bytes memory)
     {
         return abi.encodeWithSelector(
-            INVALID_POOL_OPERATOR_SHARE_ERROR_SELECTOR,
+            OPERATOR_SHARE_ERROR_SELECTOR,
+            errorCode,
             poolId,
-            poolOperatorShare
+            operatorShare
         );
     }
 
--- a/contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewardVault.sol
+++ b/contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewardVault.sol
@@ -18,10 +18,11 @@
 
 pragma solidity ^0.5.9;
 
+import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
+import "../libs/LibStakingRichErrors.sol";
 import "../interfaces/IStakingEvents.sol";
 import "../interfaces/IStakingPoolRewardVault.sol";
 import "../immutable/MixinStorage.sol";
-import "../libs/LibStakingRichErrors.sol";
 
 
 /// @dev This mixin contains logic for interfacing with the Staking Pool Reward Vault (vaults/StakingPoolRewardVault.sol)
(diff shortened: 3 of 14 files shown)


contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -21,24 +21,6 @@ pragma solidity ^0.5.9;
 
 interface IStructs {
 
-    /// @dev Allowed signature types.
-    enum SignatureType {
```
mzhu25: this must've slipped past in [2104](https://github.com/0xProject/0x-monorepo/pull/2104) 😨

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewardVault.sol
```diff
@@ -32,6 +33,38 @@ contract MixinStakingPoolRewardVault is
...
+
+        _;
+    }
+
+    /// @dev Asserts that the sender is the operator of the input pool or the input maker.
+    /// @param poolId Pool sender must be operator of.
+    /// @param makerAddress Address of a maker in the pool.
+    modifier onlyStakingPoolOperatorOrMaker(bytes32 poolId, address makerAddress) {
+        address poolOperator = rewardVault.operatorOf(poolId);
+        if (msg.sender != poolOperator && msg.sender != makerAddress) {
```
jalextowle: We can make the case where `msg.sender != makerAddress` cheaper by only calling the function `operatorOf` when we are sure that `msg.sender == makerAddress`. Solidity implements short-circuiting in high-level `if-statements`, so I think that one really clean way to refactor this would be to write:

```
address poolOperator;
if (
    msg.sender != makerAddress &&
    msg.sender != (poolOperator = rewardVault.operatorOf(poolId))
) { ... }
```

I've tested something like this, and it definitely works, so feel free to rip that off. I believe that this is as efficient as this will get, but I'd love to see any alternatives that you come up with :)

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -172,35 +173,78 @@ contract StakingPoolRewardVault is
...
+    /// Note that this is only callable by the staking contract, and will revert if the new operator
+    /// share value is greater than the old value.
+    /// @param poolId Unique Id of pool.
+    /// @param newOperatorShare The newly decreased percentage of any rewards owned by the operator.
+    function decreaseOperatorShare(bytes32 poolId, uint32 newOperatorShare)
+        external
+        onlyStakingContract
+        onlyNotInCatastrophicFailure
+    {
+        Pool memory pool = poolById[poolId];
```
jalextowle: Nit: I'm generally in favor of caching values in a `memory` structure in cases like this, but it seems like `pool` is only read from once and written to once. I think that it would be better to just use `poolById[poolId]` in this case to minimize the amount that's written to `memory`. This will save us an `SLOAD`, and potentially an `SSTORE` depending on how smart `solc` has gotten.

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -211,8 +255,8 @@ contract StakingPoolRewardVault is
         view
         returns (uint256)
     {
-        Balance memory balance = balanceByPoolId[poolId];
-        return balance.operatorBalance + balance.membersBalance;
+        Pool memory pool = poolById[poolId];
```
jalextowle: Nit: We could probably make the same change to this as the above; however, we are really only gaining minimal efficiency, whereas in the above case it actually could be significant, so this one is up to you.

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -172,35 +173,78 @@ contract StakingPoolRewardVault is
...
+    /// @param newOperatorShare The newly decreased percentage of any rewards owned by the operator.
+    function decreaseOperatorShare(bytes32 poolId, uint32 newOperatorShare)
+        external
+        onlyStakingContract
+        onlyNotInCatastrophicFailure
+    {
+        Pool memory pool = poolById[poolId];
+        uint32 oldOperatorShare = pool.operatorShare;
+
+        if (newOperatorShare > oldOperatorShare) {
```
jalextowle: I wonder if it makes sense to just revert if `newOperatorShare == oldOperatorShare`. Otherwise, it could be an expensive operation that doesn't do anything. Also, the event would be a little confusing in this case. 

Thoughts?

contracts/staking/contracts/test/TestStorageLayout.sol
```diff
@@ -73,9 +73,6 @@ contract TestStorageLayout is
             if sub(nextPoolId_slot, slot) { revertIncorrectStorageSlot() }
             slot := add(slot, 1)
 
-            if sub(poolById_slot, slot) { revertIncorrectStorageSlot() }
```
jalextowle: Should this have been removed? We still have `poolById`, right?

mzhu25: yep, the new `poolById` is the mapping formerly known as `balancesByPoolId`, which lives in `StakingPoolRewardVault`

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
mzhu25: makes sense to me

contracts/staking/contracts/test/TestStorageLayout.sol
jalextowle: Ah gotcha. Makes sense!

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -320,18 +325,20 @@ library LibStakingRichErrors {
         );
     }
 
-    function InvalidPoolOperatorShareError(
+    function OperatorShareError(
```
jalextowle: Nice consolidation!

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewardVault.sol
```diff
@@ -32,6 +33,41 @@ contract MixinStakingPoolRewardVault is
...
+    }
+
+    /// @dev Asserts that the sender is the operator of the input pool or the input maker.
+    /// @param poolId Pool sender must be operator of.
+    /// @param makerAddress Address of a maker in the pool.
+    modifier onlyStakingPoolOperatorOrMaker(bytes32 poolId, address makerAddress) {
+        address poolOperator;
+        if (
+            msg.sender != makerAddress &&
+            msg.sender != (poolOperator = rewardVault.operatorOf(poolId))
```
jalextowle: 🙏

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -91,14 +92,14 @@ contract StakingPoolRewardVault is
...
+            uint256 membersPortion
         )
     {
         // update balance of pool
-        Balance memory balance = balanceByPoolId[poolId];
-        (operatorPortion, poolPortion) = _incrementBalanceStruct(balance, amount, operatorOnly);
-        balanceByPoolId[poolId] = balance;
-        return (operatorPortion, poolPortion);
+        Pool memory pool = poolById[poolId];
+        (operatorPortion, membersPortion) = _incrementPoolBalances(pool, amount, operatorOnly);
```
jalextowle: Non-blocking: Thoughts on passing a storage pointer to `_incrementPoolBalances` instead of a memory pointer? If all we do is pass the memory pointer to `_incrementPoolBalances` and then update afterwards then I think that this makes sense.

As the tag indicates, this is not a merge blocking comment, but I do think that it could be an improvement. Guess I missed this last review

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -172,35 +173,76 @@ contract StakingPoolRewardVault is
...
+    /// Note that this is only callable by the staking contract, and will revert if the new operator
+    /// share value is greater than the old value.
+    /// @param poolId Unique Id of pool.
+    /// @param newOperatorShare The newly decreased percentage of any rewards owned by the operator.
+    function decreaseOperatorShare(bytes32 poolId, uint32 newOperatorShare)
+        external
+        onlyStakingContract
+        onlyNotInCatastrophicFailure
+    {
+        uint32 oldOperatorShare = poolById[poolId].operatorShare;
```
jalextowle: 🙏

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -172,35 +173,76 @@ contract StakingPoolRewardVault is
...
-                poolOperatorShare
+                operatorShare
             ));
         }
 
         // pool must not exist
-        Balance memory balance = balanceByPoolId[poolId];
-        if (balance.initialized) {
+        Pool memory pool = poolById[poolId];
+        if (pool.initialized) {
```
jalextowle: Slipped me during the last review: Just like below, I actually think that it would be more efficient to update the `poolById[poolId]` storage pointer directly since only one read is being performed and `SLOAD` operations are fairly expensive now. I'm going to add this to my weekly Todos to investigate more thoroughly, but I think that this would be a good change in the meantime.

contracts/staking/test/pools_test.ts
```diff
@@ -330,6 +349,38 @@ blockchainTests('Staking Pool Management', env => {
             );
             await poolOperator.addMakerToStakingPoolAsync(poolId, lastMakerAddress, revertError);
         });
+        it('Operator should successfully decrease their share of rewards', async () => {
```
jalextowle: Let's add a test for the `==` case now that it's implemented.

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
jalextowle: You can just pass in `poolById[poolId]`.

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
jalextowle: Nice! I like the update to a storage pointer! +1 for readability

contracts/staking/test/pools_test.ts
```diff
@@ -330,6 +349,56 @@ blockchainTests('Staking Pool Management', env => {
...
+            const increasedOperatorShare = operatorShare + 1;
+            const revertError = new StakingRevertErrors.OperatorShareError(
+                StakingRevertErrors.OperatorShareErrorCodes.CanOnlyDecreaseOperatorShare,
+                poolId,
+                increasedOperatorShare,
+            );
+            // decrease operator share
+            await poolOperator.decreaseStakingPoolOperatorShareAsync(poolId, increasedOperatorShare, revertError);
+        });
+        it('Should fail if operator calls decreaseStakingPoolOperatorShare but newOperatorShare == oldOperatorShare', async () => {
```
jalextowle: 🙏

PR #2116, opened 2019-08-30
Add 4 byte ids to `EIP1271Wallet.isValidSignature` data

abandeali1:
## Description

- Adds a 4 byte selector at the start of any `data` passed into `EIP1271Wallet.isValidSignature`. This can be used to differentiate which type of data is being passed into thee receiving contract (see `TestValidatorWallet` as an example)
- Removes the `examples` directory in the `exchange` package

## Checklist:


contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -206,41 +189,39 @@ contract TestValidatorWallet is
...
-        DataType dataType = _hashDataTypes[hash];
-        require(
-            dataType != DataType.None,
-            "EXPECTED_NO_DATA_TYPE"
-        );
-        if (dataType == DataType.Order) {
-            // Decode the first parameter as an `Order` type.
-            LibOrder.Order memory order = abi.decode(data, (LibOrder.Order));
+        bytes4 dataId = data.readBytes4(0);
+        if (dataId == IEIP1271Data(address(0)).OrderWithHash.selector) {
```
dorothy-zbornak: So much easier. :pray:

contracts/exchange/contracts/test/TestValidatorWallet.sol
```diff
@@ -206,41 +189,39 @@ contract TestValidatorWallet is
...
+            // Decode the order and hash
+            LibOrder.Order memory order;
+            (order, hash) = abi.decode(
+                data.slice(4, data.length - 1),
```
dorothy-zbornak: Why do we shave off an extra byte from the end here?

dorothy-zbornak: Wait, actually, should this be `data.length - 4`? I guess it doesn't really matter.

abandeali1: Whoops, off by 1 error. This should be `data.length` (it's start/end indexes, not start index + length).

PR #2121, opened 2019-09-02
Reduce Exchange codesize

abandeali1:
## Description

The bulk of this PR is focused on reducing the `Exchange` codesize by consolidating logic into internal functions. Modifiers in particular increase codesize by a lot, as their logic is inlined within each function they are used. The following changes were made as a result:

- Modifiers now make heavy use of internal functions. This ends up only adding a few jumps to the functions rather than inlining all of the modifier code. This change alone was not enough to allow the `Exchange` contract to be deployed, though. 
- `fillOrderNoThrow` was made internal (removes one public function from the interface). I can't think of a good reason someone would use this function by itself (100% of the 584 `fillOrderNoThrow` calls in the 2.1 contract have been internal).
- `getOrdersInfo` has been removed (removes one public function from the interface). This function has literally *never* been called on-chain in its existence, and it's easy enough to implement in the `DevUtils` contract (or with other better validation methods).
- A bunch of other smaller redundant functions were made internal.
- A lot of the logic for `Validator` and `EIP1271Wallet` signature types has been consolidated into internal functions (with a new `EIP1271SignatureError`).
- Fixes a bug in the `Wallet` / `Validator` / `EIP1271Wallet` signature types where the `returndatasize` could be less than 32 (this wasn't actually dangerous because `LibBytes.readBytes4` would ensure that it was at least of length 4). 
- Fixes a bug in `LibFillResults.calculateFillResults`, where we use `makerAssetFilledAmount` divided by `order.makerAssetAmount` to calculate the `makerFeePaid`. This could result in 2 small rounding errors when calculating the fee, rather than 1 rounding error if only the taker amounts are used.

## Checklist:


contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -470,37 +489,27 @@ contract MixinSignatureValidator is
     )
         private
         view
-        returns (bool isValid)
+        returns (bool)
```
mzhu25: nit: remove `isValid` from above `@return`

contracts/test-utils/src/web3_wrapper.ts
```diff
@@ -46,7 +46,7 @@ const gethConfigs = {
 };
 const ganacheConfigs = {
     shouldUseInProcessGanache: true,
-    shouldAllowUnlimitedContractSize: env.parseBoolean(EnvVars.UnlimitedContractSize),
+    shouldAllowUnlimitedContractSize: true,
```
jalextowle: Nice! With this in mind, we should probably make a task for writing tests similar to `contracts/exchange/test/codesize.ts`.

contracts/utils/contracts/src/ReentrancyGuard.sol
```diff
@@ -25,25 +25,33 @@ import "./LibRichErrors.sol";
...
 
     /// @dev Functions with this modifer cannot be reentered. The mutex will be locked
     ///      before function execution and unlocked after.
     modifier nonReentrant() {
+        _lockMutexOrThrowIfAlreadyLocked();
+        _;
+        _unlockMutex();
+    }
+
+    function _lockMutexOrThrowIfAlreadyLocked()
```
jalextowle: Out of scope: It would be nice if we had a library for automagically spinning up new mutexes on an as needed basis.

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -596,33 +578,53 @@ contract MixinSignatureValidator is
...
+        isValid = _staticCallEIP1271WalletWithReducedSignatureLength(
+            validatorAddress,
+            data,
+            signature,
+            21  // The last 21 bytes of the signature (validatorAddress + signatureType) are removed before making the staticcall
+        );
+        return isValid;
+    }
+
+    /// @dev Performs a staticcall to and EIP11271 compiant `isValidSignature` function and validates the output.
```
mzhu25:
```suggestion
    /// @dev Performs a staticcall to an EIP11271 compliant `isValidSignature` function and validates the output.
```

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -596,33 +578,53 @@ contract MixinSignatureValidator is
...
+    /// @param verifyingContractAddress Address of EIP1271Wallet or Validator contract.
+    /// @param data Arbitrary signed data.
+    /// @param signature Proof that the hash has been signed by signer. Bytes will be temporarily be popped
+    ///                  off of the signature before calling `isValidSignature`.
+    /// @param ignoredSignatureBytesLen The amount of bytes that will be temporarily popped off the the signature.
+    /// @return The validity of the signature and the raw returnData of the call.
```
mzhu25: nit: we don't actually return the raw returnData

contracts/utils/contracts/src/Refundable.sol
```diff
@@ -22,32 +22,58 @@ pragma solidity ^0.5.9;
...
-        if (shouldNotRefund) {
+        if (_areRefundsDisabled()) {
             _;
         } else {
-            shouldNotRefund = true;
+            _disableRefund();
             _;
-            shouldNotRefund = false;
-            _refundNonzeroBalance();
+            _enableRefund();
```
jalextowle: This is probably as readable as it gets, but considering inlining, I think that it probably makes sense to put the functionality of calling

```
_enableRefund();
_refundNonZeroBalance();
```

into a single internal function to save a little bit more on codesize.

contracts/utils/contracts/src/Refundable.sol
```diff
@@ -22,32 +22,58 @@ pragma solidity ^0.5.9;
...
+            _refundNonZeroBalance();
         }
     }
 
-    function _refundNonzeroBalance()
+    function _refundNonZeroBalanceIfEnabled()
+        internal
+    {
+        if (!_areRefundsDisabled()) {
```
jalextowle: Nit: Is there an issue with inlining this function? It's really short already, so we could save the `JUMPs` and `JUMPDESTs` to and from by removing the `_areRefundsDisabled` function.

contracts/exchange/contracts/src/MixinTransactions.sol
```diff
@@ -190,6 +186,20 @@ contract MixinTransactions is
         }
     }
 
+    /// @dev Sets the currentContextAddress if the current context is not msg.sender.
+    /// @param signerAddress Address of the transaction signer.
+    /// @param contextAddress The current context address.
+    function _setCurrentContextAddressIfRequired(
```
jalextowle: 🙏

contracts/utils/test/lib_bytes.ts
```diff
@@ -668,7 +668,7 @@ describe('LibBytes', () => {
     describe('writeBytesWithLength', () => {
         it('should successfully write short, nested array of bytes when it takes up the whole array', async () => {
             const testBytesOffset = new BigNumber(0);
-            const emptyByteArray = ethUtil.bufferToHex(new Buffer(shortTestBytesAsBuffer.byteLength));
+            const emptyByteArray = ethUtil.bufferToHex(Buffer.alloc(shortTestBytesAsBuffer.byteLength));
```
jalextowle: Out of scope: After doing some digging on `Buffer.alloc`, it seems like we could speed our tests up a bit by refactoring them to use [`Buffer.allocUnsafe`](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_allocunsafe_size). I think that it could be done pretty easily, but it's not high pri or even necessary.

contracts/utils/test/lib_bytes.ts
```diff
@@ -1074,5 +1074,63 @@ describe('LibBytes', () => {
...
+        it('should return a null byte array if length is set to 0', async () => {
+            const result = await libBytes.publicWriteLength.callAsync(
+                byteArrayLongerThan32Bytes,
+                constants.ZERO_AMOUNT,
+                constants.NULL_BYTES,
+            );
+            expect(result).to.eq(constants.NULL_BYTES);
+        });
+        it('should return the same byte array if length is unchanged', async () => {
+            const byteLen = (byteArrayLongerThan32Bytes.length - 2) / 2;
```
mzhu25: can we use `fromHex(byteArrayLongerThan32Bytes).length` here?

contracts/utils/test/lib_bytes.ts
```diff
@@ -1074,5 +1074,63 @@ describe('LibBytes', () => {
...
+            const newLen = new BigNumber(byteLen).dividedToIntegerBy(2);
+            const result = await libBytes.publicWriteLength.callAsync(
+                byteArrayLongerThan32Bytes,
+                newLen,
+                constants.NULL_BYTES,
+            );
+            expect(result).to.eq(
+                byteArrayLongerThan32Bytes.slice(
+                    0,
+                    newLen
```
mzhu25: might be worth introducing a couple of const functions to convert from hex string length to byte length and vice versa

contracts/exchange-forwarder/test/utils/forwarder_test_factory.ts
```diff
@@ -164,7 +164,7 @@ export class ForwarderTestFactory {
             this._forwarderFeeRecipientAddress,
         );
 
-        const ordersInfoBefore = await this._exchangeWrapper.getOrdersInfoAsync(orders);
+        const ordersInfoBefore = await Promise.all(orders.map(order => this._exchangeWrapper.getOrderInfoAsync(order)));
```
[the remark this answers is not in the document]
abandeali1: This was a quick fix, but we should probably implement `getOrdersInfo` in the `DevUtils` contract and use that everywhere where possible.

contracts/exchange/contracts/src/MixinExchangeCore.sol
```diff
@@ -483,7 +483,11 @@ contract MixinExchangeCore is
             if (address(this).balance >= protocolFee) {
                 valuePaid = protocolFee;
             }
-            IStaking(feeCollector).payProtocolFee.value(valuePaid)(order.makerAddress, takerAddress, protocolFee);
+            IStaking(feeCollector).payProtocolFee.value(valuePaid)(
```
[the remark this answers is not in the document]
abandeali1: I actually implemented this in a different branch just now, but it's going over the max codesize now =/

contracts/exchange/contracts/src/MixinSignatureValidator.sol
```diff
@@ -177,24 +177,29 @@ contract MixinSignatureValidator is
...
             signature
         );
+
+        // Any signature type that makes an external call needs to be revalidated
+        // with every partial fill
         needsRegularValidation =
             signatureType == SignatureType.Wallet ||
             signatureType == SignatureType.Validator ||
             signatureType == SignatureType.EIP1271Wallet;
+        return needsRegularValidation;
```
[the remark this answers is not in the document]
abandeali1: Just for consistency, we do this pretty much everywhere else.

contracts/exchange/contracts/src/MixinExchangeCore.sol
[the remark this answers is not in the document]
jalextowle: Maybe we should just exclude them? Pretty rough though :(

contracts/utils/contracts/src/Refundable.sol
abandeali1: Removing this adds 107 bytes of codesize, which pushes us off the edge.

contracts/utils/contracts/src/Refundable.sol
abandeali1: Nice, saved about 10 bytes :p

contracts/exchange/contracts/src/MixinExchangeCore.sol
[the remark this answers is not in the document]
abandeali1: I'll put up a draft PR and see if I can squeeze some codesize out.

contracts/utils/test/lib_bytes.ts
abandeali1: I'm going to leave this for now. After switching to `fromHex(..).length`, I think this is the only place we are directly converting from byte length to length.

contracts/utils/contracts/src/Refundable.sol
jalextowle: Wow, that's crazy. I'm going to have to look into that 🤔

contracts/utils/contracts/src/Refundable.sol
jalextowle: Every little bit counts I guess haha

contracts/utils/test/lib_bytes.ts
```diff
@@ -1067,12 +1067,84 @@ describe('LibBytes', () => {
...
+            const byteLen = fromHex(byteArrayLongerThan32Bytes).length;
+            const newLen = new BigNumber(byteLen).multipliedBy(2);
+            const result = await libBytes.publicWriteLength.callAsync(
+                byteArrayLongerThan32Bytes,
+                newLen,
+                byteArrayLongerThan32Bytes,
+            );
+            expect(result).to.eq(`${byteArrayLongerThan32Bytes}${byteArrayLongerThan32Bytes.slice(2)}`);
+        });
+        it('should result in the same byte array if length is reduced and reset', async () => {
```
jalextowle: 🙏

contracts/utils/contracts/src/Refundable.sol
jalextowle: As mentioned offline, was the addition of 107 bytes a result of removing `_refundNonZeroBalanceIfEnabled` or `_areRefundsDisabled`? The first would make sense, but the latter really wouldn't.

abandeali1: It was from removing `_areRefundsDisabled` haha

PR #2126, opened 2019-09-03
Tests for new staking mechanics

hysz:
## Description

These are the tests for #2118. I put them in this PR for readability.

## Checklist:


contracts/staking/contracts/test/LibFeeMathTest.sol
```diff
@@ -21,8 +21,6 @@
 pragma solidity ^0.5.5;
```
jalextowle: Nit: Update to `0.5.9`.

contracts/staking/test/utils/Simulation.ts
```diff
@@ -1,9 +1,11 @@
+/*
```
jalextowle: Will this be updated in this PR, or another one?

contracts/staking/test/simulations_test.ts
```diff
@@ -1,3 +1,6 @@
+/*
```
jalextowle: Same as below

contracts/staking/test/stake_test.ts
```diff
@@ -1,47 +1,357 @@
 import { ERC20ProxyContract, ERC20Wrapper } from '@0x/contracts-asset-proxy';
```
jalextowle: (File Wide Comment) These tests are amazingly readable, and after reading through the `StakerActor`, they are also rigorous 😄. The actor pattern is sweet!

contracts/staking/test/stake_test.ts
```diff
@@ -1,47 +1,357 @@
...
+            await staker.goToNextEpochAsync(); // stake is now inactive
+            await staker.goToNextEpochAsync(); // stake is now withdrawable
+            await staker.moveStakeAsync({ state: StakeState.Inactive }, { state: StakeState.Active }, amount);
+            await staker.goToNextEpochAsync(); // stake is active and not withdrawable
+            await staker.goToNextEpochAsync(); // stake is active and not withdrawable
+            await staker.unstakeAsync(amount, new StringRevertError('INSUFFICIENT_FUNDS'));
+        });
+    });
+    describe('Simulations', () => {
+        it('Simulation from Staking Spec', async () => {
```
jalextowle: It would be nice to have a simulation that is a bit shorter for debugging purposes. If this simulation finds a bug, it may be a bit non-trivial to figure out where the bug lives.

contracts/staking/test/actors/staker_actor.ts
```diff
@@ -52,95 +51,186 @@ export class StakerActor extends BaseActor {
...
-        expect(balances.stakeBalance, 'stake balance').to.be.bignumber.equal(expectedBalances.stakeBalance);
         expect(balances.stakeBalanceInVault, 'stake balance, recorded in vault').to.be.bignumber.equal(
             expectedBalances.stakeBalanceInVault,
         );
         expect(balances.withdrawableStakeBalance, 'withdrawable stake balance').to.be.bignumber.equal(
             expectedBalances.withdrawableStakeBalance,
         );
-        expect(balances.activatableStakeBalance, 'activatable stake balance').to.be.bignumber.equal(
-            expectedBalances.activatableStakeBalance,
+        expect(balances.activeStakeBalance.current, 'active stake balance (current)').to.be.bignumber.equal(
```
jalextowle: +1 for the named expect. This is super nice to read, and I'm sure it's nice to debug.

contracts/staking/test/actors/staker_actor.ts
```diff
@@ -31,19 +29,20 @@ export class StakerActor extends BaseActor {
...
         expectedStakerBalances.zrxBalance = initStakerBalances.zrxBalance.minus(amount);
-        expectedStakerBalances.stakeBalance = initStakerBalances.stakeBalance.plus(amount);
         expectedStakerBalances.stakeBalanceInVault = initStakerBalances.stakeBalanceInVault.plus(amount);
-        expectedStakerBalances.activatedStakeBalance = initStakerBalances.activatedStakeBalance.plus(amount);
+        expectedStakerBalances.activeStakeBalance.current = initStakerBalances.activeStakeBalance.current.plus(amount);
+        expectedStakerBalances.activeStakeBalance.next = initStakerBalances.activeStakeBalance.next.plus(amount);
         await this.assertBalancesAsync(expectedStakerBalances);
         // check zrx balance of vault
         const finalZrxBalanceOfVault = await this._stakingWrapper.getZrxTokenBalanceOfZrxVaultAsync();
         expect(finalZrxBalanceOfVault).to.be.bignumber.equal(initZrxBalanceOfVault.plus(amount));
```
jalextowle: Nit: Add descriptions to expects across the whole file for consistency and easier debugging. It may even be good to pass the calling functions name into `assertBalancesAsync` so that the error messages are even nicer when `assertBalancesAsync` is called by other functions in the actor.

contracts/staking/test/utils/Simulation.ts
hysz: The simulations will be the last to get updated, after the implementation is finalized and the unit tests are done. Likely, next week.

contracts/staking/test/simulations_test.ts
hysz: (see above)

contracts/staking/test/stake_test.ts
hysz: That's a good point. I worked to include all of the edge cases individually, in the tests above. I wanted to keep this just to have some semblance of a legitimate sequence of interactions.

jalextowle: Fair enough! Before writing this comment, I had the thought that the Simulations would likely not get hit in the case of a failure (because of an earlier failure). Up to you about whether to include a shorter sequence. 

Out of scope: Given the complexity of the staking logic, I think that some potential bugs could be found in pathologically long sequences (if there are any). I think it would really behoove us to do some fuzzing using the created actors.

contracts/staking/test/actors/finalizer_actor.ts
```diff
@@ -0,0 +1,190 @@
...
+        this._operatorByPoolId = _.cloneDeep(operatorByPoolId);
+        this._membersByPoolId = _.cloneDeep(membersByPoolId);
+    }
+
+    public async finalizeAsync(rewards: Reward[] = []): Promise<void> {
+        // cache initial info and balances
+        const operatorShareByPoolId = await this._getOperatorShareByPoolIdAsync(this._poolIds);
+        const rewardVaultBalanceByPoolId = await this._getRewardVaultBalanceByPoolIdAsync(this._poolIds);
+        const memberBalancesByPoolId = await this._getMemberBalancesByPoolIdAsync(this._membersByPoolId);
+        // compute expecnted changes
```
jalextowle: Nit:
```suggestion
        // compute expected changes
```

contracts/staking/test/rewards_test.ts
```diff
@@ -0,0 +1,600 @@
...
+            // this should not go do the delegator
+            await payProtocolFeeAndFinalize(rewardNotForDelegator);
+            // sanity check final balances
+            await validateEndBalances({
+                stakerEthVaultBalance_1: rewardForDelegator,
+                poolRewardVaultBalance: rewardNotForDelegator,
+                operatorRewardVaultBalance: rewardNotForDelegator,
+            });
+        });
+        it('Should stop collecting rewards after undelegating, after several epochs', async () => {
```
jalextowle: +1 for test coverage

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -198,7 +198,7 @@ contract MixinExchangeFees is
 
         // sanity check - this is a gas optimization that can be used because we assume a non-zero
         // split between stake and fees generated in the cobb-douglas computation (see below).
-        if (totalFeesCollected == 0 || totalWeightedStake == 0) {
```
dorothy-zbornak: What was the purpose of this? Seems to just add more logic.

hysz: We may no longer need this with the new math. But these were denominators in the old cobb-douglas implementation, so this avoided any division by zero.

dorothy-zbornak: I mean, why remove the check?

hysz: Ah ~ if there's no stake in a given pool then the entire reward goes to the operator. Similarly, if no pools are staked then the reward goes entirely to operators. Note that it's still in a pool's best interest to stake since alpha != 0. So if all the pools are unstaked, then one pool could come along with relatively low volume but non-zero stake and take a good portion of the total reward.

PR #2127, opened 2019-09-04
Wrap `payProtocolFee` with rich revert + reduce codesize

abandeali1:
## Checklist:


contracts/exchange/contracts/src/MixinProtocolFees.sol
```diff
@@ -51,4 +54,125 @@ contract MixinProtocolFees is
...
+        address feeCollector,
+        uint256 exchangeBalance,
+        uint256 protocolFee,
+        address makerAddress,
+        address takerAddress
+    )
+        internal
+        returns (uint256 valuePaid)
+    {
+        // Do not send a value with the call if insufficient balance
```
jalextowle:
```suggestion
        // Do not send a value with the call if the exchange has an insufficient balance
```

PR #2128, opened 2019-09-04
Read-Only mode in proxy

hysz:
## Description

This PR implements Read-Only mode for the Staking Contracts. There's a read-only proxy that emulates a static-delegate-call into the logic contract. Any state changes are reverted.

Enable this in the proxy by calling `StakingProxy.setReadOnlyMode(true)`.
Disable this in the proxy by calling `StakingProxy.setReadOnlyMode(false)`.

The nice thing about this implementation is that it does not depend on a read-only modifier.

To-dos:
1. Move the asm into a library
2. Revert when changing staking contract when in read-only mode.

## Checklist:


contracts/staking/contracts/src/ReadOnlyProxy.sol
```diff
@@ -0,0 +1,107 @@
...
+    /// @dev Executes a read-only call to the staking contract, via `revertDelegateCall`.
+    ///      By routing through `revertDelegateCall` any state changes are reverted.
+    // solhint-disable no-complex-fallback
+    function ()
+        external
+    {
+        address thisAddress = address(this);
+        bytes4 revertDelegateCallSelector = this.revertDelegateCall.selector;
+
+        assembly {
```
hysz: (@todo - move this into a library DRY)

contracts/staking/contracts/src/ReadOnlyProxy.sol
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    /// @dev Executes a delegate call to the staking contract, if it is set.
+    ///      This function always reverts with the return data.
+    function revertDelegateCall()
+        external
+    {
+        address _readOnlyProxyCallee = readOnlyProxyCallee;
+        if (_readOnlyProxyCallee == address(0)) {
```
dorothy-zbornak: Why even have this check? This function can't modify state.

contracts/staking/contracts/src/libs/LibProxy.sol
```diff
@@ -0,0 +1,109 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
+import "./LibStakingRichErrors.sol";
+
+
+library LibProxy {
```
dorothy-zbornak: This is such a cool library! I don't even care that the auditors will hate it. :trollface:

contracts/staking/contracts/src/libs/LibProxy.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    enum RevertRule {
+        REVERT_ON_ERROR,
+        ALWAYS_REVERT,
+        NEVER_REVERT
+    }
+
+    /// @dev Executes a read-only call to the staking contract, via `revertDelegateCall`.
+    ///      By routing through `revertDelegateCall` any state changes are reverted.
+    // solhint-disable no-complex-fallback
```
dorothy-zbornak: Do we still need this linter toggle?

contracts/staking/contracts/src/libs/LibProxy.sol
```diff
@@ -0,0 +1,109 @@
...
+    function proxyCall(
+        address destination,
+        RevertRule revertRule,
+        bytes4 customSelector,
+        bool ignoreSelector
+    )
+        internal
+    {
+        if (destination == address(0)) {
```
dorothy-zbornak: I kind of feel like there's no point checking for this. A wrong address is just as bad as `0x0`.

contracts/staking/contracts/src/ReadOnlyProxy.sol
hysz: See the proxy lib, which now reverts if it is address zero. Reverting is important to distinguish between a legitimate NIL response and a misconfigured proxy.

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -134,6 +134,10 @@ library LibStakingRichErrors {
     bytes4 internal constant INVALID_STAKE_STATUS_ERROR_SELECTOR =
         0xb7161acd;
 
+    // bytes4(keccak256("ProxyDestinationCannotBeNil()"))
+    bytes4 internal constant PROXY_DESTINATION_CANNOT_BE_NIL =
```
mzhu25: I think we're trying to use this optimization for rich errors that don't have any parameters:
https://github.com/0xProject/0x-monorepo/blob/88e56356c424cc834948891ecb0918ced7c56bb9/contracts/exchange-libs/contracts/src/LibMathRichErrors.sol#L8

contracts/staking/test/catastrophy_test.ts
```diff
@@ -0,0 +1,89 @@
...
+import { BigNumber } from '@0x/utils';
+import { LogWithDecodedArgs } from 'ethereum-types';
+import * as _ from 'lodash';
+
+import { StakingProxyReadOnlyModeSetEventArgs } from '../src';
+
+import { StakingWrapper } from './utils/staking_wrapper';
+
+// tslint:disable:no-unnecessary-type-assertion
+blockchainTests.resets('Catastrophy Tests', () => {
```
mzhu25:
```suggestion
blockchainTests.resets('Catastrophe Tests', () => {
```
😆

hysz: Hahahahaha this word is my nemesis!

contracts/staking/contracts/src/libs/LibProxy.sol
```diff
@@ -0,0 +1,109 @@
...
+        ALWAYS_REVERT,
+        NEVER_REVERT
+    }
+
+    /// @dev Executes a read-only call to the staking contract, via `revertDelegateCall`.
+    ///      By routing through `revertDelegateCall` any state changes are reverted.
+    function proxyCall(
+        address destination,
+        RevertRule revertRule,
+        bytes4 customSelector,
```
abandeali1: Is it correct to say that if you specify a non-zero `customSelector` then `ignoreSelector` will always be `true`?

hysz: No—but is the distinction clear if they’re renamed to `customEgressSelector` and `ignoreIngressSelector`?

contracts/staking/contracts/src/libs/LibProxy.sol
hysz: That's true... I guess this is more of an initialization sanity check than a correctness check.

hysz: We just had a cheeky chat about this in the coder cave. I think we're going to leave these sanity checks in for now (they're used throughout staking) but remove them if we do end up using constants.

contracts/staking/contracts/src/libs/LibProxy.sol
hysz: For reference:
```
                                                             | customEgressSelector | ignoreIngressSelector
StakingProxy fallback                        | no                                    | no
ReadOnlyProxy fallback                    | yes                                   | no
ReadOnlyProxy revertDelegateCall  | no                                    | yes
```

contracts/staking/test/catastrophe_test.ts
```diff
@@ -0,0 +1,89 @@
+import { ERC20ProxyContract, ERC20Wrapper } from '@0x/contracts-asset-proxy';
+import { DummyERC20TokenContract } from '@0x/contracts-erc20';
+import { blockchainTests, describe, expect, provider, web3Wrapper } from '@0x/contracts-test-utils';
```
mzhu25: with blockchainTests we no longer need to import `provider` and `web3Wrapper`. we have these in the `BlockchainTestsEnvironment`

PR #2129, opened 2019-09-05
Remove StakingWrapper

mzhu25:
## Description
`StakingApiWrapper` exposes the `StakingContract` API on proxy so that we no longer have to manually encode/decode the call data like we were doing in `StakingWrapper`. The `StakingApiWrapper` class keeps all the deployed contracts in one place and also has some convenience functions carried over from `StakingWrapper`. 
`deployAndConfigureContractsAsync` does what it sounds like and returns a `StakingApiWrapper` instance (this avoids the `...ContractIfExists` pattern)

## Types of changes

## Checklist:


contracts/staking/test/utils/staking_wrapper.ts
```diff
@@ -1,549 +0,0 @@
-import { ERC20ProxyContract } from '@0x/contracts-asset-proxy';
```
hysz: That is one mighty fine RIP!

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+            const createStakingPoolLog = txReceipt.logs[0];
+            const poolId = (createStakingPoolLog as any).args.poolId;
+            return poolId;
+        },
+
+        getZrxTokenBalanceOfZrxVaultAsync: async (): Promise<BigNumber> => {
+            return this.zrxTokenContract.balanceOf.callAsync(this.zrxVaultContract.address);
+        },
+
+        signApprovalForStakingPool: (
```
hysz: We can nix this now that makers join by calling into the contract 🤓

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+export class StakingApiWrapper {
+    public stakingProxyContract: StakingContract; // Wrapped as StakingContract to borrow API
+    public stakingContract: StakingContract; // The real Staking.sol contract
+    public zrxVaultContract: ZrxVaultContract;
+    public ethVaultContract: EthVaultContract;
+    public rewardVaultContract: StakingPoolRewardVaultContract;
+    public zrxTokenContract: DummyERC20TokenContract;
+    public utils = {
+        // Epoch Utils
+        goToNextEpochAsync: async (): Promise<TransactionReceiptWithDecodedLogs> => {
```
hysz: We can probably nix this and just move the call to `finalizeFees` into `skipToNextEpochAsync`

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+            await this.utils.fastForwardToNextEpochAsync();
+            // increment epoch in contracts
+            const txReceipt = await this.utils.goToNextEpochAsync();
+            // mine next block
+            await this._web3Wrapper.mineBlockAsync();
+            return txReceipt;
+        },
+
+        // Other Utils
+        moveStakeAsync: async (
```
hysz: If we create a constructor in `StakeInfo` that takes an optional `poolId` then we could also nix this function :)

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+import { SignedStakingPoolApproval, StakeInfo } from './types';
+
+export class StakingApiWrapper {
+    public stakingProxyContract: StakingContract; // Wrapped as StakingContract to borrow API
+    public stakingContract: StakingContract; // The real Staking.sol contract
+    public zrxVaultContract: ZrxVaultContract;
+    public ethVaultContract: EthVaultContract;
+    public rewardVaultContract: StakingPoolRewardVaultContract;
+    public zrxTokenContract: DummyERC20TokenContract;
+    public utils = {
```
hysz: I like this!

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+}
+
+/**
+ * Deploys and configures all staking contracts and returns a StakingApiWrapper instance, which
+ * holds the deployed contracts and serves as the entry point for their public functions.
+ */
+export async function deployAndConfigureContractsAsync(
+    env: BlockchainTestsEnvironment,
+    ownerAddress: string,
+    erc20ProxyContract: ERC20ProxyContract,
```
dorothy-zbornak: Should we just go ahead and deploy these ERC20/Proxy contracts here as well? Is there an integration test suite that doesn't just deploy standard implementations?

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+        this._web3Wrapper = env.web3Wrapper;
+        this._accounts = accounts;
+        this.stakingContract = stakingContract;
+        this.zrxVaultContract = zrxVaultContract;
+        this.ethVaultContract = ethVaultContract;
+        this.rewardVaultContract = rewardVaultContract;
+        this.zrxTokenContract = zrxTokenContract;
+
+        // disguise the staking proxy as a StakingContract
+        const decoderArtifacts = _.merge(artifacts, erc20Artifacts) as { [contractName: string]: ContractArtifact };
```
dorothy-zbornak: Is all this equivalent to this?
```ts
const logDecoderDependencies = _.mapValues({...artifacts, ...erc20Artifacts}, v => v.compilerOutput.abi);
```

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+        this.stakingProxyContract = new StakingContract(
+            stakingProxyContract.address,
+            env.provider,
+            {
+                ...txDefaults,
+                from: ownerAddress,
+                to: stakingProxyContract.address,
+                gas: 3000000,
```
dorothy-zbornak: Why so low? Also, can we use `3e6` notation? Seems clearer to me.

contracts/staking/test/utils/api_wrapper.ts
dorothy-zbornak: Any reason why these functions aren't just at the root level of the `StakingApiWrapper` class? Otherwise the root level object seems less of a "wrapper" and more of just data fields.

mzhu25: my rationale was that it looks a bit weird to have these functions at the same level as the contracts, i.e. I think `stakingApiWrapper.createStakingPoolAsync(...)` would look kind of weird alongside stuff like `stakingApiWrapper.stakingProxyContract.getStakingPoolIdOfMaker.callAsync(this._owner)`

contracts/staking/test/utils/api_wrapper.ts
mzhu25: These parameters were the ones in StakingWrapper, we could change it but I don't think it really matters?

contracts/staking/test/vaults_test.ts
```diff
@@ -33,26 +33,43 @@ blockchainTests('Staking Vaults', env => {
...
-            const stakingContractAddress = stakingWrapper.getStakingContract().address;
+            const poolId = await stakingApiWrapper.utils.createStakingPoolAsync(poolOperator, operatorShare);
+            const stakingContractAddress = stakingApiWrapper.stakingContract.address;
             const notStakingContractAddress = poolOperator;
             // create pool in vault
-            await stakingWrapper.rewardVaultRegisterPoolAsync(poolId, operatorShare, stakingContractAddress);
+            await stakingApiWrapper.rewardVaultContract.registerStakingPool.awaitTransactionSuccessAsync(
+                poolId,
+                operatorShare,
+                { from: stakingContractAddress },
```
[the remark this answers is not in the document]
hysz: Yeah, that's why it's skipped for now. We're going to add a Todo and address it during unit testing.

contracts/staking/test/utils/api_wrapper.ts
dorothy-zbornak: Depends on how hot of a mess finalization gets, but should be fine for now.

contracts/staking/test/utils/api_wrapper.ts
dorothy-zbornak: idk, doesn't look weird to me, but I'm possibly in the minority

contracts/staking/test/utils/api_wrapper.ts
mzhu25: good call

PR #2130, opened 2019-09-05
Feature/staking/protocol fees

jalextowle:
## Description

This pull request introduces WETH support to the staking contract. Despite the fact that supporting a new asset typically increases the complexity of a system, in this case, only the `payProtocolFee` and `finalizeFees` functions were made more complex.

### payProtocolFee
The logic is now entirely compatible with ETH or WETH, and this works whether the maker is registered in a staking pool or not. These changes are also unit tested in the file `test/protocol_fees.ts`.

### finalizeFees
Only a small change needed to be made to this function. Now this function will unwrap WETH and withdraw all of the unwrapped ETH to the staking contract. After this, finalization can carry on normally, so this is not a very big change.

## Testing instructions

`yall`

## Types of changes

## Checklist:


contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -95,15 +99,39 @@ contract MixinExchangeFees is
...
+        ) {
+            LibRichErrors.rrevert(LibStakingRichErrors.InvalidProtocolFeePaymentError(
+                protocolFeePaid,
+                msg.value
+            ));
+        }
+
+        // Transfer the protocol fee to this address if it should be paid in WETH.
+        if (msg.value == 0) {
+            erc20Proxy.transferFrom(
```
hysz: This comment is more directed to architecture opposed to the logic, but I'm wondering if we should deploy a separate `feeProxy`.

(i)
When users set their allowance on the ERC20 proxy they permit our exchange contract to transfer tokens on their behalf. The user then explicitly consents to the transfer of token X in amount Y when they sign an order.

Here we're overloading the ERC20 proxy to collect a hidden or implicit fee. We can broadcast to our community that there now exists a fee, but it's still a passive agreement that the average user may not understand. 

(ii)
The ERC20 proxy is super OP. It can transfer any ERC20 that the user has set allowance for. In this code, we really only need access to WETH.

(iii) 
This also raises a question around the proxy used to stake ZRX. On one hand, this doesn't break (i) because the user still explicitly agrees to the transfer by calling `stake()`. However (ii) does apply because in that scenario we really only need access to ZRX.

I propose we have either (i) a `wethFeeProxy` and a `zrxStakeProxy`, or (ii) a combined `stakingProxy` that users can set allowances for WETH and ZRX on. Note that this can just be the generic ERC20 Proxy, but redeployed.

@abandeali1 curious to get your thoughts.

contracts/staking/contracts/src/fees/MixinExchangeManager.sol
```diff
@@ -45,6 +45,17 @@ contract MixinExchangeManager is
         _;
     }
 
+    /// @dev Adds a new erc20 proxy.
+    /// @param erc20AssetProxy The asset proxy that will transfer erc20 tokens.
+    function addERC20AssetProxy(address erc20AssetProxy)
```
hysz: Nit - `setERC20AssetProxy` and `emit ERC20AssetProxySet` ~ for consistency w other similar functions in this package.

contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -102,4 +102,12 @@ interface IStructs {
         uint256 numerator;
         uint256 denominator;
     }
+
+    /// @dev Tracks the amount of protocol fees paid in ETH and WETH in a staking pool.
+    /// @param ethBalance The balance of the pool in ETH.
+    /// @param wethBalance The balance of the pool in WETH.
+    struct ProtocolFeeBalance {
```
hysz: I don't think this is used.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -95,15 +99,39 @@ contract MixinExchangeFees is
         payable
         onlyExchange
     {
-        uint256 amount = msg.value;
+        // If the protocol fee payment is invalid, revert with a rich error.
+        if (
+            protocolFeePaid == 0 ||
+            (msg.value != protocolFeePaid && msg.value != 0)
+        ) {
+            LibRichErrors.rrevert(LibStakingRichErrors.InvalidProtocolFeePaymentError(
```
hysz: If `protocolFeePaid == 0` then I think the error is kind of confusing:
```
InvalidProtocolFeePaymentError(
  expected: 0,
  actual: 0
)
```

contracts/staking/contracts/src/immutable/MixinConstants.sol
```diff
@@ -43,4 +45,10 @@ contract MixinConstants is
     uint64 constant internal INITIAL_TIMELOCK_PERIOD = INITIAL_EPOCH;
 
     uint256 constant internal MIN_TOKEN_VALUE = 10**18;
+
+    // The address of the canonical WETH contract -- this will be used as an alternative to ether for paying protocol fees.
```
hysz: Nit - please move this to `MixinDeploymentConstants` ~ which has constants that must be updated for each deployment.

contracts/staking/test/protocol_fees.ts
```diff
@@ -0,0 +1,361 @@
...
+                makerAddress,
+                payerAddress,
+                DEFAULT_PROTOCOL_FEE_PAID,
+                { from: exchange, value: 0 },
+            );
+
+            // Ensure that the correct number of logs were recorded.
+            expect(receipt.logs.length).to.be.eq(1);
+
+            // Ensure that the correct log was recorded.
```
hysz: Wondering if we could move this into an actor or assertion wrapper, to follow similar patterns used elsewhere in our tests. Would also reduce a lot of the code that's repeated in these tests.

contracts/staking/contracts/test/TestStaking.sol
```diff
@@ -0,0 +1,53 @@
...
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "../src/Staking.sol";
+
+
+contract TestStaking is
+    Staking
+{
+    // Stub out `payProtocolFee` to be the naive payProtocolFee function so that tests will
```
hysz: Could we say a bit more here? I'm still not sure why tests would fail against the actual implementation.

contracts/staking/contracts/test/TestProtocolFees.sol
```diff
@@ -0,0 +1,41 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "../src/Staking.sol";
+
+
+contract TestProtocolFees is
```
hysz: Nice, I really like this pattern 💯

contracts/staking/test/protocol_fees.ts
```diff
@@ -0,0 +1,361 @@
...
+
+import {
+    artifacts,
+    TestProtocolFeesContract,
+    TestProtocolFeesERC20ProxyContract,
+    TestProtocolFeesERC20ProxyTransferFromCalledEventArgs,
+} from '../src';
+
+// tslint:disable:no-unnecessary-type-assertion
+blockchainTests('Protocol Fee Unit Tests', env => {
```
hysz: Solid coverage!

contracts/staking/contracts/src/immutable/MixinConstants.sol
```diff
@@ -43,4 +45,10 @@ contract MixinConstants is
...
+    address constant internal WETH_ADDRESS = address(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+
+    // abi.encodeWithSelector(IAssetData(address(0)).ERC20Token.selector, WETH_ADDRESS)
+    bytes constant internal WETH_ASSET_DATA = hex"f47261b0000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2";
```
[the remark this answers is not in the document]
abandeali1: This isn't a merge blocker, but it's probably a good idea to add comments here with all of the testnet WETH addresses/assetDatas. We should also add WETH to the `contract-addresses` package now that it is used within the protocol.

contracts/staking/test/protocol_fees.ts
```diff
@@ -0,0 +1,361 @@
...
+    let nonExchange: string;
+    let makerAddress: string;
+    let payerAddress: string;
+
+    // The contract that will be used for testng `payProtocolFee` and `_unwrapETH`.
+    let protocolFees: TestProtocolFeesContract;
+    let proxy: TestProtocolFeesERC20ProxyContract;
+
+    // The log decoder that will be used to decode logs from TestProtocolFeesERC20Proxy.
+    let logDecoder: LogDecoder;
```
abandeali1: The wrappers should automatically decode logs now, so we shouldn't ever need to use `LogDecoder` anymore (you can just pass `artifacts` in as the last parameter when deploying a contract). We should probably make a task to remove it from all tests.

contracts/staking/contracts/test/TestStaking.sol
abandeali1: What are your thoughts on creating some internal getter function for `WETH_ASSET_DATA` (`getWethAssetData`), using that in the actual `payProtocolFee`, and then stubbing it out in this contract? That way we can test with actual WETH transfers. It's probably a couple extra gas but the tradeoff feels worth it to me.

contracts/staking/test/protocol_fees.ts
```diff
@@ -0,0 +1,361 @@
...
+            );
+            return expect(tx).to.revertWith(expectedError);
+        });
+
+        it('should revert if `protocolFeePaid` is zero with zero value sent', async () => {
+            const expectedError = new StakingRevertErrors.InvalidProtocolFeePaymentError(
+                constants.ZERO_AMOUNT,
+                constants.ZERO_AMOUNT,
+            );
+            const tx = protocolFees.payProtocolFee.sendTransactionAsync(
```
abandeali1: Nit: we should be using `awaitTransactionSuccessAsync` here.

contracts/staking/contracts/test/TestStaking.sol
abandeali1: This could actually be a cool pattern to use for all deployment constants. Not sure if we have any similar situations with any of the other constants at the moment.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
abandeali1: I have mixed feeling on it. I think I'm leaning towards having a separate `wethAssetProxy` but continuing to use the current `ERC20Proxy` for ZRX.

The good news is that we don't need to decide now since this can literally be a fork of the current `ERC20Proxy`. Let's just rename this variable `wethAssetProxy` (might as well rename ZRXVault.erc20Proxy to `zrxAssetProxy` while we're at it).

contracts/staking/contracts/src/interfaces/IStructs.sol
jalextowle: Whoops... I'll remove

contracts/staking/test/protocol_fees.ts
jalextowle: Yeah I can work on that. It should be a very simple actor for now.

contracts/staking/test/protocol_fees.ts
jalextowle: Ah, that's what I was missing. I didn't specify `artifacts`

contracts/staking/contracts/test/TestStaking.sol
jalextowle: As discussed offline, I'm a big fan and I think that we should do it on an as-needed basis.

contracts/staking/contracts/test/TestProtocolFees.sol
```diff
@@ -0,0 +1,41 @@
...
+    Staking
+{
+    function setPoolIdOfMaker(bytes32 poolId, address makerAddress)
+        external
+    {
+        poolIdByMakerAddress[makerAddress] = poolId;
+    }
+
+    function getActivePoolsByEpoch()
```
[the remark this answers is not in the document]
jalextowle: As discussed offline, a task has been made to implement this change in a separate PR, where naming conventions will also be updated.

contracts/staking/contracts/src/immutable/MixinStorage.sol
```diff
@@ -18,6 +18,8 @@
 
 pragma solidity ^0.5.9;
 
+import "@0x/contracts-asset-proxy/contracts/src/interfaces/IAssetData.sol";
```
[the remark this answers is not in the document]
jalextowle: It will actually be used now, since the check is in the constructor.

contracts/staking/contracts/test/TestStaking.sol
jalextowle: Since the integration tests aren't being implemented in this PR, I'm going to hold off on this until it's needed.

jalextowle: I ended up just removing it since they do work with the current impl

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -217,8 +263,9 @@ contract MixinExchangeFees is
             );
 
             // store pool stats
+            uint256 protocolFeeBalance = protocolFeesThisEpochByPool[poolId];
```
hysz: Is this an optimization?

contracts/staking/contracts/test/TestStaking.sol
```diff
@@ -0,0 +1,56 @@
...
+
+
+contract TestStaking is
+    Staking
+{
+    /*
+    // Stub out `payProtocolFee` to be the naive payProtocolFee function so that tests will
+    // not fail for WETH protocol fees. These tests will fail otherwise because many of them
+    // will transfer
+    function payProtocolFee(
```
hysz: Do we still need this, since it's commented out?

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
jalextowle: It looks like it actually makes it less efficient. I think that it's a holdover from how it would have worked before WETH protocol fees.

contracts/staking/contracts/test/TestStaking.sol
jalextowle: Nope, I'll remove.

PR #2131, opened 2019-09-06
Staking: Hyper-Parameters, init() Pattern, Dust Stake

dorothy-zbornak:
## Description
buttons

### `init()` Pattern
When a contract is attached to the `StakingProxy` via `attachStakingContract()`, or through the constructor, it will now `delegatecall()` the new contract's `init()` function. This function is in charge of initializing and/or migrating storage. Currently, the only initialization that happens is in `MixinScheduler` and `MixinParams`.

### Hyper-Parameters
The following hyper-parameters are no longer constants and can be configured via the `MixinHyperParameters.tune()` function:
- `epochDurationInSeconds`
- `rewardDelegatedStakeWeight`
- `minimumPoolStake`
- `maximumMakersInPool`
- `cobbDouglasAlphaNumerator`
- `cobbDouglasAlphaDenomintor`

### Dust Stake
In `payProtocolFee()`, we do no longer credit pools with `stake < minimumPoolStake` to mitigate finalization griefing from dusty pools.

## Types of changes

## Checklist:


contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -22,6 +22,24 @@ import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
```
dorothy-zbornak: Git diff is very confused by this file for some reason.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
```diff
@@ -0,0 +1,163 @@
...
+    uint256 internal cobbDouglasAlphaDenomintor = 2;
+
+    /// @dev Set all hyperparameters at once.
+    /// @param _epochDurationInSeconds Minimum seconds between epochs.
+    /// @param _rewardDelegatedStakeWeight How much delegated stake is weighted vs operator stake, in ppm.
+    /// @param _minimumPoolStake Minimum amount of stake required in a pool to collect rewards.
+    /// @param _maximumMakersInPool Maximum number of maker addresses allowed to be registered to a pool.
+    /// @param _cobbDouglasAlphaNumerator Numerator for cobb douglas alpha factor.
+    /// @param _cobbDouglasAlphaDenomintor Denominator for cobb douglas alpha factor.
+    function tune(
```
hysz: Nit - I think either `tuneHyperParameters` or `setHyperParameters` would be good to use here, as it'll provide context to the overall Staking API.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
```diff
@@ -0,0 +1,163 @@
...
+import "../libs/LibStakingRichErrors.sol";
+import "./MixinConstants.sol";
+
+
+contract MixinHyperParameters is
+    IStakingEvents,
+    MixinConstants,
+    Ownable
+{
+    // Minimum seconds between epochs.
```
hysz: We should add these to `TestStorageLayout`.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
```diff
@@ -0,0 +1,163 @@
...
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
+import "@0x/contracts-utils/contracts/src/Ownable.sol";
+import "../interfaces/IStakingEvents.sol";
+import "../libs/LibStakingRichErrors.sol";
+import "./MixinConstants.sol";
+
+
+contract MixinHyperParameters is
```
hysz: Is the terminology intended to communicate that these must be set prior to the start of an epoch? If so, it doesn't look like this is enforced so perhaps we could note this more explicitly in the function description for `tune` below.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
```diff
@@ -0,0 +1,163 @@
...
+    IStakingEvents,
+    MixinConstants,
+    Ownable
+{
+    // Minimum seconds between epochs.
+    uint256 internal epochDurationInSeconds = 2 weeks;
```
hysz: I think state upgrades will be more robust if we maintain all of the state variables in `MixinStorage`. Especially because we frequently have to reorder imports to avoid the linearization errors.

contracts/staking/contracts/src/Staking.sol
```diff
@@ -55,4 +55,16 @@ contract Staking is
...
     {}
+
+    /// @dev Initialize storage owned by this contract.
+    ///      This function should not be called directly.
+    ///      The StakingProxy contract will call it in `attachStakingContract()`.
+    function init()
+        external
+        onlyOwner
+    {
+        // DANGER! When performing upgrades, take care to modify this logic
```
hysz: If this should never be changed then it could be good to put in the `immutable` directory under `MixinInit` or something. But, it also makes sense here in place of a constructor so up to you.

contracts/staking/contracts/test/TestStorageLayout.sol
```diff
@@ -25,11 +25,13 @@ import "../src/interfaces/IStructs.sol";
...
 
 contract TestStorageLayout is
+    MixinConstants,
+    Ownable,
     MixinStorage
 {
     function assertExpectedStorageLayout()
         public
-        view
+        pure
```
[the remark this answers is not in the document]
dorothy-zbornak: Both. The slot is known at compile time and we aren't accessing state here.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    /// @dev Asserts that cobb douglas alpha values are valid.
+    function _assertValidCobbDouglasAlpha(
+        uint256 numerator,
+        uint256 denominator
+    )
+        private
+        pure
+    {
+        if (int256(numerator) < 0
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, all `FixedMath` is signed. I'll add comments about this.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
dorothy-zbornak: Not really. I don't think changing any of these mid-epoch would break things. I called them hyper-parameters because they're the only tweakable parameters in the system.

contracts/staking/contracts/src/Staking.sol
dorothy-zbornak: The opposite, actually. It SHOULD be modified for migrations because you probably don't want to clobber prior state.

contracts/staking/test/utils/types.ts
```diff
@@ -1,5 +1,25 @@
...
+export interface HyperParameters {
+    epochDurationInSeconds: BigNumber;
+    rewardDelegatedStakeWeight: BigNumber;
+    minimumPoolStake: BigNumber;
+    maxMakersInPool: BigNumber;
+    cobbDouglasAlphaNumerator: BigNumber;
+    cobbDouglasAlphaDenomintor: BigNumber;
+}
+
+export interface StakingPoolApproval {
```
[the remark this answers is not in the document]
dorothy-zbornak: Oof, I think this snuck in during a rebase.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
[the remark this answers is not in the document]
dorothy-zbornak: Actually, these alpha components should be `uint32`s anyway, so we wouldn't need this check.

contracts/staking/contracts/src/Staking.sol
hysz: Oh now that I think of it, what would you think of moving the logic from `tests/TestStorageLayout` to `src` and invoking it on each `init`?

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
hysz: Ah, gotcha. I might be slightly more in favor of a boring name, like `MixinMutableParameters` or `MixinUpgradableParameters`. But if the meaning is obvious to the other protocologists then I'm cool with it. It does have a darn nice ring to it ツ

contracts/staking/contracts/src/Staking.sol
dorothy-zbornak: Hmm it does feel a little excessive to assert compile-time knowns at run-time. Presumably, we would never deploy code that didn't pass the layout tests. Do we flatten/transform then recompile the code before deploying? If so, then, yeah it's probably a good idea.

contracts/staking/contracts/src/immutable/MixinHyperParameters.sol
dorothy-zbornak: What about just `MixinParams`, `setParams()`, and `getParams()`?

contracts/staking/contracts/src/Staking.sol
hysz: That's a good question. Historically, the deployment process has been pretty manual and not very streamlined. To my knowledge, there isn't anything mechanical that forces the tests to be run on the binary before its deployed. Perhaps we can bring it up in standup today.

contracts/staking/test/params.ts
```diff
@@ -0,0 +1,164 @@
...
+                };
+                const tx = setParamsAndAssertAsync(params);
+                const expectedError = new StakingRevertErrors.InvalidParamValueError(
+                    StakingRevertErrors.InvalidParamValueErrorCode.InvalidMaximumMakersInPool,
+                );
+                return expect(tx).to.revertWith(expectedError);
+            });
+        });
+
+        describe('cobb-douglas alpha', () => {
```
jalextowle: I feel like this would be a good candidate for `combinatorial testing`. If you agree, I think it would make sense to parameterize these tests.

contracts/staking/test/migration.ts
```diff
@@ -0,0 +1,173 @@
...
+        });
+
+        blockchainTests.resets('upgrade', async () => {
+            let proxyContract: TestStakingProxyContract;
+
+            before(async () => {
+                proxyContract = await deployStakingProxyAsync();
+            });
+
+            it('incm', async () => {
```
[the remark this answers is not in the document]
dorothy-zbornak: Lol this is a typo 😅

contracts/staking/test/params.ts
dorothy-zbornak: Hmm, I don't think it's worth it. No real logic is behind these functions. We already fuzz test the Cobb Douglas function in its tests.

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -61,9 +64,7 @@ contract StakingProxy is
         external
         onlyOwner
     {
-        stakingContract = _stakingContract;
-        readOnlyProxyCallee = _stakingContract;
-        emit StakingContractAttachedToProxy(_stakingContract);
```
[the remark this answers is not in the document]
dorothy-zbornak: I think this was accidentally removed during a rebase. :innocent:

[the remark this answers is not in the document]
dorothy-zbornak: Oh wait, it's still there. I just moved it to `_attachStakingContract()`.

contracts/staking/test/protocol_fees.ts
```diff
@@ -1,238 +1,379 @@
-import { blockchainTests, constants } from '@0x/contracts-test-utils';
```
jalextowle: Really nice job on this. These tests are super clean!

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -89,4 +90,19 @@ contract StakingProxy is
 
         emit ReadOnlyModeSet(readOnlyMode);
     }
+
+    /// @dev Attach a staking contract; future calls will be delegated to the staking contract.
+    /// @param _stakingContract Address of staking contract.
+    function _attachStakingContract(address _stakingContract)
+        private
+    {
+        stakingContract = readOnlyProxyCallee = _stakingContract;
```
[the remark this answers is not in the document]
hysz: Yeah I like this syntax too 👍

PR #2133, opened 2019-09-06
Sanity check on order offset when hashing

hysz:
## Description

A cheeky assertion to handle an exceptional internal error that should never arise.

## Checklist:

### DIFF
--- a/contracts/exchange-libs/contracts/src/LibOrder.sol
+++ b/contracts/exchange-libs/contracts/src/LibOrder.sol
@@ -67,7 +67,7 @@ library LibOrder {
         address senderAddress;          // Address that is allowed to call Exchange contract methods that affect this order. If set to 0, any address is allowed to call these methods.
         uint256 makerAssetAmount;       // Amount of makerAsset being offered by maker. Must be greater than 0.
         uint256 takerAssetAmount;       // Amount of takerAsset being bid on by maker. Must be greater than 0.
-        uint256 makerFee;               // Fee paid to feeRecipient by maker when order is filled. 
+        uint256 makerFee;               // Fee paid to feeRecipient by maker when order is filled.
         uint256 takerFee;               // Fee paid to feeRecipient by taker when order is filled.
         uint256 expirationTimeSeconds;  // Timestamp in seconds at which order expires.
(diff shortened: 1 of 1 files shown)


contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -133,6 +133,12 @@ library LibOrder {
         // ));
 
         assembly {
+            // Assert order offset (this is an internal error that should never be triggered)
+            let orderOffset := order
```
abandeali1: Why create another variable here? We already reference the offset with just `order` below.

contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -133,6 +133,12 @@ library LibOrder {
...
+            if lt(orderOffset, 32) {
+                revert(0, 0)
```
abandeali1: It's probably more correct to use `invalid()` here since this should never be triggered. I don't have a strong preference either way.

contracts/exchange-libs/contracts/src/LibOrder.sol
hysz: Yeah I guess it doesn't really make it more readable. I'll remove.

contracts/exchange-libs/contracts/src/LibOrder.sol
hysz: Neat, didn't know about that!

contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -133,6 +133,11 @@ library LibOrder {
         // ));
 
         assembly {
+            // Assert order offset (this is an internal error that should never be triggered)
```
jalextowle: Why is this just 32? Couldn't it be at least 64? It seems like it should be:  `lt(order, 0x80)`?

contracts/exchange-libs/contracts/src/LibOrder.sol
dorothy-zbornak: me too :astonished:

PR #2138, opened 2019-09-07
Update AssetProxyOwner

abandeali1:
## Description

This PR updates the `AssetProxyOwner` with the following changes:

- It can now process multiple function calls in a single transaction
- As a side effect of this, the `Transaction` struct's `data` field is now expected to ABI encode `bytes[] data, address[] destinations, uint256[] values`
- The `Transaction` struct's `destination` and `value` fields are now ignored. _However_, the `destination field must still be non-0 (this is checked when the transaction is submitted).
- This contract can now register a custom timelock to arbitrary functions. 
- If this contract ever wants to send a value _without_ calling a function, the `data` element must be `bytes4(0)`. This is a side effect of always reading the function selector from the `data` element. We could special case this, but I don't expect this contract will really ever be sending a value anyways.

## Checklist:


contracts/multisig/contracts/src/AssetProxyOwner.sol
```diff
@@ -16,93 +16,205 @@
...
+        uint256 length = data.length;
+        require(
+            length == destinations.length && length == values.length,
+            "EQUAL_LENGTHS_REQUIRED"
+        );
+
+        uint256 transactionConfirmationTime = confirmationTimes[transactionId];
+        for (uint i = 0; i != length; i++) {
+            // Ensure that each function call is past its timelock
+            _assertValidFunctionCall(
```
jalextowle: It would be valuable to split this into two `for-loops`. The first would verify that all of the function call's are valid, and the second would actually make all of the calls. While this may slightly decrease efficiency when all of the functions have elapsed their timelocks, this would save a lot of gas in the case that they haven't.

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+                reg.functionSelectors,
+                reg.destinations,
+                reg.functionCallTimeLockSeconds,
+                signerAddresses,
+                new BigNumber(REQUIRED_SIGNERS),
+                new BigNumber(DEFAULT_TIME_LOCK),
+            );
+            await expect(tx).to.revertWith(RevertReason.EqualLengthsRequired);
+        });
+        it('should fail if functionCallTimeLockSeconds.length != functionSelectors.length', async () => {
```
jalextowle: We might as well add a check for `should fail if functionCallTimeLockSeconds.length != destinations.length`, and maybe also a check where all of the lengths are different. It doesn't add much for now, but I think it would be valuable for regression protection.

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+                reg.functionSelectors,
+                reg.destinations,
+                reg.functionCallTimeLockSeconds,
+                signerAddresses,
+                new BigNumber(REQUIRED_SIGNERS),
+                new BigNumber(DEFAULT_TIME_LOCK),
             );
+            await expect(tx).to.revertWith(RevertReason.EqualLengthsRequired);
+        });
+        it('should allow no function calls to be registered', async () => {
```
jalextowle: 🙏

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+                artifacts,
+                reg.functionSelectors,
+                reg.destinations,
+                reg.functionCallTimeLockSeconds,
+                signerAddresses,
+                new BigNumber(REQUIRED_SIGNERS),
+                new BigNumber(DEFAULT_TIME_LOCK),
+            );
+            for (const [index, selector] of reg.functionSelectors.entries()) {
+                const timelock = await apOwner.functionCallTimeLocks.callAsync(selector, reg.destinations[index]);
```
jalextowle: 🙏

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+            const timelock = await apOwner.functionCallTimeLocks.callAsync(
+                reg.functionSelectors[0],
+                reg.destinations[0],
+            );
+            expect(timelock[0]).to.equal(true);
+            expect(timelock[1]).to.bignumber.equal(reg.functionCallTimeLockSeconds[0]);
+        });
+        it('should register multiple function calls', async () => {
+            const reg = createFunctionRegistration(2, 2, 2);
+            const apOwner = await TestAssetProxyOwnerContract.deployFrom0xArtifactAsync(
```
jalextowle: Optional: It would be pretty simple to take the logic from this function and the above function to create a parameterized unit test for different numbers of registered functions. I'm not sure how important fuzzing is in this case, but it would be nice to have the option open and to add unit tests up to 10 registered functions.

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
-            const notRemoveAuthorizedAddressData = erc20Proxy.addAuthorizedAddress.getABIEncodedTransactionData(
-                owners[0],
+    blockchainTests.resets('registerFunctionCall', () => {
+        it('should revert if not called by wallet', async () => {
+            const reg = createFunctionRegistration(1, 1, 1);
+            const tx = assetProxyOwner.registerFunctionCall.awaitTransactionSuccessAsync(
+                true,
+                reg.functionSelectors[0],
+                reg.destinations[0],
+                reg.functionCallTimeLockSeconds[0],
```
jalextowle: Nit: Let's add a `from` field to make it explicit who is actually calling this.

contracts/multisig/test/utils/asset_proxy_owner_wrapper.ts
```diff
@@ -1,72 +1,49 @@
...
+    public async submitConfirmAndExecuteTransactionAsync(
+        data: string[],
+        destinations: string[],
+        signerAddresses: string[],
+        increaseTimeSeconds: number,
+        opts: { values?: BigNumber[]; executeFromAddress?: string } = {},
+    ): Promise<{ executionTxReceipt: TransactionReceiptWithDecodedLogs; txId: BigNumber }> {
+        const submitResults = await this.submitTransactionAsync(data, destinations, signerAddresses[0], opts);
+        await this._assetProxyOwner.confirmTransaction.awaitTransactionSuccessAsync(submitResults.txId, {
+            from: signerAddresses[1],
```
jalextowle: Would there be a value in making this general enough to handle `signerAddresses.length > 2`?

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+                newTimeLock,
+            );
+            const timelock = await assetProxyOwner.functionCallTimeLocks.callAsync(
+                reg.functionSelectors[0],
+                reg.destinations[0],
+            );
+            expect(timelock[0]).to.equal(true);
+            expect(timelock[1]).to.bignumber.equal(newTimeLock);
+        });
+        it('should clear the function timelock if hasCustomTimeLock is set to false', async () => {
```
jalextowle: 🙏

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+            const transactionConfirmationTime = new BigNumber(latestTimestamp).minus(DEFAULT_TIME_LOCK);
+            const reg = createFunctionRegistration(1, 1, 1);
+            const result = assetProxyOwner.assertValidFunctionCall.callAsync(
+                transactionConfirmationTime,
+                reg.functionSelectors[0],
+                reg.destinations[0],
+            );
+            expect(result).to.be.fulfilled('');
+        });
+        it('should be successful if a registered function is called after the custom timelock', async () => {
```
jalextowle: Let's add a test for `should be successful if a registered function is called after the default timelock` to handle the case when `hasCustomTimelock == false`

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+            const executionLog = logs[data.length] as LogWithDecodedArgs<AssetProxyOwnerExecutionEventArgs>;
+            expect(executionLog.event).to.eq('Execution');
+            expect(executionLog.args.transactionId).to.bignumber.eq(txId);
+        }
+        it('should revert if the transaction is not confirmed by the required amount of signers', async () => {
+            const data = [hexRandom()];
+            const destinations = [receiver.address];
+            const results = await assetProxyOwnerWrapper.submitTransactionAsync(data, destinations, signerAddresses[0]);
+            const tx = assetProxyOwner.executeTransaction.awaitTransactionSuccessAsync(results.txId, {
+                from: signerAddresses[1],
```
jalextowle: I think it would be good to add a test for the case when the signer that `confirmed` is also the `from` address to cover the bases. Additionally, I think it's a more likely scenario than this one.

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+            );
+            const results = await assetProxyOwnerWrapper.submitConfirmAndExecuteTransactionAsync(
+                data,
+                destinations,
+                signerAddresses,
+                DEFAULT_TIME_LOCK,
+            );
+            assertReceiverCalledFromLogs(results.executionTxReceipt.logs, data, destinations, results.txId);
+        });
+        it('should fail if a single function has not passed the timelock', async () => {
```
jalextowle: If you make the change that I recommended regarding reverting if any of the functions have not elapsed their timelock, this unit test can be made even stronger by requiring that no `.call` operations were made. This should be an easy change since you are using the receiver pattern :)

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+            );
+            const tx = assetProxyOwnerWrapper.submitConfirmAndExecuteTransactionAsync(
+                data,
+                destinations,
+                signerAddresses,
+                newTimeLock.toNumber(),
+            );
+            expect(tx).to.revertWith(RevertReason.DefaultTimeLockIncomplete);
+        });
+        it('should be able to execute a transaction if called by any address', async () => {
```
jalextowle: I think tests for sending 0 value and also nonzero value should be included to test this scenario.

contracts/multisig/test/asset_proxy_owner.ts
```diff
@@ -1,511 +1,641 @@
...
+                destinations,
+                signerAddresses,
+                DEFAULT_TIME_LOCK,
+                { executeFromAddress: notSignerAddress },
+            );
+            assertReceiverCalledFromLogs(results.executionTxReceipt.logs, data, destinations, results.txId);
+        });
+        it('should be able to send value without data if past the default timelock', async () => {
+            // NOTE: elements of `data` must be at least 4 bytes long
+            const data = [constants.NULL_BYTES4];
```
jalextowle: It would probably be better if zero data was permissible. Would this be a hard change to make? I feel like it's a bit weird to have to send data in this kind of a scenario.

contracts/multisig/contracts/test/ContractCallReceiver.sol
```diff
@@ -0,0 +1,51 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.9;
+
+import "@0x/contracts-utils/contracts/src/LibBytes.sol";
+
+
+contract ContractCallReceiver {
```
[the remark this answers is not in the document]
abandeali1: Maybe some of these contracts can be moved into `test-utils`?

contracts/multisig/contracts/src/AssetProxyOwner.sol
```diff
@@ -16,93 +16,205 @@
...
-        if (_externalCall(txn.destination, txn.value, txn.data.length, txn.data)) {
-            emit Execution(transactionId);
+        Transaction storage transaction = transactions[transactionId];
+        transaction.executed = true;
+
+        // Decode batch transaction data from transaction.data
+        // `destination` and `value` fields of transaction are ignored
+        // Note that `destination` must be non-0, or the transaction cannot be submitted
+        // solhint-disable
+        (
```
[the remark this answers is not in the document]
abandeali1: Yep, it is now a requirement.

contracts/multisig/contracts/src/MultiSigWalletWithTimeLock.sol
```diff
@@ -63,7 +63,7 @@ contract MultiSigWalletWithTimeLock is
     /// @param _required Number of required confirmations.
     /// @param _secondsTimeLocked Duration needed after a transaction is confirmed and before it becomes executable, in seconds.
     constructor (
-        address[] _owners,
+        address[] memory _owners,
         uint256 _required,
         uint256 _secondsTimeLocked
```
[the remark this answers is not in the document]
abandeali1: Mixed feelings. There isn't really such a thing as a non-default if you look at this contract in isolation.

contracts/multisig/contracts/src/AssetProxyOwner.sol
abandeali1: Discussed offline and decided to keep as is for max flexibility.

contracts/multisig/test/asset_proxy_owner.ts
abandeali1: We'd just need to special case it. Since I don't think this contract will ever actually be used to send ETH, I decided to leave it as is.

contracts/multisig/test/asset_proxy_owner.ts
abandeali1: This would actually fail if the registered function had a timelock longer than the default.

abandeali1: Added a test for that case.

contracts/multisig/contracts/src/AssetProxyOwner.sol
```diff
@@ -16,93 +16,205 @@
...
+        for (uint i = 0; i != length; i++) {
+            // Ensure that each function call is past its timelock
+            _assertValidFunctionCall(
+                transactionConfirmationTime,
+                data[i],
+                destinations[i]
+            );
+            // Call each function
+            // solhint-disable-next-line avoid-call-value
+            (bool didSucceed,) = destinations[i].call.value(values[i])(data[i]);
```
hysz: Could there be a benefit to propagating the return value / revert reason?

contracts/multisig/contracts/test/ContractCallReceiver.sol
```diff
@@ -0,0 +1,51 @@
...
+    );
+
+    bytes4 constant internal ALWAYS_REVERT_SELECTOR = 0xF1F2F3F4;
+
+    function ()
+        external
+        payable
+    {
+        bytes4 selector = msg.data.readBytes4(0);
+        if (selector == ALWAYS_REVERT_SELECTOR) {
```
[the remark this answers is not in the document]
abandeali1: I didn't do that because the APOwner swallows the error and reverts with `FAILED_EXECUTION` anyways. Could add rich reverts in, but I don't think it's really worth the effort. In practice we shouldn't be submitting transactions that will ever fail (and will need to thoroughly test every single transaction that does get submitted).

contracts/multisig/contracts/src/AssetProxyOwner.sol
abandeali1: I'm fine with it either way (see above comment). This should be used pretty infrequently in practice as well.

contracts/multisig/contracts/src/AssetProxyOwner.sol
```diff
@@ -16,93 +16,205 @@
...
+    function _assertValidFunctionCall(
+        uint256 transactionConfirmationTime,
+        bytes memory data,
+        address destination
+    )
+        internal
+        view
+    {
+        bytes4 functionSelector = data.readBytes4(0);
+        TimeLock storage timeLock = functionCallTimeLocks[functionSelector][destination];
```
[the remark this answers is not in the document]
abandeali1: Isn't `timeLock` just a pointer to the storage slot here and not actually using an `sload`? Either way, I tested it and using `memory` is a bit cheaper. Went ahead and made the change.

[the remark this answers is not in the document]
hysz: That's right, although with a storage pointer every time you access a field it'll do an `sload`. Whereas with `memory` it'll at most do an `sload` on the first access; and it _should_ only use a single `sload` in this assignment since the entire struct fits into 1 word (although I'm not positive Solidity is that clever).

contracts/multisig/contracts/src/AssetProxyOwner.sol
hysz: That makes sense. Could be good for future-proofing, but not critical.

contracts/multisig/contracts/src/MultiSigWallet.sol
```diff
@@ -140,15 +175,17 @@ contract MultiSigWallet {
         ownerExists(owner)
     {
         isOwner[owner] = false;
-        for (uint i=0; i<owners.length - 1; i++)
+        for (uint256 i=0; i<owners.length - 1; i++) {
```
jalextowle: small nit: `i<owners.length - 1` => `i < owners.length - 1`

PR #2143, opened 2019-09-11
LibSafeDowncast unit tests

mzhu25:
## Description
Unit tests for `LibSafeDowncast` + a test wrapper contract to expose the downcast methods
Also removes a `.only` that snuck into 2142 😉 
## Checklist:

-   [ ] Prefix PR title with `[WIP]` if necessary.
-   [ ] Add tests to cover changes as needed.
-   [ ] Update documentation as needed.
-   [ ] Add new entries to the relevant CHANGELOG.jsons.

PR #2144, opened 2019-09-11
BaseUnitTestHelper class

moodlezoup:
## Description
This PR introduces a base class for a family of unit test helpers. The class abstracts away the difference between contract accessors and mutators and provides two protected methods: 
- `_verifyContractMethodExpectationsAsync` handles calls/transactions that are expected to succeed, checking results/receipts using the provided callbacks 
- `_verifyContractMethodRevertErrorAsync` handles calls/transactions that are expected to revert, checking the revert error. 

This class will replace `TransactionHelper`, but I will make a separate PR for those changes. 
See it in action [here](https://github.com/0xProject/0x-monorepo/pull/2143)

## Checklist:

### DIFF
--- a/contracts/test-utils/src/unit_test_helper.ts
+++ b/contracts/test-utils/src/unit_test_helper.ts
@@ -4,7 +4,7 @@ import { TransactionReceiptWithDecodedLogs } from 'ethereum-types';
 
 import { expect } from './chai_setup';
 
-type AsyncFunction<TArgs extends any[], TReturn> = (...args: TArgs) => PromiseLike<TReturn>;
+type AsyncFunction<TArgs extends any[], TReturn> = (...args: TArgs) => Promise<TReturn>;
 
 interface TransactionReturnData<TCallAsyncResult> {
     result: TCallAsyncResult;
@@ -27,25 +27,7 @@ export type TransactionExpectation<TAwaitTransactionSuccessAsyncArgs extends any
 ) => Promise<void> | void;
 
 export class BaseUnitTestHelper<TContract> {
-    protected static async _verifyContractMethodRevertErrorAsync<
-        TCallAsyncArgs extends any[],
-        TAwaitTransactionSuccessAsyncArgs extends any[],
-        TCallAsyncResult
-    >(
-        contractFunction: ContractWrapperFunction<TCallAsyncArgs, TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>,
-        expectedError: RevertError,
-        // tslint:disable-next-line: trailing-comma
-        ...args: TAwaitTransactionSuccessAsyncArgs
-    ): Promise<void> {
-        const tx = contractFunction.awaitTransactionSuccessAsync
-            ? contractFunction.awaitTransactionSuccessAsync(...args)
-            : contractFunction.callAsync(...((args as any) as TCallAsyncArgs));
-        return expect(tx).to.revertWith(expectedError);
-    }
-
-    constructor(protected readonly _testContract: TContract) {}
-
-    protected async _verifyContractMethodExpectationsAsync<
+    protected static async _verifyContractMethodExpectationsAsync<
         TCallAsyncArgs extends any[],
         TAwaitTransactionSuccessAsyncArgs extends any[],
         TCallAsyncResult
@@ -65,4 +47,22 @@ export class BaseUnitTestHelper<TContract> {
             await expectation({ result, receipt }, ...args);
         }
     }
+
+    protected static async _verifyContractMethodRevertErrorAsync<
+        TCallAsyncArgs extends any[],
+        TAwaitTransactionSuccessAsyncArgs extends any[],
+        TCallAsyncResult
+    >(
+        contractFunction: ContractWrapperFunction<TCallAsyncArgs, TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>,
+        expectedError: RevertError,
+        // tslint:disable-next-line: trailing-comma
+        ...args: TAwaitTransactionSuccessAsyncArgs
+    ): Promise<void> {
+        const tx = contractFunction.awaitTransactionSuccessAsync
+            ? contractFunction.awaitTransactionSuccessAsync(...args)
+            : contractFunction.callAsync(...((args as any) as TCallAsyncArgs));
+        return expect(tx).to.revertWith(expectedError);
+    }
+
+    constructor(protected readonly _testContract: TContract) {}
 }


contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export type TransactionExpectation<TAwaitTransactionSuccessAsyncArgs extends any[], TCallAsyncResult> = (
+    tx: Partial<TransactionReturnData<TCallAsyncResult>>,
+    // tslint:disable-next-line: trailing-comma
+    ...args: TAwaitTransactionSuccessAsyncArgs
+) => Promise<void> | void;
+
+export class BaseUnitTestHelper<TContract> {
+    protected readonly _testContract: TContract;
+    private readonly _logDecoder: LogDecoder;
```
jalextowle: Do we actually need `LogDecoder`? Our contract wrappers should be able to handle this, provided that the write `LogEventArg` type is passed in for casting. We could maybe avoid using this by accepting a template argument that specifies the log type that the function will be used for.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    protected static async _verifyContractMethodRevertErrorAsync<
+        TCallAsyncArgs extends any[],
+        TAwaitTransactionSuccessAsyncArgs extends any[],
+        TCallAsyncResult
+    >(
+        contractFunction: ContractWrapperFunction<TCallAsyncArgs, TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>,
```
jalextowle: I like the type safety here, but I think that it's a bit hard to read. Is there a way that you could create a `type` above that encapsulates this argument?

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+        expectedError: RevertError,
+        // tslint:disable-next-line: trailing-comma
+        ...args: TAwaitTransactionSuccessAsyncArgs
+    ): Promise<void> {
+        const tx = contractFunction.sendTransactionAsync
```
jalextowle: As mentioned earlier, I think that this should use `awaitTransactionSuccessAsync`.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+            ? contractFunction.sendTransactionAsync(...args)
+            : contractFunction.callAsync(...((args as any) as TCallAsyncArgs));
+        await expect(tx).to.revertWith(expectedError);
```
jalextowle: Optional nit: Oftentimes we will use a `return` here instead of `await` since this is at the end of the function. If possible, I think this would be a bit nicer.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+        this._testContract = testContract;
+        this._logDecoder = new LogDecoder(web3Wrapper, artifacts);
+    }
+
+    protected async _verifyContractMethodExpectationsAsync<
+        TCallAsyncArgs extends any[],
+        TAwaitTransactionSuccessAsyncArgs extends any[],
+        TCallAsyncResult
+    >(
+        contractFunction: ContractWrapperFunction<TCallAsyncArgs, TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>,
```
jalextowle: Same as above re creating new `types` for these parameters.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
...
+        expectations: Array<TransactionExpectation<TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>> = [],
+        // tslint:disable-next-line: trailing-comma
+        ...args: TAwaitTransactionSuccessAsyncArgs
+    ): Promise<void> {
+        const result = await contractFunction.callAsync(...((args as any) as TCallAsyncArgs));
+        let receipt;
+        if (contractFunction.sendTransactionAsync !== undefined) {
```
jalextowle: I think that you could use a ternary here once `LogDecoder` is removed, which would have the advantage of allowing `receipt` to be a `const`. I'll leave this up to your discretion, because it could potentially get ugly.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,82 @@
+import { RevertError } from '@0x/utils';
+import { Web3Wrapper } from '@0x/web3-wrapper';
+
+import { ContractArtifact, TransactionReceiptWithDecodedLogs } from 'ethereum-types';
+
+import { expect } from './chai_setup';
+import { LogDecoder } from './log_decoder';
+
+type AsyncFunction<TArgs extends any[], TResult> = (...args: TArgs) => Promise<TResult>;
```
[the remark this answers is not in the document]
mzhu25: creds to @dorothy-zbornak for the type, this is from `transaction_helper.ts`

contracts/test-utils/src/unit_test_helper.ts
mzhu25: Unfortunately there's not a whole lot we can do here because we need all three of the generics (to enforce the type of `expectations` and to recast `args` ti `TCallAsyncArgs`). It's definitely not ideal, but I'm going to leave as is for now

jalextowle: Sounds good. Thanks for looking into it.

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,68 @@
...
+        // tslint:disable-next-line: trailing-comma
+        ...args: TAwaitTransactionSuccessAsyncArgs
+    ): Promise<void> {
+        const tx = contractFunction.awaitTransactionSuccessAsync
+            ? contractFunction.awaitTransactionSuccessAsync(...args)
+            : contractFunction.callAsync(...((args as any) as TCallAsyncArgs));
+        return expect(tx).to.revertWith(expectedError);
+    }
+
+    constructor(protected readonly _testContract: TContract) {}
```
jalextowle: Beautiful syntax. I didn't even know you could do this :)

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,68 @@
...
+) => Promise<void> | void;
+
+export class BaseUnitTestHelper<TContract> {
+    protected static async _verifyContractMethodExpectationsAsync<
+        TCallAsyncArgs extends any[],
+        TAwaitTransactionSuccessAsyncArgs extends any[],
+        TCallAsyncResult
+    >(
+        contractFunction: ContractWrapperFunction<TCallAsyncArgs, TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>,
+        expectations: Array<TransactionExpectation<TAwaitTransactionSuccessAsyncArgs, TCallAsyncResult>> = [],
```
dorothy-zbornak: I wonder if it just makes sense to have this be a single `TransactionExpectation` callback, since it can always wrap an array of tests, and it can be more flexible about ordering or using intermediate results. But then I guess this function would pretty much just boil down to `TransactionHelper.getResultAndReceipt()` for mutator functions... but those are the only functions where we would really need this abstraction anyway? But if we're gonna just pass a single callback, then we might as well just return `result` and `receipt` and leave the responsibility to the caller to do the assertions on them... which is exactly what `TransactionHelper` does. ┐('～`;)┌

contracts/test-utils/src/unit_test_helper.ts
```diff
@@ -0,0 +1,68 @@
...
+            contractFunction.awaitTransactionSuccessAsync !== undefined
+                ? await contractFunction.awaitTransactionSuccessAsync(...args)
+                : undefined;
+
+        for (const expectation of expectations) {
+            await expectation({ result, receipt }, ...args);
+        }
+    }
+
+    protected static async _verifyContractMethodRevertErrorAsync<
```
dorothy-zbornak: Is this really worth wrapping? I mean is it really any clearer/pithier than `expect(contractFunction.callAsync(...)).to.revertWith(...)`?

contracts/test-utils/src/unit_test_helper.ts
dorothy-zbornak: I'm sure he had 100% nicer formatting for this until stupid prettier came along. :roll_eyes:

contracts/test-utils/src/unit_test_helper.ts
mzhu25: You're totally right, `TransactionHelper` does like 90% the same things as this class. 
But to answer your main question of why we need this when we have `TransactionHelper`, the tl;dr is that we need a layer of abstraction between the functionality provided by `TransactionHelper` (low level) and our unit tests (should be high level, but currently do a lot of heavy lifting) in order to avoid the rampant code duplication. You can see what having a middle layer of abstraction does for our tests [here](https://github.com/0xProject/0x-monorepo/pull/2143/files). 

Note that `BaseUnitTestHelper` is _not_ that middle layer, but I believe it serves as a better base for such a layer than `TransactionHelper` because of the modularity provided by the `expectations` abstraction. We'll be able to share a single `expectation` callback between multiple tests, so the idea is to have these `TransactionExpectation`s in the derived unit test helper classes along with other utility functions so that the actual unit test files are as clear and concise as possible.

contracts/test-utils/src/unit_test_helper.ts
mzhu25: This is mostly for parity, but I do think it's nice to completely abstract away `callAsync` vs `awaitTransactionSuccessAsync` as much as possible

PR #2152, opened 2019-09-13
Ensure pool exists before (un)delegating

mzhu25:
## Description
Check whether a pool with the given `poolId` exists at the beginning of `_delegateStake` and `_undelegateStake`, revert if not. Also removes the `bool initialized` field from `struct Pool` because we can just check whether the operator address is nil.

## Testing instructions
Tests added to `stake_test.ts`

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
+++ b/contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
@@ -193,16 +193,16 @@ contract StakingPoolRewardVault is
             ));
         }
 
-        // pool must not exist
+        // pool must not already exist
         Pool storage pool = poolById[poolId];
-        if (pool.initialized) {
-            LibRichErrors.rrevert(LibStakingRichErrors.PoolAlreadyExistsError(
-                poolId
+        if (pool.operatorAddress != NIL_ADDRESS) {
+            LibRichErrors.rrevert(LibStakingRichErrors.PoolExistenceError(
+                poolId,
+                true
             ));
         }
 
         // initialize pool
-        pool.initialized = true;
         pool.operatorAddress = operatorAddress;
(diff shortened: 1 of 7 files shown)


contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -193,16 +193,16 @@ contract StakingPoolRewardVault is
             ));
         }
 
-        // pool must not exist
+        // pool must not already exist
         Pool storage pool = poolById[poolId];
-        if (pool.initialized) {
-            LibRichErrors.rrevert(LibStakingRichErrors.PoolAlreadyExistsError(
-                poolId
+        if (pool.operatorAddress != NIL_ADDRESS) {
```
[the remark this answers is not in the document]
mzhu25: ah good catch, fsr I thought we already had a check in `createStakingPool`

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -76,6 +76,11 @@ contract MixinStakingPool is
         external
         returns (bytes32 poolId)
     {
+        // Operator address is used to check whether pool has been created or not
```
jalextowle: On second thought, this isn't necessary. No one has access to the private key for address zero, so this check is redundant. I think that I had an inaccurate picture of the flow for creating a staking pool previously, which is why I left the original comment. For some reason, I was under the impression that a pool could be created where the `sender` was not the same as the `operator`. 

Given more context, I'd say that we should remove this check. Sorry for flip-flopping on you here.

abandeali1: Agreed!

PR #2153, opened 2019-09-13
Implement Batch Execute Transaction

jalextowle:
### Solidity Changes
This PR implements a `batchExecute` function in the `StakingProxy` contract. This function will allow users to submit multiple calls to the staking logic contract in a single Ethereum transaction, which has several benefits including deterministic ordering of transactions, cheaper gas prices, and increased convenience. 

`batchExecute` was specifically implemented so that users could perform a "Stake and Move" action from the **Zrx Hub**. Essentially this action allows users to stake Zrx and immediately move the staked Zrx into a new state in the staking contracts. Even though this was the original use case for `batchExecute`, the current implementation is far more powerful and allows for extreme flexibility when using the staking contracts.

To implement `batchExecute`, a new function called `simpleProxyCallWithData` was added to `LibProxy`. This function allows a destination to be called with arbitrary calldata and simply returns the `success` and `returndata` that was returned by the call. This behavior leaves the decision of whether to `revert` or `return` up to the function that calls `simpleProxyCallWithData`, which is a useful deviation in behavior from `proxyCall`.

## Test Changes
Both `batchExecute` and `simpleProxyCallWithData` are tested in this PR, with the first being tested with a few basic integration tests and the second being rigorously tested. `batchExecute` will be unit tested in a later PR that includes unit tests for the `StakingProxy` contract; further testing is utterly essential to ensure that no significant vulnerabilities are enabled by `batchExecute`, but it should not present more of a risk than the calls that are already possible through the proxy.

## Testing instructions

`yall`

## Types of changes

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/StakingProxy.sol
+++ b/contracts/staking/contracts/src/StakingProxy.sol
@@ -22,6 +22,7 @@ import "@0x/contracts-utils/contracts/src/Ownable.sol";
 import "./libs/LibProxy.sol";
 import "./immutable/MixinStorage.sol";
 import "./interfaces/IStorageInit.sol";
+import "./interfaces/IStaking.sol";
 import "./interfaces/IStakingEvents.sol";
 import "./interfaces/IStakingProxy.sol";
 
@@ -69,6 +70,36 @@ contract StakingProxy is
         onlyOwner
     {
         _attachStakingContract(_stakingContract);
+    }
+
+    /// @dev Batch executes a series of calls to the exchange contract.
+    /// @param data An array of data that encodes a sequence of functions to
+    ///             call in the staking contracts.
+    function batchExecute(bytes[] memory data)
+        public
+        returns (bytes[] memory batchReturnData)
+    {
+        bool success;
+        bytes memory returnData;
+        batchReturnData = new bytes[](data.length);
+
+        // Execute all of the calls encoded in the provided calldata.
+        for (uint256 i = 0; i < data.length; i++) {
+            // Call the staking contract with the provided calldata.
+            (success, returnData) = stakingContract.simpleProxyCallWithData(data[i]);
+
+            // Revert on failure.
+            if (!success) {
+                assembly {
+                    revert(add(0x20, returnData), mload(returnData))
+                }
+            }
+
+            // Add the returndata to the batch returndata.
+            batchReturnData[i] = returnData;
+        }
+
+        return batchReturnData
     }
 
     /// @dev Detach the current staking contract.
--- a/contracts/staking/contracts/src/StakingProxy.sol
+++ b/contracts/staking/contracts/src/StakingProxy.sol
@@ -17,6 +17,7 @@
 */
 
 pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
 
 import "@0x/contracts-utils/contracts/src/Ownable.sol";
 import "./libs/LibProxy.sol";
@@ -72,6 +73,30 @@ contract StakingProxy is
         _attachStakingContract(_stakingContract);
     }
 
+    /// @dev Detach the current staking contract.
+    /// Note that this is callable only by this contract's owner.
+    function detachStakingContract()
+        external
+        onlyOwner
+    {
+        stakingContract = NIL_ADDRESS;
+        emit StakingContractDetachedFromProxy();
+    }
+
+    /// @dev Set read-only mode (state cannot be changed).
+    function setReadOnlyMode(bool readOnlyMode)
+        external
+        onlyOwner
+    {
+        if (readOnlyMode) {
+            stakingContract = readOnlyProxy;
+        } else {
+            stakingContract = readOnlyProxyCallee;
+        }
+
+        emit ReadOnlyModeSet(readOnlyMode);
+    }
+
     /// @dev Batch executes a series of calls to the exchange contract.
     /// @param data An array of data that encodes a sequence of functions to
     ///             call in the staking contracts.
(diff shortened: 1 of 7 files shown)


contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -95,6 +97,36 @@ contract StakingProxy is
         emit ReadOnlyModeSet(readOnlyMode);
     }
 
+    /// @dev Batch executes a series of calls to the exchange contract.
```
mzhu25: nit:
```suggestion
    /// @dev Batch executes a series of calls to the staking contract.
```

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -95,6 +97,36 @@ contract StakingProxy is
...
+        returns (bytes[] memory batchReturnData)
+    {
+        bool success;
+        bytes memory returnData;
+        batchReturnData = new bytes[](data.length);
+
+        // Execute all of the calls encoded in the provided calldata.
+        for (uint256 i = 0; i < data.length; i++) {
+            // Call the staking contract with the provided calldata.
+            (success, returnData) = stakingContract.simpleProxyCallWithData(data[i]);
```
mzhu25: nit: might make sense to just perform the `stakingContract` nil check at the beginning of this function and do the `delegatecall` directly here, otherwise we make that check for each call

contracts/staking/test/actors/staker_actor.ts
```diff
@@ -15,6 +15,42 @@ export class StakerActor extends BaseActor {
         this._poolIds = [];
     }
 
+    public async stakeAndMoveAsync(
```
mzhu25: 😎

contracts/staking/test/unit_tests/lib_proxy_unit_test.ts
```diff
@@ -72,27 +72,27 @@ blockchainTests.resets.only('LibProxy', env => {
...
 
-        describe('Failure Conditions', () => {
-            // Verifies that the result of a given call to `proxyCall()` results in `ProxyDestinationCannotBeNilError`
-            function checkDestinationZeroError(result: [boolean, string]): void {
-                const expectedError = new StakingRevertErrors.ProxyDestinationCannotBeNilError();
-                expect(result[0]).to.be.false();
-                expect(result[1]).to.be.eq(expectedError.encode());
-            }
+    // Verifies that the result of a given call to `proxyCall()` results in `ProxyDestinationCannotBeNilError`
+    function verifyDestinationZeroError(result: [boolean, string]): void {
```
mzhu25: `verify` > `check`👍

contracts/staking/contracts/src/StakingProxy.sol
jalextowle: Hmm, yeah that's a good point. I might just remove `simpleProxyCallWithData in favor of a normal `delegatecall` here.

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -95,6 +97,44 @@ contract StakingProxy is
...
+        public
+        returns (bytes[] memory batchReturnData)
+    {
+        // Initialize commonly used variables.
+        bool success;
+        bytes memory returnData;
+        batchReturnData = new bytes[](data.length);
+
+        // Ensure that a staking contract has been attached to the proxy.
+        if (stakingContract == address(0)) {
```
abandeali1: We can save a bunch of sloads here by setting `stakingContract` to a stack variable.

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -95,6 +97,44 @@ contract StakingProxy is
         emit ReadOnlyModeSet(readOnlyMode);
     }
 
+    /// @dev Batch executes a series of calls to the exchange contract.
+    /// @param data An array of data that encodes a sequence of functions to
+    ///             call in the staking contracts.
+    function batchExecute(bytes[] memory data)
+        public
```
abandeali1: Can this be made `external`, or do we get an internal compiler error?

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -95,6 +97,44 @@ contract StakingProxy is
...
+
+        // Ensure that a staking contract has been attached to the proxy.
+        if (stakingContract == address(0)) {
+            LibRichErrors.rrevert(
+                LibStakingRichErrors.ProxyDestinationCannotBeNilError()
+            );
+        }
+
+        // Execute all of the calls encoded in the provided calldata.
+        for (uint256 i = 0; i < data.length; i++) {
```
abandeali1: Nit: Set `data.length` to a stack variable and use `!=` instead of `<`. This saves a tiny amount of gas, but it's primarily for consistency.

contracts/staking/contracts/src/StakingProxy.sol
jalextowle: I made this public so that other proxy functions could make use of it (like a `stakeAndMove` function). It seems like we'll be leaving those functions out though, so I'll try to change this.

contracts/staking/contracts/src/StakingProxy.sol
mzhu25: bump

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -17,11 +17,12 @@
 */
 
 pragma solidity ^0.5.9;
-
+pragma experimental ABIEncoderV2;
```
abandeali1: Small nit: there should be a newline following this

PR #2154, opened 2019-09-13
Reference counting for rewards, plus unit tests for cumulative rewards tracking

hysz:
## Description

The Cumulative Reward mapping stores 2 words of data for a pool on every epoch that (i) a reward is earned, or (ii) a staker delegates. An entry for a given epoch is only necessary so long as there exists a delegator that depends on it for computing their portion of the reward. By reference counting these dependencies, we can eliminate entries that are no longer needed and prevent unnecessary chain bloat.

There are three primary changes in this PR:
1. Refactors the cumulative rewards code to improve readability and allow for unit testing.
2. Unit testing for cumulative reward tracking
3. Clearing cumulative rewards that no longer have any dependents.

Note: Most of the code in this PR is attributed to (1) and (2) above. If protocologists decide that reference counting adds unnecessary complexity and would favor chain bloat, please don't hesitate to say so. It would be very easy to remove (3) while keeping (1) and (2).

Please refer to [this spec](https://gist.github.com/hysz/346750f7d4646839060b987b80ec5327) if  you are unfamiliar with Cumulative Rewards or would like to see examples of how it is tracked (including reference counting).

## Checklist:


contracts/staking/contracts/src/libs/LibSafeDowncast.sol
```diff
@@ -57,4 +57,21 @@ library LibSafeDowncast {
         }
         return b;
     }
+
+    /// @dev Safely downcasts to a uint32
+    /// Note that this reverts if the input value is too large.
+    function downcastToUint32(uint256 a)
```
mzhu25: Assuming [2143](https://github.com/0xProject/0x-monorepo/pull/2143) gets merged first pls add unit tests for this guy (should be ez copypasta)

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -0,0 +1,230 @@
...
+        cumulativeRewardsByPool[poolId][epoch] = value;
+    }
+
+    /// @dev Unsets the cumulative reward for `poolId` at `epoch`.
+    /// @param poolId Unique id of pool.
+    /// @param epoch of cumulative reward to unset.
+    function _unsetCumulativeReward(bytes32 poolId, uint256 epoch)
+        internal
+    {
+        assert(cumulativeRewardsByPoolReferenceCounter[poolId][epoch] == 0);
```
jalextowle: If we adopt the encapsulation outlined below for reference counting, this assert can be removed.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -0,0 +1,230 @@
...
+        return IStructs.CumulativeRewardInfo({
+            cumulativeReward: cumulativeRewardsByPool[poolId][cumulativeRewardsLastStored],
+            cumulativeRewardEpoch: cumulativeRewardsLastStored
+        });
+    }
+
+    /// @dev Adds a dependency on a cumulative reward for a given epoch.
+    /// @param poolId Unique Id of pool.
+    /// @param epoch to remove dependency from.
+    /// @param mostRecentCumulativeRewardInfo Epoch of the most recent cumulative reward.
```
jalextowle: Nit: I think that this comment should be updated too.

contracts/staking/test/cumulative_reward_tracking_test.ts
```diff
@@ -0,0 +1,404 @@
+import { ERC20Wrapper } from '@0x/contracts-asset-proxy';
```
jalextowle: Nice test coverage on these integration tests!

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -191,51 +134,130 @@ contract MixinStakingPoolRewards is
...
+        bytes32 poolId,
+        IStructs.StoredBalance memory unsyncedDelegatedStakeToPoolByOwner,
+        uint256 currentEpoch
+    )
+        private
+        view
+        returns (uint256 totalReward)
+    {
+        // value is always zero in these two scenarios:
+        //   1. The owner's delegated is current as of this epoch: their rewards have been moved to the ETH vault.
```
jalextowle: Nit: I think the wording of this comment should be edited. `the owner's delegated stake is current` might be right?

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -191,51 +134,130 @@ contract MixinStakingPoolRewards is
...
+    function _computeRewardBalanceOfDelegator(
+        bytes32 poolId,
+        IStructs.StoredBalance memory unsyncedDelegatedStakeToPoolByOwner,
+        uint256 currentEpoch
+    )
+        private
+        view
+        returns (uint256 totalReward)
+    {
+        // value is always zero in these two scenarios:
```
jalextowle: Nit: Let's change `value` to `reward balance` here. I actually read this as `msg.value is always zero in these two scenarios` initially.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -191,51 +134,130 @@ contract MixinStakingPoolRewards is
...
+        //   1. The owner's delegated is current as of this epoch: their rewards have been moved to the ETH vault.
+        //   2. The current epoch is zero: delegation begins at epoch 1
+        if (unsyncedDelegatedStakeToPoolByOwner.currentEpoch == currentEpoch || currentEpoch == 0) return 0;
+
+        // compute reward accumulated during `delegatedStake.currentEpoch`;
```
jalextowle: Really clean implementation! The `computeMemberRewardOverInterval` function is really nice here.

contracts/staking/test/utils/cumulative_reward_tracking_simulation.ts
```diff
@@ -0,0 +1,172 @@
...
+    PayProtocolFee,
+    CreatePool,
+}
+
+interface TestLog {
+    event: string;
+    epoch: number;
+}
+
+export class CumulativeRewardTrackingSimulation {
```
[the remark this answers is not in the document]
hysz: Maybe something like `Director`?

contracts/staking/test/rewards_test.ts
```diff
@@ -602,6 +620,43 @@ blockchainTests.resets('Testing Rewards', env => {
                 membersRewardVaultBalance: rewardsForDelegator[1],
             });
         });
+        it('Should collect fees correctly when there is a payout for `currentEpochBalance` but not `nextEpochBalance`', async () => {
```
dorothy-zbornak: Q: Is it even possible to do a payout on `nextEpochBalance`?

contracts/staking/test/utils/cumulative_reward_tracking_simulation.ts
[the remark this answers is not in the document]
mzhu25: ^ I had the same idea

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -0,0 +1,292 @@
...
+        view
+        returns (uint256)
+    {
+        // sanity check
+        if (memberStakeOverInterval == 0) {
+            return 0;
+        }
+
+        // compute reward
+        IStructs.Fraction memory beginRatio = cumulativeRewardsByPool[poolId][beginEpoch];
```
dorothy-zbornak: Q: This structure is sparse, right? Is it possible, through regular operation, to come into this function where `endEpoch` is null, but `beginEpoch` is not? That would seem like a gnarly result, since it would be a negative/division by zero? Should we have some logic to handle these cases?

contracts/staking/contracts/src/sys/MixinParams.sol
```diff
@@ -102,8 +102,11 @@ contract MixinParams is
         _cobbDouglasAlphaDenomintor = cobbDouglasAlphaDenomintor;
     }
 
-    /// @dev Initialzize storage belonging to this mixin.
-    function _initMixinParams() internal {
+    /// @dev Assert param values before initializing them.
+    /// This must be updated for each migration.
+    function _assertMixinParamsBeforeInit()
```
dorothy-zbornak: :+1:

contracts/staking/test/utils/cumulative_reward_tracking_simulation.ts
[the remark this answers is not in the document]
dorothy-zbornak: Simulator seems fine to me.

contracts/staking/test/rewards_test.ts
hysz: Updated to clarify.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
hysz: There shouldn't be a scenario where this happens under normal operations, but I will add more assertions in this function to future-proof this claim. At present, there would be a divide-by-zero exception raised by `SafeMath`.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -191,51 +134,130 @@ contract MixinStakingPoolRewards is
...
-            numerator: numeratorNormalized,
-            denominator: denominatorNormalized
-        });
-        cumulativeRewardsByPoolLastStored[poolId] = epoch;
+        // store cumulative rewards and set most recent
+        _setCumulativeReward(
+            poolId,
+            epoch,
+            IStructs.Fraction({
+                numerator: numeratorNormalized,
```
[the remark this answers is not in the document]
hysz: Yeah, you're right. There could definitely be tighter bounds here. 

This is my thinking.

The max fixed-point value for a token is `10^27` (1 bn tokens with 18 decimal places), which consumes `90` bits.

The max fixed-point value for a base-10 token multiplier is `10^50` because `10^50 * 10^27 = 10^77` which consumes `256` bits.

So, theoretically, we could store the numerator and denominator up to `10^50` without loss of precision. 

Notes:
1. This upper-bound might have to be _slightly less_ than `10^50` (depending on the impl) because we're doing cross-multiplication, so during the computation the numerator could theoretically reach `2(10^27 * 10^50)`.
2. We know the numerator is always ≤ the denominator. So we could normalize by `10^k`; where `k` is either `0 if x≤50` or `(x - 50) if x>50`; where `10^x` is the nearest power of 10 that is greater than the denominator and `x > 50`.
3. We could get even tighter bounds using base-2, but it may be more difficult to reason about. Tbd!

PR #2155, opened 2019-09-13
Mutli-Block Finalization

dorothy-zbornak:
## Description
blocks-in-yo-face

At long last, multi-block finalization is here!

### Overview
Multi-Block finalization provides an asynchronous method for finalizing pool rewards that can be done over the span of multiple transactions, without halting the next epoch.
1. Every time a protocol fee is paid, the pool associated with it is credited, tracked, and emitted in an event.
2. When a keeper calls `endEpoch()`, the epoch will immediately advance and finalization begins.
    - During this time, the next epoch continues to accrue fees, as normal.
3. Keepers gather up all the pool IDs that were emitted during the previous epoch and pass them into `finalizePools()`-- potentially over several transactions.
    - Each time a pool is finalized, the rewards owed to that pool are paid to pool members (as a group) and to operators (individually).
    - If a delegator touches their stake in a pool that has yet to be finalized, it will be finalized on the spot.

### Notes
- An epoch cannot be ended until the previous epoch has completed finalization.
- In contrast to the original, synchronous finalization, epoch rewards are now paid in the context of the *new* epoch, not the epoch in which rewards were earned in. This turned out to be easier said than done and required significant adjustments to the stake accounting:
    - Prior, we maintained "cumulative reward" balances for `currentEpoch` and `currentEpoch-1`. Now we maintain them for `currentEpoch` and `currentEpoch+1` (I know, it's weird).
    - The behavior of reference counting had to be altered to properly accommodate this, and you'll witness this in their totally different tests.
    - We have to take care to factor in unfinalized rewards as well as rewards from previous epochs when computing a delegator's balance.
- There is a new mixin, [`MixinFinalizer`](../tree/feat/staking/hurts-like-a-mbf/contracts/staking/contracts/src/sys/MixinFinalizer.sol), which contains the bulk of the finalization logic. Notable functions include:
    - `endEpoch()`, which anyone can call to end an epoch and begin the finalization process.
    - `finalizePools()`, which finalizes multiple pools at once.
    - `_finalizePool()`, an internal function that finalizes a single pool. Called by `MixinStakingPoolRewards`  when syncing stake.
    - `_getUnfinalizedPoolRewards()`, an internal function that calculates the unfinalized rewards owed to a pool. Also called by `MixinStakingPoolRewards` in order to return accurate balances while pools are still being finalized.
- `MixinExchangeFees` has been pretty heavily modified to integrate the tracking we need.
- This PR comes with fresh unit tests for delegator/operator rewards, finalization, and protocol fees, ~but I'm still missing ones for the new logic added to `payProtocolFee()`. Any takers?~
- A notable change was made to `EthVault` and `StakingPoolRewardVault` to support efficient batching of finalization:
    - We're back to the `recordDepositFor()` pattern and just bulk transferring funds via their fallbacks.
- The `FinalizerActor` has been updated to accurately and more robustly compute finalization balances.
- There is now a minimum cap on stake for a pool to be considered active. It was a whopping 1 wei. I think this makes more sense than edge casing pools with zero stake.
- The `cobbDouglas()` function is now in its own library named, surprise, `LibCobbDouglas`.
- I added some dynamic scaling of values to `LibFractions` to fix precision issues I was running into.
- I clip cobb douglas results to the remaining reward balance in `_getUnfinalizedPoolRewards()` of `MixinFinalizer`. 
- My tests are ugly. Brace yourselves.

### Further Reading
That oooold quip doc I created on MBF still largely applies, if you want a natural language description of the finalization algorithm and state variables associated with it. Ignore the one titled "Marrying Delayed Staking with MBF." That was the work of a lunatic.

## Types of changes

## Checklist:


contracts/test-utils/src/lang_utils.ts
```diff
@@ -0,0 +1,9 @@
+import * as _ from 'lodash';
+
+/**
+ * _.zip() that clips to the shortest array.
+ */
+export function shortZip<T1, T2>(a: T1[], b: T2[]): Array<[T1, T2]> {
```
dorothy-zbornak: This is better than `_.zip()` because the return type isn't some weird `Array<[T1 | undefined, T2 | undefined]>`, which makes `_.zip()` pretty inconvenient to use in TS.

contracts/staking/contracts/src/libs/LibCobbDouglas.sol
```diff
@@ -0,0 +1,93 @@
...
+    /// @param ownerFees Fees attributed to the owner of the staking pool.
+    /// @param totalFees collected across all active staking pools in the epoch.
+    /// @param ownerStake Stake attributed to the owner of the staking pool.
+    /// @param totalStake collected across all active staking pools in the
+    ///        epoch.
+    /// @param alphaNumerator Numerator of `alpha` in the cobb-dougles function.
+    /// @param alphaDenominator Denominator of `alpha` in the cobb-douglas
+    ///        function.
+    /// @return ownerRewards Rewards for the owner.
+    function _cobbDouglas(
```
abandeali1: No need to prefix with `_` for libraries.

abandeali1: I suppose we should change that in `LibFixedMath` too.

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -158,36 +177,47 @@ contract MixinStake is
...
         _incrementNextBalance(_delegatedStakeByPoolId[poolId], amount);
 
-        // synchronizes reward state in the pool that the staker is delegating to
-        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner = _loadAndSyncBalance(_delegatedStakeToPoolByOwner[owner][poolId]);
-        _syncRewardsForDelegator(poolId, owner, initDelegatedStakeToPoolByOwner, finalDelegatedStakeToPoolByOwner);
+        // Synchronizes reward state in the pool that the staker is delegating
+        // to.
+        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner =
+            _loadAndSyncBalance(_delegatedStakeToPoolByOwner[owner][poolId]);
+        _syncRewardsForDelegator(
```
abandeali1: Are you using `prettier` or something similar here? Would love to get my hands on it!

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -51,40 +57,52 @@ contract MixinCumulativeRewards is
...
+        // Must be a value to unset
+        if (!_isCumulativeRewardSet(_cumulativeRewardsByPool[poolId][epoch])) {
+            return false;
+        }
+        // Must be no references to this CR
+        if (_cumulativeRewardsByPoolReferenceCounter[poolId][epoch] != 0) {
+            return false;
+        }
+        // Must not be the most recent CR.
+        if (_cumulativeRewardsByPoolLastStored[poolId] == epoch) {
```
abandeali1: Should this be `<=`?

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -264,59 +260,108 @@ contract MixinCumulativeRewards is
...
     {
-        // sanity check inputs
         if (memberStakeOverInterval == 0) {
             return 0;
         }
 
-        // sanity check interval
-        if (beginEpoch >= endEpoch) {
+        // Sanity check interval
+        if (beginEpoch > endEpoch) {
```
abandeali1: Can this ever actually trigger? We use an assertion for a similar check in `_forceSetMostRecentCumulativeRewardEpoch`

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -264,59 +260,108 @@ contract MixinCumulativeRewards is
...
                     endEpoch
                 )
             );
         }
 
-        // sanity check begin reward
-        IStructs.Fraction memory beginReward = _cumulativeRewardsByPool[poolId][beginEpoch];
+        // Sanity check begin reward
+        IStructs.Fraction memory beginReward =
+            _cumulativeRewardsByPool[poolId][beginEpoch];
```
abandeali1: I gotta say, I think whatever formatter you're using is a bit aggressive on the line lengths. Might be hard to undo this though 🤷‍♂

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -264,59 +260,108 @@ contract MixinCumulativeRewards is
...
-        uint256 reward = LibFractions.scaleFractionalDifference(
+        // Compute reward
+        reward = LibFractions.scaleFractionalDifference(
             endReward.numerator,
             endReward.denominator,
             beginReward.numerator,
             beginReward.denominator,
             memberStakeOverInterval
         );
-        return reward;
```
abandeali1: I like explicitly returning personally. We have been trying to do it throughout the codebase.

contracts/staking/contracts/src/sys/MixinParams.sol
```diff
@@ -231,6 +234,22 @@ contract MixinParams is
...
 
+    /// @dev Asserts that a minimum pool stake value is valid.
+    /// @param minimumStake Desired minimum pool stake.
+    function _assertValidMinimumPoolStake(
+        uint256 minimumStake
+    )
+        private
+        pure
+    {
+        if (minimumStake == 0) {
```
abandeali1: Do we actually want this assertion? The difference between 0 and 1 seems pretty arbitrary here. I'd be fine removing this.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -82,182 +137,268 @@ contract MixinStakingPoolRewards is
...
-                PPM_DENOMINATOR,
-                reward
-            );
+        // Split the reward between operator and members
+        (operatorReward, membersReward) =
+            _splitStakingPoolRewards(pool.operatorShare, reward, membersStake);
 
-        ethVault.depositFor.value(operatorPortion)(pool.operator);
+        // Record the operator's reward in the eth vault.
+        ethVault.recordDepositFor(pool.operator, operatorReward);
```
abandeali1: I kind of hate the `recordDepositFor` pattern. It creates the potential for vault balances to get out of sync and also adds some attack surface to the contracts. After Istanbul it also doesn't save _that_ much gas. Just making the deposits every time would add an additional ~1600 gas after [EIP-2200](https://github.com/ethereum/EIPs/pull/2200) right?

Even if it is a bit more expensive, I think the system clearly has higher security if we remove this pattern.

contracts/staking/contracts/src/sys/MixinParams.sol
dorothy-zbornak: The reward calculations explode if we allow a pool with stake 0 to get in the mix. Actually, this check should be `< 2` since delegator rewards are weighed less. I can special case these pools, but letting a pool with 0 stake collect rewards seems wrong to me?

contracts/staking/contracts/src/stake/MixinStake.sol
dorothy-zbornak: I just manually formatted  everything I touched to 80 characters. :smile:

contracts/staking/contracts/src/libs/LibCobbDouglas.sol
dorothy-zbornak: woops! :+1:

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -82,182 +137,268 @@ contract MixinStakingPoolRewards is
     )
         internal
     {
-        // transfer any rewards from the transient pool vault to the eth vault;
-        // this must be done before we can modify the owner's portion of the delegator pool.
+        // Rransfer any rewards from the transient pool vault to the eth vault;
```
hysz: * Transfer

contracts/staking/contracts/src/sys/MixinFinalizer.sol
```diff
@@ -0,0 +1,404 @@
...
+    /// @dev Finalizes pools that were active in the previous epoch, paying out
+    ///      rewards to the reward and eth vault. Keepers should call this
+    ///      function repeatedly until all active pools that were emitted in in
+    ///      a `StakingPoolActivated` in the prior epoch have been finalized.
+    ///      Pools that have already been finalized will be silently ignored.
+    ///      We deliberately try not to revert here in case multiple parties
+    ///      are finalizing pools.
+    /// @param poolIds List of active pool IDs to finalize.
+    /// @return _unfinalizedPoolsRemaining The number of unfinalized pools left.
+    function finalizePools(bytes32[] calldata poolIds)
```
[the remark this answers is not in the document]
abandeali1: +1

contracts/staking/contracts/src/sys/MixinFinalizer.sol
```diff
@@ -0,0 +1,404 @@
...
+    ///      epochs at once.
+    /// @return activePools The pools that were active in `epoch`.
+    function _getActivePoolsFromEpoch(
+        uint256 epoch
+    )
+        internal
+        view
+        returns (mapping (bytes32 => IStructs.ActivePool) storage activePools)
+    {
+        activePools = _activePoolsByEpoch[epoch % 2];
```
abandeali1: This is a pretty interesting approach!

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -26,28 +26,45 @@ import "../libs/LibStakingRichErrors.sol";
...
+    MixinStakeBalances,
+    MixinCumulativeRewards,
+    MixinStakingPoolRewards,
     MixinStakingPool
 {
     using LibSafeMath for uint256;
 
     /// @dev Pays a protocol fee in ETH or WETH.
-    ///      Only a known 0x exchange can call this method. See (MixinExchangeManager).
+    ///      Only a known 0x exchange can call this method. See
```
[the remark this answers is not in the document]
dorothy-zbornak: Human setting.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -62,21 +79,10 @@ contract MixinExchangeFees is
         payable
```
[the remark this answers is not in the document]
dorothy-zbornak: Ah, yeah, these are relics from before we had the WETH unwrapping in place.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -62,21 +79,10 @@ contract MixinExchangeFees is
...
-                    LibStakingRichErrors.ProtocolFeePaymentErrorCodes.ZeroProtocolFeePaid :
-                    LibStakingRichErrors.ProtocolFeePaymentErrorCodes.MismatchedFeeAndPayment,
-                protocolFeePaid,
-                msg.value
-            ));
-        }
+        _assertValidProtocolFee(protocolFeePaid);
 
-        // Transfer the protocol fee to this address if it should be paid in WETH.
+        // Transfer the protocol fee to this address if it should be paid in
```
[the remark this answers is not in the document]
dorothy-zbornak: I would hesitate to add an event that could potentially be triggered on every fill. I think we can actually infer this from looking at WETH `Transfer` events to the exchange, right?

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -88,271 +94,138 @@ contract MixinExchangeFees is
...
-        uint256 finalContractBalance) = _distributeFeesAmongMakerPools();
-        emit RewardsPaid(
-            totalActivePools,
-            totalFeesCollected,
-            totalWeightedStake,
-            totalRewardsPaid,
-            initialContractBalance,
-            finalContractBalance
+        // Look up the pool for this epoch.
+        uint256 currentEpoch = currentEpoch;
```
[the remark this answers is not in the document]
dorothy-zbornak: I'm surprised this even worked.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -88,271 +94,138 @@ contract MixinExchangeFees is
...
-        return address(this).balance.safeAdd(wethBalance);
+        _totalFeesCollectedThisEpoch = totalFeesCollectedThisEpoch;
     }
 
-    /// @dev Withdraws the entire WETH balance of the contract.
-    function _unwrapWETH()
-        internal
+    /// @dev Returns the total balance of this contract, including WETH.
+    /// @return totalBalance Total balance.
+    function getTotalBalance()
```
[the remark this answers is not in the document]
dorothy-zbornak: As in, add it to `totalFeesCollected`? idk, aren't they two different things? The contract balance could include subsidies, remaining rewards, etc.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
dorothy-zbornak: No, it's normal operation to clear CRs from past epochs. We just special case the current epoch because the system breaks if we don't have it.

dorothy-zbornak: Comment should be updated.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
dorothy-zbornak: If the code is written correctly, none of these conditions should ever be true. The thing is, it's super easy to make changes to the accounting that accidentally trips one, which is why I think it's valuable to have assertions in place. These gave me a ton of insight when I was working on this PR. But since these are essentially just for our benefit, maybe we should replace them with plain string reverts?

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
dorothy-zbornak: Yeah, I liked the conciseness originally but I agree it's better overall to be explicit.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -82,182 +137,268 @@ contract MixinStakingPoolRewards is
...
         );
     }
 
-    /// @dev Transfers a delegators accumulated rewards from the transient pool Reward Pool vault
-    ///      to the Eth Vault. This is required before the member's stake in the pool can be
-    ///      modified.
+    /// @dev Deposit rewards into the eth vault and reward vault for pool
+    ///      operators and members rewards, respectively. This should be called
+    ///      in tandem with `_recordStakingPoolRewards()`. We separate them
+    ///      so we can bath deposits, because ETH transfers are expensive.
```
dorothy-zbornak:
```suggestion
    ///      so we can batch deposits, because ETH transfers are expensive.
```

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -82,182 +137,268 @@ contract MixinStakingPoolRewards is
...
+        uint256 lastRewardEpoch = _cumulativeRewardsByPoolLastStored[poolId];
+
+        // If the stake has been touched since the last reward epoch,
+        // it has already been claimed.
+        if (unsyncedStake.currentEpoch >= lastRewardEpoch) {
+            return reward;
+        }
+        // From here we know: `unsyncedStake.currentEpoch < currentEpoch > 0`.
+
+        if (unsyncedStake.currentEpoch >= lastRewardEpoch) {
```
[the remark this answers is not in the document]
dorothy-zbornak: been eating too much copy pasta

contracts/staking/contracts/src/sys/MixinAbstract.sol
```diff
@@ -0,0 +1,85 @@
...
+
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "../interfaces/IStructs.sol";
+
+
+/// @dev Exposes some internal functions from various contracts to avoid
+///      cyclical dependencies.
+contract MixinAbstract {
```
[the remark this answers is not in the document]
dorothy-zbornak: idk how to fix the cyclical deps without introducing something like this or an interface, and we don't use the interface convention for mixins in this package so I went with this. :thinking:

contracts/staking/contracts/src/sys/MixinFinalizer.sol
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        // Unrwap any WETH protocol fees.
+        _unwrapWETH();
+
+        // Populate finalization state.
+        unfinalizedPoolsRemaining =
+            _unfinalizedPoolsRemaining = numActivePoolsThisEpoch;
+        unfinalizedRewardsAvailable = address(this).balance;
+        unfinalizedTotalFeesCollected = totalFeesCollectedThisEpoch;
```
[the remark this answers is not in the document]
dorothy-zbornak: They transfer into `unfinalizedTotalFeesCollected` and `unfinalizedTotalWeightedStake`, which are needed by the cobb douglas function, so we need them both. It's worth nothing that `totalWeightedStakeThisEpoch` only gets updated the *first* time a pool is credit/activated on a fill. Having both of these known saves us from needing a prefinalize step, which `endEpoch()` kind of is but with constant complexity.

contracts/staking/contracts/src/sys/MixinFinalizer.sol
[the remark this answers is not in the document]
dorothy-zbornak: If we ditch batch transfers then this is def the way to go.

contracts/staking/test/rewards_test.ts
```diff
@@ -663,6 +592,115 @@ blockchainTests.resets('Testing Rewards', env => {
...
+            await validateEndBalances({
+                stakerRewardVaultBalance_1: toBaseUnitAmount(0),
+                stakerEthVaultBalance_1: reward,
+            });
+        });
+        it(`should split payout between two delegators when undelegating`, async () => {
+            const stakeAmounts = [toBaseUnitAmount(5), toBaseUnitAmount(10)];
+            const totalStakeAmount = BigNumber.sum(...stakeAmounts);
+            // stake and delegate both
+            const stakersAndStake = _.zip(stakers.slice(0, 2), stakeAmounts) as Array<[StakerActor, BigNumber]>;
```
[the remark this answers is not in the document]
dorothy-zbornak: Ah, yeah. I introduced that thing after suffering through all these but forgot to retroactively switch.

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -103,6 +143,25 @@ export class StakingApiWrapper {
                 await this.stakingContract.getParams.callAsync(),
             ) as any) as StakingParams;
         },
+
+        cobbDouglas: async (
```
dorothy-zbornak:
```suggestion
        cobbDouglasAsync: async (
```

contracts/utils/contracts/src/LibFractions.sol
```diff
@@ -27,12 +30,25 @@ library LibFractions {
...
+            return (numerator = n1, denominator = d1);
+        }
         numerator = n1
             .safeMul(d2)
             .safeAdd(n2.safeMul(d1));
-
         denominator = d1.safeMul(d2);
-        return (numerator, denominator);
+
+        // If either the numerator or the denominator are > RESCALE_THRESHOLD,
```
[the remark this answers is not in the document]
dorothy-zbornak: Hmm. Could also move it out to a separate `rescaleFractions(threshold)` function. We should probably also drop all the `*Fractions` suffixes to these functions since it's stuttering.

contracts/staking/test/cumulative_reward_tracking_test.ts
```diff
@@ -34,371 +34,511 @@ blockchainTests.resets('Cumulative Reward Tracking', env => {
...
             await simulation.runTestAsync(
                 [
-                    TestAction.CreatePool, // creates CR in epoch 0
-                    TestAction.Delegate, // does nothing wrt CR, as there is alread a CR set for this epoch.
+                    // Creates CR for epoch 0
+                    TestAction.CreatePool,
                 ],
                 [
-                    TestAction.Delegate, // does nothing wrt CR, as there is alread a CR set for this epoch.
+                    // Updates CR for epoch 0
```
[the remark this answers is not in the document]
dorothy-zbornak: All credit to @hysz since I just updated these for the tweaked mechanics.

contracts/staking/test/actors/finalizer_actor.ts
```diff
@@ -187,8 +216,9 @@ export class FinalizerActor extends BaseActor {
     private async _getOperatorShareByPoolIdAsync(poolIds: string[]): Promise<OperatorShareByPoolId> {
         const operatorShareByPoolId: OperatorShareByPoolId = {};
         for (const poolId of poolIds) {
-            const pool = await this._stakingApiWrapper.stakingContract.getStakingPool.callAsync(poolId);
-            operatorShareByPoolId[poolId] = new BigNumber(pool.operatorShare);
+            operatorShareByPoolId[poolId] = new BigNumber(
```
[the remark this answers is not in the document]
dorothy-zbornak: Turns out this was actually less legible with `_.zipObject` because it needs to extract the `operatorShare` from the call result, so `_.zipObject(poolIds, (await Promise.all(pooIds.map(id => getPoolAsync(id))).map(p => p.operatorShare))`

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
abandeali1: I'd be fine with that personally.

contracts/staking/contracts/src/sys/MixinParams.sol
abandeali1: Sounds good. Let's move this assertion into `StakingProxy._assertValidStorageParams` now that #2166 is merged.

contracts/staking/contracts/src/sys/MixinFinalizer.sol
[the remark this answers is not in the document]
abandeali1: Ah, I see that now.

contracts/staking/test/unit_tests/delegator_reward_test.ts
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    function getDepositsFromLogs(logs: LogEntry[], poolId: string, delegator?: string): [BigNumber, BigNumber] {
+        let ethVaultDeposit = constants.ZERO_AMOUNT;
+        let rewardVaultDeposit = constants.ZERO_AMOUNT;
+        const ethVaultDepositArgs = filterLogsToArguments<EthVaultDepositEventArgs>(
+            logs,
+            TestDelegatorRewardsEvents.RecordDepositToEthVault,
+        );
+        if (ethVaultDepositArgs.length > 0 && delegator !== undefined) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Hmm, why? Lots of things don't trigger an EthVault deposit, right?

contracts/staking/test/rewards_test.ts
```diff
@@ -663,6 +592,115 @@ blockchainTests.resets('Testing Rewards', env => {
...
+            const expectedStakerRewards = stakeAmounts.map(n => reward.times(n).dividedToIntegerBy(totalStakeAmount));
+            await validateEndBalances({
+                stakerRewardVaultBalance_1: expectedStakerRewards[0],
+                stakerRewardVaultBalance_2: expectedStakerRewards[1],
+                stakerEthVaultBalance_1: toBaseUnitAmount(0),
+                stakerEthVaultBalance_2: toBaseUnitAmount(0),
+                poolRewardVaultBalance: reward,
+                membersRewardVaultBalance: reward,
+            });
+            const undelegateZeroAsync = async (staker: StakerActor) => {
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, I agree. I think these should just use `syncDelegatorRewards()` now that that's a thing.

contracts/staking/test/unit_tests/delegator_reward_test.ts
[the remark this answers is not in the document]
mzhu25: o I meant if `ethVaultDeposityArgs.length == 0` then the for loop will noop and if `delegator === undefined` then the inner if will fail

contracts/staking/test/unit_tests/finalizer_test.ts
```diff
@@ -0,0 +1,569 @@
...
+            const pool1 = await addActivePoolAsync();
+            const pool2 = await addActivePoolAsync(_.omit(pool1, 'poolId'));
+            const receipt = await testContract.endEpoch.awaitTransactionSuccessAsync();
+            const { rewardsAvailable } = getEpochEndedEvents(receipt.logs)[0];
+            const logs = await finalizePoolsAsync([pool1, pool2].map(r => r.poolId));
+            const { rewardsPaid } = getEpochFinalizedEvents(logs)[0];
+            expect(rewardsPaid).to.bignumber.lte(rewardsAvailable);
+        });
+
+        blockchainTests.optional('`rewardsPaid` fuzzing', async () => {
```
jalextowle: Very nice!

contracts/staking/test/unit_tests/finalizer_test.ts
```diff
@@ -0,0 +1,569 @@
...
+        it('computes a reward with 0% operatorShare', async () => {
+            const pool = await addActivePoolAsync({ operatorShare: 0 });
+            await testContract.endEpoch.awaitTransactionSuccessAsync();
+            return assertUnfinalizedPoolRewardsAsync(pool.poolId, {
+                totalReward: INITIAL_BALANCE,
+                membersStake: pool.membersStake,
+            });
+        });
+
+        it('computes a reward with 100% operatorShare', async () => {
```
jalextowle: Optional: It would be nice to have a test for `0% < operatorShare < 100%`

contracts/staking/test/unit_tests/delegator_reward_test.ts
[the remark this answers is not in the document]
dorothy-zbornak: ahhh, duh :+1:

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -82,182 +138,256 @@ contract MixinStakingPoolRewards is
...
+        // Get the last epoch where a reward was credited to this pool, which
+        // also happens to be when we last created a cumulative reward entry.
+        uint256 lastRewardEpoch = _cumulativeRewardsByPoolLastStored[poolId];
+
+        // If the stake has been touched since the last reward epoch,
+        // it has already been claimed.
+        if (unsyncedStake.currentEpoch >= lastRewardEpoch) {
+            return reward;
+        }
+        // From here we know: `unsyncedStake.currentEpoch < currentEpoch > 0`.
```
hysz: It looks like the intervals have been shifted. Is this on purpose?
```
# Previously the ranges were:
[unsyncedStake.currentEpoch-1 .. unsyncedStake.currentEpoch]
[unsyncedStake.currentEpoch .. lastRewardEpoch]

# Now the ranges are:
[unsyncedStake.currentEpoch .. unsyncedStake.currentEpoch + 1]
[unsyncedStake.currentEpoch + 1 .. lastRewardEpoch + 1]
```

dorothy-zbornak: Yup. This is from crediting rewards at the top of an epoch rather than at the end, and folding in unfinalized rewards.

contracts/staking/contracts/src/vaults/EthVault.sol
```diff
@@ -83,21 +91,19 @@ contract EthVault is
...
+    /// @dev Withdraw an `amount` of WETH to `owner` from the vault.
+    /// @param owner of WETH.
+    /// @param amount of WETH to withdraw.
     function _withdrawFrom(address payable owner, uint256 amount)
         internal
     {
-        // update balance
-        // note that this call will revert if trying to withdraw more
-        // than the current balance
+        //Uupdate balance.
```
mzhu25:
```suggestion
        // Update balance.
```

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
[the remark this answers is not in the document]
dorothy-zbornak: I could just remove this function. All it's doing is suming ETH and WETH balance, and we could use the space.

contracts/staking/test/utils/cumulative_reward_tracking_simulation.ts
```diff
@@ -117,20 +118,21 @@ export class CumulativeRewardTrackingSimulation {
...
+            let receipt: TransactionReceiptWithDecodedLogs | undefined;
+            let logs = [] as DecodedLogs;
             switch (action) {
                 case TestAction.Finalize:
-                    txReceipt = await this._stakingApiWrapper.utils.skipToNextEpochAsync();
+                    logs = await this._stakingApiWrapper.utils.skipToNextEpochAndFinalizeAsync();
                     break;
 
                 case TestAction.Delegate:
                     await this._stakingApiWrapper.stakingContract.stake.sendTransactionAsync(this._amountToStake, {
```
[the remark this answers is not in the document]
dorothy-zbornak: probably a typo

contracts/staking/contracts/src/vaults/EthVault.sol
dorothy-zbornak: Not touching the vaults anymore since they're getting removed in a coming PR. :man_shrugging:

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -88,206 +83,116 @@ contract MixinExchangeFees is
...
-        if (wethBalance != 0) {
-            IEtherToken(WETH_ADDRESS).withdraw(wethBalance);
-        }
+        uint256 operatorStake = getStakeDelegatedToPoolByOwner(
+            _poolById[poolId].operator,
+            poolId
+        ).currentEpochBalance;
+        membersStake = totalStake.safeSub(operatorStake);
+        weightedStake = operatorStake.safeAdd(
+            LibMath.getPartialAmountFloor(
```
hysz: The `weightedStake` should also include the operator's stake. See https://github.com/0xProject/0x-monorepo/blob/3.0/contracts/staking/contracts/src/fees/MixinExchangeFees.sol#L214

PR #2156, opened 2019-09-14
Sync Rewards + Refactored Reward Vault

hysz:
## Description

Reward Vault now only stores ETH for the delegators. The operator's reward goes directly to the ETH Vault. The reward vault has been stripped down and made much more lean. All pool logic has been moved to the staking contract.

## Checklist:


contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -0,0 +1,217 @@
+/*
```
hysz: This was all existing code that was refactored out of `MixinStakingPool.sol`

contracts/staking/contracts/src/staking_pools/MixinStakingPoolModifiers.sol
```diff
@@ -0,0 +1,81 @@
+/*
```
hysz: The code in this file already existed but was refactored into its own Mixin.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -104,14 +127,35 @@ contract MixinStakingPoolRewards is
...
+    function _handleStakingPoolReward(
         bytes32 poolId,
         uint256 reward,
         uint256 amountOfDelegatedStake,
         uint256 epoch
     )
         internal
     {
+
+        IStructs.Pool memory pool = poolById[poolId];
```
hysz: This logic was copied from the Staking Pool Vault.

contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -237,7 +237,7 @@ contract MixinExchangeFees is
 
             // compute weighted stake
             uint256 totalStakeDelegatedToPool = getTotalStakeDelegatedToPool(poolId).currentEpochBalance;
-            uint256 stakeHeldByPoolOperator = getStakeDelegatedToPoolByOwner(rewardVault.operatorOf(poolId), poolId).currentEpochBalance;
+            uint256 stakeHeldByPoolOperator = getStakeDelegatedToPoolByOwner(poolById[poolId].operator, poolId).currentEpochBalance;
```
dorothy-zbornak: :tada:

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -77,265 +57,55 @@ contract StakingPoolRewardVault is
...
 
     /// @dev Withdraw some amount in ETH of a pool member.
-    /// Note that this is only callable by the staking contract, and when
-    /// not in catastrophic failure mode.
+    /// Note that this is only callable by the staking contract.
     /// @param poolId Unique Id of pool.
     /// @param amount Amount in ETH to transfer.
-    function transferMemberBalanceToEthVault(
+    function transferToEthVault(
         bytes32 poolId,
```
dorothy-zbornak: What do we think about just passing in the `ethVault` address here? Then we wouldn't have to set up `_ethVault`, etc.

contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -95,4 +95,16 @@ interface IStructs {
         uint256 cumulativeRewardEpoch;
         IStructs.Fraction cumulativeReward;
     }
+
+    /// @dev Holds the balances and other data for a staking pool.
```
mzhu25:
```suggestion
    /// @dev Holds metadata for a staking pool.
```

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -77,202 +56,74 @@ contract MixinStakingPool is
...
-        IStructs.MakerPoolJoinStatus memory poolJoinStatus = IStructs.MakerPoolJoinStatus({
-            poolId: poolId,
-            confirmed: false
-        });
-        poolJoinedByMakerAddress[makerAddress] = poolJoinStatus;
-
-        // Maker has joined to the pool, awaiting operator confirmation
-        emit PendingAddMakerToPool(
+        // load pool and assert that we can decrease
+        IStructs.Pool memory pool = poolById[poolId];
```
mzhu25: nit: we only need `poolById[poolId].operatorShare`

contracts/staking/contracts/src/interfaces/IStakingEvents.sol
```diff
@@ -134,4 +134,24 @@ interface IStakingEvents {
     event StakingPoolRewardVaultChanged(
         address rewardVaultAddress
     );
+
+    /// @dev Emitted when a staking pool's operator share is decreased.
+    /// @param poolId Unique Id of pool.
+    /// @param oldOperatorShare Previous share of rewards owned by operator.
+    /// @param newOperatorShare Newly decreased share of rewards owned by operator.
+    event OperatorShareDecreased(
+        bytes32 poolId,
```
abandeali1: Let's index `poolId` in any events where it is used.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -104,14 +127,39 @@ contract MixinStakingPoolRewards is
     /// @param reward to record for delegators.
     /// @param amountOfDelegatedStake the amount of delegated stake that will split this reward.
     /// @param epoch at which this was earned.
-    function _recordRewardForDelegators(
+    function _handleStakingPoolReward(
```
abandeali1: Do the comments for this function need to be tweaked?

contracts/staking/contracts/src/interfaces/IStakingPoolRewardVault.sol
```diff
@@ -20,189 +20,62 @@ pragma solidity ^0.5.9;
...
-    /// @param poolId Unique Id of pool that was registered.
-    /// @param oldOperatorShare Previous share of rewards owned by operator.
-    /// @param newOperatorShare Newly decreased share of rewards owned by operator.
-    event OperatorShareDecreased(
-        bytes32 poolId,
-        uint32 oldOperatorShare,
-        uint32 newOperatorShare
+    /// @dev Emitted when the eth vault is changed
+    /// @param newEthVault address of new Eth vault.
+    event EthVaultChanged(
```
abandeali1: Our convention elsewhere is `EthVaultSet`

contracts/staking/contracts/src/interfaces/IStakingPoolRewardVault.sol
```diff
@@ -20,189 +20,62 @@ pragma solidity ^0.5.9;
...
-        bytes32 poolId,
-        uint256 amount
-    );
-
-    /// @dev Emitted when a reward is withdrawn for a pool member.
-    /// @param amount The amount in ETH withdrawn.
-    /// @param poolId The pool the reward was deposited for.
-    event RewardWithdrawnForMember(
+    event PoolRewardTransferredToEthVault(
         bytes32 poolId,
```
abandeali1: Let's index both `poolId` and `member`.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -0,0 +1,217 @@
...
+    MixinScheduler,
+    MixinStakeStorage,
+    MixinStakeBalances,
+    MixinStakingPoolRewards,
+    MixinStakingPoolModifiers
+{
+
+    using LibSafeMath for uint256;
+
+    function joinStakingPoolAsMaker(
```
abandeali1: No devdoc comment here?

contracts/staking/contracts/src/staking_pools/MixinStakingPoolModifiers.sol
```diff
@@ -0,0 +1,81 @@
...
+    MixinStakeBalances,
+    MixinStakingPoolRewards
+{
+
+    using LibSafeMath for uint256;
+
+    /// @dev Asserts that the sender is the operator of the input pool.
+    /// @param poolId Pool sender must be operator of.
+    modifier onlyStakingPoolOperator(bytes32 poolId) {
+        address poolOperator = poolById[poolId].operator;
```
abandeali1: Looks like we use `poolOperator`, `operator`, and `operatorAddress` in various places throughout these contracts.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -43,6 +46,26 @@ contract MixinStakingPoolRewards is
 
     using LibSafeMath for uint256;
 
+    /// @dev Syncs rewards for a delegator. This includes transferring rewards from
+    /// the Reward Vault to the Eth Vault, and adding/removing dependencies on cumulative rewards.
+    /// @param poolId Unique id of pool.
+    function syncDelegatorRewards(bytes32 poolId)
```
abandeali1: Under what circumstances is this called? Who is expected to call it? It feels strange to have an external function callable by anyone that modifies state.

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
```diff
@@ -48,24 +39,13 @@ contract StakingPoolRewardVault is
 {
 
     using LibSafeMath for uint256;
-    using LibSafeDowncast for uint256;
 
     // mapping from poolId to Pool metadata
-    mapping (bytes32 => Pool) internal poolById;
+    mapping (bytes32 => uint256) internal balanceByPoolId;
```
abandeali1: Nit: prefix with `_`

contracts/staking/contracts/src/vaults/StakingPoolRewardVault.sol
abandeali1: I like it a lot. Then they can't get out of sync either.

contracts/staking/contracts/src/interfaces/IStakingEvents.sol
```diff
@@ -134,4 +134,24 @@ interface IStakingEvents {
...
+        uint32 oldOperatorShare,
+        uint32 newOperatorShare
+    );
+
+    /// @dev Emitted when an operator reward is transferred to the Eth vault.
+    /// @param amount The amount in ETH withdrawn.
+    /// @param operator of the pool.
+    /// @param poolId The pool the reward was deposited for.
+    event OperatorRewardTransferredToEthVault(
```
abandeali1: Do we need this event? We already log all of this information on any deposit into the EthVault.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
hysz: Jeez, thx for catching that. Must've gotten truncated in the refactor or rebase. I'll add it back.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -0,0 +1,217 @@
...
+
+
+/// @dev This mixin contains logic for staking pools.
+contract MixinStakingPoolMakers is
+    IStakingEvents,
+    MixinConstants,
+    Ownable,
+    MixinStorage,
+    MixinZrxVault,
+    MixinStakingPoolRewardVault,
```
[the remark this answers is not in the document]
hysz: These will get added back every time we run the linearize script. I'll remove them here but we'll probably have to do this again after MBF is merged.

contracts/staking/contracts/src/staking_pools/MixinEthVault.sol
```diff
@@ -50,4 +50,32 @@ contract MixinEthVault is
     {
         return address(ethVault);
     }
+
+    /// @dev Transfers operator reward to the ETH vault.
+    /// @param poolId Unique Id of pool to transfer reward for,
+    /// @param operator of the pool.
+    /// @param amount of ETH to transfer.
+    function _transferOperatorRewardToEthVault(
```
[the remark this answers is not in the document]
hysz: RIP

contracts/staking/contracts/src/interfaces/IStakingEvents.sol
hysz: Yeah, I see what you mean. Basically, the goal was to document the flow of value as it enters and leaves a contract. That said, we can probably nix now that the vaults are set during initialization (and they also emit both ingress/egress value flow).

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
hysz: This is used by a delegator when they want to sync their rewards without delegating/undelegating. It's effectively the same as delegating zero stake. 

Note also that this uses `msg.sender` and can't force a sync for an arbitrary delegator.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -77,202 +56,74 @@ contract MixinStakingPool is
         returns (bytes32 poolId)
     {
         // note that an operator must be payable
-        address payable operatorAddress = msg.sender;
+        address payable operator = msg.sender;
```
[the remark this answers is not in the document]
hysz: Note: we chatted offline. Dropped the `Address` suffix and updated instances of `operatorAddress` and `poolOperator` to `operator`.

contracts/staking/contracts/src/interfaces/IStakingEvents.sol
hysz: (Removed)

contracts/staking/contracts/src/staking_pools/MixinStakingPoolModifiers.sol
hysz: Note: we chatted offline. Dropped the Address suffix and updated instances of operatorAddress and poolOperator to operator.

contracts/staking/contracts/src/interfaces/IStakingPoolRewardVault.sol
hysz: This event has been dropped.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -64,191 +44,83 @@ contract MixinStakingPool is
...
         );
     }
 
-    /// @dev Returns the pool id of the input maker.
-    /// @param makerAddress Address of maker
-    /// @return Pool id, nil if maker is not yet assigned to a pool.
-    function getStakingPoolIdOfMaker(address makerAddress)
+    /// @dev Returns the unique id that will be assigned to the next pool that is created.
+    /// @return Pool id.
+    function getNextStakingPoolId()
```
abandeali1: This function can be removed since `nextPoolId` is now public (reminder to remove from any interfaces as well).

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -0,0 +1,210 @@
...
+        view
+        returns (bool)
+    {
+        return poolJoinedByMakerAddress[makerAddress].confirmed;
+    }
+
+    /// @dev Returns the current number of makers in a given pool.
+    /// @param poolId Unique id of pool.
+    /// @return Size of pool.
+    function getNumberOfMakersInStakingPool(bytes32 poolId)
```
abandeali1: I'm kind of meh on having all of these getters. The rest of them are actually pretty frequently used so I'm fine leaving them in, but we should remove this one. When we're using getters for individual properties, it is easy to incur extra costs from sloads (example below).

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -0,0 +1,210 @@
...
+        if (makerPendingPoolId != poolId && makerAddress != poolById[poolId].operator) {
+            LibRichErrors.rrevert(LibStakingRichErrors.MakerPoolAssignmentError(
+                LibStakingRichErrors.MakerPoolAssignmentErrorCodes.MakerAddressNotPendingAdd,
+                makerAddress,
+                makerPendingPoolId
+            ));
+        }
+
+        // Is the pool already full?
+        if (getNumberOfMakersInStakingPool(poolId) == maximumMakersInPool) {
```
abandeali1: We already reference `poolById[poolId]` above, so this call is using an extra sload. We can load the pool in memory and reference that instead.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolModifiers.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /// @dev Asserts that the sender is the operator of the input pool or the input maker.
+    /// @param poolId Pool sender must be operator of.
+    /// @param makerAddress Address of a maker in the pool.
+    modifier onlyStakingPoolOperatorOrMaker(bytes32 poolId, address makerAddress) {
+        address operator;
+        if (
+            msg.sender != makerAddress &&
+            msg.sender != (operator = poolById[poolId].operator)
```
abandeali1: This pattern seems a bit strange to me. Why not just set `operator` when we declare it above?

dorothy-zbornak: I like this actually. Short-circuits the `sload` if the first condition fails. I think in practice makers will be passive accounts so we'll pretty much always hit the second condition, though, so maybe it's better to prioritize readability here.

hysz: This code isn't going to run very often, and I agree this will almost always be called by the operator. I think readability is more important here, although I wouldn't be opposed to using this pattern in code that gets hit frequently.

PR #2157, opened 2019-09-14
Refactored asset-swapper smart logic and renamed options

dave4506:
## Description

In favor of making `asset-swapper` logic predictable, this PR removes the smart routing logic in favor of explicit behavior based on options passed to the `SwapQuoteConsumer`. 

`useConsumerType` is renamed `useExtensionContract` in favor of clarity of the option. 

In hopes to make the expected behavior of the consumer explicit, the smart logic is moved into a subclass of `SwapQuoteConsumer`. The "smart" routing logic is now housed in a new Consumer called `SmartSwapQuoteConsumer`.

## Checklist:


dekz:
Having this logic exposed is pretty useful, for example in Launch Kit I'd rather not have the implementor figure this out and take into account all of the cases. 

Should we just remove the smart behaviour but still export the function to determine what is possible to use?

dave4506:
@dekz I agree that the logic should still be exposed at some capacity; will update the PR to add that functionality in as a util/convenience functionality.

packages/asset-swapper/src/types.ts
```diff
@@ -86,9 +87,9 @@ export interface ExchangeMarketSellSmartContractParams extends SmartContractPara
 /**
  * Represents the varying smart contracts that can consume a valid swap quote
  */
-export enum ConsumerType {
+export enum ExtensionContractType {
     Forwarder = 'FORWARDER',
-    Exchange = 'EXCHANGE',
+    None = 'NONE',
```
fragosti: I feel like the addition of another class is a bit over-kill. Would one of these approaches work?
- Expose a utility method on the consumer to help you decide.
- Have `Automatic` be a `ExtensionContractType ` enum member

dave4506: I agree with the sentiment of overkill, but in my mind it is the least worst of a few options. Here is my rationale for creating a separate class over your suggested options: 

- Expose a utility method on the consumer to help you decide.
While we can expose utility methods that can, say return `ExtensionContractType`, I think experientially, it adds more steps to the process of integrating 0x CFL, and puts the burden on the developers to using the utility method to conduct the check, and to that, the check is `async` and requires a `provider` which may increase the burden for developers. 
- Have Automatic be a `ExtensionContractType` enum member
With this, I do not think `ExtensionContractType` naming scheme accurately reflects the cases available in the enum, in my mind for functions like `swapQuoteConsumerUtils.getExtensionContractTypeForSwapQuoteAsync`, I find it would it be confusing for the type returned to possibly be `Automatic`. To solve this, we can expand the types/options naming, but I think that can get bloated on both maintaining the library and for developers

I have considered to expand the `SwapQuoteConsumer` class to have `smart` variants of the standard functions; that seemed unscalable and bloated `SwapQuoteConsumer`. 

With this in mind, I thought the best decision was to create a new `SmartSwapQuoteConsumer` class that subclasses `SwapQuoteConsumer`; to me, developers can explicit understand what behavior they are expecting from this consumer, and it doesn't bloat options, types as much, and add more steps to the CFL integration process for developers. 

Open to other ideas, but this is my general rationale with creating another class

fragosti: Taking a step back, it's important to think about how this is going to be used. In a CFL use-case, you know exactly what extension contract you want to use and having a smart method doesn't make a lot of sense because you're passing call-data into your smart contract that is likely hard-coded to the 0x exchange. Do you know if anyone is passing in the contract address as well?

So the only use-case I can imagine is the one @dekz is outlining where we want to smart-fill based on our balances, as we do in launch-kit. In launch-kit, the user is told whether they will have to wrap their ETH or not, and there are different flows that depend on their balances. So to support the launch-kit use-case we need to expose the utility method even if it is async.

dave4506: To your point, I agree that the expected behavior for CFL is for the integrators to know before hand the extension contract or sans extension setup they intend to use for their contracts. Because of that, I thought it would be clearer for `SwapQuoteConsumer` to not have any smart behavior. 

For the unique cases that the smart behavior is needed, I thought a subclass of `SwapQuoteConsumer` will provide the most intuitive interface for developers who have previously already used `asset-swapper`. If we are to expose the smart functionality as a utils, it would needed the same params as `swapQuoteConsumerUtils.getExtensionContractTypeForSwapQuoteAsync`:
```
        quote: SwapQuote,
        contractWrappers: ContractWrappers,
        provider: Provider,
        opts: Partial<SwapQuoteGetOutputOpts>
```
This would put the burden on the developers to instantiate a contractWrapper, provider, etc. I am open to creating a `Utils` class that developers can instantiate, but at that point, I feel like a subclass of `SwapQuoteConsumer` makes more sense. 

With this in mind, I imagine CFL integrators will ONLY ever use `SwapQuoteConsumer` and will never touch `SmartSwapQuoteConsumer`, that class will exist for unique implementations like launch-kit, and maybe in the future, instant. In CFL docs, I imagine `SmartSwapQuoteConsumer` will never be brought up as a feature.

I am in favor of subclassing over utils for the reasoning above.

fragosti: I'm just saying that if you can't see what the consumer is about to do, then the smart logic is not useful.

dave4506: Right, that makes a ton of sense; that blew over my head. With that, I think the utils makes more sense.

PR #2158, opened 2019-09-15
Change state var visibilities, delete redundant code

abandeali1:
## Description

This PR:

- Replaces #2112 
- Deletes redundant imports and inherited contracts
- Makes many state variables public and removes corresponding getters
    - Note: The getters that are automatically generated for public variables cannot be inherited in any interfaces, so certain getters were left in (such as `balanceOf`).
- Fixes `getTotalProtocolFeesThisEpoch` to include the contract's WETH balance (this is the only functional change in this PR).
- Adds an `IStorage` interface. This is currently unused and cannot be inherited by `MixinStorage` (as it would override the public variables), but it may be useful for external devs.

## Checklist:


contracts/staking/contracts/src/stake/MixinStakeBalances.sol
```diff
@@ -57,11 +48,11 @@ contract MixinStakeBalances is
     /// @param owner of stake.
     /// @return Active stake for owner.
     function getActiveStake(address owner)
-        public
+        external
```
mzhu25: Do we need these getters (`getActiveStake, getInactiveStake, getStakeDelegatedByOwner`)? We don't seem to be using them internally, and while doing a `_loadAndSyncBalance` can be nice, you can get the same information by just looking at the unsynced balance and `currentEpoch`
Optional for this PR, happy to address this in 2151 if we do decide to do this

mzhu25: actually after offline discussion with Greg, I think we want to keep these for convenience

abandeali1: Yeah, I like keeping these since they load from multiple storage slots in a single call.

contracts/staking/contracts/src/stake/MixinStakeStorage.sol
```diff
@@ -93,9 +85,9 @@ contract MixinStakeStorage is
         // load from storage
         balance = balancePtr;
         // sync
-        uint256 currentEpoch = getCurrentEpoch();
-        if (currentEpoch > balance.currentEpoch) {
-            balance.currentEpoch = currentEpoch.downcastToUint32();
+        uint256 currentEpoch_ = currentEpoch;
```
dorothy-zbornak: Oh, cool, I've been wanting to use this convention for local conflicts, but I wasn't sure other people were on board.

PR #2161, opened 2019-09-16
DevUtils.sol: Upgrade for v3

feuGeneA:
## Description

This contract needs to be upgraded to work with v3 and deployed for use by Mesh.

## Testing instructions

Run tests for `0x-monorepo/contracts/dev-utils` (or just watch CI).

## Types of changes

## Checklist:

-   [x] Make DevUtils deployable.  The `DevUtils` contract can’t currently be deployed because the contract size is too large. We can experiment with turning off compiler optimizations, which might lower the codesize. Otherwise we’ll have to break up the deployments into multiple contracts (@abandeali1 thinks the `OrderValidationUtils` and `OrderTransferSimulationUtils` contracts are most important).
-   [x] More tests for `OrderTransferSimulationUtils` (there was previously a fairly large bug that wasn’t caught in tests, which you can see by looking at the diff)

### DIFF
--- a/contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
+++ b/contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
@@ -24,6 +24,7 @@ import "@0x/contracts-exchange/contracts/src/interfaces/IExchange.sol";
 import "@0x/contracts-exchange/contracts/src/libs/LibExchangeRichErrorDecoder.sol";
 import "@0x/contracts-exchange-libs/contracts/src/LibExchangeRichErrors.sol";
 import "@0x/contracts-exchange-libs/contracts/src/LibOrder.sol";
+import "@0x/contracts-exchange-libs/contracts/src/LibFillResults.sol";
 import "@0x/contracts-utils/contracts/src/LibBytes.sol";
 
 
@@ -66,6 +67,13 @@ contract OrderTransferSimulationUtils is
         public
         returns (OrderTransferResults orderTransferResults)
     {
+        LibFillResults.FillResults memory fillResults = LibFillResults.calculateFillResults(
+            order,
+            takerAssetFillAmount,
+            _EXCHANGE.protocolFeeMultiplier(),
+            tx.gasprice
+        );
+
         // Create input arrays
         bytes[] memory assetData = new bytes[](4);
         address[] memory fromAddresses = new address[](4);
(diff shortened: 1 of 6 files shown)


contracts/dev-utils/contracts/src/AssetData.sol
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+pragma solidity ^0.5.5;
+pragma experimental ABIEncoderV2;
+
+import "./LibAssetData.sol";
+
+
+// solhint-disable no-empty-blocks
+contract AssetData is
```
jalextowle: Why do we need to have this contract? I can see that it's used in tests, but it doesn't seem like it's used in any Solidity contracts.

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -66,6 +67,13 @@ contract OrderTransferSimulationUtils is
         public
         returns (OrderTransferResults orderTransferResults)
     {
+        LibFillResults.FillResults memory fillResults = LibFillResults.calculateFillResults(
```
jalextowle: 🙏

contracts/dev-utils/contracts/src/TransactionDecoder.sol
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+pragma solidity ^0.5.5;
+pragma experimental ABIEncoderV2;
+
+import "./LibTransactionDecoder.sol";
+
+
+// solhint-disable no-empty-blocks
+contract TransactionDecoder is
```
jalextowle: Same question as above. What does this pattern do for us?

contracts/exchange/contracts/src/interfaces/IProtocolFees.sol
```diff
@@ -36,4 +36,16 @@ contract IProtocolFees {
     /// @param updatedProtocolFeeCollector The updated protocolFeeCollector contract address.
     function setProtocolFeeCollectorAddress(address updatedProtocolFeeCollector)
         external;
+
+    /// @dev Returns the protocolFeeMultiplier
+    function protocolFeeMultiplier()
```
jalextowle: Nice!

feuGeneA:
I tried tweaking compiler optimization options, in an attempt to reduce the bytecode size of `DevUtils`, to no avail:

image

Note that the latter rows of that spreadsheet show combined sizes of different contracts, to explore the idea of grouping related contracts together into one deployment.  The takeaway is that `OrderValidationUtils` alone is already approaching the ~25KB limit, so it can't really be combined with anything else (except for `EthBalanceChecker`, though grouping those two together doesn't really make much sense).

contracts/dev-utils/contracts/src/AssetData.sol
feuGeneA: It's important for us to expose EVM implementations of asset data encoding and decoding, in order for us to be able to easily support any language in the future.  For example, if someone wants to interact with our contracts from Java, they'd need to have a Java-native implementation of this functionality (which we don't want to provide, and which we don't want to burden them with having to write), OR, they could just call into our EVM implementation, and go on their merry way.  With this, any language that has a Web3 implementation can easily be used with our contracts.

Prior to this PR, the intention was to publish `DevUtils` as an interface to `LibAssetData`, among other things.  But given that we decided to split up the contracts, I needed a vehicle to deploy the library, so I created this contract.

contracts/dev-utils/contracts/src/TransactionDecoder.sol
feuGeneA: Same logic as https://github.com/0xProject/0x-monorepo/pull/2161#discussion_r325899210

contracts/exchange/contracts/src/interfaces/IProtocolFees.sol
feuGeneA: Credit to @abandeali1 for that one :smile:

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
feuGeneA: also @abandeali1 :)

contracts/dev-utils/package.json
```diff
@@ -12,7 +12,7 @@
     "scripts": {
         "build": "yarn pre_build && tsc -b",
         "build:ci": "yarn build",
-        "pre_build": "run-s compile contracts:gen generate_contract_wrappers",
+        "pre_build": "run-s compile quantify_bytecode contracts:gen generate_contract_wrappers",
```
jalextowle: 🎉

PR #2166, opened 2019-09-17
Staking params upgradability

abandeali1:
## Description

This PR:

- Creates a unified interface for upgrading any params within the staking contracts. `wethAssetProxy`, `ethVault`, `rewardVault`, and `zrxVault` are all now initialized in `init` and upgradable via `setParams`. Their old setter methods have been removed.
- Note that `attachStakingContract` now also requires these params to be passed in. However, they will be ignored in favor of the already stored params if null addresses are passed in.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/stake/MixinZrxVault.sol
+++ b/contracts/staking/contracts/src/stake/MixinZrxVault.sol
@@ -33,8 +33,7 @@ contract MixinZrxVault is
     function _depositFromOwnerIntoZrxVault(address owner, uint256 amount)
         internal
     {
-        IZrxVault _zrxVault = zrxVault;
-        _zrxVault.depositFrom(owner, amount);
+        zrxVault.depositFrom(owner, amount);
     }
 
     /// @dev Withdraws Zrx Tokens from to `owner` from the vault.
@@ -43,8 +42,7 @@ contract MixinZrxVault is
     function _withdrawToOwnerFromZrxVault(address owner, uint256 amount)
         internal
     {
-        IZrxVault _zrxVault = zrxVault;
-        _zrxVault.withdrawFrom(owner, amount);
+        zrxVault.withdrawFrom(owner, amount);
     }
 
     /// @dev Returns balance of `owner` in the ZRX ault.
@@ -54,7 +52,6 @@ contract MixinZrxVault is
         view
         returns (uint256)
     {
-        IZrxVault _zrxVault = zrxVault;
-        return _zrxVault.balanceOf(owner);
+        return zrxVault.balanceOf(owner);
     }
 }
--- a/contracts/staking/contracts/src/stake/MixinZrxVault.sol
+++ b/contracts/staking/contracts/src/stake/MixinZrxVault.sol
@@ -27,15 +27,6 @@ import "../immutable/MixinStorage.sol";
 contract MixinZrxVault is
     MixinStorage
 {
-    /// @dev Set the Zrx Vault.
-    /// @param zrxVaultAddress Address of the Zrx Vault.
-    function setZrxVault(address zrxVaultAddress)
-        external
-        onlyOwner
-    {
-        zrxVault = IZrxVault(zrxVaultAddress);
-    }
-
     /// @dev Deposits Zrx Tokens from the `owner` into the vault.
     /// @param owner of Zrx Tokens
     /// @param amount of tokens to deposit.
@@ -43,10 +34,6 @@ contract MixinZrxVault is
         internal
     {
         IZrxVault _zrxVault = zrxVault;
-        require(
-            address(_zrxVault) != address(0),
-            "INVALID_ZRX_VAULT"
-        );
         _zrxVault.depositFrom(owner, amount);
     }
 
@@ -57,10 +44,6 @@ contract MixinZrxVault is
         internal
     {
         IZrxVault _zrxVault = zrxVault;
-        require(
-            address(_zrxVault) != address(0),
-            "INVALID_ZRX_VAULT"
-        );
         _zrxVault.withdrawFrom(owner, amount);
     }
 
@@ -72,10 +55,6 @@ contract MixinZrxVault is
         returns (uint256)
     {
         IZrxVault _zrxVault = zrxVault;
-        require(
-            address(_zrxVault) != address(0),
-            "INVALID_ZRX_VAULT"
-        );
         return _zrxVault.balanceOf(owner);
     }
 }
--- a/contracts/staking/contracts/src/sys/MixinParams.sol
+++ b/contracts/staking/contracts/src/sys/MixinParams.sol
@@ -125,7 +125,7 @@ contract MixinParams is
         internal
     {
         // Ensure state is uninitialized.
-        _assertStorageNotInitialized();
+        _assertParamsNotInitialized();
 
         // Set up defaults.
         // These cannot be set to variables, or we go over the stack variable limit.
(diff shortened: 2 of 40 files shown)


contracts/test-utils/src/address_utils.ts
```diff
@@ -1,11 +1,9 @@
...
-        const randomBigNum = generatePseudoRandomSalt();
-        const randomBuff = crypto.solSHA3([randomBigNum]);
-        const randomAddress = `0x${randomBuff.slice(0, 20).toString('hex')}`;
-        return randomAddress;
-    },
-};
+/**
+ * Generates a random address.
+ */
+export function randomAddress(): string {
```
dorothy-zbornak: I don't know why it took us so long. :upside_down_face:

contracts/staking/test/migration.ts
```diff
@@ -122,15 +136,74 @@ blockchainTests('Migration tests', env => {
...
+                    constants.NULL_ADDRESS,
+                    constants.NULL_ADDRESS,
+                    constants.NULL_ADDRESS,
+                );
+                const logsArgs = filterLogsToArguments<TestInitTargetInitAddressesEventArgs>(
+                    receipt.logs,
+                    'InitAddresses',
+                );
+                for (const args of logsArgs) {
+                    expect(args.wethProxyAddress).to.eq(wethProxyAddress);
```
dorothy-zbornak: Why are we just asserting `wethProxyAddress` and not the others?

contracts/staking/test/migration.ts
```diff
@@ -122,15 +136,74 @@ blockchainTests('Migration tests', env => {
...
+                    ethVaultAddress,
+                    rewardVaultAddress,
+                    zrxVaultAddress,
+                );
+                const logsArgs = filterLogsToArguments<TestInitTargetInitAddressesEventArgs>(
+                    receipt.logs,
+                    'InitAddresses',
+                );
+                for (const args of logsArgs) {
+                    expect(args.wethProxyAddress).to.eq(wethProxyAddress);
```
dorothy-zbornak: Ditto

contracts/staking/test/migration.ts
abandeali1: Whoops, forgot to add those in 😅

contracts/staking/contracts/src/sys/MixinParams.sol
```diff
@@ -86,29 +92,72 @@ contract MixinParams is
...
+        _wethProxyAddress = address(wethAssetProxy);
+        _ethVaultAddress = address(ethVault);
+        _rewardVaultAddress = address(rewardVault);
+        _zrxVaultAddress = address(zrxVault);
     }
 
-    /// @dev Assert param values before initializing them.
-    /// This must be updated for each migration.
-    function _assertMixinParamsBeforeInit()
+    /// @dev Initialzize storage belonging to this mixin.
```
jalextowle: small nit: `Initialzize` => `Initialize`

contracts/staking/contracts/src/sys/MixinParams.sol
```diff
@@ -86,29 +92,72 @@ contract MixinParams is
...
+    )
         internal
     {
         // Ensure state is uninitialized.
+        _assertParamsNotInitialized();
+
+        // Set up defaults.
+        // These cannot be set to variables, or we go over the stack variable limit.
+        _setParams(
+            2 weeks,                       // epochDurationInSeconds
```
jalextowle: Nit: Rather than setting these to variables, it would probably be good to make these constants, in case we ever want to change them. Additionally, if they are constants, we can use the internal function pattern to stub out these values, if necessary.

contracts/staking/test/params.ts
```diff
@@ -20,29 +23,9 @@ blockchainTests('Configurable Parameters', env => {
     });
 
     blockchainTests.resets('setParams()', () => {
-        interface Params {
```
mzhu25: 🙏

contracts/staking/contracts/src/sys/MixinParams.sol
```diff
@@ -184,4 +281,47 @@ contract MixinParams is
             ));
         }
     }
+
+    /// @dev Asserts that passed in addresses are non-zero.
+    /// @param _wethProxyAddress The address that can transfer WETH for fees.
+    /// @param _ethVaultAddress Address of the EthVault contract.
+    /// @param _rewardVaultAddress Address of the StakingPoolRewardVault contract.
+    /// @param _zrxVaultAddress Address of the ZrxVault contract.
+    function _assertValidAddresses(
```
jalextowle: Since these addresses should all be contracts and this function will be called extremely infrequently, I think it might be worth checking that these accounts are in fact deployed contracts. I think this could be good for trustlessness and as a sanity check on the addresses that are registered.

contracts/staking/contracts/src/stake/MixinZrxVault.sol
```diff
@@ -27,26 +27,13 @@ import "../immutable/MixinStorage.sol";
...
-        zrxVault = IZrxVault(zrxVaultAddress);
-    }
-
     /// @dev Deposits Zrx Tokens from the `owner` into the vault.
     /// @param owner of Zrx Tokens
     /// @param amount of tokens to deposit.
     function _depositFromOwnerIntoZrxVault(address owner, uint256 amount)
         internal
     {
         IZrxVault _zrxVault = zrxVault;
```
mzhu25: we no longer need to cache this

contracts/staking/contracts/src/stake/MixinZrxVault.sol
```diff
@@ -57,10 +44,6 @@ contract MixinZrxVault is
...
```
mzhu25: ditto

contracts/staking/contracts/src/stake/MixinZrxVault.sol
```diff
@@ -72,10 +55,6 @@ contract MixinZrxVault is
         returns (uint256)
     {
         IZrxVault _zrxVault = zrxVault;
```
mzhu25: ditto

contracts/staking/test/migration.ts
```diff
@@ -93,21 +90,38 @@ blockchainTests('Migration tests', env => {
 
             it('throws if not called by owner', async () => {
                 const attachedAddress = randomAddress();
-                const tx = proxyContract.attachStakingContract.awaitTransactionSuccessAsync(attachedAddress, {
-                    from: notOwnerAddress,
-                });
+                const tx = proxyContract.attachStakingContract.awaitTransactionSuccessAsync(
+                    attachedAddress,
```
mzhu25: nit: `initTargetContract.address` for consistency

contracts/staking/test/utils/types.ts
```diff
@@ -4,11 +4,15 @@ import { constants } from './constants';
 
 export interface StakingParams {
     epochDurationInSeconds: BigNumber;
-    rewardDelegatedStakeWeight: BigNumber;
+    rewardDelegatedStakeWeight: number | BigNumber;
```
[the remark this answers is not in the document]
abandeali1: I was using `Numberish`, but some of the contract wrapper functions don't accept strings as inputs and had to revert back to this.

contracts/staking/contracts/src/sys/MixinParams.sol
abandeali1: Had to do this to get around stack limitations unfortunately.

contracts/staking/contracts/src/sys/MixinParams.sol
abandeali1: Is it really trustless if it's just a contract that allows someone to arbitrarily withdraw funds? 🤷‍♂

contracts/staking/contracts/src/stake/MixinZrxVault.sol
abandeali1: I'm probably going to just delete this mixin in a separate PR. It doesn't really do anything at the moment.

PR #2168, opened 2019-09-17
Storage Layout Assertions

jalextowle:
## Description
One of the issues involved with using a proxy architecture for smart contract upgradeability is that all future logic contracts must adhere to the original storage layout. Serious problems may occur if the storage layout of future logic contracts is even slightly altered. With this in mind, this PR adds assertions to the constructor of the Staking logic contract so that any future upgrades will be forced to be `storage-layout compatible` with all previous Staking logic contracts.

This was as straightforward as PRs get, but there was one interesting piece of information that came out of this. For all tightly packed storage values, we must use the `_offset` value (available in inline assembly) to determine their offset within the storage word addressed by their `_slot` value. This `_offset` value increases for storage values that are defined later, and they only increase by the combined size of the previous words (so a `uint64 b` defined after a `uint32 a` will have `b_offset == 4`).

## Testing instructions

Deploy the Staking Logic contract. If it reverts, then you changed the storage layout :)

## Types of changes

## Checklist:


contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -488,4 +505,21 @@ library LibStakingRichErrors {
             endEpoch
         );
     }
+
+    function StorageLayoutError(
```
dorothy-zbornak: I'm assuming this doesn't get included in the bytecode and is purely for reference, right? I'm on board.

contracts/staking/contracts/src/Staking.sol
```diff
@@ -66,4 +74,298 @@ contract Staking is
...
+        internal
+        pure
+    {
+        assembly {
+            let slot := 0x0
+            let offset := 0x0
+
+            /// Ownable
+
+            assertSlotAndOffset(
```
hysz: Totally optional, but it could be more readable to make an array where each element has a state variable's slot/offset. Then loop through, do the check in assembly and revert on failure in regular solidity.

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
jalextowle: This was actually just an accident 🙈

mzhu25:
Let's either kill or fix `TestStorageLayout`; it probably makes the most sense to do so in this PR

PR #2169, opened 2019-09-17
Fix breaking tests, remove MixinZrxVault

abandeali1:
## Checklist:


contracts/staking/contracts/src/stake/MixinZrxVault.sol
```diff
@@ -1,57 +0,0 @@
-/*
```
jalextowle: HUGE RIP 💯

PR #2178, opened 2019-09-20
Upgradability controls

abandeali1:
## Description

This PR:

- Moves storage param assertions directly into the `StakingProxy`. This gives stronger guarantees that the contract will not enter an invalid state.
- Adds boundaries to the epoch duration. It must now be between 5 and 30 days long.
- Changes the default epoch duration to 10 days.
- Changes all instances of `onlyOwner` to `onlyAuthorized` throughout the Staking and Vault contracts. We can also make this change in the `Exchange` contract in a separate PR.
- Refactors `Authorizable` to use internal functions. There are some redundant checks which cost some extra gas, but I'm happy with the simplicity.
- `_addAuthorizedAddress(owner)` is then called in the constructors of `MixinStorage` and the Vault contracts.

## Checklist:


contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -164,6 +164,79 @@ contract StakingProxy is
...
+    //       Asserts that a stake weight is <= 100%.
+    //       Asserts that pools allow >= 1 maker.
+    //       Asserts that all addresses are initialized.
+    function _assertValidStorageParams()
+        internal
+        view
+    {
+        // Epoch length must be between 5 and 30 days long
+        uint256 _epochDurationInSeconds = epochDurationInSeconds;
+        if (_epochDurationInSeconds < 5 days || _epochDurationInSeconds > 30 days) {
```
hysz: I always forget about this syntax lol. +1

contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -164,6 +164,79 @@ contract StakingProxy is
         return batchReturnData;
     }
 
+    /// @dev Asserts that an epoch is between 5 and 30 days long.
+    //       Asserts that 0 < cobb douglas alpha value <= 1.
+    //       Asserts that a stake weight is <= 100%.
+    //       Asserts that pools allow >= 1 maker.
+    //       Asserts that all addresses are initialized.
+    function _assertValidStorageParams()
```
hysz: Much clean. Very assert 💯

contracts/staking/contracts/src/vaults/MixinVaultCore.sol
```diff
@@ -71,12 +71,20 @@ contract MixinVaultCore is
         _;
     }
 
+    /// @dev Sets the vault owner and adds owner as an authorized address.
+    constructor()
+        public
+        Ownable()
```
hysz: Should this call the constructor for `Authorizable`?

contracts/utils/contracts/src/Authorizable.sol
```diff
@@ -43,19 +43,7 @@ contract Authorizable is
         external
         onlyOwner
     {
```
hysz: Nit - I see there isn't actually a constructor in `Authorizable`. Maybe we could add one, which for now would just call `Ownable()`. Just to abstract it away from contracts that inherit this.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolMakers.sol
```diff
@@ -177,7 +177,9 @@ contract MixinStakingPoolMakers is
         }
 
         // Is the pool already full?
-        if (pool.numberOfMakers == maximumMakersInPool) {
+        // NOTE: If maximumMakersInPool is decreased below the number of makers currently in a pool,
+        // the pool will no longer be able to add more makers.
+        if (pool.numberOfMakers >= maximumMakersInPool) {
```
mzhu25: 🙏

contracts/utils/contracts/src/Authorizable.sol
abandeali1: Makes sense. I don't think it will make a difference at the bytecode level, but good to be explicit.

PR #2188, opened 2019-09-23
Remove reference counting for cumulative rewards

hysz:
## Description

This removes reference counting for cumulative rewards.

Reference counting allowed us to clear old state that was no longer needed by the staking contract. After reviewing the tradeoffs with the protocologists, it was decided that we would remove this feature in favor of simplicity.

See #2154 for full context.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
+++ b/contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
@@ -294,10 +294,9 @@ contract MixinStakingPoolRewards is
     /// @param poolId Unique id of pool.
     /// @param unsyncedStake Unsynced delegated stake to pool by owner
     /// @param currentEpoch The epoch in which this call is executing
-    /// @param unfinalizedMembersReward Unfinalized total members reward
-    ///        (if any).
+    /// @param unfinalizedMembersReward Unfinalized total members reward (if any).
     /// @param unfinalizedMembersStake Unfinalized total members stake (if any).
-    /// @return totalReward Balance in ETH.
+    /// @return reward Balance in WETH.
     function _computeDelegatorReward(
         bytes32 poolId,
         IStructs.StoredBalance memory unsyncedStake,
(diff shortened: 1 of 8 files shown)


contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -393,44 +384,33 @@ contract MixinStakingPoolRewards is
...
+            _trySetCumulativeReward(
                 poolId,
                 _delegatedStakeToPoolByOwner.currentEpoch,
-                mostRecentCumulativeReward,
-                isDependent
+                mostRecentCumulativeReward
             );
         }
 
         // Record dependency on the next epoch
```
dorothy-zbornak: So one thing I was thinking about was instead of explicitly copying the CR to the next epoch, we just checked if we had a record for `epoch-1` if `epoch` was unset in `_computeMemberRewardOverInterval()`. I didn't want to do this before because I wasn't sure if reference counting would get in the way of this (because `epoch-1` could theoretically get deleted), but that shouldn't be a factor now, right?

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -308,10 +299,9 @@ contract MixinStakingPoolRewards is
     /// @param poolId Unique id of pool.
     /// @param unsyncedStake Unsynced delegated stake to pool by owner
     /// @param currentEpoch The epoch in which this call is executing
-    /// @param unfinalizedMembersReward Unfinalized total members reward
-    ///        (if any).
+    /// @param unfinalizedMembersReward Unfinalized total members reward (if any).
     /// @param unfinalizedMembersStake Unfinalized total members stake (if any).
-    /// @return totalReward Balance in ETH.
+    /// @return reward Balance in WETH.
     function _computeDelegatorReward(
```
abandeali1: This wasn't introduced in this PR, but the `currentEpoch` param shadows the `currentEpoch` state variable. Also, it looks like we never actually pass in anything other than the `currentEpoch` state variable, so I think we can just remove this param.

abandeali1: Don't forget to set it to a stack variable first though 😛

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -320,112 +304,121 @@ contract MixinStakingPoolRewards is
...
-            );
-        }
+
+        // 3/3 Finalized rewards earned in epoch `unsyncedStake.currentEpoch`.
+        reward = reward.safeAdd(
+            _computeMemberRewardOverInterval(
+                poolId,
+                unsyncedStake.nextEpochBalance,
+                unsyncedStakeNextEpoch,
+                _currentEpoch
```
dorothy-zbornak: Hmm, doesn't this miss rewards that have been credited an to a pool AFTER a delegator last synced? Also, isn't it possible that neither `_currentEpoch` nor `_currentEpoch-1` would have entries so this could fail?

dorothy-zbornak: discussed offline and I'm wrong :upside_down_face:

PR #2194, opened 2019-09-24
Feature/staking/storage layout assertion

jalextowle:
### Events
Some events were changed, or even deleted. Most of the changes removed event fields or changed the `indexed` status of an event field.

- `TimeLockPeriodChanged`
    - This event was removed entirely because it was not used.
- `PendingAddMakerToPool`
    - `poolId` field is now indexed. This allows this handshaking event to be filtered by `poolId`, which could be useful when analyzing pool registration behavior.
- `MakerAddedToStakingPool`
   - `poolId` is indexed, just like the above event.
- `MakerRemovedFromStakingPool`
   - `poolId` is indexed, just like the above two events.
- `ZrxDepositedIntoVault`
   - `sender` is removed entirely. This event is emitted in a function that uses the `onlyStakingProxy` modifier, and the `sender` field was always populated by `msg.sender`. Since the staking contract will rarely change, the benefit of logging this field is negligible. This field should be completely derivable from the `blocknumber` of the event emission in all but exceedingly rare cases.
- `ZrxWithdrawnFromVault`
   - `sender` is removed entirely. This was removed for the same reason as the above `sender` field.

### Storage Layout Assertions
A sequence of assertions has been added to the constructor of the contract. These assertions will lock in the storage layout for the staking contracts, which should make it easier to develop future versions that are fully compatible. 

## Types of changes

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/Staking.sol
+++ b/contracts/staking/contracts/src/Staking.sol
@@ -75,4 +75,341 @@ contract Staking is
             _zrxVaultAddress
         );
     }
+
+    /// @dev This function will fail if the storage layout of this contract deviates from
+    ///      the original staking contract's storage. The use of this function provides assurance
+    ///      that regressions from the original storage layout will not occur.
+    function _assertStorageLayout()
+        internal
+        pure
+    {
+        assembly {
+            let slot := 0x0
+            let offset := 0x0
+
+            /// Ownable
+
+            assertSlotAndOffset(
+                owner_slot,
+                owner_offset,
+                slot,
+                offset
+            )
+            slot := add(slot, 0x1)
+
+            /// Authorizable
+
+            assertSlotAndOffset(
+                authorized_slot,
+                authorized_offset,
+                slot,
+                offset
+            )
+            slot := add(slot, 0x1)
+
+            assertSlotAndOffset(
+                authorities_slot,
+                authorities_offset,
+                slot,
+                offset
+            )
+            slot := add(slot, 0x1)
+
+            /// MixinStorage
+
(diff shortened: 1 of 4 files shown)


contracts/staking/contracts/src/vaults/ZrxVault.sol
```diff
@@ -96,7 +96,7 @@ contract ZrxVault is
         _balances[owner] = _balances[owner].safeAdd(amount);
 
         // notify
-        emit ZrxDepositedIntoVault(msg.sender, owner, amount);
+        emit ZrxDepositedIntoVault(owner, amount);
```
abandeali1: I prefer simple `Deposit` and `Withdrawal` events personally. We could also conceivably remove these events entirely and just use the ZRX `Transfer` events, though I guess there could be false positives.

contracts/staking/contracts/src/Staking.sol
```diff
@@ -75,4 +75,341 @@ contract Staking is
...
+            )
+            slot := add(slot, 0x1)
+
+            assertSlotAndOffset(
+                unfinalizedState_slot,
+                unfinalizedState_offset,
+                slot,
+                offset
+            )
+
```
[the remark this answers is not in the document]
abandeali1: +1

contracts/staking/contracts/src/vaults/ZrxVault.sol
jalextowle: We could add a single `Deposit` event to prevent false positives. Such an event + the ZRX `Transfer` event should prevent false positives.

PR #2195, opened 2019-09-24
Make wethAssetProxy and zrxVault deployment constants

abandeali1:
## Description

This PR makes `wethAssetProxy` and `zrxVault` deployment constants instead of individually upgradable params. The original purpose of this change was to test how much is would reduce the codesize of the staking contract (only ~500 bytes, it turns out). I'm open to merging it or closing it. The benefit is that it would save a bunch of gas by reducing `sloads`. The downside is that upgrading these params becomes a bit less flexible, as the entire staking logic contract would need to be swapped out.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/StakingProxy.sol
+++ b/contracts/staking/contracts/src/StakingProxy.sol
@@ -36,23 +36,15 @@ contract StakingProxy is
     /// @dev Constructor.
     /// @param _stakingContract Staking contract to delegate calls to.
     /// @param _readOnlyProxy The address of the read only proxy.
-    /// @param _wethProxyAddress The address that can transfer WETH for fees.
-    /// @param _zrxVaultAddress Address of the ZrxVault contract.
     constructor(
         address _stakingContract,
-        address _readOnlyProxy,
-        address _wethProxyAddress,
-        address _zrxVaultAddress
+        address _readOnlyProxy
     )
         public
         MixinStorage()
     {
         readOnlyProxy = _readOnlyProxy;
-        _attachStakingContract(
-            _stakingContract,
-            _wethProxyAddress,
-            _zrxVaultAddress
-        );
+        _attachStakingContract(_stakingContract);
     }
 
     /// @dev Delegates calls to the staking contract, if it is set.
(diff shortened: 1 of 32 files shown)


contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -202,32 +182,11 @@ contract StakingProxy is
                     LibStakingRichErrors.InvalidParamValueErrorCode.InvalidMinimumPoolStake
             ));
         }
-
-        // ERC20Proxy and Vault contract addresses must always be initialized
-        if (address(wethAssetProxy) == NIL_ADDRESS) {
```
hysz: We should still have this test somewhere. Perhaps in the constructor for the logic contract.

contracts/staking/contracts/src/immutable/MixinDeploymentConstants.sol
```diff
@@ -50,6 +52,10 @@ contract MixinDeploymentConstants {
     // Ropsten & Rinkeby Weth Asset Data
     // bytes constant internal WETH_ASSET_DATA = hex"f47261b0000000000000000000000000c778417e063141139fce010982780140aa0cd5ab";
 
+    // @TODO SET THESE VALUES FOR DEPLOYMENT
+    address constant public WETH_ASSET_PROXY_ADDRESS = address(1);
```
hysz: What would you think about making these constants `private` and depending solely on the getters?

This would make it very easy to override for tests (and for different networks).

hysz: We could also get rid of all these constants that have been commented out.

contracts/staking/test/migration.ts
```diff
@@ -380,26 +316,6 @@ blockchainTests('Migration tests', env => {
             );
             expect(tx).to.revertWith(expectedError);
         });
-        it('reverts if wethAssetProxy is 0', async () => {
```
hysz: I think we should still have these, only in the constructor for the deployment constants mixin.

contracts/staking/contracts/src/StakingProxy.sol
dorothy-zbornak: I could live without these `NIL_ADDRESS` tests tbh, since a wrong address is just as bad and, honestly, probably more likely to happen.

abandeali1: I'll just add some basic string reverts in the `MixinDeploymentConstants` constructor for now. We can always remove them if they add too much to the codesize.

contracts/staking/contracts/src/immutable/MixinDeploymentConstants.sol
abandeali1: I like that. I'll leave the commented out ones in for testnet deployments for now.

contracts/staking/test/migration.ts
abandeali1: Added them!

PR #2196, opened 2019-09-24
Quick fixes

dekz:
## Description

* Fixed the webpack publish for asset swapper to be in line with 0x.js
* Modified the migration snapshot name to not clobber v2 until we're ready to overwrite
* Added filler transactions to keep contract addresses consistent whilst asset proxy is disabled
* Non-private package publish for orderbook

### Ganache Snapshot
```
docker run -it --rm -p 8545:8545 -e SNAPSHOT_NAME=0x_ganache_snapshot-v3-beta -e VERSION=4.3.2 0xorg/ganache-cli:latest
```
http://ganache-snapshots.0x.org.s3.amazonaws.com/0x_ganache_snapshot-v3-beta-4.3.2.zip

### DIFF
--- a/packages/orderbook/package.json
+++ b/packages/orderbook/package.json
@@ -12,6 +12,9 @@
         "type": "git",
         "url": "https://github.com/0xProject/0x-monorepo.git"
     },
+    "publishConfig": {
+        "access": "public"
+    },
     "scripts": {
         "clean": "shx rm -rf lib generated_docs",
         "test": "jest",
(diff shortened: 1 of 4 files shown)


packages/orderbook/package.json
```diff
@@ -12,6 +12,9 @@
         "type": "git",
         "url": "https://github.com/0xProject/0x-monorepo.git"
     },
+    "publishConfig": {
+        "access": "public"
```
xianny: I'm not familiar with this package. Was there a reason it was private before and now should be public? Or was it just an unintentional omission?

dekz: it was accidentally omitted and killed the publish :(.

PR #2197, opened 2019-09-24
Quick fix to abi-gen test

xianny:
## Description

Adding different error types to `@0x/base-contract` resulted in this test failing. Not sure why it wasn't failing on CI before. This PR just updates the error type in the unit test.

## Testing instructions

CI should pass. 

## Types of changes

## Checklist:

### DIFF
--- a/packages/abi-gen/test-cli/test_typescript/test/abi_gen_dummy_test.ts
+++ b/packages/abi-gen/test-cli/test_typescript/test/abi_gen_dummy_test.ts
@@ -1,13 +1,12 @@
 import { BlockchainLifecycle, devConstants, web3Factory } from '@0x/dev-utils';
 import { Web3ProviderEngine } from '@0x/subproviders';
-import { BigNumber, providerUtils } from '@0x/utils';
+import { BigNumber, providerUtils, StringRevertError } from '@0x/utils';
 import { BlockParamLiteral, Web3Wrapper } from '@0x/web3-wrapper';
 import * as chai from 'chai';
 import * as chaiAsPromised from 'chai-as-promised';
 import * as ChaiBigNumber from 'chai-bignumber';
 import * as dirtyChai from 'dirty-chai';
 import * as Sinon from 'sinon';
-import { constants } from 'websocket';
 
 import {
     AbiGenDummyContract,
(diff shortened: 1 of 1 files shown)


packages/abi-gen/test-cli/test_typescript/test/abi_gen_dummy_test.ts
```diff
@@ -1,13 +1,12 @@
...
 import { Web3ProviderEngine } from '@0x/subproviders';
-import { BigNumber, providerUtils } from '@0x/utils';
+import { BigNumber, providerUtils, StringRevertError } from '@0x/utils';
 import { BlockParamLiteral, Web3Wrapper } from '@0x/web3-wrapper';
 import * as chai from 'chai';
 import * as chaiAsPromised from 'chai-as-promised';
 import * as ChaiBigNumber from 'chai-bignumber';
 import * as dirtyChai from 'dirty-chai';
 import * as Sinon from 'sinon';
-import { constants } from 'websocket';
```
xianny: this was un-used

packages/abi-gen/test-cli/test_typescript/test/abi_gen_dummy_test.ts
```diff
@@ -81,22 +80,34 @@ describe('AbiGenDummy Contract', () => {
     });
     describe('simpleRevert', () => {
         it('should call simpleRevert', async () => {
-            expect(abiGenDummy.simpleRevert.callAsync()).to.be.rejectedWith('SIMPLE_REVERT');
+            abiGenDummy.simpleRevert.callAsync().catch(err => {
+                const expected = new StringRevertError('SIMPLE_REVERT');
+                expect(err).to.deep.equal(expected);
+            });
```
feuGeneA: I'd really like to see us keep the the `expect(...).to.be.rejectedWith(...)` syntax, and I was _sure_ it was possible, but it sure took me a while to figure it out!  How do you feel about switching it to this format?

```
            expect(abiGenDummy.simpleRevert.callAsync())
                .to.be.rejectedWith(StringRevertError)
                .deep.equal(new StringRevertError('SIMPLE_REVERT'));
```

xianny: Nice! I tried a couple things but gave up early, thanks for taking the time. I think this is also better because it fails sequentially. The way I had it might show a green check mark for a test that actually failed, and then throw an error anyway.

PR #2201, opened 2019-09-24
Rename shadowed instances of `owner` to `staker`

abandeali1:
## Checklist:


contracts/staking/contracts/src/vaults/ZrxVault.sol
```diff
@@ -70,7 +70,7 @@ contract ZrxVault is
     }
 
     /// @dev Sets the Zrx proxy.
-    /// Note that only the contract owner can call this.
+    /// Note that only the contract staker can call this.
```
jalextowle: Nit: This should probably still be `owner`. With this in mind, should this be changed to `onlyAuthorized`?

PR #2207, opened 2019-09-25
Improve readability, fix `_trySetCumulativeReward` logic

abandeali1:
## Description

This PR mostly renames variables and flattens some of the logic in `MixinStakingPoolRewards` to improve readability. In the process, I came across a bug in `_trySetCumulativeReward` where it would _always_ call `_forceSetCumulativeReward`. It would return early only iff `epoch < currentEpoch && ...`. The value passed in as `epoch` was always `_loadSyncedBalance(...).currentEpoch`, which is necessarily >= `currentEpoch`.

When reviewing this PR, please note the changes in the test cases and double check that this in intended behavior.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/stake/MixinStake.sol
+++ b/contracts/staking/contracts/src/stake/MixinStake.sol
@@ -187,6 +187,11 @@ contract MixinStake is
         // Sanity check the pool we're delegating to exists.
         _assertStakingPoolExists(poolId);
 
+        _withdrawAndSyncDelegatorRewards(
+            poolId,
+            staker
+        );
+
         // Increment how much stake the staker has delegated to the input pool.
         _increaseNextBalance(
             _delegatedStakeToPoolByOwner[staker][poolId],
@@ -195,11 +200,6 @@ contract MixinStake is
 
         // Increment how much stake has been delegated to pool.
         _increaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        _withdrawAndSyncDelegatorRewards(
-            poolId,
-            staker
-        );
     }
 
     /// @dev Un-Delegates a owners stake from a staking pool.
@@ -216,6 +216,11 @@ contract MixinStake is
         // sanity check the pool we're undelegating from exists
         _assertStakingPoolExists(poolId);
 
+        _withdrawAndSyncDelegatorRewards(
+            poolId,
+            staker
+        );
+
         // decrement how much stake the staker has delegated to the input pool
         _decreaseNextBalance(
             _delegatedStakeToPoolByOwner[staker][poolId],
@@ -224,11 +229,6 @@ contract MixinStake is
 
         // decrement how much stake has been delegated to pool
         _decreaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        _withdrawAndSyncDelegatorRewards(
-            poolId,
-            staker
-        );
     }
 
     /// @dev Returns a storage pointer to a user's stake in a given status.
--- a/contracts/staking/contracts/src/stake/MixinStake.sol
+++ b/contracts/staking/contracts/src/stake/MixinStake.sol
@@ -187,9 +187,10 @@ contract MixinStake is
         // Sanity check the pool we're delegating to exists.
         _assertStakingPoolExists(poolId);
 
-        // Cache amount delegated to pool by staker.
-        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
-            _loadUnsyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
+        _withdrawAndSyncDelegatorRewards(
+            poolId,
+            staker
+        );
 
         // Increment how much stake the staker has delegated to the input pool.
         _increaseNextBalance(
@@ -199,12 +200,6 @@ contract MixinStake is
 
         // Increment how much stake has been delegated to pool.
         _increaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        _withdrawAndSyncDelegatorRewards(
-            poolId,
-            staker,
-            initDelegatedStakeToPoolByOwner
-        );
     }
 
     /// @dev Un-Delegates a owners stake from a staking pool.
@@ -221,9 +216,10 @@ contract MixinStake is
         // sanity check the pool we're undelegating from exists
         _assertStakingPoolExists(poolId);
 
-        // Cache amount delegated to pool by staker.
-        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
-            _loadUnsyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
+        _withdrawAndSyncDelegatorRewards(
+            poolId,
+            staker
+        );
 
         // decrement how much stake the staker has delegated to the input pool
         _decreaseNextBalance(
@@ -233,12 +229,6 @@ contract MixinStake is
 
         // decrement how much stake has been delegated to pool
         _decreaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        _withdrawAndSyncDelegatorRewards(
-            poolId,
-            staker,
-            initDelegatedStakeToPoolByOwner
-        );
     }
 
     /// @dev Returns a storage pointer to a user's stake in a given status.
--- a/contracts/staking/contracts/src/stake/MixinStake.sol
+++ b/contracts/staking/contracts/src/stake/MixinStake.sol
@@ -188,7 +188,7 @@ contract MixinStake is
         _assertStakingPoolExists(poolId);
 
         // Cache amount delegated to pool by staker.
-        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
+        IStructs.StoredBalance memory initDelegatedStake =
             _loadUnsyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
 
         // Increment how much stake the staker has delegated to the input pool.
@@ -202,14 +202,14 @@ contract MixinStake is
 
         // Synchronizes reward state in the pool that the owner is delegating
         // to.
-        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner =
+        IStructs.StoredBalance memory finalDelegatedStake =
             _loadSyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
 
         _withdrawAndSyncDelegatorRewards(
             poolId,
             staker,
-            initDelegatedStakeToPoolByOwner,
-            finalDelegatedStakeToPoolByOwner
+            initDelegatedStake,
+            finalDelegatedStake
         );
     }
 
@@ -228,7 +228,7 @@ contract MixinStake is
         _assertStakingPoolExists(poolId);
 
         // cache amount delegated to pool by staker
-        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
+        IStructs.StoredBalance memory initDelegatedStake =
             _loadUnsyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
 
         // decrement how much stake the staker has delegated to the input pool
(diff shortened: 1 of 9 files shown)


contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -57,27 +57,6 @@ contract MixinCumulativeRewards is
...
-    function _trySetCumulativeReward(
-        bytes32 poolId,
-        uint256 epoch,
-        IStructs.Fraction memory value
-    )
-        internal
-    {
-        // Do nothing if it's in the past since we don't want to
-        // rewrite history.
-        if (epoch < currentEpoch
```
hysz: This was originally:
```
    if (_isCumulativeRewardSet(_cumulativeRewardsByPool[poolId][epoch])) {
        return
    }
```

The intention was: if there is a value then do not overwrite it.

The current version:
```
if (epoch < currentEpoch
            && _isCumulativeRewardSet(_cumulativeRewardsByPool[poolId][epoch]))
    {
        return;
    }
```

The intention here seems to be: it's okay to set historical values, so long as we aren't overwriting anything. And it's always okay to update the value for the current or a future epoch. Is this the intended behavior @dorothy-zbornak ?

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -95,33 +74,19 @@ contract MixinCumulativeRewards is
...
-        uint256 epoch
-    )
-        internal
-    {
-        // Check if we should do any work
-        uint256 currentMostRecentEpoch = _cumulativeRewardsByPoolLastStored[poolId];
-        if (epoch == currentMostRecentEpoch) {
-            return;
+            // Update state to reflect the most recent cumulative reward
+            _forceSetMostRecentCumulativeRewardEpoch(
```
dorothy-zbornak: Does anyone else call this function? It might be better just to inline it here because it's so trivial that it might be clearer.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
dorothy-zbornak: Doesn't the latter imply that we *can't* set historical values?
This function is gone, though, so can you point to where you derived the new conditions?

From what I can tell, it doesn't look like we ever set a CR at any epoch other than `currentEpoch`.

abandeali1: This PR removes this part of the condition: `epoch < currentEpoch`. That was _always_ evaluating to `false`, so this function would never return early.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -95,33 +74,19 @@ contract MixinCumulativeRewards is
         // Never set the most recent reward epoch to one in the future, because
         // it may get removed if there are no more dependencies on it.
         if (epoch <= currentEpoch) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Also, if this is the case, we might as well just remove the `epoch` parameter and just use `currentEpoch` to prevent accidents.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
hysz: Nvm I see the issue is that this `return` never gets triggered by an existing codepath. So, in theory, this could corrupt state if we finalized a pool and then carried out some action in the same epoch that overwrote the value. 

That said, the function that calls this overwrites using the most recent value. So, this effectively performs a NOP if the CR is already set.

So, at present, this bug is fairly benign and manifests by wasting gas.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
abandeali1: I didn't inline it because it broke too many tests, but I would be down for doing so.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -57,89 +56,28 @@ contract MixinCumulativeRewards is
...
-    function _trySetMostRecentCumulativeRewardEpoch(
-        bytes32 poolId,
-        uint256 epoch
-    )
-        internal
-    {
         // Check if we should do any work
         uint256 currentMostRecentEpoch = _cumulativeRewardsByPoolLastStored[poolId];
-        if (epoch == currentMostRecentEpoch) {
+        if (currentEpoch_ == currentMostRecentEpoch) {
```
hysz: Let's scrap this optimization. The logic local to this function doesn't know if the caller intends to overwrite.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -57,89 +56,28 @@ contract MixinCumulativeRewards is
...
-        bytes32 poolId,
-        uint256 currentMostRecentEpoch,
-        uint256 newMostRecentEpoch
-    )
-        internal
-    {
-        // Sanity check that we're not trying to go back in time
-        assert(newMostRecentEpoch >= currentMostRecentEpoch);
-        _cumulativeRewardsByPoolLastStored[poolId] = newMostRecentEpoch;
+        assert(currentEpoch_ > currentMostRecentEpoch);
```
hysz: Scrapping the above optimization, this would become `>=`

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -227,29 +216,18 @@ contract MixinStake is
...
         // decrement how much stake has been delegated to pool
-        _decrementNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        // synchronizes reward state in the pool that the owner is undelegating
-        // from
-        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner =
-            _loadSyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
+        _decreaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
 
         _withdrawAndSyncDelegatorRewards(
```
hysz: This needs to be moved above the `decrease`.

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -187,29 +187,18 @@ contract MixinStake is
...
         // Increment how much stake has been delegated to pool.
-        _incrementNextBalance(_delegatedStakeByPoolId[poolId], amount);
-
-        // Synchronizes reward state in the pool that the owner is delegating
-        // to.
-        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner =
-            _loadSyncedBalance(_delegatedStakeToPoolByOwner[staker][poolId]);
+        _increaseNextBalance(_delegatedStakeByPoolId[poolId], amount);
 
         _withdrawAndSyncDelegatorRewards(
```
hysz: This needs to moved to above the increases.

contracts/staking/contracts/src/staking_pools/MixinStakingPoolRewards.sol
```diff
@@ -118,32 +112,39 @@ contract MixinStakingPoolRewards is
...
     /// @param member of the pool.
-    /// @param initialDelegatedStakeToPoolByOwner The member's delegated
-    ///        balance at the beginning of this transaction.
-    /// @param finalDelegatedStakeToPoolByOwner The member's delegated balance
-    ///        at the end of this transaction.
     function _withdrawAndSyncDelegatorRewards(
         bytes32 poolId,
-        address member,
-        IStructs.StoredBalance memory initialDelegatedStakeToPoolByOwner,
-        IStructs.StoredBalance memory finalDelegatedStakeToPoolByOwner
```
hysz: I don't think these should be removed. As-per the function comment, these values are the delegated stake at the beginning and end of the transaction. 

Ex, before delegating and after delegating. Before withdrawing and after withdrawing. 

We're now just calling `loadUnsynced` and `loadSynced` which is not the same.

hysz: That said, since this change has already been made, we might be able to make this work by removing an optimization.

Currently we're now passing in the `init` value of delegated stake, which is needed for most operations - like computing balances. The `final` value is now only needed when determining whether we depend on the CR for this epoch or not (in `_setCumulativeRewardDependenciesForDelegator`). We _could_ just blindly say we depend on this epoch. It'll create more state bloat, but it does simplify the code.

abandeali1: > We're now just calling `loadUnsynced` and `loadSynced` which is not the same.

If you look at the codepaths where this function is called, it was _always_ using `loadUnsynced` and `loadSynced`.

abandeali1: The only changes in functionality with this PR should be the bug fix - everything else is a refactor. If you see a spot where that is not the case let me know!

hysz: @abandeali1 See [here](https://github.com/0xProject/0x-monorepo/blob/3.0/contracts/staking/contracts/src/stake/MixinStake.sol#L180). The unsynced is loaded before delegating and the synced is loaded after delegating.

abandeali1: I see. I think we can revert this commit then: https://github.com/0xProject/0x-monorepo/pull/2207/commits/22a6de48ae3bc133f7d3f21ad506f7794127e8bb

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -188,7 +188,7 @@ contract MixinStake is
         _assertStakingPoolExists(poolId);
 
         // Cache amount delegated to pool by staker.
-        IStructs.StoredBalance memory initDelegatedStake =
+        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
```
abandeali1: Nit: can we keep the `initDelegatedStake` name for this? `initDelegatedStakeToPoolByOwner` doesn't really describe this accurately (and also mirrors the name of the mapping, which is confusing imo).

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -227,8 +221,8 @@ contract MixinStake is
         // sanity check the pool we're undelegating from exists
         _assertStakingPoolExists(poolId);
 
-        // cache amount delegated to pool by staker
-        IStructs.StoredBalance memory initDelegatedStake =
+        // Cache amount delegated to pool by staker.
+        IStructs.StoredBalance memory initDelegatedStakeToPoolByOwner =
```
abandeali1: Same as above - prefer `initDelegatedStake`.

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -42,10 +42,10 @@ contract MixinStake is
         getZrxVault().depositFrom(staker, amount);
 
         // mint stake
-        _incrementCurrentAndNextBalance(_activeStakeByOwner[staker], amount);
+        _increaseCurrentAndNextBalance(_activeStakeByOwner[staker], amount);
```
jalextowle: Nice! `increase` and `decrease` are definitely more accurate than `increment` and `decrement`.

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -57,89 +56,23 @@ contract MixinCumulativeRewards is
...
-        uint256 currentMostRecentEpoch,
-        uint256 newMostRecentEpoch
-    )
-        internal
-    {
-        // Sanity check that we're not trying to go back in time
-        assert(newMostRecentEpoch >= currentMostRecentEpoch);
-        _cumulativeRewardsByPoolLastStored[poolId] = newMostRecentEpoch;
+        uint256 currentMostRecentEpoch = _cumulativeRewardsByPoolLastStored[poolId];
+        assert(currentEpoch_ >= currentMostRecentEpoch);
```
jalextowle: 🙏

contracts/staking/test/utils/cumulative_reward_tracking_simulation.ts
```diff
@@ -36,12 +36,7 @@ export class CumulativeRewardTrackingSimulation {
     private static _extractTestLogs(txReceiptLogs: DecodedLogs): TestLog[] {
         const logs = [];
         for (const log of txReceiptLogs) {
-            if (log.event === TestCumulativeRewardTrackingEvents.SetMostRecentCumulativeReward) {
```
hysz: Why are we removing this?

dorothy-zbornak: We inlined `_forceSetMostRecentBlahBlah()` so it's not hookable anymore.

PR #2212, opened 2019-09-26
Remove domain from Order and remove OrderWithoutDomain type

fragosti:
## Description

- Changes the `Order` type by removing the `domain` field and adding `exchangeAddress` and `chainId`.
- Removes the `OrderWithoutDomain` type entirely.

## Checklist:


contracts/exchange-libs/test/lib_fill_results.ts
```diff
@@ -536,18 +538,16 @@ blockchainTests('LibFillResults', env => {
             feeRecipientAddress: randomAddress(),
             expirationTimeSeconds: randomUint256(),
             salt: randomUint256(),
-            domain: {
-                verifyingContract: constants.NULL_ADDRESS,
-                chainId: 1337, // The chain id for the isolated exchange
-            },
+            exchangeAddress: constants.NULL_ADDRESS,
+            chainId: 1337, // The chain id for the isolated exchange
```
xianny: 😆

abandeali1:
@xianny we don't need to pass these in to the `Exchange` contract at all, so no changes are necessary there.

abandeali1:
Any idea why tests are failing?

fragosti:
@abandeali1 about to look into them. Also just wanted to confirm that removing the `OrdersWithoutDomain` type is safe. Seems like the code that creates the order hash uses an order schema to create the order hash (and removes other fields), and that the exchange wrappers and other wrappers have access to the ABI, and does not encode fields not in the ABI.

packages/json-schemas/schemas/order_schema.json
```diff
@@ -1,21 +1,54 @@
...
+            "$ref": "/wholeNumberSchema"
+        },
+        "feeRecipientAddress": {
+            "$ref": "/addressSchema"
+        },
+        "expirationTimeSeconds": {
+            "$ref": "/wholeNumberSchema"
+        },
+        "chainId": {
+            "type": "number "
```
feuGeneA: When running this through the Python tooling: `UNEXPECTED EXCEPTION: <SchemaError: "'number ' is not valid under any of the given schemas">`
```suggestion
            "type": "number"
```

contracts/exchange/test/utils/isolated_exchange_wrapper.ts
```diff
@@ -28,7 +28,7 @@ export interface IsolatedExchangeEvents {
     transferFromCalls: DispatchTransferFromCallArgs[];
 }
 
-export type Order = OrderWithoutDomain;
+export type Order = Order;
```
dekz: This need to be re-exported like this anymore?

fragosti: It's a testing utility class. I figured it was good to keep the abstraction in place.

PR #2216, opened 2019-09-27
Forwarder protocol fees

mzhu25:
## Description
Adds protocol fee support to Forwarder and updates tests accordingly. Also includes some cosmetic changes across a few files including:
- `expect` messages in `ForwarderTestFactory`
- Renaming some forwarder rich reverts
- Removing an unnecessary parameter in the `ExchangeWrapper` constructor

## Types of changes

## Checklist:


mzhu25:
> [Coverage Status](https://coveralls.io/builds/25965067)
> 
> Coverage decreased (-2.9%) to 72.355% when pulling **[46ff31a](https://github.com/0xProject/0x-monorepo/commit/46ff31a88b13e692db7f783e4d011e2a1bf46cca) on feature/3.0/forwarder-protocol-fees** into **[c9e03f7](https://github.com/0xProject/0x-monorepo/commit/c9e03f7d424329295cf86df709031f1fa28d8fac) on 3.0**.

🙄

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -98,29 +85,34 @@ contract MixinExchangeWrapper is
             uint256 makerAssetAcquiredAmount
         )
     {
-        // No fee or percentage fee
+        uint256 protocolFee = tx.gasprice.safeMul(EXCHANGE.protocolFeeMultiplier());
```
abandeali1: We shouldn't have to calculate the protocol fee for every order, since it will be the same for the duration of a transaction. Can it just be passed in as a parameter instead? Otherwise we'll be paying for an additional call + sload for every order.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -98,29 +85,34 @@ contract MixinExchangeWrapper is
...
+
+        // No taker fee or percentage fee
         if (order.takerFee == 0 || order.takerFeeAssetData.equals(order.makerAssetData)) {
             // Attempt to sell the remaining amount of WETH
             LibFillResults.FillResults memory singleFillResults = _fillOrderNoThrow(
                 order,
-                remainingTakerAssetFillAmount,
+                remainingTakerAssetFillAmount.safeSub(protocolFee),
```
abandeali1: Why do we want to change the amount we're filling?

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -132,6 +124,8 @@ contract MixinExchangeWrapper is
             // WETH is also spent on the taker fee, so we add it here.
             wethSpentAmount = singleFillResults.takerAssetFilledAmount.safeAdd(
                 singleFillResults.takerFeePaid
+            ).safeAdd(
```
abandeali1: Styling nit:

```
wethSpentAmount = singleFillResults.takerAssetFilledAmount
    .safeAdd(singleFillResults.takerFeePaid)
    .safeAdd(singleFillResults.protocolFeePaid);
```

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -231,9 +225,11 @@ contract MixinExchangeWrapper is
                 signature
             );
 
-            // WETH is also spent on the taker fee, so we add it here.
+            // WETH is also spent on the protocol and taker fees, so we add it here.
             wethSpentAmount = singleFillResults.takerAssetFilledAmount.safeAdd(
                 singleFillResults.takerFeePaid
+            ).safeAdd(
```
abandeali1: Same styling as above.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -98,29 +85,34 @@ contract MixinExchangeWrapper is
...
         } else if (order.takerFeeAssetData.equals(order.takerAssetData)) {
+
             // We will first sell WETH as the takerAsset, then use it to pay the takerFee.
-            // This ensures that we reserve enough to pay the fee.
+            // This ensures that we reserve enough to pay the taker and protocol fees.
             uint256 takerAssetFillAmount = LibMath.getPartialAmountCeil(
                 order.takerAssetAmount,
                 order.takerAssetAmount.safeAdd(order.takerFee),
-                remainingTakerAssetFillAmount
+                remainingTakerAssetFillAmount.safeSub(protocolFee)
```
abandeali1: I'm not sure we actually need to protect the user from not having enough to pay the protocol fee. If they did send in enough ETH to cover it with the transaction, then this logic ends up under filling the orders.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
abandeali1: After reading the logic below, I think we can completely remove this and assume the user sent enough ETH to cover the protocol fee.

mzhu25: We're wrapping all the ETH they send, so the protocol fee is always going to be paid in WETH. I think it would complicate the logic to pay the protocol fee in ETH because we don't know how many orders will be filled a priori

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
abandeali1: Sorry for the thrash on this one, but I think it would actually be cleaner to just do this calculation before calling `_marketSellSingleOrder`, rather than passing in the `protocolFee`. `remainingTakerAssetFillAmount` is always used in this way, so might as well combine them.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -57,25 +57,12 @@ contract MixinExchangeWrapper is
...
-            )
-            if success {
-                mstore(fillResults, mload(fillOrderCalldata))
-                mstore(add(fillResults, 32), mload(add(fillOrderCalldata, 32)))
-                mstore(add(fillResults, 64), mload(add(fillOrderCalldata, 64)))
-                mstore(add(fillResults, 96), mload(add(fillOrderCalldata, 96)))
-            }
+        (bool didSucceed, bytes memory returnData) = exchange.call(fillOrderCalldata);
+        if (didSucceed) {
+            assert(returnData.length == 160);
```
jalextowle: 👍

contracts/exchange-forwarder/test/utils/forwarder_test_factory.ts
```diff
@@ -2,147 +2,51 @@ import { ERC20Wrapper } from '@0x/contracts-asset-proxy';
...
-
-    return currentState;
-}
-
 // Since bignumber is not compatible with chai's within
-function expectBalanceWithin(balance: BigNumber, low: BigNumber, high: BigNumber): void {
-    expect(balance).to.be.bignumber.gte(low);
-    expect(balance).to.be.bignumber.lte(high);
+function expectBalanceWithin(balance: BigNumber, low: BigNumber, high: BigNumber, message?: string): void {
+    expect(balance, message).to.be.bignumber.gte(low);
```
jalextowle: Readability level 💯

contracts/exchange-forwarder/test/utils/forwarder_test_factory.ts
```diff
@@ -268,10 +167,9 @@ export class ForwarderTestFactory {
             await expect(tx).to.revertWith(options.revertError);
         } else {
             const gasUsed = (await tx).gasUsed;
-            const ordersInfoAfter = await Promise.all(
-                orders.map(order => this._exchangeWrapper.getOrderInfoAsync(order)),
+            const orderStatusesAfter = await Promise.all(
```
jalextowle: 🙏

PR #2220, opened 2019-09-27
ERC20BridgeProxy

dorothy-zbornak:
## Description

Say hello to the new `ERC20BridgeProxy`!

### Overview
The `ERC20BridgeProxy` essentially does just two things:

1. Call `withdrawTo()` on an instance of the `IERC20Bridge` contract.
2. Ensure that the token balance of `to` has increased by `amount` after the call.

It is up to the `IERC20Bridge` contract to actually perform the transfer. If the bridge contract needs the taker funds in order to complete the transfer (like for Eth2Dai/Uniswap bridges), the order should be constructed with `maker = IERC20Bridge`.

### AssetData
```solidity
abi.encodeWithSelector(
    // bytes4(keccak256("ERC20BridgeProxy(address,address,bytes)"))
    ERC20BridgeProxy(0).PROXY_ID,
    // The address of the ERC20 token that will be transferred from `from` to `to`
    address tokenAddress,
    // The address of an instance of the `IERC20Bridge` contract that will handle
    // the transfer.
    address bridgeAddress,
    // Arbitrary data needed by the bridge contract.
    bytes bridgeData
)
```

### IERC20Bridge
A valid bridge contract only has to export one function:
```solidity
/// @dev Transfers `amount` of the ERC20 `tokenAddress` from `from` to `to`.
/// @param tokenAddress The address of the ERC20 token to transfer.
/// @param from Address to transfer asset from.
/// @param to Address to transfer asset to.
/// @param amount Amount of asset to transfer.
/// @param bridgeData Arbitrary asset data needed by the bridge contract.
/// @return success The magic bytes `0xb5d40d78` if successful.
function transfer(
    address tokenAddress,
    address from,
    address to,
    uint256 amount,
    bytes calldata bridgeData
)
    external
    returns (bytes4 success);
```

## Checklist:


contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
```diff
@@ -0,0 +1,128 @@
...
+
+contract ERC20BridgeProxy is
+    IAssetProxy,
+    Authorizable
+{
+    using LibBytes for bytes;
+    using LibSafeMath for uint256;
+
+    // @dev Result of a successful bridge call.
+    bytes4 constant public BRIDGE_SUCCESS = 0xb5d40d78;
```
abandeali1: How was this derived?

contracts/asset-proxy/contracts/src/interfaces/IERC20Bridge.sol
```diff
@@ -0,0 +1,40 @@
...
+contract IERC20Bridge {
+
+    /// @dev Transfers `amount` of the ERC20 `tokenAddress` from `from` to `to`.
+    /// @param bridgeData Arbitrary asset data needed by the bridge contract.
+    /// @param tokenAddress The address of the ERC20 token to transfer.
+    /// @param from Address to transfer asset from.
+    /// @param to Address to transfer asset to.
+    /// @param amount Amount of asset to transfer.
+    /// @return success The magic bytes `0xb5d40d78` if successful.
+    function transfer(
```
abandeali1: Some nits about the signature of this function:

`transferFrom` is more descriptive than `transfer`. It's somewhat awkward that this would be the 3rd different `transferFrom` used in the proxies, so maybe there is a better name... but I don't love `transfer`.

I also think `bridgeData` should be the last parameter. It is optional and probably won't be used in all bridge contracts. Making it the last parameter is consistent with all of the token standards that have a similar parameter in their own transfer functions.

contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
```diff
@@ -0,0 +1,128 @@
...
+            tokenAddress,
+            from,
+            to,
+            amount
+        );
+        // Bridge must return the magic bytes to indicate success.
+        require(success == BRIDGE_SUCCESS, "BRIDGE_FAILED");
+        // Ensure that the balance of `to` has increased by at least `amount`.
+        require(
+            balanceBefore.safeAdd(amount) <= balanceOf(tokenAddress, to),
```
abandeali1: Nit: I feel like `balanceOf(tokenAddress, to) >= balanceBefore.safeAdd(amount)` is slightly more intuitive. Not a huge deal either way.

contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
```diff
@@ -0,0 +1,128 @@
...
+
+    /// @dev Retrieves the balance of `owner` for this asset.
+    /// @return balance The balance of the ERC20 token being transferred by this
+    ///         asset proxy.
+    function balanceOf(bytes calldata assetData, address owner)
+        external
+        view
+        returns (uint256 balance)
+    {
+        (address tokenAddress, ,) = abi.decode(
```
abandeali1: This can be made more efficient with:

```
(address tokenAddress) = abi.decode(
    assetData.sliceDestructive(4, assetData.length),
    (address)
);
```

dorothy-zbornak: Yeah, I was trying to sneak in some extra asset data validation this way, but I guess since it's not checking the selector, it's not even doing that very well.

contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
dorothy-zbornak: `crypto.randomBytes()`

contracts/asset-proxy/contracts/src/interfaces/IERC20Bridge.sol
dorothy-zbornak: I would like to avoid `transferFrom` because it makes it really hard to explain this pattern to people (`transferFrom` calls `transferFrom` calls `transferFrom`). Maybe `convertFrom`? `exchangeFrom`? `bridgeFrom`? `swapFrom`? `settleFrom`? I'm at a loss.

contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
abandeali1: Probably better to add some meaning to them. Maybe just reuse the `PROXY_ID`?

contracts/asset-proxy/contracts/src/ERC20BridgeProxy.sol
```diff
@@ -0,0 +1,128 @@
...
+    IAssetProxy,
+    Authorizable
+{
+    using LibBytes for bytes;
+    using LibSafeMath for uint256;
+
+    // @dev Result of a successful bridge call.
+    bytes4 constant public BRIDGE_SUCCESS = 0xb5d40d78;
+    // @dev Id of this proxy.
+    //      bytes4(keccak256("ERC20BridgeProxy(address,address,bytes)"))
```
abandeali1: This should probably use `ERC20Bridge` rather than `ERC20BridgeProxy` to be consistent with the other ids. We should also add an entry for this to `IAssetData`.

contracts/asset-proxy/contracts/src/interfaces/IERC20Bridge.sol
abandeali1: `withdrawTo` perhaps?

contracts/asset-proxy/test/erc20bridge_proxy.ts
```diff
@@ -0,0 +1,298 @@
...
+
+import {
+    artifacts,
+    ERC20BridgeProxyContract,
+    TestERC20BridgeBridgeWithdrawToEventArgs,
+    TestERC20BridgeContract,
+} from '../src';
+
+blockchainTests.resets('ERC20BridgeProxy unit tests', env => {
+    const PROXY_ID = '0xdc1600f3';
```
jalextowle: Maybe we should add this to `AssetProxyId` in `packages/types/`

contracts/asset-proxy/test/erc20bridge_proxy.ts
```diff
@@ -0,0 +1,298 @@
...
+                encodeAssetData(_opts.assetData),
+                _opts.from,
+                _opts.to,
+                new BigNumber(_opts.amount),
+                { from: caller },
+            );
+            return (logs as any) as DecodedLogs;
+        }
+
+        it('succeeds if the bridge succeeds and balance increases', async () => {
```
jalextowle: Nit: 
```suggestion
        it('succeeds if the bridge succeeds and balance increases by `amount`', async () => {
```

contracts/asset-proxy/test/erc20bridge_proxy.ts
```diff
@@ -0,0 +1,298 @@
...
+                assetData: createAssetData({
+                    bridgeData: createBridgeData({
+                        revertError,
+                    }),
+                }),
+            });
+            return expect(tx).to.revertWith(revertError);
+        });
+
+        it('fails if balance of `to` increases by `amount - 1`', async () => {
```
jalextowle: Nit: 
```suggestion
        it('fails if balance of `to` increases by less than `amount`', async () => {
```

For consistency with the above tests.

contracts/asset-proxy/test/erc20bridge_proxy.ts
```diff
@@ -0,0 +1,298 @@
...
+                assetData: createAssetData({
+                    bridgeData: createBridgeData({
+                        transferAmount: amount.minus(1),
+                    }),
+                }),
+            });
+            return expect(tx).to.revertWith('BRIDGE_UNDERPAY');
+        });
+
+        it('fails if balance of `to` decreases', async () => {
```
jalextowle: 🙏

contracts/asset-proxy/test/erc20bridge_proxy.ts
```diff
@@ -0,0 +1,298 @@
...
+                assetData: createAssetData({
+                    bridgeData: createBridgeData({
+                        returnData: hexRightPad('0x1'),
+                    }),
+                }),
+            });
+            return expect(tx).to.revertWith('BRIDGE_FAILED');
+        });
+
+        it('fails if bridge is an EOA', async () => {
```
jalextowle: Nice!

PR #2221, opened 2019-09-28
Eth2DaiBridge

dorothy-zbornak:
## Description

Here's our first ever `ERC20BridgeProxy` (#2220) bridge: `Eth2DaiBridge`!.

### Overview
This bridge contract is designed to be the `makerAddress` in all transactions, so:
1. Taker tokens get transferred to this contract.
2. `ERC20BridgeProxy` calls this contract, which:
    1. Tries to sell its entire WETH or DAI balance to Eth2Dai.
    2. Transfers everything it bought to the taker.

Because this contract should be the `makerAddress` of the order, it must sign for itself using the `Wallet` signature type. It implements the `isValidSignature()` callback but always succeeds.

## Testing instructions

There are a *whole 8 tests* for this package. There really isn't much to unit test since the balance checks are enforced by the `ERC20BridgeProxy` and we just rely on `Eth2Dai` to do the actual conversion.

## Checklist:


contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -0,0 +1,130 @@
...
+import "./ERC20Bridge.sol";
+import "../interfaces/IEth2Dai.sol";
+
+
+// solhint-disable space-after-comma
+contract Eth2DaiBridge is
+    ERC20Bridge,
+    IWallet
+{
+    bytes4 private constant LEGACY_WALLET_MAGIC_VALUE = 0xb0671381;
```
abandeali1: Maybe we should just include this in `IWallet.sol` and inherit that?

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -0,0 +1,130 @@
...
+    /* Mainnet addresses */
+    address constant public ETH2DAI_ADDRESS = 0x39755357759cE0d7f32dC8dC45414CCa409AE24e;
+    address constant public WETH_ADDRESS = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
```
abandeali1: Ideally we don't have to hard code any token addresses. It looks like Eth2Dai supports other token pairs in theory, and those should be able to be used as well. The `fromToken` can probably be passed into the `bridgeData`.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -0,0 +1,130 @@
...
+        }
+        // Try to sell all of this contract's `fromToken` balance.
+        uint256 boughtAmount = _getEth2DaiContract().sellAllAmount(
+            address(fromToken),
+            fromToken.balanceOf(address(this)),
+            address(toToken),
+            amount
+        );
+        // Transfer the converted `toToken`s to `to`.
+        toToken.transfer(to, boughtAmount);
```
abandeali1: We should probably check the return value here, just in case a token isn't 100% ERC20 compliant.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
dorothy-zbornak: interfaces can't have/expose constants though, right?

jalextowle: Abstract contracts can though :trollface:, which is what we already use for most interfaces

dorothy-zbornak: nvm, `IWallet` is actually a contract.

dorothy-zbornak: I ended up just making a copy in `asset-proxy` since it would be a circular dep.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -0,0 +1,162 @@
...
+    ///      if we haven't done so already.
+    /// @param spender The spender address.
+    /// @param tokenAddress The token address.
+    function _grantAllowanceForToken(
+        address spender,
+        address tokenAddress
+    )
+        private
+    {
+        mapping (address => bool) storage spenderHasAllowance = _hasAllowance[spender];
```
abandeali1: What happens when the allowance decreases to less than the required amount, but this is still set to true? 

I think that this check actually just ends up being an extra expense most of the time. I'd prefer just calling `IERC20Token(tokenAddress).approve(spender, uint256(-1))` every single time. If an allowance is already set, then this won't cost much extra gas at all. This also allows us to remove the `_hasAllowance` mapping altogether.

contracts/asset-proxy/contracts/src/interfaces/IWallet.sol
```diff
@@ -0,0 +1,38 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+
+contract IWallet {
+
+    bytes4 internal constant LEGACY_WALLET_MAGIC_VALUE = 0xb0671381;
```
abandeali1: I guess we can't import this due to circular dependencies?

dorothy-zbornak: Yep. The alternative is to have `exchange` import this file from the `asset-proxy` package, but that just *feels* wrong.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
dorothy-zbornak: I'm OK with this.

contracts/asset-proxy/contracts/src/interfaces/IWallet.sol
abandeali1: We should probably create a task for moving this to `exchange-libs`, but doesn't have to be in this PR.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -0,0 +1,154 @@
...
+    )
+        external
+        returns (bytes4 success)
+    {
+        // Decode the bridge data to get the `fromTokenAddress`.
+        (address fromTokenAddress) = abi.decode(bridgeData, (address));
+
+        IEth2Dai exchange = _getEth2DaiContract();
+        // Grant an allowance to the exchange to spend `fromTokenAddress` token.
+        _grantAllowanceForToken(address(exchange), fromTokenAddress);
```
abandeali1: Nit: we can probably just remove this function. It's pretty simple, and we also use `IERC20Token(fromTokenAddress).balanceOf` below.

PR #2222, opened 2019-09-29
Remove wethAssetProxy and wethAssetData from staking contracts

abandeali1:
## Description

There was some debate around if the staking contracts should use the existing `ERC20Proxy` or not for paying WETH denominated protocol fees. It's a close call, but I think requiring a separate allowance leads to greater security. It also isn't much of an inconvenience for end users, as I expect the majority of protocol fees will be paid in ETH and will therefore not require an allowance.

If the `ERC20Proxy` is not used, it doesn't really make sense to have a separate proxy at all - users can simply approve the staking proxy to spend their WETH directly. This PR removes the WETH AssetProxy from the staking codebase completely.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/fees/MixinExchangeFees.sol
+++ b/contracts/staking/contracts/src/fees/MixinExchangeFees.sol
@@ -56,11 +56,13 @@ contract MixinExchangeFees is
         // Transfer the protocol fee to this address if it should be paid in
         // WETH.
         if (msg.value == 0) {
-            getWethAssetProxy().transferFrom(
-                getWethAssetData(),
-                payerAddress,
-                address(this),
-                protocolFeePaid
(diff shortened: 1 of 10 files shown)


contracts/staking/contracts/src/fees/MixinExchangeFees.sol
```diff
@@ -56,11 +56,13 @@ contract MixinExchangeFees is
...
-                payerAddress,
-                address(this),
-                protocolFeePaid
+            require(
+                getWethContract().transferFrom(
+                    payerAddress,
+                    address(this),
+                    protocolFeePaid
+                ),
+                "WETH_TRANSFER_FAILED"
```
dorothy-zbornak: :+1:

PR #2223, opened 2019-09-30
Do not authorize owner by default

abandeali1:
## Description

The actual address used to deploy these contracts should not be authorized. The contracts are typically deployed, configured, and then their ownership is transferred to the `AssetProxyOwner`. This PR removes the owner as an authorized address in the `StakingProxy` and the `ZrxVault`.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/StakingProxy.sol
+++ b/contracts/staking/contracts/src/StakingProxy.sol
@@ -42,7 +42,15 @@ contract StakingProxy is
         MixinStorage()
     {
         readOnlyProxy = _readOnlyProxy;
+
+        // Deployer address must be authorized in order to call `init`
+        _addAuthorizedAddress(msg.sender);
+
+        // Attach the staking contract and initialize state
         _attachStakingContract(_stakingContract);
+
+        // Remove the sender as an authorized address
+        _removeAuthorizedAddressAtIndex(msg.sender, 0);
     }
 
     /// @dev Delegates calls to the staking contract, if it is set.
--- a/contracts/staking/contracts/src/ZrxVault.sol
+++ b/contracts/staking/contracts/src/ZrxVault.sol
@@ -62,8 +62,6 @@ contract ZrxVault is
         public
         Authorizable()
     {
-        _addAuthorizedAddress(owner);
-
         zrxAssetProxy = IAssetProxy(_zrxProxyAddress);
         _zrxToken = IERC20Token(_zrxTokenAddress);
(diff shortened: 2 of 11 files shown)


contracts/staking/contracts/src/StakingProxy.sol
```diff
@@ -42,7 +42,15 @@ contract StakingProxy is
         MixinStorage()
     {
         readOnlyProxy = _readOnlyProxy;
+
+        // Deployer address must be authorized in order to call `init`
+        _addAuthorizedAddress(msg.sender);
```
jalextowle: I'm actually okay with this pattern, but it does seem like it would be more ergonomic to just prevent `init()` from being called when params are set than to use `onlyAuthorized` here. This is definitely better than just authorizing the owner by default though.

contracts/staking/contracts/src/ZrxVault.sol
```diff
@@ -62,8 +62,6 @@ contract ZrxVault is
         public
         Authorizable()
     {
-        _addAuthorizedAddress(owner);
```
jalextowle: 🙏

contracts/staking/contracts/test/TestDelegatorRewards.sol
```diff
@@ -40,7 +40,9 @@ contract TestDelegatorRewards is
     }
 
     constructor() public {
+        _addAuthorizedAddress(msg.sender);
```
jalextowle: Especially after seeing this, preventing non-fresh initialization may be the move. This said, it would be one more thing to consider when re-deploying the staking logic contract, which may be something to weigh for this decision.

contracts/staking/test/unit_tests/exchange_test.ts
```diff
@@ -142,20 +128,6 @@ blockchainTests.resets('Exchange Unit Tests', env => {
             return expect(tx).to.revertWith(expectedError);
         });
 
-        it('should successfully remove an exchange if called by the (authorized) owner', async () => {
```
jalextowle: Thanks for fixing these 👍. I think it would make sense to add them back in though, so that we have test coverage to prevent the owner from being as an authorized address in the contracts.

contracts/staking/contracts/src/StakingProxy.sol
abandeali1: Discussed offline, and we are keeping this pattern the same.

PR #2224, opened 2019-09-30
3.0 testnet migrations

abandeali1:
## Description

This PR:

- Adds a `getSelector` method to each function in the auto-generated wrappers and regenerates all of the existing wrappers.
- Adds a new migration script to migrate the testnets from 2.0 to 3.0. This is pretty different than running a migration from scratch (that script did _not_ change).
- Updates `contract-addresses` with all of the new testnet addresses.

TODOs (likely in another PR):
- Update `test_contract_configs.ts` to verify the new configurations.
- Update `migration.ts` to do a fresh 3.0 deployment with proper configurations (this might break other packages).

## Checklist:


packages/contract-addresses/src/index.ts
```diff
@@ -6,6 +6,7 @@ export interface ContractAddresses {
     zrxToken: string;
     etherToken: string;
     exchange: string;
+    exchange_v3: string;
```
albrow: Convention in TypeScript is to use `camelCase` instead of `under_scores`. Maybe we could call this `exchangeV3`?

albrow: Sure, I'd be in favor of removing it.

abandeali1: I think we should leave the old exchange address as `exchangeV2`. We'll still need it for mainnet deployments, and v2 will probably continue to be live for ~3 months after v3 is deployed.

PR #2225, opened 2019-09-30
Feature/add dist tag to publish

fabioberger:
## Description

This PR adds the ability to specify an [npm dist tag](https://docs.npmjs.com/cli/dist-tag) when publishing the monorepo packages. This will allow us to publish the `3.0` packages under the `3.0` tag so that devs that install our packages using `npm install @0x/package-name` receive the `v2.1` package, and those using `npm install @0x/package-name@3.0` will receive the `3.0` pre-release package version.

## Checklist:


albrow:
@fabioberger since some packages already have a version `3.0` we'll need to use a different name for the dist tag. See https://yarnpkg.com/lang/en/docs/cli/tag/#toc-caveats.

For example:

```bash
# Installs version 3.0 of order-utils (an older version)
yarn add @0x/order-utils@3.0
```

How about something like `protocolV3`?

fabioberger:
Good point. Changed the tag to `protocolV3`

PR #2226, opened 2019-10-01
Final RC publish prep

fabioberger:
## Description

This PR:

- Updates the contract addresses from the most recent v3 testnet deploys as well as the new Ganache snapshot.
- Makes sure every package Lerna will publish has a corresponding CHANGELOG entry that specifies a `X.X.X-beta.0`-style version to be used when published. This is necessary because if no entry exists, the default is a simple patch update to the most recent version.
- Nit: Modifies `yarn build` to also exclude the coordinator and extension packages

## Checklist:


packages/contract-addresses/src/index.ts
```diff
@@ -64,12 +64,12 @@ const networkToAddresses: { [networkId: number]: ContractAddresses } = {
         devUtils: '0x3e0b46bad8e374e4a110c12b832cb120dbe4a479',
```
albrow: Looks like this [`devUtils` contract](https://ropsten.etherscan.io/address/0x3e0b46bad8e374e4a110c12b832cb120dbe4a479) was deployed 49 days ago so it's probably not compatible with v3. Maybe we could replace these with `NULL_ADDRESS` like we did for mainnet?  I don't think it's a good idea to include addresses that are not v3-compatible for some networks. Even though people aren't risking real money here we could still save developers some potential headache.

PR #2233, opened 2019-10-02
UniswapBridge

dorothy-zbornak:
## Description

Yooo here's the `UniswapBridge` for `ERC20BridgeProxy`!

Similar to the `Eth2DaiBridge` (#2221), this is intended to be the `maker` of the order so it can transfer funds to the Uniswap contract.

Also, just like `Eth2DaiBridge`, this implements the `isValidSignature()` `Wallet` type callback to always succeed.

## Checklist:


contracts/asset-proxy/test/uniswap_bridge.ts
```diff
@@ -0,0 +1,384 @@
...
+                    hexLeftPad(wethTokenAddress),
+                );
+                return expect(tx).to.revertWith('NO_UNISWAP_EXCHANGE_FOR_TOKEN');
+            });
+
+            it('fails if `WETH.deposit()` fails', async () => {
+                const revertReason = 'FOOBAR';
+                const tx = withdrawToAsync({
+                    toTokenAddress: wethTokenAddress,
+                    toTokenRevertReason: revertReason,
```
jalextowle: This pattern is slick

contracts/asset-proxy/test/uniswap_bridge.ts
```diff
@@ -0,0 +1,384 @@
...
+                expect(calls[0].exchange).to.eq(exchangeAddress);
+                expect(calls[0].tokensSold).to.bignumber.eq(opts.fromTokenBalance);
+                expect(calls[0].minTokensBought).to.bignumber.eq(0);
+                expect(calls[0].minEthBought).to.bignumber.eq(0);
+                expect(calls[0].deadline).to.bignumber.eq(blockTime);
+                expect(calls[0].recipient).to.eq(opts.toAddress);
+                expect(calls[0].toTokenAddress).to.eq(opts.toTokenAddress);
+            });
+
+            it('sets allowance for "from" token', async () => {
```
jalextowle: A test that verifies that the allowance is set on subsequent calls would be nice.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+        // Get the weth contract.
+        state.weth = _getWethContract();
+
+        // Convert from WETH to a token.
+        if (fromTokenAddress == address(state.weth)) {
+            // Unwrap the WETH.
+            state.weth.withdraw(state.fromTokenBalance);
+            // Buy as much of `toTokenAddress` token with ETH as possible and
+            // transfer it to `to`.
+            state.exchange.ethToTokenTransferInput.value(state.fromTokenBalance)(
```
jalextowle: Nit: In the case of Weth, it would be cheaper to use `address(this).balance` after unwrapping rather than `IERC20Token(fromTokenAddress).balanceOf(address(this))`

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+                0,
+                // Expires after this block.
+                block.timestamp,
+                // Recipient is `to`.
+                to
+            );
+
+        // Convert from a token to WETH.
+        } else if (toTokenAddress == address(state.weth)) {
+            // Buy as much ETH with `toTokenAddress` token as possible.
```
jalextowle: Nit: I think this should use `fromTokenAddress` instead of `toTokenAddress`.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+        );
+        // Grant an allowance to the exchange.
+        _grantExchangeAllowance(state.exchange);
+        // Get our balance of `fromTokenAddress` token.
+        state.fromTokenBalance = IERC20Token(fromTokenAddress).balanceOf(address(this));
+        // Get the weth contract.
+        state.weth = _getWethContract();
+
+        // Convert from WETH to a token.
+        if (fromTokenAddress == address(state.weth)) {
```
jalextowle: Should refunds be handled here?

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
dorothy-zbornak: But we need to know that number beforehand for the `withdraw()` call?

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
dorothy-zbornak: I think all these functions will consume the entire `fromToken` balance, unless you know otherwise?

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+            // Unwrap the WETH.
+            state.weth.withdraw(state.fromTokenBalance);
+            // Buy as much of `toTokenAddress` token with ETH as possible and
+            // transfer it to `to`.
+            state.exchange.ethToTokenTransferInput.value(state.fromTokenBalance)(
+                // No minimum buy amount.
+                0,
```
abandeali1: Shouldn't the minimum be `amount`?

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+                state.fromTokenBalance,
+                // No minimum buy amount.
+                0,
+                // Expires after this block.
+                block.timestamp
+            );
+            // Wrap the ETH.
+            state.weth.deposit.value(ethBought)();
+            // Transfer the WETH to `to`.
+            IERC20Token(toTokenAddress).transfer(to, ethBought);
```
abandeali1: `weth.transfer` is probably more clear here.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+            );
+
+        // Convert from a token to WETH.
+        } else if (toTokenAddress == address(state.weth)) {
+            // Buy as much ETH with `fromTokenAddress` token as possible.
+            uint256 ethBought = state.exchange.tokenToEthSwapInput(
+                // Sell all tokens we hold.
+                state.fromTokenBalance,
+                // No minimum buy amount.
+                0,
```
abandeali1: This should also be `amount`.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+
+        // Convert from one token to another.
+        } else {
+            // Buy as much `toTokenAddress` token with `fromTokenAddress` token
+            // and transfer it to `to`.
+            state.exchange.tokenToTokenTransferInput(
+                // Sell all tokens we hold.
+                state.fromTokenBalance,
+                // No minimum buy amount.
+                0,
```
abandeali1: Ditto.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+        return IUniswapExchangeFactory(UNISWAP_EXCHANGE_FACTORY_ADDRESS);
+    }
+
+    /// @dev Grants an unlimited allowance to the exchange for its token
+    ///      on behalf of this contract.
+    /// @param exchange The Uniswap token exchange.
+    function _grantExchangeAllowance(IUniswapExchange exchange)
+        private
+    {
+        address tokenAddress = exchange.toTokenAddress();
```
abandeali1: Can we pass in the `toTokenAddress` to save a call here?

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+        external
+        view
+        returns (bytes4 magicValue)
+    {
+        return LEGACY_WALLET_MAGIC_VALUE;
+    }
+
+    /// @dev Overridable way to get the weth contract.
+    /// @return token The WETH contract.
+    function _getWethContract()
```
abandeali1: I kinda like making this public and making the deployment variables private. That's what we do in the staking contracts.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,221 @@
...
+        internal
+        view
+        returns (IEtherToken token)
+    {
+        return IEtherToken(WETH_ADDRESS);
+    }
+
+    /// @dev Overridable way to get the uniswap exchange factory contract.
+    /// @return factory The exchange factory contract.
+    function _getUniswapExchangeFactoryContract()
```
abandeali1: Same as above.

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
dorothy-zbornak: Two things to note:
1. This should just buy as much as possible and the proxy will enforce the actual buy amount after we return.
2. These parameters seems to just be used in `assert` vyper statements after each purchase. These `assert` statements do not provide a reason string, so we'll just get a naked revert. So I think it's actually better to just let the proxy revert with an actual message.

abandeali1: That doesn't seem like a huge issue, since we catch/rethrow an error in the Exchange contract. Using the native `assert` will also save gas on failures (I believe it compiles down to REVERT).

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
```diff
@@ -0,0 +1,229 @@
...
+
+        // Get the  exchange token for the token pair.
+        state.exchangeTokenAddress = _getUniswapExchangeTokenAddressForTokenPair(
+            fromTokenAddress,
+            toTokenAddress
+        );
+        // Get the exchange.
+        state.exchange = _getUniswapExchangeForToken(state.exchangeTokenAddress);
+        // Grant an allowance to the exchange.
+        _grantExchangeAllowance(state.exchange, state.exchangeTokenAddress);
```
abandeali1: Shouldn't this be `_grantExchangeAllowance(state.exchange, fromTokenAddress)`? We also shouldn't need to call this if the `fromTokenAddress` is weth.

dorothy-zbornak: Ah, yeah, that's true. `state.exchangeTokenAddress` will always be `fromTokenAddress`. :+1:

contracts/asset-proxy/contracts/src/bridges/UniswapBridge.sol
jalextowle: I checked, and they do consume all of the `fromToken`.

contracts/asset-proxy/test/uniswap_bridge.ts
```diff
@@ -0,0 +1,365 @@
...
+    TestUniswapBridgeEvents as ContractEvents,
+    TestUniswapBridgeTokenApproveEventArgs as TokenApproveArgs,
+    TestUniswapBridgeTokenToEthSwapInputEventArgs as TokenToEthSwapInputArgs,
+    TestUniswapBridgeTokenToTokenTransferInputEventArgs as TokenToTokenTransferInputArgs,
+    TestUniswapBridgeTokenTransferEventArgs as TokenTransferArgs,
+    TestUniswapBridgeWethDepositEventArgs as WethDepositArgs,
+    TestUniswapBridgeWethWithdrawEventArgs as WethWithdrawArgs,
+} from '../src';
+
+blockchainTests.resets.only('UniswapBridge unit tests', env => {
```
jalextowle: Nit: `only`

PR #2240, opened 2019-10-03
Deployment Manager

jalextowle:
## Description

This PR adds an `integration-tests` package to `contracts/` and adds a `DeploymentManager` class that will deploy and fully configure a set of 0x smart contracts.

## Testing instructions

Run `yarn test` in the `integration-tests` package. I have added tests for the deployment manager and the end-to-end deployment test that was created earlier.

## Types of changes

* Testing change

## Checklist:


contracts/integration-tests/contracts/test/StakingWithTokens.sol
```diff
@@ -0,0 +1,53 @@
...
+        wethAddress = weth;
+    }
+
+    function setZrxVault(address vault)
+        public
+    {
+        zrxVault = vault;
+    }
+
+    function getWethAssetData()
```
abandeali1: Why do we need this function? Also, is there any reason we can't just import `TestStaking` from the staking package instead of using this contract?

contracts/integration-tests/package.json
```diff
@@ -0,0 +1,99 @@
+{
+    "name": "@0x/contracts-integration-tests",
```
abandeali1: Can we change the name of this package to `@0x/contracts-integrations`? I envision this being the package that houses custom bridge contracts, exchange wrappers, and extensions for external contracts.

jalextowle: I actually thought about naming it this. I really like this!

contracts/integration-tests/contracts/test/StakingWithTokens.sol
jalextowle: I originally included it here because this package needed something to be in the `contracts/` directory. I can remove this an add an empty contract until adding the dYdX contracts though.

contracts/integration-tests/test/deployment_manager_test.ts
```diff
@@ -0,0 +1,176 @@
...
+
+        it('should have registered the staking proxy in the zrx vault', async () => {
+            const stakingProxy = await deploymentManager.staking.zrxVault.stakingProxyAddress.callAsync();
+            expect(stakingProxy).to.be.eq(deploymentManager.staking.stakingProxy.address);
+        });
+
+        it('should have correctly set the params', async () => {
+            const params = await deploymentManager.staking.stakingWrapper.getParams.callAsync();
+            expect(params).to.be.deep.eq(
+                [10 * 60 * 60 * 24, (90 * constants.PPM_DENOMINATOR) / 100, 100 * 1e18, 10, 1, 2].map(
```
dorothy-zbornak: Is there a way we can just import these from `@0x/contracts-staking/test/utils/constants.ts` so they're less likely to get out of sync?

contracts/integration-tests/contracts/test/StakingWithTokens.sol
jalextowle: That should work, but on second thought, I will need to add functions for updating `weth` and `zrx`. It actually looks like the staking wrapper is setting these values in the staking logic contract rather than in the proxy, which is where we need the values for this test.

jalextowle: I'll just update `TestStaking`.

PR #2243, opened 2019-10-04
Make wrappers small again

xianny:
> Currently our generated contract wrappers are much too big. This adds to bundlesize bloat and also makes our MD docs much too large, causing size deploys to take over an hour and Algolia to timeout on indexing them. Let’s make them smaller.
-- @fabioberger 

## Description

I'm tagging a long list of reviewers to include everyone who might need to know about these changes. **Protocol team**, please note interface changes to `sendTransactionAsync` and `awaitTransactionSuccessAsync`.

Changes:
* makes `evmExecAsync` protected
* stop generating docs for `deployedBytecode`
* trim `deployedBytecode` so it's undefined unless a contract as pure functions
* only generate `getABIEncodedTransactionData` for non-constant functions, or if in debug mode
* only generate `getABIDecodedTransactionData` and `getABIDecodedReturnData` in debug mode
* remove `validateAndSendTransactionAsync`. Allow to provide a `shouldValidate` optional argument instead
* Introduce a `SendTransactionOpts` type for non-ethereum transaction options:
```
{ 
    shouldValidate?: boolean; // default true
    timeoutMs?: number;
    pollingIntervalMs?: number
}
```

Debug mode is enabled for all `contracts/*` packages as they are not published.

**Results**
0x.js MD file: 849KB --> 509KB
Contract-wrappers MD file: 1.4MB --> 811KB
abi-gen-wrappers `lib` size: 5.0MB -> 3.7 MB (3.8 without trimming bytecode)

## Checklist:


packages/abi-gen/templates/TypeScript/partials/method_abi_helper.handlebars
```diff
@@ -1,3 +1,26 @@
+{{^if constant~}}
+{{^if debug~}}
```
xianny: `if ((NOT constant) AND (NOT debug))`. Because we use all 3 functions `if (debug)`

packages/abi-gen/templates/TypeScript/contract.handlebars
```diff
@@ -48,6 +48,9 @@ export enum {{contractName}}Events {
 // tslint:disable:no-parameter-reassignment
 // tslint:disable-next-line:class-name
 export class {{contractName}}Contract extends BaseContract {
+    /**
+     * @ignore
```
xianny: tells typedoc to skip the next line when generating!

contracts/utils/test/ownable.ts
```diff
@@ -49,7 +49,13 @@ describe('Ownable', () => {
         });
 
         it('should transfer ownership if the specified new owner is not the zero address', async () => {
-            expect(ownable.transferOwnership.sendTransactionAsync(nonOwner, { from: owner })).to.be.fulfilled('');
+            expect(
```
xianny: not sure how this worked before 🤔

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
```diff
@@ -19,42 +19,50 @@ import {
...
 import { BigNumber, classUtils, logUtils, providerUtils } from '@0x/utils';
-import { EventCallback, IndexedFilterValues, SimpleContractArtifact } from '@0x/types';
+import { EventCallback, IndexedFilterValues, SimpleContractArtifact, TxOpts } from '@0x/types';
 import { Web3Wrapper } from '@0x/web3-wrapper';
 import { assert } from '@0x/assert';
 import * as ethers from 'ethers';
 // tslint:enable:no-unused-variable
 
-export type AbiGenDummyEventArgs = AbiGenDummyWithdrawalEventArgs | AbiGenDummySimpleEventEventArgs;
+export type AbiGenDummyEventArgs = AbiGenDummySimpleEventEventArgs | AbiGenDummyWithdrawalEventArgs;
```
feuGeneA: Is this ordering nondeterministic?

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
```diff
@@ -2124,46 +2214,46 @@ export class AbiGenDummyContract extends BaseContract {
...
-            const abiEncoder = self._lookupAbiEncoder('methodReturningMultipleValues()');
+            const abiEncoder = self._lookupAbiEncoder('simpleInputSimpleOutput(uint256)');
             // tslint:disable boolean-naming
-            const abiDecodedReturnData = abiEncoder.strictDecodeReturnValue<[BigNumber, string]>(returnData);
+            const abiDecodedReturnData = abiEncoder.strictDecodeReturnValue<BigNumber>(returnData);
             return abiDecodedReturnData;
         },
     };
-    public methodReturningArrayOfStructs = {
+    public simplePureFunction = {
```
feuGeneA: Is the order of methods in the generated wrapper non-deterministic?  That would be really unfortunate... It seems like there's a ton of "noise" in this diff due simply to a re-ordering of method declarations.  If that's the case, we should consider sorting them before generating them.  Out of scope for this PR of course, but maybe we should make an Issue about it.

packages/base-contract/src/index.ts
```diff
@@ -184,7 +184,7 @@ export class BaseContract {
         }
         return rawEncoded;
     }
-    public async evmExecAsync(input: Buffer): Promise<string> {
+    protected async _evmExecAsync(input: Buffer): Promise<string> {
```
feuGeneA: Is the reason for this just to avoid having documentation for it generated into the markdown files?  Or is there some other reason as well?

contracts/utils/test/ownable.ts
xianny: That makes sense. This is the first time I ran this test locally :)

packages/types/src/index.ts
```diff
@@ -670,6 +670,16 @@ export interface Type {
     tupleElements?: Type[];
 }
 
+/**
+ * shouldValidate: Flag indicating whether the library should make attempts to validate a transaction before
+ * broadcasting it. For example, order has a valid signature, maker has sufficient funds, etc. Default=true.
+ */
+export interface TxOpts {
```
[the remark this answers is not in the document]
xianny: Ya good point. Going with `SendTransactionOpts` (to match `sendTransaction` method name)

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -204,7 +172,11 @@ export class AssetProxyOwnerContract extends BaseContract {
          * @param txData Additional data for transaction
          * @returns The hash of the transaction
          */
-        async sendTransactionAsync(owner: string, txData?: Partial<TxData> | undefined): Promise<string> {
+        async sendTransactionAsync(
+            owner: string,
+            txData?: Partial<TxData> | undefined,
+            opts: TxOpts = { shouldValidate: true },
```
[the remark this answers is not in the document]
xianny: This was also the default behaviour a few months back before we started the whole multi-lang/abi-gen project. Neither @feuGeneA nor myself understood the importance of performing an `eth_call` first so I treated it as an optional step! But have since learned that it's a good idea to always validate before sending a tx.

packages/abi-gen-wrappers/src/generated-wrappers/asset_proxy_owner.ts
```diff
@@ -234,20 +210,19 @@ export class AssetProxyOwnerContract extends BaseContract {
...
             return new PromiseWithTransactionHash<TransactionReceiptWithDecodedLogs>(
                 txHashPromise,
                 (async (): Promise<TransactionReceiptWithDecodedLogs> => {
                     // When the transaction hash resolves, wait for it to be mined.
                     return self._web3Wrapper.awaitTransactionSuccessAsync(
                         await txHashPromise,
-                        pollingIntervalMs,
-                        timeoutMs,
+                        opts.pollingIntervalMs,
+                        opts.timeoutMs,
```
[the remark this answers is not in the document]
xianny: I intentionally didn't change `web3wrapper` because the arguments for `awaitTransactionSuccessAsync(txHash, pollingIntervalMs, timeoutMs)` don't have overlap with `sendTransactionAsync(txData)`. They are pretty different steps.

`awaitTransactionSuccessAsync` in the *contract wrappers* has been updated to take `TxOpts` though.

In the contract-wrapper class, `awaitTransactionSuccessAsync` is a convenience that calls `web3Wrapper.sendTransactionAsync(txData)`, retrieves the txhash, then calls `web3Wrapper.awaitTransactionSuccessAsync(txHash)`. Thus `sendTransactionAsync()` is a subset of `awaitTransactionSuccessAsync()`. I created the `TxOpts` object to 1. further abstract the differences between the two away so a developer can call both with the same arguments, and 2. allow us more flexibility to add or remove optional args later.

packages/abi-gen/src/index.ts
```diff
@@ -44,6 +44,10 @@ const args = yargs
         normalize: true,
         demandOption: true,
     })
+    .option('debug', {
+        describe: 'Enable debug functions',
+        type: 'count',
```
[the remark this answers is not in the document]
xianny: For some reason I thought that would imply `--debug=true` or `--debug=false`, but you are totally right, it works the same. Thank you!

packages/abi-gen/src/index.ts
```diff
@@ -73,12 +77,11 @@ const args = yargs
         default: 'TypeScript',
     })
     .example(
-        "$0 --abis 'src/artifacts/**/*.json' --out 'src/contracts/generated/' --partials 'src/templates/partials/**/*.handlebars' --template 'src/templates/contract.handlebars'",
+        "$0 --abis 'src/artifacts/**/*.json' --out 'src/contracts/generated/' --debug --partials 'src/templates/partials/**/*.handlebars' --template 'src/templates/contract.handlebars'",
```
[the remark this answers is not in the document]
xianny: Hmmmm shouldn't we make the example use all the options so that it provides more complete info? Folks can ignore whatever doesn't apply to them.

packages/base-contract/src/index.ts
xianny: Partly to avoid having docs generated, partly because it's not used anywhere else, so follows the principle of minimum exposed surfaces. 

It doesn't make a whole lotta sense to call it in an external context. It only works for functions defined in the Solidity and available via bytecode; and those should all be accessed by calling `functionName.callAsync` or similar. I have a feeling we should also avoid bypassing those to directly pass in callData even for unit tests because that would be an easy way to create holes in our testing.

packages/abi-gen/src/index.ts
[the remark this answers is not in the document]
feuGeneA: Definitely a matter of opinion.  I think of it as a one-liner "quick start guide", and the easiest way to "get started" is by accepting the defaults.  People can always find usages "in the wild" (eg our other packages) if they want to see more complex examples.  I'm open to having my mind changed though.

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
xianny: It's deterministic but the latest version of sol-compiler uses `soljson-v0.5.12`, whereas the existing artifacts were compiled with `soljson-v0.5.10`. So I assume this has something to do with the ordering.

I actually found an existing method in the dummy contract that let me do what I needed, so I removed the contract changes and reverted the artifact. We should open a separate PR just to update the artifacts though so that future diffs will be clean. I'll open an issue and do it after this one is merged.

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
xianny: see https://github.com/0xProject/0x-monorepo/pull/2243#discussion_r332724650

xianny:
Thanks for looking into the bundle size @feuGeneA ! I agree that the bundle is pretty large right now and a lot of it is due to `ethereumjs-vm`. That's a whole can of worms though, so I'll address other points first:

- tried trimming bytecode as you suggested from contracts without pure functions. This drops .1 MB from the uncompressed lib size
- changed `contract-artifacts` to a devDependency in `abi-gen-wrappers`; not sure how much this changed since I did not bundle it, but assuming it drops `contract-artifacts` from whatever needs to be downloaded from npm
- the changes here so far have changed uncompressed size from 5.0MB to 3.7MB! I updated the OP with this
- debug functions: we *are* actually keeping `getABIEncodedTransactionData` because you are right that it is useful for forwarder contracts. But I *did* remove it from pure/view functions because that doesn't apply in those cases! The debug functions are the two that @hsyz added recently, `getABIDecodedTransactionData` and `getABIDecodedReturnData` and I think those really are very specific for protocol team unit testing cases
- testing setup: Our setup currently tests our specific templates and partials. The `abi-gen` utility doesn't make a whole lot of sense without templates, so I think it's OK to provide a "safe default" and test them together. Having a hard time picturing how to test abi-gen *without* the template or partials but definitely open to ideas. I would also argue that the `debug` flag isn't really a combinatorial scenario but more of a subset/superset, so if we test the superset we're OK.

Re: ethereumjs-vm dependency, yeah it's huge :/ but the previous consensus has been that it's worth it. Maybe we need to revisit this, but that seems out of scope for this PR. From the POV of an external developer I'm also not 100% convinced that I would prefer to use `DevUtilsContract` (with ethereumjs-vm) over pre-existing `@0x/order-utils`. The trade-off is that internally it's much easier for us to maintain. I'm definitely open to having that discussion, but would rather it not block this PR, since the original goal is just to enable us to upload docs!

packages/abi-gen/src/index.ts
```diff
@@ -420,15 +420,17 @@ for (const abiFileName of abiFileNames) {
         return eventData;
     });
 
+    const shouldIncludeBytecode = methodsData.find(methodData => methodData.stateMutability === 'pure') !== undefined;
```
feuGeneA: :heart:

packages/abi-gen/templates/TypeScript/partials/method_tx.handlebars
```diff
@@ -8,6 +8,7 @@
 async sendTransactionAsync(
 {{> typed_params inputs=inputs}}
 txData?: Partial<TxData> | undefined,
+opts: SendTransactionOpts = { shouldValidate: true },
```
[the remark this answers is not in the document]
xianny: That makes complete sense. I kind of forgot about TypeScript type system ^^"

contracts/utils/test/ownable.ts
```diff
@@ -49,7 +49,13 @@ describe('Ownable', () => {
...
+                ownable.transferOwnership.awaitTransactionSuccessAsync(
+                    nonOwner,
+                    { from: owner },
+                    { timeoutMs: constants.AWAIT_TRANSACTION_MINED_MS },
```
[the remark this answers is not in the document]
xianny: Now there is :)

packages/base-contract/src/types.ts
```diff
@@ -27,7 +18,3 @@ export enum SubscriptionErrors {
     SubscriptionNotFound = 'SUBSCRIPTION_NOT_FOUND',
     SubscriptionAlreadyPresent = 'SUBSCRIPTION_ALREADY_PRESENT',
 }
-
-export interface IndexedFilterValues {
-    [index: string]: ContractEventArg;
-}
```
xianny: These were all duplicated in `@0x/types`

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
```diff
@@ -323,7 +329,7 @@ export class AbiGenDummyContract extends BaseContract {
         awaitTransactionSuccessAsync(
             wad: BigNumber,
             txData?: Partial<TxData>,
-            opts: SendTransactionOpts = { shouldValidate: true },
+            opts: AwaitTransactionSuccessOpts = { shouldValidate: true, timeoutMs: 0 },
```
fabioberger: Defaulting `timeoutMs` to 0 doesn't make any sense. From the doc comment, this is: `How long (in ms) to poll for transaction mined until aborting.`. Aborting after 0 MS makes it impossible for the poll request to complete. Am I missing something?

xianny: Ahh good catch. This should be `pollingIntervalMs` and was only set to 0 in contracts unit tests. `@0x/web3wrapper` has a default `pollingIntervalMs = 1000` already so I will leave the wrapper without a default value and keep the existing behaviour.

PR #2245, opened 2019-10-05
Enforce assetData is padded to 32 bytes

abandeali1:
## Checklist:


contracts/exchange/test/dispatcher.ts
```diff
@@ -238,37 +238,35 @@ describe('AssetProxyDispatcher', () => {
             expect(newBalances).to.deep.equal(erc20Balances);
         });
 
-        it('should perform a noop transfer if from == to', async () => {
```
mzhu25: Why are we removing this test?

abandeali1: We actually no longer noop in this case. It wasn't a very good test if it was still passing.

PR #2248, opened 2019-10-05
Simplify staking state

abandeali1:
## Description

This PR makes the following changes in order to simplify some of the accounting in the staking contracts:

- Withdrawable stake is now simply the stake that is inactive in both the current and next epoch. This considerably simplifies the logic for `unstake`. This does, however, actually change the functionality of withdrawals. Stakers can now withdraw immediately at the next epoch after moving their stake to `INACTIVE`. They will still forfeit their rewards for the epoch in which they withdraw, though. Open to hearing people's thoughts on this.

- Completely removes the `ACTIVE` stake status. Being able to think about the state in 1 less dimension removes a lot of complexity IMO. Newly staked ZRX is now `INACTIVE`. The only functional change here is that a user can deposit and withdraw in the same epoch (but will not receive any actual benefits of staking for doing so).

- Introduces an `_ownerStakeByStatus` mapping, rather than having multiple mappings for each state.

- Since there are now only 2 stake statuses, we don't need to store global stake balances for both. We now only store the global delegated stake, and inactive stake is calculated as:

```solidity
uint256 totalStake = getZrxVault().balanceOfVault();
balance.currentEpochBalance = totalStake.safeSub(balance.currentEpochBalance).downcastToUint96();
balance.nextEpochBalance = totalStake.safeSub(balance.nextEpochBalance).downcastToUint96();
```

We also finally are under the codesize limit with full optimization settings. Yay!

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/immutable/MixinStorage.sol
+++ b/contracts/staking/contracts/src/immutable/MixinStorage.sol
@@ -43,7 +43,8 @@ contract MixinStorage is
 
     // mapping from StakeStatus to the total amount of stake in that status for the entire
     // staking system.
-    mapping (uint8 => IStructs.StoredBalance) public globalStakeByStatus;
+    // (access using _loadSyncedBalance or _loadUnsyncedBalance)
+    mapping (uint8 => IStructs.StoredBalance) internal _globalStakeByStatus;
 
     // mapping from StakeStatus to address of staker to stored balance
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
--- a/contracts/staking/contracts/src/immutable/MixinStorage.sol
+++ b/contracts/staking/contracts/src/immutable/MixinStorage.sol
@@ -41,9 +41,11 @@ contract MixinStorage is
     // address for read-only proxy to call
     address public readOnlyProxyCallee;
 
-    // amount of stake currently being delegated in the system
+    // mapping from StakeStatus to gloabl stored balance
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
-    IStructs.StoredBalance internal _globalDelegatedStake;
+    // NOTE: only Status.DELEGATED is used to access this mapping, but this format
+    // is used for extensibility
+    mapping (uint8 => IStructs.StoredBalance) internal _globalStakeByStatus;
 
     // mapping from StakeStatus to address of staker to stored balance
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
--- a/contracts/staking/contracts/src/immutable/MixinStorage.sol
+++ b/contracts/staking/contracts/src/immutable/MixinStorage.sol
@@ -41,10 +41,9 @@ contract MixinStorage is
     // address for read-only proxy to call
     address public readOnlyProxyCallee;
 
-    // mapping from StakeStatus to the total amount of stake in that status for the entire
-    // staking system.
+    // amount of stake currently being delegated in the system
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
-    mapping (uint8 => IStructs.StoredBalance) internal _globalStakeByStatus;
+    IStructs.StoredBalance internal _globalDelegatedStake;
 
     // mapping from StakeStatus to address of staker to stored balance
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
--- a/contracts/staking/contracts/src/immutable/MixinStorage.sol
+++ b/contracts/staking/contracts/src/immutable/MixinStorage.sol
@@ -45,17 +45,9 @@ contract MixinStorage is
     // staking system.
     mapping (uint8 => IStructs.StoredBalance) public globalStakeByStatus;
 
-    // mapping from Owner to Amount of Active Stake
+    // mapping from StakeStatus to address of staker to stored balance
     // (access using _loadSyncedBalance or _loadUnsyncedBalance)
-    mapping (address => IStructs.StoredBalance) internal _activeStakeByOwner;
-
-    // Mapping from Owner to Amount of Inactive Stake
-    // (access using _loadSyncedBalance or _loadUnsyncedBalance)
-    mapping (address => IStructs.StoredBalance) internal _inactiveStakeByOwner;
-
-    // Mapping from Owner to Amount Delegated
-    // (access using _loadSyncedBalance or _loadUnsyncedBalance)
-    mapping (address => IStructs.StoredBalance) internal _delegatedStakeByOwner;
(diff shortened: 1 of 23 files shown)


contracts/staking/contracts/src/immutable/MixinStorage.sol
```diff
@@ -41,21 +41,13 @@ contract MixinStorage is
     // address for read-only proxy to call
     address public readOnlyProxyCallee;
 
-    // mapping from StakeStatus to the total amount of stake in that status for the entire
-    // staking system.
-    mapping (uint8 => IStructs.StoredBalance) public globalStakeByStatus;
```
hysz: It could be good to keep this format for extensibility, in case we add more statuses later.

contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -53,28 +53,19 @@ interface IStructs {
...
-        uint256 currentEpochBalance;
-        uint256 nextEpochBalance;
-    }
-
     /// @dev Statuses that stake can exist in.
+    ///      Any stake can be (re)delegated effective at the next epoch
+    ///      Inactive stake can be withdrawn if it is available in both the current and next epoch
     enum StakeStatus {
-        ACTIVE,
         INACTIVE,
```
hysz: Could possibly even simplify to just delegated & undelegated statuses for now. Also could make sense since stakers will still be able to vote when staked but not delegated.

contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -53,28 +53,19 @@ interface IStructs {
     /// Note that these balances may be stale if the current epoch
     /// is greater than `currentEpoch`.
     /// Always load this struct using _loadSyncedBalance or _loadUnsyncedBalance.
-    /// @param isInitialized
     /// @param currentEpoch the current epoch
     /// @param currentEpochBalance balance in the current epoch.
     /// @param nextEpochBalance balance in `currentEpoch+1`.
     struct StoredBalance {
-        bool isInitialized;
         uint32 currentEpoch;
```
hysz: Optional, but could up `currentEpoch` to `uint64` now that `isInitialized` is no longer needed. It'll still fit into a word.

contracts/staking/contracts/src/stake/MixinStakeBalances.sol
```diff
@@ -29,135 +29,65 @@ contract MixinStakeBalances is
...
-        returns (IStructs.StakeBalance memory balance)
-    {
-        IStructs.StoredBalance memory storedBalance = _loadSyncedBalance(
-            globalStakeByStatus[uint8(IStructs.StakeStatus.INACTIVE)]
-        );
-        return IStructs.StakeBalance({
-            currentEpochBalance: storedBalance.currentEpochBalance,
-            nextEpochBalance: storedBalance.nextEpochBalance
-        });
+        balance = _loadSyncedBalance(_globalDelegatedStake);
```
hysz: Could reduce number of `sload` ops here by moving this into a separate if block.

contracts/staking/test/stake_test.ts
```diff
@@ -324,131 +250,135 @@ blockchainTests.resets('Stake Statuses', env => {
...
-            // Later in Epoch 5: User reactivates a portion of their stake
-            await staker.moveStakeAsync(
-                new StakeInfo(StakeStatus.Inactive),
-                new StakeInfo(StakeStatus.Active),
-                toBaseUnitAmount(1),
-            );
-            // Epoch 6: Status updates (no user intervention required)
-            await staker.goToNextEpochAsync();
-        });
+        // it('Simulation from Staking Spec', async () => {
```
hysz: [reminder to uncomment this before merge]

contracts/staking/contracts/src/stake/MixinStakeBalances.sol
abandeali1: `loadSyncedBalance` copies to memory, so doesn't seem like a huge deal. This function is also never called internally.

contracts/staking/contracts/src/interfaces/IStructs.sol
abandeali1: Probably worth it.

contracts/staking/test/stake_test.ts
abandeali1: Do we actually still need this? I assume that part of the spec is going to change a lot with the changes from this PR.

abandeali1: I'm going to leave this commented out for now. Unclear to me how we will be updating the flow in the spec. The interactions become so much simpler that I'm not even sure we need this test.

contracts/staking/contracts/src/interfaces/IStructs.sol
```diff
@@ -53,28 +53,19 @@ interface IStructs {
     /// Note that these balances may be stale if the current epoch
     /// is greater than `currentEpoch`.
     /// Always load this struct using _loadSyncedBalance or _loadUnsyncedBalance.
-    /// @param isInitialized
     /// @param currentEpoch the current epoch
     /// @param currentEpochBalance balance in the current epoch.
     /// @param nextEpochBalance balance in `currentEpoch+1`.
     struct StoredBalance {
-        bool isInitialized;
```
dorothy-zbornak: Just realized we never use this. :upside_down_face:

contracts/staking/contracts/src/immutable/MixinStorage.sol
abandeali1: I guess it is nice to not have to. Also makes the state layout more consistent if we do add more statuses down the line.

contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -136,31 +128,15 @@ contract MixinStake is
...
-            ? getWithdrawableStake(staker)
-            : 0;
-
         // execute move
-        IStructs.StoredBalance storage fromPtr = _getBalancePtrFromStatus(staker, from.status);
-        IStructs.StoredBalance storage toPtr = _getBalancePtrFromStatus(staker, to.status);
-        _moveStake(fromPtr, toPtr, amount);
-
-        // update global total of stake in the statuses being moved between
+        IStructs.StoredBalance storage fromPtr = _ownerStakeByStatus[uint8(from.status)][staker];
```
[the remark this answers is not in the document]
abandeali1: They can be, but I think it's more readable this way. Should only add a tiny amount of gas for using a new stack variable.

PR #2249, opened 2019-10-06
Read-only mode backstop

abandeali1:
## Description

This creates a `ZrxVaultBackstop` contract that is intended to be added as an authorized address in the `ZrxVault`. The backstop contract can trigger catastrophic failure mode if read only mode has been set for 40 days continuously (this number is debatable and can change depending on what the timelocks end up being).

## Checklist:


contracts/staking/contracts/src/ZrxVaultBackstop.sol
```diff
@@ -0,0 +1,69 @@
...
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-utils/contracts/src/LibSafeMath.sol";
+import "./interfaces/IStructs.sol";
+import "./interfaces/IZrxVault.sol";
+import "./interfaces/IStakingProxy.sol";
+
+
+contract ZrxVaultBackstop {
```
hysz: Optional ~ `ZrxVaultFailsafe`

contracts/staking/contracts/src/interfaces/IStakingProxy.sol
```diff
@@ -17,9 +17,12 @@
 */
 
 pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
 
+import "./IStructs.sol";
 
-interface IStakingProxy /* is IStaking */
+
+contract IStakingProxy /* is IStaking */
```
jalextowle: Now that this is a contract, we could actually inherit `IStaking` here. Is there a reason not to?

contracts/staking/contracts/test/TestStorageLayoutAndConstants.sol
```diff
@@ -118,6 +118,14 @@ contract TestStorageLayoutAndConstants is
             )
             slot := add(slot, 0x1)
 
+            assertSlotAndOffset(
```
jalextowle: 🙏

contracts/staking/test/zrx_vault_backstop_test.ts
```diff
@@ -0,0 +1,50 @@
...
+        it('should revert if read-only mode has been set for less than 40 days', async () => {
+            await stakingApiWrapper.stakingProxyContract.setReadOnlyMode.awaitTransactionSuccessAsync(true, {
+                from: authorizedAddress,
+            });
+            const expectedError = new StringRevertError('READ_ONLY_MODE_DURATION_TOO_SHORT');
+            expect(
+                stakingApiWrapper.zrxVaultBackstopContract.enterCatastrophicFailureIfProlongedReadOnlyMode.awaitTransactionSuccessAsync(),
+            ).to.revertWith(expectedError);
+        });
+        it('should enter catastophic failure mode if read-only mode has been set for 40 days', async () => {
```
jalextowle: Optional Nit: A test that calls this function twice would be nice.

contracts/staking/contracts/src/ZrxVaultBackstop.sol
```diff
@@ -0,0 +1,69 @@
...
+    using LibSafeMath for uint256;
+
+    IStakingProxy public stakingProxy;
+    IZrxVault public zrxVault;
+
+    /// @dev Constructor. Sets stakingProxy and zrxVault.
+    /// @param _stakingProxyAddress Address of stakingProxy.
+    /// @param _zrxVaultAddress Address of zrxVault.
+    constructor(
+        address payable _stakingProxyAddress,
```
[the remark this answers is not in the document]
abandeali1: We have to just because the address even has the possibility of receiving value (won't compile without it).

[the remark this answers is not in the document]
jalextowle: Hmm, interesting. I wondered if that was the case.

contracts/staking/test/zrx_vault_backstop_test.ts
abandeali1: We do test for that in the ZRXVault tests.

contracts/staking/contracts/src/ZrxVaultBackstop.sol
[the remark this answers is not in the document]
abandeali1: Actually, I think we can just remove the fallback from the interface?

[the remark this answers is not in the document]
jalextowle: Yep, that should work.

contracts/staking/test/zrx_vault_backstop_test.ts
jalextowle: That's why I made this optional. I think it's nice to make sure that this function is idempotent, but it would be somewhat redundant.

contracts/staking/contracts/src/interfaces/IStakingProxy.sol
abandeali1: Actually we cannot, or the StakingProxy doesn't deploy (unimplemented functions).

PR #2250, opened 2019-10-07
Simplify Makers Interactions with Staking Pools

hysz:
## Description

This simplifies maker interactions with staking pools. The complexity was originally introduced to protect against a theoretical incentive to monopolize pools. Of the guard rails that were built, we are only keeping the 0.9 weighting on delegated stake. If someday exploiting this becomes lucrative and commonplace, then either (i) there will be a better way to enforce it that does not exist today, or (ii) it is unenforceable and we’ll have to rethink some of the economics.

Changes:
- No longer an upper limit on how many makers can be in a pool.
- No longer a handshake for a maker to join a pool.
- No longer any special powers given to makers.
- Pool Id starts at `1` and increments by `1`.

Makers now join a pool by calling the following function:
```
/// @dev Allows caller to join a staking pool as a maker.
/// @param poolId Unique id of pool.
function setMakerStakingPool(bytes32 poolId) public;
```

## Checklist:


contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -111,92 +110,17 @@ contract MixinStakingPool is
...
-            poolId,
-            makerAddress
-        );
-    }
-
-    /// @dev Returns the pool id of the input maker.
-    /// @param makerAddress Address of maker
-    /// @return Pool id, nil if maker is not yet assigned to a pool.
-    function getStakingPoolIdOfMaker(address makerAddress)
+    function setMakerStakingPool(bytes32 poolId)
```
hysz: We could include the `oldPoolId` if we're concerned about a race condition, but it complicates the `createStakingPool` function + we don't expect this to be called often.

abandeali1: I kinda like keeping this names `joinPoolAsMaker`, I think it's a bit more descriptive. Otherwise it's not obvious that the caller should be the maker.

contracts/staking/contracts/src/immutable/MixinStorage.sol
```diff
@@ -69,12 +69,10 @@ contract MixinStorage is
     mapping (address => uint256) internal _withdrawableStakeByOwner;
 
     // tracking Pool Id
-    bytes32 public nextPoolId = INITIAL_POOL_ID;
+    bytes32 public nextPoolId = bytes32(uint256(1));
```
abandeali1: I think what would be a bit cleaner is tracking `lastPoolId` so that it can default to 0. Then when creating a pool, we calculate the new `poolId`, create the pool, and update `lastPoolId`.

contracts/staking/contracts/src/staking_pools/MixinStakingPool.sol
```diff
@@ -276,7 +141,7 @@ contract MixinStakingPool is
         pure
         returns (bytes32)
     {
-        return bytes32(uint256(poolId).safeAdd(POOL_ID_INCREMENT_AMOUNT));
+        return bytes32(uint256(poolId).safeAdd(1));
```
abandeali1: This can probably be inlined in `createStakingPool` (also see my comment on `nextPoolId`)

PR #2252, opened 2019-10-09
Initial Sandstorm Framework

jalextowle:
## Description

This PR implements the basic features of the testing framework needed to build sandstorm as well as some tests that provide examples for how these features can be utilized.

### Address Manager
This class is intended to be an abstraction for address configuration. As of right now, the only functions available are:
- `addMarketMakerAsync`
- `addTakerAsync`

`addMarketMaker` will add a new `MarketMaker` object to the list of market makers. This object will include an `address` field and an `orderFactory` that will be configured to sign orders on behalf of the market maker. Additionally, the market maker will be given an initial balance of tokens in two tokens---a `mainToken` and a `feeToken`. Finally, the market maker will approve the ERC20AssetProxy, allowing the address to sign orders that can be filled in tests.

`addTakerAsync` will save a taker address. This address will be configured with initial balances in a `mainToken` and a `feeToken`. The address approves the ERC20AssetProxy, which allows the address to fill orders in the exchange.

### Function Assertion
A `FunctionAssertion` is generated wrapper function along with a `before` and `after` function. The `before` function can return data that will be passed to the `after` function, and the `after` function will receive the result of calling `callAsync` and the transaction receipt from calling `awaitTransactionSuccessAsync`.

The intended use of a `FunctionAssertion` is to write `before` and `after` functions in such a way that they serve as conditions on the outcome of calling the generated wrapper function with different arguments. These assertions will be used to write integration and fuzz tests by writing pre- and post-conditions as `before` and `after` functions and wrapping generated contract wrapper functions in these conditions.

## Testing instructions

`yarn test`

## Types of changes

## Checklist:


contracts/integrations/test/utils/cache.ts
```diff
@@ -0,0 +1,90 @@
...
+     *                to the fact that we need to serialize the arguments to "callAsync" these
+     *                arguments must be valid arguments to "getABIEncodedTransactionData".
+     * @return Either a cached value or the queried value.
+     */
+    public async callAsync(...args: any[]): Promise<any> {
+        const key = this.getter.getABIEncodedTransactionData(...args);
+        const cachedResult = this.cache[key];
+
+        if (cachedResult !== undefined) {
+            return cachedResult;
```
mzhu25: optional: I feel like there should be a way to update the cache without manually flushing it?

contracts/integrations/test/framework-unit-tests/getter_cache_test.ts
```diff
@@ -0,0 +1,158 @@
...
+        exampleContract = await TestFrameworkContract.deployFrom0xArtifactAsync(
+            artifacts.TestFramework,
+            env.provider,
+            env.txDefaults,
+            artifacts,
+        );
+    });
+
+    describe('callAsync', () => {
+        describe('() => uint', () => {
```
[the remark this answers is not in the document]
jalextowle: I wanted to write all of these tests for my benefit just to ensure that all of the types actually work with the the cache. I'd prefer to keep these tests so that the behavior of the cache doesn't regress across any of these types

contracts/integrations/test/utils/cache.ts
```diff
@@ -0,0 +1,90 @@
...
+        if (getterNames.length !== getters.length) {
+            throw new Error('Mismatched getter names and getters');
+        }
+
+        // Register all of the getters.
+        this.getters = {};
+        for (const getter of _.zip(getterNames, getters)) {
+            this.registerGetter(getter[0] as string, getter[1] as ContractGetterFunction);
+        }
+    }
```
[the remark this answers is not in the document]
jalextowle: Yeah that's a good point

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -0,0 +1,67 @@
...
+    callAsync: (...args: any[]) => Promise<any>;
+    getABIEncodedTransactionData: (...args: any[]) => string;
+}
+
+export interface ContractWrapperFunction extends ContractGetterFunction {
+    awaitTransactionSuccessAsync?: (...args: any[]) => PromiseWithTransactionHash<TransactionReceiptWithDecodedLogs>;
+}
+
+export interface Result {
+    data?: any;
```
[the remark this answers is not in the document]
jalextowle: It will be `undefined` if the wrapper function doesn't return a value (I verified this).

contracts/integrations/test/utils/cache.ts
jalextowle: I agree with this sentiment. As discussed offline, the getter cache was originally intended to fulfill a pretty specific use case (it is intended to cache values between `before` or `after` calls). In this specific case, this type of cache works well; however, a different mechanism would also be good. 

One idea that I have is to actually include the getter cache at a lower level of abstraction in our tools (probably in something like a contract wrappers). In this model, the getter caches could be flushed whenever a relevant stateful function is called. This could potentially use the `GetterCache` implementation from this PR, but it would effectively encapsulate when the cache needs to be flushed.

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,129 @@
...
+            artifacts.TestFramework,
+            env.provider,
+            env.txDefaults,
+            artifacts,
+        );
+    });
+
+    describe('runAsync', () => {
+        it('should call the before function with the provided arguments', async () => {
+            let beforeIntrospectionObject = constants.ZERO_AMOUNT;
```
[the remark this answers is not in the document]
jalextowle: Hmm, yeah that's a bit better.

contracts/integrations/test/examples/simple.ts
```diff
@@ -0,0 +1,75 @@
...
+                    // Ensure that the correct counter was returned.
+                    const counter = await exampleContract.counter.callAsync();
+                    expect(result.data).bignumber.to.be.eq(counter);
+                },
+            };
+            assertion = new FunctionAssertion(exampleContract.numberSideEffect, condition);
+        });
+
+        it('should return the correct counter', async () => {
+            await assertion.runAsync();
```
dorothy-zbornak: So normally I hate this pattern because it isn't very explicit, but because we define the assertions within the `describe` block it isn't so bad, so :+1:

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,125 @@
...
+        it('should call the before function with the provided arguments', async () => {
+            let sideEffectTarget = constants.ZERO_AMOUNT;
+            const assertion = new FunctionAssertion(exampleContract.noEffect, {
+                before: async (inputValue: BigNumber) => {
+                    sideEffectTarget = inputValue;
+                },
+                after: async (beforeInfo: any, result: Result, returnValue: BigNumber) => {},
+            });
+
+            const randomInput = generatePseudoRandom256BitNumber();
```
dorothy-zbornak: Optional: We have a `getRandomInteger()` in `@0x/contracts-test-utils` if you'd rather use that to reduce our dependency on `/package` stuff.

contracts/integrations/contracts/test/TestFramework.sol
```diff
@@ -0,0 +1,64 @@
...
+        external
+    {
+        counter = newCounter;
+    }
+
+    function revertSideEffect(uint256 returnValue)
+        external
+        returns (uint256)
+    {
+        if (counter != 0) {
```
dorothy-zbornak: This kind of confuses me as to why we have these `counter` checks. Naively, I would expect this function to just revert always. What's the reason for this pattern?

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,125 @@
...
+            const assertion = new FunctionAssertion(exampleContract.revertSideEffect, {
+                before: async (inputValue: BigNumber) => {},
+                after: async (beforeInfo: any, result: Result, returnValue: BigNumber) => {
+                    sideEffectTarget = result.data;
+                },
+            });
+            const randomInput = generatePseudoRandom256BitNumber();
+            await assertion.runAsync(randomInput);
+            const errorMessage = 'VM Exception while processing transaction: revert Revert';
+            expect(sideEffectTarget.message).to.be.eq(errorMessage);
```
dorothy-zbornak: `result.data` in this case should be a `RevertError` type, right? I wonder if we can just do `expect(sideEffectTarget).to.revertWith('Revert')` here.

contracts/integrations/test/internal-integrations-tests/exchange_integration_test.ts
```diff
@@ -0,0 +1,135 @@
...
+                    _to: makerAddress,
+                    _value: order.takerAssetAmount,
+                },
+                {
+                    _from: makerAddress,
+                    _to: takers[0],
+                    _value: order.makerAssetAmount,
+                },
+            ]);
+        });
```
dorothy-zbornak: I suspect this is more of a toy example, but we should probably also check that the ETH balance of the staking contract increased when we actually do the Exchange-Staking integration tests.

contracts/integrations/test/utils/cache.ts
mzhu25: I feel like it'd be fine to go with a simple solution (e.g. a flag for `callAsync` to update the cache) until we understand the use cases for this cache better

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
mzhu25: didn't know that existed 🙊

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,125 @@
...
+                    sideEffectTarget = result.data;
+                },
+            });
+            const randomInput = generatePseudoRandom256BitNumber();
+            await assertion.runAsync(randomInput);
+            expect(sideEffectTarget).bignumber.to.be.eq(randomInput);
+        });
+
+        it('should pass the receipt from the function call to "after"', async () => {
+            let sideEffectTarget: TransactionReceiptWithDecodedLogs = {} as TransactionReceiptWithDecodedLogs;
```
mzhu25: isn't this redundant?

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -0,0 +1,64 @@
...
+    constructor(wrapperFunction: ContractWrapperFunction, condition: Condition) {
+        this.condition = condition;
+        this.wrapperFunction = wrapperFunction;
+    }
+
+    /**
+     * Runs the wrapped function and fails if the before or after assertions fail.
+     * @param ...args The args to the contract wrapper function.
+     */
+    public async runAsync(...args: any[]): Promise<void> {
```
mzhu25: optional: As mentioned offline, it could enable some cool patterns to have `runAsync` return `beforeInfo` and `afterInfo`, should we change the return type of `after` to `Promise<any>`.

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
jalextowle: I wasn't able to get that to work because `sideEffectTarget` is not a promise. I'll mess with it a bit, and see if I can figure something out.

contracts/integrations/test/internal-integrations-tests/exchange_integration_test.ts
jalextowle: It was intended to be more of a toy example, but yeah I agree that we should add those assertions in the integration tests.

contracts/integrations/contracts/test/TestFramework.sol
jalextowle: So, I think that it's worth mentioning that this contract does not have a purpose outside of testing the framework (primarily the "GetterCache"). With this in mind, this pattern intends to change the behavior of a given function---in this case it happens to be a revert function---through side effects so that the return value will be different even if the arguments are the same. This is helpful for testing the cache because we expect the old result to be in the cache rather than the new result. 

I'm all ears if there is a simpler way to test this pattern.

contracts/integrations/test/utils/cache.ts
jalextowle: After discussing offline with @dorothy-zbornak, I'm going to remove the `GetterCache` for now since it isn't being used at all. We can add this back in later when we have a use for it (I'm going to keep the commit hash around), by which point I expect that we'll know the right way to flush the cache ergonomically.

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -0,0 +1,64 @@
+import { PromiseWithTransactionHash } from '@0x/base-contract';
+import { TransactionReceiptWithDecodedLogs } from 'ethereum-types';
+
+export interface ContractGetterFunction {
+    callAsync: (...args: any[]) => Promise<any>;
```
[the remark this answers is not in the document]
jalextowle: I initially tried to do that, and things got pretty ugly. This said, I'll give it another shot.

contracts/integrations/test/utils/cache.ts
```diff
@@ -0,0 +1,92 @@
...
+    // A dictionary of getter cache's that allow the user of the collection to reference
+    // the getter functions by name.
+    public getters: GetterCacheSet;
+
+    /**
+     * Constructs a getter collection with pre-seeded getter names and getters.
+     * @param getterNames The names of the getter functions to register.
+     * @param getters The getter functions to register.
+     */
+    constructor(getterNames: string[], getters: ContractGetterFunction[]) {
```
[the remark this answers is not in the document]
jalextowle: Hmm, I thought I had updated this to be the case.

contracts/integrations/test/utils/function_assertions.ts
jalextowle: I'm up for this change. This said, after using the framework some, I haven't had a need for this. I think we should re-evaluate whether being able to access these values is actually a good thing once we've used the framework more (we can access these values in `after` after all).

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,134 @@
...
+    hexRandom,
+} from '@0x/contracts-test-utils';
+import { BigNumber, StringRevertError } from '@0x/utils';
+import { TransactionReceiptWithDecodedLogs } from 'ethereum-types';
+
+import { artifacts, TestFrameworkContract, TestFrameworkEventEventArgs, TestFrameworkEvents } from '../../src';
+import { FunctionAssertion, Result } from '../utils/function_assertions';
+
+const ZERO = constants.ZERO_AMOUNT;
+const MAX_UINT256 = constants.MAX_UINT256;
```
mzhu25: nit: 
```suggestion
const { ZERO, MAX_UINT256 } = constants;
```

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+blockchainTests.resets('FunctionAssertion Unit Tests', env => {
```
mzhu25: I like these tests much better 👍

jalextowle: I thought (and hoped) that would be the case :)

contracts/integrations/test/framework-unit-tests/function_assertion_test.ts
jalextowle: Hmm, I have to use `ZERO_AMOUNT` if I do this restructuring. We should try to rename some of the constants so that they're more succinct.

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export interface Result {
+    data?: any;
+    success: boolean;
+    receipt?: TransactionReceiptWithDecodedLogs;
+}
+
+export interface Condition {
+    before: (...args: any[]) => Promise<any>;
+    after: (beforeInfo: any, result: Result, ...args: any[]) => Promise<any>;
```
dorothy-zbornak: It might be over-engineering to pass the call args into `after`. It also makes these callbacks unnecessarily unwieldy. If we need the call args, the `before` function could just return an object that includes them.

jalextowle: All of the assertions that I've written so far do actually make use of these arguments. This said, I understand your point and I think that it's valid. 

I'm debating whether to make this change now or "wait and see" until we've actually made more use of the framework. I'm not convinced that it would actually be nicer to have to pass the arguments to `after` since I've always used them.

PR #2253, opened 2019-10-09
Ownership Transferred Event

jalextowle:
## Description

This pull request adds an "OwnershipTransferred" event to `IOwnable`, which will be emitted by `Ownable` when the `transferOwnership` function is successfully called. This change addresses auditor feedback regarding event emission in critical functions.

## Testing instructions

`yall`

## Types of changes

## Checklist:

### DIFF
--- a/contracts/utils/contracts/src/interfaces/IOwnable.sol
+++ b/contracts/utils/contracts/src/interfaces/IOwnable.sol
@@ -1,3 +1,21 @@
+/*
+
+  Copyright 2019 ZeroEx Intl.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


contracts/utils/contracts/src/interfaces/IOwnable.sol
```diff
@@ -1,8 +1,32 @@
...
+*/
+
 pragma solidity ^0.5.9;
 
 
 contract IOwnable {
 
+    /// @dev Emitted by Ownable when ownership is transferred.
+    /// @param newOwner The new owner of the contract.
+    event OwnershipTransferred(address newOwner);
```
abandeali1: Looking at Zeppelin's implementation of `Ownable` and they use a slightly different event. Might be worth using the same one to standardize things as much as possible (though I don't feel the need to make their other changes): https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/ownership/Ownable.sol

```solidity
event OwnershipTransferred(address indexed previousOwner, address indexed newOwner);
```

PR #2254, opened 2019-10-09
Add SafeMath to Multisig

jalextowle:
## Description

Previously, the addition `confirmationTimes[transactionId] + secondsTimeLocked` could have overflown in the modifier `pastTimeLock` in `MultiSigWalletWithTimeLock`. This pull request replaces this addition with a call to `_safeAdd` in `SafeMath`.

## Types of changes

## Checklist:


contracts/multisig/contracts/src/MultiSigWalletWithTimeLock.sol
```diff
@@ -18,15 +18,18 @@
...
 import "./MultiSigWallet.sol";
 
 
 /// @title Multisignature wallet with time lock- Allows multiple parties to execute a transaction after a time lock has passed.
 /// @author Amir Bandeali - <amir@0xProject.com>
 // solhint-disable not-rely-on-time
 contract MultiSigWalletWithTimeLock is
-    MultiSigWallet
+    MultiSigWallet,
+    SafeMath
```
mzhu25: I'd prefer `using SafeMath for uint256`

jalextowle: We can't do that in this case since `SafeMath` is a contract instead of a library. I'm up for changing this to a library (`SafeMath` is typically a library); however, it may affect a lot of code.

abandeali1: Is there any reason we can't use `LibSafeMath`?

jalextowle: Nope there isn't. I just wanted to keep the changes small. I'm happy to use `LibSafeMath` instead.

contracts/utils/contracts/src/SafeMath.sol
```diff
@@ -1,90 +0,0 @@
-pragma solidity ^0.5.9;
```
dorothy-zbornak: 🎉🎉

PR #2259, opened 2019-10-10
dYdX v3 integration and tests [1/2]

mzhu25:
## Description
This PR is the first of two which aim to provide a suggested solution for dYdX <> 0x v3 interoperability, as well as dYdX <> 0x integration tests. 
Included changes:
- Contracts from dYdX's [exchange-wrappers](https://github.com/dydxprotocol/exchange-wrappers) package are pulled into `contracts/integrations/contracts/dydx`. 
- `ZeroExV3ExchangeWrapper.sol`, an updated version of dYdX's 0x exchange wrapper which works with v3
- Integration tests for the external functions in `ZeroExV3ExchangeWrapper`
- Updates `DeploymentManager` to handle multiple deployed tokens of each type (ERC20, ERC721, ERC1155). 
- `DydxDeploymentManager` class which handles the deployment of dYdX contracts on top of the deployments handled by `DeploymentManager`

Updating `ZeroExV2MultiOrderExchangeWrapper` and adding corresponding integration tests will be addressed in [2/2]

## Testing instructions

## Types of changes

## Checklist:


contracts/integrations/test/dydx_integration_test.ts
```diff
@@ -0,0 +1,441 @@
...
+
+            return {
+                erc20: erc20Balances,
+                erc721: {},
+                erc1155: {},
+                eth: ethBalances,
+            };
+        }
+
+        function getExpectedBalances(
```
mzhu25: This balance checking stuff really should be DRYed up; I'm working on a PR to use `BalanceStore` in more places, which will address this.

contracts/integrations/test/utils/deployment_mananger.ts
```diff
@@ -441,7 +479,7 @@ export class DeploymentManager {
         };
     }
 
-    private constructor(
+    constructor(
```
jalextowle: When would a user want to call this? The reason I originally made it private was that it would encourage users to just use `deployAsync` instead of rolling their own deployment.

contracts/integrations/test/utils/deployment_mananger.ts
```diff
@@ -396,26 +407,54 @@ export class DeploymentManager {
...
+                    ),
+            ),
         );
-
-        const erc1155 = await ERC1155Contract.deployFrom0xArtifactAsync(
-            ERC1155Artifacts.ERC1155,
-            environment.provider,
-            txDefaults,
-            ERC1155Artifacts,
+        const erc721 = await Promise.all(
```
jalextowle: Nice! Thanks for adding support for an arbitrary number of tokens

contracts/integrations/test/dydx_integration_test.ts
```diff
@@ -0,0 +1,441 @@
...
+            owner,
+            trustedMsgSenders: [solo],
+        });
+
+        [makerToken, takerToken, makerFeeToken, takerFeeToken] = deploymentManager.tokens.erc20;
+        weth = deploymentManager.tokens.weth;
+
+        // Configure order defaults
+        const chainId = await env.getChainIdAsync();
+        const defaultOrderParams = {
```
jalextowle: If my PR gets merged first, you should check out `AddressManager`, which should eliminate a lot of this code. It might even be useful to extend this to a `dYdXAddressManager`.

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,331 @@
...
+            requestedFillAmount,
+            order.takerAssetAmount,
+            order.takerFee
+        );
+
+        if (takerFee == 0) {
+            return;
+        }
+
+        require(
```
jalextowle: Nice!

contracts/integrations/test/utils/deployment_mananger.ts
mzhu25: oh yeah this should be protected –– `DydxDeploymentManager` calls `super(...)` in its constructor

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,331 @@
...
+        // transfer taker fee from trader if applicable
+        transferTakerFee(
+            order,
+            tradeOriginator,
+            requestedFillAmount,
+            takerFeeToken
+        );
+
+        // transfer protocol fee from trader if applicable
+        transferProtocolFee(tradeOriginator);
```
mzhu25: I should also note that the other option is for this contract always hold a WETH balance of its own.
The benefit of such an approach would be that the `tradeOriginator` wouldn't need to approve this contract to transfer WETH on their behalf.
The downsides would be that dYdX would need to periodically top up this contract's WETH balance, and might want to adjust prices to account for the fact that they are paying the protocol fee on the user's behalf (this would probably be easy if they're already doing this for gas)

contracts/integrations/test/utils/deployment_mananger.ts
jalextowle: Ok, I see why you changed it. Yeah, making it `protected` makes sense.

contracts/integrations/test/utils/deployment_mananger.ts
```diff
@@ -396,26 +407,54 @@ export class DeploymentManager {
...
     ): Promise<TokenContracts> {
-        const erc20 = await ERC20TokenContract.deployFrom0xArtifactAsync(
-            ERC20Artifacts.ERC20Token,
-            environment.provider,
-            txDefaults,
-            ERC20Artifacts,
+        const numErc1155TokensToDeploy =
+            options.numErc1155TokensToDeploy || constants.NUM_DUMMY_ERC1155_CONTRACTS_TO_DEPLOY;
+        const numErc20TokensToDeploy = options.numErc20TokensToDeploy || constants.NUM_DUMMY_ERC20_TO_DEPLOY;
+        const numErc721TokensToDeploy = options.numErc721TokensToDeploy || constants.NUM_DUMMY_ERC721_TO_DEPLOY;
```
mzhu25: these should really be `??`
wen TS 3.7 🤔

mzhu25:
re: discussion on Slack, switching to the implementation that assumes the exchange wrapper contract will keep a WETH balance

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+        if (takerFee == 0) {
+            return;
+        }
+
+        require(
+            TRUSTED_MSG_SENDER[msg.sender],
+            "ZeroExV3ExchangeWrapper#transferTakerFee: Only trusted senders can dictate the fee payer"
+        );
+
+        takerFeeToken.transferFrom(
```
abandeali1: So it is assumed that this contract has a WETH balance, correct? Given that this transfer happens after the fill is executed.

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+    ExchangeReader
+{
+    using SafeMath for uint256;
+    using TokenInteract for address;
+    using AdvancedTokenInteract for address;
+
+    // ============ Constants ============
+
+    // maximum gas price, 30 Gwei (choose value for deployment)
+    uint256 constant MAX_GAS_PRICE = 30000000000 wei;
```
abandeali1: I'm wondering how necessary this is. It could be pretty limiting in times of congestions since it can't be upgraded and feels like it doesn't actually help that much if people want to make many trades with smaller gas prices.

If we do leave it in, I think it has to be upgradable.

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+        address weth,
+        address[] memory trustedMsgSenders
+    )
+        public
+    {
+        ZERO_EX_EXCHANGE = zeroExExchange;
+        ZERO_EX_TOKEN_PROXY = zeroExProxy;
+        WETH_TOKEN = weth;
+
+        for (uint256 i = 0; i < trustedMsgSenders.length; i++) {
```
abandeali1: Probably best to make this upgradable in addition to authorizing a bunch of addresses in the constructor.

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+        transferTakerFee(
+            order,
+            tradeOriginator,
+            requestedFillAmount,
+            takerFeeToken
+        );
+
+        // set protocol fee allowance
+        IExchange v3Exchange = IExchange(ZERO_EX_EXCHANGE);
+        address protocolFeeCollector = v3Exchange.protocolFeeCollector();
```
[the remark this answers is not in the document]
mzhu25: @jalextowle suggested exposing an external function that would allow anyone (us, probably) to do this (call `v3Exchange.protocolFeeCollector()`, set the allowance). How do you feel about that approach?

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+        // validate protocol fee
+        validateProtocolFee(fill.protocolFeePaid);
+
+        // set allowance
+        makerToken.ensureAllowance(
+            receiver,
+            fill.makerAssetFilledAmount
+        );
+
+        return fill.makerAssetFilledAmount;
```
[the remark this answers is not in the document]
mzhu25: `validateTradePrice` was only in `ZeroExV2MultiOrderExchangeWrapper`; I think the `assert(fill.takerAssetFilledAmount == requestedFillAmount);` check is basically accomplishing the same thing

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
mzhu25: yes, that's the assumption

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
mzhu25: Yeah, I was considering making it upgradable. I lean towards leaving it in, but we can discuss offline

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
mzhu25: Agreed, just left this as is from the V2 implementation but I'll change it

contracts/integrations/contracts/dydx/ZeroExV3ExchangeWrapper.sol
```diff
@@ -0,0 +1,329 @@
...
+    )
+        external
+        returns (uint256)
+    {
+        // prepare the exchange
+        (LibOrder.Order memory order, address takerFeeToken) = parseOrder(orderData, makerToken, takerToken);
+        bytes memory signature = parseSignature(orderData);
+
+        // transfer taker fee from trader if applicable
+        transferTakerFee(
```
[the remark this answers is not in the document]
mzhu25: I agree –– `ZeroExV2MultiOrderExchangeWrapper` just assumes no fees altogether, but I think this might be a good middle ground. We should bring this up later

PR #2261, opened 2019-10-10
Create @0x/contracts-tests

xianny:
## Description

This PR creates a new package, `@0x/contracts-tests`, to help us get around circular dependencies. With the move to relying on DevUtils for more utility functions, some unit test setups in existing contract packages create cyclic dependencies. 

To be clear, these circular deps are **only in testing**. This [thread](https://github.com/lerna/lerna/issues/1198) suggests that lerna has no interest in making an exception for devDependencies. 🤷‍♀ 

I've managed to resolve circular deps in the past with some creativity, but ran into this hard when refactoring to use DevUtilsContract in place of `@0x/order-utils`. @mzhu25 also expressed a need for this for something he is working on. 

For simplicity, I removed the testing workflow (yarn commands etc) completely from `@0x/contracts-dev-utils` after moving it to `@0x/contracts-tests`. 

* This PR also fixes some dependency declarations in `abi-gen-wrappers/package.json`

## Testing instructions

CI should pass

Please read the README and new config files when reviewing as this PR creates a new package.

## Types of changes

## Checklist:


contracts/tests/package.json
```diff
@@ -0,0 +1,86 @@
...
+    "description": "Unit tests for 0x contracts",
+    "main": "lib/src/index.js",
+    "directories": {
+        "test": "test"
+    },
+    "scripts": {
+        "build": "tsc -b",
+        "build:ci": "yarn build",
+        "pre_build": "run-s compile contracts:gen generate_contract_wrappers",
+        "test": "yarn run_mocha",
```
mzhu25: Could we have this take an argument that specifies which packages's tests to run? In a world where we have all our contract tests in this package, I'd like to be able to do e.g. `CONTRACT=dev-utils yarn test`. Then we could maybe point the `test` script in the `dev-utils` package here to run the tests?

xianny: nice suggestion! I'll do it

contracts/tests/tsconfig.json
```diff
@@ -0,0 +1,5 @@
+{
+    "extends": "../../tsconfig",
+    "compilerOptions": { "outDir": "lib", "rootDir": ".", "resolveJsonModule": true },
+    "include": ["./src/**/*", "./test/**/*", "./generated-wrappers/**/*"]
```
jalextowle: Should this `exclude` `./lib`? It seems like that fixed some issues that I had using `yarn build`

PR #2263, opened 2019-10-14
Staking: Refactor and slightly simplify rewards tracking

dorothy-zbornak:
### Simplifying `MixinStakingPoolRewards` (a bit).

This PR refactors `MixinCumulativeRewards` and `MixinStakingPoolRewards` to better encapsulate the cumulative reward logic within `MixinCumulativeRewards`. 

This is accomplished by exposing two higher level functions:
- `_updateCumulativeReward()`
- `_addCumulativeReward()`

We also hide some low level functions:
- `_forceSetCumulativeRewards()`
- `_getCumulativeRewardAtEpoch()`

As a result, `MixinStakingPoolRewards` no longer has to deal with cumulative rewards directly.

### Removing the need to initialize cumulative rewards when creating a pool.
We now just fall back to an empty CR entry (`0/1`) in `_getCumulativeRewardAtEpoch()`, which means `MixinStakingPool` doesn't have to worry about cumulative rewards. I know, this removes some guard rails, but I think the simplification is worth it. 

### No-op when overwriting a CR.
Because `_withdrawAndSyncDelegatorRewards()` can potentially create CR entries twice (once during finalization and again at the end), we do nothing if a CR already exists at the current epoch in `_addCumulativeReward()`.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
+++ b/contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
@@ -31,21 +31,6 @@ contract MixinCumulativeRewards is
 {
     using LibSafeMath for uint256;
 
-    /// @dev Initializes Cumulative Rewards for a given pool.
-    /// @param poolId Unique id of pool.
-    function _initializeCumulativeRewards(bytes32 poolId)
-        internal
-    {
-        // Sets the default cumulative reward
-        _forceSetCumulativeReward(
-            poolId,
-            IStructs.Fraction({
-                numerator: 0,
-                denominator: MIN_TOKEN_VALUE
-            })
-        );
-    }
-
     /// @dev returns true iff Cumulative Rewards are set
     function _isCumulativeRewardSet(IStructs.Fraction memory cumulativeReward)
         internal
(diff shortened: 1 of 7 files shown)


contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -58,22 +43,71 @@ contract MixinCumulativeRewards is
...
+        if (lastStoredEpoch == currentEpoch_) {
+            return;
+        }
+
+        IStructs.Fraction memory mostRecentCumulativeReward =
+            _cumulativeRewardsByPool[poolId][lastStoredEpoch];
+
+        // Compute new cumulative reward
+        IStructs.Fraction memory cumulativeReward;
+        if (_isCumulativeRewardSet(mostRecentCumulativeReward)) {
```
abandeali1: In what cases is this false? Is this just if cumulative rewards for a pool have never been set?

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
```diff
@@ -58,22 +43,71 @@ contract MixinCumulativeRewards is
...
     }
 
+    /// @dev Sets a pool's cumulative delegator rewards for the current epoch,
+    ///      using the last stored cumulative rewards. If we've already set
+    ///      a CR for this epoch, this is a no-op.
+    /// @param poolId The pool ID.
+    function _updateCumulativeReward(bytes32 poolId)
+        internal
+    {
+        // Just create add an empty CR for this epoch, which will be added to
```
abandeali1: `Just create an empty...`

contracts/staking/contracts/src/staking_pools/MixinCumulativeRewards.sol
dorothy-zbornak: Yeah, because we don't set a CR on pool creation anymore. So if a pool is created, doesn't earn any rewards, then someone delegates, this function will get invoked, but there will be no prior CR. And since `_getCumulativeRewardAtEpoch()` now returns an empty CR if it can't find any, there's no point copying empty CRs to the places it used to look.

PR #2266, opened 2019-10-14
dYdX v3 integration and tests [2/2]

mzhu25:
## Description
Sequel to #2259
Updates `ZeroExV2MultiOrderExchangeWrapper` to work with v3 and corresponding tests in `dydx_integration_test.ts`
Note that the contract upgrade reintroduces the "trusted msg sender" pattern to the multi-order exchange wrapper, since we don't allow untrusted senders to use the contract's WETH. 

## Testing instructions

## Types of changes

## Checklist:


contracts/integrations/contracts/dydx/ZeroExV3MultiOrderExchangeWrapper.sol
```diff
@@ -65,22 +68,35 @@ contract ZeroExV2MultiOrderExchangeWrapper is
...
+        address weth,
+        address[] memory trustedMsgSenders
     )
         public
     {
         ZERO_EX_EXCHANGE = zeroExExchange;
         ZERO_EX_TOKEN_PROXY = zeroExProxy;
+        WETH_TOKEN = weth;
+
+        for (uint256 i = 0; i < trustedMsgSenders.length; i++) {
```
jalextowle: It would probably be good to add a `addTrustedAddress` function so that new addresses can be whitelisted without redeploying this contract.

contracts/integrations/contracts/dydx/ZeroExV3MultiOrderExchangeWrapper.sol
```diff
@@ -112,11 +128,21 @@ contract ZeroExV2MultiOrderExchangeWrapper is
         LibOrder.Order[] memory orders = parseOrders(orderData, makerToken, takerToken);
         bytes[] memory signatures = parseSignatures(orderData);
 
-        // ensure that the exchange can take the takerTokens from this contract
+        // ensure that the 0x exchange can take the takerTokens from this contract
         takerToken.ensureAllowance(ZERO_EX_TOKEN_PROXY, requestedFillAmount);
 
+        // make sure that the protocol fee collector can take WETH from this contract
+        IExchange v3Exchange = IExchange(ZERO_EX_EXCHANGE);
+        address protocolFeeCollector = v3Exchange.protocolFeeCollector();
```
jalextowle: I'm not a fan of the fact that the `protocolFeeCollector` will be approved redundantly here. I think that a better way to handle this approval would be to expose a public endpoint that approves the `protocolFeeCollector` endpoint from the registered exchange, so that the staking contract only needs to be approved once every upgrade.

PR #2270, opened 2019-10-17
@0x/utils: fix wrong RPC method in getChainIdAsync()

feuGeneA:
It was using `net_version`, but it should be using the `eth_chainId` method introduced in EIP-695.  I'm not sure whether/how the network ID differs from the chain ID on mainnet and the testnets, but in Ganache in particular, the network ID is 50 while the chain ID is 1337, and this difference was causing problems for Python tests.  Specifically, the Web3.py interface `Web3.eth.chainId` invokes `eth_chainId`, and the result feeds into the order hash, which wasn't lining up with the non-Python side of things.


dekz:
We're passing in the `chainId` as a parameter to Exchange deployment. So this should've been passing in `50` inside the ganache snapshot. Exchange contract keeps this value around and it doesn't actually represent the executing `chainId`, rather the cached value at deployment time (as a parameter). So it's possible to deploy on mainnet with chainId `1337` even though this shouldn't be possible right?

[Pedro's article](https://medium.com/@pedrouid/chainid-vs-networkid-how-do-they-differ-on-ethereum-eec2ed41635b) on `chainId` versus `networkId` helps understand the real world difference between the two.

dekz:
Should we bump `contract-addresses` to use `1337` over `50` in this PR? Also all of the dependencies relying on lookup by `networkId`? https://github.com/0xProject/0x-monorepo/blob/development/packages/contract-wrappers/src/contract_wrappers.ts#L114

PR #2273, opened 2019-10-18
Coordinator test refactor [1/2] + actor mixin pattern

mzhu25:
## Description
This PR introduces the actor mixin pattern: a lighter-weight, mixin version of the actor classes in `staking`. Each mixin encapsulates a different conceptual "role" in the 0x ecosystem, providing utility functions associated with that role. The mixins can be composed, e.g. `OperatorMaker` in `hybrids.ts`.

This PR also moves Coordinator integration tests to the `integrations` package. The follow-up PR will update the tests to utilize `DeploymentManager`, the new `BalanceStore`s, and the actor mixin pattern. 

Miscellanea also included in this PR:
- keeps `chainId` and `accounts` around as `DeploymentManager` instance variables. They come in handy for the `Maker` mixin
- updates `DeploymentManager` to handle multiple tokens of each standard (copied from the dYdX PR because that one might take a while to merge)
- moves `verifyEvents` into `test-utils`. I think it's p nifty.
- changes the directory structure of `integrations`: 

```
integrations/
    contracts/
    src/
    test/
        actors/
        coordinator/
        deployment/
```

## Testing instructions
The actors work, :trust:

## Types of changes

## Checklist:


contracts/integrations/test/deployment/deployment_mananger.ts
```diff
@@ -433,25 +477,21 @@ export class DeploymentManager {
...
-        staking: StakingContracts,
-        tokens: TokenContracts,
-    ) {
-        this.assetProxies = assetProxies;
-        this.assetProxyOwner = assetProxyOwner;
-        this.exchange = exchange;
-        this.staking = staking;
-        this.tokens = tokens;
-    }
+    constructor(
```
jalextowle: Nit: I still think this should be `protected`.

contracts/integrations/test/actors/pool_operator.ts
```diff
@@ -0,0 +1,60 @@
...
+            );
+
+            const createStakingPoolLog = txReceipt.logs[0];
+            const poolId = (createStakingPoolLog as any).args.poolId;
+            this.poolIds.push(poolId);
+            return poolId;
+        }
+
+        /**
+         * Joins the staking pool specified by the given ID.
```
jalextowle: Nit: 
```suggestion
         * Decreases the operator share of a specified staking pool.
```

contracts/integrations/test/actors/pool_operator.ts
```diff
@@ -0,0 +1,60 @@
...
+import { Actor, ActorConfig, Constructor } from './base';
+
+export interface PoolOperatorConfig extends ActorConfig {
+    operatorShare: number;
+}
+
+export function PoolOperatorMixin<TBase extends Constructor>(Base: TBase) {
+    return class extends Base {
+        public readonly operatorShare: number;
+        public readonly poolIds: string[] = [];
```
jalextowle: It would make more sense to me for this to just be `poolId` since an operator is only the operator of a single pool. Can you explain this design choice?

contracts/integrations/test/actors/pool_operator.ts
```diff
@@ -0,0 +1,60 @@
...
+            const createStakingPoolLog = txReceipt.logs[0];
+            const poolId = (createStakingPoolLog as any).args.poolId;
+            this.poolIds.push(poolId);
+            return poolId;
+        }
+
+        /**
+         * Joins the staking pool specified by the given ID.
+         */
+        public async decreaseOperatorShareAsync(
```
jalextowle: Nit: It probably also makes sense to update `this.operatorShare`.

contracts/integrations/test/actors/maker.ts
```diff
@@ -0,0 +1,52 @@
...
+         * Signs an order (optionally, with custom parameters) as the maker.
+         */
+        public async signOrderAsync(customOrderParams: Partial<Order> = {}): Promise<SignedOrder> {
+            return this.orderFactory.newSignedOrderAsync(customOrderParams);
+        }
+
+        /**
+         * Joins the staking pool specified by the given ID.
+         */
+        public async joinStakingPoolAsync(poolId: string): Promise<TransactionReceiptWithDecodedLogs> {
```
jalextowle: Thought: It might make sense to keep around the `poolId` that the maker has joined. Especially if a lot of makers are being used in a simulation.

contracts/integrations/test/actors/hybrids.ts
```diff
@@ -0,0 +1,5 @@
+import { Actor } from './base';
+import { MakerMixin } from './maker';
+import { PoolOperatorMixin } from './pool_operator';
+
+export const OperatorMaker = PoolOperatorMixin(MakerMixin(Actor));
```
jalextowle: 🆒

contracts/integrations/test/actors/pool_operator.ts
mzhu25: I don't think that's actually the case; a single address can serve as the operator of multiple pools, it just can't be a maker in more than one

contracts/integrations/test/actors/maker.ts
```diff
@@ -0,0 +1,54 @@
...
+    orderConfig: Partial<Order>;
+}
+
+export function MakerMixin<TBase extends Constructor>(Base: TBase) {
+    return class extends Base {
+        public poolId?: string;
+        public readonly actor: Actor;
+        public readonly orderFactory: OrderFactory;
+
+        constructor(...args: any[]) {
```
hysz: Could we document the expected input parameters?

contracts/integrations/test/actors/hybrids.ts
hysz: Yeah this is pretty neat.

mzhu25:
just thought of something: we could keep a static counter in the `Actor` class that gets incremented in the constructor, and then just set `this.address = deployment.accounts[counter]`.
Kind of sketch, kind of cool 🤷‍♂

contracts/integrations/test/actors/pool_operator.ts
jalextowle: Yep, you're completely right.

PR #2274, opened 2019-10-20
Separate finalization from withdrawing rewards

abandeali1:
## Description

- Initializes `currentEpoch` to 1 instead of 0, allowing us to remove all of the `if (currentEpoch == 0)` checks.
- Adds an assertion in `_withdrawAndSyncDelegatorRewards` that the pool has already been finalized in the previous epoch, rather than actually calling `finalizePool`. This drastically simplifies the code path of that function.
- Uses `safeMath` everywhere when possible (even if probably not necessary).

## Checklist:


contracts/staking/test/unit_tests/delegator_reward_test.ts
```diff
@@ -207,7 +207,7 @@ blockchainTests.resets('Delegator rewards unit tests', env => {
     }
 
     async function finalizePoolAsync(poolId: string): Promise<ResultWithTransfers<{}>> {
-        const receipt = await testContract.originalFinalizePool.awaitTransactionSuccessAsync(poolId);
```
dorothy-zbornak: This change actually changes the semantics of the test on [this line](../blob/f55889325bb365b79234cb2657f4470551ef101f/contracts/staking/test/unit_tests/delegator_reward_test.ts#L761). But maybe we don't care about that test.

abandeali1: Hmm, I may be able to preserve the original logic via a `delegatecall`. Had to make this change because `finalizePool` is now `external`.

abandeali1:
@dorothy-zbornak we do not finalize a pool when un/delegating, but we do still withdraw and sync rewards (by calling `__withdrawAndSyncDelegatorRewards`). We could separate that out as well, but it doesn't remove as much complexity IMO.

dorothy-zbornak:
> @dorothy-zbornak we do not finalize a pool when un/delegating, but we do still withdraw and sync rewards (by calling `__withdrawAndSyncDelegatorRewards`). We could separate that out as well, but it doesn't remove as much complexity IMO.

Ah, yeah, I see that now this PR would also cause a revert when moving stake on an unfinalized pool. Yeah, I think this is fine.

PR #2275, opened 2019-10-21
Coordinator test refactor [2/2]

mzhu25:
## Description

Part 2 of #2273 
Utilizes `DeploymentManager`, actor mixins, and `BalanceStore` in the Coordinator integration tests. 
Also implements https://github.com/0xProject/0x-monorepo/pull/2273#issuecomment-543991490 bc I think it's p cool
## Testing instructions

`@PKG=contracts-integrations yarn test`

## Types of changes

## Checklist:


contracts/integrations/test/actors/fee_recipient.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ApprovalFactory, SignedCoordinatorApproval } from '@0x/contracts-coordinator';
+import { SignatureType, SignedZeroExTransaction } from '@0x/types';
+
+import { Actor, ActorConfig, Constructor } from './base';
+
+export interface FeeRecipientConfig extends ActorConfig {
+    verifyingContract?: BaseContract;
+}
+
+export function FeeRecipientMixin<TBase extends Constructor>(Base: TBase) {
```
dorothy-zbornak: Why is this a mixin/decorator? Seems to impose very specific requirements on constructor args *and* the `TBase` type (which should be `Actor`?). This tight coupling smells like a decorator is the wrong pattern here. I kind of feel like this should just be a simple derived class of `Actor` like:

```ts
export class FeeRecipient extends Actor {
    constructor(cfg: FeeRecipientConfig) {
      super(cfg);
      ...
   }
}
````

File: contracts/integrations/test/deployment/deployment_mananger.ts [hunk 2]
```diff
@@ -122,7 +122,9 @@ export interface DeploymentOptions {
 }
 
 export class DeploymentManager {
-    public static protocolFeeMultiplier = new BigNumber(150000);
+    public static readonly protocolFeeMultiplier = new BigNumber(150000);
+    public static readonly gasPrice = new BigNumber(1000000000); // 1 Gwei
```
dorothy-zbornak:
```suggestion
    public static readonly gasPrice = new BigNumber(1e9); // 1 Gwei
```

contracts/integrations/test/actors/fee_recipient.ts
mzhu25: idk if you've seen the mixins in #2273 but the motivation is to be able to mix e.g. `Maker` and `PoolOperator` together. `FeeRecipient` will probably be less commonly used than some of the others, but one could imagine a test in which you'd want a maker that also signs Coordinator fee recipient approvals. We _could_ make FeeRecipient just derive from `Actor`, but I made it a mixin for consistency

contracts/integrations/test/coordinator/coordinator.ts
```diff
@@ -1,419 +1,457 @@
...
+            // Maker -> Fee Recipient
+            localBalanceStore.transferAsset(
+                maker.address,
+                feeRecipient.address,
+                order.makerFee,
+                order.makerFeeAssetData,
             );
+
+            // Protocol fee
+            if (msgValue.isGreaterThanOrEqualTo(DeploymentManager.protocolFee)) {
```
jalextowle: 🙏

PR #2278, opened 2019-10-22
Remove assertion that protocol fee != zero

hysz:
## Description

It is the responsibility of the exchange - not the staking contract - to determine if the protocol fee paid is correct.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
+++ b/contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
@@ -29,10 +29,6 @@ library LibStakingRichErrors {
         CanOnlyDecreaseOperatorShare
     }
 
-    enum ProtocolFeePaymentErrorCodes {
-        MismatchedFeeAndPayment
-    }
-
     enum InitializationErrorCodes {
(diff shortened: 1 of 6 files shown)


packages/order-utils/src/staking_revert_errors.ts
```diff
@@ -144,14 +144,13 @@ export class InvalidParamValueError extends RevertError {
 
 export class InvalidProtocolFeePaymentError extends RevertError {
     constructor(
-        errorCode?: ProtocolFeePaymentErrorCodes,
```
abandeali1: We can remove the `ProtocolFeePaymentErrorCodes` enum too right?

contracts/staking/contracts/src/libs/LibStakingRichErrors.sol
```diff
@@ -29,11 +29,6 @@ library LibStakingRichErrors {
         CanOnlyDecreaseOperatorShare
     }
 
-    enum ProtocolFeePaymentErrorCodes {
```
jalextowle: 🙏

PR #2279, opened 2019-10-22
Assert storage params when calling `MixinParams.setParams()`.

dorothy-zbornak:
## Description

`MixinParams.setParams()` will now do a `IStakingProxy(address(this)).assertValidStorageParams()`. This required flipping `_assertValidStorageParams()` to `public` and some minor test changes.

We could actually do this at a lower level, in `MixinParams._setParams()`, but if the staking contract is attached in the proxy constructor, the proxy code won't be externally callable. I originally had a version that tried to be clever and do an `extcodesize()` check to detect this scenario and skip the call (it gets checked anyway in `StakingProxy._attachStakingContract()`), but I guess we should steer clear of hacky code right now. :smiling_imp:   

## Checklist:


contracts/staking/test/migration_test.ts
```diff
@@ -228,7 +228,7 @@ blockchainTests('Migration tests', env => {
             const expectedError = new StakingRevertErrors.InvalidParamValueError(
                 StakingRevertErrors.InvalidParamValueErrorCodes.InvalidEpochDuration,
             );
-            expect(tx).to.revertWith(expectedError);
+            return expect(tx).to.revertWith(expectedError);
```
dorothy-zbornak: How'd we miss these? :upside_down_face:

contracts/staking/test/migration_test.ts
```diff
@@ -238,21 +238,21 @@ blockchainTests('Migration tests', env => {
...
-            expect(tx).to.be.fulfilled('');
+            return expect(tx).to.be.fulfilled('');
         });
         it('succeeds if epoch duration is 30 days', async () => {
             const tx = proxyContract.setAndAssertParams.awaitTransactionSuccessAsync({
                 ...stakingConstants.DEFAULT_PARAMS,
                 epochDurationInSeconds: thirtyDays,
             });
-            expect(tx).to.be.fulfilled('');
+            return expect(tx).to.be.fulfilled('');
```
jalextowle: Yikes, we missed all of them

PR #2280, opened 2019-10-22
Force no-op when moving stake, in some scenarios

hysz:
## Description

This forces a no-op in the following scenarios:
  i. moving zero stake, or
  ii. moving from undelegated to undelegated.

This fixes C13.

## Checklist:

### DIFF
--- a/contracts/staking/contracts/src/stake/MixinStake.sol
+++ b/contracts/staking/contracts/src/stake/MixinStake.sol
@@ -111,6 +111,17 @@ contract MixinStake is
     {
         address staker = msg.sender;
 
+        // Sanity check: no-op if no stake is being moved.
+        if (amount == 0) {
+            return;
+        }
+
+        // Sanity check: no-op if moving stake from undelegated to undelegated.
+        if (from.status == IStructs.StakeStatus.UNDELEGATED &&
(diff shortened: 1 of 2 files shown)


contracts/staking/contracts/src/stake/MixinStake.sol
```diff
@@ -111,6 +111,17 @@ contract MixinStake is
     {
         address staker = msg.sender;
 
+        // Sanity check: no-op if no stake is being moved.
```
jalextowle: 🙌

contracts/staking/test/unit_tests/stake_test.ts
```diff
@@ -373,18 +373,15 @@ blockchainTests.resets('MixinStake unit tests', env => {
             expect(increaseNextBalanceEvents).to.be.length(0);
         });
 
-        it('moves the owner stake between the same pointer when both are undelegated', async () => {
+        it('does nothing when moving the owner stake from undelegated to undelegated', async () => {
```
jalextowle: A test for the noop if amount == 0 would be nice

hysz: Lol was literally just adding that 😆

PR #2293, opened 2019-10-25
Refactor and move Forwarder tests to integrations package

mzhu25:
## Description
Pretty self-explanatory. `contracts/exchange-forwarder/test/` is no more, `ForwarderWrapper` is ded

## Types of changes

## Checklist:


contracts/integrations/test/actors/base.ts
```diff
@@ -62,6 +64,34 @@ export class Actor {
...
+     * Mints some number of ERC721 NFTs and approves a spender (defaults to the ERC721 asset proxy)
+     * to transfer the token.
+     */
+    public async configureERC721TokenAsync(
+        token: DummyERC721TokenContract,
+        spender?: string,
+        numToMint?: number,
+    ): Promise<BigNumber[]> {
+        const tokenIds: BigNumber[] = [];
+        _.times(numToMint || 1, async () => {
```
dorothy-zbornak: What if we want to mint no tokens?

contracts/exchange-forwarder/package.json
```diff
@@ -14,12 +14,11 @@
         "build:ts": "tsc -b",
         "build:ci": "yarn build",
         "pre_build": "run-s compile contracts:gen generate_contract_wrappers",
-        "test": "yarn run_mocha",
+        "test": "mocha --require source-map-support/register --require make-promises-safe '../integrations/lib/test/forwarder/**/*.js' --timeout 100000 --bail --exit",
```
dorothy-zbornak: won't this mean the forwarder tests will run twice on CI?

contracts/integrations/test/forwarder/forwarder_test_factory.ts
```diff
@@ -0,0 +1,269 @@
+import { BlockchainBalanceStore, LocalBalanceStore } from '@0x/contracts-exchange';
+import { ForwarderContract } from '@0x/contracts-exchange-forwarder';
+import { constants, expect, getPercentageOfValue, OrderStatus } from '@0x/contracts-test-utils';
+import { OrderInfo, SignedOrder } from '@0x/types';
+import { BigNumber, RevertError } from '@0x/utils';
+import * as _ from 'lodash';
```
dorothy-zbornak: Not used?

contracts/exchange-forwarder/package.json
mzhu25: ah yeah, forgot about that

contracts/integrations/test/actors/base.ts
mzhu25: all I want for christmas is nullish coalesce in our TS

contracts/exchange/test/balance_stores/local_balance_store.ts
```diff
@@ -41,6 +41,18 @@ export class LocalBalanceStore extends BalanceStore {
         this._balances.eth[senderAddress] = this._balances.eth[senderAddress].minus(amount);
     }
 
+    /**
+     * Converts some amount of the ETH balance of an address to WETH balance to simulate wrapping ETH.
+     * @param senderAddress Address whose ETH to wrap.
+     * @param amount Amount to wrap.
+     */
+    public wrapEth(senderAddress: string, wethAddress: string, amount: Numberish): void {
```
jalextowle: Very clean 👏

contracts/integrations/test/actors/base.ts
```diff
@@ -62,6 +64,34 @@ export class Actor {
         );
     }
 
+    /**
+     * Mints some number of ERC721 NFTs and approves a spender (defaults to the ERC721 asset proxy)
+     * to transfer the token.
+     */
+    public async configureERC721TokenAsync(
```
jalextowle: 👏

contracts/integrations/test/forwarder/forwarder_test.ts
```diff
@@ -0,0 +1,597 @@
...
+    let nftId: BigNumber;
+    let wethAssetData: string;
+    let makerAssetData: string;
+
+    let maker: Maker;
+    let taker: Actor;
+    let orderFeeRecipient: FeeRecipient;
+    let forwarderFeeRecipient: FeeRecipient;
+
+    before(async () => {
```
jalextowle: Optional: The situation in this `before` is **vastly** better than it was before. This said, there is still a lot of code here---albeit much simpler code than before. I think that we could probably do more work in the constructor of `maker` and `taker` like configuring token balances for some tokens as an option. It's out of scope of this PR, but I'm curious to hear any thoughts that you might have on this.

If we do implement anything like this, we'll need to use a pattern like `deployAsync` from the `DeploymentManager` so that we can use `async/await` syntax.

mzhu25: hmm,, interesting idea –– not sure if the `deployAsync` pattern would play nicely with the mixin pattern, but I'll look into it

PR #2295, opened 2019-10-27
Remove read-only mode and associated code

abandeali1:
## Description

- This PR primarily removes read-only mode from the `StakingProxy` contract. Read-only mode was originally designed to control damage in case any vulnerabilities were discovered. The `ZeroExGovernor` could 1) set ROM, 2) attach a new staking contract that fixes the bug and 3) turn off ROM. However, this seems to necessarily introduce points of centralization into the system and creates an upgrade path where users are no longer able to opt out of the new logic. The types of bugs that may exist in the current system are very isolated and would only result in minimal losses, even in the case of a critical vulnerability. This makes the tradeoff of introducing read-only mode not worth it, IMO.

- This change also allows us to remove the `ZrxVaultBackstop` contract.

- A `detachProtocolFeeCollector` function was added to the `Exchange` contract. The intent is for this function to have a timelock of 0, and calling it would become a part of standard damage control from any bugs found in the staking logic.

- Updates the default alpha from 1/2 to 2/3 (needed before deployment).

## Checklist:


contracts/exchange/test/protocol_fees_manager.ts
```diff
@@ -103,4 +103,37 @@ blockchainTests.resets('MixinProtocolFees', env => {
             expect(updatedEvent.args.updatedProtocolFeeCollector).to.be.eq(protocolFeeCollector);
         });
     });
+
+    blockchainTests.resets('detachProtocolFeeCollector', () => {
```
jalextowle: 🙏

contracts/staking/test/utils/api_wrapper.ts
```diff
@@ -277,23 +263,10 @@ export async function deployAndConfigureContractsAsync(
         artifacts,
     );
 
-    const zrxVaultBackstopContract = await ZrxVaultBackstopContract.deployFrom0xArtifactAsync(
```
jalextowle: Huge RIP

PR #2296, opened 2019-10-28
Update 3.0 testnet migrations and addresses

abandeali1:
## Description

- Productionizes the migration script a bit more in preparation for mainnet deployment
- Updates `contract-addresses` with new testnet addresses and a new schema

## Checklist:


packages/contract-addresses/src/index.ts
```diff
@@ -42,7 +44,8 @@ const networkToAddresses: { [networkId: number]: ContractAddresses } = {
         orderValidator: NULL_ADDRESS,
         zrxToken: '0xe41d2489571d322189246dafa5ebde1f4699f498',
         etherToken: '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2',
-        assetProxyOwner: NULL_ADDRESS,
+        assetProxyOwner: '0xdffe798c7172dd6deb32baee68af322e8f495ce0',
```
jalextowle: Nit: Will this be removed after the full migration?

packages/contract-addresses/src/index.ts
```diff
@@ -53,15 +56,17 @@ const networkToAddresses: { [networkId: number]: ContractAddresses } = {
...
     3: {
         erc20Proxy: '0xb1408f4c245a23c31b98d2c626777d4c0d766caa',
         erc721Proxy: '0xe654aac058bfbf9f83fcaee7793311dd82f6ddb4',
         zrxToken: '0xff67881f8d12f372d91baae9752eb3631ff0ed00',
         etherToken: '0xc778417e063141139fce010982780140aa0cd5ab',
         exchangeV2: '0xbff9493f92a3df4b0429b6d00743b3cfb4c85831',
-        exchange: '0x725bc2f8c85ed0289d3da79cde3125d33fc1d7e6',
-        assetProxyOwner: '0xdcf20f7b447d51f2b3e5499b7f6cbbf7295a5d26',
+        exchange: '0xc56388332ddfc98701fefed94535100c6166956c',
+        assetProxyOwner: NULL_ADDRESS,
```
jalextowle: Ditto

packages/contract-addresses/src/index.ts
```diff
@@ -53,15 +56,17 @@ const networkToAddresses: { [networkId: number]: ContractAddresses } = {
...
     },
     3: {
         erc20Proxy: '0xb1408f4c245a23c31b98d2c626777d4c0d766caa',
         erc721Proxy: '0xe654aac058bfbf9f83fcaee7793311dd82f6ddb4',
         zrxToken: '0xff67881f8d12f372d91baae9752eb3631ff0ed00',
         etherToken: '0xc778417e063141139fce010982780140aa0cd5ab',
         exchangeV2: '0xbff9493f92a3df4b0429b6d00743b3cfb4c85831',
-        exchange: '0x725bc2f8c85ed0289d3da79cde3125d33fc1d7e6',
-        assetProxyOwner: '0xdcf20f7b447d51f2b3e5499b7f6cbbf7295a5d26',
+        exchange: '0xc56388332ddfc98701fefed94535100c6166956c',
```
jalextowle: Nit: I think that noting that this exchange is version 3 of the protocol is worth a comment. It might be a bit confusing to the uninitiated that this exchange is really *exchangeV3* rather than *exchangeV1*.

packages/migrations/src/migration.ts
```diff
@@ -284,6 +284,8 @@ export async function runMigrationsAsync(
         exchange: exchange.address,
         // TODO (xianny): figure out how to deploy AssetProxyOwnerContract
```
jalextowle: Nit: Can this be removed, or are we still waiting for the migration to be updated for the `ZeroExGovernor`?

packages/migrations/src/testnet_migrations.ts
```diff
@@ -148,22 +252,19 @@ export async function runMigrationsAsync(supportedProvider: SupportedProvider, t
...
     await exchange.transferOwnership.awaitTransactionSuccessAsync(governor.address);
     logUtils.log('Exchange configured!');
 
+    logUtils.log('Configuring ERC20BridgeProxy...');
+    await erc20BridgeProxy.transferOwnership.awaitTransactionSuccessAsync(governor.address);
+    logUtils.log('ERC20BridgeProxy configured!');
+
     logUtils.log('Configuring ZrxVault...');
-    await zrxVault.addAuthorizedAddress.awaitTransactionSuccessAsync(governor.address);
+    await zrxVault.transferOwnership.awaitTransactionSuccessAsync(governor.address);
```
jalextowle: Nit: I think we'll still want to add the `governor` as an authorized address. A bunch of functions including `setStakingProxy` and `enterCatastrophicFailureMode` are `onlyAuthorized`. 

Is there a reason why it shouldn't be authorized?

packages/migrations/src/testnet_migrations.ts
```diff
@@ -148,22 +252,19 @@ export async function runMigrationsAsync(supportedProvider: SupportedProvider, t
...
     logUtils.log('ZrxVault configured!');
 
     logUtils.log('Configuring StakingProxy...');
-    await stakingProxy.addAuthorizedAddress.awaitTransactionSuccessAsync(governor.address);
+    await stakingProxy.transferOwnership.awaitTransactionSuccessAsync(governor.address);
```
jalextowle: Ditto

packages/contract-addresses/src/index.ts
abandeali1: Yep, this is still needed for the mainnet migrations.

packages/contract-addresses/src/index.ts
abandeali1: I think the plan was to remove `exchangeV2` after mainnet migrations, but we still need it until then.

packages/migrations/src/migration.ts
abandeali1: Probably. This file actually doesn't look very up to date. I haven't been modifying it except to get it to build.

packages/migrations/src/testnet_migrations.ts
abandeali1: These were moved inside the `ZeroExGovernor` `functionCalls` array to keep all of the configurations in one place. Doesn't really matter which we end up doing

PR #2303, opened 2019-10-29
Removed MixinAbstract

hysz:
## Description

Merged `finalizePool` into MixinStakingPoolRewards and `endEpoch` into MixinScheduler. Unlike #2264 this does not change any logic in the functions; just reorganizes them to remove the cyclic dependency.

This addresses C1.8.

## Checklist:


hysz:
We chatted about this offline and concluded that this change does improve the inheritance graph, but it deviates too far from how the code currently reads. We're going to icebox this task and do a larger refactor for the next version of staking.

PR #2304, opened 2019-10-29
Replace assetDataUtils with DevUtilsContract wherever possible

xianny:
## Description

This PR removes most uses of `assetDataUtils` everywhere and replaces them with `DevUtilsContract`. 

`@0x/instant` hasn't been included in this refactor because I'm not sure how to test it, or how to pass around the contract instance client-side. Will be pinging growth team separately to resolve this.

`@0x/testnet-faucets` also hasn't been included because we are planning to remove it soon. 

`assetDataUtils` hasn't been removed from `@0x/order-utils` yet because of the above exclusions. 

This PR also introduces three new methods in `LibAssetData.sol`:
* encodeStaticCallAssetData
* decodeStaticCallAssetData
* revertIfInvalidAssetData

**Would love a close review on these because I haven't written Solidity before!**

Reviewer hint: Most changes will look like these:
```
const assetData = assetDataUtils.encode<SOMETHING>AssetData(...);
// becomes
const assetData = await devUtils.encode<SOMETHING>AssetData.callAsync(...);
```

Since the method name doesn't change, it'll be un-highlighted in the diff. Most of these changes are pretty quick to scroll through. If you see something that breaks this pattern, that's where it could help to look closer!

## Testing instructions

Tests should pass. 

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -392,6 +402,9 @@ workflows:
             - test-exchange-ganache-3.0:
                   requires:
                       - build
+            - test-integrations-ganache-3.0:
+                  requires:
+                      - build
```
xianny: Not sure why this is showing up as a diff, but I can't get rid of it. This file is identical to the file in `development`.

packages/contract-wrappers/src/utils/assert.ts
```diff
@@ -1,36 +1,13 @@
...
-import { assetDataUtils, signatureUtils } from '@0x/order-utils';
 import { Order } from '@0x/types'; // tslint:disable-line:no-unused-variable
 import { BigNumber } from '@0x/utils'; // tslint:disable-line:no-unused-variable
 import { Web3Wrapper } from '@0x/web3-wrapper';
-import { SupportedProvider } from 'ethereum-types';
 import * as _ from 'lodash';
 
-const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
-
 export const assert = {
```
xianny: The following functions aren't used anywhere. Removed instead of refactoring.

contracts/asset-proxy/test/erc1155_proxy.ts
```diff
@@ -732,18 +736,16 @@ describe('ERC1155Proxy', () => {
...
-                tokensToTransfer,
-                valuesToTransfer,
-                receiverCallbackData,
-            );
-            // remove the function selector and contract address from check, as these change on each test
-            const offsetToTokenIds = 74;
-            const assetDataWithoutContractAddress = assetData.substr(offsetToTokenIds);
-            const expectedAssetDataWithoutContractAddress =
+
+            // hand encode optimized assetData because our tooling (based on LibAssetData.sol/encodeERC1155AssetData) does not use optimized encoding
```
xianny: @hysz can you double check that this makes sense please? 🙏

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -24,9 +25,41 @@ import {
 } from '../../src';
 
 export class FillOrderWrapper {
-    private readonly _exchange: ExchangeContract;
     private readonly _blockchainBalanceStore: BlockchainBalanceStore;
 
+    /**
```
xianny: The following two static methods are not new, but had to be moved up because static members must come first.

contracts/exchange/test/assertion_wrappers/fill_order_wrapper.ts
```diff
@@ -86,39 +119,6 @@ export class FillOrderWrapper {
         return [fillResults, fillEvent, balanceStore];
     }
 
-    /**
```
xianny: moved following two methods up because they are now the only static members.

contracts/test-utils/src/order_factory.ts
```diff
@@ -19,12 +19,12 @@ export class OrderFactory {
         customOrderParams: Partial<Order> = {},
         signatureType: SignatureType = SignatureType.EthSign,
     ): Promise<SignedOrder> {
-        const tenMinutesInSeconds = 10 * 60;
+        const fifteenMinutesInSeconds = 15 * 60;
```
xianny: `contracts/exchange/test/match_orders.ts` was taking too long and some orders were expired before the test suite finished.

contracts/asset-proxy/test/erc1155_proxy.ts
hysz: Looks good!

contracts/integrations/test/coordinator/coordinator_test.ts
```diff
@@ -53,10 +64,10 @@ blockchainTests.resets('Coordinator integration tests', env => {
             orderConfig: {
                 senderAddress: coordinator.address,
                 feeRecipientAddress: feeRecipient.address,
-                makerAssetData: assetDataUtils.encodeERC20AssetData(makerToken.address),
-                takerAssetData: assetDataUtils.encodeERC20AssetData(takerToken.address),
-                makerFeeAssetData: assetDataUtils.encodeERC20AssetData(makerFeeToken.address),
-                takerFeeAssetData: assetDataUtils.encodeERC20AssetData(takerFeeToken.address),
+                makerAssetData: await devUtils.encodeERC20AssetData.callAsync(makerToken.address),
```
jalextowle: Do you have any sense of how much this will increase build times? I understand the desire to remove `assetDataUtils` from a maintenance perspective, but it does seem like the ability to encode asset data without making an RPC call is nice from a performance perspective. Thoughts?

feuGeneA: Is it not the case that these would be executed in-process rather than RPC due to us using ethereumjs-vm?  Or did we not yet merge that in?  (I keep getting confused on where we are with that, given the mixed resistance to it.)

In any case, I took some measurements:

```bash
0x-monorepo[development]$ time yarn test:contracts
real	19m28.792s
user	19m42.732s
sys	1m42.184s
```

```bash
0x-monorepo[2137/trim-order-utils]$ time yarn test:contracts
real	24m18.599s
user	24m46.280s
sys	1m54.212s
```

(Note that I rebased this PR's branch on `development` before measuring, just so it would be as apples-to-apples as possible.)

xianny: Nice, thanks for doing that @feuGeneA . 

Yeah, all pure functions in typescript will be executed in process with ethereumjs-vm. Looks like it is about 25% slower than doing it in pure ts.

jalextowle: @feuGeneA Thanks for the benchmarking, and the reminder that we are using the ethereumjs-vm for pure functions.

jalextowle: Hmm, that is a fairly significant slowdown. I may bring some of the code in `assetDataUtils` back into `contracts-test-utils` in a later PR.

PR #2305, opened 2019-10-29
Function Assertion Sets

jalextowle:
## Description

This PR implements `ContinuousFunctionAssertionSet` and `FunctionAssertionSequence`. These are both collections of function assertions that allow us to group function assertions into higher level simulations. 

- `ContinuousFunctionAssertionSet`: Runs registered function assertions continuously and randomly.
- `FunctionAssertionSequence`: Runs all of the registered function assertions in order.
 
## Testing instructions

## Types of changes

## Checklist:

### DIFF
--- a/contracts/integrations/test/utils/function_assertions.ts
+++ b/contracts/integrations/test/utils/function_assertions.ts
@@ -29,15 +29,19 @@ export interface Result {
  *              function.
  */
 export interface Condition<TBefore extends any> {
-    before?: (...args: any[]) => Promise<TBefore>;
-    after?: (beforeInfo: TBefore, result: Result, ...args: any[]) => Promise<any>;
+    before: (...args: any[]) => Promise<TBefore>;
+    after: (beforeInfo: TBefore, result: Result, ...args: any[]) => Promise<any>;
 }
 
 /**
- *
+ * The basic unit of abstraction for testing. This just consists of a command that
+ * can be run. For example, this can represent a simple command that can be run, or
+ * it can represent a command that executes a "Hoare Triple" (this is what most of
+ * our `Assertion` implementations will do in practice).
+ * @param runAsync The function to execute for the assertion.
  */
 export interface Assertion {
-    runAsync: (...args: any[]) => Promise<any>;
+    executeAsync: (...args: any[]) => Promise<any>;
 }
 
 export interface RunResult {
@@ -45,16 +49,20 @@ export interface RunResult {
     afterInfo: any;
 }
 
+/**
+ * This class implements `Assertion` and represents a "Hoare Triple" that can be
+ * executed.
+ */
 export class FunctionAssertion<TBefore extends any> implements Assertion {
     // A condition that will be applied to `wrapperFunction`.
     // Note: `TBefore | undefined` is used because the `before` and `after` functions
     //       are optional in `Condition`.
-    public condition: Condition<TBefore | undefined>;
+    public condition: Condition<TBefore>;
 
     // The wrapper function that will be wrapped in assertions.
     public wrapperFunction: ContractWrapperFunction;
 
-    constructor(wrapperFunction: ContractWrapperFunction, condition: Condition<TBefore | undefined>) {
+    constructor(wrapperFunction: ContractWrapperFunction, condition: Condition<TBefore>) {
         this.condition = condition;
         this.wrapperFunction = wrapperFunction;
     }
@@ -63,9 +71,9 @@ export class FunctionAssertion<TBefore extends any> implements Assertion {
      * Runs the wrapped function and fails if the before or after assertions fail.
      * @param ...args The args to the contract wrapper function.
      */
-    public async runAsync(...args: any[]): Promise<RunResult> {
+    public async executeAsync(...args: any[]): Promise<RunResult> {
         // Call the before condition.
-        const beforeInfo = this.condition.before !== undefined ? await this.condition.before(...args) : undefined;
+        const beforeInfo = await this.condition.before(...args);
 
         // Initialize the callResult so that the default success value is true.
         let callResult: Result = { success: true };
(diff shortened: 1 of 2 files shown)


contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -58,11 +70,75 @@ export class FunctionAssertion {
...
+     * Run the functions in this assertion set in order.
+     */
+    public async runAsync(environment: BlockchainTestsEnvironment): Promise<void> {
+        for (let i = 0; i < this.assertions.length; i++) {
+            await this.assertions[i].runAsync(...this.inputGenerators[i]());
+        }
+    }
+}
+
+export class ContinuousFunctionAssertionSet {
```
mzhu25: I kinda liked just `FunctionAssertionSet` better tbh

hysz: Nit - could be good to add a class description. Not immediately obvious what a continuous assertion set should do.

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -9,18 +12,27 @@ export interface ContractWrapperFunction extends ContractGetterFunction {
...
-}
-
 export interface Result {
     data?: any;
-    receipt?: TransactionReceiptWithDecodedLogs;
     success: boolean;
+    receipt?: TransactionReceiptWithDecodedLogs;
+}
+
+export interface Condition {
```
hysz: Since these are so core to the testing framework it could also be good to add descriptions/usage for these interfaces.

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -58,11 +70,75 @@ export class FunctionAssertion {
...
         };
     }
 }
+
+/**
+ * Note: This can be treated like a normal `FunctionAssertion`.
+ */
+export class FunctionAssertionSequence {
+    public readonly assertions: Assertion[] = [];
+    public readonly inputGenerators: Array<() => any[]> = [];
```
mzhu25: let's have an `InputGenerator` type: `type InputGenerator = () => any[]`
and then have:
`public readonly assertionGenerators: Array<[Assertion, InputGenerator]>`

contracts/integrations/test/utils/function_assertions.ts
jalextowle: Interesting. The reason I changed this is because `FunctionAssertionSequence` terminates in a finite number of steps, whereas `ContinuousFunctionAssertionSet` will not (and will instead run continuously until the process is killed) . I was thinking that there could be a continuous variant of `FunctionAssertionSequence` and a finite version of `ContinuousFunctionAssertionSet`. With this said, I think that a more descriptive name is warranted, although maybe this isn't the right name. 

Thoughts on a better name?

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -9,18 +12,27 @@ export interface ContractWrapperFunction extends ContractGetterFunction {
...
 export interface Result {
     data?: any;
-    receipt?: TransactionReceiptWithDecodedLogs;
     success: boolean;
+    receipt?: TransactionReceiptWithDecodedLogs;
+}
+
+export interface Condition {
+    before?: (...args: any[]) => Promise<any>;
+    after?: (beforeInfo: any, result: Result, ...args: any[]) => Promise<any>;
```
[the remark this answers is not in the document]
jalextowle: I don't think so, I can update this.

contracts/integrations/test/utils/function_assertions.ts
mzhu25: o I see; we can leave it as is then

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -9,25 +12,57 @@ export interface ContractWrapperFunction extends ContractGetterFunction {
...
+ * This interface represents a condition that can be placed on a contract function.
+ * This can be used to represent the pre- and post-conditions of a "Hoare Triple" on a
+ * given contract function. The "Hoare Triple" is a way to represent the way that a
+ * function changes state.
+ * @param before A function that will be run before a call to the contract wrapper
+ *               function. Ideally, this will be a "precondition."
+ * @param after A function that will be run after a call to the contract wrapper
+ *              function.
+ */
+export interface Condition<TBefore extends any> {
```
mzhu25: nit:
```suggestion
export interface Condition<TBefore> {
```

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -9,25 +12,57 @@ export interface ContractWrapperFunction extends ContractGetterFunction {
...
+
+/**
+ * This class implements `Assertion` and represents a "Hoare Triple" that can be
+ * executed.
+ */
+export class FunctionAssertion<TBefore extends any> implements Assertion {
+    // A condition that will be applied to `wrapperFunction`.
+    // Note: `TBefore | undefined` is used because the `before` and `after` functions
+    //       are optional in `Condition`.
+    public condition: Condition<TBefore>;
```
mzhu25: This comment looks out-of-date

contracts/integrations/test/utils/function_assertions.ts
```diff
@@ -9,25 +12,57 @@ export interface ContractWrapperFunction extends ContractGetterFunction {
...
 
     // The wrapper function that will be wrapped in assertions.
     public wrapperFunction: ContractWrapperFunction;
 
-    constructor(wrapperFunction: ContractWrapperFunction, condition: Condition) {
+    constructor(wrapperFunction: ContractWrapperFunction, condition: Condition<TBefore>) {
```
mzhu25: Does this work? 
```suggestion
    constructor(wrapperFunction: ContractWrapperFunction, condition: Partial<Condition<TBefore>>) {
        this.condition = {
            before: condition.before || () => undefined,
            after: condition.after || () => undefined,
        };
```

jalextowle: After discussing offline, we decided to keep this the same so that our type-safety can be as rigorous as possible

PR #2312, opened 2019-11-01
MixinStakingPoolRewards unit tests

dorothy-zbornak:
## Description

Adds unit tests for `MixinStakingPoolRewards`.

## Checklist:


dorothy-zbornak:
@jalextowle 
> Really solid tests and test coverage! I do think that a few more of the `internal` functions in `MixinStakingPoolRewards` like `_computePoolRewardsSplit`, which may simplify some of the tests for the external functions

I think `_computePoolRewardsSplit()` is supposed to be `private`, but was left `internal` during refactors because `TestMixinFinalizer`, and it alone, uses it. So I designed the tests to cover that function as if it were `private`.

It's nbd to add those tests, so I don't have a problem adding them. But I would keep the test cases for the other functions that cover `_computePoolRewardsSplit()`, because I'm worried we'll flip the function to `private` in the future, delete its tests, and lose our coverage.

PR #2313, opened 2019-11-04
Change all instances of `networkId` to `chainId`

feuGeneA:
## Description

Merge base chosen just for ease of review, because this PR builds on top of `@0x/contract-addresses` changes in that #2284, which is based on that branch.

The `test-publish` CI check is currently failing because `@0x/mesh-rpc-client` is out of sync with this PR; specifically, `@0x/mesh-rpc-client` is importing `ContractNetworks` from `@0x/types`, but that interface has been renamed to `ContractChains`.  When `@0x/mesh-rpc-client` is updated, we can simply re-run that CI job and then see all the pretty :heavy_check_mark:'s in here.

## Testing instructions

CI!

## Types of changes

## Checklist:


packages/connect/test/http_client_test.ts
```diff
@@ -53,9 +53,9 @@ describe('HttpClient', () => {
                 assetDataA: assetData,
                 page: 3,
                 perPage: 50,
-                networkdId: 42,
+                chainId: 42,
             };
-            const urlWithQuery = `${url}?assetDataA=${assetData}&networkdId=42&page=3&perPage=50`;
+            const urlWithQuery = `${url}?assetDataA=${assetData}&chainId=42&page=3&perPage=50`;
```
fabioberger: @dekz can you change the 0x API SRA endpoint to use `chainId` instead of `networkId`? We also need to update to SRA V3 spec with this change.

packages/contract-wrappers/test/coordinator_wrapper_test.ts
```diff
@@ -138,15 +139,16 @@ describe.skip('CoordinatorWrapper', () => {
                 },
             },
             NETWORK_ID_TO_CONTRACT_ADDRESSES: {
-                [config.networkId]: contractAddresses,
+                // TODO: change to CHAIN_ID_TO_CONTRACT_ADDRESSES when @0x/coordinator-server is ready
```
fabioberger: Did you open up an issue for this so we can track it?

dekz: @hysz would be good to get this on your radar

PR #2316, opened 2019-11-05
Unit tests for MixinCumulativeRewards

hysz:
## Description

Unit tests for MixinCumulativeRewards in the Staking contracts package.

## Checklist:


contracts/staking/test/unit_tests/mixin_cumulative_rewards_test.ts
```diff
@@ -0,0 +1,258 @@
...
+            const reward = await testContract.computeMemberRewardOverInterval.callAsync(
+                testPoolId,
+                stake,
+                beginEpoch,
+                endEpoch,
+            );
+            expect(reward).to.bignumber.equal(ZERO);
+        });
+
+        it('Should revert if start is greater than the end of the interval', async () => {
```
jalextowle: Despite the fact that this should never happen, I think it could make sense to test for the case when we don't have the right "endpoints." Basically, when we are missing the cumulative reward values that we need to calculate rewards.

hysz: This is handled by `Should successfully compute reward are no cumulative reward entries` - in this case the default CR is used.

PR #2317, opened 2019-11-05
`@0x/contracts-integrations`: Pool management, stake management fuzz tests

mzhu25:
## Description
First pass at the fuzzing framework. 
- Creates a `function-assertions` directory, where all `FunctionAssertion`s shall live. Each file corresponds to a contract function, and each function within a file returns a `FunctionAssertion` corresponding to a particular class of input (e.g. valid, invalid). Only valid assertions are implemented thus far.
- Generators corresponding to each contract function are added to their respective `Actor` mixins (in this PR, `Staker` and `PoolOperator`). This generator handles input generation for its own assertion. 
- A `Simulation` encapsulates a particular flow of interactions we would like to simulate/fuzz. The `_assertionGenerator` in a `Simulation` instance will typically instantiate some actors and/or simulations and subsequently choose one of their generators for each `yield`. `StakeManagementSimulation` demonstrates how simulations can be composed.
- There is a `SimulationEnvironment` associated with each simulation, which is used to share state between simulations and actors

Some other potential features that aren't showcased in this PR:
- `FunctionAssertion`s for _invalid_ input to `stake`, `unstake`, etc.
- Multiple or hybrid actors instantiated in a single `Simulation`
- Using fancier, custom input generators in either the `Actor` generators or the `Simulation` generators
- We can also pass in ad-hoc custom input to generators as a parameter to [`next`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator/next)

## Testing instructions
`@0x/contracts-integrations yarn test`

## Checklist:

### DIFF
--- a/contracts/integrations/test/actors/pool_operator.ts
+++ b/contracts/integrations/test/actors/pool_operator.ts
@@ -1,4 +1,4 @@
-import { constants } from '@0x/contracts-staking';
+import { constants, StakingPoolById } from '@0x/contracts-staking';
 import { getRandomInteger } from '@0x/contracts-test-utils';
 import { TransactionReceiptWithDecodedLogs } from 'ethereum-types';
 import * as _ from 'lodash';
@@ -7,7 +7,6 @@ import {
     validCreateStakingPoolAssertion,
     validDecreaseStakingPoolOperatorShareAssertion,
 } from '../function-assertions';
-import { SimulationEnvironment } from '../simulation/simulation';
 import { AssertionResult } from '../utils/function_assertions';
 
 import { Actor, Constructor } from './base';
@@ -39,15 +38,11 @@ export function PoolOperatorMixin<TBase extends Constructor>(Base: TBase): TBase
             this.actor = (this as any) as Actor;
 
             // Register this mixin's assertion generators
-            if (this.actor.simulationEnvironment !== undefined) {
-                this.actor.simulationActions = {
-                    ...this.actor.simulationActions,
-                    validCreateStakingPool: this._validCreateStakingPool(this.actor.simulationEnvironment),
-                    validDecreaseStakingPoolOperatorShare: this._validDecreaseStakingPoolOperatorShare(
-                        this.actor.simulationEnvironment,
-                    ),
-                };
-            }
+            this.actor.simulationActions = {
+                ...this.actor.simulationActions,
+                validCreateStakingPool: this._validCreateStakingPool(),
+                validDecreaseStakingPoolOperatorShare: this._validDecreaseStakingPoolOperatorShare(),
+            };
         }
 
         /**
@@ -84,40 +79,29 @@ export function PoolOperatorMixin<TBase extends Constructor>(Base: TBase): TBase
             );
         }
 
-        private _getOperatorPoolIds(simulationEnvironment: SimulationEnvironment): string[] {
-            const operatorPools = _.pickBy(
-                simulationEnvironment.stakingPools,
-                pool => pool.operator === this.actor.address,
-            );
+        private _getOperatorPoolIds(stakingPools: StakingPoolById): string[] {
+            const operatorPools = _.pickBy(stakingPools, pool => pool.operator === this.actor.address);
             return Object.keys(operatorPools);
         }
 
-        private async *_validCreateStakingPool(
-            simulationEnvironment: SimulationEnvironment,
-        ): AsyncIterableIterator<AssertionResult> {
-            const assertion = validCreateStakingPoolAssertion(
-                this.actor.deployment,
-                simulationEnvironment.stakingPools,
-            );
+        private async *_validCreateStakingPool(): AsyncIterableIterator<AssertionResult> {
+            const { stakingPools } = this.actor.simulationEnvironment!;
+            const assertion = validCreateStakingPoolAssertion(this.actor.deployment, stakingPools);
             while (true) {
                 const operatorShare = getRandomInteger(0, constants.PPM);
                 yield assertion.executeAsync(operatorShare, false, { from: this.actor.address });
             }
         }
 
-        private async *_validDecreaseStakingPoolOperatorShare(
-            simulationEnvironment: SimulationEnvironment,
-        ): AsyncIterableIterator<AssertionResult | void> {
-            const assertion = validDecreaseStakingPoolOperatorShareAssertion(
-                this.actor.deployment,
-                simulationEnvironment.stakingPools,
-            );
+        private async *_validDecreaseStakingPoolOperatorShare(): AsyncIterableIterator<AssertionResult | void> {
+            const { stakingPools } = this.actor.simulationEnvironment!;
+            const assertion = validDecreaseStakingPoolOperatorShareAssertion(this.actor.deployment, stakingPools);
             while (true) {
-                const poolId = _.sample(this._getOperatorPoolIds(simulationEnvironment));
+                const poolId = _.sample(this._getOperatorPoolIds(stakingPools));
                 if (poolId === undefined) {
                     yield undefined;
                 } else {
-                    const operatorShare = getRandomInteger(0, simulationEnvironment.stakingPools[poolId].operatorShare);
+                    const operatorShare = getRandomInteger(0, stakingPools[poolId].operatorShare);
                     yield assertion.executeAsync(poolId, operatorShare, { from: this.actor.address });
                 }
             }
--- a/contracts/integrations/test/actors/pool_operator.ts
+++ b/contracts/integrations/test/actors/pool_operator.ts
@@ -1,4 +1,13 @@
+import { constants } from '@0x/contracts-staking';
+import { getRandomInteger } from '@0x/contracts-test-utils';
 import { TransactionReceiptWithDecodedLogs } from 'ethereum-types';
+import * as _ from 'lodash';
+
+import {
+    validCreateStakingPoolAssertion,
+    validDecreaseStakingPoolOperatorShareAssertion,
+} from '../function-assertions';
+import { AssertionResult } from '../utils/function_assertions';
 
 import { Actor, Constructor } from './base';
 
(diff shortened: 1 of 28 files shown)


contracts/integrations/test/actors/pool_operator.ts
```diff
@@ -33,6 +37,17 @@ export function PoolOperatorMixin<TBase extends Constructor>(Base: TBase): TBase
             // tslint:disable-next-line:no-inferred-empty-object-type
             super(...args);
             this.actor = (this as any) as Actor;
+
+            // Register this mixin's assertion generators
+            if (this.actor.simulationEnvironment !== undefined) {
```
jalextowle: Should it construct a simulation if one is undefined?

contracts/integrations/test/actors/staker.ts
```diff
@@ -24,6 +31,20 @@ export function StakerMixin<TBase extends Constructor>(Base: TBase): TBase & Con
             // tslint:disable-next-line:no-inferred-empty-object-type
             super(...args);
             this.actor = (this as any) as Actor;
+            this.stake = {
+                [StakeStatus.Undelegated]: new StoredBalance(),
+                [StakeStatus.Delegated]: { total: new StoredBalance() },
+            };
+
+            // Register this mixin's assertion generators
+            if (this.actor.simulationEnvironment !== undefined) {
```
jalextowle: Ditto.

contracts/integrations/test/actors/staker.ts
```diff
@@ -44,6 +65,79 @@ export function StakerMixin<TBase extends Constructor>(Base: TBase): TBase & Con
...
+
+                const moveableStake =
+                    from.status === StakeStatus.Undelegated
+                        ? this.stake[StakeStatus.Undelegated].nextEpochBalance
+                        : this.stake[StakeStatus.Delegated][from.poolId].nextEpochBalance;
+                const amount = getRandomInteger(0, moveableStake);
+
+                yield assertion.executeAsync(from, to, amount, { from: this.actor.address });
+            }
+        }
```
jalextowle: These are really nice. As discussed offline, I think that we will want some generators that are not as "smart." Optional for this PR though.

contracts/integrations/test/simulation/simulation.ts
```diff
@@ -0,0 +1,38 @@
+import { BlockchainBalanceStore } from '@0x/contracts-exchange';
+import { GlobalStakeByStatus, StakeStatus, StakingPoolById, StoredBalance } from '@0x/contracts-staking';
+import * as _ from 'lodash';
+
+import { DeploymentManager } from '../utils/deployment_manager';
+import { AssertionResult } from '../utils/function_assertions';
+
+// tslint:disable:max-classes-per-file
+
+export class SimulationEnvironment {
```
jalextowle: Thought: This makes a lot of sense as an environment for a staking system. This said, I think that `SimulationEnvironment` should be more abstract than this is. Maybe we can make this the `StakingSimulationEnvironment`? For example, when we extend the framework to function for the exchange contracts, we may not always want these properties.

contracts/integrations/test/simulation/stake_management_test.ts
```diff
@@ -0,0 +1,53 @@
...
+        const staker = new Staker({ name: 'Staker', deployment, simulationEnvironment: this.environment });
+        await staker.configureERC20TokenAsync(deployment.tokens.zrx);
+        balanceStore.registerTokenOwner(staker.address, staker.name);
+
+        const actions = [
+            staker.simulationActions.validStake,
+            staker.simulationActions.validUnstake,
+            staker.simulationActions.validMoveStake,
+            poolManagement.generator,
+        ];
```
jalextowle: I'm a big fan of this yield pattern. It's a lot like using closures for setup, but it seems like it will not bloat the heap as much.

contracts/integrations/test/internal-integration-tests/fillorder_test.ts
```diff
@@ -270,9 +272,7 @@ blockchainTests.resets('fillOrder integration tests', env => {
         );
 
         // The rewards are split between the operator and delegator based on the pool's operatorShare
-        const operatorReward = rewardsAvailable
-            .times(operator.operatorShares[poolId])
-            .dividedToIntegerBy(constants.PPM_DENOMINATOR);
+        const operatorReward = rewardsAvailable.times(operatorShare).dividedToIntegerBy(constants.PPM_DENOMINATOR);
```
jalextowle: Nit: I think that we can use `getPartialAmountFloor` for these calculations.

contracts/exchange/test/balance_stores/balance_store.ts
```diff
@@ -36,6 +36,17 @@ export class BalanceStore {
         };
     }
 
+    /**
+     * Registers the given token owner in this balance store. The token owner's balane will be
```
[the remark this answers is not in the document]
hysz: Hmm I kind of like `balane`. It's got panache 😎

contracts/integrations/test/actors/pool_operator.ts
hysz: Or at least raise an exception so we don't fail silently.

contracts/integrations/test/actors/pool_operator.ts
```diff
@@ -63,13 +77,51 @@ export function PoolOperatorMixin<TBase extends Constructor>(Base: TBase): TBase
...
+
+        private _getOperatorPoolIds(simulationEnvironment: SimulationEnvironment): string[] {
+            const operatorPools = _.pickBy(
+                simulationEnvironment.stakingPools,
+                pool => pool.operator === this.actor.address,
+            );
+            return Object.keys(operatorPools);
+        }
+
+        private async *_validCreateStakingPool(
```
hysz: Is there a plan to include `invalid` generators?

contracts/integrations/test/actors/staker.ts
```diff
@@ -44,6 +65,79 @@ export function StakerMixin<TBase extends Constructor>(Base: TBase): TBase & Con
...
+            simulationEnvironment: SimulationEnvironment,
+        ): AsyncIterableIterator<AssertionResult> {
+            const { zrx } = this.actor.deployment.tokens;
+            const { deployment, balanceStore, globalStake } = simulationEnvironment;
+            const assertion = validStakeAssertion(deployment, balanceStore, globalStake, this.stake);
+
+            while (true) {
+                await simulationEnvironment.balanceStore.updateErc20BalancesAsync();
+                const zrxBalance = simulationEnvironment.balanceStore.balances.erc20[this.actor.address][zrx.address];
+                const amount = getRandomInteger(0, zrxBalance);
```
hysz: What happens when `zrxBalance == 0`?

contracts/integrations/test/function-assertions/moveStake.ts
```diff
@@ -0,0 +1,111 @@
...
+            const owner = txData.from as string;
+            const updatedPools = [];
+            if (from.status === StakeStatus.Undelegated) {
+                decrementNextEpochBalance(ownerStake[StakeStatus.Undelegated], amount);
+                decrementNextEpochBalance(globalStake[StakeStatus.Undelegated], amount);
+            } else if (from.status === StakeStatus.Delegated) {
+                _.defaults(ownerStake[StakeStatus.Delegated], {
+                    [from.poolId]: new StoredBalance(),
+                });
+                decrementNextEpochBalance(ownerStake[StakeStatus.Delegated][from.poolId], amount);
```
hysz: Could be good to move this into a helper

contracts/integrations/test/simulation/stake_management_test.ts
```diff
@@ -0,0 +1,53 @@
...
+        while (true) {
+            const action = _.sample(actions);
+            yield (await action!.next()).value; // tslint:disable-line:no-non-null-assertion
+        }
+    }
+}
+
+blockchainTests.skip('Stake management fuzz test', env => {
```
hysz: Could be nice to have a way to run this w/o recompiling. Like, `yarn fuzz`

contracts/integrations/test/function-assertions/moveStake.ts
```diff
@@ -0,0 +1,111 @@
...
+ */
+export function validMoveStakeAssertion(
+    deployment: DeploymentManager,
+    globalStake: GlobalStakeByStatus,
+    ownerStake: OwnerStakeByStatus,
+    pools: StakingPoolById,
+): FunctionAssertion<{}> {
+    const { stakingWrapper } = deployment.staking;
+
+    return new FunctionAssertion<{}>(stakingWrapper.moveStake, {
```
hysz: Nit - These assertions are all pretty long. Splitting into helpers and/or adding comments would help w readability.

contracts/integrations/test/actors/staker.ts
mzhu25: since we also have `unstake` operations, this won't happen in the simulation

contracts/integrations/test/simulation/simulation.ts
mzhu25: There isn't really any overhead for the unused properties and it's already a pretty lightweight class, so I'm going to leave as is for now

contracts/integrations/test/actors/pool_operator.ts
mzhu25: yes, definitely. Will be in the next PR

contracts/integrations/test/simulation/simulation.ts
jalextowle: Okay, that's fine. I do think it should be made more abstract in a future PR though.

contracts/integrations/test/simulation/stake_management_test.ts
mzhu25: I'm encountering some weird shit with mocha when trying to do this 😕 
but pls do add to your PR if you can figure it out

contracts/integrations/test/function-assertions/moveStake.ts
mzhu25: I can't think of a clean way to do this with the way that js passes arguments by reference 😕

PR #2321, opened 2019-11-07
@0x/order-utils refactors for v3: orderParsingUtils, signatureUtils, orderHashUtils, RevertErrors, transactionHashUtils

feuGeneA:
-   [x] Move `orderParsingUtils` into `@0x/connect`.
-   [x] From `signatureUtils`, remove:
    - isValidSignatureAsync
    - isValidPresignedSignatureAsync
    - isValidWalletSignatureAsync
    - isValidValidatorSignatureAsync
    - isValidECSignature
    - isValidWalletSignatureAsync
-   [x] Remove `orderHashUtils.isValidOrderHash()`.
-   [x] Change all non-contract-test callers of `orderHashUtils.getOrderHash()` to call into the `Exchange.getOrderInfo()` contract method instead, and move `orderHashUtils.getOrderHash()` into `@0x/contracts-test-utils`.
-   [x] Move `*RevertError` types out of `@0x/order-utils` and into their respective `@0x/contracts-*` packages.
-   [x] Move `*RevertError` types out of `@0x/utils` and into their respective `@0x/contracts-*` packages.
-   [x] Delete `transactionHashUtils.isValidTransactionHash()`.
-   [x] Refactor `transactionHashUtils.getTransactionHashHex()` to use contract wrapper call.  [Note: I took a slightly different approach than what was prescribed.  The situation was very similar to the one for `getOrderHash()`.  That is, the vast majority of callers were in tests of contracts, so I moved the existing implementation into `@0x/contracts-test-utils`, exposed a new `getTransactionHash()` method on the `DevUtils` contract, and changed all non-contract-test callers to call into directly the contract method instead.)


packages/order-utils/src/eip712_utils.ts
```diff
@@ -100,18 +101,24 @@ export const eip712Utils = {
...
-    ): EIP712TypedData => {
+    ): Promise<EIP712TypedData> {
         const domain = {
             ...transaction.domain,
             name: constants.COORDINATOR_DOMAIN_NAME,
             version: constants.COORDINATOR_DOMAIN_VERSION,
             verifyingContract,
         };
-        const transactionHash = transactionHashUtils.getTransactionHashHex(transaction);
+        const transactionHash = await new DevUtilsContract('0x0000000000000000000000000000000000000000', {
```
fabioberger: @feuGeneA instantiating a `DevUtilsContract` takes up quite a bit of memory because it instantiates an EVM instance under-the-hood (according to what @xianny has told me), so we only want to instantiate it once per app and pass it around.

In _this_ particular case, we should just move the TS implementation over the `contracts/coordinator` which is the only code using it, and _keep_ the TS implementation. It would live as a test utils in that package, which is fine.

fabioberger: This goes for `eip712_utils` and `order_hash_utils`.

contracts/dev-utils/compiler.json
```diff
@@ -7,7 +7,7 @@
         "evmVersion": "constantinople",
         "optimizer": {
             "enabled": true,
-            "runs": 10000,
+            "runs": 1666,
```
fabioberger: Is this related?

contracts/dev-utils/src/wrappers.ts
```diff
@@ -4,5 +4,6 @@
  * -----------------------------------------------------------------------------
  */
 export * from '../generated-wrappers/dev_utils';
+export * from '../generated-wrappers/eth_balance_checker';
```
fabioberger: Hm, this seems like a regression. We do _not_ want to export `EthBalanceChecker` from `contracts/dev-utils`. Please check that it's not in the `publicInterfaceContracts` key in `package.json` `configs`.

contracts/exchange-libs/src/reference_functions.ts
```diff
@@ -1,8 +1,9 @@
 import { ReferenceFunctions } from '@0x/contracts-utils';
-import { LibMathRevertErrors } from '@0x/order-utils';
 import { FillResults, Order } from '@0x/types';
 import { BigNumber } from '@0x/utils';
 
+import { LibMathRevertErrors } from '.';
```
fabioberger: This pattern is disallowed. Please reference the actual file where this imported item resides in.

contracts/exchange-libs/test/lib_fill_results.ts
```diff
@@ -7,12 +7,13 @@ import {
...
 } from '@0x/contracts-test-utils';
-import { LibMathRevertErrors } from '@0x/order-utils';
+import { SafeMathRevertErrors } from '@0x/contracts-utils';
 import { FillResults, MatchedFillResults, Order } from '@0x/types';
-import { BigNumber, SafeMathRevertErrors } from '@0x/utils';
+import { BigNumber } from '@0x/utils';
 import { Web3Wrapper } from '@0x/web3-wrapper';
 import * as _ from 'lodash';
 
+import { LibMathRevertErrors } from '../src';
```
fabioberger: Never import for `./src`. See the pain it caused me in: https://github.com/0xProject/0x-monorepo/pull/2330

contracts/exchange-libs/test/lib_math.ts
```diff
@@ -6,9 +6,10 @@ import {
     testCombinatoriallyWithReferenceFunc,
     uint256Values,
 } from '@0x/contracts-test-utils';
-import { LibMathRevertErrors } from '@0x/order-utils';
-import { BigNumber, SafeMathRevertErrors } from '@0x/utils';
+import { SafeMathRevertErrors } from '@0x/contracts-utils';
+import { BigNumber } from '@0x/utils';
 
+import { LibMathRevertErrors } from '../src';
```
fabioberger: Fix all instances of this by actually importing from a specific file please.

contracts/test-utils/test/order_hash_test.ts
```diff
@@ -64,19 +63,4 @@ describe('Order hashing', () => {
...
-        it('returns false if the length is wrong', () => {
-            const isValid = orderHashUtils.isValidOrderHash('0xdeadbeef');
-            expect(isValid).to.be.false();
-        });
-        it('returns true if order hash is correct', () => {
-            const orderHashLength = 65;
-            const isValid = orderHashUtils.isValidOrderHash(`0x${Array(orderHashLength).join('0')}`);
-            expect(isValid).to.be.true();
-        });
-    });
```
fabioberger: Let's move the tests over to the `contracts-coordinator` as well.

contracts/dev-utils/compiler.json
feuGeneA: Yes, this is needed in order to reduce the size of the compiled bytecode for the DevUtils contract, to make it deployable.

contracts/exchange-libs/src/reference_functions.ts
feuGeneA: Fixed in https://github.com/0xProject/0x-monorepo/pull/2321/commits/2bba0e361fcdb63971a15b938fcb604c17965b45

contracts/exchange-libs/test/lib_fill_results.ts
feuGeneA: Fixed in https://github.com/0xProject/0x-monorepo/pull/2321/commits/2bba0e361fcdb63971a15b938fcb604c17965b45

contracts/exchange-libs/test/lib_math.ts
feuGeneA: Fixed in https://github.com/0xProject/0x-monorepo/pull/2321/commits/2bba0e361fcdb63971a15b938fcb604c17965b45

contracts/dev-utils/src/wrappers.ts
feuGeneA: Fixed in https://github.com/0xProject/0x-monorepo/pull/2321/commits/3fdadaf34a328e54737796d3f35e13d97c308a15

contracts/test-utils/test/order_hash_test.ts
feuGeneA: As discussed, these tests _were_ kept, but they got moved to `@0x/contracts-test-utils` rather than `contracts-coordinator`, because `test-utils` is where the `getOrderHashHex()` was implementation moved to.

packages/order-utils/src/eip712_utils.ts
feuGeneA: As discussed on slack, this needs to stay the way it is for now, because other things are still using `eip712Utils`.

PR #2324, opened 2019-11-07
Remove unused contract wrappers

fabioberger:
## Description

This PR:
- Removes `ZRXToken`, `DutchAuction` and all assetProxy contracts from `abi-gen-wrappers` and `contract-wrapper` packages in an effort to make these packages smaller.
- As part of refactoring the above, I needed to move `ExchangeTransferSimulator` tests out of `order-utils` in order to have the tests continue to refer to `contract-*` wrappers no longer available via `abi-gen-wrappers` without importing a bunch of wrappers into `order-utils`. We wanted to move this and it's deps anyways into `contracts-exchange` since that is the only place that still uses this functionality and we want to re-direct external devs to use `DevUtils` wrapper methods for order validation.
- Stopped deploying the old DutchAuction contract in the migrations script. It can be re-added once it's been updated.

## Checklist:


contracts/exchange/test/exchange_transfer_simulator_test.ts
```diff
@@ -95,7 +88,7 @@ describe('ExchangeTransferSimulator', async () => {
             const balanceAndProxyAllowanceLazyStore = new BalanceAndProxyAllowanceLazyStore(
                 simpleERC20BalanceAndProxyAllowanceFetcher,
             );
-            exchangeTransferSimulator = new ExchangeTransferSimulator(balanceAndProxyAllowanceLazyStore);
+            exchangeTransferSimulator = new ExchangeTransferSimulator(balanceAndProxyAllowanceLazyStore, env.provider);
```
xianny: What do you think of passing an instance of `DevUtilsContract` instead of the provider? It's what I did in `LocalBalanceStore`. 

Each instance of `DevUtilsContract` instantiates its own EVM and keeps it around in memory so having a lot of instances around might start to affect performance. I also talked to @BMillman19 about making it a convention to instantiate contract wrappers once per app wherever possible.

packages/migrations/src/migration.ts
```diff
@@ -184,14 +195,16 @@ export async function runMigrationsAsync(
...
 
-    // DutchAuction
-    const dutchAuction = await wrappers.DutchAuctionContract.deployFrom0xArtifactAsync(
-        artifacts.DutchAuction,
-        provider,
-        txDefaults,
-        artifacts,
-        exchange.address,
-    );
+    // TODO(fabio): Uncomment dutchAuction once the @0x/contracts-extensions is refactored
```
dekz: bump `dummyTransactionCount` to `6` to avoid changing all contract-addresses that follow. Once the dust has settled we'll remove this hack and ensure all of our documentation and examples reference the correct contracts for the Ganache migration.

packages/abi-gen-wrappers/package.json
```diff
@@ -21,7 +21,7 @@
         "generate_contract_wrappers": "abi-gen --abis  ${npm_package_config_abis} --output src/generated-wrappers --backend ethers"
     },
     "config": {
-        "abis": "../contract-artifacts/artifacts/@(AssetProxyOwner|DevUtils|DutchAuction|DummyERC20Token|DummyERC721Token|ERC1155Mintable|ERC20Proxy|ERC20Token|ERC721Proxy|ERC721Token|Exchange|Forwarder|IAssetProxy|IValidator|IWallet|MultiAssetProxy|OrderValidator|WETH9|ZRXToken|Coordinator|CoordinatorRegistry|EthBalanceChecker|ERC1155Proxy|StaticCallProxy|Staking|StakingProxy).json"
+        "abis": "../contract-artifacts/artifacts/@(DevUtils|DummyERC20Token|DummyERC721Token|ERC1155Mintable|ERC20Token|ERC721Token|Exchange|Forwarder|IValidator|IWallet|OrderValidator|WETH9|Coordinator|CoordinatorRegistry|EthBalanceChecker|Staking|StakingProxy).json"
```
dekz: Last I checked we still weren't using `EthBalanceChecker` any where and the functionality exists in DevUtils.

PR #2326, opened 2019-11-09
Remove chainId from @0x/connect related tooling

fragosti:
## Description

In https://github.com/0xProject/standard-relayer-api/pull/58 we are removing the `networkId` param. So we also have to remove it from all our tools. Clearly this feature will not be missed because it was never used I removed it from:
- `@0x/json-schemas`.
- `@0x/sra-spec`.
- `@0x/connect`.
- `@0x/orderbook`. 
- `@0x/asset-swapper` for when it talks to connect, but not for when it initialized contract-wrappers.

Am I missing anything? 

## Testing instructions

If tests pass, we should be good.

## Types of changes

Strictly speaking this is a breaking change because it removes an option from a lot of our packages. However, it appears this options never really worked in the first place.

## Checklist:


packages/asset-buyer/src/asset_buyer.ts
```diff
@@ -76,8 +76,8 @@ export class AssetBuyer {
     ): AssetBuyer {
         const provider = providerUtils.standardizeOrThrow(supportedProvider);
         assert.isWebUri('sraApiUrl', sraApiUrl);
-        const chainId = options.chainId || constants.DEFAULT_ASSET_BUYER_OPTS.chainId;
-        const orderProvider = new StandardRelayerAPIOrderProvider(sraApiUrl, chainId);
+        // HACK: asset-buy will be deleted, but do not pass in chainId to allow everything to compile.
```
fabioberger: @fragosti will @dave4506 be able to find all these hacks easily? E.g., is there a keyword or list of things he should grep for?

fragosti: This is the only hack in this PR, and we are deleting the entire package (task is tracked in Asana so we don't forget etc...). So I think we're good.

PR #2329, opened 2019-11-11
Add DebugSubprovider

steveklebanoff:
## Description

Adds a DebugSubprovider.  Useful for investigating RPC calls.

Inspired from the situation in which transactions were not getting generated correctly, and it took a long time to recognize that a subprovider was generating incorrect nonces.

## Testing instructions

Add a DebugSubprovider before the final subprovider (i.e. RPCSubprovider) and watch statements be logged to `console.debug`

## Types of changes

New feature

## Checklist:

### DIFF
--- a/packages/subproviders/CHANGELOG.json
+++ b/packages/subproviders/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "version": "5.1.0-beta.2",
+        "changes": [
+            {
+                "note": "Adds DebugSubprovider",
+                "pr": 2329
+            }
+        ]
+    },
+    {
         "version": "5.1.0-beta.1",
         "changes": [
             {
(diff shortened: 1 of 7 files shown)


packages/subproviders/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "5.1.0-beta.2",
```
steveklebanoff: Wasn't quite sure what the correct version number should be here, can someone confirm? CC @fabioberger

packages/subproviders/package.json
```diff
@@ -59,7 +59,7 @@
         "@0x/tslint-config": "^3.1.0-beta.1",
         "@types/bip39": "^2.4.0",
         "@types/bn.js": "^4.11.0",
-        "@types/ethereumjs-tx": "^1.0.0",
+        "@types/ethereumjs-tx": "^1.0.1",
```
steveklebanoff: I needed to do this to get the `gasPrice` and `gasLimit` attributes

packages/subproviders/CHANGELOG.json
fabioberger: This is correct! Thanks for checking!

PR #2330, opened 2019-11-11
Reduce bundle-size of contracts-* packages

fabioberger:
## Description

This PR addresses the issue of the `@0x/contracts-*` packages bundle sizes with several refactors:
1. Add a `.npmignore` in each package that excludes everything from publish except the `lib/src` dir (until now we've been publishing _everything_: https://unpkg.com/browse/@0x/contracts-exchange@2.1.14/). 
2. Give (1), code within the `test` dir is not published. Because of this, any exports from this dir that we still want exported were moved to the `src` dir.
3. Blanket dir exports were removed from `index.ts` files. We want our public interface to be as small and concise as possible. These were replaced with specific imports. Protocologists, please develop a sensitivity to bundle size and what constitutes the public interface of a package 🙏 .
4. In order to prevent publishing `artifacts` and `wrappers` for _every_ mixin, interface and test contract, `@0x/contracts-gen` now generates all the`artifacts` and `wrappers` into the `test` dir of each package. Each `package.json` now contains a `publicInterfaceContracts` config where we then pick which contracts we want to expose via the public interface. These contract artifacts/wrappers then get copied from the `test` dir to the `src` dir as part of the build process.

This refactor was made more difficult due to the prolific use of `import * from '../src'` (imports from index.ts) within packages. Please stop using this convention as it makes future refactors harder. It doesn't take that much work to explicitly import like one does with external imports. 

#### @0x/contracts-exchange

Before: 
```
npm notice name:          @0x/contracts-exchange
npm notice version:       2.2.0-beta.1
npm notice filename:      0x-contracts-exchange-2.2.0-beta.1.tgz
npm notice package size:  6.7 MB
npm notice unpacked size: 72.5 MB
npm notice shasum:        986383da082765999c4a48377cfa404d73a92811
npm notice integrity:     sha512-9VD+ekw5gJ8kU[...]eTAEbNu4EhD6g==
npm notice total files:   757
```
After:
```
npm notice name:          @0x/contracts-exchange
npm notice version:       2.2.0-beta.1
npm notice filename:      0x-contracts-exchange-2.2.0-beta.1.tgz
npm notice package size:  261.8 kB
npm notice unpacked size: 3.2 MB
npm notice shasum:        851cef55f3f006c0bc17fe8f2cc7660a004d493f
npm notice integrity:     sha512-wewtcutvVc9tN[...]qZf2tGPDjtu8A==
npm notice total files:   69
```

#### @0x/contracts-asset-proxy

Before:
```
npm notice name:          @0x/contracts-asset-proxy
npm notice version:       2.3.0-beta.1
npm notice filename:      0x-contracts-asset-proxy-2.3.0-beta.1.tgz
npm notice package size:  1.6 MB
npm notice unpacked size: 13.8 MB
npm notice shasum:        f75727770147ac24f172f97e6dab3a8c13720980
npm notice integrity:     sha512-dZ3wMzYrEeZWF[...]ySfG7/nNheojA==
npm notice total files:   408
```
After:
```
npm notice name:          @0x/contracts-asset-proxy
npm notice version:       2.3.0-beta.1
npm notice filename:      0x-contracts-asset-proxy-2.3.0-beta.1.tgz
npm notice package size:  318.6 kB
npm notice unpacked size: 2.5 MB
npm notice shasum:        ee1e50d0470345d6624c367e5976e8c35c90ecab
npm notice integrity:     sha512-NfvYnjHrYJ2mm[...]ljSQmxARVT7rw==
npm notice total files:   99
```

All packages experienced close to an order of magnitude size reduction.

## Future improvements

There are still a handful of testing utilities that are exported from `contracts-*` packages. We should try and refactor the packages such that this no longer is the case. This should be done in a follow-up PR.

## Checklist:


contracts/integrations/src/deployment_manager.ts
```diff
@@ -9,13 +9,7 @@ import {
 import { artifacts as ERC1155Artifacts, ERC1155MintableContract } from '@0x/contracts-erc1155';
```
dorothy-zbornak: Why is this file being exported? Isn't `DeploymentManager` only used by the tests in this package?

fabioberger: Good catch, reverted this change.

PR #2335, opened 2019-11-13
`@0x/contracts-integrations`: Restructure directories, welcome core.ts and balance stores 

mzhu25:
## Description
Moves a bunch of files and directories around in the integrations package. The updated structure:
Screen Shot 2019-11-13 at 13 43 40
I'd potentially like to have balance stores use stuff from the integrations test framework (DeploymentManager, actors, simulations, etc.), which required moving balance stores out of the exchange package. 
This required also moving `core.ts` and `FillOrderWrapper`. This in turn required moving some signature validation tests from `core` to `signature_validation.ts`. 🙈 And I also deleted `TestSignatureValidator.sol` while I was there.
I didn't do a whole refactor of `core.ts` in this PR, just wanted to get the move done. 

Also changes a bunch of imports to be directly from the file instead of one-off modules, per #2330, and brings us one file closer to killing `ExchangeWrapper` 🎊 

## Checklist:


contracts/integrations/test/framework/assertions/function_assertion.ts
```diff
@@ -106,79 +106,3 @@ export class FunctionAssertion<TBefore, ReturnDataType> implements Assertion {
         };
     }
 }
-
-export type IndexGenerator = () => number;
```
jalextowle: RIP. Can't believe we never even tried these.

contracts/integrations/test/exchange/core_test.ts
```diff
@@ -249,115 +232,16 @@ blockchainTests.resets('Exchange core', () => {
             signedOrder = await orderFactory.newSignedOrderAsync();
         });
 
-        describe('callback signature types', () => {
```
jalextowle: Why were these removed?

mzhu25: they were moved to `signature_validator.ts` in `@0x/contracts-exchange` because they rely on `TestValidatorWalletContract`

contracts/test-utils/src/types.ts
```diff
@@ -201,3 +201,14 @@ export enum ExchangeFunctionName {
     SetProtocolFeeCollectorAddress = 'setProtocolFeeCollectorAddress',
     DetachProtocolFeeCollector = 'detachProtocolFeeCollector',
 }
+
+export enum ValidatorWalletAction {
```
dorothy-zbornak: Do we use these anywhere other than the signature validator tests?

contracts/exchange/test/signature_validator.ts
```diff
@@ -1118,21 +1118,168 @@ blockchainTests.resets('MixinSignatureValidator', env => {
...
+                ExchangeSignatureValidatorApprovalEventArgs
             >;
             const logArgs = log.args;
             expect(logArgs.signerAddress).to.equal(signerAddress);
             expect(logArgs.validatorAddress).to.equal(validatorWallet.address);
             expect(logArgs.isApproved).to.equal(approval);
         });
     });
+
+    describe('fillOrder integration tests', () => {
```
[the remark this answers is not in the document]
mzhu25: yeah these are pretty gross but the cases in `isolated_fill_order.ts` use different signatures, whereas these use the same signature but change the validator wallet's state ☹️ 
If we ever do a tech debt demolition sprint I've got a list of files to delet or refactor and this'll be one of them

contracts/test-utils/src/types.ts
mzhu25: nope, good catch
will move

PR #2343, opened 2019-11-15
abi-gen-wrappers: Move shared logic into BaseContract

dekz:
## Description

* Re-enables `estimateGas` if `gas` is not provided for `sendTransaction`
* Removes `lodash` dependency from `BaseContract`
* Refactors some shared logic into `BaseContract`

Before
```
212K dev_utils.ts
343K exchange.ts
231K staking.ts
```

After
```
163K dev_utils.ts
237K exchange.ts
138K staking.ts
```

Diff
```
-23% dev_utils.ts
-31% exchange.ts
-40% staking.ts
```

### Future

Next best optimization would be to sideload named tuples to reduce Exchange. I.e `Order` can be stored once and unrolled when queried.

This would involve extracting repeated `tuples` and storing them once and referencing them when needed. For example, all functions in exchange which reference `Order` or `FillResults` can reference the same stored structure.

```typescript
const orderComponent = [ { name: 'makerAssetAmount', type: 'uint256' } ... ];
const fillResultsComponent = [ { name: 'makerAssetFilledAmount', type: 'uint256' } ... ];
const abi = [
  {
      name: 'batchFillOrKillOrders',
      inputs: [ { name: 'orders', type: 'tuple[]', components: orderComponent } ],
      outputs: [ { name: 'fillResults', type: 'tuple[]', components: fillResultsComponent } ],
  }
...
];
```

A [manual naive approach](https://gist.github.com/dekz/9d84f193525b2a89d22779079170cd10) reduces Exchange by -26%.

```
237K exchange.ts
174K exchange_sideloaded.ts
```


packages/base-contract/src/utils/filter_utils.ts
```diff
@@ -19,7 +18,8 @@ export const filterUtils = {
         abi: ContractAbi,
         blockRange?: BlockRange,
     ): FilterObject {
-        const eventAbi = _.find(abi, { name: eventName }) as EventAbi;
+        // tslint:disable:next-line no-unnecessary-type-assertion
```
fabioberger: Can you add a comment for why this was required?

PR #2347, opened 2019-11-16
docs: Generate MD docs for contracts-* packages and a few others

fabioberger:
## Description

This PR refactors our markdown docs generation script so that it can generate docs for packages in the `contracts` dir. It will now generate MD docs for several `contracts-*` packages, as well as a `@0x/orderbook` and `@0x/dev-utils`. It _will not_ upload and render these docs on the website, but it will generate them at publish time, so a history of MD docs will be available in the git history with release commits corresponding to the docs a particular version.

The docs feature checks that the packages export all classes, interfaces, enums that are exposed in the package's public interface. Otherwise you could ship a library that exports the following function `export function doStuff(task Task): bool {...}` and forget to export the type `Task` and library users would be none the wiser. CI `test-doc-generation` will catch these types of issues and notify devs. At dev time, you can much more quickly test for these issues by running:

```
yarn generate_doc --package @0x/dev-utils
```

## Checklist:


contracts/integrations/test/framework/balances/types.ts
```diff
@@ -6,7 +6,7 @@ import { BigNumber } from '@0x/utils';
 // alias for clarity
 type address = string;
 
-interface TokenData<TERC20, TERC721, TERC1155> {
+export interface TokenData<TERC20, TERC721, TERC1155> {
```
xianny: Can we actually skip exporting everything in this file? I grepped and it looks like they're not used anywhere. We're avoiding importing `contracts/integrations` into anything because it's supposed to be used to resolve dependency cycles.

fabioberger: Yep, will fix.

contracts/coordinator/src/index.ts
```diff
@@ -1,5 +1,41 @@
...
+    LibConstantsContract,
+    LibCoordinatorApprovalContract,
+    LibCoordinatorRichErrorsContract,
+    LibEIP712CoordinatorDomainContract,
+} from './wrappers';
 export import CoordinatorRevertErrors = require('./revert_errors');
 export { ApprovalFactory } from './approval_factory';
 export { SignedCoordinatorApproval } from './types';
+export { SignatureType, SignedZeroExTransaction, EIP712DomainWithDefaultSchema } from '@0x/types';
+export {
```
jalextowle: Why is this being exported in both `@0x/contracts-dev-utils` and this package? Do we only expect most consumers to use one of these?

contracts/erc20/src/index.ts
```diff
@@ -1,2 +1,42 @@
...
+    DummyERC20TokenTransferEventArgs,
+    ERC20TokenEventArgs,
+    ERC20TokenEvents,
+    ERC20TokenTransferEventArgs,
+    ERC20TokenApprovalEventArgs,
+    ERC20TokenContract,
+} from './wrappers';
+export { artifacts } from './artifacts';
+export {
+    ContractArtifact,
```
jalextowle: Same question here as for the `coordinator` `src/index.ts` file

contracts/coordinator/src/index.ts
jalextowle: This actually seems to be the case for many contract packages. Do we just want the important bits of `ethereum-types` to be accessible from all of the contracts packages?

fabioberger: @jalextowle it's actually more complicated then that.

First, unlike vanilla `tsdoc`, our docs only render things exported from `index.ts`. This lets us define what the public interface of our package is and hide the complexities of it's internal from the end developer.

In order to support this behavior, we render the docs for the entire package with `tsdoc` and  then remove anything not exported in `index.ts`. For all things exported in `index.ts`, we also remove private methods/properties, etc... For all that is left, we then make sure all parameter & return value types are also exported. Why? So that the docs are comprehensive, and we don't reference a type for which the docs don't render a corresponding declaration. Imagine a method referencing type `LotsOfOpts` without defining what those opts are. Second, it makes sure the end developer can define `LotsOfOpts` and then pass it into the function. Third, imagine a class that takes a `Web3ProviderEngine` instance as a constructor arg but the package doesn't export the `Web3ProviderEngine` class. Our checks will catch this as well and make sure a package has all the requisite exports to make the interface usable.

Without these checks, we'd be publishing a bunch of libraries that expect specific types in their interfaces without making sure the packages _also_ provide those types to the end developer.

jalextowle: Thanks for the thoughtful reply. That makes sense! 

@fabioberger Also explained offline that duplicated exports from multiple packages are de-duplicated if they are of the same version, which answers a follow-up to these questions.

PR #2348, opened 2019-11-17
Refactor/3.0/coordinator client

xianny:
## Description

This PR removes the hand-written coordinator client/wrapper from @0x/contract-wrappers and moves it to @0x/contracts-coordinator. The motivation for this is that we are planning to deprecate @0x/contract-wrappers and encourage developers to import directly from @0x/contracts-* packages.

This PR also updates the client to be compatible with V3. 

The diff is messed up because we [disabled the coordinator wrapper earlier](https://github.com/0xProject/0x-monorepo/commit/b4b43a9e9ebb0fded51c9e5291d7091791f04239)
This is a comparison with the commit BEFORE it was disabled: https://github.com/0xProject/0x-monorepo/compare/df97b20913a6ad901e858517a82b7f91cc0dd870..refactor/3.0/coordinator-client
Unfortunately it looks like the diff is too big AND the file was renamed so github isn't showing a nice diff :(

Specific changes, in addition to V3 refactor:
- `packages/contract-wrappers/.../coordinator_wrapper.ts` -> `contracts/coordinator/src/client/index.ts`
- Simplified some of the code in executing transactions in `client/index.ts`
- Add `await` to unit tests so test failures don't get missed

## Testing instructions
CI 🙏 

## Types of changes

## Checklist:


contracts/coordinator/src/client/index.ts
```diff
@@ -0,0 +1,892 @@
...
+
+// tx expiration time will be set to (now + default_approval - time_buffer)
+const DEFAULT_APPROVAL_EXPIRATION_TIME_SECONDS = 90;
+const DEFAULT_EXPIRATION_TIME_BUFFER_SECONDS = 30;
+
+/**
+ * This class includes all the functionality related to filling or cancelling orders through
+ * the 0x V2 Coordinator extension contract.
+ */
+export class CoordinatorClient {
```
hysz: Nice! 🥂

contracts/coordinator/src/client/index.ts
```diff
@@ -0,0 +1,892 @@
...
+     * @param rightOrder             Second order to match.
+     * @param leftSignature          Proof that order was created by the left maker.
+     * @param rightSignature         Proof that order was created by the right maker.
+     * @param   txData               Transaction data. The `from` field should be the user Ethereum address who would like
+     *                               to fill these orders. Must be available via the Provider supplied at instantiation.
+     * @param   sendTxOpts           Optional arguments for sending the transaction.
+     * @return  Transaction hash.
+     */
+    @decorators.asyncZeroExErrorHandler
+    public async matchOrdersAsync(
```
hysz: Order matching is not supported by the coordinator server.

contracts/coordinator/src/client/index.ts
```diff
@@ -0,0 +1,892 @@
...
+     * @param rightOrders       Set of orders to match against `leftOrders`
+     * @param leftSignatures    Proof that left orders were created by the left makers.
+     * @param rightSignatures   Proof that right orders were created by the right makers.
+     * @param txData            Transaction data. The `from` field should be the user Ethereum address who would like
+     *                          to fill these orders. Must be available via the Provider supplied at instantiation.
+     * @param sendTxOpts        Optional arguments for sending the transaction.
+     * @return  Transaction hash.
+     */
+    @decorators.asyncZeroExErrorHandler
+    public async matchOrdersWithMaximalFillAsync(
```
hysz: Same as above.

contracts/integrations/test/coordinator/client_test.ts
```diff
@@ -0,0 +1,634 @@
...
+            expect(result.status).to.equal(200);
+            expect(await result.text()).to.equal('pong');
+
+            result = await fetchAsync(`${coordinatorEndpoint}${anotherCoordinatorPort}/v2/ping`);
+            expect(result.status).to.equal(200);
+            expect(await result.text()).to.equal('pong');
+        });
+    });
+    // fill handling is the same for all fill methods so we can test them all through the fillOrder and batchFillOrders interfaces
+    describe('#fillOrderAsync', () => {
```
hysz: I think it would be a worthwhile sanity check to test all the functions exposed by the client. It will also future-proof the interface.

packages/asset-swapper/test/swap_quote_consumer_test.ts
```diff
@@ -9,7 +10,6 @@ import { SwapQuote, SwapQuoteConsumer } from '../src';
 import { ExtensionContractType } from '../src/types';
 
 import { chaiSetup } from './utils/chai_setup';
-import { migrateOnceAsync } from './utils/migrate';
```
fabioberger: Why were changes introduced to `asset-swapper`?

packages/monorepo-scripts/src/utils/doc_generate_utils.ts
```diff
@@ -264,7 +264,7 @@ export class DocGenerateUtils {
         if (!_.isEmpty(missingReferences)) {
             throw new Error(
                 `${this._packageName} package needs to export: \n${missingReferences.join(
-                    '\n',
+                    ',\n',
```
fabioberger: 👏

packages/asset-swapper/test/swap_quote_consumer_test.ts
xianny: `migrateOnceAsync` was identical in two files and the unit tests here also needed it. So instead of making a third file I moved it into `@0x/migrations`

contracts/coordinator/src/client/index.ts
xianny: chatted on Slack, `matchOrders` and `matchOrdersWithMaximalFill` are supported, just not the batch versions. https://github.com/0xProject/0x-monorepo/blob/development/contracts/coordinator/contracts/src/MixinCoordinatorApprovalVerifier.sol#L113-L114

PR #2350, opened 2019-11-18
asset-swapper V3 rebase, pay protocol fees, etc.

dorothy-zbornak:
## Description

- Rebases #2272 against `development`.
- `@0x/asset-swapper` now pays protocol fees (now that staking gets hooked up during migration).
- `@0x/asset-swapper` removed dependency on `@0x/contract-wrappers`.
- `@0x/migrations` now deploys `Forwarder` AFTER staking is attached to the Exchange.

## Checklist:


packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
```diff
@@ -182,59 +182,50 @@ export class ForwarderSwapQuoteConsumer implements SwapQuoteConsumerBase<Forward
...
+            txHash = await this._forwarder
+                .marketSellOrdersWithEth(orders, orders.map(o => o.signature), formattedFeePercentage, feeRecipient)
+                .sendTransactionAsync({
+                    from: finalTakerAddress,
+                    gas: gasLimit,
+                    gasPrice,
+                    value: providedEthAmount || ethAmountWithFees,
+                });
         }
+        // TODO(dorothy-zbornak): Handle signature request denied
```
dorothy-zbornak: I removed the `try/catch` that was here because it was doing nothing. `ContractError.SignatureRequestDenied` didn't actually get thrown by contract wrappers, and `sendTransactionAsync()` only mines the TX immediately on ganache, so you won't detect a failure on mainnet. 

`ContractError.SignatureRequestDenied` denied is easy enough to detect, but maybe the logic should live in another package. Detecting a `CompleteFillFailed` will have to be accomplished by doing a `callAsync()` at the same block the TX was mined. Let's add tasks for this.

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
```diff
@@ -135,41 +132,38 @@ export class ExchangeSwapQuoteConsumer implements SwapQuoteConsumerBase<Exchange
...
+            txHash = await this._exchangeContract
+                .marketSellOrdersFillOrKill(orders, takerAssetFillAmount, orders.map(o => o.signature))
+                .sendTransactionAsync({
+                    from: finalTakerAddress,
+                    gas: gasLimit,
+                    gasPrice,
+                    value,
+                });
         }
+        // TODO(dorothy-zbornak): Handle signature request denied
```
dorothy-zbornak: I removed the `try/catch` that was here because it was doing nothing. `ContractError.SignatureRequestDenied` didn't actually get thrown by contract wrappers, and `sendTransactionAsync()` only mines the TX immediately on ganache, so you won't detect a failure on mainnet. 

`ContractError.SignatureRequestDenied` denied is easy enough to detect, but maybe the logic should live in another package. Detecting a `IncompleteFill` will have to be accomplished by doing a `callAsync()` at the same block the TX was mined. Let's add tasks for this.

packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
```diff
@@ -129,23 +131,21 @@ export class ForwarderSwapQuoteConsumer implements SwapQuoteConsumerBase<Forward
...
         }
-        const methodAbi = utils.getMethodAbiFromContractAbi(
-            this._contractWrappers.forwarder.abi,
-            methodName,
-        ) as MethodAbi;
+        const methodAbi = utils.getMethodAbiFromContractAbi(this._forwarder.abi, methodName) as MethodAbi;
 
+        const ethAmountWithFees = affiliateFeeUtils
+            .getTotalEthAmountWithAffiliateFee(worstCaseQuoteInfo, feePercentage)
+            .plus(worstCaseQuoteInfo.protocolFeeInEthAmount);
```
dorothy-zbornak: Not sure if this is what we want, but here I just add protocol fees to the total payable amount.

packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
```diff
@@ -182,59 +182,50 @@ export class ForwarderSwapQuoteConsumer implements SwapQuoteConsumerBase<Forward
...
+        const { orders, worstCaseQuoteInfo } = quote; // tslint:disable-line:no-unused-variable
 
         // get taker address
         const finalTakerAddress = await swapQuoteConsumerUtils.getTakerAddressOrThrowAsync(this.provider, opts);
-        // if no ethAmount is provided, default to the worst totalTakerTokenAmount
-        const value = ethAmount || worstCaseQuoteInfo.totalTakerTokenAmount;
+        // if no ethAmount is provided, default to the worst totalTakerAssetAmount
+        const ethAmountWithFees = affiliateFeeUtils
+            .getTotalEthAmountWithAffiliateFee(worstCaseQuoteInfo, feePercentage)
+            .plus(worstCaseQuoteInfo.protocolFeeInEthAmount);
```
dorothy-zbornak: ditto

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
```diff
@@ -69,62 +70,58 @@ export class ExchangeSwapQuoteConsumer implements SwapQuoteConsumerBase<Exchange
...
-            methodName,
-        ) as MethodAbi;
+        const methodAbi = utils.getMethodAbiFromContractAbi(this._exchangeContract.abi, methodName) as MethodAbi;
 
         return {
             params,
-            toAddress: this._contractWrappers.exchange.address,
+            toAddress: this._exchangeContract.address,
             methodAbi,
+            ethAmount: quote.worstCaseQuoteInfo.protocolFeeInEthAmount,
```
dorothy-zbornak: Added to pay protocol fees.

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
```diff
@@ -135,41 +132,38 @@ export class ExchangeSwapQuoteConsumer implements SwapQuoteConsumerBase<Exchange
...
-            return txHash;
-        } catch (err) {
-            if (_.includes(err.message, ContractError.SignatureRequestDenied)) {
-                throw new Error(SwapQuoteConsumerError.SignatureRequestDenied);
-            } else if (_.includes(err.message, ForwarderError.CompleteFillFailed)) {
-                throw new Error(SwapQuoteConsumerError.TransactionValueTooLow);
-            } else {
-                throw err;
-            }
+        const value = ethAmount || quote.worstCaseQuoteInfo.protocolFeeInEthAmount;
```
dorothy-zbornak: Added to pay protocol fees.

packages/asset-swapper/test/sorting_utils_test.ts
```diff
@@ -0,0 +1,134 @@
...
+import { chaiSetup } from './utils/chai_setup';
+import { testOrderFactory } from './utils/test_order_factory';
+
+chaiSetup.configure();
+const expect = chai.expect;
+
+const FAKE_ERC20_TAKER_ASSET_DATA = '0xf47261b22222222222222222222222222222222222222222222222222222222222222222';
+const FAKE_ERC20_MAKER_ASSET_DATA = '0xf47261b11111111111111111111111111111111111111111111111111111111111111111';
+
+describe('sortingUtils', () => {
```
hysz: Nice work! Very cleanly communicates / tests the sorting.

packages/asset-swapper/src/utils/order_prune_utils.ts
```diff
@@ -0,0 +1,96 @@
...
+            };
+        });
+        // take orders + on chain information and find the valid orders and fillable makerAsset or takerAsset amounts
+        const prunedOrders = this._filterForFillableAndPermittedFeeTypeOrders(unsortedOrders, ordersOnChainMetadata);
+
+        return prunedOrders;
+    }
+
+    // tslint:disable-next-line: prefer-function-over-method
+    private _filterForFillableAndPermittedFeeTypeOrders(
```
hysz: There's a lot going on in this function. Could be good to add some inline comments.

packages/asset-swapper/test/sorting_utils_test.ts
dorothy-zbornak: I didn't write like 99% of this PR (it just looks that way because of the rebase), so props to @dave4506

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
dave4506: nice catch, I completely forgot about adding that in.

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
dave4506: Hmm, yeah that try-catch was residual from `asset-buyer`, I do want to ask though, not very familiar with the nuances of the wrappers, how are errors handled? One of the large feedbacks I received at EthWaterloo was the vague errors thrown, don't want us to make `asset-swapper` more opaque with errors.

packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
dave4506: `getTotalEthAmountWithAffiliateFee` will consider protocol fees since affiliateFee is dependent on protocolFee (as of current forwarder design).

packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
dave4506: same case, do not need to add protocolFee

packages/asset-swapper/src/quote_consumers/forwarder_swap_quote_consumer.ts
dorothy-zbornak: ahh, yeah, that makes sense

packages/asset-swapper/src/quote_consumers/exchange_swap_quote_consumer.ts
dorothy-zbornak: `SignatureRequestDenied` *was* just running a regex against the thrown error message (by Metamask or w/e). 

If the use `awaitTransactionSuccessAsync()` instead of `sendTransactionAsync()`, it will throw if the TX fails. Then we can use `callAsync()` to replay the TX and get the actual rich revert, which gives us deeper detail.

PR #2352, opened 2019-11-19
KyberBridge

dorothy-zbornak:
sammy

## Description

It's `KyberBridge`! The triforce is now complete.

## Checklist:


contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+    struct TradeState {
+        IKyberNetworkProxy kyber;
+        IEtherToken weth;
+        address fromTokenAddress;
+        uint256 fromTokenBalance;
+        uint256 payableAmount;
+        uint256 minConversionRate;
+    }
+
+    /// @dev Address of the WETH contract.
```
jalextowle: Optional: It would be cool to use the `DeploymentConstants` pattern here.

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+        state.kyber = _getKyberContract();
+        state.weth = _getWETHContract();
+        // Decode the bridge data to get the `fromTokenAddress`.
+        (state.fromTokenAddress) = abi.decode(bridgeData, (address));
+        state.fromTokenBalance = IERC20Token(state.fromTokenAddress).balanceOf(address(this));
+        if (state.fromTokenBalance == 0) {
+            // Do nothing if no input tokens.
+            return BRIDGE_SUCCESS;
+        }
+        state.minConversionRate = (10 ** 18) * amount / state.fromTokenBalance;
```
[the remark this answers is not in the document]
jalextowle: Also, could we move this closer to where `trade` is called? I think it would help a little bit with comprehension.

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+            // to spend them.
+            LibERC20Token.approve(state.fromTokenAddress, address(state.kyber), uint256(-1));
+        } else {
+            // If the input token is WETH, unwrap it and attach it to the call.
+            state.fromTokenAddress = KYBER_ETH_ADDRESS;
+            state.payableAmount = state.fromTokenBalance;
+            state.weth.withdraw(state.fromTokenBalance);
+        }
+
+        // Try to sell all of this contract's input token balance.
```
hysz: Does `trade` always revert on failure? Is there a scenario where it just returns  an invalid bought amount, like zero?

contracts/asset-proxy/contracts/src/interfaces/IKyberNetworkProxy.sol
```diff
@@ -0,0 +1,46 @@
...
+    /// @param sellTokenAddress Token to sell.
+    /// @param sellAmount Amount of tokens to sell.
+    /// @param buyTokenAddress Token to buy.
+    /// @param recipientAddress Address to send bought tokens to.
+    /// @param maxBuyTokenAmount A limit on the amount of tokens to buy.
+    /// @param minConversionRate The minimal conversion rate. If actual rate
+    ///        is lower, trade is canceled.
+    /// @param walletId The wallet ID to send part of the fees
+    /// @return boughtAmount Amount of tokens bought.
+    function trade(
```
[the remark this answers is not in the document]
dorothy-zbornak: This is the contract we talk to: https://github.com/KyberNetwork/smart-contracts/blob/master/contracts/KyberNetworkProxy.sol#L33

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
abandeali1: Should we have a canonical `DeploymentConstants` that we can use if we're going to use this pattern more frequently? Unused internal functions shouldn't affect the bytecode of the contracts, right?

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+    {
+        TradeState memory state;
+        state.kyber = _getKyberContract();
+        state.weth = _getWETHContract();
+        // Decode the bridge data to get the `fromTokenAddress`.
+        (state.fromTokenAddress) = abi.decode(bridgeData, (address));
+        state.fromTokenBalance = IERC20Token(state.fromTokenAddress).balanceOf(address(this));
+        if (state.fromTokenBalance == 0) {
+            // Do nothing if no input tokens.
+            return BRIDGE_SUCCESS;
```
abandeali1: Shouldn't this actually revert? I wouldn't say it's successful if the `toToken` isn't actually transferred.

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+        }
+
+        // Try to sell all of this contract's input token balance.
+        uint256 boughtAmount = state.kyber.trade.value(state.payableAmount)(
+            // Input token.
+            state.fromTokenAddress,
+            // Sell amount.
+            state.fromTokenBalance,
+            // Output token.
+            toTokenAddress == address(state.weth) ?
```
abandeali1: Nit: this equality check is done multiple times, it might be worth assigning to a `bool` (if the stack limit allows for it)

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,160 @@
...
+                address(uint160(to)),
+            // Buy as much as possible.
+            uint256(-1),
+            // Minimum conversion rate.
+            state.minConversionRate,
+            // No affiliate address.
+            address(0)
+        );
+        // Wrap ETH output and transfer to recipient.
+        if (toTokenAddress == address(state.weth)) {
```
abandeali1: So Kyber will send this contract ETH instead of WETH, right? We'll need a fallback function to accept it in that case.

dorothy-zbornak: Oh woops! I totally forgot! This actually explains a confusing revert I kept getting while writing the tests. :upside_down_face:

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
dorothy-zbornak: I guess. The asset proxy will revert on the balance check as well, but we might as well revert early.

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
dorothy-zbornak: Should revert but I guess with a bad enough conversion rate it could return zero. The asset proxy should catch it on the balance check, if it's too low.

contracts/asset-proxy/contracts/src/bridges/KyberBridge.sol
```diff
@@ -0,0 +1,166 @@
...
+        uint256 boughtAmount = state.kyber.trade.value(state.payableAmount)(
+            // Input token.
+            state.fromTokenAddress,
+            // Sell amount.
+            state.fromTokenBalance,
+            // Output token.
+            isToTokenWeth ? KYBER_ETH_ADDRESS : toTokenAddress,
+            // Transfer to this contract if converting to ETH, otherwise
+            // transfer directly to the recipient.
+            isToTokenWeth ? address(uint160(address(this))) : address(uint160(to)),
```
abandeali1: Nit: can't `address(uint160(to))` just be `to`?

PR #2353, opened 2019-11-20
Asset-swapper aggregator utils

dorothy-zbornak:
yoink

## Description

Adds aggregator utils to `@0x/asset-swapper`, which can be found in [`src/utils/aggregate.ts`](../blob/feat/asset-swapper/ERC20BridgeAggregator/packages/asset-swapper/src/utils/aggregate.ts).

### Improving Quotes
The main functions of interest are `improveMarketSellAsync()`and  `improveMarketBuyAsync()`. Each takes native orders, a fill amount, and some options. They will poll on-chain DEXes and return a mixture of native and bridge orders that can be passed into `marketBuy()` or `marketFill()` on the Exchange contract.

```ts
/**
 * Improve a market sell operation by (potentially) merging native orders with
 * generated bridge orders.
 * @param samplerContract The `IERC20BridgeSamplerContract` contract.
 * @param nativeOrders Native orders.
 * @param takerAmount Amount of taker asset to sell.
 * @param opts Options object.
 * @return Improved orders.
 */
export async function improveMarketSellAsync(
    samplerContract: IERC20BridgeSamplerContract,
    nativeOrders: SignedOrderWithoutDomain[],
    takerAmount: BigNumber,
    opts?: Partial<ImproveOrdersOpts>,
): Promise<SignedOrderWithoutDomain[]>;

/**
 * Improve a market buy operation by (potentially) merging native orders with
 * generated bridge orders.
 * @param samplerContract The `IERC20BridgeSamplerContract` contract.
 * @param nativeOrders Native orders.
 * @param makerAmount Amount of maker asset to buy.
 * @param opts Options object.
 * @return Improved orders.
 */
export async function improveMarketBuyAsync(
    samplerContract: IERC20BridgeSamplerContract,
    nativeOrders: SignedOrderWithoutDomain[],
    makerAmount: BigNumber,
    opts?: Partial<ImproveOrdersOpts>,
): Promise<SignedOrderWithoutDomain[]>;
```

### Options

An options object can be passed into either function, which control slippage on bridge orders, sources used, and performance of the mixing algorithm.

```ts
/**
 * Options for `improveMarketBuyAsync()` and `improveMarketSellAsync()`.
 */
export interface ImproveOrdersOpts {
    /**
     * Liquidity sources to exclude. Default is none.
     */
    excludedSources: ERC20BridgeSource[];
    /**
     * Whether to prevent mixing Kyber orders with Uniswap and Eth2Dai orders.
     */
    noConflicts: boolean;
    /**
     * Complexity limit on the search algorithm, i.e., maximum number of
     * nodes to visit. Default is 1024.
     */
    runLimit: number;
    /**
     * When generating bridge orders, we use
     * sampled rate * (1 - bridgeSlippage)
     * as the rate for calculating maker/taker asset amounts.
     * This should be a small positive number (e.g., 0.0005) to make up for
     * small discrepancies between samples and truth.
     * Default is 0.0005 (5 basis points).
     */
    bridgeSlippage: number;
    /**
     * Number of samples to take for each DEX quote.
     */
    numSamples: number;
    /**
     * Dust amount, as a fraction of the fill amount.
     * Default is 0.01 (100 basis points).
     */
    dustThreshold: number;
}
```

### Changes to the `ERC20BridgeSampler` contract.
Previously, `queryOrdersAndSampleSells/Buys()` would return the `OrderInfo` for each order passed in, along with DEX samples. Now it calls `DevUtils.getOrderRelevantStates()` and returns the *actual* fillable maker or taker amount (depending on whether it's a buy or sell). Also, if the order status is not `FILLABLE` or the signature is invalid, this amount will be zero.

### TODO
-   [x] Fold this library into `asset-swapper`'s quoting logic.
-   [x] The addresses of the bridge contracts in `aggregate.ts` are wrong and will need to be updated once the final versions of the bridges have been deployed.
-   [x] Incorporate native order fees
-   [x] Fix outstanding/breaking TODOs (mainly the redundant work)
-   [ ] Fix test coverage
-   [x] Add Sampler to `@0x/contract-addresses`
-   [x] If no `nativeOrders` pass in a "dummy order" such that swapper now uses only other DEX sources to prepare quote.
-   [ ] Gas is not accounted for. Not sure how we want to deal with these.

## Checklist:


packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+    opts?: Partial<ImproveOrdersOpts>,
+): Promise<OrderWithoutDomain[]> {
+    if (nativeOrders.length === 0) {
+        throw new Error(AggregationError.EmptyOrders);
+    }
+    const _opts = {
+        ...IMPROVE_ORDERS_OPTS_DEFAULTS,
+        ...opts,
+    };
+    const [orderInfos, dexQuotes] = await queryNetworkAsync(
```
dekz: Thoughts on using [`DevUtils.getOrderRelevantState`](https://github.com/0xProject/0x-monorepo/blob/development/contracts/dev-utils/contracts/src/OrderValidationUtils.sol#L51) in the Sampler and returning `[orderInfos, fillableTakerAssetAmount, isValidSignature]`.

We [perform this](https://github.com/0xProject/0x-monorepo/blob/development/packages/asset-swapper/src/utils/order_prune_utils.ts#L24) to ensure the order is valid and check how much is fillable. We could skip that check a by combining it in the Sampler and reduce an RPC call.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+ * generated bridge orders.
+ * @param nativeOrders Native orders.
+ * @param takerAmount Amount of taker asset to sell.
+ * @param opts Options object.
+ * @return Improved orders.
+ */
+export async function improveMarketSellAsync(
+    nativeOrders: OrderWithoutDomain[],
+    takerAmount: BigNumber,
+    opts?: Partial<ImproveOrdersOpts>,
```
dekz: `provider` is required but `opts` is optional. I'd suggest pulling out the provider explicitly.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+    nativeOrders: OrderWithoutDomain[],
+    sampleAmounts: BigNumber[],
+    sources: ERC20BridgeSource[],
+    provider?: SupportedProvider,
+    isBuy?: boolean,
+): Promise<[OrderInfo[], DexSample[][]]> {
+    if (!provider) {
+        throw new Error(AggregationError.MissingProvider);
+    }
+    const sampler = new ERC20BridgeSamplerContract(SAMPLER_ADDRESS, provider);
```
dekz: I'd even consider just passing in the entire `ERC20BridgeSampler`, won't have to deal with optional providers here or in `ImproveOrdersOpts`

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+        { name: 'bridgeAddress', type: 'address' },
+        { name: 'bridgeData', type: 'bytes' },
+    ]);
+    return encoder.encode({ tokenAddress, bridgeAddress, bridgeData });
+}
+
+/**
+ * Create ERC20Proxy asset data.
+ */
+export function createERC20AssetData(tokenAddress: string): string {
```
dekz: we can re-use `DevUtils` here which performs this `pure` in memory.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+        makerAssetData: createBridgeAssetData(makerToken, bridgeAddress, createBridgeData(takerToken)),
+        takerAssetData: createERC20AssetData(takerToken),
+        ...createCommonOrderFields(makerAssetAmount, takerAssetAmount, slippage, isBuy),
+    };
+}
+
+/**
+ * Create ERC20Bridge asset data.
+ */
+export function createBridgeAssetData(tokenAddress: string, bridgeAddress: string, bridgeData: string): string {
```
dekz: We should add an asana task to get this encoding/decoding into `DevUtils` and re-use from that.

packages/asset-swapper/package.json
```diff
@@ -43,6 +43,7 @@
         "@0x/assert": "^2.2.0-beta.2",
         "@0x/contract-addresses": "^3.3.0-beta.4",
         "@0x/abi-gen-wrappers": "^5.4.0-beta.3",
+        "@0x/contracts-erc20-bridge-sampler": "^1.0.0-beta.1",
```
dekz: TODO add this to `abi-gen-wrappers`/export from `contract-wrappers`. These have pruned artifacts and won't bloat out the asset-swapper bundle size.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+ * @param nativeOrders Native orders.
+ * @param takerAmount Amount of taker asset to sell.
+ * @param opts Options object.
+ * @return Improved orders.
+ */
+export async function improveMarketSellAsync(
+    nativeOrders: OrderWithoutDomain[],
+    takerAmount: BigNumber,
+    opts?: Partial<ImproveOrdersOpts>,
+): Promise<OrderWithoutDomain[]> {
```
dekz: Is this correct in terms of turning it into a `WalletSignature`?

```typescript
        const improvedOrders = await improveMarketBuyAsync(prunedOrders, assetFillAmount, {
            provider: this.provider,
        });
        const improvedSignedOrders = improvedOrders.map(i => ({
            ...i,
            signature: '0x04',
            chainId: this.chainId,
            exchangeAddress: this._contractAddresses.exchange,
        }));
```

IIRC 
```
Wallet = 0x04
EIP1271Wallet = 0x07
```

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,780 @@
...
+    if (nativeOrders.length === 0) {
+        throw new Error(AggregationError.EmptyOrders);
```
dekz: What is the flow for just retrieving a bunch of `BridgeOrders` if there are no `NativeOrders` for this asset pair?

Just use `sampleBuys`/`sampleSells` directly?

dorothy-zbornak: I would just pass in an empty order to keep the entry point the same.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: yap!

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: Yeah, I'm down for something like that. It looks like there's some more complex pruning rules being applied. So maybe the sampler contract can query those states and pass them into an `orderFilter` callback in `opts`.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: I hate `DevUtils` but OK :roll_eyes:

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: There's a good chance it won't fit. `DevUtils` needs so much TLC. Like why is `chai` a non-dev dep for `DevUtils`? :woman_shrugging:

dorothy-zbornak: (if it fits)

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: Well, for the bridge orders. You still have to map the native orders to their respective (probably non-`Wallet`) signatures. I'll handle signatures in the upcoming iteration, though, so it'll be moot.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: Hmm, really all we care about is how much taker asset is fillable. I think I'll just have the contract return that quantity, which  will also be zero if the signature is invalid or if the order status is not `FILLABLE`.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: On second thought, how about we wait until `DevUtils` is sorted out since switching this one function won't really do me any favors right now.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
+import { IERC20BridgeSamplerContract } from '@0x/contracts-erc20-bridge-sampler';
```
dave4506: Jacob highlighted above, but will need to rebase this PR with `development` we switched back to using `@0x/contract-wrappers` for bundle size reasons.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+import { generatePseudoRandomSalt } from '@0x/order-utils';
+import { SignedOrderWithoutDomain } from '@0x/types';
+import { AbiEncoder, BigNumber } from '@0x/utils';
+
+import { constants } from '../constants';
+
+const { NULL_BYTES, NULL_ADDRESS } = constants;
+const ZERO = new BigNumber(0);
```
dave4506: nit. these should be `constants.ts`

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+    exclusionMask: number;
+    // Input fill amount (taker asset amount in a sell, maker asset amount in a buy).
+    input: BigNumber;
+    // Output fill amount (maker asset amount in a sell, taker asset amount in a buy).
+    output: BigNumber;
+    // Fill that must precede this one. This enforces certain fills to be contiguous.
+    parent?: Fill;
+    // Arbitrary data to store in this Fill object. Used to reconstruct orders
+    // from paths.
+    data?: any;
```
dave4506: Feel like a more verbose name here other than data would help

dave4506: This should actually just be `FillData`, if we need more data entries, we can always add it.

We can use a union type to capture all of the 'types' that `data` can be.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+    bridgeSlippage: number;
+    /**
+     * Number of samples to take for each DEX quote.
+     */
+    numSamples: number;
+    /**
+     * Dust amount, as a fraction of the fill amount.
+     * Default is 0.01 (100 basis points).
+     */
+    dustThreshold: number;
```
dave4506: nit. prefer `dustAmountThreshhold`

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+            data: {
+                source: ERC20BridgeSource.Native,
+                order,
+            },
+        });
+    }
+    return path;
+}
+
+function createBuyPathFromNativeOrders(orders: SignedOrderWithoutDomain[], fillableAmounts: BigNumber[]): Fill[] {
```
dave4506: Can we reuse some of the logic from above? (as in dedup the logic) Not a big issue

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+        this._currentRunCount = 0;
+        this._optimalPath = optimalPath;
+        this._optimalPathOutput = optimalPath ? getPathOutput(optimalPath, input) : ZERO;
+        const ctx = {
+            currentPath: [],
+            currentPathInput: ZERO,
+            currentPathOutput: ZERO,
+            currentPathFlags: 0,
+        };
+        // Visit all valid combintations of fills to find the optimal path.
```
dave4506:
```suggestion
        // Visit all valid combinations of fills to find the optimal path.
```

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+    // If native orders are allowed, splice them in.
+    if (!(_opts.excludedSources as ERC20BridgeSource[]).includes(ERC20BridgeSource.Native)) {
+        allPaths.splice(0, 0, nativePath);
+        allFills.splice(0, 0, ...nativePath);
+    }
+    const optimizer = new FillsOptimizer(_opts.runLimit);
+    const optimalPath = optimizer.optimize(
+        sortFillsByPrice(allFills),
+        takerAmount,
+        pickOptimalPath(allPaths, takerAmount),
```
dave4506: is it possible to move this `pickOptimalPath` logic into `FillsOptimizer`? Seems like it makes more sense there.

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+function getPartialFillOutput(fill: Fill, partialInput: BigNumber): BigNumber {
+    return BigNumber.min(fill.output, fill.output.div(fill.input).times(partialInput)).integerValue(
+        BigNumber.ROUND_DOWN,
+    );
+}
+
+/**
+ * Class for finding optimized fill paths.
+ */
+export class FillsOptimizer {
```
hysz: I know that AMM's like the one used by Uniswap will give you a worse price the more you fill. Is this the case for all of our sources?

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: Eh?

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: idk, we seem to use `amount` for flat amounts, whereas this is a %. Maybe `dustFractionThreshold`?

packages/asset-swapper/src/utils/aggregate.ts
```diff
@@ -0,0 +1,800 @@
...
+    Eth2Dai,
+    Kyber,
+    NumSources,
+}
+
+/**
+ * Convert a source to a canonical address used by the sampler contract.
+ */
+export const SOURCE_TO_ADDRESS: { [key: string]: string } = {
+    [ERC20BridgeSource.Eth2Dai]: '0x39755357759cE0d7f32dC8dC45414CCa409AE24e',
```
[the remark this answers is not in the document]
dorothy-zbornak: Yes, for sure, once these are actually all deployed.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: `pickOptimalPath` just picks the shortest path of those passed in, but `FillsOptimizer`'s job is to actually generate new paths so idk if that's a great fit.

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: As a rule, yes. In practice, we often see quotes up to and beyond $10k, for high liquidity pairs, that are virtually linear (i.e., rates are essentially constant).

dorothy-zbornak: Well, as a rule except for maybe Kyber, which is why we fall back to the flood fill.

dorothy-zbornak:
> This looks awesome! Only minor feedback on the design of the algorithm.
> 
> Thought though: @dekz @dorothy-zbornak
> While this algorithm searches for optimal paths with random walks, can we optimize further with a linear algebra interpretation of the paths?
> 
> I am thinking that with some pruning we can get a rather sparse matrix representation of the edges and Fill Paths, with some modified form of dijkstra's algorithm or some other optimization algorithm and an efficient sparse matrix typescript library, we may be able to see performance gains?
> 
> Just a shower thought, maybe for a later upgrade if performance becomes an issue.

Yeah, there is no doubt we can optimize this algo to reduce the space complexity. The pruning done here is not super aggressive. One obvious one we can do for buys is avoid traversing paths that would exceed the global minimum cost. There are some issues with using LA to find ideal splits, like not being able to partially fill orders (except for the last) with market functions, so we might have to go down the route of having a custom forwarder for `asset-swapper`. 

Another thing worth noting is that in most cases (knock on wood), this algo does approximate the globally optimal solution early on, because the solution for combining convex quotes should just be discoverable using a simple greedy algorithm. Because the fill nodes are sorted by price, this solution is walked over sooner rather than later.

packages/asset-swapper/src/utils/aggregate.ts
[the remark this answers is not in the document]
dorothy-zbornak: Just realized these are DEX addresses. Yeah, idk, feels weird to put non 0x stuff in `contract-addresses`. :thinking:

packages/asset-swapper/src/utils/aggregate.ts
dorothy-zbornak: Turns out these are subtly different enough to make it not worth it.

dave4506:
Adding back the [WIP] tag, to finish the work with integrating `asset-swapper` with the aggregator utils and wrapping up the TODO tasks.

packages/contract-wrappers/CHANGELOG.json
```diff
@@ -5,6 +5,10 @@
             {
                 "note": "Add wrapper for `IAssetDataContract`",
                 "pr": 2373
+            },
+            {
+                "note": "Add `IERC20BridgeSampler` wrapper",
+                "pr": 2353
```
xianny: I think this needs a new entry; current entry is already published

packages/contract-artifacts/CHANGELOG.json
```diff
@@ -5,6 +5,10 @@
             {
                 "note": "Add `IAssetData` artifact",
                 "pr": 2373
+            },
```
xianny: this needs a new entry right? current entry is already published

packages/types/CHANGELOG.json
```diff
@@ -14,6 +14,10 @@
             {
                 "note": "Add `ERC20BridgeAssetData`",
                 "pr": 2373
```
xianny: same; this needs a new entry

dekz:
Thanks for catching that @xianny. I pushed a fix to that.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -0,0 +1,401 @@
...
+        if (d.gt(0)) {
+            return 1;
+        } else if (d.lt(0)) {
+            return -1;
+        }
+        return 0;
+    });
+}
+
+function getOrderTokens(order: SignedOrder): [string, string] {
```
dekz: TODO `assetDataUtils` is back

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface OrderDomain {
+    chainId: number;
+    exchangeAddress: string;
+}
+
+/**
+ * Common exception messages thrown by aggregation logic.
+ */
+export enum AggregationError {
+    NoOptimalPath = 'no optimal path',
```
dekz:
```suggestion
    NoOptimalPath = 'NO_OPTIMAL_PATH',
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -0,0 +1,116 @@
...
+    EmptyOrders = 'empty orders',
```
dekz:
```suggestion
    EmptyOrders = 'EMPTY_ORDERS',
```

packages/asset-swapper/src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,50 @@
...
+    private readonly _devUtils: DevUtilsContract;
+
+    constructor(devUtils: DevUtilsContract) {
+        this._devUtils = devUtils;
+    }
+
+    public async getSignedOrdersWithFillableAmountsAsync(
+        signedOrders: SignedOrder[],
+    ): Promise<SignedOrderWithFillableAmounts[]> {
+        const signatures = _.map(signedOrders, o => o.signature);
```
dekz:
```suggestion
        const signatures = signedOrders.map(o => o.signature);
```

packages/asset-swapper/src/utils/order_state_utils.ts
```diff
@@ -0,0 +1,50 @@
+import { DevUtilsContract } from '@0x/contract-wrappers';
+import { orderCalculationUtils } from '@0x/order-utils';
+import { OrderStatus, SignedOrder } from '@0x/types';
+import * as _ from 'lodash';
```
dekz:
```suggestion

```

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -0,0 +1,722 @@
...
+            ],
+        });
+    }
+
+    function createSamplerFromBuyRates(rates: Numberish[][]): MockSamplerContract {
+        return new MockSamplerContract({
+            queryOrdersAndSampleBuys: (orders, signatures, sources, fillAmounts) => [
+                orders.map(o => o.makerAssetAmount),
+                sources.map(s =>
+                    shortZip(fillAmounts, rates[getSourceFromAddress(s)]).map(([f, r]) =>
```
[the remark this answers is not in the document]
dorothy-zbornak: I can rewrite this to be clearer.

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -0,0 +1,722 @@
...
+            makerAssetAmount: getRandomInteger(1, 1e18),
+            takerAssetAmount: getRandomInteger(1, 1e18),
+            makerFee: getRandomInteger(1, 1e17),
+            takerFee: getRandomInteger(1, 1e17),
+            signature: hexUtils.random(),
+            ...overrides,
+        };
+    }
+
+    function getSourceFromAssetData(assetData: string): ERC20BridgeSource {
```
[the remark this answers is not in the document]
dorothy-zbornak: Probably not this version though; it's not very thorough.

packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
```diff
@@ -0,0 +1,168 @@
...
+        takerAssetData: assetDataUtils.encodeERC20AssetData(takerToken),
+        ...createCommonOrderFields(orderDomain, makerAssetAmount, takerAssetAmount, slippage, isBuy),
+    };
+}
+
+/**
+ * Create ERC20Bridge asset data.
+ */
+export function createBridgeAssetData(tokenAddress: string, bridgeAddress: string, bridgeData: string): string {
+    const encoder = AbiEncoder.createMethod('ERC20Bridge', [
```
dorothy-zbornak: TODO once we get bridge stuff in `assetDataUtils`.

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -0,0 +1,722 @@
...
+            default:
+                break;
+        }
+        throw new Error(`Unknown bridge address: ${bridgeAddress}`);
+    }
+
+    function getSourceFromAddress(sourceAddress: string): ERC20BridgeSource {
+        for (const k of Object.keys(SOURCE_TO_ADDRESS)) {
+            if (SOURCE_TO_ADDRESS[k].toLowerCase() === sourceAddress.toLowerCase()) {
+                return parseInt(k, 10) as ERC20BridgeSource;
```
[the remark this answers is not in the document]
dorothy-zbornak: Made the change, although what are you even trying to do there, you lunatic.

packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
xianny: It's already in there :)

PR #2356, opened 2019-11-20
`@0x/contracts-exchange-forwarder`: ERC20Bridge buy support in Forwarder

mzhu25:
## Description
Adds support for ERC20Bridge orders in the Forwarder. The changes to the Forwarder contracts themselves are pretty small, the only notable exception is the addition of an `_isPercentageFee` function which checks whether the `takerFeeAssetData` is equalish to the `makerAssetData`, e.g. we support orders where the maker asset if from an ERC20 bridge and the taker fee asset is the non-bridge equivalent. 
The bulk of the code additions are from test stuff, including:
- unit tests for `_isPercentageFee` and `_transferAssetToSender`
- Forwarder <> bridge integration tests. The tests basically use the real bridge proxy and bridge contracts, which interact with stubbed out versions of Uniswap and Eth2Dai. These integration tests also required a bunch of scattered changes
- Added the `ERC20BridgeProxy` to the `DeploymentManager`

## Checklist:


contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -316,4 +316,40 @@ contract MixinExchangeWrapper is
...
+        bytes memory makerAssetData
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        bytes4 takerFeeAssetProxyId = takerFeeAssetData.readBytes4(0);
+        // If the takerFee asset is not ERC20, it cannot be a percentage fee (and will revert with
+        // an UnsupportedFeeError in the calling function).
+        if (takerFeeAssetProxyId != ERC20_PROXY_ID) {
```
abandeali1: There's probably some edge case where it's possible that the `takerFeeAssetProxyId == ERC20_BRIDGE_PROXY_ID`. It's probably simpler if we just check:

```
if (
    (takerFeeAssetProxyId == ERC20_PROXY_ID || takerFeeAssetProxyId == ERC20_PROXY_ID) &&
    (makerAssetProxyId == ERC20_PROXY_ID || makerAssetProxyId == ERC20_PROXY_ID)
) { ... }
```

contracts/exchange-forwarder/contracts/src/libs/LibConstants.sol
```diff
@@ -27,28 +27,26 @@ contract LibConstants {
 
     using LibBytes for bytes;
 
-    bytes4 constant internal ERC20_DATA_ID = bytes4(keccak256("ERC20Token(address)"));
-    bytes4 constant internal ERC721_DATA_ID = bytes4(keccak256("ERC721Token(address,uint256)"));
+    bytes4 constant internal ERC20_PROXY_ID = bytes4(keccak256("ERC20Token(address)"));
```
abandeali1: We can probably remove these in favor of the `IAssetData(address(0)).ERC20Token.selector` pattern using https://github.com/0xProject/0x-monorepo/blob/development/contracts/asset-proxy/contracts/src/interfaces/IAssetData.sol.

contracts/exchange-forwarder/contracts/src/MixinAssets.sol
```diff
@@ -85,10 +81,12 @@ contract MixinAssets is
...
         bytes4 proxyId = assetData.readBytes4(0);
 
-        if (proxyId == ERC20_DATA_ID) {
+        if (proxyId == ERC20_PROXY_ID) {
             _transferERC20Token(assetData, amount);
-        } else if (proxyId == ERC721_DATA_ID) {
+        } else if (proxyId == ERC721_PROXY_ID) {
             _transferERC721Token(assetData, amount);
+        } else if (proxyId == ERC20_BRIDGE_PROXY_ID) {
+            _transferERC20BridgeAsset(assetData, amount);
```
abandeali1: Can we just reuse `_transferERC20Token` here?

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -88,7 +88,7 @@ contract MixinExchangeWrapper is
         )
     {
         // No taker fee or percentage fee
-        if (order.takerFee == 0 || order.takerFeeAssetData.equals(order.makerAssetData)) {
+        if (order.takerFee == 0 || _isPercentageFee(order.takerFeeAssetData, order.makerAssetData)) {
```
abandeali1: It seems unlikely but possible that one of the taker `assetDatas` are using the `ERC20BridgeProxy`, in which case we should use this function for the `if (order.takerFeeAssetData.equals(order.takerAssetData))` check too.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -316,4 +316,40 @@ contract MixinExchangeWrapper is
...
         }
     }
+
+    /// @dev Checks whether an order's taker fee is effectively denominated in the maker asset.
+    ///      This is the case if they have the same ERC20Proxy asset data, or if the makerAssetData
+    ///      is the ERC20Bridge equivalent of the takerFeeAssetData.
+    /// @param takerFeeAssetData Byte array encoded for the takerFee asset proxy.
+    /// @param makerAssetData Byte array encoded for the maker asset proxy.
+    /// @return isPercentageFee Whether or not the taker fee asset matches the maker asset.
+    function _isPercentageFee(
```
abandeali1: I might rename this something more generic, like `areAssetDataTransfersEqual`. I could see it being used elsewhere down the line.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
mzhu25: I thought it was impossible for the `takerAssetData` or `takerFeeAssetData` to use the bridge proxy? The bridges require a balance of the opposing asset when performing a `bridgeTransferFrom`, but since the Forwarder contract is the taker, it will be the one that holds a `makerAsset` balance when the `ERC20BridgeProxy` is dispatched

abandeali1: This is only true if using the DEX aggregation pattern where the bridge is also the maker. However, other bridge contracts can require some sort of allowances and pull from the user's account (which would be true of margin limit orders).

mzhu25: ah yeah, I forgot about that side of things

contracts/exchange-forwarder/CHANGELOG.json
```diff
@@ -1,4 +1,14 @@
 [
+    {
+        "version": "3.1.0-beta.4",
+        "changes": [
+            {
+                "note": "Added buy support for ERC20Bridge",
+                "pr": 2356
+            }
+        ],
+        "timestamp": 1574461784
```
dorothy-zbornak: No timestamp for unpublished versions.

contracts/integrations/test/forwarder/bridge_test.ts
```diff
@@ -0,0 +1,378 @@
...
+import { Taker } from '../framework/actors/taker';
+import { actorAddressesByName } from '../framework/actors/utils';
+import { BlockchainBalanceStore } from '../framework/balances/blockchain_balance_store';
+import { DeploymentManager } from '../framework/deployment_manager';
+import { TestEth2DaiContract, TestUniswapExchangeContract } from '../wrappers';
+
+import { deployForwarderAsync } from './deploy_forwarder';
+import { ForwarderTestFactory } from './forwarder_test_factory';
+
+blockchainTests.resets.only('Forwarder <> ERC20Bridge integration tests', env => {
```
dorothy-zbornak:
```suggestion
blockchainTests.resets('Forwarder <> ERC20Bridge integration tests', env => {
```

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -173,6 +190,15 @@ contract MixinExchangeWrapper is
                 remainingTakerAssetFillAmount
             );
 
+            // Account for the ERC20Bridge transfering more of the maker asset than expected.
+            if (makerAssetProxyId == erc20BridgeProxyId) {
+                uint256 balanceAfter = IERC20Token(tokenAddress).balanceOf(address(this));
+                makerAssetAcquiredAmount = LibSafeMath.max256(
```
dorothy-zbornak: Why don't we just set `makerAssetAcquiredAmount` to `max(makerAssetAcquiredAmount, balanceAfter)`? Would that be so wrong?

mzhu25: two edge cases: 
- someone accidentally transfers tokens to the Forwarder
- `makerAsset` is WETH, which seems silly but given the flexibility of the bridge pattern I could see there being a use case for this at some point 🤷‍♂

PR #2362, opened 2019-11-22
Re-export the RevertErrors

dekz:
## Description

Re-export the `RevertErrors` from `@0x/utils` to assist in decoding when integrating/running in production.

Possible future plan can be found here: https://gist.github.com/dekz/3f6a586b4de6da9bd6d43da9addcbe42

### Explanation

When a `eth_call` is performed, the [result is checked](https://github.com/0xProject/0x-monorepo/blob/development/packages/base-contract/src/index.ts#L291) for a `RevertError`. 

RevertError looks up its global registry for the `bytes4` revert. If it is found it is decoded and thrown.

RevertErrors are loaded when the [file is imported](https://github.com/0xProject/0x-monorepo/blob/development/contracts/exchange-libs/src/lib_math_revert_errors.ts#L30). This means that for Exchange wrapper to decode a `LibBytesError` the developer must've loaded a `contracts-*` package which exported the `LibBytesError`. When using `abi-gen-wrappers/contract-wrappers`, these `contract-*` packages are not used directly (due to large ABI output files). A developer strictly using `abi-gen-wrappers/contract-wrappers` who encounters an `eth_call` RevertError would not be able to decode the error as they have not been registered. RevertError exists, but no errors have been registered.

This PR moves the RevertErrors into Utils, along side RevertError so anyone using the Utils package to decode RevertErrors (base-contract) is able to decode them.


contracts/exchange-libs/src/reference_functions.ts
```diff
@@ -2,10 +2,10 @@ import { ReferenceFunctions } from '@0x/contracts-utils';
 import { FillResults, Order } from '@0x/types';
 import { BigNumber } from '@0x/utils';
 
-import LibMathRevertErrors = require('./lib_math_revert_errors');
-
 const { safeAdd, safeSub, safeMul, safeDiv } = ReferenceFunctions;
 
+import { LibMathRevertErrors } from '../src';
```
fabioberger: Please do not import from `../src` anywhere. This makes future refactors difficult and is an anti-pattern.

dekz: Why is it an anti-pattern? It's also testing that it is actually exported in the `index.ts` for consumers.

abandeali1: I'm not sure why it's considered an anti-pattern, but it does seem cleaner and more readable to just import from `@0x/utils` to me.

contracts/utils/test/lib_rich_errors.ts
```diff
@@ -1,39 +0,0 @@
-import { blockchainTests, expect, hexRandom } from '@0x/contracts-test-utils';
-import { coerceThrownErrorAsRevertError, StringRevertError } from '@0x/utils';
-
-import { artifacts } from './artifacts';
-import { TestLibRichErrorsContract } from './wrappers';
-
-blockchainTests('LibRichErrors', env => {
```
abandeali1: Why was this file deleted?

PR #2365, opened 2019-11-22
dYdX Bridge

hysz:
## Description

This PR implements the dYdX bridge contract. The bridge can execute two actions on an account: deposit or withdrawal.

Usage:
```
Order: {
...
takerAssetData: erc20Transfer()
makerAssetData: bridgeTransfer([deposit to dydx from maker, withdraw from dydx to taker])
}
```

Notes:
```
1. This bridge must be set as an operator of the input dydx account.
2. The `bridgeTransferFrom` may only be called in the context of the 0x Exchange.
3. The maker or taker of the 0x order must be the dydx account owner.
4. Deposits into dydx are made from the `from` address.
5. Withdrawals from dydx are made to the `to` address.
```

We also explored the option of making this bridge contract the maker, which would relieve the need for the Multi-Asset Proxy. This worked by implementing `isValidSignature` in the bridge contract; however, this was very limiting. Asset data had to be encoded in a very specific way in order for us to decode and validate on-chain. Instead, we defer signature validation to the exchange by only allowing calls from the ERC20Bridge. 

## Checklist:


contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,179 @@
...
+        // Create dydx amount.
+        IDydx.AssetAmount memory amountToWithdraw = IDydx.AssetAmount({
+            sign: true,                                     // true if positive.
+            denomination: IDydx.AssetDenomination.Wei,      // Wei => actual token amount held in account.
+            ref: IDydx.AssetReference.Target,               // Target => an absolute amount.
+            value: amount                                   // amount to withdraw.
+        });
+
+        // Create withdraw action.
+        IDydx.ActionArgs memory withdrawAction = IDydx.ActionArgs({
```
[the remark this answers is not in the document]
abandeali1: Thinking about this more, I like keeping this simple for now and not adding additional actions.

contracts/asset-proxy/contracts/src/interfaces/IDydxBridge.sol
```diff
@@ -0,0 +1,35 @@
...
+interface IDydxBridge {
+
+    enum BridgeAction {
+        Deposit,                    // Deposit tokens into dydx account.
+        Withdraw                    // Withdraw tokens from dydx account.
+    }
+
+    struct BridgeData {
+        BridgeAction action;        // Action to run on dydx account.
+        address accountOwner;       // The owner of the dydx account.
```
abandeali1: I think we can probably get away with removing this.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,179 @@
...
+    /// @return success The magic bytes if successful.
+    function bridgeTransferFrom(
+        address,
+        address from,
+        address to,
+        uint256 amount,
+        bytes calldata encodedBridgeData
+    )
+        external
+        onlyAuthorized
```
abandeali1: I wonder if we should use `onlyAuthorized` here vs hard coding a `require(msg.sender == EXCHANGE)`. I don't think this contract needs to be upgradable or support multiple authorized addresses. There's probably an argument to be made for that, but my guess is this will be the first of many iterations before we want to worry about upgradability.

hysz: Sure I'm down to nix the dependency and use something more lightweight for now.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,179 @@
...
+    {
+        // Decode bridge data.
+        (BridgeData memory bridgeData) = abi.decode(encodedBridgeData, (BridgeData));
+
+        // Cache dydx contract.
+        IDydx dydx = IDydx(_getDydxAddress());
+
+        // Assert that `from` is the owner or an operator of the dydx account.
+        require(
+            from == bridgeData.accountOwner || dydx.getIsLocalOperator(bridgeData.accountOwner, from),
```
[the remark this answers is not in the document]
hysz: Mm yeah you're right, we can only allow `from` to be the account owner. So we can get rid of this check altogether. [source](https://github.com/dydxprotocol/solo/blob/master/contracts/protocol/impl/OperationImpl.sol#L318)

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,179 @@
...
+        internal
+        pure
+        returns (IDydx.ActionArgs memory)
+    {
+        // Create dydx amount.
+        IDydx.AssetAmount memory amountToDeposit = IDydx.AssetAmount({
+            sign: true,                                 // true if positive.
+            denomination: IDydx.AssetDenomination.Wei,  // Wei => actual token amount held in account.
+            ref: IDydx.AssetReference.Target,           // Target => an absolute amount.
+            value: amount                               // amount to deposit.
```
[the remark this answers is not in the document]
hysz: This introduces potential rounding errors because the `amount` parameter has already been rounded down in the exchange. Digging into this, it would have been introduced by the MAP version as well.

```
## In Exchange
The `amount` parameter is computed by the Exchange as:
  amount = floor[(takerAssetFillAmount * makerAssetAmount) / takerAssetAmount]

The amount to deposit is `takerAssetFillAmount`, which could be recomputed in the Bridge as:
  takerAssetFillAmount = (amount * takerAssetAmount) / makerAssetAmount

However since `amount` is rounded we cannot precisely extract the original `takerAssetAmount`:
  {
      makerAssetAmount: 3,
      takerAssetAmount: 11,
      takerAssetFillAmount: 5
  }

  In Exchange:
  amount = floor[(5 * 3) / 11] = floor[1.3636..] = 1

  In Bridge:
  amountToDeposit = takerAssetFillAmount = (1 * 11) / 3 = 3.6666 (neither rounding up nor down would give 5)
```

If we split the operations between the maker and taker asset data.
Doing this we transfer the exact values computed by the Exchange:
```
Order: {
  ...,
  takerAssetData: [ERC20 token to Maker, Bridge deposits from Maker (Maker=`to`)]
  makerAssetData: [Bridge withdraws to Taker (Maker=`from`)]
}
```

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,186 @@
...
+        // Cache dydx contract.
+        IDydx dydx = IDydx(_getDydxAddress());
+
+        // Create dydx action.
+        IDydx.AccountInfo[] memory accounts = new IDydx.AccountInfo[](1);
+        IDydx.ActionArgs[] memory actions = new IDydx.ActionArgs[](1);
+        if (bridgeData.action == BridgeAction.Deposit) {
+            // We interpret `to` as the 0x order maker and `from` as the taker.
+            accounts[0] = IDydx.AccountInfo({
+                owner: to,
```
dorothy-zbornak: Q: What happens if the someone creates an order with a deposit dydx asset data in the maker or taker *fee* asset data? Can filling the order effectively force the fee recipient into performing a deposit to their account?

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,186 @@
...
+        // Decode bridge data.
+        (BridgeData memory bridgeData) = abi.decode(encodedBridgeData, (BridgeData));
+
+        // Cache dydx contract.
+        IDydx dydx = IDydx(_getDydxAddress());
+
+        // Create dydx action.
+        IDydx.AccountInfo[] memory accounts = new IDydx.AccountInfo[](1);
+        IDydx.ActionArgs[] memory actions = new IDydx.ActionArgs[](1);
+        if (bridgeData.action == BridgeAction.Deposit) {
```
dorothy-zbornak: Actually, how would a deposit work? Wouldn't the bridge proxy revert because no tokens were transferred?

abandeali1: Hmm that's a good point. This seems to break the rules of the proxy. Can we write a test for this case before merging?

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
hysz: Ah yeah, that would be annoying. It's probably worth using the more complex logic to avoid people doing this. I can revert to the previous version that only allowed operations on the account owned by the `from` address.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
hysz: I've reverted to the previous design iteration, which always used `from` as the account owner and allowed actions to be bundled. It's a bit more complex than this one, but it's very flexible and should address both of your comments @dorothy-zbornak.

The 0x Order in this design looks like this:
```
{
...
takerAssetData: erc20Transfer()
makerAssetData: bridgeTransfer([deposit to dydx from maker, withdraw from dydx to taker])
}
```

hysz: Worth noting that the motivation for this ^ design was to make the bridge flexible and extensible. This allows us to iterate on the impl and experiment with other dydx actions, like executing callbacks. It also allows one 0x order to execute across several of the maker's dydx accounts. 

The alternative would be to have a single workflow that just does a basic deposit & withdrawal. I'm open to this bc it's simpler, but I do like the flexible version. Am looking forward to playing with it irl and potentially finding new use cases.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,239 @@
...
+    )
+        internal
+        pure
+        returns (
+            IDydx.ActionArgs memory withdrawAction
+        )
+    {
+        // Create dydx amount.
+        IDydx.AssetAmount memory amountToWithdraw = IDydx.AssetAmount({
+            sign: true,                                     // true if positive.
```
dorothy-zbornak: Could be wrong, but I was under the impression that this needed to be `false` for withdraws. Ref https://github.com/dydxprotocol/solo/blob/master/contracts/protocol/impl/OperationImpl.sol#L377

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,239 @@
...
+            denomination: IDydx.AssetDenomination.Wei,      // Wei => actual token amount held in account.
+            ref: IDydx.AssetReference.Target,               // Target => an absolute amount.
+            value: amount                                   // amount to withdraw.
+        });
+
+        // Create withdraw action.
+        withdrawAction = IDydx.ActionArgs({
+            actionType: IDydx.ActionType.Withdraw,          // withdraw tokens.
+            amount: amountToWithdraw,                       // amount to withdraw.
+            accountId: bridgeAction.accountId,              // index in the `accounts` when calling `operate`.
```
dorothy-zbornak: I guess this might also need validation like the deposit vector.

contracts/asset-proxy/contracts/src/interfaces/IDydxBridge.sol
```diff
@@ -0,0 +1,42 @@
...
+        Deposit,                    // Deposit tokens into dydx account.
+        Withdraw                    // Withdraw tokens from dydx account.
+    }
+
+    struct BridgeAction {
+        BridgeActionType actionType;            // Action to run on dydx account.
+        uint256 accountId;                      // Index in `BridgeData.accountNumbers` for this action.
+        uint256 marketId;                       // Market to operate on.
+        uint256 conversionRateNumerator;        // Optional. If set, transfer amount is scaled by (conversionRateNumerator/conversionRateDenominator).
+        uint256 conversionRateDenominator;      // Optional. If set, transfer amount is scaled by (conversionRateNumerator/conversionRateDenominator).
```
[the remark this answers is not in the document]
hysz: Yeah I thought about condensing this but figured this way would be more obvious for clients - although, I guess if this is how Kyber does it then I'm cool w it!

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,236 @@
...
+    DeploymentConstants
+{
+
+    using LibSafeMath for uint256;
+
+    /// @dev Callback for `IERC20Bridge`. Deposits or withdraws tokens from a dydx account.
+    ///      Notes:
+    ///         1. This bridge must be set as an operator of the input dydx account.
+    ///         2. This function may only be called in the context of the 0x Exchange.
+    ///         3. The maker or taker of the 0x order must be the dydx account owner.
```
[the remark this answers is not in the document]
hysz: That's the likely use case, although the taker could technically be the owner if this were used in the `takerAssetData` or `takerFeeAssetData`. I'll add a note re withdrawals.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -0,0 +1,236 @@
...
+            value: amount                               // amount to deposit.
+        });
+
+        // Create dydx deposit action.
+        depositAction = IDydx.ActionArgs({
+            actionType: IDydx.ActionType.Deposit,           // deposit tokens.
+            amount: dydxAmount,                             // amount to deposit.
+            accountId: bridgeAction.accountId,              // index in the `accounts` when calling `operate`.
+            primaryMarketId: bridgeAction.marketId,         // indicates which token to deposit.
+            otherAddress: depositFrom,                      // deposit from the account owner.
```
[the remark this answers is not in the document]
hysz: > Does dydx have any kind of whitelisting for which addresses can deposit into accounts?
Yes, see [here](https://github.com/dydxprotocol/solo/blob/master/contracts/protocol/impl/OperationImpl.sol#L316)
(i) The `otherAddress` must be either the sender or account owner, and
(ii) The sender must be an account operator.

> Is it practical for mesh, etc, to validate that the depositFrom address is somehow linked to the accountId?

The `accountId` here is actually just an offset into the `accounts` array (see [here](https://github.com/dydxprotocol/solo/blob/master/contracts/protocol/lib/Actions.sol#L284)), which we construct in this contract. A client-side check could validate that `from` is the account owner and that this bridge is a valid operator. However, we don't perform these checks here because dydx will revert in this scenario.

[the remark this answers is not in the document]
dorothy-zbornak: Ah, great :+1:

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
dorothy-zbornak: I guess this one is moot too since `accountId` is always going to point to one by the maker.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
hysz: I looked back at my test transactions and it looks like the encoding is different when operating on a portion of the balance vs the entire balance. This does need to be `false` for partial; I've updated this plus set the `ref` to `Delta` for each.

hysz: Mm maybe their client differentiates in order to avoid rounding errors in the contract.

contracts/asset-proxy/contracts/src/interfaces/IDydxBridge.sol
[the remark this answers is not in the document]
hysz: Actually, I think I prefer this so that the client just sets the numerator/denominator to the `takerAssetAmount` and `makerAssetAmount` (or vise versa). Plus we get to use the `safeGetPartialAmountFloor` to check for rounding errors.

contracts/asset-proxy/test/dydx_bridge.ts
```diff
@@ -0,0 +1,244 @@
...
+            const expectedOperateActionEvents = [];
+            for (const action of bridgeData.actions) {
+                expectedOperateActionEvents.push({
+                    actionType: action.actionType as number,
+                    accountId: action.accountId,
+                    amountSign: action.actionType === BridgeActionType.Deposit ? true : false,
+                    amountDenomination: weiDenomination,
+                    amountRef: deltaAmountRef,
+                    amountValue: action.conversionRateDenominator.gt(0)
+                        ? amount.times(action.conversionRateNumerator).div(action.conversionRateDenominator)
```
dorothy-zbornak: nit:
```suggestion
                        ? amount.times(action.conversionRateNumerator).div(action.conversionRateDenominator).integer(BigNumber.ROUND_DOWN)
```

PR #2366, opened 2019-11-23
Pool Member Fuzz Tests

jalextowle:
## Description

This PR introduces fuzz testing for members of staking pools. The actions that are tested in this fuzz test are a superset of the actions from the staking and pool operator fuzz tests. New actions include:
- joining staking pools
- filling orders

At the moment, the fill order implementation is pretty hacky and has some flaws. This said, this should provide a good base for future development.

## Checklist:


contracts/integrations/test/fuzz_tests/pool_membership_test.ts
```diff
@@ -0,0 +1,89 @@
...
+import { PoolMember } from '../framework/actors/pool_member';
+import { AssertionResult } from '../framework/assertions/function_assertion';
+import { BlockchainBalanceStore } from '../framework/balances/blockchain_balance_store';
+import { DeploymentManager } from '../framework/deployment_manager';
+import { Simulation, SimulationEnvironment } from '../framework/simulation';
+
+import { PoolManagementSimulation } from './pool_management_test';
+import { StakeManagementSimulation } from './stake_management_test';
+
+class PoolMembershipSimulation extends Simulation {
```
mzhu25: maybe like "MakerTakerSimulation" or something? idk

contracts/integrations/test/fuzz_tests/pool_membership_test.ts
```diff
@@ -0,0 +1,89 @@
...
+    protected async *_assertionGenerator(): AsyncIterableIterator<AssertionResult | void> {
+        const { deployment } = this.environment;
+
+        const poolManagement = new PoolManagementSimulation(this.environment);
+        const stakeManagement = new StakeManagementSimulation(this.environment);
```
mzhu25: I don't think we need this here, since stake shouldn't really affect joining pools

contracts/integrations/test/framework/assertions/function_assertion.ts
```diff
@@ -1,11 +1,15 @@
...
 
 // tslint:disable:max-classes-per-file
-
 export type GenericContractFunction<T> = (...args: any[]) => ContractFunctionObj<T>;
 
+export interface FunctionArguments<TArgs extends any[]> {
+    args: TArgs;
+    txData: Partial<TxData>;
+}
+
```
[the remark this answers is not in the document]
jalextowle: Yeah, we can probably do that now that args isn't a `rest` parameter. Ill try it out.

contracts/integrations/test/framework/assertions/fillOrder.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * A function assertion that verifies that a complete and valid fill succeeded and emitted the correct logs.
+ */
+/* tslint:disable:no-unnecessary-type-assertion */
+export function validFillOrderCompleteFillAssertion(
+    deployment: DeploymentManager,
+): FunctionAssertion<[Order, BigNumber, string], {}, FillResults> {
+    const exchange = deployment.exchange;
+
+    return new FunctionAssertion<[Order, BigNumber, string], {}, FillResults>(exchange.fillOrder.bind(exchange), {
```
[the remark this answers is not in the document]
jalextowle: We had a lifetime issue where the `self` and `this` variables were dropping important wrapper functions (like "_strictEncodeArgs").

contracts/integrations/test/fuzz_tests/pool_membership_test.ts
jalextowle: I like `PoolMembership` more than that 🤷‍♂

contracts/integrations/test/framework/assertions/fillOrder.ts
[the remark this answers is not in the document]
jalextowle: Thinking about this a little more, we need to re-engineer `FunctionAssertion` such that users don't have to worry about lifetime issues when using the framework. I don't think that it's high-pri, but the best way to accomplish this is probably to pass in the wrapper and the function name of the wrapper function that we care about. A hackier solution would also be to let `FunctionAssertion` "own" the contract wrapper as well as the contract wrapper function. This would ensure that the two objects have the same lifetime, but it's a little icky.

What do you think would be nicer? I can probably put this change into this PR.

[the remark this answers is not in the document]
mzhu25: Yeah the first sounds nicer to me, but we can probably punt on it for now

contracts/integrations/test/framework/assertions/fillOrder.ts
```diff
@@ -0,0 +1,97 @@
...
+        after: async (
+            _beforeInfo,
+            result: FunctionResult,
+            args: [Order, BigNumber, string],
+            txData: Partial<TxData>,
+        ) => {
```
hysz: I noticed that in the other assertions we validate both events and on-chain state, but here only events. Would it be beneficial to assert any on-chain state here, say balances?

jalextowle: Discussed offline: I'm trying to keep this minimal so that Michael has a fairly clean slate to work with. Adding a TODO

PR #2369, opened 2019-11-25
Python backports for OpenSea

feuGeneA:
OpenSea is integrating with 0x v2 (for now), and they've hit a bug and a lack of needed functionality, both of which have already been fixed/implemented in the v3/development branch, so this PR is to backport that functionality for them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -166,7 +166,7 @@ jobs:
         working_directory: ~/repo
         docker:
             - image: nikolaik/python-nodejs:python3.7-nodejs8
-            - image: 0xorg/ganache-cli:2.2.2
+            - image: 0xorg/ganache-cli:4.4.0-beta.1
             - image: 0xorg/launch-kit-backend:74bcc39
               environment:
                   RPC_URL: http://localhost:8545
--- a/python-packages/sra_client/test/relayer/docker-compose.yml
+++ b/python-packages/sra_client/test/relayer/docker-compose.yml
@@ -2,7 +2,7 @@
 version: '3'
 services:
     ganache:
-        image: "0xorg/ganache-cli:2.2.2"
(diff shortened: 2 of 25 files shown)


feuGeneA:
image

.circleci/config.yml
```diff
@@ -166,7 +166,7 @@ jobs:
         working_directory: ~/repo
         docker:
             - image: nikolaik/python-nodejs:python3.7-nodejs8
-            - image: 0xorg/ganache-cli:2.2.2
+            - image: 0xorg/ganache-cli:4.4.0-beta.1
```
fabioberger: Was this intentional?

python-packages/sra_client/test/relayer/docker-compose.yml
```diff
@@ -2,7 +2,7 @@
 version: '3'
 services:
     ganache:
-        image: "0xorg/ganache-cli:2.2.2"
+        image: "0xorg/ganache-cli:4.4.0-beta.1"
```
fabioberger: Was this intentional?

.circleci/config.yml
feuGeneA: Yes.  Adding in the test of `build_transaction()` ran me into a problem with Web3.py trying to do something with `eth_chainId`, and that wasn't a supported RPC method in the version of ganache-cli used in the 2.2.2 image.  The 4.4.0-beta.1 image has the updated ganache-cli, as well as the snapshot needed to support the tests, so I ran with it.

python-packages/sra_client/test/relayer/docker-compose.yml
feuGeneA: Yes, to stay in sync with the `.circleci/config.yml`

PR #2370, opened 2019-11-25
Deprecate abi-gen-wrappers

xianny:
## Description

This PR enables generating contract wrappers directly in `@0x/contract-wrappers` and deprecates `@0x/abi-gen-wrappers`. The motivation is to simplify the updating steps when new contracts are deployed, and reduce the overall number of different packages.

TypeScript interfaces remain the same, only developer workflow has changed.

The post-deployment workflow is mostly unchanged:
1. Update `@0x/contract-addresses`
2. Copy the contract artifact into `@0x/contract-artifacts`. If updating all the artifacts at once, you can use `cd packages/contract-artifacts && yarn artifacts_update`. If manually coping an artifact, make sure to use `yarn artifacts_transform` to remove unwanted fields.
3. Regenerate the wrappers. `cd packages/contract-wrappers && yarn rebuild`
4. Update changelogs in `contract-artifacts` and `contract-wrappers`

## Testing instructions

CI 🙏 

## Types of changes

* Bug fix 

## Checklist:


packages/contract-wrappers/src/index.ts
```diff
@@ -11,50 +48,61 @@ export {
...
-    CoordinatorRegistryContract,
-    IValidatorContract,
-    IWalletContract,
+    ExchangeProtocolFeeCollectorAddressEventArgs,
+    ExchangeProtocolFeeMultiplierEventArgs,
+    ExchangeTransactionExecutionEventArgs,
+} from './generated-wrappers/exchange';
+export { ForwarderContract } from './generated-wrappers/forwarder';
+export { IValidatorContract } from './generated-wrappers/i_validator';
+export { IWalletContract } from './generated-wrappers/i_wallet';
```
fabioberger: Let's remove IWalletContract, IValidatorContract, most devs don't use those. They _can_ get them from `contracts-*` if need be.

packages/contract-wrappers/src/index.ts
```diff
@@ -11,50 +48,61 @@ export {
...
+export {
+    StakingProxyAuthorizedAddressAddedEventArgs,
+    StakingProxyAuthorizedAddressRemovedEventArgs,
+    StakingProxyContract,
+    StakingProxyEventArgs,
+    StakingProxyEvents,
+    StakingProxyOwnershipTransferredEventArgs,
+    StakingProxyStakingContractAttachedToProxyEventArgs,
+    StakingProxyStakingContractDetachedFromProxyEventArgs,
+} from './generated-wrappers/staking_proxy';
```
fabioberger: Feeling similarly about staking contracts. Thoughts?

packages/contract-wrappers/src/index.ts
```diff
@@ -11,50 +48,61 @@ export {
...
-} from '@0x/abi-gen-wrappers';
-
+} from './generated-wrappers/weth9';
+export { CoordinatorContract } from './generated-wrappers/coordinator';
+export {
+    CoordinatorRegistryEventArgs,
+    CoordinatorRegistryEvents,
+    CoordinatorRegistryCoordinatorEndpointSetEventArgs,
+    CoordinatorRegistryContract,
+} from './generated-wrappers/coordinator_registry';
```
fabioberger: ...and CoordinatorRegistry.

xianny:
@fabioberger , do non-exported things get omitted from the docs? If so then yeah agree with removing seldom-used things. If not then I think there's no reason *not* to export all the generated wrappers from here so that they're all available if needed without the large size of `contracts-*` packages.

packages/contract-wrappers/src/index.ts
```diff
@@ -1,7 +1,31 @@
 export { ContractAddresses } from '@0x/contract-addresses';
 
 export { ContractWrappers } from './contract_wrappers';
-
+export { DevUtilsContract } from './generated-wrappers/dev_utils';
+export {
+    ERC1155MintableContract,
```
abandeali1: Why are we exporting the `Mintable` version of this contract rather than the regular version? We only export the regular versions for ERC20/721.

packages/contract-wrappers/src/index.ts
xianny: @dekz weighed in and let me know Staking/StakingProxy are needed!

packages/contract-wrappers/src/index.ts
xianny: Hm I don't know. This was copied from `abi-gen-wrappers`/ derived from `@0x/contract-artifacts`. In `contract-artifacts` there's only `ERC1155Mintable` and *not* `ERC1155Token`.

packages/contract-wrappers/src/index.ts
xianny: removed!

packages/contract-wrappers/src/index.ts
xianny: I think we can just remove `ERC1155MintableContract`. But should we add `ERC1155TokenContract`?

fabioberger: Yep, let's remove it and add ERC1155Token

packages/contract-wrappers/src/contract_wrappers.ts
```diff
@@ -98,7 +95,7 @@ export class ContractWrappers {
         this.weth9 = new WETH9Contract(contractAddresses.etherToken, this.getProvider());
         this.exchange = new ExchangeContract(contractAddresses.exchange, this.getProvider());
         this.forwarder = new ForwarderContract(contractAddresses.forwarder, this.getProvider());
-        this.orderValidator = new OrderValidatorContract(contractAddresses.orderValidator, this.getProvider());
+        this.staking = new StakingContract(contractAddresses.staking, this.getProvider());
```
abandeali1:
```suggestion
        this.staking = new StakingContract(contractAddresses.stakingProxy, this.getProvider());
```

PR #2373, opened 2019-11-27
Remove assetDataUtils everywhere

xianny:
## Description

I started removing uses of `assetDataUtils` a while ago, but left the task incomplete as it was easier to wait for things to be removed from the monorepo rather than refactor them.

This PR replaces the last few uses of `assetDataUtils` with `DevUtilsContract` and removes `assetDataUtils` completely. 

After making a few needed fixes, this PR has grown somewhat unwieldy:

- moves `hex_utils.ts` from @0x/contracts-test-utils to @0x/utils
- removes `assetDataUtils` from `@0x/order-utils`
- removes/replaces uses of `assetDataUtils` everywhere in monorepo
- re-implements `decodeAssetDataOrThrow` in `@0x/order-utils`, this time with ERC20Bridge support
- adds `ERC20BridgeAssetData` type and interface to @0x/types

Apologies for combining the `hex_utils` rename in this PR. Trying to get a fix in sooner rather than later as the currently published `assetDataUtils` does not support `ERC20Bridge`. 

## Checklist:


packages/instant/src/index.umd.ts
```diff
@@ -108,15 +108,15 @@ export interface ZeroExInstantConfig extends ZeroExInstantOverlayProps {
     shouldDisablePushToHistory?: boolean;
 }
 
-export const render = (config: ZeroExInstantConfig, selector: string = DEFAULT_ZERO_EX_CONTAINER_SELECTOR) => {
```
xianny: I couldn't find where this was being called, so didn't rename it to `renderAsync`. Worried this breaks something else. Where is this used? @fragosti @steveklebanoff

packages/instant/src/util/dev_utils_contract.ts
```diff
@@ -0,0 +1,5 @@
+import { DevUtilsContract } from '@0x/contract-wrappers';
+import { NULL_ADDRESS } from '@0x/utils';
+
```
fragosti: Maybe a little comment about what is going on.

packages/instant/src/index.umd.ts
```diff
@@ -156,14 +156,17 @@ export const ERC721_PROXY_ID = AssetProxyId.ERC721;
...
 };
 
-export const assetDataForERC721TokenAddress = (tokenAddress: string, tokenId: string | number): string => {
+export const assetDataForERC721TokenAddressAsync = async (
+    tokenAddress: string,
+    tokenId: string | number,
+): Promise<string> => {
     assert.isETHAddressHex('tokenAddress', tokenAddress);
-    return assetDataUtils.encodeERC721AssetData(tokenAddress, new BigNumber(tokenId));
+    return devUtilsContract.encodeERC721AssetData(tokenAddress, new BigNumber(tokenId)).callAsync();
```
fragosti: This is a significant breaking change. Obviously I'm lacking context on this decision but it seems odd that you can no longer get assetDatas in a synchronous way.

fabioberger: That is a good point @fragosti. @xianny we could change the method signature in the wrapper to be synchronous if the method is pure, no?

xianny: Not easily, we currently use `putContractCode` and `runCall` from `ethereumjs-vm` and they are async. 

@jalextowle showed me a synchronous work-around using `IAssetDataContract` though:

```
const assetDataEncoder = new IAssetDataContract(NULL_ADDRESS, fakeProvider);
const encoded = assetDataEncoder.ERC20Token(tokenAddress).getABIEncodedTransactionData()
```

packages/instant/src/index.umd.ts
```diff
@@ -158,12 +158,12 @@ export const ERC20_PROXY_ID = AssetProxyId.ERC20;
 
 export const assetDataForERC20TokenAddress = (tokenAddress: string): string => {
     assert.isETHAddressHex('tokenAddress', tokenAddress);
-    return assetDataUtils.encodeERC20AssetData(tokenAddress);
+    return assetDataEncoder.ERC20Token(tokenAddress).getABIEncodedTransactionData();
```
fragosti: 😍 so much better!

packages/instant/src/util/asset_data_encoder.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Returns the first four bytes of a given hex string.
+ * Does not have any validations, so the hex string MUST
+ * be a valid asset data, or this function returns garbage
+ * without throwing errors.
+ * @param assetData A hex string where the first four bytes are a valid Asset Proxy Id
+ *
+ */
+export function decodeAssetProxyId(assetData: string): string {
+    return hexSlice(assetData, 0, 4);
+}
```
fragosti: Shouldn't something like this be in `order-utils`

fragosti:
Still feel like I'm lacking context though. Why are we removing `assetDataUtils`? What is the standard way of creating asset datas now?

xianny:
@fragosti the motivation is to stop maintaining separate typescript libs for encoding/decoding asset data, and only have the code implemented and tested once in Solidity. Cutting unnecessary overhead is gonna save us time in the future; e.g. in this instance we already neglected to add ERC20Bridge support, but it's already implemented in Solidity.

@fabio @dekz btw, I think we should encourage devs to use `IAssetDataContract` instead of `DevUtilsContract`, especially if they are only going to be using DUC for asset data encoding/decoding. It's much faster because it doesn't have to instantiate the EVM and load the ABI.

packages/instant/src/util/asset_data_encoder.ts
xianny: Maybe but right now nothing else uses it and `instant` doesn't use `order-utils` as a dependency yet...

contracts/coordinator/test/mixins.ts
```diff
@@ -4,15 +4,13 @@ import {
...
-    hexConcat,
-    hexSlice,
     randomAddress,
     TransactionFactory,
     transactionHashUtils,
 } from '@0x/contracts-test-utils';
 import { LibBytesRevertErrors } from '@0x/contracts-utils';
 import { SignatureType, SignedOrder } from '@0x/types';
-import { BigNumber, CoordinatorRevertErrors } from '@0x/utils';
+import { BigNumber, CoordinatorRevertErrors, hexConcat, hexSlice } from '@0x/utils';
```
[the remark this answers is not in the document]
xianny: done!

PR #2375, opened 2019-12-01
Fix for typo in constructor and gas price to apply to meta transaction.

Arctek:
## Description

Fix for typo in constructor and gas price to apply to meta transaction for the coordinator client.


abandeali1:
Thanks for the fixes!

PR #2383, opened 2019-12-03
Python publish for v3

### LINKED ISSUE #2299: Checklist for final Python publish of 3.0 support
This issue is a checklist of items that need to be resolved or corrected before publishing finalized Python packages that support v3 of the protocol.

-   [ ] Update release dates in all CHANGELOGs
-   [ ] Publish docs and update doc links everywhere. For example, in contract_wrappers/src/index.rst, there are lots of references to https://0x.org/docs/contracts#..., which seem to be broken.
-   [ ] Consider revising docker commands in documentation (docstrings) because they don't work as stated right now (because you need to specify the ganache version and snapshot name).  Or skip that because it will "just work" later when we publish a stable v3 snapshot/docker-image?
-   [ ] Consider revising example SRA URL in `python-packages/sra_client/src/zero_ex/sra_client/__init__.py`.  Currently it points to `https://api.radarrelay.com/0x/v3` (changed recently from `.../v2`).  Does that even necessarily exist?  Would it be better to use a 0x API URL instead?
-   [ ] Remove `--pre` flag (which says "use pre-release versions") from all tox.ini files' `run_tests_against_deployment` targets.
-   [ ] Remove v3-specific tags on docker image references, in both `sra_client/test/relayer/docker-compose.yml` and `.circleci/config.yml`, referring just to `latest` tags.
-   [ ] Remove all `setup.py` version pins for `python-packages/` interdependencies.

feuGeneA:
Fixes #2299


python-packages/contract_wrappers/src/zero_ex/contract_wrappers/__init__.py
```diff
@@ -21,7 +21,7 @@
 To replicate these examples, one can use the `0xorg/ganache-cli`:code: docker
 image, which comes with the 0x contracts pre-deployed.  To start it::
 
-    docker run docker run -d -p 8545:8545 0xorg/ganache-cli
+    docker run -d -p 8545:8545 -e SNAPSHOT_NAME=0x_ganache_snapshot-v3-beta 0xorg/ganache-cli
```
fabioberger: Ah, can we rename the snapshot without the v3-beta suffix?

fabioberger: Maybe open a follow-up issue so we can roll this out?

feuGeneA: Done.

feuGeneA: I had wanted to do that initially, but the ganache image wasn't ready for that yet, but now it is so I changed it.

PR #2386, opened 2019-12-04
Fix unbound method in provider standardizer

johnrjj:
## Checklist:

Yesterday, Metamask (as of 7.7.0 https://github.com/MetaMask/metamask-extension/releases/tag/v7.7.0) changed the internals of their provider (using prototypes) [see 
 [https://github.com/MetaMask/metamask-inpage-provider/blob/master/index.js#L285-L301](https://github.com/MetaMask/metamask-inpage-provider/blob/master/index.js#L285-L301) ] and the way we pull off the methods of the provider object when standardizing the providing make JavaScript lose the `this` context without binding, which breaks our implementation. This PR fixes that.

### DIFF
--- a/packages/utils/src/provider_utils.ts
+++ b/packages/utils/src/provider_utils.ts
@@ -42,6 +42,10 @@ export const providerUtils = {
             enable: (supportedProvider as any).enable,
             sendAsync: _.noop.bind(_), // Will be replaced
         };
+        if (provider.enable) {
+            // Need to bind, metamask can lose reference to function without binding as of 7.7.0
+            provider.enable.bind(supportedProvider);
+        }
         // Case 1: We've already converted to our ZeroExProvider so noop.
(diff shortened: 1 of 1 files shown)


packages/utils/src/provider_utils.ts
```diff
@@ -42,6 +42,10 @@ export const providerUtils = {
             enable: (supportedProvider as any).enable,
             sendAsync: _.noop.bind(_), // Will be replaced
         };
+        if (provider.enable) {
+            // Need to bind, metamask can lose reference to function without binding as of 7.7.0
+            provider.enable.bind(supportedProvider)
```
kimpers: I think bind creates a new fn so you may need to reassign `provider.enable`
```suggestion
            provider.enable = provider.enable.bind(supportedProvider)
```

johnrjj: blah! that's what I meant to do :) thanks

PR #2387, opened 2019-12-04
`@0x/contracts-integrations`: Staking rewards fuzz test

mzhu25:
## Description
The much-anticipated 4th installation of the fuzz test series: staking rewards. This PR introduces:
- function assertions for `endEpoch`, `finalizePool`, and `withdrawDelegatorRewards`
- corresponding assertion generators in the `Keeper` and `Staker` actor mixins
- `staking_rewards_test.ts`, where it all comes together

In addition, this PR updates the `Taker` actor mixin to fix an issue in the market making/taking fuzz test that causes takers to run out of ETH after a while

## Testing instructions
```
PKG=@0x/contracts-integrations yarn build
cd contracts/integrations
FUZZ_TEST=staking_rewards yarn test:fuzz
```

## Checklist:


contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,269 @@
...
+            ] = await stakingWrapper.getMostRecentCumulativeReward(poolId).callAsync();
+            expect(cumulativeRewardsLastStored).to.bignumber.equal(currentEpoch);
+            let [numerator, denominator] = ReferenceFunctions.LibFractions.add(
+                beforeInfo.mostRecentCumulativeRewards.numerator,
+                beforeInfo.mostRecentCumulativeRewards.denominator,
+                membersReward,
+                beforeInfo.poolStats.membersStake,
+            );
+            [numerator, denominator] = ReferenceFunctions.LibFractions.normalize(numerator, denominator);
+            // There's a bug in our reference functions, but I can't figure it out :/
```
mzhu25: I suspect that the bug is that the reference `safeDiv` truncates in base 10, while solidity truncates in binary

contracts/exchange-libs/src/reference_functions.ts
```diff
@@ -113,3 +113,30 @@ export function calculateFillResults(
         protocolFeePaid: safeMul(protocolFeeMultiplier, gasPrice),
     };
 }
+
+export const LibFractions = {
```
hysz: I feel like @dorothy-zbornak added a math library to `contracts-utils` or `contracts-test-utils` that we could merge this into.

contracts/integrations/test/framework/actors/keeper.ts
```diff
@@ -75,6 +84,51 @@ export function KeeperMixin<TBase extends Constructor>(Base: TBase): TBase & Con
...
+                    yield assertion.executeAsync([], { from: this.actor.address });
+                }
+            }
+        }
+
+        private async _fastForwardToNextEpochAsync(): Promise<void> {
+            const { stakingWrapper } = this.actor.deployment.staking;
+
+            // increase timestamp of next block by how many seconds we need to
+            // get to the next epoch.
```
hysz: Is this logic removed from the staking api wrapper? O/w maybe we could call it here #DRY?

contracts/integrations/test/framework/actors/taker.ts
```diff
@@ -61,32 +64,69 @@ export function TakerMixin<TBase extends Constructor>(Base: TBase): TBase & Cons
...
+            const { actors, balanceStore } = this.actor.simulationEnvironment!;
+            const assertion = validFillOrderAssertion(this.actor.deployment, this.actor.simulationEnvironment!);
             while (true) {
-                const maker = Pseudorandom.sample(marketMakers);
+                // Choose a maker to be the other side of the order
+                const maker = Pseudorandom.sample(filterActorsByRole(actors, Maker));
                 if (maker === undefined) {
-                    yield undefined;
+                    yield;
                 } else {
```
hysz: Nit - this is a pretty long nested block. Could probably just omit the `else` and put this logic after the if smt. Or move it to a helper perhaps.

contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,269 @@
...
+} from '@0x/contracts-test-utils';
+import { BigNumber } from '@0x/utils';
+
+import { DeploymentManager } from '../deployment_manager';
+import { SimulationEnvironment } from '../simulation';
+
+import { FunctionAssertion, FunctionResult } from './function_assertion';
+
+const PRECISION = 15;
+const ALPHA_NUMERATOR = 1;
```
hysz: If you wanted to use the deployed values they're in [staking/src/constants.ts](https://github.com/0xProject/0x-monorepo/tree/development/contracts/staking/src/constants.ts). I believe the deployed value for alpha is `1/3`.

contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,269 @@
...
+            };
+        },
+        after: async (beforeInfo: FinalizePoolBeforeInfo, result: FunctionResult, args: [string]) => {
+            // Ensure that the tx succeeded.
+            expect(result.success, `Error: ${result.data}`).to.be.true();
+
+            const logs = result.receipt!.logs; // tslint:disable-line:no-non-null-assertion
+
+            // // Compute relevant epochs
+            // uint256 currentEpoch_ = currentEpoch;
```
hysz: Should these be removed?

contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,269 @@
...
+            // uint256 prevEpoch = currentEpoch_.safeSub(1);
+            const { stakingPools, currentEpoch } = simulationEnvironment;
+            const prevEpoch = currentEpoch.minus(1);
+            const [poolId] = args;
+            const pool = stakingPools[poolId];
+
+            // // Load the aggregated stats into memory; noop if no pools to finalize.
```
hysz: Should this be removed? Or if you wanted to keep the Solidity then could be good to add a note for future devs.

contracts/integrations/test/fuzz_tests/staking_rewards_test.ts
```diff
@@ -0,0 +1,125 @@
...
+            stakeManagement.generator,
+        ];
+        while (true) {
+            const action = Pseudorandom.sample(actions);
+            yield (await action!.next()).value; // tslint:disable-line:no-non-null-assertion
+        }
+    }
+}
+
+blockchainTests('Staking rewards fuzz test', env => {
```
hysz: Do we need a `.skip` here? (I recall it was previously included w other fuzz tests).

contracts/integrations/test/framework/assertions/withdrawDelegatorRewards.ts
```diff
@@ -0,0 +1,76 @@
...
+                WETH9Events.Transfer,
+            );
+            const expectedPoolRewards =
+                transferEvents.length > 0
+                    ? beforeInfo.poolRewards.minus(transferEvents[0]._value)
+                    : beforeInfo.poolRewards;
+            const poolRewards = await stakingWrapper.rewardsByPoolId(poolId).callAsync();
+            expect(poolRewards).to.bignumber.equal(expectedPoolRewards);
+
+            // TODO: Check CR
```
hysz: Can we just ensure there's an asana task for this so it won't get lost?

hysz: Could also be good to check delegator's balance. Or maybe check for an ERC20 event.

contracts/integrations/test/framework/assertions/stake.ts
```diff
@@ -50,33 +39,34 @@ export function validStakeAssertion(
...
             const ownerUndelegatedStake = await stakingWrapper
-                .getOwnerStakeByStatus(txData.from!, StakeStatus.Undelegated) // tslint:disable-line:no-non-null-assertion
-                .callAsync();
-            const expectedOwnerUndelegatedStake = expectedUndelegatedStake(ownerStake, amount);
-            expect(ownerUndelegatedStake, 'Owner undelegated stake').to.deep.equal(expectedOwnerUndelegatedStake);
-            // Updates local state accordingly
-            ownerStake[StakeStatus.Undelegated] = expectedOwnerUndelegatedStake;
-
-            // Checks that the global undelegated stake has also increased by the stake amount
-            const globalUndelegatedStake = await stakingWrapper
```
hysz: Why remove the global checks?

contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,269 @@
...
+            ] = await stakingWrapper.getMostRecentCumulativeReward(poolId).callAsync();
+            return {
+                poolStats,
+                aggregatedStats,
+                poolRewards,
+                cumulativeRewardsLastStored,
+                mostRecentCumulativeRewards,
+            };
+        },
+        after: async (beforeInfo: FinalizePoolBeforeInfo, result: FunctionResult, args: [string]) => {
```
hysz: Optional nit - could be more readable to split this function up. It's a couple hundy lines long.

contracts/integrations/test/framework/assertions/finalizePool.ts
hysz: 🤔 ... I'm down to look into this w you tomorrow!

contracts/exchange/test/reference_functions.ts
```diff
@@ -107,23 +107,6 @@ describe('Reference functions', () => {
             ).to.throw(expectedError.message);
         });
 
-        it('reverts if `order.makerAssetAmount` is 0', () => {
```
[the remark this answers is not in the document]
mzhu25: this is actually not the correct behavior for the reference function, since the `makerAssetFilledAmount` is computed as `makerAssetAmount * takerAssetFilledAmount / takerAssetAmount`

contracts/integrations/test/fuzz_tests/staking_rewards_test.ts
mzhu25: nah the `before` block ensures that it only runs if the `FUZZ_TEST` env variable is set

contracts/integrations/test/framework/actors/keeper.ts
mzhu25: I'd like to eventually get rid of the staking api wrapper altogether, so I'm going to leave as is for now

contracts/integrations/test/framework/assertions/finalizePool.ts
mzhu25: eh yeah looking at this again I think I'll just replace these with normal comments

contracts/integrations/test/framework/assertions/stake.ts
mzhu25: we aren't actually updating the global undelegated stake in `stake` and `unstake` right? I need to investigate why this check was passing before actually :p

contracts/integrations/test/framework/assertions/withdrawDelegatorRewards.ts
mzhu25: made a task 👍

contracts/integrations/test/framework/assertions/stake.ts
mzhu25: o wait I'm trippin
back they go

contracts/integrations/test/framework/assertions/finalizePool.ts
mzhu25: nvm @dorothy-zbornak figured it out, it was a bug in the `LibFractions.normalize` reference function

contracts/staking/src/types.ts
```diff
@@ -67,6 +67,62 @@ export class StoredBalance {
...
 }
 
+/**
+ * Simulates _loadCurrentBalance. `shouldMutate` flag specifies whether or not to update the given
+ * StoredBalance instance.
+ */
+export function loadCurrentBalance(
+    balance: StoredBalance,
+    epoch: BigNumber,
+    shouldMutate: boolean = false,
```
dorothy-zbornak: Bit of a nit. Does this really make things cleaner? I mean if we didn't have it we'd just do a `balance = loadCurrentBalance(balance, ...)` vs `loadCurrentBalance(balance, ..., true)`. But the first seems clearer.

contracts/staking/src/types.ts
```diff
@@ -67,6 +67,62 @@ export class StoredBalance {
...
+export function increaseCurrentAndNextBalance(balance: StoredBalance, amount: Numberish, epoch: BigNumber): void {
+    loadCurrentBalance(balance, epoch, true);
+    balance.currentEpochBalance = balance.currentEpochBalance.plus(amount);
+    balance.nextEpochBalance = balance.nextEpochBalance.plus(amount);
+}
+
+/**
+ * Simulates _decreaseCurrentAndNextBalance
+ */
+export function decreaseCurrentAndNextBalance(balance: StoredBalance, amount: Numberish, epoch: BigNumber): void {
```
dorothy-zbornak: Man, yeah, idk how I feel about these functions that mutate in place. I get that they kind of mimic the way the solidity works but it's kind of rough to follow in TS.

contracts/integrations/test/fuzz_tests/staking_rewards_test.ts
```diff
@@ -0,0 +1,125 @@
...
+    before(function(): void {
+        if (process.env.FUZZ_TEST !== 'staking_rewards') {
+            this.skip();
```
dorothy-zbornak: Wow, never knew about this. Stealing.

contracts/integrations/test/framework/assertions/finalizePool.ts
```diff
@@ -0,0 +1,217 @@
...
+} from '@0x/contracts-test-utils';
+import { BigNumber } from '@0x/utils';
+
+import { DeploymentManager } from '../deployment_manager';
+import { SimulationEnvironment } from '../simulation';
+
+import { FunctionAssertion, FunctionResult } from './function_assertion';
+
+const PRECISION = 15;
+const COBB_DOUGLAS_ALPHA = toDecimal(stakingConstants.DEFAULT_PARAMS.cobbDouglasAlphaNumerator).dividedBy(
```
dorothy-zbornak: Is it worth just pulling this in a `before()` block to reduce the chance of it getting out of sync?

contracts/integrations/test/framework/assertions/fillOrder.ts
```diff
@@ -30,68 +50,172 @@ function verifyFillEvents(
...
-                makerFeePaid: constants.ZERO_AMOUNT,
-                takerFeePaid: constants.ZERO_AMOUNT,
-                protocolFeePaid: DeploymentManager.protocolFee,
+                ...fillResults,
             },
         ],
         ExchangeEvents.Fill,
     );
 
+    const expectedTransferEvents = [
```
dorothy-zbornak: Won't some of these not happen if the amounts are zero?

contracts/integrations/test/framework/actors/taker.ts
```diff
@@ -61,32 +63,24 @@ export function TakerMixin<TBase extends Constructor>(Base: TBase): TBase & Cons
...
+        private async *_validFillOrder(): AsyncIterableIterator<AssertionResult | void> {
+            const { actors } = this.actor.simulationEnvironment!;
+            const assertion = validFillOrderAssertion(this.actor.deployment, this.actor.simulationEnvironment!);
             while (true) {
-                const maker = Pseudorandom.sample(marketMakers);
+                // Choose a maker to be the other side of the order
+                const maker = Pseudorandom.sample(filterActorsByRole(actors, Maker));
                 if (maker === undefined) {
-                    yield undefined;
+                    yield;
```
dorothy-zbornak: Q: Do you need an `else` if you are just `yield`ing empty?

contracts/integrations/test/framework/actors/taker.ts
```diff
@@ -61,32 +63,24 @@ export function TakerMixin<TBase extends Constructor>(Base: TBase): TBase & Cons
...
-                    yield assertion.executeAsync([order, order.takerAssetAmount, order.signature], {
+                    // Maker creates and signs a fillable order
+                    const order = await maker.createFillableOrderAsync(this.actor);
+                    // Taker fills the order by a random amount (up to the order's takerAssetAmount)
+                    const fillAmount = Pseudorandom.integer(order.takerAssetAmount);
+                    // Taker executes the fill with a random msg.value, so that sometimes the
+                    // protocol fee is paid in ETH and other times it's paid in WETH.
+                    yield assertion.executeAsync([order, fillAmount, order.signature], {
                         from: this.actor.address,
+                        value: Pseudorandom.integer(DeploymentManager.protocolFee.times(2)),
```
dorothy-zbornak: Why times 2?

contracts/integrations/test/framework/actors/taker.ts
mzhu25: no, but without the `else` block tsc complains that `maker` could be undefined 🙄

contracts/integrations/test/framework/actors/taker.ts
mzhu25: want to test for refunds when msg.value > protocol fee
2 is arbitrary

contracts/integrations/test/fuzz_tests/pool_management_test.ts
```diff
@@ -46,8 +43,13 @@ blockchainTests('Pool management fuzz test', env => {
         });
         const balanceStore = new BlockchainBalanceStore({}, {});
 
-        const simulationEnv = new SimulationEnvironment(deployment, balanceStore);
-        const simulation = new PoolManagementSimulation(simulationEnv);
+        const simulationEnvironment = new SimulationEnvironment(deployment, balanceStore);
+        // tslint:disable:no-unused-expression
+        new PoolOperator({ deployment, simulationEnvironment, name: 'Operator 1' });
```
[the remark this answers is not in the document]
mzhu25: yeah, I'll make this more explicit

contracts/integrations/test/framework/assertions/finalizePool.ts
mzhu25: I think I'll leave as is for now, and add the staking params to the `SimulationEnvironment` instance when I implement the `Governor` fuzz tests

PR #2389, opened 2019-12-05
Allow mainnet fork to be used for contract tests

abandeali1:
## Description

- Adds a `fork` option to `describe` and `blockchainTests`. This defaults to `skip` if the `FORK_RPC_URL` env variable isn't set and changes to `only` if it is set (which means tests must be run both with and without `FORK_RPC_URL` in order to run all).
- Adds tests for all mainnet contract configs, except for `ZeroExGovernor`.
 

## Checklist:


contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -116,6 +122,13 @@ export const blockchainTests: BlockchainContextDefinition = _.assign(
         optional(description: string, callback: BlockchainSuiteCallback): ISuite | void {
             return defineBlockchainSuite(description, callback, process.env.TEST_ALL ? describe : describe.skip);
         },
+        fork(description: string, callback: BlockchainSuiteCallback): ISuite | void {
+            return defineBlockchainSuite(
+                description,
+                callback,
+                process.env.FORK_RPC_URL ? describe.only : describe.skip,
```
dorothy-zbornak: Neat!

contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -97,6 +99,10 @@ export const describe = _.assign(mochaDescribe, {
         const describeCall = process.env.TEST_ALL ? mochaDescribe : mochaDescribe.skip;
         return describeCall(description, callback);
     },
+    fork(description: string, callback: SuiteCallback): ISuite | void {
```
dorothy-zbornak: I wonder if this should have a signature like `fork(description, opts, callback)` so we can pass unlocked accounts and fork block numbers in the tests. I feel like the tests should own these parameters. Thoughts?

abandeali1: I'd like something like that, but it's not actually very easy to instantiate a new provider for `blockchainTests` without a large-ish refactor. Let's backlog it?

contracts/integrations/test/mainnet_configs_test.ts
```diff
@@ -0,0 +1,200 @@
...
+        it('ERC20BridgeProxy should be registered', async () => {
+            const erc20BridgeProxy = await multiAssetProxy.getAssetProxy(AssetProxyId.ERC20Bridge).callAsync();
+            expect(erc20BridgeProxy).to.eq(contractAddresses.erc20BridgeProxy);
+        });
+        it('StaticCallProxy should be registered', async () => {
+            const staticCallProxy = await multiAssetProxy.getAssetProxy(AssetProxyId.StaticCall).callAsync();
+            expect(staticCallProxy).to.eq(contractAddresses.staticCallProxy);
+        });
+    });
+    describe('StakingProxy', () => {
```
hysz: Could be good to test the storage layout here as well.

abandeali1: 👍 Added a task to do this in a separate PR

Issue #2380, opened 2019-12-02
General: Handle a batch order of v2 and v3 orders in v3 Forwarder


satosheth:
Feature Request: Be able to handle the filling of a batch of v2 and v3 orders in one transaction within the forwarder.

Problem:
We would like to switch over to 0x v3 as soon as possible, but not all other relayers will be switching over right away. One of the key reasons we will be switching over is for the batch filling of ERC721 orders. However, if we cannot batch fill the orders we get from v2 relayers (and our current orders) with our new v3 orders, it will be impossible to provide this functionality for our users as it will be too confusing of an interface.

My specific use case is only for ERC721 tokens, I am not sure of the demand for this functionality for other tokens.

PR #2390, opened 2019-12-09
Make Forwarder backwards compatible with v2

abandeali1:
## Description

- Makes the `Forwarder` backwards compatible with v2 if the order's `makerFeeAssetData` field is set to `0x770501f8` (first 4 bytes of the v2 order schema hash). Note that no fees are supported for v2 orders in this implementation, even if the order involves ZRX on either side of the trade.
- Adds some integration tests for the `Forwarder` using Ganache's forking functionality. These are _extremely_ finicky and tests will sometimes fail without reason. There is also a set of tests that are skipped because protocol fees are not working properly on the fork.
- Changes the Solidity EVM target to Istanbul and redeploys bridge contracts

## Checklist:

#2380


contracts/integrations/test/forwarder/forwarder_mainnet_test.ts
```diff
@@ -0,0 +1,201 @@
...
+                .callAsync({
+                    from: takerAddress,
+                    value: ethSellAmount,
+                });
+            expect(wethSpentAmount).to.bignumber.eq(ethSellAmount);
+            expect(makerAssetAcquiredAmount).to.bignumber.eq(
+                BigNumber.sum(orders[0].makerAssetAmount, orders[1].makerAssetAmount.dividedToIntegerBy(2)),
+            );
+        });
+        it.skip('should fill multiple v2/v3 orders', async () => {
```
dorothy-zbornak: Tried to play with this myself and couldn't figure out why it was reverting. The behavior is so bizarre. Are we sure the contract at `0x9ea309d4f96d6433c75a6b9c5a596cf19a87edb8` is the same one in this PR?

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -401,4 +483,15 @@ contract MixinExchangeWrapper is
...
+
+    /// @dev Checks whether an order is a v2 order.
+    /// @param order Order struct containing order specifications.
+    /// @return True if the order's `makerFeeAssetData` is set to the v2 order id.
+    function _isV2Order(LibOrder.Order memory order)
+        internal
+        pure
+        returns (bool)
+    {
+        return order.makerFeeAssetData.readBytes4(0) == EXCHANGE_V2_ORDER_ID;
```
dorothy-zbornak: This will revert if the order has no maker fee asset.

contracts/integrations/test/forwarder/forwarder_mainnet_test.ts
abandeali1: I actually tried to get it to work by deploying a fresh Forwarder on the fork and running tests against that, same exact errors. I also selectively placed reverts to try to figure out what was going on and it appeared that`fillResults.protocolFeesPaid` was always 0 for v3 fills, despite getting past a `require(tx.gasprice != 0)` statement.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
abandeali1: Good catch!

packages/instant/.production.discharge.json
```diff
@@ -9,5 +9,6 @@
     "aws_profile": "0xproject",
     "aws_region": "us-east-1",
     "cdn": true,
-    "dns_configured": true
+    "dns_configured": true,
+    "should_keep_files_in_s3": true
```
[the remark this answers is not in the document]
abandeali1: Nope, must have accidentally committed this.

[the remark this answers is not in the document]
abandeali1: Ah I see, I guess I rebased from `development` at some point. This should go away once the base branch is up to date.

PR #2391, opened 2019-12-10
Generate (complete) solidity docs

dorothy-zbornak:
### [*See sample generated contract docs here!*](https://gist.github.com/dorothy-zbornak/ec4e35a459186f620d1464a8a1ad7f33)

Rewrote `sol-doc`completely to generate more complete solidity documentation.
Instead of relying on `solc`'s natspec output we now walk the AST and manually extract documentation. This allows us to gather documentation for things that are not usually covered by natspec, like structs, enums, events, etc. We also fallback to non-natspec comments when natspec is not available. This gives us much larger documentation coverage out of the box because we have a mixture of documentation styles throughout our solidity codebase.

`sol-doc` can output a JSON documentation artifact and/or markdown output. It will intelligently fold inherited contracts (so we can just display a single, composed contract instead of multiple mixins)  and will only output documentation for types that are exposed via some public function or event. It will also generate accessor functions for public state variables. 

There are two interesting package scripts in `sol-doc`:
- `generate-protocol-docs`: generates markdown docs for the `Exchange`, `Forwarder`, `Coordinator`, and `Staking`. The package `config` entries can be modified to add more targets.
- `s3:sync_md_docs`: uploads the generated docs to an s3 bucket at `s3://docs-markdown/@0x/contracts-docs/v3.0.0.0` (I think, lol). 

### Extended documentation styles
We now support an extended variety of docstrings on non-function types.

**Examples with `struct`s**
```solidity
/// @dev Docstring for `StructA`.
/// @param field Docstring for `StructA.field`.
struct StructA {
  bytes32 field;
}

// @dev Docstring for `StructC`.
struct StructC {
  // Docstring for `StructC.field`.
  bytes32 field;
}

// Docstring for `StructB`.
struct StructB {
   bytes32 field; // Docstring for `StructB.field`
}
```

**Examples with `enum`s**
```solidity
/// @dev Docstring for `EnumA`.
/// @param Value Docstring for `EnumA.Value`.
enum EnumA {
  Value
}

/// @dev Docstring for `EnumA`.
enum EnumA {
  // Docstring for `EnumA.Value`.
  Value
}

// Docstring for `EnumA`.
enum EnumA {
  Value // Docstring for `EnumA.Value`.
}

```
**Examples with state variables**
These will be converted to accessor functions..

```solidity
/// @dev Docstring for `stateA`.
bytes32 stateA;

// Docstring for `stateB`.
bytes32 stateB;

bytes32 stateC; // Docstring for `stateC`.

/// @dev Docstring for `arrayA`.
/// @param 0 Docstring for the first indexer.
/// @param 1 Docstring for the second indexer.
/// @return 0 Docstring for the final `bytes32` value.
bytes32[][] arrayA;

/// @dev Docstring for `mapA`.
/// @param 0 Docstring for the first `address` key.
/// @param 1 Docstring for the second `uint256` key.
/// @return 0 Docstring for the final `bytes32` value.
mapping (address => mapping (uint256 => bytes32)) mapA;
```

## Checklist:


contracts/staking/contracts/src/immutable/MixinStorage.sol
```diff
@@ -72,7 +76,9 @@ contract MixinStorage is
     // mapping from Pool Id to Epoch
     mapping (bytes32 => uint256) internal _cumulativeRewardsByPoolLastStored;
 
-    // registered 0x Exchange contracts
+    /// @dev Registered 0x Exchange contracts, capable of paying protocol fees.
+    /// @param 0 The address to check.
+    /// @return 0 Whether the address is a registered exchange.
```
hysz: > /// @return 0 Whether
Why `0` is this required for doc gen?

hysz: (If this corresponds to return parameter at index `0`, could be nice to just default to 0)

contracts/exchange-libs/contracts/src/LibOrder.sol
```diff
@@ -78,8 +79,9 @@ library LibOrder {
     }
     // solhint-enable max-line-length
 
+    /// @dev Order information returned by `getOrderInfo()`.
     struct OrderInfo {
-        uint8 orderStatus;                    // Status that describes order's validity and fillability.
+        OrderStatus orderStatus;                    // Status that describes order's validity and fillability.
```
hysz: 👍

contracts/exchange-libs/contracts/src/LibEIP712ExchangeDomain.sol
```diff
@@ -29,9 +29,11 @@ contract LibEIP712ExchangeDomain {
     // EIP712 Exchange Domain Version value
     string constant internal _EIP712_EXCHANGE_DOMAIN_VERSION = "3.0.0";
 
-    // Hash of the EIP712 Domain Separator data
-    // solhint-disable-next-line var-name-mixedcase
+    // solhint-disable var-name-mixedcase
+    /// @dev Hash of the EIP712 Domain Separator data
+    /// @return 0 Domain hash.
```
hysz: Ah, so this is pulled in for the public function for this variable?

contracts/staking/contracts/src/immutable/MixinStorage.sol
dorothy-zbornak: Hmm. The syntax becomes ambiguous that way, because we can't determine whether the first word of the comment is the parameter name or just part of the comment.

contracts/exchange-libs/contracts/src/LibEIP712ExchangeDomain.sol
dorothy-zbornak: Yup.

contracts/staking/contracts/src/immutable/MixinStorage.sol
hysz: Oh I see, I misinterpreted the `0` as the index but this the name we give in the docs. Maybe it would read better if we just gave them names?
```
/// @param exchangeAddress The address to check.
/// @return isRegistered Whether the address is a registered exchange.
```
Not sure if this would have any unintended consequences though. I'm sure either is fine.

dorothy-zbornak: Yeah it would be cool to give them names but they won't jive with the ABI or the wrappers where they're just indexed. I don't know if that matters. We also have to be mindful of parameter ordering with this approach.

PR #2398, opened 2019-12-17
`@0x/contracts-exchange-forwarder`: StaticCall support, MultiAsset buy support

mzhu25:
## Description
Adds for MultiAsset `makerAssetData` and StaticCall `makerAssetData`/`makerFeeAssetData`/`takerFeeAssetData` in the Forwarder.
MultiAsset buy support is straightforward enough, but StaticCall `takerFeeAssetData` introduces some corner cases wrt how the Forwarder computes how much WETH has been spent. 
We could simplify the code a bit by restricting the user from providing StaticCall `takerFeeAssetData`, but one could plausibly have some weird use case where you'd want that due to for settlement order reasons 🤷‍♂
Edit: No longer supports StaticCall `takerFeeAssetData` for cleanliness' sake

## Testing instructions
Unit tests: `PKG=@0x/contracts-exchange-forwarder yarn test`
Integration tests: `PKG=@0x/contracts-integrations yarn test`

## Checklist:

### DIFF
--- a/contracts/exchange-forwarder/contracts/src/MixinAssets.sol
+++ b/contracts/exchange-forwarder/contracts/src/MixinAssets.sol
@@ -146,6 +146,9 @@ contract MixinAssets is
         );
     }
 
+    /// @dev Decodes MultiAsset assetData and recursively transfers assets to sender.
+    /// @param assetData Byte array encoded for the respective asset proxy.
+    /// @param amount Amount of asset to transfer to sender.
     function _transferMultiAsset(
         bytes memory assetData,
         uint256 amount
--- a/contracts/exchange-forwarder/contracts/src/MixinAssets.sol
+++ b/contracts/exchange-forwarder/contracts/src/MixinAssets.sol
@@ -20,6 +20,7 @@ pragma solidity ^0.5.9;
 
 import "@0x/contracts-utils/contracts/src/LibBytes.sol";
 import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
+import "@0x/contracts-utils/contracts/src/LibSafeMath.sol";
 import "@0x/contracts-utils/contracts/src/Ownable.sol";
 import "@0x/contracts-erc20/contracts/src/LibERC20Token.sol";
 import "@0x/contracts-erc721/contracts/src/interfaces/IERC721Token.sol";
@@ -35,6 +36,7 @@ contract MixinAssets is
     IAssets
 {
     using LibBytes for bytes;
+    using LibSafeMath for uint256;
 
     /// @dev Withdraws assets from this contract. It may be used by the owner to withdraw assets
     ///      that were accidentally sent to this contract.
@@ -95,7 +97,9 @@ contract MixinAssets is
             _transferERC20Token(assetData, amount);
         } else if (proxyId == IAssetData(address(0)).ERC721Token.selector) {
             _transferERC721Token(assetData, amount);
-        } else {
+        } else if (proxyId == IAssetData(address(0)).MultiAsset.selector) {
+            _transferMultiAsset(assetData, amount);
+        } else if (proxyId != IAssetData(address(0)).StaticCall.selector) {
             LibRichErrors.rrevert(LibForwarderRichErrors.UnsupportedAssetProxyError(
                 proxyId
             ));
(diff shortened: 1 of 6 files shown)


contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -494,4 +496,27 @@ contract MixinExchangeWrapper is
...
         return order.makerFeeAssetData.length > 3 && order.makerFeeAssetData.readBytes4(0) == EXCHANGE_V2_ORDER_ID;
     }
+
+    /// @dev Checks whether one asset is effectively equal to another asset.
+    ///      This is the case if they have the same ERC20Proxy/ERC20BridgeProxy asset data, or if
+    ///      one is the ERC20Bridge equivalent of the other.
+    /// @param takerFee Byte array encoded for the takerFee asset proxy.
+    /// @param takerFeeAssetData Byte array encoded for the maker asset proxy.
+    /// @return Whether or not the underlying assets are equal.
+    function _noTakerFee(
```
abandeali1: What do we actually gain from special casing the `StaticCallProxy` here? I feel like it makes the logic inconsistent from how the `StaticCallProxy` works elsewhere. Also in the interest of keeping this codebase as lean as possible, I'd probably opt to remove all of these checks completely.

mzhu25: If the `takerFeeAssetData` is StaticCall and we _don't_ treat the `takerFeePaid` amount as zero, we'd be subtracting some arbitrary non-zero amount from either the `wethSpentAmount` or the `makerAssetAcquiredAmount`; either way the results would be incorrect/misleading. So the options are either:
- this (`_noTakerFee`)
- adding another `else if` block to `_marketSellSingleOrder` and `_marketBuySingleOrder`
- enforcing that the `takerFeeAssetData` is not StaticCall

I think the current option is a bit cleaner than the 2nd option, but I'd be down for the 3rd.

abandeali1: The different scenarios where this is currently used are:

- marketBuy where `order.takerFee == 0`. `takerFeePaid` will always be 0 here anyways.
- marketBuy where both `order.takerFeeAssetData` and `order.makerAssetData` are the same `StaticCall` data. The amount doesn't really matter in the case, so I don't see a huge need for extra logic.
- marketSell where `order.takerFee == 0`. `takerFeePaid` will always be 0 here anyways.
- marketSell where both `order.takerFeeAssetData` and `order.takerAssetData` are the same `StaticCall` data. Note this case will never be hit in the current contract because `takerAssetData` is always WETH. In future iterations where this is possible, the amount also doesn't really matter 🤷‍♂ .

I know it's kind of funky, but I think it's simpler to be as agnostic as possible when working with different `assetData` formats and only special case things when absolutely necessary. Otherwise the code becomes much less maintainable, so I'd still argue for removing this.

contracts/exchange-forwarder/contracts/src/MixinAssets.sol
```diff
@@ -95,7 +97,9 @@ contract MixinAssets is
             _transferERC20Token(assetData, amount);
         } else if (proxyId == IAssetData(address(0)).ERC721Token.selector) {
             _transferERC721Token(assetData, amount);
-        } else {
+        } else if (proxyId == IAssetData(address(0)).MultiAsset.selector) {
+            _transferMultiAsset(assetData, amount);
+        } else if (proxyId != IAssetData(address(0)).StaticCall.selector) {
```
hysz: Could probably just keep this as `else { revert }`.

contracts/exchange-forwarder/contracts/src/MixinAssets.sol
```diff
@@ -141,4 +145,23 @@ contract MixinAssets is
             tokenId
         );
     }
+
+    function _transferMultiAsset(
```
hysz: Nit - add function comment.

contracts/exchange-forwarder/test/asset_test.ts
```diff
@@ -115,8 +126,49 @@ blockchainTests('Supported asset type unit tests', env => {
...
+            const result = await forwarder.areUnderlyingAssetsEqual(multiAssetData, multiAssetData).callAsync();
+            expect(result).to.be.true();
+        });
+        it('returns false if assetData1 != assetData2 are MultiAsset', async () => {
+            const differentMultiAssetData = assetDataEncoder
+                .MultiAsset([getRandomInteger(0, constants.MAX_UINT256)], [erc721AssetData])
+                .getABIEncodedTransactionData();
+            const result = await forwarder
+                .areUnderlyingAssetsEqual(multiAssetData, differentMultiAssetData)
+                .callAsync();
```
hysz: Could be worthwhile to validating equality of `MultiAsset` with multiple nested assets.

PR #2399, opened 2019-12-17
0x-sra-client.py: Fix bug in config_order, and other small improvements

### LINKED ISSUE #2385: sra_client.py: broken order_config interface, and other minor issues
* The `get_order_config()` method is broken, in that it doesn't allow you to pass in an order to be configured.
* The library needs an easier entry point.  Currently you have to do:
    ```python
    from zero_ex.sra_client import ApiClient, Configuration, DefaultApi
    config = Configuration()
    config.host = "http://localhost:3000"
    relayer = DefaultApi(ApiClient(config))
    relayer.post_order(...)
    ```
    But there should be a simpler entry point, like just
    ```python
    from zero_ex.sra_client import SRAClient
    relayer = SRAClient("http://localhost:3000")
    relayer.post_order(...)
    ```
* Fix doc to include documentation for RelayerApiOrdersResponseSchema and for all those other Schema types.  They're referenced by name in the current doc, but there's currently no way to get any info on them without digging into the source code.

feuGeneA:
## Description

Fixes #2385 

See generated docs at https://104044-92181371-gh.circle-artifacts.com/0/root/repo/python-packages/sra_client/build/docs/html/index.html

## Type of changes:


python-packages/sra_client/src/zero_ex/sra_client/__init__.py
```diff
@@ -155,7 +182,7 @@
  'order': {'chainId': 1337,
            'exchangeAddress': '0x...',
            'expirationTimeSeconds': '...',
-           'feeRecipientAddress': '0x0000000000000000000000000000000000000000',
+           'feeRecipientAddress': '0x0000000000000000000000000000000000000001',
```
PirosB3: @feuGeneA is this a bug? were the 4 changes below (`0x0000000000000000000000000000000000000001`) changed by mistake?

.gitignore
```diff
@@ -160,33 +160,7 @@ contracts/exchange-forwarder/generated-wrappers/
...
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/i_wallet/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/multi_asset_proxy/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/order_validator/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/staking/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/staking_proxy/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/static_call_proxy/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/weth9/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/zrx_token/__init__.py
-python-packages/contract_wrappers/src/zero_ex/contract_wrappers/zrx_vault/__init__.py
+python-packages/contract_wrappers/src/zero_ex/contract_wrappers/*/__init__.py
```
PirosB3: 👍

python-packages/contract_wrappers/setup.py
```diff
@@ -128,11 +105,27 @@ def run(self):
...
+            )
+        ):
             try:
-                remove(f"src/zero_ex/contract_wrappers/{contract}/__init__.py")
+                remove(
+                    path.join(
+                        "src",
+                        "zero_ex",
+                        "contract_wrappers",
+                        f"{contract}",
```
PirosB3: why do you have this f-string? can't you simply use the variable `contract`?

python-packages/contract_wrappers/setup.py
```diff
@@ -128,11 +105,27 @@ def run(self):
...
+        rmtree(
+            path.join("src", "0x_contract_wrappers.egg-info"),
+            ignore_errors=True,
+        )
         # generated files:
-        for contract in CONTRACTS_TO_BE_WRAPPED:
+        print("Removing src/zero_ex/contract_wrappers/*/__init__.py...")
+        for contract in glob(
+            path.join(
+                ".", "src", "zero_ex", "contract_wrappers", "*", "__init__.py"
```
PirosB3: small nit here - this path is relative to the directory the current process is in, and not relative to the directory the current file is in. I think you want the latter?

**Example**

If I ran..
```
cd /tmp/
python path/to/monorepo/python-packages/contract_wrappers/setup.py
```

LOC 114 would result in `/tmp/src/zero_ex/contract_wrappers/*/__init__.py` while I think you wanted it to be `/tmp/path/to/monorepo/python-packages/contract_wrappers/src/zero_ex/contract_wrappers/*/__init__.py`

python-packages/sra_client/src/zero_ex/sra_client/__init__.py
feuGeneA: Not a bug, not a mistake.  We're now asking launch-kit-backend to configure the order for us, and it's responding with a 0x...1 fee address, so this is the new expectation.

python-packages/contract_wrappers/setup.py
feuGeneA: Good call.  This used to be a `/`-delimited path, and i converted it to `path.join()`.  The old string needed the `f`, but now I see what you mean.  Will update.

python-packages/contract_wrappers/setup.py
feuGeneA: You're absolutely right on this... but I don't think it's worth accommodating at this point.  The script already generally lacks support for this type of interaction.

PirosB3: yup yup I agree

PirosB3:
@feuGeneA the more I try out this code, the more I feel that `sra_client` library won't work anywhere. Given that we are deprecating LaunchKit Backend (which requires a `/v3/` prefix), and given that `api.0x.org` does not require a `/v3/` prefix in the path, `sra_client` won't work anywhere because it hard-codes a `/v3/` prefix in the path.

Is there any way that the current `sra_client` implementation works with 0x API?

PirosB3:
Aside from this, you code works great. I tested it using:

```python

>>> relayer = RelayerApi('https://v3.kovan.backend.sra.0x.org')
>>> order_config = relayer.get_order_config(relayer_api_order_config_payload_schema=my_order)
>>> print(order_config)
{'fee_recipient_address': '0x0000000000000000000000000000000000000000',
 'maker_fee': '0',
 'sender_address': '0x0000000000000000000000000000000000000000',
 'taker_fee': '0'}
```

PR #2403, opened 2019-12-19
Fix Circle builds

hysz:
## Description

Circle builds are failing when installing yarn:
```
#!/bin/bash -eo pipefail
npm install --global yarn@1.17.0
npm ERR! code EEXIST
npm ERR! syscall symlink
npm ERR! path ../lib/node_modules/yarn/bin/yarn.js
npm ERR! dest /usr/bin/yarn
npm ERR! errno -17
npm ERR! EEXIST: file already exists, symlink '../lib/node_modules/yarn/bin/yarn.js' -> '/usr/bin/yarn'
npm ERR! File exists: /usr/bin/yarn
npm ERR! Remove the existing file and try again, or run npm
npm ERR! with --force to overwrite files recklessly.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2019-12-19T08_40_19_522Z-debug.log
```

See relevant issue [here](https://github.com/yarnpkg/yarn/issues/7773). Rather than waiting for a fix we can just force install.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,7 @@ jobs:
             #       command: npm set prefix=/home/circleci/npm && echo 'export PATH=$HOME/circleci/npm/bin:$PATH' >> /home/circleci/.bashrc
             - run:
                   name: install-yarn
-                  command: npm install --global yarn@1.17.0
+                  command: npm install --force --global yarn@1.17.0
             - run:
                   name: yarn
                   command: yarn --frozen-lockfile --ignore-engines install || yarn --frozen-lockfile --ignore-engines install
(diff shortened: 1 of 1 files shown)


feuGeneA:
> LGTM, the `nikolaik/python-nodejs` image preinstalls yarn using apt: https://hub.docker.com/r/nikolaik/python-nodejs/dockerfile
> 
> Would it make sense for us to also prepare our own image without:
> ` apt-get install -yqq yarn && \` ?

@opaolini doesn't that mean that we would have to maintain that image going forward?  If so, I'd prefer not to do that.  The less maintenance overhead the better :)

@hysz why not just remove the `install-yarn` step?  The only reason I can think of to keep it there is to "document" our dependency on `yarn`, but we have `package.json` for that.

PR #2406, opened 2019-12-19
Fix asset-swapper bugs and misc improvements.

dorothy-zbornak:
### Ignore zero sample results from the sampler contract.
The sampler contract now returns zero when pulling a quote fails, such as when the market does not exist. We weren't ignoring these samples so buy quotes would see these as the best deals (something for nothing).

### ~Allow skipping Uniswap when dealing with low precision amounts with `minUniswapDecimals` option.~
~Uniswap is unstable at low precision amounts. We now skip over any uniswap quotes with low-precision inputs our outputs. An open question is whether to apply this special casing to Kyber as well if they pull from Uniswap...~

### Increase default `runLimit` from `1024` to `4096`.
Tests on my machine show little runtime penalty with this higher setting. Could be different on the server though, but let's see how it deals with it.

Higher `runLimit` means more thorough path-finding, which we definitely need with a higher sampling rate.

### Increase default `numSamples` from `8` to `10`.
We can get tighter pricing with more samples. I think this is still reasonable.

### Fix ordering of optimized orders.
Orders returned by the fills optimizer were out of order. This meant that best/worst quotes were sometimes flipped and, more importantly, market fills would potentially have a worse final conversion because more expensive orders were being filled first.

### Fix rounding of quoted asset amounts.
The quoted best and worst asset amounts were often off by `1` wei, due to rounding. This changes the rounding strategy to be more exact.

### Undo `bridgeSlippage` in best case quote.
Technically the most optimistic price for bridge orders is at `price / (1-bridgeSlippage)`, since we generate orders at `price * (1-bridgeSlippage)`. Previously we were just taking the price of the generated orders. 

## Checklist:


packages/asset-swapper/src/constants.ts
```diff
@@ -83,4 +83,5 @@ export const constants = {
     NULL_ERC20_ASSET_DATA,
     PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS,
     MARKET_UTILS_AMOUNT_BUFFER_PERCENTAGE,
+    SAMPLER_CONTRACT_GAS_LIMIT: 10e6,
```
dave4506: I'm fine with having this constant in this `constant.ts` file, but I feel like the `GAS_LIMIT` constant makes sense more to be in the `utils/market_operations_utils/constants.ts` file

dave4506:
@dorothy-zbornak Added a minor bug fix with a `quote` metadata, that surfaced when I was playing around with instant. Fixed case where first order in `resultOrders` is a bridge order and the quote metadata uses the bridge assetData instead of makerAssetData.

PR #2407, opened 2019-12-19
Add aggregator mainnet tests

dorothy-zbornak:
## Description

This adds mainnet aggregator tests (via asset-swapper) to the `contracts/integrations` package.

The purpose of these tests is to test fill-ability of the generated orders. We accomplish this by using a [fat wallet](https://etherscan.io/address/0xAb5801a7D398351b8bE11C439e05C5B3259aeC9B) (Vitalik's lol), `eth_call`, and a custom forwarder contract, deployed to mainnet. Basically we do the following for each market fill:

1. Pull ETH -> takerAsset orders (`takerOrders`) from SRA(s).
2. Pull takerAsset -> makerAsset orders (`makerOrders`) from `asset-swapper`.
3. `eth_call` the forwarder from the context of the fat wallet, attaching lots of ether.
  a. Wrap the passed in ether into WETH.
  b. Approve the ERC20 asset proxy for all relevant tokens.
  c. Buy up as much of takerAsset from `takerOrders` using the passed ether.
  d. Perform a market fill on `makerOrders` using the bought takerAssets.
  e. Return before and after balances, as well as fill results.
4. Perform assertions on the results. 

These tests are wholly non-deterministic and will probably fail from time to time as liquidity fluctuates, but they've already revealed some insidious bugs in the bridges and asset-swapper so obv they're useful.

These tests perform swaps on all combinations of (W)ETH, DAI, USDC, and FOAM at 1, 10, 100, 1000, 10000, 25000, and 50000 quantities.

### `blockchainTests` extensions
You can now do a `blockchainTests.live(...)` in tests to get a provider that points to the mainnet backend defined by `LIVE_RPC_URL` (contrast this with `FORK_RPC_URL` used by `fork()`).

Additionally, I went ahead and added modifiers to the existing `blockchainTests.fork()`. So now you can do `blockchainTests.fork.only(...)`, etc.

BONUS: You can now run `fork()` and `live()` tests alongside regular ganache tests because they all create separate providers!

## Checklist:


contracts/integrations/test/aggregation/tokens.ts
```diff
@@ -0,0 +1,77 @@
+export const tokens: { [symbol: string]: { address: string; decimals: number; price: number } } = {
```
hysz: Sweet!

contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -122,13 +193,69 @@ export const blockchainTests: BlockchainContextDefinition = _.assign(
...
             return defineBlockchainSuite(description, callback, process.env.TEST_ALL ? describe : describe.skip);
         },
-        fork(description: string, callback: BlockchainSuiteCallback): ISuite | void {
-            return defineBlockchainSuite(
-                description,
-                callback,
-                process.env.FORK_RPC_URL ? describe.only : describe.skip,
-            );
-        },
+        fork: _.assign(
```
hysz: goodguy.gif

contracts/integrations/test/aggregation/fill_test.ts
```diff
@@ -0,0 +1,218 @@
...
+            ...env.txDefaults,
+            gasPrice: GAS_PRICE,
+            gas: 10e6,
+        });
+        swapQuoter = SwapQuoter.getSwapQuoterForStandardRelayerAPIUrl(env.provider, 'https://api.0x.org/sra');
+        // Pool orderbooks because we're desperate for liquidity.
+        orderbooks.push(
+            swapQuoter.orderbook,
+            Orderbook.getOrderbookForPollingProvider({
+                httpEndpoint: 'https://sra.bamboorelay.com/0x/v3',
```
hysz: Do you know what happens if this endpoint is non-responsive?

dorothy-zbornak: The tests throw up. This is admittedly a little fragile while we wait for liquidity to come to the party. I'll add some try/catch clauses to handle these cases more gracefully.

contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -88,6 +98,71 @@ export class BlockchainTestsEnvironmentSingleton {
...
+    protected constructor() {
+        const snapshotHandlerAsync = async (): Promise<void> => {
+            throw new Error('Snapshots are not supported with a live provider.');
+        };
+        this.blockchainLifecycle = {
+            startAsync: snapshotHandlerAsync,
+            revertAsync: snapshotHandlerAsync,
+        } as any;
+        this.txDefaults = txDefaults;
+        if (process.env.LIVE_RPC_URL) {
```
abandeali1: This provider is instantiated differently than the `FORK_RPC_URL` provider. Maybe we should just consolidate all of that here?

contracts/test-utils/src/mocha_blockchain.ts
```diff
@@ -58,56 +58,98 @@ export interface BlockchainTestsEnvironment {
...
+        }
+        return ForkedBlockchainTestsEnvironmentSingleton._instance;
+    }
+
+    protected static _createWeb3Provider(forkHost: string): Web3ProviderEngine {
+        return web3Factory.getRpcProvider({
+            ...providerConfigs,
+            fork: forkHost,
+            blockTime: 0,
+            locked: false,
```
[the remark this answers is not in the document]
dorothy-zbornak: This replaced the `unlocked_accounts` option that was previously set for the governor, etc. I tried it without, though, and the tests ran fine so it seems `false` is actually the default, so we can drop it from here.

PR #2421, opened 2020-01-03
asset-swapper: Compare equivalent asset data

dekz:
## Description

Compares asset data equivalent rather than the exact string. Allowing for Bridge Orders to be equivalent to erc20 asset data when compared. 

We often validate quotes and orders from the user specified asset data. Previously a [quote could be invalid](https://github.com/0xProject/0x-monorepo/blob/9d5724e1a0f2ffcd33622b285fdef477ec36ecc6/packages/asset-swapper/src/utils/assert.ts#L39) for DAI if the only order returned was a bridge order. And [paying taker fees](https://github.com/0xProject/0x-monorepo/blob/9d5724e1a0f2ffcd33622b285fdef477ec36ecc6/packages/asset-swapper/src/utils/utils.ts#L30) on an order with a bridge could result in validation failure.

There are now two functions available to use:
```typescript
// Rough equivalency, i.e makerAssetData MUST be ERC20(WETH) OR ERC20BridgeOrder(WETH)
utils.isAssetDataEquivalent 
// Exact match, i.e takerAssetData MUST be ERC20(WETH)
utils.isExactAssetData
```

This does not currently handle the case where `ERC20(token) === MAP(ERC20(token), StaticCall())` etc which we will add in the future.

Prior to https://github.com/0xProject/0x-monorepo/pull/2406 it could return ERC20BridgeOrders as the `makerAssetData`.  That PR returns the input order (defaults to Native or DummyBridge), though this still fails validation.


packages/asset-swapper/src/utils/protocol_fee_utils.ts
```diff
@@ -21,8 +21,10 @@ export class ProtocolFeeUtils {
         return constants.PROTOCOL_FEE_MULTIPLIER;
     }
 
-    // tslint:disable-next-line: prefer-function-over-method
     public async getGasPriceEstimationOrThrowAsync(shouldHardRefresh?: boolean): Promise<BigNumber> {
+        if (this.gasPriceEstimation.eq(constants.ZERO_AMOUNT)) {
```
dorothy-zbornak: :sunglasses:

PR #2425, opened 2020-01-03
Exchange signature validation fuzz tests

dorothy-zbornak:
## Description

This adds snake oil I MEAN FUZZ TESTS for the signature validation functions on the Exchange contract:
- `isValidHashSignature()`
- `isValidOrderSignature()`
- `isValidTransactionSignature()`

Most of these tests focus on negative testing (bad signatures) because, you know. 

Some small modifications were made to the fuzzing framework to accommodate:
- Add `Pseudorandom.hex()` for generating pseudo-random hex bytes.
- Add `Simulation.resets` property that automatically resets the blockchain state after every step.
- Add `DeplomentManager.web3Wrapper` to facilitate ^.

Otherwise everything (actors, etc) is pretty much contained within the [exchange_signature_validation_test.ts](../blob/fuzz/exchange/signature-validation/contracts/integrations/test/fuzz_tests/exchange_signature_validation_test.ts) file because I doubt we'd ever reuse any of these actors.

## Checklist:


contracts/integrations/test/fuzz_tests/exchange_signature_validation_test.ts
```diff
@@ -0,0 +1,832 @@
...
+
+    protected async _presignHashAsync(signer: string, hash: string): Promise<void> {
+        await this.deployment.exchange.preSign(hash).awaitTransactionSuccessAsync({ from: signer });
+    }
+
+    protected async _approveValidatorAsync(signer: string, validator: string, approved: boolean = true): Promise<void> {
+        await this.deployment.exchange
+            .setSignatureValidatorApproval(validator, approved)
+            .awaitTransactionSuccessAsync({ from: signer });
+    }
```
mzhu25: do these need to be wrapped in an actor at all? I think we can just have these as standalone generators

contracts/integrations/test/fuzz_tests/exchange_signature_validation_test.ts
```diff
@@ -0,0 +1,832 @@
...
+            });
+            mangled.hash = await transactionHashUtils.getTransactionHashHex(mangled.transaction);
+            break;
+        default:
+            throw new Error(`Unhandled mangle mode: ${mode}`);
+    }
+    return mangled;
+}
+
+class HashSignatureValidatorActor extends SignatureValidatorActor {
```
mzhu25: ditto

dorothy-zbornak:
> Looks p good to me, but can we get rid of the Actors? Like you said, we're unlikely to reuse these actors anywhere else, and the simulation doesn't really make use of any actor functionality afaict. Might as well make them standalone functions, or instance methods of `SignatureValidationSimulation`

Oh absolutely. I was just trying to be idiomatic since I was new to these fuzz tests and had no clue what I was doing. :woman_shrugging:

contracts/integrations/test/fuzz_tests/exchange_signature_validation_test.ts
```diff
@@ -0,0 +1,725 @@
...
+            let result;
+            try {
+                result = await exchange.isValidOrderSignature(params.order, params.signature).callAsync();
+            } catch (err) {
+                if (params.isValid) {
+                    throw err;
+                }
+                return;
+            }
+            expect(result).to.eq(!!params.isValid);
```
hysz: Whoa what does the `!!` mean?

dorothy-zbornak: Just coerces it into a boolean.

PR #2427, opened 2020-01-06
ERC20BridgeSampler: Allow for batching multiple buy/sell samples

dekz:
## Description

To discover the price of certain pairs we need an efficient way to perform multiple quote samples at once.

Initial investigation:

```typescript
await sampler
    .queryBatchOrdersAndSampleBuys(
        _.times(times, i => orders),
        _.times(times, i => orders.map(o => o.signature)),
        [
            SOURCE_TO_ADDRESS[ERC20BridgeSource.Eth2Dai],
            SOURCE_TO_ADDRESS[ERC20BridgeSource.Uniswap],
        ],
        [new BigNumber('1000000000000000000')],
    )
    .callAsync();
```

We query with 16 Native orders for each asset pair. 

**Note**:

* it's possible we will have more and less orders in the real world and will need to batch based on the Native orders.
* we're only querying for 1 asset amount
* this queries the same orders multiple times, our ABI encoder may optimize this and the results will not be as accurate as real world testing. This also excludes Kyber as it is buying.

```
16 'orders'
query 1: 1643.028ms
query 1: 1617.886ms
query 1: 1630.122ms
query 1: 1924.910ms
query 1: 1665.817ms
query 1: 1523.556ms
done 1

query 20: 1936.026ms
query 20: 1873.733ms
query 20: 1631.222ms
query 20: 1681.635ms
query 20: 1694.278ms
query 20: 1848.476ms
done 20

query 40: 1908.213ms
query 40: 2097.256ms
query 40: 1911.376ms
query 40: 1886.049ms
query 40: 2103.286ms
query 40: 1846.552ms
done 40

query 80: 2300.164ms
query 80: 2530.159ms
query 80: 2365.349ms
query 80: 3428.551ms
query 80: 2612.497ms
query 80: 3319.264ms
done 80

query 160: 3940.097ms
query 160: 3531.208ms
query 160: 3800.738ms
query 160: 3716.978ms
query 160: 4308.913ms
query 160: 4676.135ms
done 160

query 320: 8395.693ms
query 320: 8389.707ms
query 320: 6770.865ms
query 320: 6132.578ms
query 320: 8007.835ms
query 320: 8511.705ms
done 320
```

So in the initial investigation we can query 320 distinct asset pairs in 6-8 seconds. We currently have ~62 tokens in 0xApi metadata and would likely only consider a few base pairs (WETH, DAI). Currently we'd be looking at ~120 asset pairs which can be served in under 5 seconds.


contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -41,6 +41,61 @@ contract ERC20BridgeSampler is
     uint256 constant internal UNISWAP_SAMPLE_CALL_GAS = 150e3;
     uint256 constant internal ETH2DAI_SAMPLE_CALL_GAS = 250e3;
 
+     struct OrdersAndSample {
+        uint256[] orderFillableTakerAssetAmounts;
+        uint256[][] makerTokenAmountsBySource;
+    }
+
+    function queryMultipleOrdersAndSampleBuys(
```
dorothy-zbornak: I think it's probably clearer to rename these functions to something just like `getTokenBuy/SellPrices()`, and return prices instead of fill amounts. For every pair, we could just return the price of the first fillable native order + the price from each DEX. Price could be expressed either as an 18 decimal fraction or a maker/taker amount pair.

As is, it's kind of hard to understand what the intent of these functions are.

packages/orderbook/src/orderbook.ts
```diff
@@ -84,6 +84,26 @@ export class Orderbook {
             o => o.order.makerAssetData === makerAssetData && o.order.takerAssetData === takerAssetData,
         );
     }
+    public async getBatchOrdersAsync(makerAssetDatas: string[], takerAssetDatas: string[]): Promise<APIOrder[][]> {
+        for (const [mi, makerAssetData] of makerAssetDatas.entries()) {
```
dorothy-zbornak: Would it be better to explicitly pass in pairs instead of generating combinations to avoid doing more work than necessary?

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -41,6 +41,71 @@ contract ERC20BridgeSampler is
...
+    /// @dev Query batches of native orders and sample sell quotes on multiple DEXes at once.
+    /// @param orders Batches of Native orders to query.
+    /// @param orderSignatures Batches of Signatures for each respective order in `orders`.
+    /// @param sources Address of each DEX. Passing in an unsupported DEX will throw.
+    /// @param takerTokenAmounts Batches of Taker token sell amount for each sample.
+    /// @return ordersAndSamples How much taker asset can be filled
+    ///         by each order in `orders`. Maker amounts bought for each source at
+    ///         each taker token amount. First indexed by source index, then sample
+    ///         index.
+    function queryBatchOrdersAndSampleSells(
```
dorothy-zbornak: Do we actually need this?

packages/orderbook/src/orderbook.ts
dekz: mm yeah could be a good idea. Let me think on this

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
dekz: If we want to change it up to ensure we hit Kyber then we'll need this. Added for consistency at least in the contract.

dorothy-zbornak: We should write unit tests for these then. I can assign myself the task after the merge.

PR #2432, opened 2020-01-09
Update contract wrappers (esp Forwarder + asset-swapper)

dorothy-zbornak:
## Description
(⊃｡•́‿•̀｡)⊃ *Chillax, most of the files changed are just generated artifacts.*

This PR regenerates `contract-artifacts` and `contract-wrappers` to use the latest contracts, most notably the Forwarder, which is now backwards compatible with V2 and supports flat/multiple affiliate fees. This has a couple implications:

- Altered deployment logic in `@0x/migrations`.
- All contract artifacts are compiled for `istanbul` now so the ganache provider created by `@0x/dev-utils` defaults to the `istanbul` evm. You'll get mystery reverts that will rob you of hours of pointless debugging otherwise (I'm not bitter).
- `@0x/asset-swapper` calculates affiliate fees slightly differently now (see below).

### `asset-swapper` Changes
The Forwarder no longer pays affiliate fees as a percentage of eth spent, in favor of a flat fee. We've elected to emulate the previous percentage-based behavior in this release, so `asset-swapper` naively computes a percentage -> flat fee based on the taker asset amount of the worst quote. So we do potentially overestimate the fee right now.

I also did some editing in this PR by removing `getSmartContractParamsOrThrow()` methods on `SwapQuoteConsumer`s. They were a nice code RIP, don't seem to be used by us, and I can't really imagine why someone would want this function in the first place.

## Checklist:


packages/asset-swapper/CHANGELOG.json
```diff
@@ -1,4 +1,17 @@
 [
+    {
+        "version": "3.1.0",
```
dekz:
```suggestion
        "version": "4.0.0",
```

due to removal of public function `getSmartContractParamsOrThrow`.

packages/migrations/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "5.2.0",
```
dekz:
```suggestion
        "version": "6.0.0",
```

I'd advocate for a major bump here as interacting with this `Forwarder` is a breaking change.

dorothy-zbornak:
> Do the existing addresses in `@0x/contract-addresses` correspond to deployments of the updated artifacts? If so, LGTM!

yup yup

PR #2433, opened 2020-01-09
Contract wrappers: Catch empty reverts on live networks

dorothy-zbornak:
## Description

Previously, performing a `callAsync()` on a live network/provider that reverts with no data (invalid opcode or empty revert) would appear to succeed, returning default values. This PR tries to fix that.

When an `eth_call` reverts with no data, we get a `0x` result (and that's it). The changes to `base-contract` will look for this result and, if we are expected valid return data`, will throw an error.

One thing to note is we can't detect reverts on functions that actually have no return values via `callAsync()`, because they should return `0x` anyway. However, we would never actually `eth_call` a function that returned no data except to see if it succeeds. In these cases, `estimateGas()` should be the preferred method since it already throws on all reverts (`sendTransactionAsync()` with no gas option already does this for you). 

## Checklist:


packages/utils/test/abi_encoder/return_values_test.ts
```diff
@@ -173,32 +174,29 @@ describe('ABI Encoder: Return Value Encoding/Decoding', () => {
...
-        it('Should decode NULL as default value (single; static)', async () => {
+        it('Should fail to decode NULL (single; static)', async () => {
             // Generate Return Value
             const method = new AbiEncoder.Method(ReturnValueAbis.singleStaticReturnValue);
-            const returnValue = '0x00000000';
             const encodedReturnValue = '0x';
-            const decodedReturnValue = method.strictDecodeReturnValue<string>(encodedReturnValue);
+            const decodeReturnValue = () => method.strictDecodeReturnValue<string>(encodedReturnValue);
             // Validate decoded return value
-            expect(decodedReturnValue).to.be.deep.equal(returnValue);
```
feuGeneA: For my own edification if nothing else, there's no reason to use `deep.equal` here, since `returnValue` is a string...right?

(Same comment applies to line 201.)

packages/base-contract/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "6.1.0",
+        "changes": [
+            {
+                "note": "Catch empty reverts on live networks.",
```
feuGeneA: Nit: shouldn't this be just "_Support_ catching..." ?  Since you just added a function but never call it, within `base-contract` itself.

dorothy-zbornak: tru dat

packages/utils/test/abi_encoder/return_values_test.ts
dorothy-zbornak: Yurrp. I think it was just done for uniformity.

hysz: Yeah, there was a lot of copy pasta in these tests.

PR #2440, opened 2020-01-12
Add ChaiBridge order balance/allowance checks to DevUtils

abandeali1:
## Description

This PR adds balance/allowance checks for `ChaiBridge` orders in `LibAssetData`.

## Checklist:


contracts/test-utils/src/web3_wrapper.ts
```diff
@@ -20,6 +20,7 @@ export let providerConfigs: Web3Config = {
     shouldUseInProcessGanache: true,
     shouldAllowUnlimitedContractSize: true,
     hardfork: 'istanbul',
+    unlocked_accounts: ['0x6cc5f688a315f3dc28a7781717a9a798a59fda7b', '0x55dc8f21d20d4c6ed3c82916a438a413ca68e335'],
```
dorothy-zbornak: Oh, I thought all addresses were unlocked on ganache by default?

abandeali1: You don't need to unlock the first 10 addresses associated with Ganache's private key or any addresses that you are only using with `eth_call`. I believe you still need to unlock the rest.

PR #2442, opened 2020-01-14
Add 1155 support to Forwarder

abandeali1:
## Description

- Adds 1155 support to the `Forwarder` contract.
- Refactors all transfer logic into `LibAssetDataTransfer`. This also adds `transferIn` and `transferOut` functions, which set the stage for adding arbitrary token sell support (using `transferIn` + marketBuy/marketSell)
- Note: tests do not use `verifyEventsFromLogs` because our ABI decoder currently cannot decode arrays in logs.

## Checklist:


contracts/exchange-forwarder/contracts/src/libs/LibAssetDataTransfer.sol
```diff
@@ -0,0 +1,258 @@
...
+import "@0x/contracts-utils/contracts/src/LibRichErrors.sol";
+import "@0x/contracts-utils/contracts/src/LibSafeMath.sol";
+import "@0x/contracts-erc20/contracts/src/LibERC20Token.sol";
+import "@0x/contracts-erc721/contracts/src/interfaces/IERC721Token.sol";
+import "@0x/contracts-erc1155/contracts/src/interfaces/IERC1155.sol";
+import "@0x/contracts-asset-proxy/contracts/src/interfaces/IAssetData.sol";
+import "./LibForwarderRichErrors.sol";
+
+
+library LibAssetDataTransfer {
```
dorothy-zbornak: :raised_hands:

contracts/exchange-forwarder/contracts/src/libs/LibAssetDataTransfer.sol
```diff
@@ -0,0 +1,258 @@
...
+    /// @param from Address to transfer asset from.
+    /// @param to Address to transfer asset to.
+    /// @param amount Amount of asset to transfer to sender.
+    function _transferERC20Token(
+        bytes memory assetData,
+        address from,
+        address to,
+        uint256 amount
+    )
+        internal
```
dorothy-zbornak: Could these just as well be `private`?

contracts/exchange-forwarder/contracts/src/libs/LibAssetDataTransfer.sol
```diff
@@ -0,0 +1,258 @@
...
+            LibRichErrors.rrevert(LibForwarderRichErrors.UnsupportedAssetProxyError(
+                proxyId
+            ));
+        }
+    }
+
+    ///@dev Transfer asset from sender to this contract.
+    /// @param assetData Byte array encoded for the respective asset proxy.
+    /// @param amount Amount of asset to transfer to sender.
+    function transferIn(
```
dorothy-zbornak: Do we use this? Is this in anticipation of property-based orders?

dorothy-zbornak:
> Do we want to use the asset proxies for transferring assets in so users don't need to set additional approvals?

Doesn't that require adding the forwarder as an authority, time-locks and all?

mzhu25:
> > Do we want to use the asset proxies for transferring assets in so users don't need to set additional approvals?
> 
> Doesn't that require adding the forwarder as an authority, time-locks and all?

yeah –– I guess it wouldn't be too annoying to set a single approval assuming we're only going to be using the transfer in functionality for the GU 1155 wrapper, but might be nice if we decide to use it for other assets too

contracts/exchange-forwarder/contracts/src/libs/LibAssetDataTransfer.sol
abandeali1: Yep, this is in anticipation of property based orders and arbitrary token sell support. We do not use this yet.

abandeali1:
I don't think a single allowance is a huge deal for now. If this turns out to be a popular use case we can consider adding in an upgradability path later on.

contracts/exchange-forwarder/test/asset_test.ts
```diff
@@ -215,19 +232,75 @@ blockchainTests.resets('Supported asset type unit tests', env => {
...
+            const amount = new BigNumber(1);
+            const ids = [await erc1155Wrapper.mintFungibleTokensAsync([forwarder.address], values)];
+            const assetData = assetDataEncoder
+                .ERC1155Assets(erc1155Token.address, ids, values, constants.NULL_BYTES)
+                .getABIEncodedTransactionData();
+            const txReceipt = await forwarder
+                .transferOut(assetData, amount)
+                .awaitTransactionSuccessAsync({ from: receiver });
+            const logArgs = (txReceipt.logs[0] as LogWithDecodedArgs<ERC1155TransferBatchEventArgs>).args;
+            expect(logArgs.operator).to.eq(forwarder.address);
```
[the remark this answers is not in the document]
abandeali1: Our log decoder is currently broken for events that contain arrays, so I went with this for now.

contracts/exchange-forwarder/contracts/src/libs/LibAssetDataTransfer.sol
abandeali1: I'm fine with it either way 🤷‍♂ It's possible we want to directly access these if we're limiting support to certain asset types.

dorothy-zbornak: Then should we drop the underscore?

PR #2445, opened 2020-01-17
asset-swapper: provide more accurate best quote price

dorothy-zbornak:
## Description

In certain scenarios, asset-swapper's `bestQuoteInfo` price could actually be way off. This is most noticeable with large fills, when there's a sizable bridge order at the end of the quote, and a nonzero slippage buffer. 

The reason this happens is because, during optimization, we collapse the DEX samples into a single bridge order, flattening the prices of the samples into a single constant price, which is the average price of all the samples. When we have a slippage buffer, this collapsed bridge order is at the average price of the worst case fill, because the bridge orders are generated to accommodate it.

To get the real best-case quote, we need to compute the integral over the price curve. We can approximate this by just computing the fill over the nested samples, effectively treating the bridge order as a collection of sub-orders at different prices.

*Note:* This only affected the quote price. The order would actually fill at the correct rate.

### Results
One particularly vulnerable pair is ETH/SNX, because the quote is almost always either *just* kyber or uniswap. After this change, we see a **+2%** improvement on this quote (at $100k).

I also simulated mainnet fills on $10-$50,000 buys and sells on combinations of ETH, USDC, DAI, and FOAM (ikr).  Out of 84 fills, the median *realized* fill price was 0.037% less than the best quote price (the mean was 0.034%). So, this improved calculation seems to be fairly close to reality.

## Checklist:


packages/asset-swapper/test/swap_quote_calculator_test.ts
```diff
@@ -65,6 +65,9 @@ const createSamplerFromSignedOrdersWithFillableAmounts = (
     return sampler;
 };
 
+// TODO(dorothy-zbornak): Replace these tests entirely with unit tests because
+// omg they're a nightmare to maintain.
+
```
dorothy-zbornak: Nearly ragequit life updating these tests. :tired_face:

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -79,6 +79,48 @@ export interface Fill {
...
+
+/**
+ * Optimized orders to fill.
+ */
+export interface OptimizedMarketOrder extends SignedOrderWithFillableAmounts {
+    /**
+     * The optimized fills that generated this order.
+     */
+    fill: CollapsedFill;
+}
```
dorothy-zbornak: So now `getMarketSellOrdersAsync()` & fam return an array of these, so the quote calculator will have access to the individual fill/samples metadata.

packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -239,99 +246,155 @@ export class SwapQuoteCalculator {
...
+                        .div(partialTakerAssetAmount)
+                        .times(partialMakerAssetAmount)
+                        .integerValue(BigNumber.ROUND_DOWN);
+                    takerAssetAmount = takerAssetAmount.plus(partialTakerAssetFillAmount);
+                    makerAssetAmount = makerAssetAmount.plus(partialMakerAssetFillAmount);
+                }
+            }
+            totalMakerAssetAmount = totalMakerAssetAmount.plus(makerAssetAmount);
+            totalTakerAssetAmount = totalTakerAssetAmount.plus(takerAssetAmount);
+            totalFeeTakerAssetAmount = totalFeeTakerAssetAmount.plus(feeTakerAssetAmount);
```
dekz: TODO: start to think about when fees are in different assets rather than a constant assumption.

If we cross pairs for a better price fees might be whack in multiple tokens. If orders come from different Relayers they could use a different Fee structure (mF = mA or mF = tA).

packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -239,99 +246,155 @@ export class SwapQuoteCalculator {
...
-            },
-            {
-                totalMakerAssetAmount: constants.ZERO_AMOUNT,
-                totalTakerAssetAmount: constants.ZERO_AMOUNT,
-                totalFeeTakerAssetAmount: constants.ZERO_AMOUNT,
-                remainingMakerAssetFillAmount: makerAssetBuyAmount,
-            },
-        );
+                // Infer the bridge slippage from the difference between the fill
+                // size and the atual order asset amounts.
```
dekz:
```suggestion
                // size and the actual order asset amounts.
```

packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -239,99 +246,155 @@ export class SwapQuoteCalculator {
...
+            } else {
+                // This is a collapsed bridge order.
+                // Because collapsed bridge orders actually fill at different rates,
+                // we can iterate over the uncollapsed fills to get the actual
+                // asset amounts transfered.
+                // We can also assume there are no fees and the order is not
+                // partially filled.
+
+                // Infer the bridge slippage from the difference between the fill
+                // size and the atual order asset amounts.
```
dekz:
```suggestion
                // size and the actual order asset amounts.
```

PR #2447, opened 2020-01-17
fix bug in OrderTransferSimulationUtils causing failures for 721 assets

mzhu25:
## Checklist:


contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -82,13 +82,13 @@ contract OrderTransferSimulationUtils is
         // Transfer `makerAsset` from maker to taker
         assetData[0] = order.makerAssetData;
         fromAddresses[0] = order.makerAddress;
-        toAddresses[0] = takerAddress;
+        toAddresses[0] = takerAddress == address(0) ? address(1) : takerAddress;
```
jalextowle: This looks good to me. My only suggestion would be that we could put this logic into a function, but I think it's fine either way.

mzhu25:
moved to #2449

PR #2449, opened 2020-01-20
Fix regression in DevUtils

jalextowle:
## Description

This PR fixes some issues in DevUtils that were reintroduced after fixing the `takerAddress == address(0x0)` bug. 

## Checklist:


contracts/integrations/test/dev-utils/order_validation_utils.ts
```diff
@@ -155,6 +155,19 @@ blockchainTests.resets('OrderValidationUtils/OrderTransferSimulatorUtils', env =
                 .callAsync();
             expect(fillableTakerAssetAmount).to.bignumber.equal(constants.ZERO_AMOUNT);
         });
+        it.only('should correctly validate fillable order', async () => {
```
dekz:
```suggestion
        it('should correctly validate fillable order', async () => {
```

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -134,19 +134,19 @@ contract OrderTransferSimulationUtils is
         // Transfer `makerAsset` from maker to taker
         assetData[1] = order.makerAssetData;
         fromAddresses[1] = order.makerAddress;
-        toAddresses[1] = takerAddress;
+        toAddresses[1] = takerAddress == address(0) ? address(1) : takerAddress;
```
dekz: is this because some tokens prevent "burning" by sending to 0 address?

jalextowle: Yeah, tokens typically `require(to != address(0x0))`. Since we're using the `OrderTransferSimulations` contract as part of validation now, we don't want the transfer to fail because the recipient is address zero (since a taker won't actually be address zero).

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -82,13 +82,13 @@ contract OrderTransferSimulationUtils is
         // Transfer `makerAsset` from maker to taker
         assetData[0] = order.makerAssetData;
         fromAddresses[0] = order.makerAddress;
-        toAddresses[0] = takerAddress;
+        toAddresses[0] = takerAddress == address(0) ? address(1) : takerAddress;
```
abandeali1: I feel like we should us a random address instead of `address(1)`. While there is a low probability of `address(1)` being special cased for anything, it is almost infinitely more likely than any random address. 

We should also assign it to a constant that describes its purpose.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -133,6 +133,18 @@ contract OrderValidationUtils is
...
 
+        if (!_isAssetDataValid(order.takerAssetData)) {
+            fillableTakerAssetAmount = 0;
+        }
+
+        if (order.takerFee != 0 && !_isAssetDataValid(order.takerFeeAssetData)) {
+            fillableTakerAssetAmount = 0;
+        }
+
+        if (orderInfo.orderStatus == LibOrder.OrderStatus.CANCELLED) {
```
abandeali1: Is there any reason this shouldn't be `orderInfo.orderStatus != LibOrder.OrderStatus.FILLABLE`?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,47 @@ contract OrderValidationUtils is
...
+    ///      multiAssetData.
+    /// @param assetData The asset data that should be validated.
+    /// @return Whether or not the order should be considered valid.
+    function _isAssetDataValid(bytes memory assetData)
+        internal
+        pure
+        returns (bool)
+    {
+        // Only process the taker asset data if it is multiAssetData.
+        bytes4 assetProxyId = assetData.readBytes4(0);
```
abandeali1: Nit: we should probably check that `assetData.length > 3 && assetData%4 == 0` before doing anything else here.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,47 @@ contract OrderValidationUtils is
...
+        if (assetProxyId != IAssetData(address(0)).MultiAsset.selector) {
```
abandeali1: The same edge case exists for the `ERC1155Proxy`. We don't have to handle it in this PR, but we should add a TODO.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
jalextowle: Nope, good call.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
jalextowle: Shouldn't this condition be `assetData.length > 3 && assetData % 32 == 4`?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,77 @@ contract OrderValidationUtils is
...
+            } else if (nestedAssetProxyId == IAssetData(address(0)).ERC1155Assets.selector) {
+                hasSeenERC1155 = true;
+                if (_isAssetDataDuplicated(nestedAssetData, i)) {
+                    return false;
+                }
+            }
+
+            if (!hasSeenERC721 && !hasSeenERC1155) {
+                break;
+            }
```
mzhu25: What about the case where the taker asset is MAP(ERC20, Duplicate721, Duplicate721) ?

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,77 @@ contract OrderValidationUtils is
...
+                hasSeenERC721 = true;
+                if (_isAssetDataDuplicated(nestedAssetData, i)) {
+                    return false;
+                }
+            } else if (nestedAssetProxyId == IAssetData(address(0)).ERC1155Assets.selector) {
+                hasSeenERC1155 = true;
+                if (_isAssetDataDuplicated(nestedAssetData, i)) {
+                    return false;
+                }
+            }
```
mzhu25: what if it's a fungible 1155 asset?

mzhu25: what if it's a fungible 1155 asset?

jalextowle: Hmm good point. I'm going to leave this out for this PR in that case (implementing ERC1155 is optional and I'm not sure I have time this week to solve the general problem). I'll add a `TODO`

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
jalextowle: Good point. I'm just going to remove this for now.

packages/contract-addresses/addresses.json
```diff
@@ -13,13 +13,14 @@
...
         "coordinator": "0x38a795580d0f687e399913a00ddef6a17612c722",
+        "libTransactionDecoder": "0x5f20e82643ce007d87692eb1b3d3fc059588b224",
         "multiAssetProxy": "0xef701d5389ae74503d633396c4d654eabedc9d78",
         "staticCallProxy": "0x3517b88c19508c08650616019062b898ab65ed29",
         "erc1155Proxy": "0x7eefbd48fd63d441ec7435d024ec7c5131019add",
         "zrxVault": "0xba7f8b5fb1b19c1211c5d49550fcd149177a5eaf",
         "staking": "0x2a17c35ff147b32f13f19f2e311446eeb02503f3",
         "stakingProxy": "0xa26e80e7dea86279c6d778d702cc413e6cffa777",
-        "devUtils": "0x5f53f2aa72cb3a9371bf3c58e8fb3a313478b2f4",
+        "devUtils": "0xf8ba284450dcd7518bc5c68a4d93b60f734c3d72",
```
jalextowle: Reminder: The `devUtils` addresses will need to be updated to reflect the changes introduced because of @mzhu25's feedback

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
fabioberger: @jalextowle Augur is using ERC1155. Can you elaborate on what you mean by it being optional?

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
```diff
@@ -41,6 +41,10 @@ contract OrderTransferSimulationUtils is
         TransfersSuccessful       // All transfers in the order were successful
     }
 
+    // NOTE(jalextowle): This is a random address that we use to avoid issues that addresses like `address(1)`
+    // may cause later.
+    address constant internal NONZERO = address(0x377f698C4c287018D09b516F415317aEC5919332);
```
hysz: Nitty - this name is a bit confusing. Perhaps something that describes how it is used? Ex, `RANDOM_TAKER` or `RANDOM_ACTOR` or `UNUSED_ADDRESS`.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,63 @@ contract OrderValidationUtils is
...
+    function _isAssetDataDuplicated(
+        bytes[] memory nestedAssetData,
+        uint256 startIdx
+    )
+        internal
+        pure
+        returns (bool)
+    {
+        uint256 length = nestedAssetData.length;
+        for (uint256 i = startIdx + 1; i != length; i++) {
```
hysz: Nit - since `startIdx` is set by the caller it may be better be defensive and use `i < length`. 
(the practice of using `!=` over `<` came from the days where we overemphasized golfing)

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
```diff
@@ -185,4 +197,63 @@ contract OrderValidationUtils is
...
+        if (assetProxyId != IAssetData(address(0)).MultiAsset.selector) {
+            return true;
+        }
+
+        // Get array of values and array of assetDatas
+        (, uint256[] memory assetAmounts, bytes[] memory nestedAssetData) = decodeMultiAssetData(assetData);
+
+        uint256 length = nestedAssetData.length;
+        for (uint256 i = 0; i != length; i++) {
+            // TODO(jalextowle): Implement similar validation for non-fungible ERC1155 asset data.
```
hysz: Should we handle the other cases now and return `false` or is it convention to let it pass through?

contracts/dev-utils/contracts/src/OrderTransferSimulationUtils.sol
jalextowle: Hmm, I was using `NONZERO` because the only thing that is important about this address is *that it is nonzero*. That said, I would be okay with `UNUSED`.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
jalextowle: When you say "handle the other cases now", do you mean in this PR, or at that part of the code? We can't just return false, because then everything will be marked as invalid.

hysz: Ah, okay. Yeah I meant wrt validation. So basically if we don't recognize something then it is automatically valid?

contracts/dev-utils/contracts/src/DevUtils.sol
```diff
@@ -26,14 +26,12 @@ import "@0x/contracts-utils/contracts/src/LibEIP712.sol";
...
 import "./OrderValidationUtils.sol";
 import "./OrderTransferSimulationUtils.sol";
-import "./LibTransactionDecoder.sol";
 import "./EthBalanceChecker.sol";
 
 
 // solhint-disable no-empty-blocks
 contract DevUtils is
     OrderValidationUtils,
-    LibTransactionDecoder,
```
dorothy-zbornak: Would it make more sense to turn this into a public library and expose it through `DevUtils`?

jalextowle: I thought about that. In short, I'm not sure what would make more sense. I'm happy to do that.

contracts/dev-utils/contracts/src/OrderValidationUtils.sol
jalextowle: Yep. Otherwise, we'd have to recognize asset data for it to be valid, which I don't think makes sense. As of right now, we are never going to be able to recognize all asset data. This is important when you consider that people can't post orders to 0x-mesh unless they are valid. I don't think that we want to spend a ton of time *right now* writing more bespoke validation logic, so I think the best course of action is probably to implement a permissive version.

I'm very open to thoughts and criticisms of this argument though.

contracts/dev-utils/contracts/src/DevUtils.sol
dorothy-zbornak: Just seems like this approach could get untenable having to juggle so many addresses. But I don't even know if our deployment tools are up to the task of external linkage. Maybe just punt on it for now and I'll play with it when I'm doing the dydx validation.

jalextowle: I agree that we should probably make this a public library. I'm going to punt on this for this PR so that we can get this fix to TokenTrove, but feel free to put this on my plate when it comes up again.

PR #2453, opened 2020-01-23
[base-contract] throw string revert error

dekz:
Was throwing inside a try/catch and immediately getting caught. Exception was then re-thrown as:
```js
{
  message: 'VM execution error.',
  data:
   'Reverted 0x08c379a000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000025455243323042726964676553616d706c65722f494e56414c49445f544f4b454e5f50414952000000000000000000000000000000000000000000000000000000' 
}
```

### DIFF
--- a/packages/base-contract/src/index.ts
+++ b/packages/base-contract/src/index.ts
@@ -142,13 +142,13 @@ export class BaseContract {
         let revertError: RevertError;
         try {
             revertError = decodeThrownErrorAsRevertError(error);
-            // Re-cast StringRevertErrors as plain Errors for backwards-compatibility.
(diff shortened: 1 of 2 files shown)


packages/base-contract/src/index.ts
```diff
@@ -143,13 +143,13 @@ export class BaseContract {
         try {
             revertError = decodeThrownErrorAsRevertError(error);
             // Re-cast StringRevertErrors as plain Errors for backwards-compatibility.
```
dorothy-zbornak: Just move this down and g2g!

PR #2461, opened 2020-01-31
Stop hardcoding contract addresses

xianny:
## Description

In `@0x/asset-swapper`, some modules are instantiated by referring to addresses hardcoded in `@0x/contract-addresses`. This creates difficulty in testing because some contracts are not included on the Ganache snapshot or migration script e.g. ERC20BridgeSampler. 

This PR adds an optional `contractAddresses` param. If not provided, will fallback to previous behaviour of fetching from `@0x/contract-addresses`.

## Testing instructions

Tests should pass

## Types of changes

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/swap_quoter.ts
+++ b/packages/asset-swapper/src/swap_quoter.ts
@@ -159,7 +159,7 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
-        this._contractAddresses = getContractAddressesForChainOrThrow(chainId);
+        this._contractAddresses = options.contractAddresses || getContractAddressesForChainOrThrow(chainId);
         this._devUtilsContract = new DevUtilsContract(this._contractAddresses.devUtils, provider);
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -159,7 +159,7 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
-        this._contractAddresses = getContractAddressesForChainOrThrow(chainId);
+        this._contractAddresses = options.contractAddresses || getContractAddressesForChainOrThrow(chainId);
```
dorothy-zbornak: I wonder if it would be more useful to instead accept a `Partial<ContractAddresses>` and doing something like:
```ts
this._contractAddresses = {
    ...NULL_ADDRESSES,
    ...(() => {
        try {
            return getContractAddressesForChainOrThrow(chainId),
        } catch (err) {
            return {};
        }
    })(),
    ...options.contractAddresses
};
```
Though I guess the caller could just do that on their end. :man_shrugging:

packages/asset-swapper/src/types.ts
```diff
@@ -1,3 +1,4 @@
+import { ContractAddresses } from '@0x/migrations';
```
dekz: oops, fixing this in another PR

PR #2463, opened 2020-01-31
Support contracts with unlinked libraries

dorothy-zbornak:
*This is the second in a series of PRs that splits #2456 into more manageable pieces.*

## Description

`DevUtils` is at capacity and needs to be refactored into using public libraries if we hope to continue expanding it. This PR adds support to our infra for these contracts.

This is how public library linking works:
- `solc` compiles calls to public libraries as `delegatecall`s and generates bytecode with placeholder symbols in place of the library address.
- References to these symbols are maintained in a `linkReferences` object in `evm.bytecode.linkReferences` and `evm.deployedBytecode.linkReferences`.
- `abi-gen` now generates a new static function `deployWithLibrariesFrom0xArtifact()` for deploying contracts with unlinked libraries.
  - In addition to the usual, it takes a map of artifacts to find the artifacts for dependent  libraries.
  - It recursively walks the dependency graph, deploying each library depth-first.
  - It then fills in the address symbols in the bytecode of the calling contract with the deployed library address and deploys that contract.

Some notes:
- These contracts don't have fully formed bytecode, so they can't use an in-process vm to execute `pure` functions. This is now checked for in the generated wrappers.

## Checklist:


packages/base-contract/src/index.ts
```diff
@@ -77,6 +77,9 @@ export class PromiseWithTransactionHash<T> implements Promise<T> {
     public catch<TResult>(onRejected?: (reason: any) => Promise<TResult>): Promise<TResult | T> {
         return this._promise.catch(onRejected);
     }
+    public finally(onFinally?: (() => void) | null): Promise<T> {
+        return this._promise.finally(onFinally);
+    }
```
dorothy-zbornak: Sneaking this one in for compat with newer versions of TS.

packages/abi-gen/test-cli/output/typescript/abi_gen_dummy.ts
```diff
@@ -10,6 +10,7 @@ import {
     SubscriptionManager,
```
dorothy-zbornak: This file is generated, so you can just ignore it.

packages/abi-gen/test-cli/output/typescript/lib_dummy.ts
```diff
@@ -9,6 +9,7 @@ import {
     BaseContract,
```
dorothy-zbornak: This file is generated, so you can just ignore it.

packages/abi-gen/test-cli/output/typescript/test_lib_dummy.ts
```diff
@@ -9,6 +9,7 @@ import {
     BaseContract,
```
dorothy-zbornak: This file is generated, so you can just ignore it.

packages/base-contract/src/utils.ts
```diff
@@ -38,3 +38,30 @@ export function methodAbiToFunctionSignature(methodAbi: MethodAbi): string {
...
+export function linkLibrariesInBytecode(
+    bytecodeArtifact: EvmBytecodeOutput,
+    libraryAddresses: { [libraryName: string]: string },
+): string {
+    let bytecode = bytecodeArtifact.object.substr(2);
+    for (const link of Object.values(bytecodeArtifact.linkReferences)) {
+        for (const [libraryName, libraryRefs] of Object.entries(link)) {
+            const libraryAddress = libraryAddresses[libraryName];
+            if (!libraryAddress) {
+                continue;
```
mzhu25: I feel like we should be throwing here instead of in the contract wrapper, ideally with both the name of the contract whose bytecode we're modifying and the name of the missing library.

PR #2464, opened 2020-01-31
DevUtils refactor into public libraries

dorothy-zbornak:
*This is the third in a series of PRs that splits #2456 into more manageable pieces.*

## Description

Chill, most of these files are just regenerated artifacts/wrappers.

#### `@0x/contracts-dev-utils`:
- Refactors components of the `DevUtils` contract into public libraries, which get linked at deployment time.
- Deletes the standalone `LibTransactionDecoder` contract that was introduced in #2449  and folds it back into `DevUtils`.
- Adds `allowance()` and `balanceOf()` to `LibERC20Token`.
- Adds a `_getTransferableConvertedMakerAssetAmount()`, which currently just wraps `_getTransferableAssetAmount()`, but will be used for dydx maker assets in the next PR. This is necessary because dydx validation depends on the taker asset as well as the maker asset.

#### `@0x/migrations`:
- Migrations now uses the contract artifacts instead of `@0x/contract-wrappers`, even for ganache migrations.

## Checklist:


packages/migrations/src/migration.ts
```diff
@@ -28,6 +32,18 @@ import { SupportedProvider, TxData } from 'ethereum-types';
 import { constants } from './utils/constants';
 import { erc20TokenInfo, erc721TokenInfo } from './utils/token_info';
 
+const allArtifacts = {
```
hysz: ➕

packages/contract-addresses/addresses.json
```diff
@@ -127,21 +123,20 @@
...
         "coordinator": "0x4d3d5c850dd5bd9d6f4adda3dd039a3c8054ca29",
         "multiAssetProxy": "0xcfc18cec799fbd1793b5c43e773c98d4d61cc2db",
         "staticCallProxy": "0x6dfff22588be9b3ef8cf0ad6dc9b84796f9fb45f",
-        "devUtils": "0xa31e64ea55b9b6bbb9d6a676738e9a5b23149f84",
+        "devUtils": "0x74341e87b1c4db7d5ed95f92b37509f2525a7a90",
         "exchangeV2": "0x48bacb9266a570d521063ef5dd96e61686dbe788",
-        "zrxVault": "0x1941ff73d1154774d87521d2d0aaad5d19c8df60",
-        "staking": "0x0d8b0dd11f5d34ed41d556def5f841900d5b1c6b",
-        "stakingProxy": "0x38ef19fdf8e8415f18c307ed71967e19aac28ba1",
+        "zrxVault": "0xc4df27466183c0fe2a5924d6ea56e334deff146a",
```
hysz: Why did these need to be re-deployed?

contracts/dev-utils/contracts/src/AssetBalance.sol
```diff
@@ -0,0 +1,374 @@
+/*
```
hysz: Did any of the logic change between the code in contracts/src/LibAssetData.sol and this, or just refactored?

contracts/erc20/contracts/src/LibERC20Token.sol
```diff
@@ -107,6 +108,50 @@ library LibERC20Token {
         }
     }
 
+    /// @dev Retrieves the allowance for a token, owner, and spender.
+    ///      Returns `0` if the call reverts.
+    /// @param token The address of the token contract.
+    /// @param owner The owner of the tokens.
+    /// @param spender The address the spender.
+    /// @return allowance The allowance for a token, owner, and spender.
+    function allowance(address token, address owner, address spender)
```
hysz: Why do we return 0 instead of reverting if the call fails?

packages/contract-addresses/addresses.json
dorothy-zbornak: These are ganache snapshot addresses. Every public library in `DevUtils` is a separate deployment now, so it moves the deployer's nonce so we end up with different addresses.

contracts/dev-utils/contracts/src/AssetBalance.sol
dorothy-zbornak: Mostly refactored. `_getConvertibleMakerBalanceAndAssetProxyAllowance()` got added but it's just a no-op placeholder in this PR, so you'll see it again.

contracts/erc20/contracts/src/LibERC20Token.sol
dorothy-zbornak: This library is all about very permissive interactions. Plus I figured we really don't want to revert in DevUtils if we can help it.

PR #2465, opened 2020-01-31
Add liquidity source breakdown to asset-swapper

dave4506:
## Description

Adds an addition `sourceBreakdown` entry to `SwapQuote` provides a breakdown of the orders by liquidity source.

Example:
```
{
    "Kyber": 0.2,
    "Native": 0.4,
    "Uniswap": 0.1,
    "Eth2Dai": 0.3
}
```

## Checklist:

### DIFF
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@0x/asset-swapper",
-    "version": "4.1.0",
+    "version": "4.0.1",
     "engines": {
         "node": ">=6.12"
     },
--- a/packages/asset-swapper/package.json
+++ b/packages/asset-swapper/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@0x/asset-swapper",
-    "version": "4.0.1",
+    "version": "4.1.0",
     "engines": {
         "node": ">=6.12"
     },
--- a/packages/asset-swapper/src/utils/swap_quote_calculator.ts
+++ b/packages/asset-swapper/src/utils/swap_quote_calculator.ts
@@ -413,9 +413,9 @@ export class SwapQuoteCalculator {
         const breakdown: SwapQuoteOrdersBreakdown = {};
 
         // total asset amount (accounting for slippage protection)
-        const totalAssetAmount = orders.reduce((amount: BigNumber, order: OptimizedMarketOrder): BigNumber => {
-            return amount.plus(operation === MarketOperation.Buy ? order.makerAssetAmount : order.takerAssetAmount);
-        }, constants.ZERO_AMOUNT);
+        const totalAssetAmount = BigNumber.sum(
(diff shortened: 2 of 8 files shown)


packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -266,6 +279,49 @@ export class SwapQuoteCalculator {
...
+        return Object.keys(aggregatedFillAmountBySource).reduce((acc: SwapQuoteOrdersBreakdown, source: string): SwapQuoteOrdersBreakdown => {
+            return {
+                ...{
+                    [source]: aggregatedFillAmountBySource[source].dividedBy(totalAssetAmount),
+                },
+                ...acc,
+            };
+        }, breakdown);
+    }
 }
```
dorothy-zbornak: There's actually a more direct way to get this info.  `getMarketSellOrdersAsync()` now returns a `OptimizedMarketOrder[]` type, which each have a `fill` field that specifies the source and asset amount.

dave4506: <face palm>

packages/asset-swapper/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@0x/asset-swapper",
-    "version": "4.0.1",
+    "version": "4.1.0",
```
dorothy-zbornak: Eh? Are we publishing this separately or something?

packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -398,6 +403,31 @@ export class SwapQuoteCalculator {
...
+        orders: OptimizedMarketOrder[],
+        operation: MarketOperation,
+    ): SwapQuoteOrdersBreakdown {
+        // HACK: to shut up linter
+        const breakdown: SwapQuoteOrdersBreakdown = {};
+
+        // total asset amount (accounting for slippage protection)
+        const totalAssetAmount = orders.reduce((amount: BigNumber, order: OptimizedMarketOrder): BigNumber => {
+            return amount.plus(operation === MarketOperation.Buy ? order.makerAssetAmount : order.takerAssetAmount);
+        }, constants.ZERO_AMOUNT);
```
dorothy-zbornak: Nit
```ts
const totalAssetAmount = BigNumber.sum(
    ...[0, orders.map(o => operation === MarketOperation.Buy ? o.makerAssetAmount : o.takerAssetAmount)],
);
```

packages/asset-swapper/package.json
dave4506: Ope, I was changing that when debugging something, my bad

packages/asset-swapper/src/utils/swap_quote_calculator.ts
dave4506: clean af

fragosti:
Some of the other dex aggregators have a return format of:
```
[
{
  name: 'Kyber',
  amount: ...
  proprotion: ...
}
]
```
I feel like this is more extensible which is what we want in the API. Can we change the format to be more like this?

PR #2466, opened 2020-02-01
DevUtils: DydxBridge validation

dorothy-zbornak:
*This is the fourth and final in a series of PRs that splits #2456 into more manageable pieces.*

## Description

This adds validation for dydx bridge orders. The heavy lifting is done in [LibDydxBalance](https://github.com/0xProject/0x-monorepo/blob/feat/contracts/dev-utils/dydx-validation/contracts/dev-utils/contracts/src/LibDydxBalance.sol).

Fillable amount calculation will zero out if any of these conditions aren't met:
- Should only have a dydx bridge in the *maker* asset. We don't handle it in any other assets, which would fail by default.
- Must have actions in the form: `[...deposits, withdraw]`, i.e., 0 or more deposit actions and one withdraw action at the end. There are some additional restrictions on these actions.
  - The `marketId` (token) for the withdraw action must be the same as the bridge token.
  - You can have multiple deposits, but `marketId`s must be unique amongst them.
- The bridge must be an "operator" for the maker.
- The rate of withdrawal must be >= 1.0, i.e., the order must return >= the maker asset amount asked of the bridge proxy.

Afterwards, we compute the fillable maker asset amount, which is the the minimum of:
- `_getDepositableMakerAmount() -> m_0`: The maker asset quantity after which the deposit actions can no longer be performed.
  - This depends on the the respective token balance & allowance of the maker for the deposit market and, if the token is also the taker asset, the rate at which the taker asset is exchanged to the maker.
- `__getSolventMakerAmount() -> m_1`: The maker asset quantity after which the margin account will fall below the minimum collateralization ratio.
  - The collateralization ratio of an account is the USD value of all deposited tokens over the USD value of all borrowed tokens.
  - This also takes into account the taker asset transfer rate.

Computing `m_1` is given by:

getSolventMakerAmount formula

Where
- `Ω` is the minimum collateralization ratio required for an account to be solvent.
- `α` is the current USD value all collateral deposited into the account.
- `ɣ` is the current USD value of all tokens borrowed by the account.
- `δα` is the rate at which USD value is deposited into the account.
- `δɣ` is the rate at which USD value is withdrawn from the account.

Of course, all this means we have...

#### More Crappy Solidity Integer Math
The `D18` library has been added to `dev-utils`. It used to have a longer name but math is easier to follow with short names, imo.

Another thing we need to be mindful of when building these orders is that the precision of the calculations is essentially limited by the lowest precision token involved. Like when doing tests on USDC -> DAI orders, the maximum precision I can get out of the calculations is around 5 or 6 decimal places. Scaling the `conversionRateNumerator` and `conversionRateDenominator` by some constant factor might increase this precision. 

#### What's Missing
These cases aren't currently supported but should be technically possible with some extra effort:

- Deposits of the same token into multiple accounts.
- Withdraws from multiple accounts.

## Checklist:


contracts/dev-utils/contracts/src/LibDydxBalance.sol
```diff
@@ -0,0 +1,418 @@
...
+    {
+        bytes memory rawBridgeData;
+        (, info.makerTokenAddress, info.bridgeAddress, rawBridgeData) =
+            LibAssetData.decodeERC20BridgeAssetData(order.makerAssetData);
+        info.dydx = IDydx(dydx);
+        info.makerAddress = order.makerAddress;
+        if (order.takerAssetData.readBytes4(0) == IAssetData(0).ERC20Token.selector) {
+            (, info.takerTokenAddress) =
+                LibAssetData.decodeERC20AssetData(order.takerAssetData);
+        }
```
hysz: I wonder if we should revert if we cannot extract the taker token address. O/w it seems the other functions may not perform as expected. Failing here may be more graceful.

contracts/dev-utils/contracts/src/LibDydxBalance.sol
```diff
@@ -0,0 +1,418 @@
...
+import "@0x/contracts-utils/contracts/src/LibBytes.sol";
+import "@0x/contracts-utils/contracts/src/LibSafeMath.sol";
+import "./LibAssetData.sol";
+import "./D18.sol";
+
+
+library LibDydxBalance {
+
+    using LibBytes for bytes;
+    using LibSafeMath for uint256;
```
hysz: Nit - `using D18 for uint256` and `using D18 for int256` could improve readability and consistency w other code. But it's not too bad so up to you 🤷‍♂.

contracts/dev-utils/contracts/src/D18.sol
```diff
@@ -0,0 +1,249 @@
...
+    /// @dev Safely add two signed integers.
+    function _add(int256 a, int256 b)
+        private
+        pure
+        returns (int256 r)
+    {
+        r = a + b;
+        require(
+            !((a < 0 && b < 0 && r > a) || (a > 0 && b > 0 && r < a)),
+            "D18/DECIMAL_MUL_OVERFLOW"
```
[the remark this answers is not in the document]
dorothy-zbornak: Eesh, idk where my head was at when I wrote this file.

contracts/dev-utils/contracts/src/LibDydxBalance.sol
```diff
@@ -0,0 +1,418 @@
...
+                action.conversionRateNumerator,
+                action.conversionRateDenominator
+            );
+    }
+
+    /// @dev Returns the USD value of an action based on its conversion rate
+    ///      and market prices.
+    /// @param info State from `_getBalanceCheckInfo()`.
+    /// @param action A `BridgeAction`.
+    function _getActionRateValue(
```
hysz: Nit - I think a name like `_getMonetaryValue` could be more descriptive and aligned with naming used by dYdX.

contracts/dev-utils/contracts/src/LibDydxBalance.sol
```diff
@@ -0,0 +1,418 @@
...
+            //      `cr = (supplyValue + t * dd) / (borrowValue + t * db)`
+            // Solving for `t` gives us:
+            //      `t = (supplyValue - cr * borrowValue) / (cr * db - dd)`
+            int256 t = D18.div(
+                D18.sub(supplyValue, D18.mul(minCr, borrowValue)),
+                D18.sub(D18.mul(minCr, db), dd)
+            );
+            solventMakerAmount = LibSafeMath.min256(
+                solventMakerAmount,
+                // `t` is in maker token units, so convert it to maker wei.
```
hysz: I may be thinking about this in totally the wrong way, but isn't `t` unitless here?

Suppose:
- Unit of `supplyValue` is `a`. The unit of `dd` is then also `a`.
- Unit of `borrowValue` is `b`. The unit of `db` is then also `b`.
- Unit of `cr` is then `a/b`. 

Find the unit of measure of `t`:
```
t = (supplyValue - cr * borrowValue) / (cr * db - dd)
→
t unit of measure = (a - (a/b) * b) / ((a/b) * b - a)
```
The units all cancel out, leaving `t` as just a scalar. 

If `t` were unitless then I think we would interpret it as the # of times we can execute the trade:
* At `t=1` we can execute the entire trade.
* At `t<1` we cannot execute the trade.
* At `t>1` we can execute the trade more than once.

So any value of `t≥1` would be valid for an order that buys `supplyValue` and sells `borrowValue`.

dorothy-zbornak: Ye right. The `_toWei()` on the next line converts it back to maker units. The comment is wrong.

contracts/dev-utils/contracts/src/AssetBalance.sol
```diff
@@ -366,6 +369,17 @@ contract AssetBalance is
         if (order.makerAssetData.length < 4) {
             return (0, 0);
         }
+        bytes4 assetProxyId = order.makerAssetData.readBytes4(0);
+        // Handle dydx bridge assets.
+        if (assetProxyId == IAssetData(address(0)).ERC20Bridge.selector) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Technically this isn't *really* validating the taker side, just the expectations on the maker side.

I think massively simplifying the validation is definitely worth it on our side, but it probably does hurt other teams. I can't speak for Mesh, but I don't think we would use these orders at all in the swapper if we couldn't estimate how fillable they are. (Though whether we even use them remains to be seen.)

contracts/dev-utils/contracts/src/LibDydxBalance.sol
dorothy-zbornak: Actually, I kind of like the long-hand since it makes it clearer that these `uint256`s are expressed in a different unit.

contracts/asset-proxy/contracts/src/interfaces/IDydx.sol
```diff
@@ -71,6 +71,36 @@ interface IDydx {
...
 
+    struct D256 {
+        uint256 value;
+    }
+
+    struct Value {
+        uint256 value;
+    }
+
+    struct Price {
```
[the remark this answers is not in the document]
dorothy-zbornak: I hate this too, but it makes it easier to cross reference with the Dydx contracts which use the same named structs.

contracts/dev-utils/contracts/src/LibDydxBalance.sol
dorothy-zbornak: Actually, this already would revert but I think we should do the opposite and not revert. The other functions do check if the taker token is null.

PR #2468, opened 2020-02-04
Bump maxBuffer?

fragosti:
Just applying the working fix from #2460


dekz:
won't you need this as well? https://github.com/0xProject/0x-monorepo/pull/2460/files#diff-cf5889edb0d5c8f06a52d29dad477ec2R29

PR #2481, opened 2020-02-12
Asset-Swapper: Incorporate fees into fill path optimization

dorothy-zbornak:
## Description

Small orders are hurting the bottom line on small swaps because of the fees involved in filling them. `asset-swapper` needs to account for gas and protocol fees to intelligently skip over these orders if they are not profitable. This will also help wrt gas-heavy DEXes like Kyber.
 

### Approach
The basic idea is to express the gas + protocol fees, which are denominated in ETH, as an equivalent amount of maker or taker asset (depending on whether we're doing a buy or sell). This "penalty" is then added or subtracted to the fill output value during the optimization phase, which proceeds as normal.

To express the fees as the maker or taker asset, we ask the sampler contract for an additional sell quote of 1 ETH->X, where X is either the maker (sells) or taker (buys) asset. This allows us to convert ETH-denominated fees into the equivalent "output" quantity for optimization.

### `fees` option
There is now  a `fees` option. This is a mapping of `ERC20BridgeSource` -> ETH fee. This is up to the caller to supply as by default it will assume no fees for any fill. This seems more flexible/scalable than baking it into `asset-swapper` since we'll probably be tweaking these values.

Each fee supplied should be the estimated cost of filling an order on each liquidity source, as gas. For example, you might compute the fee for native orders as: `(AVERAGE_FILL_GAS + PROTOCOL_FEE_MULTIPLIER)`. This will get scaled by the gas price automatically.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
@@ -3,7 +3,6 @@ import { assetDataUtils, generatePseudoRandomSalt } from '@0x/order-utils';
 import { AbiEncoder, BigNumber } from '@0x/utils';
 
 import { constants } from '../../constants';
-import { sortingUtils } from '../../utils/sorting_utils';
 
 import { constants as marketOperationUtilConstants } from './constants';
 import {
@@ -50,7 +49,7 @@ export class CreateOrderUtils {
                 );
             }
         }
-        return sortingUtils.sortOrders(orders);
+        return orders;
     }
 
     // Convert buy fills into orders.
@@ -79,7 +78,7 @@ export class CreateOrderUtils {
                 );
             }
         }
-        return sortingUtils.sortOrders(orders);
+        return orders;
     }
 
     private _getBridgeAddressFromSource(source: ERC20BridgeSource): string {
--- a/packages/asset-swapper/src/utils/market_operation_utils/fill_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fill_optimizer.ts
@@ -10,7 +10,7 @@ const { ZERO_AMOUNT } = constants;
 interface FillsOptimizerContext {
     currentPath: Fill[];
     currentPathInput: BigNumber;
-    currentPathOutput: BigNumber;
+    currentPathAdjustedOutput: BigNumber;
     currentPathFlags: number;
 }
 
@@ -22,7 +22,7 @@ export class FillsOptimizer {
     private readonly _shouldMinimize: boolean;
     private _currentRunCount: number = 0;
     private _optimalPath?: Fill[] = undefined;
-    private _optimalPathOutput: BigNumber = ZERO_AMOUNT;
+    private _optimalPathAdjustedOutput: BigNumber = ZERO_AMOUNT;
 
     constructor(runLimit: number, shouldMinimize?: boolean) {
         this._runLimit = runLimit;
@@ -32,24 +32,27 @@ export class FillsOptimizer {
     public optimize(fills: Fill[], input: BigNumber, upperBoundPath?: Fill[]): Fill[] | undefined {
         this._currentRunCount = 0;
         this._optimalPath = upperBoundPath;
-        this._optimalPathOutput = upperBoundPath ? getPathOutput(upperBoundPath, input) : ZERO_AMOUNT;
+        this._optimalPathAdjustedOutput = upperBoundPath ? getPathAdjustedOutput(upperBoundPath, input) : ZERO_AMOUNT;
         const ctx = {
             currentPath: [],
             currentPathInput: ZERO_AMOUNT,
-            currentPathOutput: ZERO_AMOUNT,
+            currentPathAdjustedOutput: ZERO_AMOUNT,
             currentPathFlags: 0,
         };
         // Visit all valid combinations of fills to find the optimal path.
         this._walk(fills, input, ctx);
-        return this._optimalPath;
+        if (this._optimalPath) {
+            return sortFillsByAdjustedRate(this._optimalPath, this._shouldMinimize);
+        }
+        return undefined;
     }
 
     private _walk(fills: Fill[], input: BigNumber, ctx: FillsOptimizerContext): void {
-        const { currentPath, currentPathInput, currentPathOutput, currentPathFlags } = ctx;
+        const { currentPath, currentPathInput, currentPathAdjustedOutput, currentPathFlags } = ctx;
 
         // Stop if the current path is already complete.
         if (currentPathInput.gte(input)) {
-            this._updateOptimalPath(currentPath, currentPathOutput);
+            this._updateOptimalPath(currentPath, currentPathAdjustedOutput);
             return;
         }
 
@@ -67,8 +70,8 @@ export class FillsOptimizer {
             }
             const nextPath = [...currentPath, nextFill];
             const nextPathInput = BigNumber.min(input, currentPathInput.plus(nextFill.input));
-            const nextPathOutput = currentPathOutput.plus(
-                getPartialFillOutput(nextFill, nextPathInput.minus(currentPathInput)),
+            const nextPathAdjustedOutput = currentPathAdjustedOutput.plus(
+                getPartialFillOutput(nextFill, nextPathInput.minus(currentPathInput)).minus(nextFill.fillPenalty),
             );
             // tslint:disable-next-line: no-bitwise
             const nextPathFlags = currentPathFlags | nextFill.flags;
@@ -80,7 +83,7 @@ export class FillsOptimizer {
                 {
                     currentPath: nextPath,
                     currentPathInput: nextPathInput,
-                    currentPathOutput: nextPathOutput,
+                    currentPathAdjustedOutput: nextPathAdjustedOutput,
                     // tslint:disable-next-line: no-bitwise
                     currentPathFlags: nextPathFlags,
                 },
@@ -88,10 +91,10 @@ export class FillsOptimizer {
         }
     }
 
-    private _updateOptimalPath(path: Fill[], output: BigNumber): void {
-        if (!this._optimalPath || this._compareOutputs(output, this._optimalPathOutput) === 1) {
+    private _updateOptimalPath(path: Fill[], adjustedOutput: BigNumber): void {
+        if (!this._optimalPath || this._compareOutputs(adjustedOutput, this._optimalPathAdjustedOutput) === 1) {
             this._optimalPath = path;
-            this._optimalPathOutput = output;
+            this._optimalPathAdjustedOutput = adjustedOutput;
         }
     }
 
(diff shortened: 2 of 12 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -63,8 +72,14 @@ export class MarketOperationUtils {
...
         const [makerToken, takerToken] = getOrderTokens(nativeOrders[0]);
-        const [fillableAmounts, dexQuotes] = await this._sampler.executeAsync(
+        const [fillableAmounts, ethToMakerAssetRate, dexQuotes] = await this._sampler.executeAsync(
             DexOrderSampler.ops.getOrderFillableTakerAmounts(nativeOrders),
+            DexOrderSampler.ops.getMedianSellRate(
+                difference(FEE_QUOTE_SOURCES, _opts.excludedSources),
+                this._wethAddress,
+                makerToken,
+                ONE_ETHER,
+            ),
```
dorothy-zbornak: Right now this does not use native orders in the interest of response time (don't want an additional query to mesh). It probably doesn't really matter though since we really just need a reasonable estimate.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -78,19 +93,16 @@ export class MarketOperationUtils {
             MarketOperation.Sell,
         );
 
-        const prunedNativePath = pruneDustFillsFromNativePath(
-            createSellPathFromNativeOrders(nativeOrdersWithFillableAmounts),
-            takerAmount,
-            _opts.dustFractionThreshold,
-        );
-        const clippedNativePath = clipPathToInput(sortFillsByPrice(prunedNativePath), takerAmount);
```
dorothy-zbornak: Clipping native orders is dumb and inaccurate because a native order can only be clipped at fill time if it's actually positioned last in the order queue.

packages/asset-swapper/src/utils/market_operation_utils/fill_optimizer.ts
```diff
@@ -32,24 +32,27 @@ export class FillsOptimizer {
...
+            currentPathAdjustedOutput: ZERO_AMOUNT,
             currentPathFlags: 0,
         };
         // Visit all valid combinations of fills to find the optimal path.
         this._walk(fills, input, ctx);
-        return this._optimalPath;
+        if (this._optimalPath) {
+            return sortFillsByAdjustedRate(this._optimalPath, this._shouldMinimize);
+        }
+        return undefined;
```
dorothy-zbornak: Curiously, there was no strong guarantee that the optimizer would return fills in descending exchange rate. This was fine because later, after creating the orders, we would sort them by rate there. Sorting the orders afterwards won't take into account tx + gas costs though, so they now need to be done at the fill level.

packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
```diff
@@ -50,7 +49,7 @@ export class CreateOrderUtils {
                 );
             }
         }
-        return sortingUtils.sortOrders(orders);
```
dorothy-zbornak: Re: My previous comment about needing to sort orders at the fill level.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -331,31 +372,59 @@ function createBuyPathFromNativeOrders(orders: SignedOrderWithFillableAmounts[])
...
+                exclusionMask: opts.noConflicts ? sourceToExclusionMask(sample.source) : 0,
                 fillData: { source: sample.source },
             });
             prevSample = quote[i];
         }
+        // Don't push empty paths.
         if (path.length > 0) {
-            // Don't push empty paths.
+            // Only the first fill in a DEX path incurs a penalty.
+            path[0].fillPenalty = ethToOutputAssetRate.times(opts.fees[path[0].fillData.source] || 0);
```
dekz:
```suggestion
            path[0].fillPenalty = ethToOutputAssetRate.times(opts.fees[path[0].fillData.source] || 0).integerValue();
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -164,4 +161,8 @@ export interface GetMarketOrdersOpts {
      * Default: 1.25.
      */
     sampleDistributionBase: number;
+    /**
+     * Fees for each liquidity source, expressed in ETH wei.
```
dekz: This should be in gas*gasPrice right since it is simply multiplied by the base unit amount for the `fillPenalty`? I.e not `240000` but `240000*1e9`.

So the 0x penalty for `USDC/DAI` is:
```
ethToMakerAssetRate 0.000000000266756647
fillPenalty 640215 = 2400000000000000 * 0.000000000266756647
```

Without the gwei multiplier the penalty is too small.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -63,8 +72,14 @@ export class MarketOperationUtils {
...
         };
         const [makerToken, takerToken] = getOrderTokens(nativeOrders[0]);
-        const [fillableAmounts, dexQuotes] = await this._sampler.executeAsync(
+        const [fillableAmounts, ethToMakerAssetRate, dexQuotes] = await this._sampler.executeAsync(
             DexOrderSampler.ops.getOrderFillableTakerAmounts(nativeOrders),
+            DexOrderSampler.ops.getMedianSellRate(
+                difference(FEE_QUOTE_SOURCES, _opts.excludedSources),
+                this._wethAddress,
+                makerToken,
+                ONE_ETHER,
```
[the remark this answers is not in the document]
dorothy-zbornak: Good catch! The token args are backwards here.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -287,19 +315,25 @@ function difference<T>(a: T[], b: T[]): T[] {
...
+        const makerAmount = fillableAmountsUtils.getMakerAssetAmountSwappedAfterOrderFees(order);
+        const takerAmount = fillableAmountsUtils.getTakerAssetAmountSwappedAfterOrderFees(order);
         // Native orders can be filled in any order, so they're all root nodes.
         path.push({
             flags: FillFlags.SourceNative,
             exclusionMask: 0,
             input: takerAmount,
             output: makerAmount,
+            // Every fill from native orders incurs a penalty.
+            fillPenalty: ethToOutputAssetRate.times(opts.fees[ERC20BridgeSource.Native] || 0),
```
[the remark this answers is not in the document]
dorothy-zbornak: I thought the only fees we return are protocol fees?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
dorothy-zbornak: I'm kind of leaning towards keeping these full precision since they're just used for comparison.

packages/asset-swapper/src/utils/market_operation_utils/types.ts
dorothy-zbornak: Yeah, the gas price needs to be incorporated in these because the optimizer won't assume one. I'll update the comment to be more clear.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -136,8 +148,14 @@ export class MarketOperationUtils {
...
         const [makerToken, takerToken] = getOrderTokens(nativeOrders[0]);
-        const [fillableAmounts, dexQuotes] = await this._sampler.executeAsync(
+        const [fillableAmounts, ethToTakerAssetRate, dexQuotes] = await this._sampler.executeAsync(
             DexOrderSampler.ops.getOrderFillableMakerAmounts(nativeOrders),
+            DexOrderSampler.ops.getMedianSellRate(
+                difference(FEE_QUOTE_SOURCES, _opts.excludedSources),
+                this._wethAddress,
+                makerToken,
+                ONE_ETHER,
+            ),
```
dorothy-zbornak: This one is also wrong because it's actually getting maker -> ETH (we want ETH -> taker).

packages/asset-swapper/src/utils/market_operation_utils/index.ts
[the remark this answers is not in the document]
dekz: Sorry I mean, higher up the stack, potentially even higher in 0xApi, we should be setting the fee penalty based on the current gas price determined.

I.e set it with every request to be more accurate to the market, than set it globally once.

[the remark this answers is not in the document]
dorothy-zbornak: Oh, yes.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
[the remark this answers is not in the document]
dekz: This should also handle the case where `makerToken` is WETH otherwise the Sampler will throw, in the case the rate should just be `1`.

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -167,4 +164,12 @@ export interface GetMarketOrdersOpts {
...
+     * Gas price is not assumed by the optimizer, so these fees should be
+     * pre-scaled by the gas price if one wishes to factor it in.
+     * E.g, if gas costs for a source is 150k and gas price is 1 gwei, the fee
+     * given here should be 150k * 1 gwei.
+     */
+    fees: { [source: string]: BigNumber };
```
dekz: Hmm on second thought I think we should keep this in terms of gas. And in `swap_quoter.ts` we scale it up before giving it to the `swapQuoteCalculator`. Since at this point we have a `gasPrice` from either the user or from ethgasstation.

Something like: 
```typescript
 let swapQuote: SwapQuote;
        const swapQuoteWithAdjustedFees = {
            ...calculateSwapQuoteOpts,
            fees: Object.assign({}, ...Object.entries(calculateSwapQuoteOpts.fees).map(([k, v]) => ({ [k]: v.times(gasPrice) }))),
        };

        if (marketOperation === MarketOperation.Buy) {
            swapQuote = await this._swapQuoteCalculator.calculateMarketBuySwapQuoteAsync(
                prunedOrders,
                assetFillAmount,
                slippagePercentage,
                gasPrice,
                swapQuoteWithAdjustedFees,
            );
        } else {
            swapQuote = await this._swapQuoteCalculator.calculateMarketSellSwapQuoteAsync(
                prunedOrders,
                assetFillAmount,
                slippagePercentage,
                gasPrice,
                swapQuoteWithAdjustedFees,
            );
        }
```

packages/asset-swapper/src/utils/market_operation_utils/fill_optimizer.ts
```diff
@@ -32,24 +32,27 @@ export class FillsOptimizer {
...
             currentPathInput: ZERO_AMOUNT,
-            currentPathOutput: ZERO_AMOUNT,
+            currentPathAdjustedOutput: ZERO_AMOUNT,
             currentPathFlags: 0,
         };
         // Visit all valid combinations of fills to find the optimal path.
         this._walk(fills, input, ctx);
-        return this._optimalPath;
+        if (this._optimalPath) {
+            return sortFillsByAdjustedRate(this._optimalPath, this._shouldMinimize);
```
dekz: This can return ALL native orders if the `runLimit` is reached. Problem with that is the amount of liquidity available can far out weigh the input liquidity. I.e a 10 USDC -> DAI sell can return 20+ 0x orders. This has some downstream issues with protocol fee calculation and gas costs.

packages/asset-swapper/src/utils/market_operation_utils/types.ts
merklejerk: Yeah, I was trying to be agnostic to future fees that aren't necessarily gas price dependent but we don't really need that abstraction right now.

packages/asset-swapper/src/utils/market_operation_utils/fill_optimizer.ts
dorothy-zbornak: Added order pruning (but not clipping) to ameliorate this issue.

PR #2480, opened 2020-02-11
Curve ERC20Bridge

dekz:
## Description

Trades USDC<>DAI on Curve

PR #2483, opened 2020-02-13
ERC20BridgeSampler Curve

dekz:
## Description

Samples the CurveBridge

https://github.com/0xProject/0x-monorepo/pull/2480


packages/asset-swapper/src/utils/market_operation_utils/sampler.ts
```diff
@@ -128,29 +147,39 @@ const samplerOperations = {
...
-                return samplerOperations.getEth2DaiSellQuotes(makerToken, takerToken, takerFillAmounts);
+                batchedOperation = samplerOperations.getEth2DaiSellQuotes(makerToken, takerToken, takerFillAmounts);
             } else if (source === ERC20BridgeSource.Uniswap) {
-                return samplerOperations.getUniswapSellQuotes(makerToken, takerToken, takerFillAmounts);
+                batchedOperation = samplerOperations.getUniswapSellQuotes(makerToken, takerToken, takerFillAmounts);
             } else if (source === ERC20BridgeSource.Kyber) {
-                return samplerOperations.getKyberSellQuotes(makerToken, takerToken, takerFillAmounts);
+                batchedOperation = samplerOperations.getKyberSellQuotes(makerToken, takerToken, takerFillAmounts);
+            } else if (source === ERC20BridgeSource.Curve) {
+                // TODO(dekz) best way to pass this data around, in the world of multiple curves at once?
```
dekz: Not happy with this at the moment, there can be multiple Curve sources with the same tokens, like right now there is USDC/DAI and USDT/USDC/DAI. So it is not easy to tell the `create_order` which `curveAddress` should be used. 

@dorothy-zbornak any thoughts on the right abstraction here? Pass it through `DexSample`? What will PLP look like as well?

contracts/erc20-bridge-sampler/test/mainnet_tests.ts
```diff
@@ -0,0 +1,39 @@
...
+        const curveAddress = '0x2e60CF74d81ac34eB21eEff58Db4D385920ef419';
+        const daiTokenIdx = new BigNumber(0);
+        const usdcTokenIdx = new BigNumber(1);
+
+        it('samples sells from Curve DAI->USDC', async () => {
+            const samples = await testContract
+                .sampleSellsFromCurve(curveAddress, daiTokenIdx, usdcTokenIdx, [toBaseUnitAmount(1)])
+                .callAsync();
+            expect(samples.length).to.be.bignumber.greaterThan(0);
+            expect(samples[0]).to.be.bignumber.greaterThan(0);
```
dorothy-zbornak: I guess one of the things I get paranoid about is whether these endpoints are expressing the correct units. So maybe we can check that `samples[0] > 0 && samples[0] < toBaseUnitAmount(2, 6)`

packages/asset-swapper/src/utils/market_operation_utils/create_order.ts
```diff
@@ -106,13 +109,26 @@ function createBridgeOrder(
...
     isBuy: boolean = false,
 ): OptimizedMarketOrder {
-    return {
-        makerAddress: bridgeAddress,
-        makerAssetData: assetDataUtils.encodeERC20BridgeAssetData(
+    let makerAssetData;
+    if (fill.source === ERC20BridgeSource.CurveUsdcDai || fill.source === ERC20BridgeSource.CurveUsdcDaiUsdt) {
+        const { curveAddress, tokens } = constants.DEFAULT_CURVE_OPTS[fill.source];
+        const fromTokenIdx = tokens.indexOf(takerToken);
+        const toTokenIdx = tokens.indexOf(takerToken);
```
dorothy-zbornak:
```suggestion
        const toTokenIdx = tokens.indexOf(makerToken);
```

PR #2489, opened 2020-02-20
[PLP] Adds a simple wrapper around the PLP registry that provides caching layer

PirosB3:
## Description

Adds a simple wrapper around the `RegistryContract` interface (that is loosely coupled to the IPLPRegistryContract). This wrapper exposes a wrapper around `getPoolForMarketAsync()` and provides a useful caching layer that is meant to drastically reduce the amount of on-chain contract calls.  

## Testing instructions

Aside from adding unit tests, I’ve also done some integration testing: I've created a simple `main.ts` file (not committed) that initializes a `PLPRegistry` instance with a `IPLPRouterContract` instance, that was initialized with a dummy PLP contract that I deployed on-chain. All integration tests worked as expected.

## Types of changes

## Checklist:


packages/asset-swapper/test/plp_registry_test.ts
```diff
@@ -0,0 +1,150 @@
+// tslint:disable: custom-no-magic-numbers
+import { ContractFunctionObj } from '@0x/base-contract';
+import { randomAddress } from '@0x/contracts-test-utils';
+import { BlockParamLiteral, CallData } from '@0x/dev-utils';
+import * as chai from 'chai';
+import 'mocha';
+import * as TypeMoq from 'typemoq';
```
dorothy-zbornak: Never used this library before. Very cool!

packages/asset-swapper/test/plp_registry_test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+import { PLPRegistry, RegistryContract } from '../src/quote_consumers/plp_registry';
+
+import { chaiSetup } from './utils/chai_setup';
+
+chaiSetup.configure();
+const expect = chai.expect;
+const ONE_MINUTE_IN_MILLIS = 1000 * 60;
+
+class MockRegistryContracy implements RegistryContract {
```
dorothy-zbornak:
```suggestion
class MockRegistryContract implements RegistryContract {
```

packages/asset-swapper/src/quote_consumers/plp_registry.ts
```diff
@@ -0,0 +1,77 @@
+import { ContractFunctionObj } from '@0x/contract-wrappers';
```
dekz: This should live in `utils` rather than `quote_consumers`

packages/asset-swapper/src/quote_consumers/plp_registry.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export class PLPRegistry {
+    private readonly _contract: RegistryContract;
+    private readonly _cache: { [key: string]: CacheValue };
+
+    /**
+     * Instantiates a new PLPRegistry.
+     * @param contract an instance of IPLPRouterContract
+     */
+    constructor(contract: RegistryContract) {
```
dekz: Any thoughts on making this more generic so it can adapt any on-chain call to be cached for a time period?

Similar concept with protocol fee multiplier (150,000) where we had discussed checking this every so often.

packages/asset-swapper/src/quote_consumers/plp_registry.ts
PirosB3: I moved the file to `utils`. Thank you for the suggestion

packages/asset-swapper/test/plp_registry_test.ts
PirosB3: This was fixed, thank you

packages/asset-swapper/src/quote_consumers/plp_registry.ts
PirosB3: Here is what I came up with: https://gist.github.com/PirosB3/74cb65c65f110927c773e98f6150a7ba

The first file generic_cache_wrapper.ts contains the generic “abstract” classes. The second file plp_registry_cache_wrapper.ts contains an example implementation of the cache wrapper for the PLP Registry. I also added a third file sample_execution.ts which shows how you would “call” the implemented PLP Registry cache wrapper

NOTE: I don’t know TypeScript too well, so some things may be obviously wrong

PR #2493, opened 2020-02-24
Redeploy DevUtils

dorothy-zbornak:
## Description

Redeploys `DevUtils` on all networks with all sources verified.

## Checklist:


packages/contract-addresses/addresses.json
```diff
@@ -107,7 +107,7 @@
         "multiAssetProxy": "0xf6313a772c222f51c28f2304c0703b8cf5428fd8",
         "staticCallProxy": "0x48e94bdb9033640d45ea7c721e25f380f8bffa43",
         "erc1155Proxy": "0x64517fa2b480ba3678a2a3c0cf08ef7fd4fad36f",
-        "devUtils": "0xb1a3d901bad1df7d710fc8d008db7cdd6bbbffe6",
+        "devUtils": "0x9402639a828bdf4e9e4103ac3b69e1a6e522eb59",
```
dorothy-zbornak: You aren't seeing double. I had the same nonce on both of these networks.

jalextowle: I did that for my last few deployments too 😂

contracts/utils/test/artifacts.ts
```diff
@@ -6,6 +6,7 @@
 import { ContractArtifact } from 'ethereum-types';
 
 import * as Authorizable from '../test/generated-artifacts/Authorizable.json';
+import * as D18 from '../test/generated-artifacts/D18.json';
```
jalextowle: D18? I must be missing some context here

merklejerk: Just yet another fixed point library to jive with dydx.

jalextowle: Fun 🎉

PR #2502, opened 2020-02-27
Install gitpkg in asset-swapper

fragosti:
Works pretty well?

Screen Shot 2020-02-27 at 1 50 56 PM

It uploads the package as built to https://github.com/0xProject/gitpkg-registry/releases and uses the version in `package.json` as the tag. We need to establish some conventions around how to do this but seems useful?


fragosti:
@dorothy-zbornak I changed this to add the git SHA through a fork. That should serve us for now.

PR #2497, opened 2020-02-25
Update resolver errors to include parent filename

HenryNguyen5:
## Description

Add parent filename when resolving on the parents dependencies fails. This makes it so that it's easy to figure out which file has invalid import when multiple contracts has the same failing imports. Commonly happens during sweeping refactors.

## Testing instructions

Run compiler on a smart contract that has an incorrect import path for a dependency, inspect that the error message now reports the smart contract file within the error description.

## Types of changes

- Add parent filename when resolving on the parents dependencies fails.

I didnt modify the changelog since I dont know if 0x wants to include extra features in the next release.

PR #2504, opened 2020-02-29
Update resolver errors to include parent filename

jalextowle:
Duplicate of: https://github.com/0xProject/0x-monorepo/pull/2497. Moved to this PR so that a small change could be made.


jalextowle:
@HenryNguyen5 Would you mind giving this change a quick review? Btw, I had to move the PR since I didn't have access to the Chainlink fork.

PR #2508, opened 2020-03-02
Add JSON RPC ID namespace options  to web3wrapper + other packages

dave4506:
## Description

To help 0x-API (and other future products made with 0x tooling) to originate JSON rpc requests for debugging and alerting, added a `jsonRPCIdNameSpace` option to 0x tooling. 

Packages that have been updated: 
- `@0x/web-wrapper`
- `@0x/base-contract`
- `@0x/abi-gen`
- `@0x/contract-wrappers`
- `@0x/asset-swapper` 

By passing a `jsonRPCIdNameSpace`, all JSON RPC requests' `id` key to a node will be prepended with the namespace with a `.` delimiter.

Ex:
```
0x-asset-swapper.1
```

## Checklist:

### DIFF
--- a/packages/abi-gen/templates/TypeScript/contract.handlebars
+++ b/packages/abi-gen/templates/TypeScript/contract.handlebars
@@ -392,8 +392,9 @@ export class {{contractName}}Contract extends BaseContract {
         txDefaults?: Partial<TxData>,
         logDecodeDependencies?: { [contractName: string]: ContractAbi },
         deployedBytecode: string | undefined = {{contractName}}Contract.deployedBytecode,
+        jsonRpcIdNameSpace?: string,
     ) {
-        super('{{contractName}}', {{contractName}}Contract.ABI(), address, supportedProvider, txDefaults, logDecodeDependencies, deployedBytecode);
+        super('{{contractName}}', {{contractName}}Contract.ABI(), address, supportedProvider, txDefaults, logDecodeDependencies, deployedBytecode, jsonRpcIdNameSpace);
         classUtils.bindAll(this, ['_abiEncoderByFunctionSignature', 'address', '_web3Wrapper']);
         {{#if events~}}
         this._subscriptionManager = new SubscriptionManager<{{contractName}}EventArgs, {{contractName}}Events>(
--- a/packages/web3-wrapper/src/web3_wrapper.ts
+++ b/packages/web3-wrapper/src/web3_wrapper.ts
@@ -56,6 +56,7 @@ export class Web3Wrapper {
     // Raw provider passed in. Do not use. Only here to return the unmodified provider passed in via `getProvider()`
     private readonly _supportedProvider: SupportedProvider;
     private readonly _callAndTxnDefaults: Partial<CallData> | undefined;
+    private _jsonRpcRequestNameSpace: string;
     private _jsonRpcRequestId: number;
     /**
      * Check if an address is a valid Ethereum address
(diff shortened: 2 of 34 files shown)


packages/abi-gen/templates/TypeScript/contract.handlebars
```diff
@@ -392,8 +392,9 @@ export class {{contractName}}Contract extends BaseContract {
         txDefaults?: Partial<TxData>,
         logDecodeDependencies?: { [contractName: string]: ContractAbi },
         deployedBytecode: string | undefined = {{contractName}}Contract.deployedBytecode,
+        jsonRpcIdNameSpace?: string,
     ) {
-        super('{{contractName}}', {{contractName}}Contract.ABI(), address, supportedProvider, txDefaults, logDecodeDependencies, deployedBytecode);
+        super('{{contractName}}', {{contractName}}Contract.ABI(), address, supportedProvider, txDefaults, logDecodeDependencies, deployedBytecode, jsonRpcIdNameSpace);
```
xianny: can we provide a default value based on the contract name, so that this change doesn't break all existing instantiations?

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -679,7 +685,8 @@ export class Web3Wrapper {
     public async sendRawPayloadAsync<A>(payload: Partial<JSONRPCRequestPayload>): Promise<A> {
         const sendAsync = this._provider.sendAsync.bind(this._provider);
         const payloadWithDefaults = {
-            id: this._jsonRpcRequestId++,
+            id: `${!!this._jsonRpcRequestNameSpace ? `${this._jsonRpcRequestNameSpace}.` : ''}${this
```
xianny: nit: we could make this a little more readable:

```
const prefix = this._jsonRpcRequestNameSpace ? `${this._jsonRpcRequestNameSpace}.` : ''

...
id: `${prefix}${this._jsonRpcRequestId++}`
```

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -149,12 +150,17 @@ export class Web3Wrapper {
...
+        supportedProvider: SupportedProvider,
+        callAndTxnDefaults: Partial<CallData> = {},
+        jsonRpcIdNameSpace?: string,
+    ) {
         this.abiDecoder = new AbiDecoder([]);
         this._supportedProvider = supportedProvider;
         this._provider = providerUtils.standardizeOrThrow(supportedProvider);
         this._callAndTxnDefaults = callAndTxnDefaults;
         this._jsonRpcRequestId = 1;
+        this._jsonRpcRequestNameSpace = jsonRpcIdNameSpace;
```
xianny: nit: if there's no reason to change the name I think it's better to keep it consistent as `jsonRpcIdNameSpace` throughout

dave4506:
> Is this just for Alchemy? Is it possible to implement this with a Subprovider and using a provider engine? https://0x.org/docs/tools/subproviders/v6.0.8

That is possible to do so, just at a loss of granularity; as I take it, providers are instantiated usually once and used through out an app. With that, one provider would only allow one namespace to exist for all its uses in Web3Wrappers and in ContractWrappers. 

Possibly, we can use some kind of `NamespaceProvider` middleware object instead of the additional option. Every time we pass provider into an object we instantiate `NamespaceProvider` with the provider and the namespace to be used in a specific wrapper. 

Open to thoughts to both ways @fragosti

fragosti:
I'm not sure this solves the issue.
Copying Oskar's response:
```
So, I don't think this solves the ASANA issue.
It does add a namespace, but from the 0x-API/application point of view, our requests are already namespaced because of Alchemy Apps and Keys.
My issue that's on the ASANA Task was basically for the 0x-API to log the requestID of the request that failed. The ideal flow would be:
1. at T1 /swap/quote failed with Network time out on alchemy.
2. Throws an error eth_call (requestID: 1234) failed with network timeout
3. I grab the logs, correlate requestID: 1234 with our alchemy dashboard around time T1 and dive deeper, and maybe ask alchemy if it's something on their end with those requests.
Although I see the value of segregating the namespace for internal libraries that utilize the Eth RPC, but in our case we definitely know what the big bois (eth_call) are.
For the task at hand, I think that requestID: 0x-asset-swapper.1234 is not a bigger help than requestID: 1234.
Ideally the requestIDs would be UUIDs, but a semi-random / unique integer at a known time also works.
```

opaolini:
Sorry about not responding directly, the above response sums up my thoughts, but to add to that:

The namespace here will be definitely useful if on start of `0x-API` we would be able to specify the namespace like `69b5bd988b` (which is an example 0x-API instance hash), then we could easily attribute that to the specific instance of `0x-API`.

Now we just need to propagate the error such that we would get:
```
network timeout at: https://<provider>. (request.id: 69b5bd988b.1337)
```

Now we just see:
```
network timeout at: https://<provider>.
```

PR #2511, opened 2020-03-04
`@0x/contracts-extensions`: Maximum gas price contract

mzhu25:
## Description
Contract includes a parameter-less version which uses a hard-coded max of 20 Gwei, and a version that allows makers to specify a custom max value. 
Includes tooling for encoding/decoding the StaticCall asset data, as well as some basic unit tests

## Checklist:


contracts/extensions/contracts/src/MaximumGasPrice.sol
```diff
@@ -0,0 +1,52 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.16;
```
jalextowle: 👀

contracts/extensions/contracts/src/MaximumGasPrice.sol
```diff
@@ -0,0 +1,52 @@
...
+        require(
+            tx.gasprice <= DEFAULT_MAX_GAS_PRICE,
+            "MaximumGasPrice/GAS_PRICE_EXCEEDS_20_GWEI"
+        );
+    }
+
+    /// @dev Checks that the current transaction's gas price is less than
+    ///      the specified maximum value.
+    /// @param data Encodes the maximum gas price.
+    function checkGasPrice(bytes calldata data)
```
abandeali1: Why not just take a `uint256` here?

contracts/extensions/contracts/src/MaximumGasPrice.sol
```diff
@@ -0,0 +1,52 @@
...
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.5.16;
+pragma experimental ABIEncoderV2;
```
abandeali1: I don't think we actually need to use ABIv2 here. I'd just run tests with/without to see if there are any gas differences.

contracts/extensions/src/max_gas_price_utils.ts
```diff
@@ -0,0 +1,44 @@
...
+const defaultGasPriceEncoder = AbiEncoder.createMethod('checkGasPrice', []);
+
+const ONE_GWEI = new BigNumber(10 ** 9);
+export const TWENTY_GWEI = ONE_GWEI.times(20);
+
+/**
+ * Encodes the given stop limit data parameters into StaticCall asset data so that it can be used
+ * in a 0x order.
+ */
+export function encodeMaxGasPriceStaticCallData(maxGasPriceContractAddress: string, maxGasPrice?: BigNumber): string {
```
steveklebanoff: Thank you for writing this helper 🙌

contracts/extensions/contracts/src/MaximumGasPrice.sol
mzhu25: didn't realize that would work 🙈

contracts/extensions/contracts/src/MaximumGasPrice.sol
mzhu25: looks like without is _slightly_ cheaper for the parametrized version

contracts/extensions/contracts/src/MaximumGasPrice.sol
```diff
@@ -38,12 +36,10 @@ contract MaximumGasPrice {
 
     /// @dev Checks that the current transaction's gas price is less than
     ///      the specified maximum value.
-    /// @param data Encodes the maximum gas price.
-    function checkGasPrice(bytes calldata data)
+    /// @param maxGasPrice The maximum gas price allowed for the current transaction.
+    function checkGasPrice(uint256 maxGasPrice)
         external
-        view
```
abandeali1: Why did we get rid of `view` here?

mzhu25: oops that was a remnant of some tests

PR #2512, opened 2020-03-05
Bridges: USDT/KNC fix + ERC20BridgeTransfer events

dorothy-zbornak:
### Use `LibERC20Token.approveIfBelowMax()`
USDT (and apparently KNC) reverts in `approve()` if the allowance is not zero. Because we re-approve tokens on every fill in the DEX bridges, this means bridges only work the first time for USDT. Now the bridges will use `LibERC20Token.approveIfBelow()`, which only calls `approve(-1)` if the allowance is less than some amount.

This does mean we will most certainly never re-up the allowance on legitimate tokens. This should be fine since, for example, you would have to transfer the entire supply of KNC 10^51 times to exhaust the initial allowance.

### Emit `ERC20BridgeTransfer` events
It's difficult to accurately track fills from DEX bridges because they can transfer more than the order dictates, and fill results will not catch this. To help address this, bridges now emit an `ERC20BridgeTransfer` event:

```solidity
    /// @dev Emitted when a bridge transfer is completed.
    /// @param bridge The bridge address.
    /// @param fromToken The address of the "from" token.
    /// @param toToken The address of the "to" token.
    /// @param fromTokenAmount The "from" token amount consumed.
    /// @param toTokenAmount The "to" token amount transferred.
    /// @param from Supplier of "fromToken".
    /// @param to Receiver of "toToken".
    event ERC20BridgeTransfer(
        address indexed bridge,
        address fromToken,
        address toToken,
        uint256 fromTokenAmount,
        uint256 toTokenAmount,
        address from,
        address to
    );
```

Outside of DEX bridges, not all fields are guaranteed or have the same meaning. For example, neither Dydx nor Chai necessarily convert from a taker token. They also aren't privy to a  "from" token balance. It's a bit of an odd fit, but I figure some data is better than no data.

### TODO
- Redeploy after everyone agrees this PR is copacetic.

## Checklist:


contracts/asset-proxy/contracts/src/bridges/CurveBridge.sol
```diff
@@ -55,32 +63,32 @@ contract CurveBridge is
...
     {
         // Decode the bridge data to get the Curve metadata.
-        (address curveAddress, int128 fromCoinIdx, int128 toCoinIdx, int128 version) = abi.decode(bridgeData, (address, int128, int128, int128));
-        ICurve exchange = ICurve(curveAddress);
+        CurveBridgeData memory data = abi.decode(bridgeData, (CurveBridgeData));
 
-        address fromTokenAddress = exchange.underlying_coins(fromCoinIdx);
+        address fromTokenAddress = ICurve(data.curveAddress).underlying_coins(data.fromCoinIdx);
         require(toTokenAddress != fromTokenAddress, "CurveBridge/INVALID_PAIR");
+        uint256 fromTokenBalance = IERC20Token(fromTokenAddress).balanceOf(address(this));
```
abandeali1: Any reason not to use `LibERC20Token` here?

contracts/asset-proxy/contracts/src/bridges/CurveBridge.sol
```diff
@@ -89,6 +97,16 @@ contract CurveBridge is
         uint256 toTokenBalance = IERC20Token(toTokenAddress).balanceOf(address(this));
```
abandeali1: Same as above - should probably use `LibERC20Token`.

contracts/asset-proxy/contracts/src/bridges/Eth2DaiBridge.sol
```diff
@@ -56,18 +57,29 @@ contract Eth2DaiBridge is
         (address fromTokenAddress) = abi.decode(bridgeData, (address));
 
         IEth2Dai exchange = IEth2Dai(_getEth2DaiAddress());
+        uint256 fromTokenBalance = IERC20Token(fromTokenAddress).balanceOf(address(this));
```
abandeali1: `LibERC20Token`

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
```diff
@@ -81,6 +82,18 @@ contract DydxBridge is
 
         // Run operation. This will revert on failure.
         IDydx(_getDydxAddress()).operate(accounts, actions);
+
+        emit ERC20BridgeTransfer(
```
abandeali1: I'm wondering if we should even have these events for the non-DEX bridges. It's pretty awkward with both this and `ChaiBridge`, and I'm not sure it really makes data collection any easier without some sort of rate.

contracts/asset-proxy/contracts/src/bridges/ChaiBridge.sol
```diff
@@ -70,6 +70,16 @@ contract ChaiBridge is
         // This will never fail if the `draw` call was successful
         IERC20Token(_getDaiAddress()).transfer(to, amount);
 
+        emit ERC20BridgeTransfer(
```
abandeali1: As with the `DydxBridge`, I'd consider just not emitting an event here.

contracts/erc20/contracts/src/LibERC20Token.sol
```diff
@@ -47,6 +47,25 @@ library LibERC20Token {
...
+    /// @param spender The address that receives an allowance.
+    /// @param amount The minimum allowance needed.
+    function approveIfBelow(
+        address token,
+        address spender,
+        uint256 amount
+    )
+        internal
+    {
+        if (IERC20Token(token).allowance(address(this), spender) < amount) {
```
abandeali1: We can just use the `allowance` built into this library here.

contracts/asset-proxy/contracts/src/bridges/CurveBridge.sol
dorothy-zbornak: `LibERC20Token` is too permissive with `balanceOf/allowance` (won't propagate reverts). Those were created basically for DevUtils' sake.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
dorothy-zbornak: I'll check with kroeger.

alexkroeger: do we have another viable option for collecting the precise trade amounts for those bridges?

contracts/asset-proxy/contracts/src/interfaces/IERC20Bridge.sol
```diff
@@ -21,7 +21,25 @@ pragma solidity ^0.5.9;
...
+    /// @dev Emitted when a bridge transfer is completed.
+    /// @param bridge The bridge address.
+    /// @param fromToken The address of the "from" token.
+    /// @param toToken The address of the "to" token.
+    /// @param fromTokenAmount The "from" token amount consumed.
+    /// @param toTokenAmount The "to" token amount transferred.
+    /// @param from Supplier of "fromToken".
+    /// @param to Receiver of "toToken".
+    event ERC20BridgeTransfer(
+        address indexed bridge,
```
abandeali1: Do we actually need to log the address of each bridge? Seems uncommon for events to do this.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
dorothy-zbornak: In practicality, Chai and Dydx orders will have their `toAmount`s match up with those in the `Fill` event. All the other fields will also be visible in the `Fill` event. So I'm leaning towards removing them.

dorothy-zbornak: I guess the argument for keeping it would be to make tracking (successful) fills from these bridges easier since you won't have to parse the asset data. But I think we already have the infra for that?

contracts/asset-proxy/contracts/src/interfaces/IERC20Bridge.sol
dorothy-zbornak: Oh, no, you're right. I forgot receipts will include the logger address.

contracts/asset-proxy/contracts/src/bridges/DydxBridge.sol
alexkroeger: yea I'm not worried about that, we have infra for parsing asset data in all fills. If fill events will accurately reflect traded amounts in Chai and Dydx bridge trades then I'm fine with removing.

PR #2513, opened 2020-03-06
Asset-swapper: Fallback orders + refactors

dorothy-zbornak:
### Generate fallback orders
If the `allowFallback` option is passed into `SwapQuoter`, fallback orders can be generated. How this works is we first generate an optimized quote. Then if this quote includes native orders, we will generate a second quote for the same size as the native orders and append it to the original quote. This fallback quote will not reuse any sources in the original quote.

We *do*, however, allow Kyber conflicts in the fallback quote because, even though in theory this should reduce reverts, in reality it increases reverts because with conflicting sources omitted we very often cannot find an fallback path. Even with conflicts, early simulations (w/ bridgeSliipage = 1%) never seem to fail on these quotes, so I think we good?

### Gas estimates
Because `eth_estimateGas` will always fail to correctly anticipate the gas cost of fallback orders, AS now attempts to provide a `gas` estimate in the `bestCaseQuoteInfo` and `worstCaseQuoteInfo` fields. I'm lazy so they're both the same right now but in the future we can get more clever with the best case.
 
### New(ish) optimizer algorithm
I tweaked the algorithm and cleaned it up a lot. We still try to explore as many fill combinations as we can (we have to because fees make price curves non-convex). The algo should also run a bit faster now thanks to Remco's suggestion of iteratively solving two paths at a time, rather than all at once. This is important because we now essentially run it twice in order to generate the fallback orders.

### Massive refactors to `market_operation_utils`
Yeah, that was turning into a hot mess. All the code in that directory are much nicer and simpler now.

### Changes to swap quote options
Removed:
- `slippagePercentage`: The fallback orders essentially double the possible fill amount already.
- `dustFractionThreshold`: No longer necessary because the optimizer can handle a larger # of paths.
- `noConflicts`: We never set this to false so it's now always on.

Added:
- `allowFallback`: If `true` (default), allows the generation of fallback orders
- `gasSchedule`: Estimated gas usage of filling a liquidity source-- similar to `feeSchedule` but expressed in gas instead of ETH.
- `maxFallbackSlippage`: The maximum rate % change allowed in the fallback (compared to the optimal quote). If the fallback's slippage exceeds this %, no fallback will be provided.

Renamed:
- `fees` -> `feeSchedule`

### Ordering of orders in quote
Because of how the new optimizer works, there is no longer a strong guarantee that orders will be sorted by decreasing rate. If you relied on this, sorry boo.

### Bugs
- I found a small bug related to how DEX paths were generated, which may have led to slightly worse pricing.
- Because the code is so different, I went ahead and addressed #2517 in here as well.

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -240,373 +198,114 @@ export class MarketOperationUtils {
...
-            if (prevFill.source === source) {
-                prevFill.totalMakerAssetAmount = prevFill.totalMakerAssetAmount.plus(makerAssetAmount);
-                prevFill.totalTakerAssetAmount = prevFill.totalTakerAssetAmount.plus(takerAssetAmount);
-                prevFill.subFills.push({ makerAssetAmount, takerAssetAmount });
-                continue;
-            }
+        // Find a fallback path from sources not used in the first path.
+        let fallbackPath: Fill[] = [];
+        if (opts.allowFallback) {
+            fallbackPath = findOptimalPath(side, getUnusedSourcePaths(optimalPath, paths), inputAmount) || [];
```
dekz: I believe this would pickup Kyber where Uniswap was previously in the Optimal Path. These would be exclusive as if the Uniswap path were to fail then the Kyber path would probably fail, at least at that price? 

Seems risky.

packages/asset-swapper/src/constants.ts
```diff
@@ -43,7 +42,7 @@ const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {
         orderRefreshIntervalMs: 10000, // 10 seconds
     },
     ...DEFAULT_ORDER_PRUNER_OPTS,
-    samplerGasLimit: 59e6,
+    samplerGasLimit: 250e6,
```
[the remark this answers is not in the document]
dorothy-zbornak: What's the benefit?

dorothy-zbornak:
> Taken a quick peek but wanted to discuss some things to get the clear in my mind.
> 
> I feel like there are 2 approaches to the goal of lowering reverts with a backup path.
> 
> 1. Find the best path using on-chain sources, use 0x to improve price if available. So we have a backup to 0x failing.
> 2. Find the best path, create another backup path incase the best path fails. This feels a bit difficult to reason about. It could be 100% Uniswap, then 100% 0x as a backup (previously this would be just Uniswap, so little to be gained in this scenario). It could also be 100% Uniswap then 100% Kyber as a backup, which could be a conflict. It can also be the inverse of 100% Kyber with 100% Uniswap backup, also a potential conflict.
> 
> I am biased to the 1) approach as that seems simple, but keen to hear your thoughts. In my head 0x should never be the backup as I find it difficult to reason the case where Uniswap would fail and 0x orders would still be fillable.
> 
> When a path includes Curve, and a backup path includes another Curve, this is now 2million+ faux gas estimate as a "just in case".

What do you think of:
1. Find the optimal quote.
2. If the optimal quote contains a 0x order:
  1. Generate a fallback quote with the unused sources (+ factoring in Kyber conflicts).

The separate quotes thing is to ensure the optimizer is solving for as close to the actual fill amount as possible, which gives us better pricing.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -240,373 +194,120 @@ export class MarketOperationUtils {
...
+                findOptimalPath(side, getFallbackSourcePaths(optimalPath, paths), fallbackInputAmount, opts.runLimit) ||
+                [];
         }
-        collapsed.push({
-            source: fill.fillData.source,
-            totalMakerAssetAmount: makerAssetAmount,
-            totalTakerAssetAmount: takerAssetAmount,
-            subFills: [{ makerAssetAmount, takerAssetAmount }],
-            nativeOrder: (fill.fillData as NativeFillData).order,
+        return createOrdersFromPath([...optimalPath, ...fallbackPath], {
```
dekz: Still of a mind to have protection here where it's not worth filling this backup path if the price differs from the best path by X amount (aka `slippagePercentage`?)

If that is the case I'd rather eat a revert then slip 10%. Though this is very difficult to simulate at the moment.

I'd be down to fail silently and handle the case if there is no decent fallback path, we tried our best. So the user only gets the optimal path.

dorothy-zbornak: Added a `maxSlippage` option, but it only applies to the fallback, (reasons discussed offline).

packages/instant/src/util/swap_quote_updater.ts
```diff
@@ -37,18 +36,13 @@ export const swapQuoteUpdater = {
...
-            asset.metaData.assetProxyId === AssetProxyId.ERC20
-                ? ERC20_SWAP_QUOTE_SLIPPAGE_PERCENTAGE
-                : ERC721_SWAP_QUOTE_SLIPPAGE_PERCENTAGE;
         try {
             const gasInfo = await gasPriceEstimator.getGasInfoAsync();
             newSwapQuote = await swapQuoter.getMarketBuySwapQuoteForAssetDataAsync(
                 asset.assetData,
                 wethAssetData,
                 baseUnitValue,
                 {
```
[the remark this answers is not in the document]
dorothy-zbornak: Where do you still see `slippagePercentage`?

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -176,7 +167,16 @@ export interface GetMarketOrdersOpts {
     /**
      * Fees for each liquidity source, expressed in gas.
      */
-    fees: { [source: string]: BigNumber };
+    feeSchedule: { [source: string]: BigNumber };
+    /**
+     * Estimated gas consumed by each liquidity source.
+     */
+    gasSchedule: { [source: string]: number };
```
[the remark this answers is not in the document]
dorothy-zbornak: I could go either way. It's *probably* not good design for AS to assume the protocol fee, even though it already does in a few places (we should remove these).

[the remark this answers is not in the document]
dorothy-zbornak: It's also not particularly hard for the caller to do the operation you described theirself.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -191,6 +191,9 @@ export const samplerOperations = {
         takerFillAmount: BigNumber,
         liquidityProviderRegistryAddress?: string | undefined,
     ): BatchedOperation<BigNumber> {
+        if (makerToken.toLowerCase() === takerToken.toLowerCase()) {
+            return samplerOperations.constant(new BigNumber(1));
```
dekz: nice

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -0,0 +1,305 @@
...
+            if (sample.output.eq(0)) {
+                break;
+            }
+            const input = sample.input.minus(prevSample ? prevSample.input : 0);
+            const output = sample.output.minus(prevSample ? prevSample.output : 0);
+            const penalty =
+                i === 0 // Only the first fill in a DEX path incurs a penalty.
+                    ? ethToOutputRate.times(fees[source] || 0)
+                    : ZERO_AMOUNT;
+            const adjustedOutput = side === MarketOperation.Sell ? output.minus(penalty) : output.plus(penalty);
```
dekz: We may want to `break` if adjustedOutput is `lte(0)`, I guess optimizer would kick it out if there were other sources available.

dorothy-zbornak: I considered this at first but realized it could drop entire paths because the fee in the first fill can drive it to negative (esp if we increase samples). This might be viable for native paths, though.

dorothy-zbornak: Oh we already do that for native paths.

dekz: ah I see, that makes sense, especially for small fills and 20 samples.

PR #2521, opened 2020-03-12
Forwarder Market sell specified amount or throw

dekz:
## Description

We wish to sell a specific amount or throw. This is then consistent with `marketBuyOrdersWithEth` and our usage of `FillOrKill` functions in asset-swapper. 

`marketSellOrdersWithEth` does not throw and is a surprise to many users and inconsistent when selling DAI versus selling ETH via asset-swapper/0xApi. Generally we view fills as having funds for the `takerAssetAmount` and an additional source of funds for the protocolFee. Asset-swapper is no longer guaranteeing that the orders are always sorted by adjusted price (due to redundant path), so this opens up a potential issue where we partially fill some "ok" orders. 

One thing to note is how `ethSellAmount` interacts with protocol fees. We wish to sell 50 ETH for example and not 49.9 ETH and 0.1 ETH of protocol fee. In normal conditions these are distinct (WETH and ETH in `value`, WETH and WETH (allowance)). This function also treats them as seperate pools, (`ethSellAmount` ETH and  protocol fee ETH also in `value`). To sell 50 ETH using this function the user will need to specify `ethSellAmount=50` and send `value=50.1` to sell and pay protocol fee.

In the event of a successful marketSell which encountered unfillable orders, the additional protocolFees are returned, rather than sold in following orders.

## TODO
- [x] clean up as there is a bit of duplicate functionality
- [ ] use the struct everywhere for consistency?


contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -223,24 +223,96 @@ contract MixinExchangeWrapper {
...
+        uint256 wethSellAmount,
+        bytes[] memory signatures
+    )
+        internal
+        returns (
+            uint256 totalWethSpentAmount,
+            uint256 totalMakerAssetAcquiredAmount
+        )
+    {
+        bytes4 erc20BridgeProxyId = IAssetData(address(0)).ERC20Bridge.selector;
```
dorothy-zbornak: I don't think there's any benefit to putting this in a variable. idk why we do it in the other function either.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -223,24 +223,96 @@ contract MixinExchangeWrapper {
...
+            );
+
+            // Account for the ERC20Bridge transfering more of the maker asset than expected.
+            if (makerAssetProxyId == erc20BridgeProxyId) {
+                uint256 balanceAfter = IERC20Token(tokenAddress).balanceOf(address(this));
+                sellFillResults.makerAssetAcquiredAmount = LibSafeMath.max256(
+                    balanceAfter.safeSub(balanceBefore),
+                    sellFillResults.makerAssetAcquiredAmount
+                );
+            }
```
dorothy-zbornak: This all looks identical to what we do in `_marketSellNoThrow()`. Maybe we should just bake this logic into `_marketSellSingleOrder()`.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -223,24 +223,96 @@ contract MixinExchangeWrapper {
...
+
+        for (uint256 i = 0; i != orders.length; i++) {
+            // Preemptively skip to avoid division by zero in _marketSellSingleOrder
+            if (orders[i].makerAssetAmount == 0 || orders[i].takerAssetAmount == 0) {
+                continue;
+            }
+
+            // The remaining amount of WETH to sell
+            uint256 remainingTakerAssetFillAmount = wethSellAmount
+                .safeSub(totalWethSpentAmount);
```
dorothy-zbornak: I think we also want to `safeSub(totalProtocolFeePaid)`?

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
```diff
@@ -223,24 +223,96 @@ contract MixinExchangeWrapper {
...
+
```
dorothy-zbornak: What if we also did something like this here:

```solidity
if (remainingTakerAssetFillAmount > singleProtocolFee) {
    // Do not count the protocol fee as part of the fill amount.
    remainingTakerAssetFillAmount = remainingTakerAssetFillAmount.safeSub(singleProtocolFee);
} else {
    // Stop if we don't have at least enough ETH to pay another protocol fee.
    break;
}
```

dorothy-zbornak: My hunch is line 206 (`.safeSub(_isV2Order(orders[i]) ? 0 : protocolFee);`)  in `_marketSellNoThrow()` is probably where I was seeing those rigo + USDC + high gas reverts, so adding something similar to that function could also fix that.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
mzhu25: i kinda like it for readability, but we can put it in a constant

contracts/integrations/test/forwarder/forwarder_test.ts
```diff
@@ -157,7 +157,7 @@ blockchainTests('Forwarder integration tests', env => {
             );
         });
     });
-    blockchainTests.resets('marketSellOrdersWithEth without extra fees', () => {
+    blockchainTests.resets.only('marketSellOrdersWithEth without extra fees', () => {
```
mzhu25: don't forget to remove the `.only` 🐘

contracts/integrations/test/forwarder/forwarder_test.ts
```diff
@@ -448,6 +448,25 @@ blockchainTests('Forwarder integration tests', env => {
             });
         });
     });
+    blockchainTests.resets('marketSellAmountWithEth', () => {
+        const protocolFee = new BigNumber(150000).times(DeploymentManager.gasPrice);
```
mzhu25: nit: we have `DeploymentManager.protocolFee`

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
dekz: Perhaps the naming is off right now, but `wethSellAmount` in the parameter is exclusive of protocol fees. So we don't want to subtract the protocolFee as it is not included in `wethSellAmount`. I.e A user calls `marketSellAmountWithEth` with 2 ETH and provides 2.1 ETH in value. We then proceed to sell 2ETH, not 2.1 ETH.

contracts/exchange-forwarder/contracts/src/MixinExchangeWrapper.sol
dekz: Perhaps the naming is off right now or I am missing something, but `wethSellAmount` in the parameter is exclusive of protocol fees. So we don't want to subtract the protocolFee as it is not included in `wethSellAmount`. I.e A user calls `marketSellAmountWithEth` with 2 ETH and provides 2.1 ETH in value. We then proceed to sell 2ETH, not 2.1 ETH.

dekz: Perhaps the naming is off right now or I am missing something, but `wethSellAmount` in the parameter is exclusive of protocol fees. So we don't want to subtract the protocolFee as it is not included in `wethSellAmount`. I.e A user calls `marketSellAmountWithEth` with 2 ETH and provides 2.1 ETH in value. We then proceed to sell 2ETH, not 2.1 ETH.

contracts/exchange-forwarder/contracts/src/Forwarder.sol
```diff
@@ -151,6 +151,72 @@ contract Forwarder is
...
+        ) = _marketSellExactAmountNoThrow(
+            orders,
+            ethSellAmount,
+            signatures
+        );
+        // Ensure we sold the specified amount (note: wethSpentAmount includes fees)
+        if (wethSpentAmount < ethSellAmount) {
+            LibRichErrors.rrevert(LibForwarderRichErrors.CompleteSellFailedError(
+                ethSellAmount,
+                wethSpentAmount
```
dorothy-zbornak: Now I'm wondering if it would be more useful to know how much we actually sold minus the protocol fee. What do you think?

PR #2523, opened 2020-03-13
Gas Tokens in Bridges

dekz:
## Description

When a user makes a trade against a Bridge like Curve (10e5 gas). They spend approximately `(10e5 * 8e9) * 1e-18 * $150` $1.20 in gas. We can reduce this by approximately 40% by freeing gas tokens if we have an available balance. So the user only pays $0.72 to fill on Curve.  

Calculations are from gastoken.io
```javascript
// GST2
var FREE_BASE = 14154;
var FREE_TOKEN = 6870;
var REIMBURSE = 24000;
Math.floor((gasUsed + FREE_BASE) / (2 * REIMBURSE - FREE_TOKEN));
```

| bridge  | gas consumed | GST2 freed |
| ------- | ------------ | ---------- |
| Uniswap | 3e5          | 0.03       |
| Oasis   | 5e5          | 0.12       |
| Curve   | 10e5          | 0.24       |

### freeFrom
Rather than have each bridge use its own balance, we chose to use `freeFromUpTo` to allow a EOA (GST Collector) to hold the GST balance, preventing a drain if GST2 was traded via a Bridge. This keeps the balances of tokens traded distinct from the gst balance. The collector can whitelist certain contracts, and remove contracts if required. If there is no allowance set it does not revert. 

We also don't loose any unspent (and sent) tokens when we re-deploy bridges. 

### Process 
* GST Collector holds the balance of GST2. 
* User deploys new Bridge Contract specifying the GST Collector as the address to free from
* Until Collector approves the Bridge Contract, 0 GST will be freed
* Collector approves Bridge Contract

### Notes
* This does not refund on a revert
* This does not cover 0x overhead (only bridge overhead)


contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
```diff
@@ -0,0 +1,41 @@
...
+contract MixinGasToken is
+    DeploymentConstants
+ {
+
+    /// @dev Burns gas tokens based on the amount of gas consumed in the function
+    modifier burnsGasTokens {
+        uint256 gasBefore = gasleft();
+        _;
+        IGasToken gst = IGasToken(_getGstAddress());
+        if (address(gst) != address(0) && msg.sender == _getERC20BridgeProxyAddress()) {
```
abandeali1: Do we actually need any of these checks? `address(gst) == address(0)` will probably only be true on testnets, and `msg.sender == _getERC20BridgeProxyAddress()` is usually enforced through different checks. This modifier may end up being useful elsewhere without these.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
```diff
@@ -0,0 +1,41 @@
...
+            // (gasUsed + FREE_BASE) / (2 * REIMBURSE - FREE_TOKEN)
+            //            14154             2400        6870
+            uint256 value = (gasBefore - gasleft() + 14154) / 41130;
+            gst.freeUpTo(value);
```
abandeali1: What happens if this contract has an insufficient balance of gas tokens?

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
abandeali1: Ah nvm, I guess we don't check if `msg.sender == _getERC20BridgeProxyAddress()` on all of the bridges.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
```diff
@@ -0,0 +1,41 @@
...
+ {
+
+    /// @dev Burns gas tokens based on the amount of gas consumed in the function
+    modifier burnsGasTokens {
+        uint256 gasBefore = gasleft();
+        _;
+        IGasToken gst = IGasToken(_getGstAddress());
+        if (address(gst) != address(0) && msg.sender == _getERC20BridgeProxyAddress()) {
+            // (gasUsed + FREE_BASE) / (2 * REIMBURSE - FREE_TOKEN)
+            //            14154             2400        6870
```
hysz: Nitty - `24000`

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
dekz: Do we want to these bridges (Uni,Oasis,Curve,Kyber) `bridgeTransferFrom` only callable via BridgeProxy address?

Happy to create a new modifier if we can guarantee that assumption

dorothy-zbornak: I personally like these being permissionless.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
dekz: It's handled gracefully but wastes a bit of gas.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
dekz: I've removed these checks, if a bridge wants em it can add em.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /// @dev Frees gas tokens based on the amount of gas consumed in the function
+    modifier freesGasTokens {
+        uint256 gasBefore = gasleft();
+        _;
+        IGasToken gst = IGasToken(_getGstAddress());
+        if (address(gst) != address(0)) {
+            // (gasUsed + FREE_BASE) / (2 * REIMBURSE - FREE_TOKEN)
+            //            14154             24000        6870
+            uint256 value = (gasBefore - gasleft() + 14154) / 41130;
```
dorothy-zbornak: This tries to refund the entire amount of gas consumed, right? Maybe we should just try to refund half. I believe a quirk of the EVM is that refunds are capped at half the gas consumed, so any excess might be wasted.

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
```diff
@@ -0,0 +1,55 @@
...
+    /// @dev Frees gas tokens using the balance of `from`. Amount freed is based
+    ///     on the gas consumed in the function
+    modifier freesGasTokensFromCollector() {
+        uint256 gasBefore = gasleft();
+        _;
+        IGasToken gst = IGasToken(_getGstAddress());
+        if (address(gst) != address(0)) {
+            // (gasUsed + FREE_BASE) / (2 * REIMBURSE - FREE_TOKEN)
+            //            14154             24000        6870
+            uint256 value = (gasBefore - gasleft() + 14154) / 41130;
```
dorothy-zbornak: Ditto

contracts/asset-proxy/contracts/src/bridges/MixinGasToken.sol
dekz: Yes this calculation gets us to maximum efficiency. You can also play with the calculator (https://gastoken.io/) to confirm.

```
gasUsed = 2,000,000
value = 48 =  (gasUsed + 14154) / 41130 

SELFDESTRUCT = -24000
SELFDESTRUCT_COST = 700 + 5000
refund = -878,400 = SELFDESTRUCT * 48 + SELFDESTRUCT_COST * 48

newGasUsed = 1,121,600 > gasUsed * 0.5
```

PR #2525, opened 2020-03-18
DexForwarderBridge

dorothy-zbornak:
##  Description
We can reduce swap protocol fees by unifying bridge orders under a single fill. We can take advantage of the fact that DEX bridges are permissionless. This PR introduces the `DexForwarderBridge` bridge, which essentially performs market fills over a sequence of bridge calls.

## Checklist:


contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+        (
+            state.inputToken,
+            state.calls
+        ) = abi.decode(bridgeData, (address, BridgeCall[]));
+
+        state.initialInputTokenBalance = IERC20Token(state.inputToken).balanceOf(address(this));
+
+        for (uint256 i = 0; i < state.calls.length; ++i) {
+            // Stop if the we've sold all our input tokens.
+            if (state.totalInputTokenSold >= state.initialInputTokenBalance) {
```
dorothy-zbornak: I had a version where we also could stop based on output tokens bought (market buy), but ran into some issues:
- we can't use `amount` because we underestimate this value in AS w/ `bridgeSlippage`
- this contract can end up with a balance of input tokens, so would we refund those to `to`?
- I'm not sure this is the same behavior you would get if *not* using the DFB (individual bridge orders)

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+        external
+        returns (bytes4 success)
+    {
+        TransferFromState memory state;
+        (
+            state.inputToken,
+            state.calls
+        ) = abi.decode(bridgeData, (address, BridgeCall[]));
+
+        state.initialInputTokenBalance = IERC20Token(state.inputToken).balanceOf(address(this));
```
mzhu25: might as well use LibERC20Token for balances

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+
+    event DexForwarderBridgeCallFailed(
+        address indexed target,
+        address inputToken,
+        address outputToken,
+        uint256 inputTokenAmount,
+        uint256 outputTokenAmount
+    );
+
+    /// @dev Executes a series of calls, forwarding .
```
mzhu25: missing the latter half of this sentence

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+    /// @param outputToken The token being bought.
+    /// @param to The recipient of the bought tokens.
+    /// @param bridgeData The abi-encoeded input token address.
```
mzhu25:
```suggestion
    /// @param bridgeData The abi-encoded input token address.
```

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+                // Increase the amount of tokens sold.
+                state.totalInputTokenSold = state.totalInputTokenSold.safeAdd(
+                    state.callInputTokenAmount
+                );
+            }
+        }
+        // Revert if we were not able to sell our entire input token balance.
+        require(
+            state.totalInputTokenSold >= state.initialInputTokenBalance,
+            "DexForwaderBridge/INCOMPLETE_FILL"
```
mzhu25:
```suggestion
            "DexForwarderBridge/INCOMPLETE_FILL"
```
🇨🇦 and elsewhere too

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,210 @@
...
+                call.inputTokenAmount,
+                state.initialInputTokenBalance.safeSub(state.totalInputTokenSold)
+            );
+            state.callOutputTokenAmount = LibMath.getPartialAmountFloor(
+                state.callInputTokenAmount,
+                call.inputTokenAmount,
+                call.outputTokenAmount
+            );
+
+            (bool didSucceed, ) = address(this)
```
mzhu25: Maybe we should decode the return data here and check that it equals BRIDGE_SUCCESS?

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
mzhu25: yeah I think this is probably cleaner

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
dorothy-zbornak: `LibERC20Token.balanceOf()` swallows reverts. Not my jam.

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
dorothy-zbornak: `execteBridgeCall()` doesn't return anything. It just succeeds or reverts.

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
```diff
@@ -0,0 +1,213 @@
...
+                    state.inputToken,
+                    outputToken,
+                    state.callInputTokenAmount,
+                    state.callOutputTokenAmount,
+                    call.bridgeData
+                ));
+
+            if (!didSucceed) {
+                // Log errors.
+                emit DexForwarderBridgeCallFailed(
```
dekz: Is this just for internal data gathering/insights? 

AS makes heavy use of FillOrKill so the only case where it'd reasonably be emitted is when succeed with enough slippage/backup path?

contracts/asset-proxy/src/dex_forwarder_bridge.ts
```diff
@@ -0,0 +1,27 @@
...
+export interface DexForwaderBridgeCall {
+    target: string;
+    inputTokenAmount: BigNumber;
+    outputTokenAmount: BigNumber;
+    bridgeData: string;
+}
+
+export interface DexForwaderBridgeData {
+    inputToken: string;
+    calls: DexForwaderBridgeCall[];
```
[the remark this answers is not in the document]
dorothy-zbornak: help me jesus

contracts/asset-proxy/contracts/src/bridges/DexForwarderBridge.sol
dorothy-zbornak: Yeah, that's what I'm thinking. These won't be going through `fillOrder()` or the ERC20BridgeProxy so it's a less obvious path to detecting them. Maybe it's not worth it, though, since the bridge itself can revert if not enough input tokens are sold. We'd have to rely on traces for those situations (and the market fn failing).

Now I'm leaning towards nixing this event.

PR #2526, opened 2020-03-20
AssetSwapper: Fix quote optimizer fee bug

dorothy-zbornak:
## Description

When the optimizer was building a quote, it wasn't clipping the input/output amounts of a tail-end fill by the remaining input amount. So a large enough fill (usually native orders) could be attributed a higher return rate than it really should. This fix also apparently improves quoted and realized prices slightly, even though I wasn't trying to.

Also updated the ethgasstation poll rate to ~half the block time, because gas price is a lot more mercurial during the end of the world.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
@@ -34,7 +34,6 @@ function mixPaths(
     targetInput: BigNumber,
     maxSteps: number = 2 ** 15,
 ): Fill[] {
-    const allFills = [...pathA, ...pathB].sort((a, b) => b.rate.comparedTo(a.rate));
     let bestPath: Fill[] = [];
     let bestPathInput = ZERO_AMOUNT;
     let bestPathRate = ZERO_AMOUNT;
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -77,6 +91,14 @@ function isPathComplete(path: Fill[], targetInput: BigNumber): boolean {
     return input.gte(targetInput);
 }
 
+function clipFillAdjustedOutput(fill: Fill, remainingInput: BigNumber): BigNumber {
+    if (fill.input.lte(remainingInput)) {
+        return fill.adjustedOutput;
+    }
+    const penalty = fill.adjustedOutput.minus(fill.output);
+    return fill.output.times(remainingInput.div(fill.input)).plus(penalty);
```
dekz: This is negative right? If so it would be clearer if it was: 

```suggestion
    const penalty = fill.output.minus(fill.adjustedOutput);
    return fill.output.times(remainingInput.div(fill.input)).minus(penalty);
```

dorothy-zbornak: Nah, it can be positive for buys.

PR #2528, opened 2020-03-23
Update solc to 0.6.x

HenryNguyen5:
## Description
Update solc to 0.6.x

Solc 0.6.x addresses a rate limiting fix that prevents downstream users
from downloading the required version of solc-js more often.
See: https://github.com/ethereum/solc-js/pull/446

## Testing instructions
Should just pass CI.

## Types of changes
- Change typescript typings to reflect new solc breaking changes to API.
- Update `sol-compiler` to use `compile` instead of `compileStandardWrapper`. It has been renamed to be the same thing.

### DIFF
--- a/packages/sol-compiler/package.json
+++ b/packages/sol-compiler/package.json
@@ -91,7 +91,7 @@
         "pluralize": "^7.0.0",
         "require-from-string": "^2.0.1",
         "semver": "5.5.0",
-        "solc": "^0.5.5",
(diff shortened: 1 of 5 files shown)


dorothy-zbornak:
Thanks for getting the ball rolling on this! I've put up a PR (#2532) that incorporates this change + a little more so I'll close this one in the meantime.

PR #2532, opened 2020-03-27
sol-compiler: 0.6 support

dorothy-zbornak:
## Description

Addresses #2528 

This adds solidity 0.6 support to `sol-compiler`. I've also done some decent-sized refactors to (hopefully) make compatibility with new versions/features easier to integrate, and to tidy up the code. There are some other small changes listed below.

### Parallelization
I rewrote a bunch of the code to be parallelized, to speed up build times.

### Artifact Changes
- solc version string in artifact `settings` object is now normalized to `{VERSION}+commit.{COMMIT_HASH}`. Previously it could take on many forms, depending on where your compiler comes from (docker, npm, direct path).
- Paths for source files in the output artifacts are now expressed in dot-notation relative paths. Previously they were expressed as path segments, which made it difficult to tell if a source file was an npm module or not.
- Input artifacts now include the solidity `version` in their `settings` object. This makes automated verification a bit easier because we don't have to look up the solidity version used from the output artifacts as well. 

## Checklist:


packages/sol-compiler/src/solc_wrapper_v05.ts
```diff
@@ -0,0 +1,108 @@
...
+            sources: {},
+            settings: {
+                remappings: [],
+                ...this._compilerSettings,
+            },
+        };
+        for (const [contractPath, contractContent] of Object.entries(contractsByPath)) {
+            input.sources[contractPath] = { content: contractContent };
+        }
+        for (const [prefix, _path] of Object.entries(importRemappings)) {
```
jalextowle: Have we always supported import remappings? This is a pretty nice feature

packages/sol-compiler/src/solc_wrapper_v06.ts
```diff
@@ -0,0 +1,25 @@
+import { CompilerOptions, StandardOutput } from 'ethereum-types';
+import solc = require('solc');
+
+import { compileSolcJSAsync, getSolcJSAsync } from './utils/compiler';
+
+import { SolcWrapperV05 } from './solc_wrapper_v05';
+
+// 0.6.x has a `compile()` function in lieu of `compileStandardWrapper`.
+type SolcV06 = solc.SolcInstance & { compile(input: string): string };
```
jalextowle: Should this be a union type rather than an intersection type? It would seem that this would remove all type information since `solc.SolcInstance` doesn't actually have a `compile` function with the same function signature as `compile(input: string): string`.

Source:
```
    export interface SolcInstance {
        compile(
            sources: InputSources,
            optimizerEnabled: number,
            findImports: (importPath: string) => ImportContents,
        ): CompilationResult;
        compileStandardWrapper(input: string, findImports?: (importPath: string) => ImportContents): string;
    }
```

Please let me know if I'm missing something here. If I am not, it may be necessary to add the new `compile` function to `typescript-typings`

packages/sol-compiler/src/utils/compiler.ts
```diff
@@ -152,18 +153,42 @@ export async function compileSolcJSAsync(
 
 /**
  * Compiles the contracts and prints errors/warnings
- * @param solcVersion Version of a solc compiler
+ * @param solidityVersion Solidity version
```
jalextowle: 🙏

packages/sol-compiler/src/utils/compiler.ts
```diff
@@ -152,18 +153,42 @@ export async function compileSolcJSAsync(
...
 ): Promise<solc.StandardOutput> {
     const standardInputStr = JSON.stringify(standardInput, null, 2);
-    const dockerCommand = `docker run -i -a stdin -a stdout -a stderr ethereum/solc:${solcVersion} solc --standard-json`;
-    const standardOutputStr = execSync(dockerCommand, { input: standardInputStr }).toString();
-    const compiled: solc.StandardOutput = JSON.parse(standardOutputStr);
-    return compiled;
+    const dockerArgs = [
+        'run',
+        '-i',
+        '-a',
```
jalextowle: Optional nit: It could be clearer to group the options with their arguments like:

```
'-a stdin'
```

I'm pretty sure that's kosher (the docs for stream don't define a schema for the strings that can be elements of `args`), but please correct me if I'm wrong.

packages/sol-compiler/src/solc_wrapper_v05.ts
```diff
@@ -0,0 +1,108 @@
...
+    protected readonly _compilerSettings: solc.CompilerSettings;
+
+    public static normalizeOutput(
+        output: StandardOutput,
+        importRemappings: ImportPrefixRemappings,
+        opts: CompilerOptions,
+    ): StandardOutput {
+        const _output = _.cloneDeep(output);
+        // _output.sources = makeContractPathsRelative(_output.sources, opts.contractsDir!, importRemappings);
+        // _output.contracts = makeContractPathsRelative(_output.contracts, opts.contractsDir!, importRemappings);
```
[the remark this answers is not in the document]
dorothy-zbornak: oops

packages/sol-compiler/src/solc_wrapper_v05.ts
dorothy-zbornak: Yeah, this is nothing new.

packages/sol-compiler/src/compiler.ts
```diff
@@ -83,60 +70,75 @@ interface ContractData {
...
-    private readonly _useDockerisedSolc: boolean;
     private readonly _isOfflineMode: boolean;
     private readonly _shouldSaveStandardInput: boolean;
+    private readonly _solcWrappersByVersion: { [version: string]: SolcWrapper } = {};
+
+    public static async getCompilerOptionsAsync(
+        overrides: Partial<CompilerOptions> = {},
+        file: string = 'compiler.json',
+    ): Promise<CompilerOptions> {
+        // TODO: Look for config file in parent directories if not found in current directory
```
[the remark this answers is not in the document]
dorothy-zbornak: This TODO was from before, but I agree. An `extend` pattern would do the same with explicitness.

packages/sol-compiler/src/compiler.ts
```diff
@@ -83,60 +70,75 @@ interface ContractData {
     contractName: string;
 }
 
+// tslint:disable no-non-null-assertion
```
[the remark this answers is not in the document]
dorothy-zbornak: I think it's mostly just in test code.

packages/sol-compiler/src/compiler.ts
```diff
@@ -83,60 +70,75 @@ interface ContractData {
...
+        assert.doesConformToSchema('compiler.json', fileConfig, compilerOptionsSchema);
+        return {
+            ...fileConfig,
+            ...overrides,
+        };
+    }
+
+    private static _createDefaultResolver(
+        contractsDir: string,
+        // tslint:disable-next-line: trailing-comma
```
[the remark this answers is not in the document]
dorothy-zbornak: TS throws a fit w/trailing commas on a rest param

packages/sol-compiler/src/utils/compiler.ts
dorothy-zbornak: Yeah the args have to be broken up. Otherwise it would be executed as `docker '-a stdin' ...`. I can group them on the same line and disable the prettier.

packages/sol-compiler/src/solc_wrapper_v06.ts
dorothy-zbornak: It's my understanding that intersection types _are_ unions. Union types are either-or. idk why TS picked those horrible names.  Feels weird to add the new compile method to `SolcInstance` because that's the 0.4/0.5 interface in my mind.

packages/sol-compiler/src/compiler.ts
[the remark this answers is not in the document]
jalextowle: Yep, I saw that. I only commented because it was kept around.

packages/sol-compiler/src/utils/compiler.ts
jalextowle: I see, up to you. I'm fine with it as-is.

packages/sol-compiler/src/solc_wrapper_v06.ts
jalextowle: Yeah that is a *horrible* name. I can't believe they chose that.

Carry on...

packages/sol-compiler/src/cli.ts
```diff
@@ -36,13 +36,16 @@ const SEPARATOR = ',';
...
-        contracts,
-        isOfflineMode: process.env.SOLC_OFFLINE ? true : undefined,
-    };
-    const compiler = new Compiler(opts);
+    const opts = _.omitBy(
+        {
+            contractsDir: argv.contractsDir,
+            artifactsDir: argv.artifactsDir,
+            contracts,
+            isOfflineMode: process.env.SOLC_OFFLINE ? true : undefined,
```
hysz: 🙏

packages/sol-compiler/src/utils/compiler.ts
```diff
@@ -437,3 +474,47 @@ export function getDependencyNameToPackagePath(
     });
     return dependencyNameToPath;
 }
+
+/**
+ * Extract the solidity version (e.g., '0.5.9') from a solc version (e.g., `0.5.9+commit.34d3134f`).
+ */
+export function getSolidityVersionFromSolcVersion(solcVersion: string): string {
+    const m = /(\d+\.\d+\.\d+)\+commit\.[a-f0-9]{8}/.exec(solcVersion);
```
hysz: Not sure if relevant, but we may want to handle uppercase hex with `[a-fA-F0-9]`. We may also be able to replace the hex expression with `[[:xdigit:]]`.

PR #2536, opened 2020-04-01
asset-swapper: DFB support + refactors

dorothy-zbornak:
## Description

This adds DexForwarderBridge support AS. When creating the orders, we batch _contiguous_ bridge paths into a single DFB order, so it only pays a single protocol fee. This functionality can be disabled with the `shouldBatchBridgeOrders` SwapQuoter option.

I've also massively refactored `swap_quote_calculator.ts`. I created some new functions in `quote_simulation.ts` whose responsibility is to simulate filling the quote. These first transform the quote orders into a side-agnostic format to remove the need for all the nasty branching logic. So instead of filling a straight maker -> taker asset order, it ends up filling  input -> output orders.

We also now have best-case gas estimation and best-case source composition (i.e., what sources were actually used). So now `bestCaseQuoteInfo.gas <= worstCaseQuoteInfo.gas`, where they were previously equivalent. Also `sourceBreakdown` now _only_ includes the sources consumed in the best case, so no fallback orders.

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -127,7 +127,7 @@ export interface OptimizedMarketOrder extends SignedOrderWithFillableAmounts {
     /**
      * The optimized fills that generated this order.
      */
-    fill: CollapsedFill;
+    fills: CollapsedFill[];
```
dorothy-zbornak, line 130: So now orders have an array of collapsed fills instead of a single one. Fun.

packages/asset-swapper/src/utils/swap_quote_calculator.ts
```diff
@@ -173,324 +166,75 @@ export class SwapQuoteCalculator {
...
-            feeTakerAssetAmount: takerAssetAmountWithFees.minus(takerAssetAmount),
-        };
-    }
+function fillResultsToQuoteInfo(fr: QuoteFillResult): SwapQuoteInfo {
     return {
-        feeTakerAssetAmount: constants.ZERO_AMOUNT,
-        takerAssetAmount: takerAssetAmountWithFees,
+        makerAssetAmount: fr.totalMakerAssetAmount,
+        takerAssetAmount: fr.takerAssetAmount,
+        totalTakerAssetAmount: fr.totalTakerAssetAmount,
```
dorothy-zbornak, line 235: A consequence of this new approach is that _both_ the maker (in buys) and taker (in sells) asset amounts can theoretically deviate from the input amount slightly when fees are involved. In practice they're probably almost always the same though. So if we want to reference a fixed buy or sell amount in the API, we should be using the `quote.makerAssetAmount` or `quote.takerAssetFillAmount`, not the best or worst case values. Just something to note.

PR #2540, opened 2020-04-07
ZeroEx: Universal proxy

dorothy-zbornak:
## Description

tenor

This is the MVP of the [ZeroEx proxy contract](https://forum.0x.org/t/a-flexible-and-sustainable-exchange-contract-architecture). For the uninitiated, this is a per-function proxy pattern.

Some things have deviated from the first version of the spec:
- No more ingress `msgSender` tracking :cry: .
  - This is because we can't write to state when a read-only function is called, making it impossible to guarantee that `msgSender` will be correct. For example, the scenario `Alice -> Exchange (rw) -> Bob -> Exchange (ro) -> Exchange (ro)` would have a `msgSender` of `Alice` because we can't write to state in the ro calls.
  - We could establish a best practice of read-only functions should never trust `msgSender`, but it could be difficult to police if the read occurs deeper in the call stack.
  - We can still live in a world where function wrappers are easier by registering `onlySelf` internal fill functions that take an explicit taker (e.g., `fillOrderInternal(address taker...)`).
- `Ownable` is also feature.
- `bootstrap()` is also a feature, and happens outside the constructor because turns out you can't external call yourself in a constructor.
- Added a new `Migrate` feature for migrations.

For deeper deets and diagrams, check the update spec doc.

Odds and ends:
- This version required some contracts from `@0x/contracts-utils` to be upgraded to 0.6. So now there are 0.6-compatible copies of those contracts in `contracts/utils/contracts/src/v06`.
- `@0x/sol-compiler` had to be tweaked to omit `receive` functions in generated 0.6 ABIs because `ethers.js` gets very noisy when it encounters these entries in artifacts.

## Checklist:


contracts/zero-ex/contracts/src/errors/LibCommonRichErrors.sol
```diff
@@ -0,0 +1,36 @@
...
+
+    // solhint-disable func-name-mixedcase
+
+    function OnlyCallableBySelfError(address sender)
+        internal
+        pure
+        returns (bytes memory)
+    {
+        return abi.encodeWithSelector(
+            bytes4(keccak256("OnlyCallableBySelfError(address)")),
```
dorothy-zbornak, line 32: ya, I started inlining this computation:
- It only saves <100 gas to precompute the selector
- It was so annoying to precompute the selector
- This is less error prone
- This code rarely executes anyway

contracts/zero-ex/contracts/src/features/Ownable.sol
```diff
@@ -0,0 +1,79 @@
...
+    IOwnable,
+    IZeroExBootstrapper,
+    FixinOwnable
+{
+    // solhint-disable const-name-snakecase
+
+    /// @dev Name of this feature.
+    string constant public override FEATURE_NAME = "Ownable";
+    /// @dev Version of this feature.
+    uint256 constant public override FEATURE_VERSION = (1 << 64) | (0 << 32) | (0);
```
dorothy-zbornak, line 43: This is a pattern for versioning feature contracts. A consumer can use this in conjunction with `getFunctionImplementation` and `IFeature` to "version" a function.

contracts/zero-ex/contracts/src/features/SimpleFunctionRegistry.sol
```diff
@@ -0,0 +1,134 @@
...
+    string constant public override FEATURE_NAME = "SimpleFunctionRegistry";
+    /// @dev Version of this feature.
+    uint256 constant public override FEATURE_VERSION = (1 << 64) | (0 << 32) | (0);
+
+    /// @dev Initializes the feature implementation registry.
+    /// @param impl The actual address of this feature contract.
+    function bootstrap(address impl) external override {
+        // Register the registration functions (inception vibes).
+        _extend(this.extend.selector, impl);
+        _extend(this.extendSelf.selector, impl);
```
dorothy-zbornak, line 50: This is temporary and will get de-registered in the `BasicMigration` migrator.

packages/sol-compiler/src/solc_wrapper_v06.ts
```diff
@@ -22,4 +22,15 @@ export class SolcWrapperV06 extends SolcWrapperV05 {
         solcInstance.compileStandardWrapper = solcInstance.compile;
         return compileSolcJSAsync(solcInstance, input);
     }
+
+    protected _normalizeOutput(output: StandardOutput): StandardOutput {
+        const _output = super._normalizeOutput(output);
+        // Filter out 'receive' ABI item types until ethers supports it.
+        for (const contracts of Object.values(_output.contracts)) {
+            for (const contract of Object.values(contracts)) {
+                contract.abi = contract.abi.filter(v => v.type !== 'receive');
```
dorothy-zbornak, line 31: If we don't filter these out, `ethers.js` will spam the console with `WARNING`s. Waiting for a fix in the meantime, but it shouldn't make a difference if we leave it out for now.

packages/ethereum-types/src/index.ts
```diff
@@ -91,9 +91,9 @@ export interface MethodAbi {
     name: string;
     inputs: DataItem[];
     outputs: DataItem[];
-    constant: boolean;
+    constant?: boolean;
     stateMutability: StateMutability;
-    payable: boolean;
+    payable?: boolean;
```
dorothy-zbornak, line 96: `constant` and `payable` are optional in 0.6 ABIs

.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ jobs:
                   path: ~/repo/packages/abi-gen/test-cli/output
```
dorothy-zbornak, line 38: git diff does a crap job w/this file so I'l summarize what went on:
- CI was running out of memory in `test-contracts-rest` so I moved staking to its own job
- I removed the `3.0` suffixes

contracts/zero-ex/contracts/src/ZeroEx.sol
```diff
@@ -0,0 +1,122 @@
...
+
+    /// @dev Fallback for just receiving ether.
+    receive() external payable {}
+
+    // solhint-enable state-visibility
+
+    /// @dev Bootstrap the initial feature set of this contract.
+    ///      This can only be called once by the deployer of this contract.
+    /// @param bootstrappers Array of bootstrapping contracts to delegatecall into.
+    function bootstrap(IZeroExBootstrapper[] calldata bootstrappers) external {
```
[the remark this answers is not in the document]
dorothy-zbornak: No. We need to bootstrap to set up the minimum viable featureset (`extend()` + `transferOwnership()`). If we don't have these, the proxy is essentially non-interactive.

contracts/zero-ex/contracts/src/ZeroEx.sol
```diff
@@ -0,0 +1,122 @@
...
+                LibProxyRichErrors.InvalidBootstrapCallerError(msg.sender, bootstrapCaller)
+            );
+        }
+        // Prevent calling `bootstrap()` again.
+        stor.bootstrapCaller = address(0);
+
+        // Call the bootstrap contracts.
+        for (uint256 i = 0; i < bootstrappers.length; ++i) {
+            // Delegatecall into the bootstrap contract.
+            (bool success, bytes memory resultData) = address(bootstrappers[i])
```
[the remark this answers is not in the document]
dorothy-zbornak: Currently they (SimpleFunctionRegistry and Ownable) have nothing useful to return, bootstrap() only happens once in a lifetime, and the bootstrappers are totally deterministic.

contracts/zero-ex/contracts/src/features/SimpleFunctionRegistry.sol
```diff
@@ -0,0 +1,139 @@
...
+        // Register the rollback function.
+        _extend(this.rollback.selector, impl);
+    }
+
+    /// @dev Roll back to the last implementation of a function.
+    ///      Only directly callable by an authority.
+    /// @param selector The function selector.
+    function rollback(bytes4 selector)
```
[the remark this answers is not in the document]
dorothy-zbornak: By version, do you mean implementation address? So like `rollback(FN_SELECTOR, address(OldFnImplementation))`? I think that's useful.

I'm surprised the multisig allows execution of txs out of order. We should correct this in time for governance.

packages/sol-compiler/src/solc_wrapper_v06.ts
dorothy-zbornak: You mean add a longer comment?

contracts/zero-ex/contracts/src/ZeroEx.sol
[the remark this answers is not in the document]
hysz: Ah yeah, saw this below. Makes sense!

packages/sol-compiler/src/solc_wrapper_v06.ts
hysz: Ah nvm, misread and thought there were two separate issues. Inline comment looks good.

contracts/zero-ex/contracts/src/ZeroEx.sol
```diff
@@ -0,0 +1,122 @@
...
+    constructor() public {
+        // Set the `bootstrap()` caller to the deployer.
+        LibProxyStorage.getStorage().bootstrapCaller = msg.sender;
+    }
+
+    // solhint-disable state-visibility
+
+    /// @dev Forwards calls to the appropriate implementation contract.
+    fallback() external payable {
+        bytes4 selector = msg.data.readBytes4(0);
```
abandeali1, line 48: Solidity supports native calldata array slices now FYI: https://solidity.readthedocs.io/en/v0.6.0/types.html#array-slices. I think that means we can get rid of `LibBytesV06`?

dorothy-zbornak: Best news I've heard all month.

contracts/zero-ex/contracts/src/storage/LibOwnableStorage.sol
```diff
@@ -0,0 +1,41 @@
...
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+
+/// @dev Storage helpers for the `Ownable` feature.
+library LibOwnableStorage {
+
+    /// @dev Globally unique offset for the storage bucket.
+    bytes32 constant internal STORAGE_ID =
+        0xeef73acb590dd70cb88ccc8e9832ea7f198de2f3c87ff92d610497d647795b3c;
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah.

contracts/zero-ex/contracts/src/ZeroEx.sol
dorothy-zbornak: > As of now, array slices are only implemented for calldata arrays.

womp, womp

dorothy-zbornak: Still can't cast to `bytes4` without assembly and we'll probably use LibBytes sooner than we think so I think I'll keep it for now.

contracts/zero-ex/test/storage_uniqueness_test.ts
```diff
@@ -0,0 +1,34 @@
+import { describe } from '@0x/contracts-test-utils';
+import { promises } from 'fs';
+import { basename, resolve } from 'path';
+
+const { readdir, readFile } = promises;
+
+describe('Storage ID uniqueness test', () => {
```
dorothy-zbornak, line 7: I added tests to detect if storage IDs aren't unique. This works so long as we follow the basic pattern of the storage library files.

contracts/zero-ex/contracts/src/features/SimpleFunctionRegistry.sol
```diff
@@ -0,0 +1,184 @@
...
+        address[] storage history = stor.implHistory[selector];
+        uint256 i = history.length;
+        for (; i > 0; --i) {
+            address impl = history[i - 1];
+            history.pop();
+            if (impl == targetImpl) {
+                break;
+            }
+        }
+        if (i == 0) {
```
hysz, line 87: Wouldn't this get triggered below when we try to rollback `extendSelf`?

dorothy-zbornak: No, because `extendSelf` gets added through `_extend()`, which pushes the old implementation (null) to the stack, so it has a history at that point.

packages/utils/src/revert_error.ts
```diff
@@ -113,7 +113,9 @@ export abstract class RevertError extends Error {
         const instance = new type();
         try {
             const values = decoder(_bytes);
-            return _.assign(instance, { values });
+            _.assign(instance, { values });
+            instance.message = instance.toString();
```
dorothy-zbornak, line 117: This allows revert errors to finally print their full payloads in the stack traces. You're welcome.

contracts/zero-ex/contracts/src/storage/LibOwnableStorage.sol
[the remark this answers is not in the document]
hysz: What would you guys think of a pattern like this? 
```
library LibStorage {
    enum Buckets {
        Proxy,
        Ownable,
        SimpleFunctionRegistry
    }
}

library LibProxyStorage {
    function getStorage() internal pure returns (Storage storage stor) {
        uint256 storageId = uint256(LibStorage.Buckets.Proxy);
        assembly { stor_slot := storageId }
    }
}
```

Would make it easier to verify that code is not accidentally writing to the wrong slot in review. Also tracks which bucket points to which slot when a bucket is deprecated (ie, would change `Proxy` to like `Proxy_Deprecated_v200`).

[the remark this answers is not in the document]
dorothy-zbornak: We could move the IDs into a single file but I'm leery of using an enum because the slots would be right next to each other. So might two structs with more than one value field overlap?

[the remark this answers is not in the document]
dorothy-zbornak: Yeah, this reverts.

```solidity
pragma solidity ^0.6.5;

contract BucketTest {
    struct Bucket1 {
        uint256 a;
        uint256 b;
    }
    
    struct Bucket2 {
        uint256 c;
    }
    
    function test() external {
        Bucket1 storage b1;
        Bucket2 storage b2;
        assembly {
            b1_slot := 1
            b2_slot := 2
        }
        b2.c = 123;
        b1.a = 456;
        b1.b = 789;
        require(b2.c == 123, 'OVERLAP_DETECTED');
    }
}
```

It would work if we hashed the enum right beforehand but that introduces overhead and these get called a lot.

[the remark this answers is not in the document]
dorothy-zbornak: Or how about we just multiply the storage ID by a big number like `storageId * 1e6` to space them out?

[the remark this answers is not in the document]
mzhu25: why not just `bytes32 constant internal STORAGE_ID = keccak256(Buckets.Proxy)` ?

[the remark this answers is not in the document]
hysz: Ah that's right, good catch! I'm down for `storageId * 1e6`.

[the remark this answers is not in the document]
dorothy-zbornak: Because keccak in a constant still gets run every time in 0.6 😟

contracts/zero-ex/contracts/src/storage/LibStorage.sol
```diff
@@ -0,0 +1,49 @@
...
+
+    /// @dev Get the storage offset given a storage ID.
+    /// @param storageId An entry in `StorageId`
+    /// @return offset The storage offset.
+    function getStorageOffset(StorageId storageId)
+        internal
+        pure
+        returns (uint256 offset)
+    {
+        return uint256(storageId) * STORAGE_OFFSET_MULTIPLIER;
```
hysz, line 47: Would probably be good to use LibSafeMath as a sanity check.

dorothy-zbornak: If we're gonna use LibSafeMath then we might lose the gas advantage over a keccak? The uint256 cast will also do a range check on the enum, so the overflow could realistically only occur if the constant is too large. We'd have to have something like 10^51 enum values for this to happen with the current multiplier.

contracts/zero-ex/contracts/src/features/Migrate.sol
```diff
@@ -0,0 +1,94 @@
...
+    function bootstrap(address impl) external returns (bytes4 success) {
+        // Register feature functions.
+        ISimpleFunctionRegistry(address(this)).extend(this.migrate.selector, impl);
+        ISimpleFunctionRegistry(address(this)).extend(this.getMigrationOwner.selector, impl);
+        return LibBootstrap.BOOTSTRAP_SUCCESS;
+    }
+
+    /// @dev Execute a migration function in the context of the ZeroEx contract.
+    ///      The result of the function being called should be the magic bytes
+    ///      0x2c64c5ef. Only callable by the owner.
```
hysz, line 58: How are the magic bytes throughout these files generated? Are they fn selectors?

dorothy-zbornak: It's the hash of `MIGRATE_SUCCESS` and `BOOTSTRAP_SUCCESS`. I never really understood the rationale behind using fn selectors. Wouldn't you want a distinct value that you can't accidentally return?

hysz: Yeah, this works. Could we just add a cheeky comment about how it's generated so future devs know to use the same pattern?

contracts/zero-ex/contracts/src/storage/LibStorage.sol
hysz: Ah that's right we have the protective range checks now. Could we add a comment about this implicit protection for future readers?

contracts/zero-ex/test/storage_uniqueness_test.ts
```diff
@@ -0,0 +1,36 @@
+import { describe } from '@0x/contracts-test-utils';
+import { readdir, readFile } from 'fs';
+import { basename, resolve } from 'path';
+import { promisify } from 'util';
+
+describe('Storage ID uniqueness test', () => {
```
mzhu25, line 6: we can delete these now

dorothy-zbornak: I actually updated the regex to extract the `StorageId` enum used in `getStorage()` but forgot to also update the test case descriptions :upside_down_face:. I would never have noticed if you hadn't said anything.

PR #2541, opened 2020-04-07
asset-swapper: RFQ-T firm quotes

feuGeneA:
## Description

These changes support https://github.com/0xProject/0x-api/pull/162

Every commit is crafted to be completely independent and atomic. I suggest you review them one at a time, in order to give better context, rather than just reviewing all of the changes at once.

## Testing instructions

While there aren't any tests here yet, this code is currently being exercised by a 0x-api test, which you can see in the RFQ-T commit at https://github.com/0xProject/0x-api/pull/162 , and which is using the `asset-swapper`, and `contract-addresses` packages from this branch.

The deployment of ERC20BridgeSampler was necessary not just for the RFQ-T test but also for the vanilla `/swap/v0/quote` endpoint test as well.  Before that deployment, the `market_operation_utils` code was reverting when calling into the sampler, resulting in a HTTP 400 response from the 0x API.

## Types of changes

## Checklist:

-   [x] All Comments addressed or cited below.
-   [x] https://github.com/0xProject/0x-monorepo/pull/2541#discussion_r405837623
-   [ ] https://github.com/0xProject/0x-monorepo/pull/2541#pullrequestreview-391267634


packages/asset-swapper/src/types.ts
```diff
@@ -213,6 +217,8 @@ export interface SwapQuoterOpts extends OrderPrunerOpts {
     contractAddresses?: ContractAddresses;
     samplerGasLimit?: number;
     liquidityProviderRegistryAddress?: string;
+    rfqtTakerApiKeyWhitelist: string[];
```
feuGeneA, line 220: make this optional

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+        marketOperation: MarketOperation,
+        intentOnFilling: boolean,
+        takerApiKey: string,
+        takerAddress: string,
+    ): Promise<SignedOrder[]> {
+        const makerEndpointMaxResponseTimeMs = 1000;
+
+        const getTokenAddressOrThrow = (assetData: string): string => {
+            const decodedAssetData = assetDataUtils.decodeAssetDataOrThrow(assetData);
+            if (decodedAssetData.hasOwnProperty('tokenAddress')) {
```
feuGeneA, line 31: try to use types instead of forced casting

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const responsePromises: Array<Promise<AxiosResponse<SignedOrder> | undefined>> = [];
+        for (const rfqtMakerEndpoint of this._rfqtMakerEndpoints) {
+            responsePromises.push(
+                Axios.get(`${rfqtMakerEndpoint}/quote`, {
+                    headers: { '0x-api-key': takerApiKey },
+                    params: {
+                        sellToken,
+                        buyToken,
+                        buyAmount: marketOperation === MarketOperation.Buy ? assetFillAmount.toString() : undefined,
```
feuGeneA, line 61: We'll be telling external makers that they should only focus on serving quotes with buyAmount.  Can optimize latency here by only hitting quote providers if a buyAmount was provided; why waste time on a sellAmount request if we know quote providers won't give a quote anyways.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const responsesIfDefined: Array<AxiosResponse<SignedOrder> | undefined> = await Promise.all(responsePromises);
+
+        const responses: Array<AxiosResponse<SignedOrder>> = responsesIfDefined.filter(
+            (respIfDefd): respIfDefd is AxiosResponse<SignedOrder> => respIfDefd !== undefined,
+        );
+
+        const ordersWithStringInts = responses.map(response => response.data); // not yet BigNumber
+
+        const orders: SignedOrder[] = ordersWithStringInts.map(orderWithStringInts => {
```
feuGeneA, line 85: Find a helper function do this for us.  @steveklebanoff ?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsePromises: Array<Promise<AxiosResponse<SignedOrder> | undefined>> = [];
+        for (const rfqtMakerEndpoint of this._rfqtMakerEndpoints) {
+            responsePromises.push(
+                Axios.get(`${rfqtMakerEndpoint}/quote`, {
```
steveklebanoff, line 56: We're not doing any validation about what's being returned from the RFQT provider here.

Some things to consider:
- What if this URL returns a non-200 response?
- What if the RFQT provider returns something other than a `SignedOrder`?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -144,7 +148,12 @@ export class SwapQuoter {
      *
      * @return  An instance of SwapQuoter
      */
-    constructor(supportedProvider: SupportedProvider, orderbook: Orderbook, options: Partial<SwapQuoterOpts> = {}) {
+    constructor(
+        supportedProvider: SupportedProvider,
+        orderbook: Orderbook,
+        options: Partial<SwapQuoterOpts> = {},
+        quoteRequestor?: QuoteRequestor,
```
dorothy-zbornak, line 155: Why not just add `quoteRequestor` to `SwapQuoterOpts`?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,10 +535,29 @@ export class SwapQuoter {
             gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
         }
         // get the relevant orders for the makerAsset
-        let prunedOrders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
+        let orders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
+        if (options.intentOnFilling && options.apiKey) {
+            if (!this.rfqtTakerApiKeyWhitelist.includes(options.apiKey)) {
+                throw new Error('API key not permissioned for RFQ-T');
+            }
+            if (!options.takerAddress || options.takerAddress === '0x0000000000000000000000000000000000000000') {
```
dorothy-zbornak, line 543: There is a `NULL_ADDRESS` in `constants.ts`

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+    constructor(rfqtMakerEndpoints: string[]) {
+        this._rfqtMakerEndpoints = rfqtMakerEndpoints;
+    }
+
+    public async requestRfqtFirmQuotesAsync(
+        makerAssetData: string,
+        takerAssetData: string,
+        assetFillAmount: BigNumber,
+        marketOperation: MarketOperation,
+        intentOnFilling: boolean,
```
dorothy-zbornak, line 23: What is the point of this param? To get here this has to be `true`, and it never gets used in this function.

packages/asset-swapper/src/utils/quote_requestor.ts
dorothy-zbornak: Why do makers prefer buys over sells?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsePromises: Array<Promise<AxiosResponse<SignedOrder> | undefined>> = [];
+        for (const rfqtMakerEndpoint of this._rfqtMakerEndpoints) {
+            responsePromises.push(
```
dorothy-zbornak, line 55: Nit: We can maybe do this in a more idomatic fashion with an async lambda like:

```ts
const responsesIfDefined = await Promise.all(this._rfqtMakerEndpoints.map(async endpoint => {
    try {
        return await Axios.get(...);
    } catch (err) {
        ...
    }
}));
```

TS should also be able to infer the return type this way also.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,10 +535,29 @@ export class SwapQuoter {
...
+                throw new Error('RFQ-T requests must specify a taker address');
+            }
+            orders = orders.concat(
```
dorothy-zbornak, line 546: Optional nit:
```ts
            orders.push(...(await ...));
```

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,10 +535,29 @@ export class SwapQuoter {
             gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
         }
         // get the relevant orders for the makerAsset
-        let prunedOrders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
+        let orders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
```
dorothy-zbornak, line 538: Why can't we do this call and the RFQT request in the sample `Promise.all()`?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+        takerAssetData: string,
+        assetFillAmount: BigNumber,
+        marketOperation: MarketOperation,
+        intentOnFilling: boolean,
+        takerApiKey: string,
+        takerAddress: string,
+    ): Promise<SignedOrder[]> {
+        const makerEndpointMaxResponseTimeMs = 1000;
+
+        const getTokenAddressOrThrow = (assetData: string): string => {
```
dorothy-zbornak, line 29: Why is this a closure? It could be useful elsewhere.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,99 @@
...
+            if (decodedAssetData.hasOwnProperty('tokenAddress')) {
+                // type cast necessary here as decodeAssetDataOrThrow returns
+                // an AssetData object, which doesn't necessarily contain a
+                // token address.  (it could possibly be a StaticCallAssetData,
+                // which lacks an address.)  so we'll just assume it's a token
+                // here.  should be safe, with the enclosing guard condition
+                // and subsequent error.
+                // tslint:disable-next-line:no-unnecessary-type-assertion
+                return (decodedAssetData as ERC20AssetData).tokenAddress;
+            } else {
```
dorothy-zbornak, line 40: Nit: Unecessary `else`

packages/asset-swapper/src/types.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/5602aacbd21494f38d9ac6b01b6248f552768aa1

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/1670b21123d52ce1e5cb1fd71d477f9d89d9e0c5

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/98fc780ade384f732bb71c0e62f77d5bec769fb9 .  Is that what you had in mind?

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/37390e03b41a9032fe2a8f772235e8b69dacb6be

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: We can and we should!  Per the spec we wrote!  Thank you for noticing this!  Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/5c3929180944eb4c706f6401a9ef81ab411fa264

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/7096a70bc422bffecf7476110024fbb8cd78885c

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I just wrote it "as local as possible."  Figured we'd graduate it if/when it becomes useful elsewhere.  (Though how we'd remember it exists when it comes time to use it is undefined.)  I'd be happy to move it wherever you see fit.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Is it?  Is it not possible that an API client could pass in a `StaticCallAssetData` and have it flow all the way through to here?

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/63bfd233108f3fb9a918d722f2e4f0461e838b37

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/aadcf8fed0ec5ecc202645892180934c02845671

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: It's not used yet because so far we're only supporting firm quotes.  This function will change such that if `intentOnFilling` is true, then it hits the maker's `/quote` endpoint, but if it's false then it will hit the maker's `/price` endpoint instead.

dorothy-zbornak: I see. Will that be in this PR or soon after? I'm not usually a fan of placeholder code because sometimes we pivot and forget.

packages/asset-swapper/src/utils/quote_requestor.ts
dorothy-zbornak: I think it makes sense to just make this a regular, unexported function in this file.

packages/asset-swapper/src/utils/quote_requestor.ts
dorothy-zbornak: I mean we return in the `if` clause so nothing after it can execute anyway.

packages/asset-swapper/src/utils/quote_requestor.ts
dorothy-zbornak: Partly. Does it not work with `try/catch`?

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Very soon after.  We're not putting off indicative quotes; they're a key part of RFQ-T.  I'm doing firm quotes in one milestone, and indicative in the next.

packages/asset-swapper/src/constants.ts
```diff
@@ -43,6 +43,8 @@ const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {
     },
     ...DEFAULT_ORDER_PRUNER_OPTS,
     samplerGasLimit: 250e6,
+    rfqtTakerApiKeyWhitelist: [],
```
dekz, line 46: consider pushing these down into an `rfqt` namespace.
```javascript
rfqt: {
  apiKeys: [],
  endpoints: [],
}
```

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -161,10 +165,13 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
+        this.rfqtTakerApiKeyWhitelist = options.rfqtTakerApiKeyWhitelist || [];
+        this.rfqtMakerEndpoints = options.rfqtMakerEndpoints || [];
```
dekz, line 169: `rfqtMakerEndpoints` seems to only be used to construct a `QuoteRequestor`, does it need to be set as a public accessor?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -36,12 +37,15 @@ export class SwapQuoter {
     public readonly expiryBufferMs: number;
     public readonly chainId: number;
     public readonly permittedOrderFeeTypes: Set<OrderPrunerPermittedFeeTypes>;
+    public readonly rfqtTakerApiKeyWhitelist: string[];
```
dekz, line 40: Do these need to be public?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,26 +530,48 @@ export class SwapQuoter {
             gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
         }
         // get the relevant orders for the makerAsset
-        let prunedOrders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
+        const orderPromises: Array<Promise<SignedOrder[]>> = [
+            this._getSignedOrdersAsync(makerAssetData, takerAssetData),
+        ];
+        if (options.intentOnFilling && options.apiKey) {
+            if (!this.rfqtTakerApiKeyWhitelist.includes(options.apiKey)) {
+                throw new Error('API key not permissioned for RFQ-T');
```
dekz, line 538: Is this a throwable validation at this layer? Or should it fail silently and give a normal quote? 

If someone gets delisted they're going to start getting Errors rather than quotes, seems quite disruptive.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,26 +530,48 @@ export class SwapQuoter {
...
+            if (!options.takerAddress || options.takerAddress === constants.NULL_ADDRESS) {
+                throw new Error('RFQ-T requests must specify a taker address');
+            }
+            orderPromises.push(
+                this._quoteRequestor.requestRfqtFirmQuotesAsync(
+                    makerAssetData,
+                    takerAssetData,
+                    assetFillAmount,
+                    marketOperation,
+                    options.intentOnFilling,
```
dekz, line 549: Consider an `opts` param

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,95 @@
...
+
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsesIfDefined: Array<void | AxiosResponse<SignedOrder>> = await Promise.all(
+            this._rfqtMakerEndpoints.map(async rfqtMakerEndpoint =>
+                Axios.get<SignedOrder>(`${rfqtMakerEndpoint}/quote`, {
```
dekz, line 55: we should be able try/catch
```
try {
  await Axios.get<SignedOrder>
} catch (err) {
  // log
}
```

packages/asset-swapper/src/swap_quoter.ts
dekz: Are we asserting they are `uris`?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,95 @@
...
+    public async requestRfqtFirmQuotesAsync(
+        makerAssetData: string,
+        takerAssetData: string,
+        assetFillAmount: BigNumber,
+        marketOperation: MarketOperation,
+        intentOnFilling: boolean,
+        takerApiKey: string,
+        takerAddress: string,
+    ): Promise<SignedOrder[]> {
+        const makerEndpointMaxResponseTimeMs = 1000;
```
dekz, line 27: This feels like a configuration option that should be exposed globally.

packages/asset-swapper/src/utils/quote_requestor.ts
dekz: Also interested to hear the preference for `buyAmount`. I would consider them both equally required, or if only one can be provided, then `sellAmount`

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,95 @@
...
+        };
+
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsesIfDefined: Array<void | AxiosResponse<SignedOrder>> = await Promise.all(
+            this._rfqtMakerEndpoints.map(async rfqtMakerEndpoint =>
```
dekz, line 54: We should consider whether a network blip should be handled with a retry.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,95 @@
...
+                Axios.get<SignedOrder>(`${rfqtMakerEndpoint}/quote`, {
+                    headers: { '0x-api-key': takerApiKey },
+                    params: {
+                        sellToken,
+                        buyToken,
+                        buyAmount: marketOperation === MarketOperation.Buy ? assetFillAmount.toString() : undefined,
+                        sellAmount: marketOperation === MarketOperation.Sell ? assetFillAmount.toString() : undefined,
+                        takerAddress,
```
dekz, line 62: When `takerAddress` is the `Forwarder`, how does the MM know the real `from`?

packages/asset-swapper/src/utils/quote_requestor.ts
dekz: I have a feeliing they exist in `@0x/connect`

packages/asset-swapper/src/types.ts
```diff
@@ -192,6 +193,10 @@ export interface SwapQuoteOrdersBreakdown {
  */
 export interface SwapQuoteRequestOpts extends CalculateSwapQuoteOpts {
     gasPrice?: BigNumber;
+    takerAddress?: string;
```
dekz, line 196: Consider pushing these down into `rfqt` since they're all grouped together and only make sense in that context.

packages/asset-swapper/src/swap_quoter.ts
fabioberger: Yeah, agree that we should fallback to providing a regular quote in this case.

packages/asset-swapper/src/types.ts
```diff
@@ -192,6 +193,10 @@ export interface SwapQuoteOrdersBreakdown {
...
+    apiKey?: string;
+    enableRfqt?: boolean;
```
fabioberger, line 198: This doesn't seem to be used anywhere.

packages/asset-swapper/src/utils/quote_requestor.ts
fabioberger: I agree with @dorothy-zbornak, let's not add placeholder code (in the same that we don't leave commented out code).

packages/asset-swapper/src/utils/quote_requestor.ts
fabioberger: Does only `ERC20AssetData` include a `tokenAddress` property? Also not sure if that's always hold into the future.

feuGeneA: No, `ERC721AssetData` and `ERC1155AssetData` both also include token addresses.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I misspoke in the initial comment here.  We're recommending serving quotes with sellAmount, not buyAmount.

We're just focusing on sells because buys are less than 20% of transactions, because many liquidity providers don't support buys, so we've been driving integrators towards sells.  We can expand later and support buys in the future.  Just reasoning about sells will be easier for market makers.

packages/asset-swapper/package.json
```diff
@@ -11,7 +11,7 @@
         "build": "yarn tsc -b",
         "watch": "tsc -w -p tsconfig.json",
         "build:ci": "yarn build",
-        "lint": "tslint --format stylish --project .",
+        "lint": "tslint --format stylish --project . && prettier --check 'src/**/*.{ts,tsx,json,md}' --config ../../.prettierrc",
```
steveklebanoff, line 14: We have the established pattern of having `prettier` as it's own script: https://github.com/0xProject/0x-monorepo/blob/development/packages/orderbook/package.json#L25

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -161,10 +165,13 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
+        this.rfqtTakerApiKeyWhitelist = options.rfqtTakerApiKeyWhitelist || [];
+        this.rfqtMakerEndpoints = options.rfqtMakerEndpoints || [];
         this._contractAddresses = options.contractAddresses || getContractAddressesForChainOrThrow(chainId);
         this._devUtilsContract = new DevUtilsContract(this._contractAddresses.devUtils, provider);
         this._protocolFeeUtils = new ProtocolFeeUtils(constants.PROTOCOL_FEE_UTILS_POLLING_INTERVAL_IN_MS);
         this._orderStateUtils = new OrderStateUtils(this._devUtilsContract);
+        this._quoteRequestor = options.quoteRequestor || new QuoteRequestor(this.rfqtMakerEndpoints);
```
steveklebanoff: For testing purposes it may be a lot easier to take in an instantiated `QuoteRequestor` as a param, instead of constructing it in the class from `this.rfqtMakerEndpoints`.  

I don't have a good sense of the other changes that would be needed to support that though.  Do you think it may be worthwhile or not?

packages/asset-swapper/package.json
```diff
@@ -53,6 +53,7 @@
         "@0x/orderbook": "^2.2.5",
         "@0x/utils": "^5.4.1",
         "@0x/web3-wrapper": "^7.0.7",
+        "axios": "^0.19.2",
```
steveklebanoff, line 56: When I ran `yarn` from the monorepo root locally, I got the root-level `yarn.lock` changes.  Do you have the same on your machine?

```
+++ b/yarn.lock
@@ -3076,6 +3076,12 @@ axios@^0.18.0:
     follow-redirects "^1.3.0"
     is-buffer "^1.1.5"

+axios@^0.19.2:
+  version "0.19.2"
+  resolved "https://registry.yarnpkg.com/axios/-/axios-0.19.2.tgz#3ea36c5d8818d0d5f8a8a97a6d36b86cdc00cb27"
+  dependencies:
+    follow-redirects "1.5.10"
+
 babel-code-frame@^6.22.0, babel-code-frame@^6.26.0:
   version "6.26.0"
   resolved "https://registry.yarnpkg.com/babel-code-frame/-/babel-code-frame-6.26.0.tgz#63fd43f7dc1e3bb7ce35947db8fe369a3f58c74b"
@@ -7659,6 +7665,12 @@ flush-write-stream@^1.0.0, flush-write-stream@^1.0.2:
     inherits "^2.0.1"
     readable-stream "^2.0.4"

+follow-redirects@1.5.10:
+  version "1.5.10"
+  resolved "https://registry.yarnpkg.com/follow-redirects/-/follow-redirects-1.5.10.tgz#7b7a9f9aea2fdff36786a94ff643ed07f4ff5e2a"
+  dependencies:
+    debug "=3.1.0"
+
```

packages/asset-swapper/src/types.ts
```diff
@@ -193,9 +193,11 @@ export interface SwapQuoteOrdersBreakdown {
  */
 export interface SwapQuoteRequestOpts extends CalculateSwapQuoteOpts {
     gasPrice?: BigNumber;
-    takerAddress?: string;
     apiKey?: string;
-    intentOnFilling?: boolean;
+    rfqt?: {
+        takerAddress?: string;
```
feuGeneA, line 198: `takerAddress` should not be optional in this context.  When it was its own key in `SwapQuoteRequestOpts` it made sense, but now that it's within the optional `rfqt` key, it should be required.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I removed this param for you guys, in https://github.com/0xProject/0x-monorepo/pull/2541/commits/227676c150a87404bce7e6d2e9b351ff2be137fa

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/121d51b414e698d94eb8d1a3a9d3130ff7431dc3 . Was there anything else I missed?

packages/asset-swapper/src/constants.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/93872ad7a30cd98bca8c4fe14087cb5b732b42b9

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed the `rfqtMakerEndpoints` bit in https://github.com/0xProject/0x-monorepo/pull/2541/commits/3c795d365d7680144c91234d0e2822d5156f9446.

We are not _directly_ asserting that they are URI's.  We're just passing them through to the quote requestor, and letting it fail silently if they're not valid.

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/fa617d2e9d194b501f132129e9be08ef1d727e6f

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/5f4778c852ed3d2dbed8db04d92f0786343e4c78

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: I dropped this param for now, because of [this comment thread](https://github.com/0xProject/0x-monorepo/pull/2541#discussion_r405901859), but when I add it back, in the next milestone, I will consider this.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/121d51b414e698d94eb8d1a3a9d3130ff7431dc3

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/8cdc05f58269c9a52368f649804c3891daee1402

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: They don’t, but they do know the API key, and that’s the important thing for them to know.

packages/asset-swapper/src/types.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/70add44fc1b5418fab24274beb9c7692bc4c8859

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Would you mind if we deferred that to another PR?  I made a note about it in [a list of Discovered Risks](https://0xproject.quip.com/qn1xA43e3ANE/0x-API-RFQ-T-Tech-Spec#dVHACAO8thX).

packages/asset-swapper/src/types.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/commit/23684285877f9bf12294056b0fad1a1d9b034358

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/39c2a757be66359f602400b610a8f303faf01f2f

packages/asset-swapper/src/utils/quote_requestor.ts
blakehendo: As long as they know the API key, they know which application is responsible for sending in the request.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,100 @@
...
+                            buyAmount: marketOperation === MarketOperation.Buy ? assetFillAmount.toString() : undefined,
+                            sellAmount:
+                                marketOperation === MarketOperation.Sell ? assetFillAmount.toString() : undefined,
+                            takerAddress,
+                        },
+                        timeout: makerEndpointMaxResponseTimeMs,
+                    });
+                } catch (err) {
+                    logUtils.warn(
+                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}: ${JSON.stringify(
```
steveklebanoff, line 71: When I actually got an error from a RFQT provider locally, `JSON.stringify(err)` didn't actually output anything useful:

```
Failed to get RFQ-T quote from market maker endpoint https://1338.0.0.1 for API key my-ko0l-api-key for taker address 0xd209925defc99488e3afff1174e48b4fa628302a: {}
```

Consider just doing a `logUtils.warn(err)` after this line

packages/asset-swapper/src/types.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/264407b7072b3dec01c27c872c2f43c5af40d690

packages/asset-swapper/package.json
feuGeneA: I introduced a `prettier` script in https://github.com/0xProject/0x-monorepo/pull/2541/commits/eb5ec58453bca786806c25dfe1b9f4269fca4ed4, but I also kept its invocation in the `lint` script.

I feel pretty strongly that `yarn lint` should run everything that CI's `static-tests` runs.  I'm open to having my mind changed if anyone else feels strongly about it, but that's why I did it this way.

packages/asset-swapper/src/types.ts
```diff
@@ -192,6 +193,11 @@ export interface SwapQuoteOrdersBreakdown {
  */
 export interface SwapQuoteRequestOpts extends CalculateSwapQuoteOpts {
     gasPrice?: BigNumber;
+    apiKey?: string;
```
dekz, line 196: Feels like apiKey should also be under `rfqt`, unless we're planning on using outside of the rfqt context.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,100 @@
...
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsesIfDefined: Array<void | AxiosResponse<SignedOrder>> = await Promise.all(
+            this._rfqtMakerEndpoints.map(async rfqtMakerEndpoint => {
+                try {
+                    return Axios.get<SignedOrder>(`${rfqtMakerEndpoint}/quote`, {
```
dekz, line 57: Since we're not awaiting this it won't land in the try/catch right? We should await here to make sure it does lang as expected.

We could probably have a test on this behaviour that when RFQT HTTP request does blow up, the user still gets a quote

dekz: We could also just do:
```
const response = await Axios.get<SignedOrder>...
return response.data
```
Then we won't need to `filter` and `map` iin the lines below, just filter for `undefined`.

dekz: @steveklebanoff maybe it's too late but do you have any thoughts on an idempotency of this request to the MM? Let's say we retry due to a network blip, we'd retry with the same `id` so the MM (if they did receive the first request, but we failed to get the response) would understand we're just asking for the same thing again, rather than a second quote for the taker.

This idempotency id could be in the params or as a header.

packages/asset-swapper/src/types.ts
feuGeneA: @fabioberger ?  will you be using `apiKey` here for meta transactions?  i coded it this way thinking that he would.

fabioberger: Let's leave it top-level, we will be using it for other initiatives, including meta-txns.

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: I think this is a cool idea for a follow-up iteration of RFQT, but out of scope for this first MVP iteration.

I added this idea to our working "RFQT Spec v3 Potential Changes" document 👍

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: I'm a little confused.  We _are_ enabling the client to pass in an instantiated `QuoteRequestor`.  But, it's within the options constructor param rather than having its own constructor param.

Having said that, while I wrote this a mind for injecting that dependency from the test scenario, but I didn't actually write a test to use that mechanism, so I may have been off base here.

feuGeneA: Maybe we should just scrap the whole notion.  I only wrote it that way because I thought we'd mock out the QuoteRequestor instance, to be able to test everything outside of it.  But if we're mocking out the Axios layer, then our tests can be exercising the QuoteRequestor as well, at which point there's no need to inject this dependency into the SwapQuoter.

packages/asset-swapper/package.json
feuGeneA: Just to close the loop in this context: Steve and spoke elsewhere, and while we don't fully understand the issue (it's perhaps due to Steve's older version of `yarn`), it is not a blocker to him developing locally.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: @dekz I committed https://github.com/0xProject/0x-monorepo/commit/026bd5802cf774654debd6549a220fc9d861c482 to address the try/catch issue you pointed out.  Is that what you had in mind?

feuGeneA: @dekz I committed https://github.com/0xProject/0x-monorepo/pull/2541/commits/0cb5e4553b4de9cb7f9503bce4410d1d366a0dd0 to address the try/catch issue you pointed out. Is that what you had in mind?

packages/asset-swapper/package.json
steveklebanoff: I think it's pretty weird to have different `lint` behavior for `asset-swapper` compared to all other packages.   IMHO it's important to be consistent here.

Unless we want to make this change wholesale across all repos (a can of worms I don't think we should open right now), I think we should make this consistent.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,26 +529,49 @@ export class SwapQuoter {
...
+                    makerAssetData,
+                    takerAssetData,
+                    assetFillAmount,
+                    marketOperation,
+                    options.apiKey,
+                    options.rfqt.takerAddress,
+                ),
+            );
+        }
+        const orders: SignedOrder[] = ([] as SignedOrder[]).concat(...(await Promise.all(orderPromises)));
```
steveklebanoff, line 555: Curious as to what's up with the craziness in the line?  Can we just do the following?

```
const orders: SignedOrder[] = await Promise.all(orderPromises);
```

steveklebanoff, line 555: Would the following work & be more readable?

```
const resolvedPromises = await Promise.all(orderPromises);
const orders = _.flatten(resolvedPromises);
```

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -523,26 +529,49 @@ export class SwapQuoter {
             gasPrice = await this._protocolFeeUtils.getGasPriceEstimationOrThrowAsync();
         }
         // get the relevant orders for the makerAsset
-        let prunedOrders = await this._getSignedOrdersAsync(makerAssetData, takerAssetData);
+        const orderPromises: Array<Promise<SignedOrder[]>> = [
+            this._getSignedOrdersAsync(makerAssetData, takerAssetData),
+        ];
+        if (
+            options.rfqt &&
```
steveklebanoff, line 536: Nit but it might be nice to assign this big `&&` into a var, i.e:

```ts
const isRfqtEnabled = options.rfqt && options.rfqt.intentOnFilling && options.apiKey && this._rfqtTakerApiKeyWhitelist.includes(options.apiKey)
if (isRfqtEnabled) {
   ....
}
```

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,100 @@
...
+    ): Promise<SignedOrder[]> {
+        const { makerEndpointMaxResponseTimeMs } = _.merge({}, constants.DEFAULT_RFQT_FIRM_QUOTE_REQUEST_OPTS, options);
+
+        const buyToken = getTokenAddressOrThrow(makerAssetData);
+        const sellToken = getTokenAddressOrThrow(takerAssetData);
+
+        // create an array of promises for quote responses, using "undefined"
+        // as a placeholder for failed requests.
+
+        const responsesIfDefined: Array<void | AxiosResponse<SignedOrder>> = await Promise.all(
```
steveklebanoff, line 54: A little confused here, I expected this to be `Array<undefined | AxiosResponse<SignedOrder>>`.  What am I missing where this is `void` instead of `undefined`?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,100 @@
...
+            this._rfqtMakerEndpoints.map(async rfqtMakerEndpoint => {
+                try {
+                    return await Axios.get<SignedOrder>(`${rfqtMakerEndpoint}/quote`, {
```
steveklebanoff, line 57: It seems really important that we validate that the response is actually a `SignedOrder`.

You can see @dekz do something like this in the orderbook package: 

https://github.com/0xProject/0x-monorepo/blob/development/packages/connect/src/utils/relayer_response_json_parsers.ts#L35

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,100 @@
...
+                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}: ${JSON.stringify(
+                            err,
+                        )}`,
+                    );
+                    return undefined;
+                }
+            }),
+        );
+
+        const responses: Array<AxiosResponse<SignedOrder>> = responsesIfDefined.filter(
```
steveklebanoff, line 80: Can we use https://lodash.com/docs/4.17.15#compact here?

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: @dekz Are you open to exporting `typeConverters` or `orderParsingUtils` from the `@0x/orderbook` package? 

https://github.com/0xProject/0x-monorepo/blob/master/packages/connect/src/utils/type_converters.ts#L24

Or is there somewhere else where this is already defined & exported?

steveklebanoff:
@dekz Here's a start to our testing effort: https://github.com/0xProject/0x-monorepo/pull/2549

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Great catch.  This is an artifact from when I was doing `map(async endpoint => Axios.get(...).catch(...)`.  Jacob and Lawrence had me change it to be `map(async endpoint => try { Axios.get(...) } catch {...}`.  As an array of potentially rejected promises, `tsc` demanded void.  When I switched to `try{}catch{}`, I introduced a `return undefined;` to the catch handler, which brought everything into alignment on the `undefined` as placeholder notion.  But I neglected to change this `void` to an `undefined`.  Fix is under test...

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/d55108ab605378e897b516c138ecef30e1348851

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/ccc9e18132f0c66d4d97ce6fe94b2759f5290102

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/bb15f78af0d715d64dca64df1f3c2354688ea89b

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: What do you think of this? https://github.com/0xProject/0x-monorepo/pull/2541/commits/27ca75d94f2a874ef8c54b3ce7f50b34643b0d97

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I cleaned this up a _little_ bit.  See https://github.com/0xProject/0x-monorepo/pull/2541/commits/b854fcdb72f619a0a406fccf127547a7e9e15084.  What do you think?

In general I'm loath to introduce a 3rd-party lib to do something the language can handle, but if you feel strongly about it, or if there's some established convention around it, I'm happy to defer.

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: IMHO I prefer it this way.  But I could be swayed.

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: As I said in the other comment, I generally would rather not introduce 3rd-party libs where it's not necessary.  And, I believe I have addressed the readability issue.  Please let me know what you think.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,104 @@
...
+        );
+
+        const responses = responsesIfDefined.filter(
+            (respIfDefd): respIfDefd is AxiosResponse<SignedOrder> => respIfDefd !== undefined,
+        );
+
+        const ordersWithStringInts = responses.map(response => response.data); // not yet BigNumber
+
+        const validatedOrdersWithStringInts = ordersWithStringInts.filter(order =>
+            this._schemaValidator.isValid(order, schemas.orderSchema),
```
steveklebanoff, line 87: I think we actually want `signedOrderSchema` here

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,104 @@
...
+                    return undefined;
+                }
+            }),
+        );
+
+        const responses = responsesIfDefined.filter(
+            (respIfDefd): respIfDefd is AxiosResponse<SignedOrder> => respIfDefd !== undefined,
+        );
+
+        const ordersWithStringInts = responses.map(response => response.data); // not yet BigNumber
```
steveklebanoff, line 84: I think the number values are strings, not ints at this point

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: Given that lodash is already a dependency of 0x-api, it feels natural to leverage it here. That being said, I'm ok with the code as-is

packages/asset-swapper/src/swap_quoter.ts
steveklebanoff: No biggie -- feel free to leave as-is

packages/asset-swapper/src/swap_quoter.ts
steveklebanoff: New commit is looking better, still a little hard to read though.   

As mentioned tho -- we already have lodash as a dependency, and we can import just the function we care about (dont need to import whole library), i'd prefer just to use the lodash function here.  That being said, I'm ok as-is

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,104 @@
...
+
+        const validatedOrdersWithStringInts = ordersWithStringInts.filter(order =>
+            this._schemaValidator.isValid(order, schemas.orderSchema),
+        );
+
+        const orders: SignedOrder[] = validatedOrdersWithStringInts.map(orderWithStringInts => {
```
steveklebanoff, line 90: What will happen if the MM returned a valid `SignedOrder` but for the wrong assets?  Seems like we should test for that and not include the order if it doesn't match what we expect wrt maker and taker assets.

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: whoops, i misread the variable name here.  this var name seems OK

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: I did this here: https://github.com/0xProject/0x-monorepo/pull/2552

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: I fixed this in this PR: https://github.com/0xProject/0x-monorepo/pull/2552

packages/asset-swapper/src/types.ts
feuGeneA: Fabio and I discussed this elsewhere.  He only needs to use an API key in the 0x API, not in asset-swapper, so I ended up putting `apiKey` under `rfqt` after all.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Steve fixed this here: https://github.com/0xProject/0x-monorepo/pull/2541/commits/84adbcb683be847ea8217658c6edacdf85c2ceea#diff-951722c8cfc47f2ea3eebf81342ae48dR73

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Steve addressed this in #2552

feuGeneA:
@dekz 
> If possible I'd try and take one of the pre-existing Asset-swapper tests and test that nothing goes awry when the RFQT system is misconfigured. I.e test the user gets a swap when no MM quote endpoint can be reached.
> 
> Means we don't have to set up a golden path environment, but still get to exercise a large chunk of the logic around this.

We were able to get solid coverage of the new `QuoteRequestor` class (thanks @steveklebanoff!).  What we did not cover in this repo is any of the changes to `SwapQuoter`, which do have some complex conditional logic, but are otherwise relatively trivial.  However, we have integration tests in the 0x API repo that _do_ exercise these `SwapQuoter` changes.

I feel this PR is ready for a final review.  If you guys think any follow-up work is warranted, please be specific, and I'll be sure to track it.

packages/asset-swapper/src/utils/rfqt_mocker.ts
```diff
@@ -0,0 +1,37 @@
+import axios from 'axios';
```
dekz, line 1: should this be under `test/utils`?

feuGeneA: No because it's exported for external use.  Specifically, we're using it from within the 0x API tests.

packages/asset-swapper/package.json
```diff
@@ -53,6 +54,8 @@
...
+        "axios-mock-adapter": "^1.18.1",
```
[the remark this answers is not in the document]
feuGeneA: No because clients that use the exported mock need these deps.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -0,0 +1,116 @@
...
+    public async requestRfqtFirmQuotesAsync(
+        makerAssetData: string,
+        takerAssetData: string,
+        assetFillAmount: BigNumber,
+        marketOperation: MarketOperation,
+        takerApiKey: string,
+        takerAddress: string,
+        options?: Partial<RfqtFirmQuoteRequestOpts>,
+    ): Promise<SignedOrder[]> {
+        const { makerEndpointMaxResponseTimeMs } = _.merge({}, constants.DEFAULT_RFQT_FIRM_QUOTE_REQUEST_OPTS, options);
```
steveklebanoff, line 47: So you like lodash now, huh? 😆   If you're using it here, would prefer to use `flatten` as well 🤓 

I'm ok leaving as it though ❤️

packages/migrations/package.json
```diff
@@ -11,7 +11,7 @@
         "build": "tsc -b",
         "build:ci": "yarn build",
         "clean": "shx rm -rf lib ${npm_package_config_snapshot_name} ${npm_package_config_snapshot_name}-*.zip",
-        "lint": "tslint --format stylish --project .",
+        "lint": "tslint --format stylish --project . && prettier --check 'src/**/*.{ts,tsx,json,md}' --config ../../.prettierrc",
```
steveklebanoff, line 14: It seems like you agreed to keep `prettier` as a separate script in the `asset-swapper` package, can we do the same here?

steveklebanoff:
LGTM pending last nit: https://github.com/0xProject/0x-monorepo/pull/2541/files#r408991756

packages/migrations/package.json
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2541/commits/1da8f6887177629b9c6099dad84f8c348113a4ab

steveklebanoff:
Let's go 💪 💪 💪 !!!

PR #2542, opened 2020-04-08
Always use DFB in asset-swapper

dekz:
## Description

Use DFB even when there is only one source.

* Gives us a single entry point to monitor
* Allows burning of GST over all sources

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/orders.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/orders.ts
@@ -38,7 +38,7 @@ interface DexForwaderBridgeData {
     }>;
 }
 
-const dexForwarderBridgeDataEncoder = AbiEncoder.create([
+export const dexForwarderBridgeDataEncoder = AbiEncoder.create([
     { name: 'inputToken', type: 'address' },
     {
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -16,6 +16,7 @@ import * as _ from 'lodash';
```
dorothy-zbornak, line 16: It doesn't seem like we need any of the changes to this file?

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -38,7 +38,7 @@ interface DexForwaderBridgeData {
     }>;
 }
 
-const dexForwarderBridgeDataEncoder = AbiEncoder.create([
+export const dexForwarderBridgeDataEncoder = AbiEncoder.create([
```
dorothy-zbornak, line 41: So we don't need to export this either.

dekz: Yeah it's true, we seem to mostly disable the option in the tests

PR #2545, opened 2020-04-09
ZeroEx: TransformERC20, TokenSpender

dorothy-zbornak:
## Description

This adds the `TransformERC20` and `TokenSpender` feature to the ZeroEx contract. Please refer to the internal spec for background on these features.

transform-example (1)

This PR does *not* include the transformers themselves. Those are in #2576.

## Checklist:


contracts/erc20/contracts/src/v06/LibERC20TokenV06.sol
```diff
@@ -0,0 +1,212 @@
...
+*/
+
+pragma solidity ^0.6.5;
+
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-utils/contracts/src/v06/LibBytesV06.sol";
+import "./IERC20TokenV06.sol";
+
+
+library LibERC20TokenV06 {
```
dorothy-zbornak, line 26: This library has significant changes from the original (0.5) version.

contracts/zero-ex/contracts/src/features/IPuppetPool.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /// @dev Create a new, free puppet to add to the pool. Anyone can call this.
+    /// @return puppet The new puppet's address.
+    function createFreePuppet() external returns (address puppet);
+
+    /// @dev Acquire a new puppet instance. This removes the puppet from the
+    ///      pool. If one is not available, a new one will be deployed.
+    ///      Only callable from within.
+    /// @return puppet The acquired puppet.
+    function _acquirePuppet() external returns (IPuppet puppet);
```
dorothy-zbornak, line 40: Proposing prefixing "internal" registered functions with `_`.

contracts/zero-ex/contracts/src/features/ITransformERC20.sol
```diff
@@ -0,0 +1,103 @@
...
+        address payable taker,
+        IERC20TokenV06 inputToken,
+        IERC20TokenV06 outputToken,
+        uint256 inputTokenAmount,
+        uint256 minOutputTokenAmount,
+        Transformation[] calldata transformations
+    )
+        external
+        payable
+        returns (uint256 outputTokenAmount);
```
dorothy-zbornak, line 102: I expose this function (internally) so we can wrap it with friendlier market fill functions down the line.

dorothy-zbornak:
> You don't know the address of the puppet that will be acquired a priori, so wouldn't you need three txns to start using a puppet? (1) acquire a puppet, (2) set allowance, (3) execute transformation transaction
> 
> It's also kinda sketch that you could:
> 
> 1. Acquire a puppet
> 2. Set allowances on it
> 3. Release it, forgetting to remove allowances on it
> 4. An attacker acquires that puppet and steals funds
> 
> Can we use `create2` instead? Instead of a common pool of puppets, each user address gets a unique puppet, deployed by using `create2` salted with their address. Since now you'd know the address a priori, you can reduce the 3 txns above to 2, and we remove the risk of sharing puppets.

No allowances get set on the puppets from puppet pool. Those puppets get physically transferred tokens and hold them for the duration of the swap. The only puppet that gets allowances is the one owned by TokenSpender, which is permanent and known, and only used to transfer from taker to another address.

contracts/zero-ex/contracts/src/features/SimpleFunctionRegistry.sol
```diff
@@ -55,16 +55,15 @@ contract SimpleFunctionRegistry is
     }
 
     /// @dev Initializes this feature.
-    /// @param impl The actual address of this feature contract.
     /// @return success Magic bytes if successful.
-    function bootstrap(address impl) external returns (bytes4 success) {
```
dorothy-zbornak, line 60: Dunno why I was passing in `impl` when `immutable _implementation` was a thing.

packages/utils/src/revert_error.ts
```diff
@@ -9,7 +9,16 @@ import { BigNumber } from './configured_bignumber';
 
 // tslint:disable: max-classes-per-file
 
-type ArgTypes = string | BigNumber | number | boolean | BigNumber[] | string[] | number[] | boolean[];
```
dorothy-zbornak, line 12: Add `Numberish` array support.

packages/web3-wrapper/package.json
```diff
@@ -47,7 +47,7 @@
         "chai-as-promised": "^7.1.0",
         "chai-bignumber": "^3.0.0",
         "dirty-chai": "^2.0.1",
-        "ganache-core": "^2.9.0-istanbul.0",
+        "ganache-core": "^2.10.2",
```
dorothy-zbornak, line 50: Fixes a weird invalid opcode bug.

contracts/zero-ex/contracts/src/migrations/InitialMigration.sol
```diff
@@ -47,8 +53,13 @@ contract InitialMigration {
     ///      transfers ownership to `owner`, then self-destructs.
     ///      Only callable by `deployer` set in the contstructor.
     /// @param owner The owner of the contract.
+    /// @param features Features to bootstrap into the proxy.
     /// @return zeroEx The deployed and configured `ZeroEx` contract.
-    function deploy(address payable owner) public virtual returns (ZeroEx zeroEx) {
+    function deploy(address payable owner, BootstrapFeatures memory features)
```
dorothy-zbornak, line 58: So now the migrator does not deploy the feature contracts for you. Turns out some of these are quite large.

contracts/zero-ex/contracts/src/external/Puppet.sol
```diff
@@ -0,0 +1,155 @@
...
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-utils/contracts/src/v06/OwnableV06.sol";
+import "../errors/LibPuppetRichErrors.sol";
+import "./IPuppet.sol";
+
+
+/// @dev A contract that can execute arbitrary calls from its owner.
+contract Puppet is
+    IPuppet,
+    OwnableV06
```
dorothy-zbornak, line 31: Probably going to ditch `Ownable` and storing the owner in an `immutable` so this contract is 100% stateless.

dorothy-zbornak: I did it.

contracts/zero-ex/contracts/src/external/Puppet.sol
```diff
@@ -0,0 +1,175 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-utils/contracts/src/v06/errors/LibRichErrorsV06.sol";
+import "@0x/contracts-utils/contracts/src/v06/errors/LibOwnableRichErrorsV06.sol";
+import "../errors/LibPuppetRichErrors.sol";
+import "./IPuppet.sol";
+
+
+/// @dev A contract that can execute arbitrary calls from its owner.
+contract Puppet is
```
hysz, line 29: I'm curious to hear what other protocologists think, but the terminology has been tripping me up a bit. It's definitely clever, but I wonder if it's really necessary. The puppet and token spender are basically just implementing the proxy pattern. I might consider more familiar (albeit boring) names like `CallProxy` and `UniversalAssetProxy`.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+
+/// @dev Feature to composably transform between ERC20 tokens.
+contract TransformERC20 is
+    IFeature,
+    ITransformERC20,
+    FixinCommon
+{
+    // solhint-disable const-name-snakecase
+    /// @dev Name of this feature.
+    string constant public override FEATURE_NAME = "TransformERC20";
```
hysz, line 48: Do we have tests for accidental duplicate feature names? I'm manually checking, but I could easily see this getting overlooked.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+    /// @return deployer The transform deployer address.
+    function getTransformerDeployer()
+        external
+        override
+        view
+        returns (address deployer)
+    {
+        return transformDeployer;
+    }
+
```
hysz, line 93: Nit double whitespace.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+        uint256 minOutputTokenAmount,
+        Transformation[] memory transformations
+    )
+        public
+        payable
+        returns (uint256 outputTokenAmount)
+    {
+        // If the input token amount is -1, transform the taker's entire
+        // spendable balance.
+        if (inputTokenAmount == uint256(-1)) {
```
hysz, line 204: We should also document this special value in the description, both in this fn and above.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+        // Pull input tokens from the taker to the puppet and transfer attached ETH.
+        _transferInputTokensAndAttachedEth(inputToken, taker, address(puppet), inputTokenAmount);
+
+        // Perform transformations.
+        for (uint256 i = 0; i < transformations.length; ++i) {
+            _executeTransformation(puppet, transformations[i], taker, callDataHash);
+        }
+
+        // Ensure enough output token has been sent to the taker.
+        if (minOutputTokenAmount != 0) {
```
hysz, line 224: We should also document this special value in the description, both in this fn and above.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+            // Compute how much output token has been transferred to the taker.
+            uint256 takerOutputTokenBalanceAfter =
+                LibERC20Transformer.getTokenBalanceOf(outputToken, taker);
+            if (takerOutputTokenBalanceAfter > takerOutputTokenBalanceBefore) {
+                outputTokenAmount = takerOutputTokenBalanceAfter.safeSub(
```
hysz, line 229: It would seem to me that we would always want to compute `outputTokenAmount`, no?

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+                    takerOutputTokenBalanceBefore
+                );
+            }
```
hysz, line 232: Should we revert if `takerOutputTokenBalanceBefore <= takerOutputTokenBalanceAfter`? It seems like this would be unexpected behavior.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+        override
+        view
+        returns (IPuppet puppet)
+    {
+        return LibTransformERC20Storage.getStorage().puppet;
+    }
+
+    /// @dev Transfer input tokens from the taker and any attached ETH to `to`
+    /// @param inputToken The token to pull from the taker.
+    /// @param taker The taker address.
```
hysz, line 266: It's sort of confusing having a `taker` in this context. Perhaps just `from` and `to`?

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+        private
+    {
+        // Transfer any attached ETH.
+        if (msg.value != 0) {
+            to.transfer(msg.value);
+        }
+        // Transfer input tokens.
+        if (!LibERC20Transformer.isTokenETH(inputToken)) {
+            // Token is not ETH, so pull the ERC20 from the taker.
+            ITokenSpender(address(this))._spendERC20Tokens(
```
hysz, line 284: I might consider using the conventional `transferFrom` name here.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,365 @@
...
+                inputToken,
+                taker,
+                to,
+                amount
+            );
+        } else {
+            // Token is ETH, so the caller must attach enough ETH to the call.
```
hysz, line 291: Nit - squash into an `else if`

contracts/zero-ex/contracts/src/external/IPuppet.sol
```diff
@@ -0,0 +1,61 @@
...
+
+/// @dev A contract that can execute arbitrary calls from its owner.
+interface IPuppet {
+
+    /// @dev Execute an arbitrary call. Only an authority can call this.
+    /// @param target The call target.
+    /// @param callData The call data.
+    /// @param value Ether to attach to the call.
+    /// @return resultData The data returned by the call.
+    function execute(
```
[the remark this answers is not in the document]
dorothy-zbornak: Isn't `call` a reserved keyword?

contracts/zero-ex/contracts/src/external/Puppet.sol
```diff
@@ -0,0 +1,175 @@
...
+    IPuppet
+{
+    // solhint-disable no-unused-vars,indent,no-empty-blocks
+    using LibRichErrorsV06 for bytes;
+
+    // solhint-disable
+    /// @dev Store the owner/deployer as an immutable to make this contract stateless.
+    address public override immutable owner;
```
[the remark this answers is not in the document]
dorothy-zbornak: Features don't have owners. The proxy does. Using an immutable would mean you couldn't transfer ownership, which would make migrations awkward since they rely on this mechanic.

contracts/zero-ex/contracts/src/features/ITransformERC20.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// @dev Defines a transformation to run in `transformERC20()`.
+    struct Transformation {
+        // The transformation handler.
+        // Can receive the entire balance of `tokens`.
+        IERC20Transformer transformer;
+        // RLP-encoded deployment nonce of the deployer when `transformer`
+        // was deployed.
+        // This is used to verify that the transformer was deployed
+        // by a contract we trust.
+        bytes rlpNonce;
```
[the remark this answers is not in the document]
dorothy-zbornak: The nonce isn't a known value. It increments for every contract. If a client screws it up, the tx just reverts.

contracts/zero-ex/contracts/src/features/TokenSpender.sol
```diff
@@ -0,0 +1,145 @@
...
+contract TokenSpender is
+    IFeature,
+    ITokenSpender,
+    FixinCommon
+{
+    // solhint-disable const-name-snakecase,indent
+    /// @dev Name of this feature.
+    string constant public override FEATURE_NAME = "TokenSpender";
+    /// @dev Version of this feature.
+    uint256 constant public override FEATURE_VERSION = (1 << 64) | (0 << 32) | (0);
```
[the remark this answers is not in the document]
dorothy-zbornak: I think I tried this and found out solidity doesn't support functions in initializers.

contracts/zero-ex/contracts/src/features/TokenSpender.sol
```diff
@@ -0,0 +1,145 @@
...
+        address owner,
+        address to,
+        uint256 amount
+    )
+        external
+        override
+        onlySelf
+    {
+        IAllowanceTarget spender = LibTokenSpenderStorage.getStorage().allowanceTarget;
+        // Have the puppet spender execute an ERC20 `transferFrom()`.
```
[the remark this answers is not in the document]
dorothy-zbornak: This comment is old. It shouldn't be calling it a puppet. AllowanceTarget is an entirely different contract.

contracts/zero-ex/contracts/src/features/TokenSpender.sol
```diff
@@ -0,0 +1,145 @@
...
+                address(token),
+                abi.encodeWithSelector(
+                    IERC20TokenV06.transferFrom.selector,
+                    owner,
+                    to,
+                    amount
+                )
+            )
+        );
+        if (didSucceed) {
```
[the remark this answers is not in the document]
dorothy-zbornak: It's the raw revert data.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: No, but naming and versioning features isn't essential to anything right now.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: It'll stay at zero if the balance does not increase. We can't represent decreases with a uint256.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: Do you mean after < before? This should be caught in the final assertion if the taker expects a non-zero increase.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: That's kind of assuming that TokenSpender will only ever support ERC20s, right? I was thinking it would be upgraded to support other token types so I chose a descriptive name.

contracts/zero-ex/contracts/src/external/Puppet.sol
dorothy-zbornak: Technically the puppet can do a regular call as well. We don't use it right now though so we might as well drop that feature and go with your naming convention.

contracts/erc20/contracts/src/v06/IERC20TokenV06.sol
```diff
@@ -0,0 +1,95 @@
+/*
```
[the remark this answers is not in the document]
dorothy-zbornak: Feels bad bumping up the major version when I've only upgraded a handful of the contracts.

This would also freeze this contract package from the context of its dependents. Lerna will choose to install this package from npm for them. So we could never make changes to the previous version from within the monorepo and have it be reflected in its dependents, because it no longer technically exists in the monorepo.

contracts/zero-ex/contracts/src/external/Puppet.sol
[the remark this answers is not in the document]
dorothy-zbornak: Puppet isn't a feature, if that's a point of confusion. It's a standalone contract deployed by the TransformERC20 feature.

contracts/zero-ex/contracts/src/external/IPuppet.sol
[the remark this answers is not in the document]
hysz: Doesn't appear to be, just verified in Remix.

contracts/zero-ex/contracts/src/features/TokenSpender.sol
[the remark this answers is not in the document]
hysz: Just verified in Remix, you just have to make the `FEATURE_VERSION` immutable instead of constant.

contracts/zero-ex/contracts/src/external/Puppet.sol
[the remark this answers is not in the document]
hysz: Ah, gotcha. Makes sense for the proxy. If we do eventually create `onlyOwner` functions in features I could be in favor of this pattern, but suppose we'll x that bridge when we come to it 🤓.

contracts/zero-ex/contracts/src/features/ITransformERC20.sol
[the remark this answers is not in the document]
hysz: Yeah I was thinking along the lines of hardcoding the unique nonce into each contract before it's deployed. Then having an automatic check before or after the call, either by querying the transformer or standardizing the return data. Ex:
```
require(
    isZeroExTransformer(transformer.getNonce()),
    "ONLY_ZERO_EX_TRANSFORMERS"
);
transformer.transform(...);
```
OR
```
returnData = transformer.transform(...);
require(
    isZeroExTransformer(returnData.getNonce()),
    "ONLY_ZERO_EX_TRANSFORMERS"
);
```

contracts/erc20/contracts/src/v06/IERC20TokenV06.sol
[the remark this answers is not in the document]
hysz: Thanks, yeah that does sound kind of hectic. Another option could be to keep this in a branch separate from `development` until the transition is complete, similar to how we used to do for new versions of the Exchange. What do you think, would that be more or less cumbersome?

contracts/zero-ex/contracts/src/features/TokenSpender.sol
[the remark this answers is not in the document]
hysz: Yeah, honestly not sure where I was going with this question in retrospect. Makes sense, carry on lol.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
hysz: For some more context, suppose `minOutputTokenAmount == 0`. Then the user doesn't mind how much they receive, but wouldn't we still want to emit and return the value? Additionally, if `takerOutputTokenBalanceAfter < takerOutputTokenBalanceBefore` then the `safeSub` will just revert.

dorothy-zbornak: Yeah, that makes sense. We probably still want to handle a negative change because SafeMath rich reverts are obnoxiously opaque.

contracts/zero-ex/contracts/src/features/TokenSpender.sol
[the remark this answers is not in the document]
dorothy-zbornak: Nice!

contracts/erc20/contracts/src/v06/IERC20TokenV06.sol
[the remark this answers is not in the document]
dorothy-zbornak: idk, I think that worked for 3.0 because we weren't immediately productionizing the changes. We'd be folding this stuff into asset-swapper and other packages pretty much right away so we'd be looking for a solution in 2 weeks again.

contracts/zero-ex/contracts/src/features/ITransformERC20.sol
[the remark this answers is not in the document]
dorothy-zbornak: I like the latter. One less call. Just returning the nonce alone should be sufficient to indicate success since the RLP encoding is never empty and always unique.

contracts/zero-ex/contracts/src/external/IPuppet.sol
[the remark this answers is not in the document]
dorothy-zbornak: Do we actually want this close resemblance? Semantically they're different and that could lead to confusion.
```solidity
// Call the `call()` function on `addr`.
IPuppet(addr).call(target, callData);
// vs
// Call a raw payload on `addr`
addr.call(callData);
```

And it's even more nuanced with `delegatecall`:
```solidity
// Call the `delegatecall()` function on `addr`.
// Executes in the context of `addr`.
IPuppet(addr).delegatecall(target, callData)
// vs
// Execute the raw `callData` in the context of `this`.
addr.delegatecall(callData)
```

contracts/zero-ex/contracts/src/external/Puppet.sol
mzhu25: I kinda prefer FlashWallet? (flash loans so hot right now)

contracts/zero-ex/contracts/src/features/TransformERC20.sol
mzhu25: I kinda like having the option of not querying balances twice, e.g. for (un)wrapping WETH it'd be a waste of gas to do so

contracts/zero-ex/contracts/src/external/IPuppet.sol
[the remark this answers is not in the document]
mzhu25: how about `execute` and `delegateExecute`?

contracts/zero-ex/contracts/src/external/Puppet.sol
dorothy-zbornak: I like FlashWallet

contracts/zero-ex/contracts/src/features/TokenSpender.sol
```diff
@@ -0,0 +1,145 @@
...
+import "../migrations/LibMigrate.sol";
+import "../external/IAllowanceTarget.sol";
+import "../storage/LibTokenSpenderStorage.sol";
+import "./ITokenSpender.sol";
+import "./IFeature.sol";
+import "./ISimpleFunctionRegistry.sol";
+
+
+/// @dev Feature that allows spending token allowances.
+contract TokenSpender is
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, I'm thinking we'll just update this feature and keep the same allowance target.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: I do too, but I also think it's pretty useful for this function to return the amount transferred regardless.

contracts/zero-ex/contracts/src/external/IPuppet.sol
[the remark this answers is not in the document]
dorothy-zbornak: Let's go with `executeCall()` and `executeDelegateCall()` so everyone is left equally unimpressed.

contracts/zero-ex/test/features/token_spender_test.ts
```diff
@@ -0,0 +1,154 @@
...
+            ),
+        });
+        feature = new TokenSpenderContract(zeroEx.address, env.provider, env.txDefaults, abis);
+        token = await TestTokenSpenderERC20TokenContract.deployFrom0xArtifactAsync(
+            artifacts.TestTokenSpenderERC20Token,
+            env.provider,
+            env.txDefaults,
+            artifacts,
+        );
+        allowanceTarget = new IFlashWalletContract(
```
mzhu25, line 44: this should be `IAllowanceTargetContract`, though I think we only need the address

contracts/zero-ex/test/features/token_spender_test.ts
```diff
@@ -0,0 +1,154 @@
...
+            const tokenOwner = randomAddress();
+            await token
+                .setBalanceAndAllowanceOf(tokenOwner, balance, allowanceTarget.address, allowance)
+                .awaitTransactionSuccessAsync();
+            const spendableBalance = await feature.getSpendableERC20BalanceOf(token.address, tokenOwner).callAsync();
+            expect(spendableBalance).to.bignumber.eq(BigNumber.min(balance, allowance));
+        });
+    });
+
+    describe('getAllowanceTarget()', () => {
```
mzhu25, line 148: this test seems unnecessary

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -0,0 +1,371 @@
...
+        }
+
+        // Compute how much output token has been transferred to the taker.
+        uint256 takerOutputTokenBalanceAfter =
+            LibERC20Transformer.getTokenBalanceOf(outputToken, taker);
+        if (takerOutputTokenBalanceAfter > takerOutputTokenBalanceBefore) {
+            outputTokenAmount = takerOutputTokenBalanceAfter.safeSub(
+                takerOutputTokenBalanceBefore
+            );
+        }
```
hysz, line 244: I wonder if we should revert if the final balance is less than the initial balance? What's a scenario where the user would want this?

dorothy-zbornak: Yeah, I can't really think of a compelling scenario.

dorothy-zbornak: We now revert.

PR #2547, opened 2020-04-10
[WIP] RFQT Mockable Axios

steveklebanoff:
First pass on creating an easy way (which we can export) to mock out RFQT requests.

Still a WIP as this may change with actually using it as part of 0x API.  Also I may change the interface of `withMockedRfqtRequests` as it may not be the most natural approach.

### DIFF
--- a/packages/asset-swapper/src/utils/quote_requestor.ts
+++ b/packages/asset-swapper/src/utils/quote_requestor.ts
@@ -24,6 +24,11 @@ export class QuoteRequestor {
         takerApiKey: string,
         takerAddress: string,
     ): Promise<SignedOrder[]> {
+        if (marketOperation !== MarketOperation.Sell) {
+            // Only sell operations are supported by RFQT makers right now
+            return [];
+        }
+
         const makerEndpointMaxResponseTimeMs = 1000;
 
         const getTokenAddressOrThrow = (assetData: string): string => {
@@ -57,8 +62,7 @@ export class QuoteRequestor {
                     params: {
                         sellToken,
                         buyToken,
-                        buyAmount: marketOperation === MarketOperation.Buy ? assetFillAmount.toString() : undefined,
-                        sellAmount: marketOperation === MarketOperation.Sell ? assetFillAmount.toString() : undefined,
+                        sellAmount: marketOperation === assetFillAmount.toString(),
                         takerAddress,
                     },
                     timeout: makerEndpointMaxResponseTimeMs,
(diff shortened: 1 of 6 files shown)


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -24,6 +24,11 @@ export class QuoteRequestor {
         takerApiKey: string,
         takerAddress: string,
     ): Promise<SignedOrder[]> {
+        if (marketOperation !== MarketOperation.Sell) {
```
steveklebanoff, line 27: This was helpful so I didn't have to deal with a null `takerAmount` param in the tests.  I think this should be done regardless though.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -57,17 +62,15 @@ export class QuoteRequestor {
...
                     },
                     timeout: makerEndpointMaxResponseTimeMs,
                 }).catch(err => {
                     logUtils.warn(
-                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}: ${JSON.stringify(
-                            err,
-                        )}`,
+                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}`,
                     );
+                    logUtils.warn(err);
```
steveklebanoff, line 73: `JSON.stringify(err)` was just outputting `{}` so I needed to add this to see the actual error

packages/asset-swapper/test/utils/rfqt_mocker.ts
```diff
@@ -0,0 +1,65 @@
...
+    takerAddress: string;
+}
+export interface MockedSuccessfulRfqtRequest {
+    endpoint: string;
+    expectedApiKey: string;
+    expectedParams: RfqtMakerParams;
+    response: SignedOrder;
+}
+export const rfqtMocker = {
+    withMockedRfqtRequests: async (
```
steveklebanoff, line 32: Instead of creating all this abstraction, perhaps I should let the consumer of this function mock any arbitrary URL, param and header combo with any arbitrary data

steveklebanoff:
Replaced by https://github.com/0xProject/0x-monorepo/compare/rfq-t...rfq-t-mock-axios-generic-mock

PR #2549, opened 2020-04-10
asset-swapper: Mockable axios for QuoteRequestor

steveklebanoff:
* Tests `QuoteRequestor` using a mock-axios-adapter.
* Exports a way for other packages to mock-out RFQT requests


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -59,19 +64,16 @@ export class QuoteRequestor {
...
                         timeout: makerEndpointMaxResponseTimeMs,
                     });
                 } catch (err) {
                     logUtils.warn(
-                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}: ${JSON.stringify(
-                            err,
-                        )}`,
+                        `Failed to get RFQ-T quote from market maker endpoint ${rfqtMakerEndpoint} for API key ${takerApiKey} for taker address ${takerAddress}`,
                     );
+                    logUtils.warn(err);
```
steveklebanoff, line 76: `JSON.stringify(err)` which was used in the code outputted `{}` for an actual error.  I needed to add this line to actually see any error thrown

PR #2550, opened 2020-04-13
[WIP] Add state override set parameters to eth_call

opaolini:
## Description

These changes add support for optional `state override set` parameter in [go-ethereum](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set) that allows to ephemerally override state prior to executing `eth_call`.

## Testing instructions

Ganache does not support `state override` - https://github.com/trufflesuite/ganache-core/issues/554, nor does Parity.

These changes have been tested manually and with a simple `web3-wrapper` test against AlchemyAPI (their `v2` geth endpoint and rinkeby, as described [here](https://geth.ethereum.org/docs/rpc/ns-eth#override-example) ) as well as making sure the call does not break any current tests.

## Types of changes

  * Added `StateOverrideSet` to `@0x/ethereum-types`
  * Modified `@0x/web3-wrapper` to accept `StateOverrideSet` as an optional third parameter
  * Modified `@0x/abi-gen` to add an optional third parameter to `callAsync`
  * Modified `@0x/base-contracts`

## Checklist:

### DIFF
--- a/packages/abi-gen/README.md
+++ b/packages/abi-gen/README.md
@@ -143,7 +143,7 @@ Run unit tests and check diffs of generated wrappers vs known wrappers:
 yarn test_cli
 ```
 
-Known-good wrappers have been previously committed and are kept in `test-cli/expected-output/{language}`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/output/{language}`, and then manually copy it to `test-cli/expected-output/{language}`.
+Known-good wrappers have been previously committed and are kept in `test-cli/output/{language}`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/output/{language}`. Currently the `yarn test` script diffs the `test-cli/output/{language}` folder with the version previously committed exiting with an error if it doesnt match. Therefore if the output has changed you will need need to commit the changes first.
 
 Run linters against generated code:
 
--- a/packages/abi-gen/README.md
+++ b/packages/abi-gen/README.md
@@ -143,7 +143,7 @@ Run unit tests and check diffs of generated wrappers vs known wrappers:
 yarn test_cli
 ```
 
-Known-good wrappers have been previously committed and are kept in `test-cli/output/{language}`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/output/{language}`. Currently the `yarn test` script diffs the `test-cli/output/{language}` folder with the version previously committed exiting with an error if it doesnt match. Therefore if the output has changed you will need need to commit the changes first.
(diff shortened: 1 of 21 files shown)


packages/abi-gen/templates/Python/partials/method_class.handlebars
```diff
@@ -26,7 +26,7 @@ class {{toPythonClassname this.languageSpecificName}}Method(ContractMethod):
         return ({{> params }})
 
     {{/if}}
-    def call(self, {{#if inputs}}{{> typed_params inputs=inputs}}, {{/if}}tx_params: Optional[TxParams] = None) -> {{> call_return_type outputs=outputs type='call'~}}:
+    def call(self, {{#if inputs}}{{> typed_params inputs=inputs}}, {{/if}}tx_params: Optional[TxParams] = None, state_override: Optional[StateOverride] = None) -> {{> call_return_type outputs=outputs type='call'~}}:
```
fabioberger, line 29: Ah cool, didn't know you were also adding it to the Python wrappers. Cool.

Could we potentially just pass a `custom_params` param in, which is an object with a single key to start (`state_override`)? Just thinking that we might as well future-proof this so if we need to pass in additional optional params to `call` in the future, we don't need to introduce a breaking change. thoughts?

opaolini: So it turns out that the Python web3 library does not allow to pass in the optional third parameter to call [0], therefore I removed `StateOverride` from the python templates altogether.

We could probably add it, but that would require calling `self.web3.manager.request_blocking` directly [1].

Should I add support in Python for this?

[0] - https://web3py.readthedocs.io/en/stable/web3.eth.html
[1] - https://github.com/ethereum/web3.py/blob/master/web3/eth.py#L428

packages/abi-gen/README.md
```diff
@@ -143,7 +143,7 @@ Run unit tests and check diffs of generated wrappers vs known wrappers:
 yarn test_cli
 ```
 
-Known-good wrappers have been previously committed and are kept in `test-cli/expected-output/{language}`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/output/{language}`, and then manually copy it to `test-cli/expected-output/{language}`.
+Known-good wrappers have been previously committed and are kept in `test-cli/output/{language}`. They are intended to provide sample output and should be kept in sync with the generating code. When making changes to this project or `@0x/abi-gen-templates`, run `yarn test_cli:prebuild` to generate fresh code into `test-cli/output/{language}`. Currently the `yarn test` script diffs the `test-cli/output/{language}` folder with the version previously committed exiting with an error if it doesnt match. Therefore if the output has changed you will need need to commit the changes first.
```
fabioberger, line 146: 🙏

packages/contract-wrappers-test/test/web3_wrapper_integration_test.ts
```diff
@@ -0,0 +1,74 @@
...
+import { chaiSetup } from './utils/chai_setup';
+chaiSetup.configure();
+
+const { expect } = chai;
+
+describe('Web3Integration tests', () => {
+    const providerEngine = new Web3ProviderEngine();
+    const providerURL = process.env.ETH_RPC_URL_RINKEBY;
+    if (providerURL === undefined) {
+        return;
```
fabioberger, line 18: Perhaps print out that this test is being skipped?

PR #2552, opened 2020-04-14
RFQ-T: validate assetData & add more tests

steveklebanoff:
* Validates taker and maker asset data
* Validates signedOrderSchema instead of just orderSchema
* Adds some more logs
* Adds some more tests

### DIFF
--- a/packages/asset-swapper/src/utils/quote_requestor.ts
+++ b/packages/asset-swapper/src/utils/quote_requestor.ts
@@ -83,7 +83,7 @@ export class QuoteRequestor {
         const ordersWithStringInts = responses.map(response => response.data); // not yet BigNumber
 
         const validatedOrdersWithStringInts = ordersWithStringInts.filter(order => {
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -83,9 +82,22 @@ export class QuoteRequestor {
 
         const ordersWithStringInts = responses.map(response => response.data); // not yet BigNumber
 
-        const validatedOrdersWithStringInts = ordersWithStringInts.filter(order =>
-            this._schemaValidator.isValid(order, schemas.orderSchema),
-        );
+        const validatedOrdersWithStringInts = ordersWithStringInts.filter(order => {
+            const hasValidSchema = this._schemaValidator.isValid(order, schemas.signedOrderSchema);
```
steveklebanoff, line 86: Note that this changed from `orderSchema` to `signedOrderSchema`

PR #2555, opened 2020-04-16
asset-swapper: RFQ-T indicative quotes

feuGeneA:
When a request specifies a whitelisted API key, and a taker address, and does _not_ specify `intentToFill=true`, the `SwapQuoter` passes its `QuoteRequestor` into `MarketOperationUtils`, so that it can call `QuoteRequestor.getRfqtIndicativeQuotesAsync()`, which it transforms into weird pseudo-native-orders, which in turn get fed into order routing along with all of the other native orders.

This code is not yet covered with tests here, but is being exercised in the end-to-end test in https://github.com/0xProject/0x-api/pull/172, which is integrated with CI there, and which is living at https://github.com/0xProject/0x-api/pull/172/commits/66c9d9dd3aa71b96b2a3b85000f9c0a44bc0bffd#diff-491d90fb8eb25e485b3d8c9429e6754cR326.

-   [x] Add tests
-   [x] Updated CHANGELOGs


packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -150,6 +163,54 @@ function dexQuotesToPaths(
...
+        // Skip orders with rates that are <= 0.
+        if (adjustedRate.lte(0)) {
+            continue;
+        }
+        path.push({
+            input,
+            output,
+            rate,
+            adjustedRate,
+            adjustedOutput: output,
```
dekz, line 194: Is there a reason this isn't a calculated `adjustedOutput`? Basically we adjust the `output` by the cost it takes to fill it. Without it then RFQT quotes are basically free.

Penalty should be the same as a normal native order.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -92,10 +88,25 @@ export class MarketOperationUtils {
                 this._liquidityProviderRegistry,
             ),
         );
+        const rfqtPromise =
+            opts !== undefined && _opts.rfqt !== undefined && _opts.rfqt.quoteRequestor !== undefined
```
dekz, line 92: Since we merge defaults into `_opts`, and never access it further down, do we need the `opts !== undefined` check?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -16,7 +16,9 @@ export const samplerOperations = {
                     .getABIEncodedTransactionData();
             },
             handleCallResultsAsync: async (contract, callResults) => {
-                return contract.getABIDecodedReturnData<BigNumber[]>('getOrderFillableTakerAssetAmounts', callResults);
+                return contract
+                    .getABIDecodedReturnData<BigNumber[]>('getOrderFillableTakerAssetAmounts', callResults)
```
dekz, line 20: I'm assuming they're not signing the orders where the intent to fill is false? Rather than use `takerAddress` as a conditional can we determine this off the signature, or lack there of?

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -150,6 +163,54 @@ function dexQuotesToPaths(
     return paths;
 }
 
+function rfqtIndicativeQuotesToPath(
```
dorothy-zbornak, line 166: Why can't RFQT orders just be passed into `nativeOrdersToPath()`? Making them a separate source creates a lot of unnecessary work for the optimizer.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -150,6 +163,54 @@ function dexQuotesToPaths(
...
+            flags: 0,
+            index: 0, // TBD
+            parent: undefined, // TBD
+            source: ERC20BridgeSource.Rfqt,
+            fillData: { quote },
+        });
+    }
+    // Sort by descending adjusted rate.
+    path = path.sort((a, b) => b.adjustedRate.comparedTo(a.adjustedRate));
+    /* copy-pasted the following from nativeOrdersToPath(). not really sure i need it.
```
dorothy-zbornak, line 204: Yeah, you need to do this because the indexes and parents are all wrong.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -35,6 +35,7 @@ export const DEFAULT_GET_MARKET_ORDERS_OPTS: GetMarketOrdersOpts = {
     gasSchedule: {},
     allowFallback: true,
     shouldBatchBridgeOrders: true,
+    rfqt: undefined,
```
steveklebanoff, line 38: Is this change actually required? I would expect us not having to populate this key with `undefined` since it's defined as `rfqt?` https://github.com/0xProject/0x-monorepo/blob/6c62c3342ca21c89f6e3da527d0b0b83e4040033/packages/asset-swapper/src/utils/market_operation_utils/types.ts#L192

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
steveklebanoff: If we could just pass in the RFQT orders to `nativeOrdersToPath` like @dorothy-zbornak suggested it feels like that would solve this and make the code a lot simpler.  I'm not 100% confident on the implications here though

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -92,10 +88,25 @@ export class MarketOperationUtils {
...
+                ? _opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(
+                      nativeOrders[0].makerAssetData,
+                      nativeOrders[0].takerAssetData,
+                      takerAmount,
+                      MarketOperation.Sell,
+                      _opts.rfqt,
+                  )
+                : Promise.resolve([] as RfqtIndicativeQuoteResponse[]);
```
steveklebanoff, line 100: Can you do `Promise.resolve<RfqtIndicativeQuoteResponse[]>([])` instead?  Feels a bit cleaner than the typecasting done here

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -165,11 +171,26 @@ export class MarketOperationUtils {
                 this._liquidityProviderRegistry,
             ),
         );
+        const rfqtPromise =
```
steveklebanoff, line 174: At what point do we want to short-circuit considering RFQT MMers will not suppot buys to begin with?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -16,7 +16,9 @@ export const samplerOperations = {
...
+                    .map((amt, i) => (orders[i].takerAddress === NULL_ADDRESS ? amt : orders[i].takerAssetAmount));
```
steveklebanoff, line 21: Can you please add a comment in the code base explaining what's going on here?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -27,6 +34,40 @@ function getTokenAddressOrThrow(assetData: string): string {
...
+}
+
+function inferQueryParams(
+    marketOperation: MarketOperation,
+    makerAssetData: string,
+    takerAssetData: string,
+    assetFillAmount: BigNumber,
+): { buyToken: string; sellToken: string; buyAmount: string | undefined; sellAmount: string | undefined } {
+    if (marketOperation === MarketOperation.Buy) {
+        return {
```
steveklebanoff, line 55: Nit but it'd be nice to define the shared things at the top. 

Also I don't think we need the explicit `: undefined` definitions

```
const baseResult = { buyToken: getTokenAddressOrThrow(makerAssetData), sellToken:  getTokenAddressOrThrow(takerAssetData)};

if (marketOperation === MarketOperation.Buy) {
   return { ...baseResult, buyAmount: assetFillAmount.toString()};
else {
   return { ...baseResult, sellAmount: assetFillAmount.toString()};
}
```

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -40,35 +81,29 @@ export class QuoteRequestor {
...
+                        headers: { '0x-api-key': _opts.apiKey },
                         params: {
-                            sellToken,
-                            buyToken,
-                            buyAmount: marketOperation === MarketOperation.Buy ? assetFillAmount.toString() : undefined,
-                            sellAmount:
-                                marketOperation === MarketOperation.Sell ? assetFillAmount.toString() : undefined,
-                            takerAddress,
+                            takerAddress: _opts.takerAddress,
+                            ...inferQueryParams(marketOperation, makerAssetData, takerAssetData, assetFillAmount),
```
steveklebanoff, line 98: nice abstraction 👍

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -113,4 +148,81 @@ export class QuoteRequestor {
 
         return orders;
     }
+
+    public async requestRfqtIndicativeQuotesAsync(
```
steveklebanoff, line 152: Seems like a lot of duplicated code with `requestRfqtFirmQuotesAsync`. Any thoughts on abstracting some shared logic?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -113,4 +148,81 @@ export class QuoteRequestor {
...
+                ...response,
+                makerAssetAmount: new BigNumber(response.makerAssetAmount),
+                takerAssetAmount: new BigNumber(response.takerAssetAmount),
+            };
+        });
+
+        return responses;
+    }
+
+    private _isValidRfqtIndicativeQuoteResponse(response: RfqtIndicativeQuoteResponse): boolean {
```
steveklebanoff, line 212: I think this is OK for now, but it may be nice in the future to introduce a proper JSON `schema` file for this.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
feuGeneA: This is moot now.  This function has gone away.  Now the indicative quotes are transformed into native orders much earlier in the pipeline, before this method is used.

feuGeneA: See here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/d6d4d29257c4cb0810c280f8edc33939f0e6092e#diff-bdb1465a97651026c09ef607e5de9e6fR323

packages/asset-swapper/src/utils/market_operation_utils/index.ts
feuGeneA: This was a bug, that I noticed too, and fixed here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/d6d4d29257c4cb0810c280f8edc33939f0e6092e#diff-bdb1465a97651026c09ef607e5de9e6fR97

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
feuGeneA: I'm sorry, I wrote this commit while blindly following the spec I wrote.  In retrospect, this idea, of making the sampler bypass fillability checks because these orders are unsigned, is moot now.

This is because we're getting indicative quotes in parallel with the on-chain call, not prior to it (like we do for firm quotes), and we're piping the quotes directly into the order router, and there's no further fillability check after routing*, so there's no need for such a hack.

It was only ever consideration when we were trying to decide whether to request quotes from `SwapQuoter` or from `MarketOperationUtils`, but ultimately we decided to do both (for latency reasons), alleviating the need for this.

*: Of course there's the gas estimation / transaction validation stage that happens in the 0x-api side, but you'll see [there](https://github.com/0xProject/0x-api/commit/66c9d9dd3aa71b96b2a3b85000f9c0a44bc0bffd#diff-aa8361c6f3b8881b3948134ce5805f8eR53) that I'm forcing `skipValidation` to true when handling `/swap/price` requests and delegating them to the handler for `/swap/quote`.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
feuGeneA: They can. I was just hacking my way through it to get it under test, and then iterate on it.  I through out all of the stuff you're referring to.  See https://github.com/0xProject/0x-monorepo/pull/2555/commits/d6d4d29257c4cb0810c280f8edc33939f0e6092e

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
feuGeneA: It's not required, thank you for noticing it. I think I hadn't yet made it optional in a previous iteration of hacking on it.  I got rid of this.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
feuGeneA: My TypeScript ignorance precedes me.  Thank you for the idiom.  I've fixed this here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/d6d4d29257c4cb0810c280f8edc33939f0e6092e#diff-bdb1465a97651026c09ef607e5de9e6fR105

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
feuGeneA: I got rid of this commit.  Please see my explanation here: https://github.com/0xProject/0x-monorepo/pull/2555#discussion_r410040348

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Please clarify what you'd like to see moved to the top.  Is it the return value type of this function?

Regarding the `undefined`s, `tsc` complains if I remove them:

```
$ /home/gene/dev/0x-monorepo/node_modules/.bin/tsc -b
src/utils/quote_requestor.ts:55:9 - error TS2322: Type '{ buyToken: string; sellToken: string; buyAmount: string; }' is not assignable to type '{ buyToken: string; sellToken: string; buyAmount: string | undefined; sellAmount: string | undefined; }'.
  Property 'sellAmount' is missing in type '{ buyToken: string; sellToken: string; buyAmount: string; }'.

 55         return {
            ~~~~~~~~
 56             buyToken: getTokenAddressOrThrow(makerAssetData),
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
 58             buyAmount: assetFillAmount.toString(),
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 59         };
    ~~~~~~~~~~
```

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I think, at least in this iteration if not the last (I can't remember...), these functions are pretty distinct.  They're both limited in scope to dealing with the specific endpoints they hit, with the bulk of them coming from unpacking the response types, which differ in each case.  I do still see some patterns of duplication, but IMO they're limited.  Let me know what you think in this round.

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: Ah sorry for not being clear here -- my suggestion was for the `buyToken` and `sellToken` assignments to be shared at the top of the function. 

Also, my bad on the `undefined` note.  Here's my suggested change:

```ts
const baseResult = { buyToken: getTokenAddressOrThrow(makerAssetData), sellToken:  getTokenAddressOrThrow(takerAssetData)};

if (marketOperation === MarketOperation.Buy) {
   return { ...baseResult, buyAmount: assetFillAmount.toString(), sellAmount: undefined };
else {
   return { ...baseResult, buyAmount: undefined, sellAmount: assetFillAmount.toString()};
}
```

This is def a nit tho -- totally OK with me if you want to keep as-is

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -34,6 +35,7 @@ export enum ERC20BridgeSource {
     CurveUsdcDaiUsdtTusd = 'Curve_USDC_DAI_USDT_TUSD',
     LiquidityProvider = 'LiquidityProvider',
     CurveUsdcDaiUsdtBusd = 'Curve_USDC_DAI_USDT_BUSD',
+    Rfqt = 'Rfqt',
```
steveklebanoff, line 38: Is this used anywhere?

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -358,3 +359,32 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
...
+            fillableTakerAssetAmount: quote.takerAssetAmount,
+            makerAssetAmount: quote.makerAssetAmount,
+            takerAssetAmount: quote.takerAssetAmount,
+            makerAssetData: quote.makerAssetData,
+            takerAssetData: quote.takerAssetData,
+            takerAddress: NULL_ADDRESS,
+            makerAddress: NULL_ADDRESS,
+            senderAddress: NULL_ADDRESS,
+            feeRecipientAddress: NULL_ADDRESS,
+            salt: ZERO_AMOUNT, // generatePseudoRandomSalt(),
```
steveklebanoff, line 378: What's up with this comment? Can we leave as zero?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -92,10 +93,25 @@ export class MarketOperationUtils {
...
```
steveklebanoff, line 96: When would `_opts` ever be undefined?

Can make either make this ternary shorter, or break it out into something else?  This feels a bit jarring to read

One option (still a pretty long statement for ternary tho)
```ts
const rfqtIndicativeQuotes = _opts.rfqt && _opts.rfqt.quoteRequestor ? _opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(...) : Promise.resolve<RfqtIndicativeQuoteResponse[]>([]);
```

Another option
```ts
let rfqtIndicativeQuotesPromise = Promise.resolve<RfqtIndicativeQuoteResponse[]>([]);
if (_opts.rfqt && _opts.rfqt.quoteRequestor) {
   rfqtIndicativeQuotesPromise = _opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(...)
}
```

Third option (I like this best)
```ts
const rfqtIndicativeQuotes = async (): Promise<RfqtIndicativeQuoteResponse[]> => {
   if (_opts.rfqt && _opts.rfqt.quoteRequestor) {
       return _opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(...);
   }
  return [];
}
```

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -290,7 +318,10 @@ export class MarketOperationUtils {
         const paths = createFillPaths({
             side,
             // Augment native orders with their fillable amounts.
-            orders: createSignedOrdersWithFillableAmounts(side, opts.nativeOrders, opts.orderFillableAmounts),
+            orders: [
+                ...createSignedOrdersWithFillableAmounts(side, opts.nativeOrders, opts.orderFillableAmounts),
+                ...createSignedOrdersFromRfqtIndicativeQuotes(opts.rfqtIndicativeQuotes),
```
steveklebanoff, line 323: nice 🙌

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -358,3 +359,32 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
...
+            expirationTimeSeconds: new BigNumber(Math.floor(Date.now() / ONE_SECOND_MS) + ONE_HOUR_IN_SECONDS),
```
steveklebanoff, line 379: In a future revision I def think we should expect an expirationTimeSeconds from the MMer and use that here

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -358,3 +359,32 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
...
+            makerFeeAssetData: NULL_BYTES,
+            takerFeeAssetData: NULL_BYTES,
+            makerFee: ZERO_AMOUNT,
+            takerFee: ZERO_AMOUNT,
+            fillableTakerFeeAmount: ZERO_AMOUNT,
+            signature: WALLET_SIGNATURE,
+            chainId: 0, // HACK !!!!!!!!! how can we get at this from this context?
```
steveklebanoff, line 386: I think you can pass it in from `MarketOperationUtils`'s `this._orderDomain` but will let @dekz or @dorothy-zbornak chime in here if there's a better way

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -358,3 +359,32 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
...
+            exchangeAddress: NULL_ADDRESS, // HACK !!!!!!!!! how can we get at this from this context?
```
steveklebanoff, line 387: I think you can pass it in from MarketOperationUtils's `this.contractAddresses`

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -358,3 +359,32 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
...
+    quotes: RfqtIndicativeQuoteResponse[],
+): SignedOrderWithFillableAmounts[] {
+    return quotes.map(quote => {
+        return {
+            fillableMakerAssetAmount: quote.makerAssetAmount,
+            fillableTakerAssetAmount: quote.takerAssetAmount,
+            makerAssetAmount: quote.makerAssetAmount,
+            takerAssetAmount: quote.takerAssetAmount,
+            makerAssetData: quote.makerAssetData,
+            takerAssetData: quote.takerAssetData,
```
steveklebanoff, line 373: super nit (feel free to leave as-is) but for the values that we are passing over verbatim I'd do:

```
const { makerAssetAmount, takerAssetAmount , makerAssetData, takerAssetData } = quote;
return {
   makerAssetAmount,
   takerAssetAmount,
   makerAssetData,
   takerAssetData,
   <rest of obj>
}
```

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: Good point, they are more distinct then I realized on first pass -- i'm cool with leaving them as-is 👍

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: Oh good eye.  This method was basically copied from https://github.com/0xProject/0x-monorepo/blob/110e1afa8e088288fb920a1f5689fe3a7f7b754a/packages/asset-swapper/src/utils/market_operation_utils/orders.ts#L338 , and I changed it to zero here, but preserved that call in the comments just in case it ended up not working for some reason.  At this point I can say it's working, but I'd love some feedback from @dorothy-zbornak on whether we should be concerned about this.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -27,6 +34,40 @@ function getTokenAddressOrThrow(assetData: string): string {
     throw new Error(`Decoded asset data (${JSON.stringify(decodedAssetData)}) does not contain a token address`);
 }
 
+function assertTakerAddressOrThrow(takerAddress: string | undefined): void {
+    if (
+        takerAddress === undefined ||
+        takerAddress === '' ||
```
dorothy-zbornak, line 40:
```suggestion
        !takerAddress ||
```

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
dorothy-zbornak, line 387: Neither of these really matter since these orders aren't valid anyway, so I'm fine with leaving it as-is and making the comments less frightening lol

packages/asset-swapper/src/utils/market_operation_utils/types.ts
feuGeneA: No, it's not.  It's a remnant from the first draft.  I've removed it here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/4cc9ceabd21edecc28a430973d90f7ad96c0a2a0

packages/asset-swapper/src/utils/market_operation_utils/index.ts
feuGeneA: This is my attempt at making it look nicer: https://github.com/0xProject/0x-monorepo/pull/2555/commits/5effc6ec90e5a90402e0efb01cefd0b6a6a8ef1b

packages/asset-swapper/src/utils/market_operation_utils/index.ts
feuGeneA: I cleaned this up here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/5effc6ec90e5a90402e0efb01cefd0b6a6a8ef1b

Please let me know what you think.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -165,11 +176,26 @@ export class MarketOperationUtils {
                 this._liquidityProviderRegistry,
             ),
         );
+        const rfqtPromise =
+            opts !== undefined && _opts.rfqt !== undefined && _opts.rfqt.quoteRequestor !== undefined
```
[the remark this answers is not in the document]
feuGeneA: Fixed by consolidating this crazy conditional with the other instance of it here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/5effc6ec90e5a90402e0efb01cefd0b6a6a8ef1b

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: I've cleaned up this function here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/ad7868ebe10b902c04f34cd2cfdac64b72dfb8bc

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: I put it at zero for now: https://github.com/0xProject/0x-monorepo/pull/2555/commits/ad7868ebe10b902c04f34cd2cfdac64b72dfb8bc

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: I just set them both to zero for now, per https://github.com/0xProject/0x-monorepo/pull/2555#discussion_r411527107

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: I did just that: https://github.com/0xProject/0x-monorepo/pull/2555/commits/ad7868ebe10b902c04f34cd2cfdac64b72dfb8bc

Thank you

packages/asset-swapper/src/types.ts
```diff
@@ -188,16 +187,19 @@ export interface SwapQuoteOrdersBreakdown {
     [source: string]: BigNumber;
 }
 
+export interface RfqtRequestOpts {
+    takerAddress: string;
+    apiKey: string;
+    intentOnFilling?: boolean;
```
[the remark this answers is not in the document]
feuGeneA: I made this change here: https://github.com/0xProject/0x-monorepo/pull/2555#discussion_r411083497

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -92,10 +93,25 @@ export class MarketOperationUtils {
...
+        const rfqtPromise =
+            _opts !== undefined && _opts.rfqt !== undefined && _opts.rfqt.quoteRequestor !== undefined
+                ? _opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(
+                      nativeOrders[0].makerAssetData,
+                      nativeOrders[0].takerAssetData,
+                      takerAmount,
+                      MarketOperation.Sell,
+                      _opts.rfqt,
+                  )
+                : Promise.resolve<RfqtIndicativeQuoteResponse[]>([]);
```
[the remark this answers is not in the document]
feuGeneA: Great idea, thank you, did it here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/5effc6ec90e5a90402e0efb01cefd0b6a6a8ef1b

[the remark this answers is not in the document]
feuGeneA: I tried to remove it and got errors.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I made this change here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/83289bc801013e1b7d2bc0b6f5696d886be90784

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -566,19 +565,30 @@ export class SwapQuoter {
 
         let swapQuote: SwapQuote;
 
+        const calcOpts: CalculateSwapQuoteOpts = opts;
+        if (
+            // we should request indicative quotes:
+            calcOpts.rfqt &&
+            !calcOpts.rfqt.intentOnFilling &&
+            calcOpts.rfqt.apiKey &&
+            this._rfqtTakerApiKeyWhitelist.includes(calcOpts.rfqt.apiKey)
```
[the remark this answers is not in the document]
feuGeneA: I simplified this here: https://github.com/0xProject/0x-monorepo/pull/2555/commits/ba2ac6a7b5b362dd0b62179aa44921c4ad7ada75

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -27,6 +34,40 @@ function getTokenAddressOrThrow(assetData: string): string {
...
+        throw new Error('RFQ-T requires the presence of a taker address');
+    }
+}
+
+function inferQueryParams(
+    marketOperation: MarketOperation,
+    makerAssetData: string,
+    takerAssetData: string,
+    assetFillAmount: BigNumber,
+): { buyToken: string; sellToken: string; buyAmount: string | undefined; sellAmount: string | undefined } {
```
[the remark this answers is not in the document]
feuGeneA: I made this change here: https://github.com/0xProject/0x-monorepo/commit/2c97208e7455656866189a68cd0ff2cab3c64e74

steveklebanoff:
I feel good about all the outstanding changes I requested -- nice work!

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -26,6 +32,26 @@ import {
...
+    if (opts.rfqt && opts.rfqt.quoteRequestor) {
+        return opts.rfqt.quoteRequestor.requestRfqtIndicativeQuotesAsync(
+            makerAssetData,
+            takerAssetData,
+            assetFillAmount,
+            marketOperation,
+            opts.rfqt,
+        );
+    } else {
+        return Promise.resolve<RfqtIndicativeQuoteResponse[]>([]);
```
steveklebanoff, line 51: Can't we just `return [];` here?

Note: this is a nit, don't need to change

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -590,7 +590,12 @@ export class SwapQuoter {
         return swapQuote;
     }
     private _shouldEnableIndicativeRfqt(opts: CalculateSwapQuoteOpts['rfqt']): boolean {
-        return opts !== undefined && !opts.intentOnFilling && this._rfqtTakerApiKeyWhitelist.includes(opts.apiKey);
+        return (
+            opts !== undefined &&
+            opts.isIndicative !== undefined &&
```
steveklebanoff, line 595: dont need this undefined check

packages/asset-swapper/src/types.ts
```diff
@@ -188,16 +187,20 @@ export interface SwapQuoteOrdersBreakdown {
     [source: string]: BigNumber;
 }
 
+export interface RfqtRequestOpts {
+    takerAddress: string;
+    apiKey: string;
+    intentOnFilling: boolean;
+    isIndicative?: boolean;
```
steveklebanoff, line 194: Nit but would prefer for this to be required

PR #2564, opened 2020-04-23
ZeroEx: Merge Migrate and Ownable Features

dorothy-zbornak:
## Description
This merges the Migrate and Ownable features in ZeroEx into just Ownable.

The rationale is that Migrate was reaching into Ownable's storage, which meant they were tightly coupled, so they might as well be one feature.

- I also removed Migrate's `getMigrationOwner()` function because the original owner can just be encoded in the calldata during a migration.
- `migrate()` now accepts a third parameter: the new owner. Previously a `migrate()` call could not also set a new owner because it would restore the original before returning. Now it sets it to `newOwner` before returning.

#### Maintenance
I also snuck in a fix to some asset-swapper tests I wrote a while back that are infrequently failing on CI.

## Checklist:


contracts/zero-ex/contracts/src/features/Ownable.sol
```diff
@@ -34,26 +35,36 @@ contract Ownable is
...
+    // solhint-enable const-name-snakecase
+
+    // solhint-disable
+    /// @dev The deployed address of this contract.
+    address immutable private _implementation;
+    // solhint-enable
+
+    constructor() public {
+        _implementation = address(this);
+    }
```
dorothy-zbornak, line 53: Also switched to this pattern instead of explicitly passing in `impl` in `bootstrap()`. This is the pattern we will also follow for `migrate()`.

PR #2565, opened 2020-04-24
[asset-swapper] clip native fill data

dekz:
## Description

There is a scenario where perfectly sized Native orders were penalized more aggressively than they should have been. 

The previous behaviour was as follows:
```
penaltyMakerUnits = 10
input = 2000

SMALL_ORDER
makerAmount = 100
takerAmount = 200

adjustedOutput = 90
adjustedRate = 0.45


BIG_ORDER
makerAmount = 1000
takerAmount = 2000

adjustedOutput = 990
adjustedRate = 0.495
```

In the scenario where the user wishes to sell `2000` the `SMALL_ORDER` should be penalized over `BIG_ORDER` and `BIG_ORDER` should be favoured. 

In the scenario where the user wishes to sell `200` both `SMALL_ORDER` and `BIG_ORDER` are sufficient and either is acceptable to fill.

```
penaltyMakerUnits = 10
input = 200

SMALL_ORDER
makerAmount = 100
takerAmount = 200

adjustedOutput = (min(takerAmount, input) / takerAmount) * makerAmount - penaltyMakerUnits = 90
adjustedRate = 0.45


BIG_ORDER
makerAmount = 1000
takerAmount = 2000

adjustedOutput = (min(takerAmount, input) / takerAmount) * makerAmount - penaltyMakerUnits = 90
adjustedRate = 0.45
```

We effectively bring large orders in line with small orders when they completely cover the requested amount. 

### Issue
The following behaviour caused perfectly sized orders to disappear:
1. Perfectly sized orders were penalized more harshly than large orders
2. Native orders are sorted by `adjustedRate`, then a subset of native orders are chosen (enough to fill the requested amount).
3. Since better `rate` orders were dropped in 2), they would no longer be present when finding a path.

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -74,7 +74,7 @@ function nativeOrdersToPath(
         const penalty = ethToOutputRate.times(fees[ERC20BridgeSource.Native] || 0);
         const rate = makerAmount.div(takerAmount);
         // targetInput can be less than the order size
-        // whilst the penalty is constant, it effects the adjusted output
+        // whilst the penalty is constant, it affects the adjusted output
         // only up until the target has been exhausted.
         // A large order and an order at the exact target should be penalized
         // the same.
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -35,7 +35,7 @@ export function createFillPaths(opts: {
     const dexQuotes = opts.dexQuotes || [];
     const ethToOutputRate = opts.ethToOutputRate || ZERO_AMOUNT;
     // Create native fill paths.
-    const nativePath = nativeOrdersToPath(side, orders, ethToOutputRate, feeSchedule);
+    const nativePath = nativeOrdersToPath(side, orders, opts.targetInput, ethToOutputRate, feeSchedule);
     // Create DEX fill paths.
     const dexPaths = dexQuotesToPaths(side, dexQuotes, ethToOutputRate, feeSchedule);
     return filterPaths([...dexPaths, nativePath].map(p => clipPathToInput(p, opts.targetInput)), excludedSources);
@@ -60,6 +60,7 @@ function filterPaths(paths: Fill[][], excludedSources: ERC20BridgeSource[]): Fil
 function nativeOrdersToPath(
     side: MarketOperation,
     orders: SignedOrderWithFillableAmounts[],
+    targetInput: BigNumber = POSITIVE_INF,
     ethToOutputRate: BigNumber,
     fees: { [source: string]: BigNumber },
 ): Fill[] {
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -1017,5 +1019,58 @@ describe('MarketOperationUtils tests', () => {
...
+                flags: 0,
+                index: 0,
+                source: ERC20BridgeSource.Native,
+            };
+            const fillPaths = [[uni1, uni2, uni3], [nativeOrder]];
+            const optimalPath = findOptimalPath(MarketOperation.Sell, fillPaths, nativeOrder.input);
+            expect(BigNumber.max(nativeOrder.rate, uni1.rate, uni2.rate, uni3.rate)).to.be.bignumber.eq(
+                nativeOrder.rate,
+            );
+            expect((optimalPath as Fill[])[0].source).to.eq(ERC20BridgeSource.Native);
```
dorothy-zbornak, line 1072: Wouldn't this have been true before this patch, since it's the only native order in the path? Don't we want a test case following the scenario in the issue?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -92,11 +92,12 @@ function isPathComplete(path: Fill[], targetInput: BigNumber): boolean {
 }
 
 function clipFillAdjustedOutput(fill: Fill, remainingInput: BigNumber): BigNumber {
+    const penalty = fill.adjustedOutput.minus(fill.output);
+    const output = remainingInput.times(fill.rate).plus(penalty);
```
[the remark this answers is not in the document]
dekz: No good reason to keep it, but I was logging penalty and all of the info for both cases.

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -1017,5 +1019,58 @@ describe('MarketOperationUtils tests', () => {
             });
         });
     });
+
+    describe('Path Optimizer', () => {
+        it('optimizes exact native order correctly', () => {
```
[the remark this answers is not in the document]
dekz: Will clean them up so they're a bit nicer to parse. This came from direct real world data

packages/asset-swapper/test/market_operation_utils_test.ts
dekz: Yep this particular test is not relevant.  The issue was really around clipping native paths and the adjusted output. 

This test was triangulating the issue, I'll follow up with a better one.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -71,19 +72,26 @@ function nativeOrdersToPath(
         const input = side === MarketOperation.Sell ? takerAmount : makerAmount;
         const output = side === MarketOperation.Sell ? makerAmount : takerAmount;
         const penalty = ethToOutputRate.times(fees[ERC20BridgeSource.Native] || 0);
-        const adjustedOutput = side === MarketOperation.Sell ? output.minus(penalty) : output.plus(penalty);
         const rate = makerAmount.div(takerAmount);
+        // targetInput can be less than the order size
+        // whilst the penalty is constant, it affects the adjusted output
+        // only up until the target has been exhausted.
+        // A large order and an order at the exact target should be penalized
```
PirosB3, line 79: should *not* be penalized the same?

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -71,19 +72,26 @@ function nativeOrdersToPath(
...
+        // the same.
+        const clippedInput = BigNumber.min(targetInput, input);
+        // scale the clipped output inline with the input
+        const clippedOutput = clippedInput.dividedBy(input).times(output);
```
PirosB3, line 83: should we be concerned about rounding errors with this operation? of does BigNumber prevent rounding issues?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -96,7 +96,7 @@ function clipFillAdjustedOutput(fill: Fill, remainingInput: BigNumber): BigNumbe
         return fill.adjustedOutput;
     }
     const penalty = fill.adjustedOutput.minus(fill.output);
-    return fill.output.times(remainingInput.div(fill.input)).plus(penalty);
+    return remainingInput.times(fill.rate).plus(penalty);
```
PirosB3, line 99: Why was this function modified? based on my understanding, it should not impact `nativeOrdersToPath` which is the core function that you changed

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
dekz: they **should** be penalized the same, since the cost is constant and the amount is constant. Where having a bigger order size is no advantage.

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
dekz: I find it just easier to reason about. `input * rate = output`, versus scaling the previous output by the proportion of remaining vs total.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
dekz: These remain very precise, but I could go either way in having these as whole `base units` which would lose precision but not necessarily have too much of an impact.

PR #2573, opened 2020-04-29
ZeroEx: InitialMigration Race Condition/Self-destruct

dorothy-zbornak:
## Description

The `InitialMigration` contract had a race condition where anyone could call `deploy()` first and prevent the governor from calling it. Not a vuln, just a nuisance.

Now the caller to `deploy()` is set in the constructor of `InitialMigration` AND the contract self-destructs before returning. The old comments actually stated this latter behavior but the code never actually did it.

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/migrations/InitialMigration.sol
+++ b/contracts/zero-ex/contracts/src/migrations/InitialMigration.sol
@@ -28,25 +28,42 @@ import "./LibBootstrap.sol";
 
 /// @dev A contract for deploying and configuring a minimal ZeroEx contract.
 contract InitialMigration {
-    /// @dev The deployed ZereEx instance.
-    ZeroEx private _zeroEx;
+
+    /// @dev The allowed caller of `deploy()`. In production, this would be
+    ///      the governor.
+    address public immutable deployer;
+    /// @dev The real address of this contract.
+    address private immutable _implementation;
+
+    /// @dev Instantiate this contract and set the allowed caller of `deploy()`
+    ///      to `deployer_`.
+    /// @param deployer_ The allowed caller of `deploy()`.
+    constructor(address deployer_) public {
+        deployer = deployer_;
+        _implementation = address(this);
+    }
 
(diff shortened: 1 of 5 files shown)


packages/contract-wrappers-test/package.json
```diff
@@ -15,7 +15,7 @@
         "fix": "tslint --fix --format stylish --project .--exclude **/lib/**/*",
         "test:circleci": "run-s test:coverage",
         "test": "yarn run_mocha",
-        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js lib/test/global_hooks.js --timeout 10000 --bail --exit",
+        "run_mocha": "mocha --require source-map-support/register --require make-promises-safe lib/test/**/*_test.js lib/test/global_hooks.js --timeout 20000 --bail --exit",
```
dorothy-zbornak, line 18: I've seen this test time out on CI like a dozen times in the last month.

contracts/zero-ex/contracts/src/migrations/InitialMigration.sol
```diff
@@ -28,25 +28,42 @@ import "./LibBootstrap.sol";
...
+    /// @dev The allowed caller of `deploy()`. In production, this would be
+    ///      the governor.
+    address public immutable deployer;
+    /// @dev The real address of this contract.
+    address private immutable _implementation;
+
+    /// @dev Instantiate this contract and set the allowed caller of `deploy()`
+    ///      to `deployer_`.
+    /// @param deployer_ The allowed caller of `deploy()`.
+    constructor(address deployer_) public {
```
hysz, line 41: Not a blocker here, but it could be good to rethink our naming conventions. We've had feedback that our use of underscores is confusing. In one of the other repos we've nixed the `_` for variables altogether and prefixed globals with `g`.

dorothy-zbornak: Yeah I'm down to make this change down the line.

PR #2574, opened 2020-04-29
asset-swapper: RFQ-T follow ups

feuGeneA:
## Description

Lots of tasks that needed doing after the initial implementation of RFQ-T.  I recommend reviewing commit by commit.

## Testing instructions

Changes are covered by tests running in CI, with the following exception:
* https://github.com/0xProject/0x-monorepo/commit/dad25b86e841a20c06a746b11b59445eb00064f3 is not tested here, but it _is_ covered, over in the corresponding 0x API PR.

## Types of changes

## Checklist:


packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -81,14 +82,14 @@ function hasExpectedAssetData(
...
 export class QuoteRequestor {
-    private readonly _rfqtMakerEndpoints: string[];
     private readonly _schemaValidator: SchemaValidator = new SchemaValidator();
-    private readonly _warningLogger: (s: string) => void;
 
-    constructor(rfqtMakerEndpoints: string[], logger: (s: string) => void = s => logUtils.warn(s)) {
-        this._rfqtMakerEndpoints = rfqtMakerEndpoints;
-        this._warningLogger = logger;
-    }
+    constructor(
```
steveklebanoff, line 87: Never seen constructor assignment like this before, interesting

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -81,14 +82,14 @@ function hasExpectedAssetData(
...
+        private readonly _rfqtAssetOfferings: RfqtMakerAssetOfferings,
+        private readonly _warningLogger: (a: any) => void = a => logUtils.warn(a),
+        private readonly _infoLogger: (a: any) => void = () => undefined,
```
steveklebanoff, line 90: Do you think this should default to `(a) => logUtils.info(a)`?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -150,6 +156,21 @@ export class QuoteRequestor {
                 return false;
             }
 
+            if (order.takerAddress !== _opts.takerAddress) {
+                this._warningLogger(`Unexpected takerAddress in RFQ-T order, filtering out: ${JSON.stringify(order)}`);
+                return false;
+            }
+
+            const minExpiry = Math.round((Date.now() + this._expiryBufferMs) / constants.ONE_SECOND_MS);
```
steveklebanoff, line 164: It seems weird to take in a `milliseconds` param just round it to `seconds` -- esp. since this would round down `400 ms` to `0 seconds`.  What do you think about just accepting the param in `seconds`?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -102,28 +103,33 @@ export class QuoteRequestor {
...
+                            } for taker address ${_opts.takerAddress}`,
+                        );
+                        this._warningLogger(err);
+                        return undefined;
+                    }
                 }
+                return undefined;
             }),
         );
+        this._infoLogger({ aggregatedRfqtLatencyMs: Date.now() - timeBeforeAwait });
```
steveklebanoff, line 132: In a world where there are multiple RFQT providers: if this number is large -- how do we know which MM is responsible for it being large? I think ideally we would log the latency of each individual endpoint.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -150,6 +156,21 @@ export class QuoteRequestor {
                 return false;
             }
 
+            if (order.takerAddress !== _opts.takerAddress) {
```
steveklebanoff, line 159: Just in case there is a casing mismatch, it might be worthwhile to call `.toLowerCase()` on each of these

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -224,6 +250,15 @@ export class QuoteRequestor {
                 );
                 return false;
             }
+            const minExpiry = Math.round((Date.now() + this._expiryBufferMs) / constants.ONE_SECOND_MS);
+            if (new BigNumber(response.expirationTimeSeconds).isLessThan(minExpiry)) {
```
steveklebanoff, line 254: What do you think about putting this in a helper fn?

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -165,14 +167,19 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
-        this._rfqtTakerApiKeyWhitelist = options.rfqt ? options.rfqt.takerApiKeyWhitelist || [] : [];
+        this._rfqtTakerApiKeyWhitelist = rfqt ? rfqt.takerApiKeyWhitelist || [] : [];
+        this._rfqtSkipBuyRequests =
+            // tslint:disable-next-line:no-non-null-assertion these are clearly defined in the declaration
+            (rfqt && rfqt.skipBuyRequests) || constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt!.skipBuyRequests!;
```
steveklebanoff, line 173: This feels a little weird: if `constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt.skipBuyRequests` was true, but `rfqt` was explicitly defined with `{ skipBuyRequests: false }`, this line would still evaluate to `true`.

Could this instead be `rfqt ? rfqt.skipBuyRequests : constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt!.skipBuyRequests!`

steveklebanoff: Also -- I suggest removing `: SwapQuoterOpts` in the `const DEFAULT_SWAP_QUOTER_OPTS: SwapQuoterOpts = {` line in `constants.ts`

This would allow `constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt!.skipBuyRequests!` to be changed to `constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt.skipBuyRequests`

packages/asset-swapper/test/quote_requestor_test.ts
```diff
@@ -102,9 +114,40 @@ describe('QuoteRequestor', async () => {
...
+            });
+
+            // Shouldn't ping an RFQ-T provider when they don't support the requested asset pair.
+            // (see how QuoteRequestor constructor parameters below don't list
+            // any supported asset pairs for this maker.)
+            mockedRequests.push({
+                endpoint: 'https://426.0.0.1',
+                requestApiKey: apiKey,
+                requestParams: expectedParams,
+                responseData: successfulOrder1,
```
steveklebanoff, line 140: Should this return a different successful order to ensure it doesn't get included?

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I think it shouldn't.

One reason is that we're outputting a JSON object in the calls to `_infoLogger`, not `stringify`ed JSON, for compatibility with Kibana.  So, if we defaulted to `logUtils.log(a)`, then the output would just say `[Object]` or whatever.  We _could_ default to `logUtils.log(JSON.stringify(a))`.

But the other reason is just that I didn't want to sully the output (of the tests, or of anything else integrating with asset-swapper) with a bunch of non-natural language (JSON) output.

I could be convinced to reconsider...

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: I cleaned this up a bunch here: https://github.com/0xProject/0x-monorepo/pull/2574/commits/125c53827bbde841a0d978c405452a14008dffd6

Let me know what you think.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: When we first discussed logging this latency (I believe it was @PirosB3 who suggested it), I'm pretty sure we talked about doing it in aggregate like this.

My understanding is that gathering latency on individual responses isn't possible, because we're awaiting all of the responses in parallel.  If we were to position the code to be able to capture individual response times, I believe that we would be sacrificing the parallelism.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Done, here: https://github.com/0xProject/0x-monorepo/pull/2574/commits/ed3d194c905c96924c79cea8fc3cab263d71287b

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Love it, done here: https://github.com/0xProject/0x-monorepo/pull/2574/commits/125c53827bbde841a0d978c405452a14008dffd6#diff-951722c8cfc47f2ea3eebf81342ae48dR318

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Great catch, thank you.  Unfortunately `skipBuyRequests` is itself optional, so I couldn't use your suggestion verbatim, but it's pretty close: https://github.com/0xProject/0x-monorepo/pull/2574/commits/a9e72085dd5c62409e0433f44a2d9da21c386d70

As for removing that type annotation in `constants.ts`...  In my mind this is a `tslint` bug: it should be looking at the actual declaration of that `constants.DEFAULT...` and seeing that these things actually are defined.  Removing a type annotation just to satisfy a linter doesn't seem right to me.

packages/asset-swapper/test/quote_requestor_test.ts
feuGeneA: Good catch, fixed here: https://github.com/0xProject/0x-monorepo/pull/2574/commits/2922ebd09583303f571f9b27e884d4a38f93f456

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: It is possible without sacrificing the parallelism -- I will slack you about this

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: Good point 👍

packages/asset-swapper/src/swap_quoter.ts
steveklebanoff: The `skipBuyRequests` looks better 👍 

With regards to `constants.ts`: this is not a tslint bug.  This is TypeScript obeying what was defined.  `constants.DEFAULT_SWAP_QUOTER_OPTS` is explicitly defined by us as type `SwapQuoterOpts `, and `SwapQuoterOpts` considers `rfqt` optional.  Thus -- when you reference this var, it doesn't know if `rfqt` is defined or not, because it's relying on the `SwapQuoterOpts` type.

The two ways to fix this are:
- Remove the `: SwapQuoterOpts` casting as I described.  This forces TS to consider `DEFAULT_SWAP_QUOTER_OPTS` to be of the type of the object it is assigned -- which has `rfqt` populated
- Change `DEFAULT_SWAP_QUOTER_OPTS` type assignment to not be `SwapQuoterOpts` but instead be something like `SwapQuoterOptsWithRequiredRfqtParam` -- but there's not a quick utility type to do this

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: It still seems weird to take in `ms` and then round to `seconds`, esp when `400 ms` can round to `0` seconds.    That being said -- I see the expiry buffer is already defined in `ms` (expiryBufferMs), so perhaps we can't get around this.

packages/asset-swapper/test/quote_requestor_test.ts
```diff
@@ -102,9 +114,29 @@ describe('QuoteRequestor', async () => {
                 responseData: unsignedOrder,
                 responseCode: StatusCodes.Success,
             });
+            // A successful response code and good order but for the wrong takerAddress
+            const orderWithNullTaker = testOrderFactory.generateTestSignedOrder({
```
steveklebanoff, line 118: If we wanted to get super granular we could actually test to see if a warning was generated for this case ;)

packages/asset-swapper/src/swap_quoter.ts
steveklebanoff: FWIW I'm ok with this as-is if need be -- this shouldn't be a blocker to getting hte PR merged

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2574/commits/e42701599ad9d50d3f5ff258c826fcbbda9b1b26

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: I still feel weird about removing that type annotation on the constant declaration.  I think I found a no-so-verbose way to restore type safety to this: https://github.com/0xProject/0x-monorepo/pull/2574/commits/352b1b43f2523a307689d91755734533679c31f8

Please let me know what you think

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -165,14 +167,20 @@ export class SwapQuoter {
         this.orderbook = orderbook;
         this.expiryBufferMs = expiryBufferMs;
         this.permittedOrderFeeTypes = permittedOrderFeeTypes;
-        this._rfqtTakerApiKeyWhitelist = options.rfqt ? options.rfqt.takerApiKeyWhitelist || [] : [];
+        this._rfqtTakerApiKeyWhitelist = rfqt ? rfqt.takerApiKeyWhitelist || [] : [];
+        this._rfqtSkipBuyRequests =
```
steveklebanoff, line 171: Looks like this works, but is a little confusing to read.  Does the following work? I think it reads a little easier.

```ts
const skipBuyRequests = (rfqt && rfqt.skipBuyRequests === true) || (constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt && constants.DEFAULT_SWAP_QUOTER_OPTS.rfqt.skipBuyRequests === true)
```

That being said, I think we've noodled on this enough and I'm cool w/ whatever you think :P

PR #2575, opened 2020-04-30
feat: ERC20BridgeSampler Unlock Kyber collisions

dekz:
## Description

Detect when Kyber is using a Uniswap or Eth2Dai bridge and return 0. When the reserve isn't this bridge we return the `makerAmount`. 

Since a Kyber `DAI/USDC` trade is actually `DAI->ETH, ETH->USDC` the Sampler now samples these individual legs rather than `DAI/USDC`. As a result we're sampling closer to the behaviour of the `trade` function. I.e Sampling `DAI/USDC` isn't the same as `DAI->ETH, ETH->USDC` and results vary.

When sampling Kyber via the ERC20BridgeSampler it will ONLY return amounts when Kyber Reserves are hit. This means its possible for the Kyber sampler to return:
```
0, // UniswapReserve
0, // UniswapReserve
100, // KyberReserve
200, // KyberReserve
0, // UniswapReserve
```

We also move Native to the front and on-chain to the end for a new fallback path strategy. If a Native order was previously after an on-chain at the end, it is dropped.

Since Native orders are now all at the front, they cannot split an on-chain fill and incur an additional protocol fee.

Fallback is now the optimal route of on-chain sources (previously a 3rd source). Native orders are then simply prepended to the front. This means that in the optimal case (no Native reverts) a middle on-chain source may get filled more than optimal (by the Native amount). 

For example, picture a sell of 100 DAI where the optimal route is a 3 way split between 0x Native, Uniswap and Kyber. The fallback is 50/50 Uni/Kyber. The route is adjusted to fill: `33 Native, 50 Uni, 50  Kyber`. In the event of Native reverts it fills at `50 Uni, 50 Kyber`. In the event of no reverts it fills at `33 Native, 50 Uni, 17 Kyber`

In a 2 way split this matters less. `33 Native, 77 Uniswap` -> `33 Native, 100 Uniswap`

**TODO**
- [x] redeploy Mainnet sampler with verified source

**Follow ups:**
- [ ] permissioned vs permissionless Kyber Reserves. Requires redeploy of KyberBridge


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -343,32 +336,38 @@ export class MarketOperationUtils {
...
             );
-            if (fallbackSlippage > maxFallbackSlippage) {
-                fallbackPath = [];
+            if (fallbackSlippage <= maxFallbackSlippage) {
+                // If the last fill is Native and penultimate is not
+                // then this intention is a partial fill
+                // In this case we drop it as we don't want it to fail
+                // at the end and we don't want to fully fill it
+                // tslint:disable-next-line: no-unused-variable
+                const [last, penultimate, ..._rest] = optimalPath.reverse();
```
dorothy-zbornak, line 361: We can just drop the `_rest` var, yeah?

dorothy-zbornak: Even though we don't use the value of `optimalPath` again in the code, might be a god idea to do `optimalPath.slice().reverse()` since `reverse()` is a mutator and might trip up other people expecting `optimalPath` to still be the same at this point.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -343,32 +336,38 @@ export class MarketOperationUtils {
...
+                inputAmount,
                 getPathAdjustedRate(side, optimalPath, inputAmount),
             );
-            if (fallbackSlippage > maxFallbackSlippage) {
-                fallbackPath = [];
+            if (fallbackSlippage <= maxFallbackSlippage) {
+                // If the last fill is Native and penultimate is not
+                // then this intention is a partial fill
+                // In this case we drop it as we don't want it to fail
+                // at the end and we don't want to fully fill it
```
dorothy-zbornak, line 359: Couldn't a single order at the end also be a partial fill?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -343,32 +336,38 @@ export class MarketOperationUtils {
...
+                // tslint:disable-next-line: no-unused-variable
+                const [last, penultimate, ..._rest] = optimalPath.reverse();
+                const finalNativeFillOrUndefined =
+                    last.source === ERC20BridgeSource.Native && penultimate.source !== ERC20BridgeSource.Native
```
dorothy-zbornak, line 363: What happens if `penultimate` is `undefined`, like when the optimalPath is just a single native order?

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -807,4 +793,53 @@ contract ERC20BridgeSampler is
...
+                ));
+        uint256 rate = 0;
+        address reserve;
+        if (didSucceed) {
+            (reserve, rate) = abi.decode(resultData, (address, uint256));
+        } else {
+            return makerTokenAmount;
+        }
+        if (reserve != 0x31E085Afd48a1d6e51Cc193153d625e8f0514C7F || // Uniswap
+            reserve != 0x1E158c0e93c30d24e918Ef83d1e0bE23595C3c0f)   // Eth2Dai
```
dorothy-zbornak, line 835: nit: put these into constants

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -807,4 +793,53 @@ contract ERC20BridgeSampler is
...
+        address _makerToken = makerToken == _getWethAddress() ? KYBER_ETH_ADDRESS : makerToken;
+        uint256 takerTokenDecimals = _getTokenDecimals(takerToken);
+        uint256 makerTokenDecimals = _getTokenDecimals(makerToken);
+        (bool didSucceed, bytes memory resultData) = _getKyberNetworkProxyAddress().staticcall.gas(DEFAULT_CALL_GAS)(
+            abi.encodeWithSelector(
+                IKyberNetwork(0).kyberNetworkContract.selector
+            ));
+        if (!didSucceed) {
+            return makerTokenAmount;
+        }
```
dorothy-zbornak, line 816: It's probably safe to just do a regular `address kyberNetworkContract = IKyberNetwork(_getKyberNetworkProxyAddress()).kyberNetworkContract();`. Kyber is on kovan, right?

contracts/erc20-bridge-sampler/contracts/src/IKyberNetwork.sol
```diff
@@ -21,6 +21,8 @@ pragma solidity ^0.5.9;
 
 interface IKyberNetwork {
```
dorothy-zbornak, line 22: I think we should just rename this contract `IKyberNetworkProxy` and the new one `IKyberNetwork`, if it's not too much trouble.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -119,18 +119,15 @@ function dexQuotesToPaths(
     fees: { [source: string]: BigNumber },
 ): Fill[][] {
     const paths: Fill[][] = [];
-    for (const quote of dexQuotes) {
+    for (let quote of dexQuotes) {
         const path: Fill[] = [];
+        // Drop any non-zero entries. This can occur if the
+        // first few fills on Kyber were UniswapReserves
+        quote = quote.filter(q => !q.output.isZero());
```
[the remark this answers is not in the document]
dekz: I thought this too originally but didn't seem to encounter it. 

Let's say the samples for Kyber looks like: `[0, 0, 1000]` Where `0`s are other reserves, we only unlock kyber at the third stop (enough volume for it to switch). In the routing we'd only select Kyber for at least this amount, so we're going to hit the right reserve. But that could all change at execution time.

I expected to see it in simbot, but it didn't seem to occur, so I removed sorting Kyber last.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
dekz: Yeah normally it is a partial fill. And re-ordering (later on) will make it a complete fill.

It's easier to reason about dropping it than handling the case where it reverts.

I'll make this comment clearer why it reads "fully fill" (due to re-ordering of Native first).

packages/asset-swapper/src/utils/market_operation_utils/index.ts
dekz: 💥

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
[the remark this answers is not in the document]
dorothy-zbornak: If we can gather compelling data on pairs where this scenario is most likely then I'm fine with this hack. We should def document this reasoning and accepted risk in the code though. Otherwise I'll forget one day and revert this change.

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
dekz: It's not in ganache

contracts/erc20-bridge-sampler/contracts/src/ERC20BridgeSampler.sol
dekz: I kinda feel bad putting everything into `DeploymentConstants`, you agree with just a constant in the sampler?

dorothy-zbornak: Yeah that's fine. Do we need to also have the addresses handy for Kovan?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -343,32 +324,42 @@ export class MarketOperationUtils {
...
                 side,
-                fallbackPath,
-                fallbackInputAmount,
+                nonNativeOptimalPath,
+                inputAmount,
                 getPathAdjustedRate(side, optimalPath, inputAmount),
             );
-            if (fallbackSlippage > maxFallbackSlippage) {
-                fallbackPath = [];
+            if (fallbackSlippage <= maxFallbackSlippage) {
```
dorothy-zbornak, line 345: How about we do
```suggestion
            if (nativeSubPath.length === optimalPath.length || fallbackSlippage <= maxFallbackSlippage) {
```
so we always provide a fallback if the optimal path is purely native?

dekz: I like it but I'm also scared of it. Let me think on it...

PR #2576, opened 2020-05-05
ZeroEx: ERC20 Transformers

dorothy-zbornak:
## Description

tenor

Here are the first 3 transformers for use in the `TransformERC20` feature debuted in #2545:

- **WethTransformer**
  - Wraps ETH or unwraps WETH.
- **PayTakerTransformer**
  - Transfer tokens/ETH to the taker.
- **FillQuoteTransformer**
  - Fills a buy or sell quote.

So for an ETH -> DAI quote, we might autobots assemble:
1. `WethTransformer({token: ETH})`
2. `FillQuoteTransformer({sellToken: WETH, buyToken: DAI})`
3. `PayTakerTransformer({tokens: [DAI, ETH]})` (ETH to refund protocol fees)

Or for DAI -> ETH:
1. `FillQuoteTransformer({sellToken: DAI, buyToken: WETH})`
2. `WethTransformer({token: WETH})`
3. `PayTakerTransformer({tokens: [ETH]})`

Or for DAI -> USDC
1. `FillQuoteTransformer({sellToken: DAI, buyToken: USDC})`
2. `PayTakerTransformer({tokens: [USDC, ETH]})` (ETH to refund protocol fees)

## Checklist:


contracts/zero-ex/contracts/src/transformers/LibERC20Transformer.sol
```diff
@@ -70,4 +70,45 @@ library LibERC20Transformer {
     {
         return isTokenETH(token) ? owner.balance : token.balanceOf(owner);
     }
+
+    /// @dev RLP-encode a 32-bit or less account nonce.
+    /// @param nonce A positive integer in the range 0 <= nonce < 2^32.
+    /// @return rlpNonce The RLP encoding.
+    function rlpEncodeNonce(uint256 nonce)
```
dorothy-zbornak, line 77: Yes, this is actually less gas than an SLOAD lol.

contracts/zero-ex/contracts/src/transformers/Transformer.sol
```diff
@@ -0,0 +1,80 @@
...
+    constructor(uint256 deploymentNonce_) public {
+        deploymentNonce = deploymentNonce_;
+        deployer = msg.sender;
+        _implementation = address(this);
+    }
+
+    /// @dev Destruct this contract. Only callable by the deployer and will not
+    ///      succeed in the context of a delegatecall (from another contract).
+    /// @param ethRecipient The recipient of ETH held in this contract.
+    function die(address payable ethRecipient)
```
dorothy-zbornak, line 52: we need to be able to murder transformers since there's no way to un-register them.

contracts/zero-ex/contracts/src/vendor/v3/IExchange.sol
```diff
@@ -0,0 +1,107 @@
...
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+
+/// @dev Interface to the V3 Exchange.
+interface IExchange {
```
mzhu25, line 24: jw what's the etymology of this "vendor" directory?

contracts/zero-ex/contracts/test/TestDelegateCaller.sol
```diff
@@ -0,0 +1,37 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+*/
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+
```
mzhu25, line 22: extra newline

contracts/zero-ex/contracts/test/TestFillQuoteTransformerExchange.sol
```diff
@@ -0,0 +1,145 @@
...
+contract TestFillQuoteTransformerExchange {
+
+    struct FillBehavior {
+        // How much of the order is filled, in taker asset amount.
+        uint256 filledTakerAssetAmount;
+        // Scaling for maker assets minted, in 1e18.
+        uint256 makerAssetMintRatio;
+    }
+
+    uint256 private constant PROTOCOL_FEE_MULTIPLIER = 1337;
```
mzhu25, line 38: ZEIP? 😝

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+        Transformer(deploymentNonce_)
+    {
+        exchange = exchange_;
+        erc20Proxy = exchange_.getAssetProxy(ERC20_ASSET_PROXY_ID);
+    }
+
+    /// @dev Sell this contract's entire balance of of `sellToken` in exchange
+    ///      for `buyToken` by filling `orders`. Protocol fees should be attached
+    ///      to this call. `buyToken` and excess ETH will be transferred back to the caller.
+    ///      This function cannot be re-entered.
```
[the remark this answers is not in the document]
dorothy-zbornak: Hopes and wishes (old comment).

contracts/zero-ex/contracts/src/vendor/v3/IExchange.sol
dorothy-zbornak: Vendor is convention for third party stuff. Right now it's just the exchange but in the future I would put uniswap, oasis, kyber interfaces in there. https://softwareengineering.stackexchange.com/questions/123305/what-is-the-difference-between-the-lib-and-vendor-folders

contracts/zero-ex/contracts/test/TestFillQuoteTransformerExchange.sol
dorothy-zbornak: I'll hit up Theo.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+        returns (FillOrderResults memory results)
+    {
+        IERC20TokenV06 takerFeeToken = order.takerFeeAssetData.length == 0
+            ? IERC20TokenV06(address(0))
+            : _getTokenFromERC20AssetData(order.takerFeeAssetData);
+
+        uint256 takerTokenFillAmount = sellAmount;
+
+        if (order.takerFee != 0) {
+            if (takerFeeToken == makerToken) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah we can just quickly deploy a new one when we have more interesting orders coming out of the API.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+                // Taker fee is payable in the maker token.
+                // Increase the fill amount to account for maker tokens being
+                // lost to the taker fee.
+                // makerTokenFillAmount' =
+                //  (order.makerAssetAmount * makerTokenFillAmount) /
+                //  (order.makerAssetAmount - order.takerFee)
+                makerTokenFillAmount = LibMathV06.getPartialAmountCeil(
```
hysz, line 323: We could reduce the computation and potential for rounding errors by computing the `takerTokenFillAmount` here. I believe it works out to just:
```
takerTokenFillAmount  = LibMathV06.getPartialAmountCeil(
    makerTokenFillAmount,
    order.makerAssetAmount.safeSub(order.takerFee), 
    order.takerAssetAmount
);  
```

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+        uint256 takerAssetFillAmount,
+        uint256 protocolFee,
+        IERC20TokenV06 makerToken,
+        bool isTakerFeeInTakerToken
+    )
+        private
+        returns (FillOrderResults memory results)
+    {
+        // Track changes in the maker token balance.
+        results.makerTokenBoughtAmount = makerToken.balanceOf(address(this));
```
hysz, line 383: Nit - the semantics here is sort of confusing. The way this variable is used within this function, the maker amount bought is actually zero at this point. Might just do:
```
uint256 initBalance = makerToken.balanceOf(address(this));
```

contracts/zero-ex/contracts/src/transformers/PayTakerTransformer.sol
```diff
@@ -0,0 +1,89 @@
...
+
+        // Transfer tokens directly to the taker.
+        for (uint256 i = 0; i < data.tokens.length; ++i) {
+            // The `amounts` array can be shorter than the `tokens` array.
+            // Missing elements are treated as `uint256(-1)`.
+            uint256 amount = data.amounts.length > i ? data.amounts[i] : uint256(-1);
+            if (amount == uint256(-1)) {
+                amount = data.tokens[i].getTokenBalanceOf(address(this));
+            }
+            if (amount != 0) {
```
hysz, line 83: I wonder if we should error out when `amount == 0` 🤔. Is there a use case for this?

contracts/zero-ex/contracts/src/transformers/Transformer.sol
```diff
@@ -0,0 +1,80 @@
...
+        virtual
+    {
+        // Only the deployer can call this.
+        if (msg.sender != deployer) {
+            LibTransformERC20RichErrors
+                .OnlyCallableByDeployerError(msg.sender, deployer)
+                .rrevert();
+        }
+        // Must be executing our own context.
+        if (address(this) != _implementation) {
```
hysz, line 63: +1

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+        bytes calldata data_
+    )
+        external
+        override
+        returns (bytes memory rlpDeploymentNonce)
+    {
+        TransformData memory data = abi.decode(data_, (TransformData));
+
+        // Validate data fields.
+        if (data.sellToken.isTokenETH() ||
```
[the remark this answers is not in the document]
dorothy-zbornak: How about different error codes?

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,424 @@
...
+
+    /// @dev The Exchange ERC20Proxy ID.
+    bytes4 constant private ERC20_ASSET_PROXY_ID = 0xf47261b0;
+
+    /// @dev The Exchange contract.
+    IExchange public immutable exchange;
+    /// @dev The ERC20Proxy address.
+    address public immutable erc20Proxy;
+
+    using LibERC20TokenV06 for IERC20TokenV06;
```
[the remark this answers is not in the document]
dorothy-zbornak: idk I kind of like having them down before the first function since that's when the syntactic sugar can actually get used. The rest of the code base is like this as well so I'd have to touch nearly every other contract. We can do a separate re-styling PR prior to deployment if we agree it's better.

contracts/zero-ex/contracts/src/transformers/PayTakerTransformer.sol
dorothy-zbornak: Yeah. The `PayTakerTransformer` is also used to refund unspent protocol fees, which could end up being `0` if they were all used.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dorothy-zbornak: nice!

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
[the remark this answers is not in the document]
hysz: Okay yeah not a blocker, but we should probably put these at the top since they apply to the entire contract.

contracts/zero-ex/contracts/src/transformers/WethTransformer.sol
```diff
@@ -0,0 +1,98 @@
...
+/// @dev A transformer that wraps or unwraps WETH.
+contract WethTransformer is
+    Transformer
+{
+    /// @dev Transform data to ABI-encode and pass into `transform()`.
+    struct TransformData {
+        // The token to wrap/unwrap. Must be either ETH or WETH.
+        IERC20TokenV06 token;
+        // Amount of `token` to wrap or unwrap.
+        // `uint(-1)` will unwrap the entire balance.
```
hysz, line 39: I think we should try to make these API's as explicit as possible. For example, we could standardize that if a transformer has more than one action we'll use an Enum to differentiate:
```
enum Action {
    UNWRAP,
    WRAP
}
```

dorothy-zbornak: idk, is `1` or `0` any more explicit than passing in 2 distinct addresses?

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,437 @@
...
+                // Market sell.
+                results = _sellToOrder(
+                    data.buyToken,
+                    data.sellToken,
+                    data.orders[i],
+                    data.signatures[i],
+                    data.fillAmount.safeSub(soldAmount).min256(
+                        data.maxOrderFillAmounts.length > i
+                        ? data.maxOrderFillAmounts[i]
+                        : uint256(-1)
```
hysz, line 184: Nit - use a constant here to help readability.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -0,0 +1,437 @@
...
+            // Market sell check.
+            if (soldAmount < data.fillAmount) {
+                LibTransformERC20RichErrors
+                    .IncompleteFillSellQuoteError(
+                        address(data.sellToken),
+                        soldAmount,
+                        data.fillAmount
+                    ).rrevert();
+            }
+        } else {
```
hysz, line 221: Is there an automatic check when decoding that enforces `data.side` is `Side.Sell` or `Side.Buy`? If not we should probably do `...else if (data.side == Side.Buy) {..} else {revert()}`

contracts/zero-ex/contracts/src/transformers/WethTransformer.sol
hysz: Are you thinking of like a separate transformer for wrapping and unwrapping? That's probably fine too.

hysz: Btw I think the confusion just arises here because we're overloading the `uint(-1)`. I think this is generally used to mean like the whole balance, but here we also use it to flip the operation. I don't feel too strongly either way, but worth considering.

dorothy-zbornak: We don't use `uint256(-1)` to flip the operation, we specify either the WETH address or the ETH address (0xeee...).

dorothy-zbornak: `{token: 0xeeee..., amount: uint256(-1)`: Wrap all ETH.
`{token: WETH_ADDRESS, amount: uint256(-1)`: Unwrap all WETH.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dorothy-zbornak: Yeah, the `abi.decode()` will do a range check on the enum value.

contracts/zero-ex/contracts/src/transformers/WethTransformer.sol
hysz: Ah okay, that makes more sense. +1

PR #2582, opened 2020-05-15
asset-swapper: Use @0x/quote-server, not local typedefs; and adapt to renamed taker request parameters

feuGeneA:
## Description

`@0x/quote-server` is now the canonical reference for the interface between asset-swapper and RFQ maker endpoints.  This PR stops using locally defined types for that interface, and imports them from `@0x/quote-server` instead.

This PR also adapts to changes introduced to `@0x/quote-server` in [its PR #5](https://github.com/0xProject/quote-server/pull/5).

## Testing instructions

Tests are integrated with CI.

## Types of changes

## Checklist:

-   [x] Figure out why `test-publish` is failing in CI.  It's saying:
```
ERROR in package @0x/asset-swapper:
Command failed: /root/repo/node_modules/typescript/bin/tsc


node_modules/@0x/asset-swapper/lib/src/utils/quote_requestor.d.ts:2:33 - error TS2307: Cannot find module '@0x/quote-server'.

2 import { IndicativeQuote } from '@0x/quote-server';
                                  ~~~~~~~~~~~~~~~~~~

node_modules/@0x/asset-swapper/lib/src/utils/quote_requestor.d.ts:6:33 - error TS2307: Cannot find module '@0x/quote-server'.

6 export { IndicativeQuote } from '@0x/quote-server';
                                  ~~~~~~~~~~~~~~~~~~

Error: Command failed: /root/repo/node_modules/typescript/bin/tsc
```


packages/asset-swapper/src/index.ts
```diff
@@ -70,5 +70,5 @@ export {
 } from './utils/market_operation_utils/types';
 export { affiliateFeeUtils } from './utils/affiliate_fee_utils';
 export { ProtocolFeeUtils } from './utils/protocol_fee_utils';
-export { QuoteRequestor, RfqtIndicativeQuoteResponse } from './utils/quote_requestor';
+export { QuoteRequestor, IndicativeQuote } from './utils/quote_requestor';
```
steveklebanoff, line 73: Why not just export IndicativeQuote from `quote-server` directly?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -1,24 +1,20 @@
...
 import { assetDataUtils, orderCalculationUtils, SignedOrder } from '@0x/order-utils';
+import { FirmQuote, IndicativeQuote } from '@0x/quote-server';
 import { ERC20AssetData } from '@0x/types';
 import { BigNumber, logUtils } from '@0x/utils';
 import Axios, { AxiosResponse } from 'axios';
 
 import { constants } from '../constants';
 import { MarketOperation, RfqtMakerAssetOfferings, RfqtRequestOpts } from '../types';
 
+export { SignedOrder } from '@0x/order-utils';
```
steveklebanoff, line 11: Do we need to export these here?

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -4,7 +4,7 @@ import { ERC20BridgeAssetData, SignedOrder } from '@0x/types';
 import { AbiEncoder, BigNumber } from '@0x/utils';
 
 import { MarketOperation, SignedOrderWithFillableAmounts } from '../../types';
-import { RfqtIndicativeQuoteResponse } from '../quote_requestor';
+import { IndicativeQuote } from '../quote_requestor';
```
steveklebanoff, line 7: Would prefer to import this directly from `@0x/quote-server`

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -121,10 +117,16 @@ export class QuoteRequestor {
...
-            const hasValidSchema = this._schemaValidator.isValid(order, schemas.signedOrderSchema);
-            if (!hasValidSchema) {
-                this._warningLogger(order, 'Invalid RFQ-t order received, filtering out');
+            try {
+                const hasValidSchema = this._schemaValidator.isValid(order, schemas.signedOrderSchema);
+                if (!hasValidSchema) {
+                    throw new Error('order not valid');
+                }
+            } catch (err) {
+                this._warningLogger(order, `Invalid RFQ-t order received, filtering out. ${err.message}`);
```
steveklebanoff, line 129: When would this throw instead of `hasValidSchema` just returning `false`?

feuGeneA: When the input isn't even valid JSON.  Returning false implies that the response is valid JSON but doesn't conform to the schema.  See test case I added for this in the commit that introduced this change: https://github.com/0xProject/0x-monorepo/pull/2582/commits/b9984b6df4555f4d4417d0589dff2f4214ad76a6

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: My logic was that a client should import an interface's dependent types from the module that exports that interface.  Again, happy to defer to convention as you see fit, but this is what made sense to me.

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: My logic was that a module should export all types that its clients will need to reference in order to use that module's API.  We use both of these types in the interfaces for `QuoteRequestor`'s public methods `requestRfqtFirmQuotesAsync()` and `requestRfqtIndicativeQuotesAsync()`.

packages/asset-swapper/src/index.ts
feuGeneA: My logic was that we should import the type from the module that requires us to reference that type.  But if there's some convention here then I'm happy to follow it.

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: Got it, makes sense -- thanks!

packages/asset-swapper/src/index.ts
steveklebanoff: Lines 1 through 16 of this file seem to follow the pattern of exporting directly from the package, so I'd prefer to do this here as well, i.e. `export { IndicativeQuote } from '@0x/quote-server`

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
steveklebanoff: It feels more direct to import directly from `@0x/quote-server`, and I see us using this pattern elsewhere

packages/asset-swapper/src/utils/quote_requestor.ts
steveklebanoff: This feels a bit strange to me considering we already export SignedOrder in `asset-swapper`'s index.ts for any consumer of this package: https://github.com/0xProject/0x-monorepo/blob/b9984b6df4555f4d4417d0589dff2f4214ad76a6/packages/asset-swapper/src/index.ts#L15

I'd prefer to remove this

packages/asset-swapper/package.json
```diff
@@ -52,6 +52,7 @@
         "@0x/json-schemas": "^5.0.7",
         "@0x/order-utils": "^10.2.4",
         "@0x/orderbook": "^2.2.5",
+        "@0x/quote-server": "^1.0.0",
```
steveklebanoff, line 55: Is the plan to merge & publish https://github.com/0xProject/quote-server/pull/5 and then bump up the version in this PR?

steveklebanoff: It looks like once this happens, we can modify many `IndicativeQuote` types to be more explicitly `RFQTIndicativeQuote` types, correct?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -31,7 +31,7 @@ async function getRfqtIndicativeQuotesAsync(
     marketOperation: MarketOperation,
     assetFillAmount: BigNumber,
     opts: Partial<GetMarketOrdersOpts>,
-): Promise<RfqtIndicativeQuoteResponse[]> {
+): Promise<IndicativeQuote[]> {
```
steveklebanoff, line 34: Would it make sense to use `RFQTIndicativeQuote` instead of `IndicativeQuote` type in this file?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -51,20 +44,20 @@ function inferQueryParams(
     makerAssetData: string,
     takerAssetData: string,
     assetFillAmount: BigNumber,
-): { buyToken: string; sellToken: string; buyAmount?: string; sellAmount?: string } {
+): Pick<TakerRequest, 'buyTokenAddress' | 'sellTokenAddress' | 'buyAmountBaseUnits' | 'sellAmountBaseUnits'> {
```
steveklebanoff, line 47: Nice use of `TakerRequest` type 👍

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -286,6 +285,7 @@ export class QuoteRequestor {
             Object.keys(this._rfqtAssetOfferings).map(async url => {
                 if (this._makerSupportsPair(url, makerAssetData, takerAssetData)) {
                     const requestParams = {
+                        apiKey: options.apiKey,
```
steveklebanoff, line 288: Curious - why did this have to get added?

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -112,7 +105,7 @@ export class QuoteRequestor {
         const _opts: RfqtRequestOpts = { ...constants.DEFAULT_RFQT_REQUEST_OPTS, ...options };
         assertTakerAddressOrThrow(_opts.takerAddress);
 
-        const ordersWithStringInts = await this._getQuotesAsync<SignedOrder>( // not yet BigNumber
+        const firmQuotes = await this._getQuotesAsync<FirmQuote>( // not yet BigNumber
```
steveklebanoff, line 108: `RFQTFirmQuote`?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2582/commits/7162935028a676a93d98e1360be07a7122fb5e56

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: It didn't, thank you for catching this.  This was the result of my confusing the presence of an `apiKey` in the `TakerRequest` in `@0x/quote-server` as indicating that the server expects the API key as a query parameter, but I've since realized that that `TakerRequest` structure actually represents the _output_ of request parsing, at which point the header and query params have been merged into one structure.

Addressed in https://github.com/0xProject/0x-monorepo/pull/2582/commits/f7cd7110ea7dae6023127b9b6cb674a0cf1e1735

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2582/commits/ad868af96e13f3efa68d5836d898c53a6a5f8252

packages/asset-swapper/package.json
feuGeneA: Yes.  The that quote-server version has been published, and this PR now refers to it (via [this commit](https://github.com/0xProject/0x-monorepo/pull/2582/commits/b91cc3781de259e8647792ccd51db688419f7309)), and the more  RFQT-specific types are now in use, via [this commit](https://github.com/0xProject/0x-monorepo/pull/2582/commits/ad868af96e13f3efa68d5836d898c53a6a5f8252) and [this one](https://github.com/0xProject/0x-monorepo/pull/2582/commits/7162935028a676a93d98e1360be07a7122fb5e56).

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -412,7 +412,7 @@ function createNativeOrder(fill: CollapsedFill): OptimizedMarketOrder {
 }
 
 export function createSignedOrdersFromRfqtIndicativeQuotes(
-    quotes: RfqtIndicativeQuoteResponse[],
+    quotes: IndicativeQuote[],
```
steveklebanoff, line 415: Why `IndicativeQuote` instead of `RFQTIndicativeQuote`?

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -50,7 +51,7 @@ export interface NativeFillData extends FillData {
 }
 
 export interface RfqtFillData extends FillData {
-    quote: RfqtIndicativeQuoteResponse;
+    quote: IndicativeQuote;
```
steveklebanoff, line 54: Why IndicativeQuote instead of RFQTIndicativeQuote?

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
feuGeneA: Good catch, thank you, fixed in https://github.com/0xProject/0x-monorepo/pull/2582/commits/c18f3f0b339adf3c2f315f363a8a6cf366f71104

packages/asset-swapper/src/utils/market_operation_utils/types.ts
feuGeneA: This code is actually unused, so I removed it in https://github.com/0xProject/0x-monorepo/pull/2582/commits/187dd2fdc3aab3901084466e47c0975e0331e426

packages/asset-swapper/test/quote_requestor_test.ts
```diff
@@ -54,7 +55,7 @@ describe('QuoteRequestor', async () => {
                 endpoint: 'https://1337.0.0.1',
                 requestApiKey: apiKey,
                 requestParams: expectedParams,
-                responseData: successfulOrder1,
+                responseData: { signedOrder: successfulOrder1, quoteExpiry: successfulOrder1.expirationTimeSeconds },
```
steveklebanoff, line 58: Nit but can we remove `quoteExpiry` from here so these are vanilla RFQT firm quotes?

PR #2590, opened 2020-05-27
UniswapV2Bridge

xianny:
## Description

Implement UniswapV2Bridge contract. Allows us to swap tokens through the UniswapV2 Router. 

## Testing instructions

Please review unit tests and comment if you think any cases should be added. 

## Types of changes

## Checklist:


contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
```diff
@@ -0,0 +1,140 @@
...
+        if (state.fromTokenAddress == toTokenAddress) { // NOT TESTED
+            LibERC20Token.transfer(state.fromTokenAddress, to, amount);
+            return BRIDGE_SUCCESS;
+        }
+
+        // Get our balance of `fromTokenAddress` token.
+        state.fromTokenBalance = IERC20Token(state.fromTokenAddress).balanceOf(address(this));
+
+        // Grant the Uniswap router an allowance.
+        LibERC20Token.approve(
```
dekz, line 82: There should be a `LibERC20Token .approveIfBelow` so we keep it as high as possible and only write the state (expensive) if we have to.

contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
```diff
@@ -0,0 +1,140 @@
...
+            amount,
+            // Convert `fromTokenAddress` to `toTokenAddress`.
+            path,
+            // Recipient is `to`.
+            to,
+            // Expires after this block.
+            block.timestamp
+        );
+
+        state.boughtAmount = amounts[0];
```
dekz, line 109:
```

| amounts | uint[] memory | The input token amount and all subsequent output token amounts.|

```
https://uniswap.org/docs/v2/smart-contracts/router/#swapexacttokensfortokens

does this mean the first element is amountSold?

dorothy-zbornak: Yeah seems like we want the second element here.

dorothy-zbornak: Could also just do `state.boughtAmount = router.swapExactTokensForTokens(...)[1]`.

contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
```diff
@@ -0,0 +1,140 @@
...
+
+
+// solhint-disable space-after-comma
+// solhint-disable not-rely-on-time
+contract UniswapV2Bridge is
+    IERC20Bridge,
+    IWallet,
+    DeploymentConstants
+{
+    using LibAddressArray for address[];
```
dorothy-zbornak, line 39: Think you don't need this.

contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
```diff
@@ -0,0 +1,140 @@
...
+        LibERC20Token.approve(
+            state.fromTokenAddress,
+            _getUniswapV2Router01Address(),
+            state.fromTokenBalance
+        );
+
+        // Convert directly from fromTokenAddress to toTokenAddress
+        address[] memory path = new address[](2);
+        path[0] = state.fromTokenAddress;
+        path[1] = toTokenAddress;
```
dorothy-zbornak, line 91: I thought there would be multihop support out of the gate? Should we just encode an `address[] path` as the bridge data?

xianny: Ya let's do that. I'll change it.

contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
xianny: okiedokie will switch

contracts/asset-proxy/contracts/src/bridges/UniswapV2Bridge.sol
xianny: Oops my bad. Looks like we want the second element (last element if more than 2 tokens in the path).

PR #2592, opened 2020-06-02
Exchange Proxy: TransformerDeployer

dorothy-zbornak:
## Description

A contract for deploying and killing ERC20 transformers. Only authorized addresses can call `deploy()` or `kill()`.

## Checklist:


contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
```diff
@@ -0,0 +1,226 @@
...
+            keccak256(bytes('TransformerDeployer')),
+            keccak256(bytes('1.0.0')),
+            chainId,
+            address(this)
+        ));
+        _addMembers(members);
+    }
+
+    /// @dev Registers new members. Must attach majority votes.
+    function addMembers(Vote[] memory votes, bytes[] memory signatures)
```
mzhu25, line 93: I'd prefer that we change the `Vote` struct to:
```
    struct Vote {
        uint256 expirationTime;
        bytes signature;
    }
```
and the function signature to:
```suggestion
    function addMembers(address[] memory members, Vote[] memory votes)
```
so that we don't need to check that the `members` hash is the same for all the votes

contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
```diff
@@ -0,0 +1,226 @@
...
+        require(votes.length >= memberCount / 2 + 1, "TransformerDeployer/INSUFFICIENT_VOTES");
+        bytes32 membersHash = keccak256(abi.encode(votes[0].members));
+        address[] memory signers = new address[](votes.length);
+        for (uint256 i = 0; i < votes.length; ++i) {
+            // Ensure the vote isn't expired.
+            require(votes[i].expirationTime > block.timestamp, "TransformerDeployer/VOTE_EXPIRED");
+            // Ensure the members are the same across all votes.
+            require(
+                membersHash == keccak256(abi.encode(votes[i].members)),
+                "TransformerDeployer/NONHOMOGENOUS_VOTE"
```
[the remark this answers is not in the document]
dorothy-zbornak: omg I feel like I've been spelling this wrong my entire life

contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
dorothy-zbornak: moot now, since we're now just going with  a regular authorizable pattern

PR #2593, opened 2020-06-02
Add MultiBridge support to AssetSwapper

mzhu25:
## Description

- Adds a function to the Sampler contract to sample from MultiBridge
- Adds corresponding sampler operation to AS
- Some other MB-specific logic in AS
- Regenerates contract artifacts/wrappers

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -81,29 +82,45 @@ export class MarketOperationUtils {
...
             // Get ETH -> maker token price.
             DexOrderSampler.ops.getMedianSellRate(
-                difference(FEE_QUOTE_SOURCES, _opts.excludedSources).concat(
-                    this._liquidityProviderSourceIfAvailable(_opts.excludedSources),
+                difference(
+                    FEE_QUOTE_SOURCES.concat(
+                        this._liquidityProviderRegistry ? [ERC20BridgeSource.LiquidityProvider] : [],
+                        this._multiBridgeRegistry ? [ERC20BridgeSource.MultiBridge] : [],
+                    ),
+                    _opts.excludedSources,
```
dorothy-zbornak, line 102: Should probably just put this in a variable since it's used twice.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -161,29 +179,45 @@ export class MarketOperationUtils {
...
                 ),
```
dorothy-zbornak, line 200: ditto

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -648,14 +651,14 @@ describe('MarketOperationUtils tests', () => {
                     { excludedSources: SELL_SOURCES, numSamples: 4, bridgeSlippage: 0, shouldBatchBridgeOrders: false },
                 );
                 expect(result.length).to.eql(1);
-                expect(result[0].makerAddress).to.eql(liquidityProviderAddress);
+                expect(result[0].makerAddress).to.eql(DEX_FORWARDER_BRIDGE_ADDRESS);
```
dorothy-zbornak, line 654: Why does this happen? `shouldBatchBridgeOrders` is `false`.

dorothy-zbornak:
Can you explain to me the reasoning behind the MB registry? Will there be different MB contracts for different pairs?

packages/asset-swapper/test/market_operation_utils_test.ts
mzhu25: ah yes that would explain the failing test, idk how this slipped in

mzhu25:
> Can you explain to me the reasoning behind the MB registry? Will there be different MB contracts for different pairs?
probably not –– I decided on the registry layer so we could swap out MB implementations without disrupting the API. Also considered using a proxy pattern, but decided against it for PLP-related security reasons

dorothy-zbornak:
> > Can you explain to me the reasoning behind the MB registry? Will there be different MB contracts for different pairs?
> 
> probably not –– I decided on the registry layer so we could swap out MB implementations without disrupting the API. Also considered using a proxy pattern, but decided against it for PLP-related security reasons

I can appreciate seamless updates, but I wonder if it's easier to update AS + API than hunt down the person with the MB registry private key. I guess it could be worth it if MB is getting frequent updates. Not a blocker, either way.

dorothy-zbornak:
> > > Can you explain to me the reasoning behind the MB registry? Will there be different MB contracts for different pairs?
> > 
> > 
> > probably not –– I decided on the registry layer so we could swap out MB implementations without disrupting the API. Also considered using a proxy pattern, but decided against it for PLP-related security reasons
> 
> I can appreciate seamless updates, but I wonder if it's easier to update AS + API than hunt down the person with the MB registry private key. I guess it could be worth it if MB is getting frequent updates. Not a blocker, either way.

Actually now I'm slightly leaning against the registry. I think there might be value in AS/API opting into a version of MB. This gives us the ability to easily rollback to an older MB if necessary and also makes it easier to AB test one MB integration vs another.

mzhu25:
yeah that makes sense. will update

packages/contract-addresses/addresses.json
```diff
@@ -20,7 +20,7 @@
         "devUtils": "0x74134cf88b21383713e096a5ecf59e297dc7f547",
         "erc20BridgeProxy": "0x8ed95d1746bf1e4dab58d8ed4724f1ef95b20db0",
         "uniswapBridge": "0x36691c4f426eb8f42f150ebde43069a31cb080ad",
-        "erc20BridgeSampler": "0xc0154b14cc60a6661171fdc817f759429a57e184",
+        "erc20BridgeSampler": "0xd7223e59a11bbeb663c77625f06963ff42fdf97d",
```
dekz, line 23: Can we get this source verified or redeploy if the artifact is lost prior to merge?

mzhu25: I'll redeploy, verify, and update before merging

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -75,35 +77,37 @@ export class MarketOperationUtils {
         }
         const _opts = { ...DEFAULT_GET_MARKET_ORDERS_OPTS, ...opts };
         const [makerToken, takerToken] = getNativeOrderTokens(nativeOrders[0]);
+        const optionalSources = (this._liquidityProviderRegistry ? [ERC20BridgeSource.LiquidityProvider] : []).concat(
+            this._multiBridge === NULL_ADDRESS ? [ERC20BridgeSource.MultiBridge] : [],
```
dekz, line 81: should this be
```suggestion
            this._multiBridge !== NULL_ADDRESS ? [ERC20BridgeSource.MultiBridge] : [],
```

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -47,14 +47,16 @@ async function getRfqtIndicativeQuotesAsync(
...
 export class MarketOperationUtils {
     private readonly _wethAddress: string;
+    private readonly _multiBridge: string;
 
     constructor(
         private readonly _sampler: DexOrderSampler,
         private readonly contractAddresses: ContractAddresses,
         private readonly _orderDomain: OrderDomain,
-        private readonly _liquidityProviderRegistry: string = NULL_ADDRESS,
+        private readonly _liquidityProviderRegistry?: string,
```
dekz, line 56: This change does seem a little inconsistent with the following multiBridge. What are your thoughts on this change?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -76,51 +77,38 @@ export class MarketOperationUtils {
         }
         const _opts = { ...DEFAULT_GET_MARKET_ORDERS_OPTS, ...opts };
         const [makerToken, takerToken] = getNativeOrderTokens(nativeOrders[0]);
+        const optionalSources = (this._liquidityProviderRegistry !== NULL_ADDRESS
+            ? [ERC20BridgeSource.LiquidityProvider]
+            : []
+        ).concat(this._multiBridge !== NULL_ADDRESS ? [ERC20BridgeSource.MultiBridge] : []);
```
dorothy-zbornak, line 83: I wonder if we should just make `optionalSources` a `readonly` field so we only see this ugly bit of code once.

PR #2594, opened 2020-06-03
Exchange Proxy: AffiliateFeeTransformer

dorothy-zbornak:
## Description

These are flat fees instead of percentage-based. Though I'm thinking AS would accept the affiliate fee as a % and it'll be approximated as a safe lower-bound. Flat fees seem less likely to botch a quote fill due to precision errors, especially if the fee is in ETH.

## Checklist:


contracts/zero-ex/contracts/src/transformers/AffiliateFeeTransformer.sol
```diff
@@ -0,0 +1,90 @@
...
+    )
+        external
+        override
+        returns (bytes memory rlpDeploymentNonce)
+    {
+        TransformData memory data = abi.decode(data_, (TransformData));
+
+        // All arrays must be the same length.
+        if (data.tokens.length != data.amounts.length ||
+            data.tokens.length != data.recipients.length)
```
mzhu25, line 75: an array of structs seems like it would be cleaner here

contracts/zero-ex/contracts/src/transformers/AffiliateFeeTransformer.sol
```diff
@@ -0,0 +1,86 @@
...
+    )
+        external
+        override
+        returns (bytes4 success)
+    {
+        TokenFee[] memory fees = abi.decode(data, (TokenFee[]));
+
+        // Transfer tokens to recipients.
+        for (uint256 i = 0; i < fees.length; ++i) {
+            uint256 amount = fees[i].amount == MAX_UINT256
```
hysz, line 76: Building on the feedback from our last audit, we should probably use `if/else` opposed to the ternary operator. This will avoid unnecessarily evaluating `getTokenBalanceOf`.

PR #2597, opened 2020-06-04
Exchange Proxy: Address audit feedback (2)

dorothy-zbornak:
## Description

Aside from documentation stuff:

### TransformERC20 feature
- `IERC20Transformer.transform()` goes back to returning `TRANSFORMER_SUCCESS`.
- Instead of passing a transformer address into a `Transformation` struct, we pass a `uint32 deploymentNonce` and derive the transformer address before calling it.
- Make `AllowanceTarget.executeCall` *not* `payable`.
- Make `_transformERC20Private()` `private`.
- Add an `onlyOwner` `setTransformerDeployer()` function to upgrade the deployer.

## Checklist:

### DIFF
--- a/contracts/asset-proxy/CHANGELOG.json
+++ b/contracts/asset-proxy/CHANGELOG.json
@@ -102,7 +102,7 @@
             },
             {
                 "note": "Add `setOperators()` to `IDydx`",
-                "pr": "TODO"
+                "pr": 2462
             }
         ],
         "timestamp": 1581204851
--- a/contracts/zero-ex/contracts/src/features/TransformERC20.sol
+++ b/contracts/zero-ex/contracts/src/features/TransformERC20.sol
@@ -44,13 +44,19 @@ contract TransformERC20 is
     FixinCommon
 {
 
+    /// @dev Stack vars for `_transformERC20Private()`.
+    struct TransformERC20PrivateState {
+        IFlashWallet wallet;
+        address transformerDeployer;
+        uint256 takerOutputTokenBalanceBefore;
+        uint256 takerOutputTokenBalanceAfter;
+    }
+
     // solhint-disable
     /// @dev Name of this feature.
     string public constant override FEATURE_NAME = "TransformERC20";
     /// @dev Version of this feature.
     uint256 public immutable override FEATURE_VERSION = _encodeVersion(1, 0, 0);
-    /// @dev The trusted deployer for all transformers.
-    address public immutable transformDeployer;
     /// @dev The implementation address of this feature.
     address private immutable _implementation;
     // solhint-enable
@@ -58,14 +64,14 @@ contract TransformERC20 is
     using LibSafeMathV06 for uint256;
     using LibRichErrorsV06 for bytes;
 
-    constructor(address trustedDeployer_) public {
+    constructor() public {
         _implementation = address(this);
-        transformDeployer = trustedDeployer_;
     }
 
     /// @dev Initialize and register this feature.
     ///      Should be delegatecalled by `Migrate.migrate()`.
-    function migrate() external returns (bytes4 success) {
+    /// @param transformerDeployer The trusted deployer for transformers.
+    function migrate(address transformerDeployer) external returns (bytes4 success) {
         ISimpleFunctionRegistry(address(this))
             .extend(this.getTransformerDeployer.selector, _implementation);
         ISimpleFunctionRegistry(address(this))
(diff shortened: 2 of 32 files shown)


contracts/asset-proxy/CHANGELOG.json
```diff
@@ -98,7 +98,7 @@
             },
             {
                 "note": "Add `setOperators()` to `IDydx`",
-                "pr": "TODO"
+                "pr": 2462
```
dorothy-zbornak, line 101: Retroactively fixing this changelog.

contracts/utils/contracts/src/v06/AuthorizableV06.sol
```diff
@@ -35,13 +35,13 @@ contract AuthorizableV06 is
...
+    // @dev Whether an address is authorized to call privileged functions.
+    // @param 0 Address to query.
+    // @return 0 Whether the address is authorized.
     mapping (address => bool) public override authorized;
-    /// @dev Whether an address is authorized to call privileged functions.
-    /// @param 0 Index of authorized address.
-    /// @return 0 Authorized address.
+    // @dev Whether an address is authorized to call privileged functions.
+    // @param 0 Index of authorized address.
+    // @return 0 Authorized address.
```
dorothy-zbornak, line 44: solidity 0.6.9 now freaks out if you try to devdoc public storage vars :cry:

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -44,51 +44,71 @@ contract TransformERC20 is
...
+    constructor() public {
         _implementation = address(this);
-        transformDeployer = trustedDeployer_;
     }
 
     /// @dev Initialize and register this feature.
     ///      Should be delegatecalled by `Migrate.migrate()`.
-    function migrate() external returns (bytes4 success) {
+    /// @param transformerDeployer The trusted deployer for transformers.
+    function migrate(address transformerDeployer) external returns (bytes4 success) {
```
mzhu25, line 74: `@return`

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -44,51 +44,71 @@ contract TransformERC20 is
...
         ISimpleFunctionRegistry(address(this))
             .extend(this._transformERC20.selector, _implementation);
         createTransformWallet();
+        LibTransformERC20Storage.getStorage().transformerDeployer = transformerDeployer;
         return LibMigrate.MIGRATE_SUCCESS;
     }
 
+    /// @dev Replace the allowed deployer for transformers.
+    ///      Only callable by the owner.
+    function setTransformerDeployer(address transformerDeployer)
```
mzhu25, line 94: `@param`

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -316,20 +348,27 @@ contract TransformERC20 is
...
     function _executeTransformation(
         IFlashWallet wallet,
         Transformation memory transformation,
+        address transformerDeployer,
         address payable taker,
         bytes32 callDataHash
     )
         private
     {
+        // Derive the transformer address from the deployment nonce.
```
mzhu25, line 363: so this means that if we ever replace the `transformerDeployer` address, all transformers deployed by the old deployer become inaccessible?

dorothy-zbornak: Yup. We'll have to redeploy the existing ones.

mzhu25: what was the rationale for this change?

dorothy-zbornak: The code here or the deployer being replaceable?

mzhu25: well the deployer being replaceable and the fact that old deployments become defunct once a deployer is replaced

dorothy-zbornak: I decided to make it replaceable based off internal discussions and the uncertainty over how we want to manage the deployer at this stage. Previously we would have had to redeploy this entire feature. The fact that the transformers would die with the deployer is not new, because we always used the nonce to assert validity. This might actually be a good thing since it's a universal off switch for transformers if the deployer account is compromised. Ideally we won't be changing the deployer often, and we'll probably only have a handful of transformers early on.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -220,31 +241,44 @@ contract TransformERC20 is
                 .getSpendableERC20BalanceOf(inputToken, taker);
         }
 
-        IFlashWallet wallet = getTransformWallet();
+        TransformERC20PrivateState memory state;
```
hysz, line 244: Was this to work around a stack too deep error? The general downside to this approach is that we're now declaring w/o initializing, so a developer may accidentally reference a field before it's been populated.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -316,20 +350,27 @@ contract TransformERC20 is
...
     )
         private
     {
+        // Derive the transformer address from the deployment nonce.
+        address payable transformer = LibERC20Transformer.getDeployedAddress(
+            transformerDeployer,
+            transformation.deploymentNonce
+        );
         // Call `transformer.transform()` as the wallet.
         bytes memory resultData = wallet.executeDelegateCall(
```
hysz, line 371: Optional. If this will be reused then could be good to encapsulate in a library, like `LibFlashWallet`, that both executes the call and validates the return value.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: Yeah. I don't think there's a way around that problem.

PR #2600, opened 2020-06-11
Update wrappers and artifacts for ERC20BridgeSampler

xianny:
## Description
These are an indirect dependency of asset-swapper. Updating here to publish on gitpkg so they can be used in 0x-API.

## Checklist:


xianny:
@mzhu25 is doing this in https://github.com/0xProject/0x-monorepo/pull/2593 instead

PR #2608, opened 2020-06-19
feat: ExchangeProxy FillQuoteTransformer bridge direct

dekz:
## Description

Directly fills the Bridge order with the on-chain source.


contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +392,105 @@ contract FillQuoteTransformer is
...
-                    results.takerTokenSoldAmount.safeAdd(fillResults.takerFeePaid);
+        bytes4 makerAssetProxyId = LibBytesV06.readBytes4(order.makerAssetData, 0);
+        // If it is a Bridge order we fill this directly
+        // rather than filling via 0x Exchange
+        if (makerAssetProxyId == ERC20_BRIDGE_PROXY_ID) {
+            try
+                this.fillBridgeOrder(order,
+                    signature,
+                    takerAssetFillAmount,
+                    makerToken)
```
dekz, line 403: we intentionally ignore fee related settlement, so no need to adjust `takerTokenSoldAmount`

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +392,109 @@ contract FillQuoteTransformer is
...
+    ///      or the amount filled was not sufficient this reverts.
+    /// @param order The bridge order to fill.
+    /// @param inputTokenAmount How much taker asset to fill.
+    /// @param outputTokenAmount How much maker asset to receive.
+    /// @param makerToken The maker token.
+    function fillBridgeOrder(
+        IExchange.Order calldata order,
+        uint256 inputTokenAmount,
+        uint256 outputTokenAmount,
+        IERC20TokenV06 makerToken
```
dekz, line 452: We intentionally keep the expected makerToken here rather than use the derived one in the `makerAssetData`.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +392,109 @@ contract FillQuoteTransformer is
...
+            address bridgeAddress,
+            bytes memory bridgeData
+        ) = abi.decode(
+            LibBytesV06.sliceDestructive(
+                order.makerAssetData,
+                4,
+                order.makerAssetData.length),
+            (address, address, bytes)
+        );
+        require(bridgeAddress != address(this), "INVALID_BRIDGE_ADDRESS");
```
dekz, line 471: I'm open to better revert reasons here, but ultimately these are just caught and ignored, not surfaced to the user.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +387,110 @@ contract FillQuoteTransformer is
...
-                .safeSub(initialMakerTokenBalance);
-            // We can trust the other fill result quantities.
-            results.protocolFeePaid = fillResults.protocolFeePaid;
-            results.takerTokenSoldAmount = fillResults.takerAssetFilledAmount;
-            // If the taker fee is payable in the taker asset, include the
-            // taker fee in the total amount sold.
-            if (isTakerFeeInTakerToken) {
-                results.takerTokenSoldAmount =
-                    results.takerTokenSoldAmount.safeAdd(fillResults.takerFeePaid);
+        bytes4 makerAssetProxyId = order.makerAssetData.readBytes4(0);
```
dekz, line 390: Might need to bump this up to `_sellToOrder` and `_buyFromOrder` to prevent adjustments in amounts if fees are set on an order. Otherwise the behaviour is pretty undefined.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -161,13 +168,6 @@ contract FillQuoteTransformer is
                 }
             }
 
-            // Ensure we have enough ETH to cover the protocol fee.
```
dekz, line 164: Discussion: If we continue rather than revert we can attempt to handle the gas price adjustment with a fallback. If we revert explicitly early we cannot.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +387,108 @@ contract FillQuoteTransformer is
...
+            address(this),
+            outputTokenAmount, // amount to transfer back from the bridge
+            bridgeData
+        );
+        uint256 afterMakerTokenBalance = makerToken.balanceOf(address(this));
+        // Ensure that the maker token balance has increased by the expected amount
+        require(
+            afterMakerTokenBalance >= initialMakerTokenBalance.safeAdd(outputTokenAmount),
+            "BRIDGE_UNDERPAY"
+        );
```
dorothy-zbornak, line 486: What do we think of removing this check since we have the important one at the end anyway?

contracts/zero-ex/contracts/src/transformers/FillOrdersTransformer.sol
```diff
@@ -0,0 +1,436 @@
...
+        if (data.side == Side.Sell && data.fillAmount == MAX_UINT256) {
+            // If `sellAmount == -1 then we are selling
+            // the entire balance of `sellToken`. This is useful in cases where
+            // the exact sell amount is not exactly known in advance, like when
+            // unwrapping Chai/cUSDC/cDAI.
+            data.fillAmount = data.sellToken.getTokenBalanceOf(address(this));
+        }
+
+        // Approve the ERC20 proxy to spend `sellToken`.
+        data.sellToken.approveIfBelow(erc20Proxy, data.fillAmount);
```
dorothy-zbornak, line 143: Approval will be wrong for buys because `data.fillAmount` is in the maker asset. Will need to reflect this in the other transformer too.
```suggestion
 {
            uint256 initialSellTokenBalance =
                data.sellToken.getTokenBalanceOf(address(this));
            if (data.side == Side.Sell && data.fillAmount == MAX_UINT256) {
                // If `sellAmount == -1 then we are selling
                // the entire balance of `sellToken`. This is useful in cases where
                // the exact sell amount is not exactly known in advance, like when
                // unwrapping Chai/cUSDC/cDAI.
                data.fillAmount = initialSellTokenBalance;
            }
            // Approve the ERC20 proxy to spend `sellToken`.
            data.sellToken.approveIfBelow(erc20Proxy, initialSellTokenBalance);
        }
```

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dekz: We only check that we sold X amount or bought Y amount. This checks that we sold to the bridge at a reasonably expected price. If we didn't revert we could've sold 100 for 1 instead of 100 for 99. This needs to stay for safe bridge market sells.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -388,31 +387,108 @@ contract FillQuoteTransformer is
...
+                )
+            );
+            if (success) {
+                results = abi.decode(data, (FillOrderResults));
             }
-        } catch (bytes memory) {
             // Swallow failures, leaving all results as zero.
+        } else {
+            // Ensure we have enough ETH to cover the protocol fee.
+            if (address(this).balance < protocolFee) {
```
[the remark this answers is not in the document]
dekz: still cheaper then a fill which reverts at the end for protocol fee error  😂

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dorothy-zbornak: Won't the `transformERC2()` check assert the price?
https://github.com/0xProject/0x-monorepo/blob/development/contracts/zero-ex/contracts/src/features/TransformERC20.sol#L285

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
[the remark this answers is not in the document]
dorothy-zbornak: I mean can we just check the balance once and decrement it after every successful fill?

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dekz: Yeah looks like it 😂

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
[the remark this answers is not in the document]
dekz: Yeah we'd just need to pass down `ethRemaining` down to fillOrder. Only want to revert if it's a Native which is determined further down.

contracts/zero-ex/contracts/src/transformers/FillOrdersTransformer.sol
dekz: As discussed this shouldn't have an impact at the approval is set to UINT256.

dorothy-zbornak: I think we still need to make this change in case another transformer doesn't set the approval to unlimited.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -237,14 +258,14 @@ contract FillQuoteTransformer is
     /// @param order The order to fill.
     /// @param signature The signature for `order`.
     /// @param sellAmount Amount of taker token to sell.
-    /// @param protocolFee The protocol fee needed to fill `order`.
+    /// @param state The state.
```
dekz, line 261: @dekz make a better doc comment mate

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -372,47 +389,119 @@ contract FillQuoteTransformer is
...
+    ///      or the amount filled was not sufficient this reverts.
+    /// @param order The bridge order to fill.
+    /// @param inputTokenAmount How much taker asset to fill clamped to the available balance.
+    /// @param outputTokenAmount How much maker asset to receive.
+    /// @param makerToken The maker token.
+    function fillBridgeOrder(
+        IExchange.Order calldata order,
+        uint256 inputTokenAmount,
+        uint256 outputTokenAmount,
+        FillState calldata state,
```
dorothy-zbornak, line 467: All this really needs is `boughtAmount`, right? Args encoding is probably cheaper if it didn't have to encode a full struct.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -282,17 +303,15 @@ contract FillQuoteTransformer is
         }
 
         // Clamp fill amount to order size.
-        takerTokenFillAmount = LibSafeMathV06.min256(
-            takerTokenFillAmount,
-            order.takerAssetAmount
-        );
+        takerTokenFillAmount = takerTokenFillAmount.min256(order.takerAssetAmount);
+        takerTokenFillAmount = takerTokenFillAmount.min256(state.takerBalanceRemaining);
```
dorothy-zbornak, line 307: Should we just move these to `_fillOrder()`?

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -372,47 +389,119 @@ contract FillQuoteTransformer is
...
-                    results.takerTokenSoldAmount.safeAdd(fillResults.takerFeePaid);
+        // If it is a Bridge order we fill this directly
+        // rather than filling via 0x Exchange
+        if (order.makerAssetData.readBytes4(0) == ERC20_BRIDGE_PROXY_ID) {
+            uint256 outputTokenAmount = LibMathV06.getPartialAmountFloor(
+                takerAssetFillAmount,
+                order.takerAssetAmount,
+                order.makerAssetAmount
+            );
+            (bool success, bytes memory data) = address(_implementation).delegatecall(
```
dorothy-zbornak, line 416: spooked me for a second :sweat_smile:

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -372,47 +389,119 @@ contract FillQuoteTransformer is
...
+    /// @param makerToken The maker token.
+    function fillBridgeOrder(
+        IExchange.Order calldata order,
+        uint256 inputTokenAmount,
+        uint256 outputTokenAmount,
+        FillState calldata state,
+        IERC20TokenV06 makerToken
+    )
+        external
+        returns (FillOrderResults memory results)
```
dorothy-zbornak, line 471: Do we even really need a return value here? Can't we figure this all out in `fillOrder()` and avoid encoding/decoding a return value? Would also prefer to keep this function simple to audit since it executes as a delegetecall.

contracts/zero-ex/contracts/src/vendor/v3/IGasToken.sol
```diff
@@ -0,0 +1,37 @@
+/*
+
+  Copyright 2019 ZeroEx Intl.
```
dorothy-zbornak, line 3:
```suggestion
  Copyright 2020 ZeroEx Intl.
```

contracts/zero-ex/contracts/src/transformers/FillOrdersTransformer.sol
```diff
@@ -0,0 +1,436 @@
...
+import "@0x/contracts-utils/contracts/src/v06/LibSafeMathV06.sol";
+import "@0x/contracts-utils/contracts/src/v06/LibMathV06.sol";
+import "../errors/LibTransformERC20RichErrors.sol";
+import "../vendor/v3/IExchange.sol";
+import "./Transformer.sol";
+import "./LibERC20Transformer.sol";
+
+
+/// @dev A transformer that fills an ERC20 market sell/buy quote.
+contract FillOrdersTransformer is
```
dorothy-zbornak, line 35: I now kind of feel like we should just delete this contract lol. Just doesn't seem worth the headache to maintain them both when realistically no one is going to be using this. We can always resurrect it from history if the need arises.

dekz:
### Before
image

### After
image

### Diff
| source         | Exchange | FQT   | diff |
| -------------- | -------- | ----- | ---- |
| Curve_Compound | 1228K    | 1100K | 128K |
| Curve_SUSD     | 462K     | 434K  | 28K  |
| Kyber          | 893K     | 805K  | 88K  |
| Uniswap        | 366K     | 353K  | 13K  |
| UniswapV2      | 375K     | 354K  | 21K  |

contracts/zero-ex/contracts/src/fixins/FixinGasToken.sol
```diff
@@ -0,0 +1,46 @@
+/*
+
+  Copyright 2019 ZeroEx Intl.
```
mzhu25, line 3:
```suggestion
  Copyright 2020 ZeroEx Intl.
```

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
mzhu25: wait why do we need to do a delegatecall here? If it's just to swallow errors, I'd prefer to just swallow errors from the `transfer` and `bridgeTransferFrom` calls in `fillBridgeOrder` to save a bit of gas.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -456,16 +462,17 @@ contract FillQuoteTransformer is
...
+        bytes calldata makerAssetData,
+        bytes calldata takerAssetData,
         uint256 inputTokenAmount,
-        uint256 outputTokenAmount,
-        FillState calldata state,
-        IERC20TokenV06 makerToken
+        uint256 outputTokenAmount
     )
         external
         returns (FillOrderResults memory results)
```
dorothy-zbornak, line 478: not used

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dorothy-zbornak: it's because this contract is already a `delegatecall` from the `FlashWallet`, so a `call` to `this` would fail.

mzhu25: sure, but I mean can't we just have a private version of `fillBridgeOrder` which we use here? i.e. replace `(bool success, bytes memory data) = address(_implementation).delegatecall(...);` with: `results = _fillBridgeOrder(...);`

dorothy-zbornak: I think the issue is we want to revert the transfer if the `bridgeTransferFrom()` fails as well, so these need to be atomic.

mzhu25: ah got it, that's annoying... we should just translate all our bridges into transformers 🙃

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dekz: :see_no_evil:

PR #2610, opened 2020-06-20
EP: MetaTransactions

dorothy-zbornak:
## Description

Introduces two new features:
- `MetaTransactions`
  - Unlike the Exchange, we decode and translate the calldata from the external function call to the internal (`onlySelf`) version, where we can override the taker address, rather than just delegatecalling the raw calldata.
- `SignatureValidator`
  - We only support EIP712 and eth_sign signatures right now.

And one new fixin:
- `FixinEIP712`
  - Utilities for generating EIP712 hashes.

Other stuff:
- Refactor `address immutable _implementation` state variable into `FixinCommon`
- Refactor `_implementation = address(this)` out of feature constructors into `FixinCommon`.
- Add a `registerFeatureFunction(selector)` function to `FixinCommon`. This is shorthand for `        ISimpleFunctionRegistry(address(this)).extend(selector, _implementation)`.
- Tweak the flow of initial migration/bootstrapping slightly.
  - The `ZeroEx` contract is now deployed in a separate step and passed into `InitialMigration.migrate()`, whereas before `InitialMigration` would deploy that contract itself.
- Added tooling to `@0x/order-utils`, etc., for hashing and signing this new `ExchangeProxyMetaTransaction` type.
 
post-merge TODO:
  - Publish new ganache snapshot.
  - Update EP spec for updated initial migration/bootstrap process.

## Checklist:

- [x] Feature has updated version information.
- [x] implements IFeature interface.
- [x] Feature contracts are stateless (including inherited contracts).
- [x] onlySelf feature functions are prefixed with _.
- [x] Feature functions are added to full_migration_tests.
- [x] No delegatecalls from inside a feature. Call other features through the router.
- [x] No self-destruct in features.
- [x] No intentionally persistent (non-atomic) balances on the Exchange Proxy.
- [x] No allowances to be set on Exchange Proxy by users.
- [x] No direct access to another feature’s storage bucket without strong justification.
- [x] No direct access to AllowanceTarget (must go through TokenSpender).
- [x] No executing arbitrary calldata from the context of the Exchange Proxy.
- [x] Features use unique StorageIds. 
- [x] Document functions with execution contexts outside of the Exchange Proxy.
- [x] Document feature dependencies in checklist doc.
- [x] Document reentrant functions in checklist doc.
- [x] Document temporary balances.


contracts/zero-ex/contracts/src/ZeroEx.sol
```diff
@@ -34,11 +34,12 @@ contract ZeroEx {
...
+    ///      by `bootstrap()` to seed the initial feature set.
+    /// @param bootstrapper Who can call `bootstrap()`.
+    constructor(address bootstrapper) public {
         // Temporarily create and register the bootstrap feature.
         // It will deregister itself after `bootstrap()` has been called.
-        Bootstrap bootstrap = new Bootstrap(msg.sender);
+        Bootstrap bootstrap = new Bootstrap(bootstrapper);
         LibProxyStorage.getStorage().impls[bootstrap.bootstrap.selector] =
             address(bootstrap);
     }
```
dorothy-zbornak, line 45: Previously it was the job of `InitialMigration` to deploy the `ZeroEx` contract. Now we deploy the `ZeroEx` contract (along with all the features) in a separate tx, but we still want `InitialMigration` to be the one to wire them all together. So we need to be able to decouple the deployer of `ZeroEx` from the bootstrapper (`InitialMigration`).

dorothy-zbornak: Impetus for this change is that `FixinEIP712` wants the `ZeroEx` contract address to init its immutables, so we need* to have `ZeroEx` deployed beforehand.

*unless we want to predict the address via hash.

contracts/zero-ex/contracts/src/features/MetaTransactions.sol
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    function _executeTransformERC20Call(ExecuteState memory state)
+        private
+        returns (bytes memory returnData)
+    {
+        // HACK(dorothy-zbornak): `abi.decode()` with the individual args
+        // will cause a stack overflow. But we can prefix the call data with an
+        // offset to transform it into the encoding for the equivalent single struct arg.
+        // Decoding a single struct consumes far less stack space.
```
dorothy-zbornak, line 332: This caused me so much grief. Especially because solc would sometimes be able to get around the stack overflow if I just did a `yarn clean && yarn compile`.

This is definitely hacky so check my assumptions here. Alternatively, we could do the entire translation to a `_transformERC20()` call in assembly, but that seems more error prone since we're dealing with a nested array of structs. We can always come back to this later in an update.

mzhu25: this looks correct to me. wen EIP-663 tho 😭

contracts/zero-ex/test/features/meta_transactions_test.ts
```diff
@@ -0,0 +1,518 @@
...
+                        mtxHash,
+                        signers[0],
+                        signature,
+                    ).encode(),
+                ),
+            );
+        });
+    });
+
+    describe('executeMetaTransactions()', () => {
```
mzhu25, line 428: maybe add a check that you can't execute the same mtx twice via `executeMetaTransactions`

contracts/zero-ex/contracts/src/features/MetaTransactions.sol
```diff
@@ -0,0 +1,412 @@
...
+        state.selfBalance  = address(this).balance;
+        if (state.mtx.value > state.selfBalance) {
+            LibMetaTransactionsRichErrors
+                .MetaTransactionInsufficientEthError(
+                    state.hash,
+                    state.selfBalance,
+                    state.mtx.value
+                ).rrevert();
+        }
+        // Must be signed by signer.
```
mzhu25, line 321: should we skip the signature check if `state.sender == state.mtx.signer`?

dorothy-zbornak: Good point. We do skip it in the Exchange. I wonder if that's a vulnerability though since you can trick a mtx relayer into executing a transaction as itself, if they aren't paying attention.

mzhu25: yeah that might be true –– in that case I think we should just add a generic `batchExecute` feature

contracts/zero-ex/contracts/src/features/MetaTransactions.sol
```diff
@@ -0,0 +1,412 @@
...
+    function _executeTransformERC20Call(ExecuteState memory state)
+        private
+        returns (bytes memory returnData)
+    {
+        // HACK(dorothy-zbornak): `abi.decode()` with the individual args
+        // will cause a stack overflow. But we can prefix the call data with an
+        // offset to transform it into the encoding for the equivalent single struct arg.
+        // Decoding a single struct consumes far less stack space.
+        TransformERC20Args memory args;
+        {
```
hysz, line 359: I think this looks okay but I'll have to spend more time validating it. Could we add more documentation around the expected memory layouts? See [here](https://github.com/0xProject/0x-monorepo/blob/development/contracts/asset-proxy/contracts/src/ERC20Proxy.sol#L67) as an example.

contracts/zero-ex/contracts/src/ZeroEx.sol
dorothy-zbornak: No need to redeploy anything. Bootstrapper doesn't exist in the wild since it self-destructs.

contracts/zero-ex/contracts/src/features/IMetaTransactions.sol
```diff
@@ -0,0 +1,127 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-erc20/contracts/src/v06/IERC20TokenV06.sol";
+
+
+/// @dev Meta-transactions feature.
+interface IMetaTransactions {
+
+    /// @dev Describes an exchange proxy meta transaction.
+    struct MetaTransactionData {
```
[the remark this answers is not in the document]
dorothy-zbornak: The only optional fields are the fees I think.

contracts/zero-ex/contracts/src/migrations/FullMigration.sol
```diff
@@ -62,25 +66,39 @@ contract FullMigration {
...
     /// @param migrateOpts Parameters needed to initialize features.
     function deploy(
         address payable owner,
+        ZeroEx zeroEx,
         Features memory features,
         MigrateOpts memory migrateOpts
     )
         public
-        returns (ZeroEx zeroEx)
+        returns (ZeroEx _zeroEx)
```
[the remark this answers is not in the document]
dorothy-zbornak: Not really other than something I noticed during migrations is it's nice to have return values so you can feel confident that an eth_call succeeded. With old versions of geth it wasn't clear whether a function with no result  reverted or succeeded because both return a `0x`.

contracts/zero-ex/contracts/src/features/MetaTransactions.sol
```diff
@@ -0,0 +1,412 @@
...
+                .MetaTransactionGasPriceError(
+                    state.hash,
+                    tx.gasprice,
+                    state.mtx.minGasPrice,
+                    state.mtx.maxGasPrice
+                ).rrevert();
+        }
+        // Must have enough ETH.
+        state.selfBalance  = address(this).balance;
+        if (state.mtx.value > state.selfBalance) {
```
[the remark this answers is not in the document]
dorothy-zbornak: Yeah, this is specifically for batch transactions, since `msg.value` will probably no longer be accurate after each execution.

contracts/zero-ex/contracts/src/features/IMetaTransactions.sol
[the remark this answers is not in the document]
hysz: Oh interesting, I think we could probably make more of these optional. Not necessary for this PR, but for example:
- No `minGasPrice`/`maxGasPrice` range → no gas limitations.
- No `sender` → anyone can send.
- No `expirationTimeSeconds` → mtx doesn't expire.
- No `value` → taker doesn't send ETH.

I also realized that the entire `optional` struct would need to be encoded one a single optional field is set. So, the flat structure is probably better in the contracts anyway.

PR #2613, opened 2020-06-26
BalancerBridge

mzhu25:
## Description
Adds the BalancerBridge contract and corresponding AssetSwapper support. 
AssetSwapper uses the Balancer SOR package to get pool addresses and compute quotes off-chain. 
Also includes some related AssetSwapper refactors, most notably merging all Curves into a single `ERC20BridgeSource`. To account for the fact that different Curves have significantly different gas costs, we change the fee schedule so that each source is given a fee estimate function instead of a constant. 

## Types of changes

## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/curve_utils.ts
```diff
@@ -0,0 +1,9 @@
+import * as _ from 'lodash';
+
+import { MAINNET_CURVE_CONTRACTS } from './constants';
+
+export const getCurveAddressesForPair = (takerToken: string, makerToken: string): string[] => {
+    return Object.keys(
+        _.pickBy(MAINNET_CURVE_CONTRACTS, tokens => tokens.includes(takerToken) && tokens.includes(makerToken)),
```
xianny, line 7: nit: possible to use native Array.filter instead of lodash?

mzhu25: we need to filter out keys based on their values, is there a way to do that with Array.filter?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -343,181 +381,199 @@ export const samplerOperations = {
...
+                                const intermediateToken = getMultiBridgeIntermediateToken(takerToken, makerToken);
+                                return samplerOperations.getMultiBridgeSellQuotes(
+                                    multiBridgeAddress,
+                                    makerToken,
+                                    intermediateToken,
+                                    takerToken,
+                                    takerFillAmounts,
+                                );
+                            case ERC20BridgeSource.Balancer:
+                                const pools = await getBalancerPoolsForPairAsync(takerToken, makerToken);
```
dekz, line 456: this'll probably throw if TheGraph is down. We should handle this so it doesn't take down the entire sampling operation.

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    weightIn: BigNumber;
+    weightOut: BigNumber;
+    swapFee: BigNumber;
+    spotPrice?: BigNumber;
+    slippage?: BigNumber;
+    limitAmount?: BigNumber;
+}
+
+export const getBalancerPoolsForPairAsync = async (takerToken: string, makerToken: string): Promise<BalancerPool[]> => {
+    return parsePoolData(await getPoolsWithTokens(takerToken, makerToken), takerToken, makerToken);
```
dorothy-zbornak, line 18: the plan is to cache this, right?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,9 +1,25 @@
...
+    BalancerFillData,
+    BatchedOperation,
+    CurveFillData,
+    DexSample,
+    FakeBuyOpts,
+    SourceQuoteOperation,
+    UniswapV2FillData,
+} from './types';
 
 /**
```
dorothy-zbornak, line 24: Starting to feel like this needs to be a class so it can hold state like registry addresses and cached balancer pools and calls to some of these ops can be more legible. I think it'll clean up some of the tests as well.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -273,19 +297,33 @@ export const samplerOperations = {
             },
         };
     },
-    getMedianSellRate(
+    getBalancerSellQuotes(pool: BalancerPool, takerFillAmounts: BigNumber[]): SourceQuoteOperation<BalancerFillData> {
+        return {
+            source: ERC20BridgeSource.Balancer,
+            fillData: { poolAddress: pool.id },
+            ...samplerOperations.constant(takerFillAmounts.map(_.curry(computeBalancerSellQuote)(pool))),
```
dorothy-zbornak, line 304: why not drop the lodash and just `map(amt => computeBalancerSellQuote(pool, amt))`?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -273,19 +297,33 @@ export const samplerOperations = {
...
+        };
+    },
+    getBalancerBuyQuotes(pool: BalancerPool, makerFillAmounts: BigNumber[]): SourceQuoteOperation<BalancerFillData> {
+        return {
+            source: ERC20BridgeSource.Balancer,
+            fillData: { poolAddress: pool.id },
+            ...samplerOperations.constant(makerFillAmounts.map(_.curry(computeBalancerBuyQuote)(pool))),
```
dorothy-zbornak, line 311: 92f1a91a67bcf75871e0b2ff518049175b89b7ff_128

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -343,181 +381,199 @@ export const samplerOperations = {
...
+                                    makerFillAmounts,
+                                    fakeBuyOpts,
+                                );
+                            case ERC20BridgeSource.Balancer:
+                                const pools = await getBalancerPoolsForPairAsync(takerToken, makerToken);
+                                return pools.map(pool =>
+                                    samplerOperations.getBalancerBuyQuotes(pool, makerFillAmounts),
+                                );
+                            default:
+                                throw new Error(`Unsupported sell sample source: ${source}`);
```
dorothy-zbornak, line 554:
```suggestion
                                throw new Error(`Unsupported buy sample source: ${source}`);
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -29,16 +29,12 @@ export enum ERC20BridgeSource {
...
     Uniswap = 'Uniswap',
     UniswapV2 = 'Uniswap_V2',
-    UniswapV2Eth = 'Uniswap_V2_ETH',
     Eth2Dai = 'Eth2Dai',
     Kyber = 'Kyber',
-    CurveUsdcDai = 'Curve_USDC_DAI',
-    CurveUsdcDaiUsdt = 'Curve_USDC_DAI_USDT',
-    CurveUsdcDaiUsdtTusd = 'Curve_USDC_DAI_USDT_TUSD',
-    CurveUsdcDaiUsdtBusd = 'Curve_USDC_DAI_USDT_BUSD',
-    CurveUsdcDaiUsdtSusd = 'Curve_USDC_DAI_USDT_SUSD',
```
dorothy-zbornak, line 39: Unbelievably excited to see these squashed.

packages/asset-swapper/src/utils/market_operation_utils/constants.ts
```diff
@@ -60,56 +50,45 @@ export const DEFAULT_FAKE_BUY_OPTS: FakeBuyOpts = {
 /**
  * Sources to poll for ETH fee price estimates.
  */
-export const FEE_QUOTE_SOURCES = SELL_SOURCES;
+export const FEE_QUOTE_SOURCES = [
+    ERC20BridgeSource.Uniswap,
+    ERC20BridgeSource.UniswapV2,
+    ERC20BridgeSource.Eth2Dai,
+    ERC20BridgeSource.Kyber,
+    ERC20BridgeSource.Curve,
```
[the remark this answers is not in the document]
mzhu25: yeah no Balancer on purpose, I figured making the GraphQL query was kinda overkill for estimating the ETH price

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,41 @@
...
+    balanceOut: BigNumber;
+    weightIn: BigNumber;
+    weightOut: BigNumber;
+    swapFee: BigNumber;
+    spotPrice?: BigNumber;
+    slippage?: BigNumber;
+    limitAmount?: BigNumber;
+}
+
+export const getBalancerPoolsForPairAsync = async (takerToken: string, makerToken: string): Promise<BalancerPool[]> => {
```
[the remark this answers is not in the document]
mzhu25: o yeah I was going to fill in the docs, I was just procrastinating :p

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -49,13 +45,28 @@ export interface NativeFillData extends FillData {
...
+
+export interface UniswapV2FillData extends FillData {
+    tokenAddressPath: string[];
+}
+
 /**
  * Represents an individual DEX sample from the sampler contract.
  */
-export interface DexSample {
+export interface DexSample<TFillData extends FillData = FillData> {
```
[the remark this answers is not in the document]
mzhu25: I feel like having the type parameter default here is nice, since most sources (Uniswap, Eth2Dai, Kyber, LiquidityProvider, MultiBridge) aren't using the optional `fillData` field for now. Saves some `<Filldata>`s

packages/asset-swapper/src/utils/quote_simulation.ts
```diff
@@ -148,12 +145,12 @@ export function fillQuoteOrders(
         if (remainingInput.lte(0)) {
             break;
         }
+        result.gas += getTotalGasUsedByFills(fo.order.fills, gasSchedule);
```
dorothy-zbornak, line 148: Why was this calculation moved here? Isn't this wrong? We might not need to hit all the fills/sources.

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -13,12 +13,43 @@ export interface BalancerPool {
...
+const THIRTY_MINUTES_MS = 30 * 60 * 1000; // tslint:disable-line:custom-no-magic-numbers
+export class BalancerPoolsCache {
+    constructor(
+        private readonly _cache: { [key: string]: CacheValue } = {},
+        public cacheExpiryMs: number = THIRTY_MINUTES_MS,
+    ) {}
+    public async getPoolsForPairAsync(takerToken: string, makerToken: string): Promise<BalancerPool[]> {
+        const key = JSON.stringify([takerToken, makerToken].sort());
+        const value = this._cache[key];
+        const minTimestamp = new Date().getTime() - this.cacheExpiryMs;
```
dorothy-zbornak, line 29:
```suggestion
        const minTimestamp = Date.now() - this.cacheExpiryMs;
```

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -13,12 +13,43 @@ export interface BalancerPool {
...
+        if (value.timestamp < minTimestamp) {
+            const pools = await this._fetchPoolsForPairAsync(takerToken, makerToken);
+            const timestamp = new Date().getTime();
```
dorothy-zbornak, line 32:
```suggestion
            const timestamp = Date.now();
```

packages/asset-swapper/src/utils/quote_simulation.ts
mzhu25: oops, you're right

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    pools: BalancerPool[];
+}
+const THIRTY_MINUTES_MS = 30 * 60 * 1000; // tslint:disable-line:custom-no-magic-numbers
+export class BalancerPoolsCache {
+    constructor(
+        private readonly _cache: { [key: string]: CacheValue } = {},
+        public cacheExpiryMs: number = THIRTY_MINUTES_MS,
+    ) {}
+    public async getPoolsForPairAsync(takerToken: string, makerToken: string): Promise<BalancerPool[]> {
+        const key = JSON.stringify([takerToken, makerToken].sort());
```
dekz, line 27: is `balanceIn` relative to the first argument in fetchPoolsForPair? And sorting these would result in an invalid rate in the opposite direction?

E.g 

query `DAI, WETH`, `balanceIn` is DAI. Result is stored as `DAI, WETH`. 
query `WETH, DAI`, `balanceIn` is WETH`. Result is already stored and retrieved from `DAI, WETH`

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    swapFee: BigNumber;
+    spotPrice?: BigNumber;
+    slippage?: BigNumber;
+    limitAmount?: BigNumber;
+}
+interface CacheValue {
+    timestamp: number;
+    pools: BalancerPool[];
+}
+const THIRTY_MINUTES_MS = 30 * 60 * 1000; // tslint:disable-line:custom-no-magic-numbers
```
dekz, line 20: IMO 30 mins is fine for a cache of existing pools. I.e a new pool added takes 30m till it is served by AssetSwapper.

30 mins is too long for rates within the pool and determining the route based on these values. Especially with sub 1% slippage.

I'd suggest we cache for at most a minute and move towards a model where the pools are discovered (cached) but queried on-demand.

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,76 @@
...
+            const poolData = (await getPoolsWithTokens(takerToken, makerToken)).pools;
+            return parsePoolData(poolData, takerToken, makerToken);
+        } catch (err) {
+            return [];
+        }
+    }
+}
+
+// tslint:disable completed-docs
+export function computeBalancerSellQuote(pool: BalancerPool, takerFillAmount: BigNumber): BigNumber {
```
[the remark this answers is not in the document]
mzhu25: I have the short circuit in `computeBalancerBuyQuote` but I figured you can _sell_ any amount to the pool (at least in theory?)
Not sure what's going on in `bpowApprox`, will investigate

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
mzhu25: ah yeah you're right. I'll just add some logic to store two cache entries per pair (one for [takerToken, makerToken], one for [makerToken, takerToken])

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
mzhu25: Agreed. I wonder if it'd be worth differentiating pool discovery vs pool state. idk if there would be a substantial difference in latency between querying the state of the cached pool addresses versus calling `getPoolsWithTokens`

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -0,0 +1,76 @@
...
+    constructor(
+        private readonly _cache: { [key: string]: CacheValue } = {},
+        public cacheExpiryMs: number = THIRTY_MINUTES_MS,
+    ) {}
+    public async getPoolsForPairAsync(takerToken: string, makerToken: string): Promise<BalancerPool[]> {
+        const key = JSON.stringify([takerToken, makerToken].sort());
+        const value = this._cache[key];
+        const minTimestamp = Date.now() - this.cacheExpiryMs;
+        if (value === undefined || value.timestamp < minTimestamp) {
+            const pools = await this._fetchPoolsForPairAsync(takerToken, makerToken);
```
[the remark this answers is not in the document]
mzhu25: yeah probably –– I guess top 10 by out balances makes sense?

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
mzhu25: yep, you're right

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
dekz: yeah I assume TheGraph is fairly quick, which is fine for pool discovery. 

But ultimately I think we'd want to get the rates as live as possible to reduce our potential for revert.

packages/utils/src/configured_bignumber.ts
```diff
@@ -34,4 +34,7 @@ if (isNode) {
     };
 }
 
+const orig = BigNumber.config;
+BigNumber.config = (_args?: any) => orig({});
```
dorothy-zbornak, line 38: Should probably document why we do this and also clobber `BigNumber.set()`.

Also, maybe this is neater?
```ts
(orig => BigNumber.config = (...args: any) => orig({}))(BigNumber.config);
BigNumber.set = BigNumber.config;
```

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -567,18 +572,22 @@ export const samplerOperations = {
                 ),
             ),
         );
-
+        const samplerOps = subOps.filter(op => op.source !== ERC20BridgeSource.Balancer);
+        const nonSamplerOps = subOps.filter(op => op.source === ERC20BridgeSource.Balancer);
```
dorothy-zbornak, line 576: Is this change just to reduce the load on the sampler contract?

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -41,7 +45,13 @@ export class BalancerPoolsCache {
...
     protected async _fetchPoolsForPairAsync(takerToken: string, makerToken: string): Promise<BalancerPool[]> {
         try {
-            return parsePoolData(await getPoolsWithTokens(takerToken, makerToken), takerToken, makerToken);
+            const poolData = (await getPoolsWithTokens(takerToken, makerToken)).pools;
+            // Sort by maker token balance (descending)
+            const pools = parsePoolData(poolData, takerToken, makerToken).sort((a, b) =>
+                b.balanceOut.minus(a.balanceOut).toNumber(),
+            );
+            // tslint:disable-next-line:custom-no-magic-numbers
+            return pools.length > 10 ? pools.slice(0, 10) : pools;
```
dorothy-zbornak, line 54: I think it's worth putting `10` into a constant.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
mzhu25: the sampler was reverting because `batchCall` didn't know what to do with empty bytes. we could change `batchCall` instead to skip over if `callDatas[i] == 0`, but 🤷

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -3,11 +3,13 @@ import { BigNumber } from '@0x/utils';
 import { MarketOperation } from '../../types';
 
 import { ZERO_AMOUNT } from './constants';
-import { getPathSize, isValidPath } from './fills';
+import { getPathAdjustedSize, getPathSize, isValidPath } from './fills';
 import { Fill } from './types';
 
 // tslint:disable: prefer-for-of custom-no-magic-numbers completed-docs
 
+const RUN_LIMIT_DECAY_FACTOR = 0.8;
```
dorothy-zbornak, line 11: Wonder how detrimental `0.5` would be...

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -102,31 +103,50 @@ export class MarketOperationUtils {
...
+
         const rfqtPromise = getRfqtIndicativeQuotesAsync(
             nativeOrders[0].makerAssetData,
             nativeOrders[0].takerAssetData,
             MarketOperation.Sell,
             takerAmount,
             _opts,
         );
+
+        const balancerPromise = this._sampler.executeAsync(
```
dorothy-zbornak, line 128: Seems like balancer shouldn't belong to the sampler anymore?

PR #2620, opened 2020-07-01
geth eth_call support.

dorothy-zbornak:
## Description

This add the types necessary to support geth eth_calls in `web3-wrapper` and therefore our generated contract wrappers.

Example: 
```ts
const erc20 = new IERC20TokenContract(TOKEN_ADDRESS, provider, ...);
erc20.balanceOf(...).callAsync({
    overrides: {
        // Override some state of `TOKEN_ADDRESS` during this call to `balanceOf()`.
        // All fields are optional.
        [TOKEN_ADDRESS]: {
            // Override deployed bytecode at `TOKEN_ADDRESS`
            code: '0x....',
            // Override the ETH balance of `TOKEN_ADDRESS`
            balance: 12345,
            // Override the nonce of `TOKEN_ADDRESS`
            nonce: 3,
        },
    },
});
```

## Checklist:


packages/json-schemas/schemas/call_data_schema.json
```diff
@@ -19,6 +19,24 @@
...
             "minimum": 0
+        },
+        "overrides": {
+            "type": "object",
+            "additionalProperties": {
+                "type": "object",
+                "properties": {
+                    "code": {
+                        "type": "string",
+                        "pattern": "^0x[0-9a-f]*$"
```
hysz, line 30: Not sure if we'd encounter this in practice but `[0-9a-fA-F]` 🤷

packages/web3-wrapper/src/web3_wrapper.ts
```diff
@@ -562,9 +562,10 @@ export class Web3Wrapper {
         }
         const marshalledDefaultBlock = marshaller.marshalBlockParam(defaultBlock);
         const callDataHex = marshaller.marshalCallData(callData);
+        const overrides = marshaller.marshalCallOverrides(callData.overrides || {});
         const rawCallResult = await this.sendRawPayloadAsync<string>({
             method: 'eth_call',
-            params: [callDataHex, marshalledDefaultBlock],
+            params: [callDataHex, marshalledDefaultBlock, overrides],
```
mzhu25, line 568: will this cause any problems for non-geth clients?

mzhu25: ok the CI failures tell me yes –– check if overrides is empty before adding it to params?

packages/json-schemas/schemas/call_data_schema.json
```diff
@@ -19,6 +19,24 @@
...
+                    },
+                    "nonce": {
+                        "oneOf": [{ "$ref": "/numberSchema" }, { "$ref": "/jsNumberSchema" }]
+                    },
+                    "value": {
```
dekz, line 35: is this `balance` and a `wholeNumberSchema` (wei)?

packages/web3-wrapper/src/web3_wrapper.ts
dorothy-zbornak: doh

packages/json-schemas/schemas/call_data_schema.json
dorothy-zbornak: d-doh

PR #2624, opened 2020-07-07
TransformerDeployer: Fix `kill()` interface mismatch.

dorothy-zbornak:
## Description

The current `TransformerDeployer.die()` is calling `Transformer.kill()`, which does not exist. We actually want to call `Transformer.kill(address payable ethRecipient)`. This PR fixes this.

We also now throw a revert error in `deploy()` if the `create()` opcode returns a null address, which indicates a constructor revert.

#### TODO before merging:
- [ ] Redeploy `TransformerDeployer`
- [ ] Have governor call `TransformERC20.setTransformerDeployer()` to new `TransformerDeployer` instance on all networks.
- [ ] Update `contract-addresses`.

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
+++ b/contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
@@ -24,7 +24,7 @@ import "@0x/contracts-utils/contracts/src/v06/AuthorizableV06.sol";
 
 /// @dev A contract with a `die()` function.
 interface IKillable {
-    function die() external;
+    function die(address payable ethRecipient) external;
 }
 
 /// @dev Deployer contract for ERC20 transformers.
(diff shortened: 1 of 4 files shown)


contracts/zero-ex/contracts/src/external/TransformerDeployer.sol
```diff
@@ -24,7 +24,7 @@ import "@0x/contracts-utils/contracts/src/v06/AuthorizableV06.sol";
 
 /// @dev A contract with a `die()` function.
 interface IKillable {
-    function die() external;
+    function die(address payable ethRecipient) external;
```
dorothy-zbornak, line 27: :innocent:

dorothy-zbornak:
Merged into #2657

PR #2626, opened 2020-07-08
ZeroEx: transformERC20 call data validation

dorothy-zbornak:
## Description

This adds signed calldata validation to the `TransformERC20` and `MetaTransactions` features.

0x-API should sign calldata like this:

```ts
export function signCallData(callData: string, privateKey: string): string {
    const {r, s, v} = ethjs.ecsign(
        ethjs.hashPersonalMessage(ethjs.toBuffer(hexUtils.hash(callData))),
        ethjs.toBuffer(privateKey),
    );
    const prefix = ethjs.keccak256('SignedCallDataSignature(bytes)');
    return hexUtils.concat(callData, prefix, v, r, s, SignatureType.EthSign);
}
```

The `TransformERC20` feature will only pass a non-null `callDataHash` parameter to transformers if the calldata has been signed by `getQuoteSigner()`.

Also, we can now override contract call `data` in `callAsync()`, `awaitTransactionSuccessAsync()`, and `estimateGasAsync()`.

### TODO
- [x] Tests
- [x] Update specs
- [x] Add `TransformERC20` dependency on `SignatureValidator` in checklist doc
- [x] Regenerate contract-wrappers

## Checklist:

- [x] Feature has updated version information.
- [x] implements IFeature interface.
- [x] Feature contracts are stateless (including inherited contracts).
- [x] onlySelf feature functions are prefixed with _.
- [x] Feature functions are added to full_migration_tests.
- [x] No delegatecalls from inside a feature. Call other features through the router.
- [x] No self-destruct in features.
- [x] No intentionally persistent (non-atomic) balances on the Exchange Proxy.
- [x] No allowances to be set on Exchange Proxy by users.
- [x] No direct access to another feature’s storage bucket without strong justification.
- [x] No direct access to AllowanceTarget (must go through TokenSpender).
- [x] No executing arbitrary calldata from the context of the Exchange Proxy.
- [x] Features use unique StorageIds. 
- [x] Document functions with execution contexts outside of the Exchange Proxy.
- [x] Document feature dependencies in checklist doc.
- [x] Document reentrant functions in checklist doc.
- [x] Document temporary balances.


contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
```diff
@@ -0,0 +1,67 @@
...
+library LibSignedCallData {
+    using LibBytesV06 for bytes;
+
+    // bytes4(keccak256('SignedCallDataSignature(bytes)'))
+    bytes4 constant private SIGNATURE_SELECTOR = 0xf86d1d92;
+
+    /// @dev Try to parse potentially signed calldata into its hash and signature
+    ///      components. Signed calldata has signature data appended to it.
+    /// @param callData the raw call data.
+    /// @return callDataHash The hash of the signed callData, or 0x0 if no signature
```
moodlezoup: we still return `keccak256(callData)` if no signature is present

contracts/zero-ex/contracts/src/features/SignatureValidator.sol
```diff
@@ -106,11 +106,18 @@ contract SignatureValidator is
         override
         returns (bool isValid)
     {
-        try this.validateHashSignature(hash, signer, signature) {
-            isValid = true;
-        } catch (bytes memory) {
-            isValid = false;
-        }
+        // HACK: `validateHashSignature()` is stateless so we can just perform
```
hysz, line 109: 👍

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
```diff
@@ -0,0 +1,67 @@
...
+        // ```
+        //   abi.encodePacked(
+        //     callData,
+        //     bytes4(keccak256('SignedCallDataSignature(bytes)')),
+        //     signature // 66 bytes
+        //   );
+        // ```
+
+        if (
+            // Signed callData has to be at least 70 bytes long.
```
hysz, line 58: I'm not sure whether optimized Solidity is always guaranteed to always execute `callData.length < 70` and only then `SIGNATURE_SELECTOR != callData.readBytes4(callData.length - 70)`. 🤔

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
```diff
@@ -0,0 +1,67 @@
...
+            callData.length < 70 ||
+            // The bytes4 at offset -70 should equal `SIGNATURE_SELECTOR`.
+            SIGNATURE_SELECTOR != callData.readBytes4(callData.length - 70)
```
hysz, line 61: A bit concerned about false positives arising from using the selector as a delimiter within arbitrary calldata. One way to mitigate this would be prepending the signature. Then we'd be able to guarantee no false positive, so long as we do not have a function with the same selector as a signature.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
```diff
@@ -242,55 +223,65 @@ contract TransformERC20 is
...
+                args.callDataSignature
             );
+            // Perform transformations.
+            for (uint256 i = 0; i < args.transformations.length; ++i) {
+                _executeTransformation(
+                    state.wallet,
+                    args.transformations[i],
+                    state.transformerDeployer,
+                    args.taker,
+                    // Transformers will receive a null calldata hash if
```
hysz, line 250: Curious about the benefit of passing `hash|null` opposed to a `bool`, how will the hash used by the transformer?

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
dorothy-zbornak: idk I think the odds of 4 specific bytes at a specific position in a typically large payload seems pretty low? Prepending seems like it would potentially be more of a headache for integrations and our own maintenance since the calldata will be unusable unless you are aware of the signing scheme (which we might change). We've also been using a similar scheme for metadata tagging thus far with no issues. :man_shrugging:

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
dorothy-zbornak: From the docs:

> The operators || and && apply the common short-circuiting rules. This means that in the expression f(x) || g(y), if f(x) evaluates to true, g(y) will not be evaluated even if it may have side-effects.

contracts/zero-ex/contracts/src/features/TransformERC20.sol
dorothy-zbornak: My reasoning is that it comes for free and is functionally the same right now, but conceivably it could be used to track per-quote state or help augment data pipeline events.

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
moodlezoup: yeah I agree that appending the signature is somewhat sketch, but afaict it _should_ be fine here, plus the stakes should be pretty low for the time being (worst case someone other than the intended taker fills an RFQT order?)

contracts/zero-ex/test/features/transform_erc20_test.ts
```diff
@@ -101,7 +112,29 @@ blockchainTests.resets('TransformERC20 feature', env => {
...
+
+        it('owner can set the quote signer with `setQuoteSigner()`', async () => {
+            const newSigner = randomAddress();
+            const receipt = await feature.setQuoteSigner(newSigner).awaitTransactionSuccessAsync({ from: owner });
+            verifyEventsFromLogs(receipt.logs, [{ quoteSigner: newSigner }], TransformERC20Events.QuoteSignerUpdated);
+            const actualSigner = await feature.getQuoteSigner().callAsync();
+            expect(actualSigner).to.eq(newSigner);
+        });
+
+        it('non-owner cannot set the transformer deployer with `setTransformerDeployer()`', async () => {
```
moodlezoup:
```suggestion
        it('non-owner cannot set the quote signer with `setQuoteSigner()`', async () => {
```

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
hysz: Awesome, thanks for looking up!

contracts/zero-ex/contracts/src/features/libs/LibSignedCallData.sol
hysz: I think for the current use case it's probably not a huge deal. Although I imagine this will be pointed out in the audits, so probably good to document the reasoning and worst case here in the code. I believe the worst case is just that we'll think the calldata is signed when it isn't, which would lead to a revert down the line.

dorothy-zbornak: In meta-transactions there's a possibility that we could unintentionally  truncate the calldata if the 4 byte selector is found at `-70` (but it is not actually signed calldata), which might alter the results of the `abi.decode()`, and thus the parameters passed to `_transformERC20()`. This could work out detrimentally for the taker. But I think it's really unlikely and in most cases, with conventional encoding algorithms, would just cause `abi.decode()` to revert.

dorothy-zbornak: Oh wait, nvm. We `abi.decode()` the raw calldata in MetaTransactions, so we should be good :zany_face:
The guy who wrote this is a genius.

PR #2627, opened 2020-07-08
asset-swapper: QuoteReport response

steveklebanoff:
## Description

Like https://github.com/0xProject/0x-monorepo/pull/2606 but with a refactor that involves **no passing around a QuoteReporter object 🎉   **

* We keep track of the orderHash => makerUri on the QuoteRequestor object itself.  this allows us to not pass around a QuoteReporter, and just pass everything in to generate the report in _generateOptimizedOrders
    * In order to support this, we generate a new `QuoteRequestor` on each swap quot erequest

* We return the QuoteReport from _generateOptimizedOrders and pass it up the call stack (wasn't as hairy as I thought it'd be)

## Checklist:

-   [ ] Make sure works with latest 0x API


packages/asset-swapper/src/types.ts
```diff
@@ -155,6 +156,7 @@ export interface SwapQuoteBase {
     bestCaseQuoteInfo: SwapQuoteInfo;
     worstCaseQuoteInfo: SwapQuoteInfo;
     sourceBreakdown: SwapQuoteOrdersBreakdown;
+    quoteReport?: QuoteReport;
```
steveklebanoff, line 159: Made this optional because I didn't know if/how we should have a `QuoteReport` return for the `getBatchMarket`... requests

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -19,6 +19,7 @@ import {
     SwapQuote,
     SwapQuoteRequestOpts,
     SwapQuoterOpts,
+    SwapQuoterRfqtOpts,
 } from './types';
 import { assert } from './utils/assert';
 import { calculateLiquidity } from './utils/calculate_liquidity';
```
steveklebanoff, line 26: All the changes to this class are just to support the generation of a new `QuoteRequestor` for each request, instead of using one `QuoteRequestor` for every request.   This is because we now store state on the `QuoteRequestor`

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -589,8 +588,8 @@ export class SwapQuoter {
 
         const calcOpts: CalculateSwapQuoteOpts = opts;
 
-        if (calcOpts.rfqt !== undefined && this._shouldEnableIndicativeRfqt(calcOpts.rfqt, marketOperation)) {
-            calcOpts.rfqt.quoteRequestor = this._quoteRequestor;
+        if (calcOpts.rfqt !== undefined) {
```
steveklebanoff, line 623: @feuGeneA I've changed this to _always_ pass through the `QuoteRequestor` (even if it is not an indicative quote.  Seems harmless from the code I've read through but I wanted to confirm w/ you

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -549,18 +539,27 @@ export class SwapQuoter {
...
+            this.expiryBufferMs,
+        );
+
         if (
             opts.rfqt &&
             opts.rfqt.intentOnFilling &&
             opts.rfqt.apiKey &&
-            this._rfqtTakerApiKeyWhitelist.includes(opts.rfqt.apiKey) &&
-            !(marketOperation === MarketOperation.Buy && this._rfqtSkipBuyRequests)
+            this._rfqtTakerApiKeyWhitelist().includes(opts.rfqt.apiKey) &&
```
dekz: `this._rfqtTakerApiKeyWhitelist().includes(opts.rfqt.apiKey)` -> `this._isApiKeyWhitelisted(ops.rfqt.apiKey)`

seems like the only usage of it.

packages/asset-swapper/src/types.ts
dekz: wouldn't there be one per swap quote? `getBatch` is just performing N many swap quotes at once.

packages/asset-swapper/src/types.ts
```diff
@@ -246,13 +256,7 @@ export interface SwapQuoterOpts extends OrderPrunerOpts {
...
     multiBridgeAddress?: string;
     ethGasStationUrl?: string;
-    rfqt?: {
-        takerApiKeyWhitelist: string[];
-        makerAssetOfferings: RfqtMakerAssetOfferings;
-        skipBuyRequests?: boolean;
-        warningLogger?: LogFunction;
-        infoLogger?: LogFunction;
-    };
+    rfqt?: SwapQuoterRfqtOpts;
```
dekz, line 264: 👌

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+            liquiditySource: ERC20BridgeSource.Native,
+            makerAmount: nativeOrder.makerAssetAmount,
+            takerAmount: nativeOrder.takerAssetAmount,
+            fillableTakerAmount: this._orderHashesToFillableAmounts[orderHash],
+            nativeOrder,
+            orderHash,
+        };
+
+        // if we find this is an rfqt order, label it as such and associate makerUri
+        const foundRfqtMakerUri = this._quoteRequestor && this._quoteRequestor.getMakerUriForOrderHash(orderHash);
```
dekz, line 144: ah I see, nice that we just keep it contained in there.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -394,6 +402,15 @@ export class MarketOperationUtils {
             multiBridgeAddress: opts.multiBridgeAddress,
             shouldBatchBridgeOrders: !!opts.shouldBatchBridgeOrders,
         });
+        const quoteReport = new QuoteReportGenerator(
```
dekz, line 465: nice

packages/asset-swapper/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "4.7.0",
```
steveklebanoff, line 3: ⚠️  this should probably be a major version bump since we change the return type of a public function

packages/asset-swapper/src/swap_quoter.ts
steveklebanoff: good call -- 1c84709db

packages/asset-swapper/src/types.ts
steveklebanoff: i didn't feel as though it was worth going into fns like `_calculateBatchBuySwapQuoteAsync` and returning `QuoteReport`s because I don't think there would actually be a use for it.  happy to revisit if you think it's important or could be useful though

packages/asset-swapper/src/swap_quoter.ts
feuGeneA: Yeah I guess I was just being defensive here when I wrote this originally.  Probably an over-optimization.

packages/asset-swapper/src/types.ts
```diff
@@ -230,6 +232,14 @@ export interface RfqtMakerAssetOfferings {
 
 export { LogFunction } from './utils/quote_requestor';
 
+export interface SwapQuoterRfqtOpts {
+    takerApiKeyWhitelist: string[];
+    makerAssetOfferings: RfqtMakerAssetOfferings;
+    skipBuyRequests?: boolean;
```
feuGeneA: Didn't Daniel recently remove this flag?  Or was that somewhere else?  If it's not too inconvenient, and if "skip buy requests" isn't a thing anymore, we should consider removing this.

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+export class QuoteReportGenerator {
+    private readonly _dexQuotes: DexSample[];
+    private readonly _nativeOrders: SignedOrder[];
+    private readonly _orderHashesToFillableAmounts: { [orderHash: string]: BigNumber };
+    private readonly _marketOperation: MarketOperation;
+    private readonly _collapsedFills: CollapsedFill[];
+    private readonly _quoteRequestor?: QuoteRequestor;
+
+    constructor(
+        marketOperation: MarketOperation,
```
feuGeneA, line 59: Nit: Consider using the constructor shorthand for member declaration and assignment.  (See [here](https://dev.to/satansdeer/typescript-constructor-shorthand-3ibd).)  It'd cut 11 lines here, and also IMHO it would improve readability.  When I'm reading code like this, and I know such a shorthand exists, and I don't see it used, it makes me think there's probably a reason that it wasn't used, and it makes me start watching out for what that reason might be.  If there is no such reason, then it seems like an unnecessary hurdle (albeit minor) to following along with the code.

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+
+        // convert order fillable amount array to easy to look up hash
+        if (orderFillableAmounts.length !== nativeOrders.length) {
+            // length mismatch, abort
+            this._orderHashesToFillableAmounts = {};
+            return;
+        }
+        const orderHashesToFillableAmounts: { [orderHash: string]: BigNumber } = {};
+        nativeOrders.forEach((nativeOrder, idx) => {
+            orderHashesToFillableAmounts[orderHashUtils.getOrderHash(nativeOrder)] = orderFillableAmounts[idx];
```
feuGeneA, line 80: Is this using a native JavaScript hash implementation, or is it using the `DevUtils` contract?  And if it's the latter, is it incurring an RPC call for for every hash calculation, or is it running the solidity in an in-memory EVM?

Not sure whether the answers to these questions would change the approach here or not, but it seems super important to be aware of them.

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+        });
+        this._orderHashesToFillableAmounts = orderHashesToFillableAmounts;
```
feuGeneA, line 82: For my own education, why did you have to declare a local var first and then assign that to the member?  Is it just so you could pre-initialize the value with `{}` without running into an error about the `readonly` nature of the var?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    constructor(
+        marketOperation: MarketOperation,
+        dexQuotes: DexSample[],
+        nativeOrders: SignedOrder[],
+        orderFillableAmounts: BigNumber[],
+        collapsedFills: CollapsedFill[],
+        quoteRequestor?: QuoteRequestor,
+    ) {
+        this._dexQuotes = dexQuotes;
```
feuGeneA, line 66: Have I mentioned how I loathe this name, `dexQuotes`?  0x is a DEX too, right!?

I know you're just using it here to correspond with the name already used outside of this class, and that's a good enough reason to leave this as is.  But please consider whether using a better name in here might improve reading comprehension.

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+        const orderHashesToFillableAmounts: { [orderHash: string]: BigNumber } = {};
+        nativeOrders.forEach((nativeOrder, idx) => {
+            orderHashesToFillableAmounts[orderHashUtils.getOrderHash(nativeOrder)] = orderFillableAmounts[idx];
+        });
+        this._orderHashesToFillableAmounts = orderHashesToFillableAmounts;
+    }
+
+    public generateReport(): QuoteReport {
+        const dexReportSourcesConsidered = this._dexQuotes.map(ds => this._dexSampleToReportSource(ds));
```
feuGeneA: Super nit: I'd prefer `dq` over `ds` here, to correspond with `_dexQuotes` just like the next line's `no` corresponds to `_nativeOrders`.  I got tripped up, saying "what does `ds` stand for?" ... "dex source?" ... "but the next line is a something-something source too, and it didn't use `-s`..."

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+export interface BridgeReportSource {
+    liquiditySource: Exclude<ERC20BridgeSource, ERC20BridgeSource.Native>;
+    makerAmount: BigNumber;
+    takerAmount: BigNumber;
+}
+
+interface NativeReportSourceBase {
+    liquiditySource: ERC20BridgeSource.Native;
+    makerAmount: BigNumber;
+    takerAmount: BigNumber;
```
feuGeneA, line 20: We don't need the `makerAmount` and `takerAmount` fields in this interface because they're included in the `nativeOrder` field ... right?  Or what am I missing?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -0,0 +1,150 @@
...
+        if (this._marketOperation === MarketOperation.Buy) {
+            return {
+                makerAmount: ds.input,
+                takerAmount: ds.output,
+                liquiditySource,
+            };
+        } else if (this._marketOperation === MarketOperation.Sell) {
+            return {
+                makerAmount: ds.output,
+                takerAmount: ds.input,
```
feuGeneA, line 123: Are you sure these Buy/Sell//maker/taker//input/output correspondences are correct?  I tried to double check it, but I couldn't make any sense of it in this context.  I would appreciate some feedback on how you came to this logic, even if just for my own education.

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: Good call-out!  Luckily, this is a native JS implementation:

https://github.com/0xProject/0x-monorepo/blob/development/packages/order-utils/src/order_hash_utils.ts#L8
https://github.com/0xProject/0x-monorepo/blob/development/packages/utils/src/hex_utils.ts#L93
https://github.com/0xProject/0x-monorepo/blob/development/packages/utils/src/sign_typed_data_utils.ts#L14

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: good callout -- it looks like just assigning directly compiles -- this just felt a little more declarative to me -- I don't feel super strongly tho

image

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: right on -- 876371359

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: you are correct -- makerAmount is the same as `makerAssetAmount` and takerAmount is the same as `takerAssetAmount` - but i think there is a big benefit to having shared consistent base attributes here (i.e. there always is a `makerAmount` and `takerAmount` regardless of the `liquiditySource`) for processing and comparing later in the data pipeline

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: I figured this out from the comment here: https://github.com/0xProject/0x-monorepo/blob/feature/new-order-reporter/packages/asset-swapper/src/utils/market_operation_utils/types.ts#L78:L78 and confirmed w/ @dekz

packages/asset-swapper/src/utils/quote_report_generator.ts
steveklebanoff: although the constructor shorthand does save lines of code, I've never seen it actually used in practice outside of some code you wrote recently.  I can't put my finger on it, but it does look a bit strange to me, and I haven't seen us adopt us a pattern to-date, so I wrote it using this assignment pattern.  I'm cool w/ converting if we want to establish this is a pattern though -- esp. if there is a corresponding lint rule

packages/asset-swapper/src/types.ts
steveklebanoff: I believe he removed it from being passed in from 0x-api.  I agree it should be removed, but I don't want to creep the scope of this task.  I added an Asana ticket

packages/asset-swapper/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "version": "5.0.0",
```
dekz: This isn't a breaking change is it?

steveklebanoff: This changes the return type of `getMarketSellOrdersAsync` and `getMarketBuyOrdersAsync ` which are public functions in `market_operation_utils/index.ts`

steveklebanoff:
Proof of this working on 0x-api: https://github.com/0xProject/0x-api/pull/289

PR #2635, opened 2020-07-15
fix: asset-swapper allow a empty overrides to signal no default

dekz:
## Description

Borked the merge, its really difficult to avoid the default merged in, i.e in Ganache. 

```
samplerOverrides = { overrides: {} }
```

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/swap_quoter.ts
+++ b/packages/asset-swapper/src/swap_quoter.ts
@@ -193,8 +193,7 @@ export class SwapQuoter {
                   [this._contractAddresses.erc20BridgeSampler]: { code: samplerBytecode },
               }
             : {};
-        const samplerOverrides = _.merge(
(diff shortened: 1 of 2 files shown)


packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -193,8 +193,7 @@ export class SwapQuoter {
                   [this._contractAddresses.erc20BridgeSampler]: { code: samplerBytecode },
               }
             : {};
-        const samplerOverrides = _.merge(
-            {},
+        const samplerOverrides = _.assign(
```
dorothy-zbornak, line 196: Don't we actually want `_.defaults()`?

Like
```ts
_.defaults(options.samplerOverrides, { block: ..., overrides: defaultOverrides })
```

PR #2640, opened 2020-07-21
asset-swapper: Quicker path-finding

dorothy-zbornak:
## Description

We can reduce the number of iterations we go through (`runLimit`) to find a decent path if we:
- Pre-sort all the individual (unmixed) paths by descending adjusted rate
- When mixing two paths, sort the fill nodes such that:
  - Fill nodes are always contiguous (form a valid subpath)
  - Subpaths (segments of the same source) are sorted by descending rate.    
- Compute the "rate" of a path based off `targetInput` (fill amount) instead of `min(targetInput, pathSize)`.
  - This favors paths that cover more of the total fill (not sure about this).
-  Increase the `runLimit` decay rate to `0.5` (prev `0.8`). So we halve the number of iterations for each subsequent call to `mixPaths()`.
- Start the `bestPath` in `mixPaths()` off with the left path (the previous mixed path), instead of starting with none.

This makes it more likely that we walk the viable paths first, so we can cut `runLimit` down to something like `2**8` (previously `2**13`).

This results in a much faster (2-3x) response time with small change in quotes:

ab-response-time

ab-realized

*Notes*: 
- WETH/DAI/USDC sims buy AND sell
- orange is this version w/ `runLimit = 2**8`
- blue is original version w/ `runLimit = 2**13`

We can probably tweak `runLimit` to improve results. This should buy us a little breathing room until we can entirely replace the algorithm with something better.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/fills.ts
@@ -1,4 +1,4 @@
-import { BigNumber } from '@0x/utils';
+import { BigNumber, hexUtils } from '@0x/utils';
 
 import { MarketOperation, SignedOrderWithFillableAmounts } from '../../types';
 import { fillableAmountsUtils } from '../../utils/fillable_amounts_utils';
@@ -56,6 +56,7 @@ function nativeOrdersToPath(
     ethToOutputRate: BigNumber,
     fees: FeeSchedule,
 ): Fill[] {
+    const sourcePathId = hexUtils.random();
     // Create a single path from all orders.
     let path: Fill[] = [];
     for (const order of orders) {
@@ -84,6 +85,7 @@ function nativeOrdersToPath(
             continue;
         }
         path.push({
+            sourcePathId,
             input: clippedInput,
             output: clippedOutput,
             rate,
@@ -114,6 +116,7 @@ function dexQuotesToPaths(
 ): Fill[][] {
     const paths: Fill[][] = [];
     for (let quote of dexQuotes) {
+        const sourcePathId = hexUtils.random();
         const path: Fill[] = [];
         // Drop any non-zero entries. This can occur if the any fills on Kyber were UniswapReserves
         // We need not worry about Kyber fills going to UniswapReserve as the input amount
@@ -136,6 +139,7 @@ function dexQuotesToPaths(
             const adjustedRate = side === MarketOperation.Sell ? adjustedOutput.div(input) : input.div(adjustedOutput);
 
             path.push({
+                sourcePathId,
                 input,
                 output,
                 rate,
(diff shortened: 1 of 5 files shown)


packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -43,9 +44,10 @@ function mixPaths(
     targetInput: BigNumber,
     maxSteps: number,
 ): Fill[] {
-    let bestPath: Fill[] = [];
-    let bestPathInput = ZERO_AMOUNT;
-    let bestPathRate = ZERO_AMOUNT;
+    const _maxSteps = Math.max(maxSteps, 16);
```
dorothy-zbornak, line 45: Also cap the minimum so everyone gets *some* action.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -285,11 +285,11 @@ export function getFallbackSourcePaths(optimalPath: Fill[], allPaths: Fill[][]):
...
 
 export function getPathAdjustedRate(side: MarketOperation, path: Fill[], targetInput: BigNumber): BigNumber {
-    const [input, output] = getPathAdjustedSize(path, targetInput);
-    if (input.eq(0) || output.eq(0)) {
+    const [, output] = getPathAdjustedSize(path, targetInput);
+    if (output.eq(0)) {
         return ZERO_AMOUNT;
     }
-    return side === MarketOperation.Sell ? output.div(input) : input.div(output);
+    return side === MarketOperation.Sell ? output.div(targetInput) : targetInput.div(output);
```
moodlezoup, line 273: hm yeah idk about this –– have you AB tested this change alone?

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -109,6 +109,10 @@ export enum FillFlags {
  * Represents a node on a fill path.
  */
 export interface Fill<TFillData extends FillData = FillData> {
+    // Unique ID of the original source path this fill belongs to.
+    // This is generated when the path is generated and is useful to distinguish
+    // paths that have the same `source` IDs but are distinct (e.g., Curves).
+    sourcePathId: string;
```
dorothy-zbornak, line 126: New field to tell consolidated sources apart.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -112,6 +114,7 @@ function dexQuotesToPaths(
     ethToOutputRate: BigNumber,
     fees: FeeSchedule,
 ): Fill[][] {
+    const sourcePathId = hexUtils.random();
```
moodlezoup: this should be inside the for loop I think

moodlezoup (commented):
overall approach look good to me

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
dorothy-zbornak: ~I want it to be unique to the whole path. So we can tell which fill belongs to which curve.~

dorothy-zbornak: nvm ur rite

dorothy-zbornak: yay that fixed my sims

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -57,29 +59,27 @@ function mixPaths(
...
+                if (!isValidPath(nextPath, true)) {
                     continue;
                 }
                 // Remove this fill from the next list of candidate fills.
                 const nextAllFills = allFills.slice();
                 nextAllFills.splice(i, 1);
                 // Recurse.
                 _walk(
-                    childPath,
+                    [...path, fill],
```
dorothy-zbornak:
```suggestion
                    nextPath,
```

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
dorothy-zbornak: `7001`: Using `targetInput` as `input` (this PR)
`7002`: Using `input` as `input` (original recipe)

ab-realized-rates

moodlezoup: 🤷  whatever wins I guess

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -745,9 +739,9 @@ describe('MarketOperationUtils tests', () => {
                 expect(improvedOrders).to.be.length(3);
                 const orderFillSources = improvedOrders.map(o => o.fills.map(f => f.source));
                 expect(orderFillSources).to.deep.eq([
-                    [ERC20BridgeSource.Uniswap],
+                    [ERC20BridgeSource.Uniswap, ERC20BridgeSource.Curve],
                     [ERC20BridgeSource.Native],
-                    [ERC20BridgeSource.Eth2Dai, ERC20BridgeSource.Curve],
+                    [ERC20BridgeSource.Eth2Dai],
```
moodlezoup, line 818: wait what's going on here?

dorothy-zbornak: Yeah it's a little funky that Curve comes right after Uniswap now, but both ways are technically valid so I'm like w/e :shrug:

moodlezoup: sus but ok

moodlezoup (approved):

packages/asset-swapper/test/market_operation_utils_test.ts
dorothy-zbornak: welcome to asset-swapper lol

PR #2641, opened 2020-07-21
feat: asset-swapper market depth

dekz:
## Description

Adds `getBidAskLiquidityForMakerTakerAssetPairAsync ` which returns the raw liquidity from asset-swapper 

* `getMarket{Buy|Sell}LiquidityAsync` does all the liquidity finding things and returns it pretty raw as a MarketSideLiquidity interface
* Attached PLP address to the PLP fillData so we don't have to pass it all the way down as a parameter 
* Attached MultiBridge address to the MultiBridge fillData so we don't have to pass it all the way down as a parameter 

## Checklist:


dorothy-zbornak (commented):
makes sense

PR #2647, opened 2020-07-28
Feature/bunny hop

moodlezoup:
## Checklist:


packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -585,6 +623,146 @@ export const samplerOperations = {
...
+                                intermediateToken,
+                                takerToken,
+                                takerFillAmounts,
+                            );
+                        case ERC20BridgeSource.Balancer:
+                            return this.balancerPoolsCache
+                                .getCachedPoolAddressesForPair(takerToken, makerToken)
+                                .map(poolAddress =>
+                                    this.getBalancerSellQuotes(poolAddress, makerToken, takerToken, takerFillAmounts),
+                                );
```
moodlezoup: note that this will _only_ be used for two-hop quotes.
Recall that `ERC20BridgeSource.Balancer` gets excluded from the sources passed into `getSellQuotes` and instead we use the BalancerPoolCache/GraphQL query to compute Balancer single-hop quotes. 
For two-hop, the balancer pool addresses need to be known _before_ the sampler RPC call. I want to avoid performing the GraphQL query + RPC call synchronously, so instead we just pull whatever pool addresses are in the cache and pass those to the sampler.

moodlezoup: We could also adopt this approach of using BalancerPoolsCache/GraphQL to fetch pool addresses and BalancerSampler to fetch balances/compute quotes for single-hop quotes, but then we would also potentially need to make a GraphQL query + RPC call synchronously (though we'd dial the cache expiry way up, so the GraphQL queries would be less frequent)

contracts/erc20-bridge-sampler/contracts/src/BalancerSampler.sol
```diff
@@ -0,0 +1,143 @@
...
+
+*/
+
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "./interfaces/IBalancer.sol";
+
+
+contract BalancerSampler {
```
dorothy-zbornak: :raised_hands:

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -109,18 +109,46 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
...
-                buyToken,
-                side: isBuyQuote(quote) ? FillQuoteTransformerSide.Buy : FillQuoteTransformerSide.Sell,
-                fillAmount: isBuyQuote(quote) ? quote.makerAssetFillAmount : quote.takerAssetFillAmount,
-                maxOrderFillAmounts: [],
-                orders: quote.orders,
-                signatures: quote.orders.map(o => o.signature),
-            }),
-        });
+        if (quote.isTwoHop) {
+            const [firstHopOrder, secondHopOrder] = quote.orders;
```
dorothy-zbornak, line 112: So only all in two hop for now?

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -109,18 +109,46 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
...
+            transforms.push({
+                deploymentNonce: this.transformerNonces.fillQuoteTransformer,
+                data: encodeFillQuoteTransformerData({
+                    sellToken,
+                    buyToken: getTokenFromAssetData(firstHopOrder.makerAssetData),
+                    side: FillQuoteTransformerSide.Sell,
```
dorothy-zbornak, line 118: Only sells for now?

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -109,18 +109,46 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
...
+                    fillAmount: firstHopOrder.takerAssetAmount,
+                    maxOrderFillAmounts: [],
+                    orders: [firstHopOrder],
+                    signatures: [firstHopOrder.signature],
+                }),
+            });
+            transforms.push({
```
dorothy-zbornak, line 125: Never even thought of stacking this transformer. 😎

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -334,10 +334,8 @@ export class QuoteRequestor {
                             switch (quoteType) {
                                 case 'firm':
                                     return 'quote';
-                                    break;
```
dorothy-zbornak, line 358: :see_no_evil:

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -420,7 +425,7 @@ export class SwapQuoter {
         const takerAssetData = assetDataUtils.encodeERC20AssetData(takerTokenAddress);
         let [sellOrders, buyOrders] =
             options.excludedSources && options.excludedSources.includes(ERC20BridgeSource.Native)
-                ? Promise.resolve([[], []])
```
dorothy-zbornak: :see_no_evil:

contracts/erc20-bridge-sampler/contracts/src/TwoHopSampler.sol
```diff
@@ -0,0 +1,124 @@
...
+        uint256 intermediateAssetAmount = 0;
+        for (uint256 i = 0; i != firstHopCalls.length; ++i) {
+            firstHopCalls[i].writeUint256(firstHopCalls[i].length - 32, sellAmount);
+            (bool didSucceed, bytes memory returnData) = address(this).staticcall(firstHopCalls[i]);
+            if (didSucceed) {
+                uint256 amount = returnData.readUint256(64);
+                if (amount > intermediateAssetAmount) {
+                    intermediateAssetAmount = amount;
+                    firstHop.sourceIndex = i;
+                    firstHop.returnData = returnData;
```
dorothy-zbornak: Ah OK, so it's all in one source both hops, with no respecting gas/fees?

packages/asset-swapper/src/types.ts
```diff
@@ -205,9 +211,14 @@ export interface SwapQuoteInfo {
 /**
  * percentage breakdown of each liquidity source used in quote
  */
-export interface SwapQuoteOrdersBreakdown {
-    [source: string]: BigNumber;
-}
+export type SwapQuoteOrdersBreakdown = Partial<
+    { [key in Exclude<ERC20BridgeSource, typeof ERC20BridgeSource.MultiHop>]: BigNumber } & {
```
dorothy-zbornak, line 215: :crystal_ball:

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,337 +1,530 @@
...
-                    .getOrderFillableTakerAssetAmounts(orders, orders.map(o => o.signature), devUtilsAddress)
-                    .getABIEncodedTransactionData();
-            },
-            handleCallResultsAsync: async (contract, callResults) => {
-                return contract.getABIDecodedReturnData<BigNumber[]>('getOrderFillableTakerAssetAmounts', callResults);
-            },
-        };
-    },
-    getOrderFillableMakerAmounts(orders: SignedOrder[], devUtilsAddress: string): BatchedOperation<BigNumber[]> {
+export class SamplerOperations {
```
dorothy-zbornak, line 37: H O O R A Y

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,337 +1,530 @@
...
+                return result;
             },
         };
-    },
-    getUniswapV2SellQuotes(
+    }
+
+    constructor(
+        protected readonly _samplerContract: ERC20BridgeSamplerContract,
+        public readonly balancerPoolsCache: BalancerPoolsCache = new BalancerPoolsCache(),
```
dorothy-zbornak, line 52: Feel like there's a ton of other things we can pass in and store here.

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -261,17 +266,39 @@ export interface GetMarketOrdersOpts {
...
+    implements SourceQuoteOperation<TFillData> {
+    constructor(
+        private readonly _samplerContract: ERC20BridgeSamplerContract,
+        public readonly source: ERC20BridgeSource,
+        private readonly _samplerFunction: (...params: TParams) => ContractFunctionObj<TReturn>,
+        private readonly _params: any[],
+        public fillData: TFillData = {} as TFillData, // tslint:disable-line:no-object-literal-type-assertion
+        public handleCallResultsAsync: (callResults: string) => Promise<BigNumber[]> = async (callResults: string) => {
+            return _samplerContract.getABIDecodedReturnData<BigNumber[]>(_samplerFunction.name, callResults);
+        },
```
dorothy-zbornak: This many args should really go in an object.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -585,6 +623,146 @@ export const samplerOperations = {
...
+        sources: ERC20BridgeSource[],
+        makerToken: string,
+        takerToken: string,
+        makerFillAmounts: BigNumber[],
+        wethAddress: string,
+        liquidityProviderRegistryAddress?: string,
+    ): SourceQuoteOperation[] {
+        return _.flatten(
+            sources.map(
+                (source): SourceQuoteOperation | SourceQuoteOperation[] => {
```
dorothy-zbornak, line 813: Do any of these actually return a singular `SourceQuoteOperation`?

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { FeeSchedule, SourceInfo, TokenAdjacencyGraph } from './types';
+
+/**
+ * Given a token pair, returns the intermediate tokens to consider for two-hop routes.
+ */
+export function getIntermediateTokens(
+    makerToken: string,
+    takerToken: string,
+    tokenAdjacencyGraph: TokenAdjacencyGraph,
```
dorothy-zbornak, line 14: What will this table look like? There are no X/ETH pairs?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -499,12 +503,42 @@ export class MarketOperationUtils {
...
+                side,
+                _.flatten(dexQuotes),
+                twoHopQuotes,
+                nativeOrders,
+                orderFillableAmounts,
+                bestTwoHopQuote!,
+                opts.quoteRequestor,
+            );
+            return { optimizedOrders: twoHopOrders, quoteReport: twoHopQuoteReport, isTwoHop: true };
+        }
```
dorothy-zbornak: Maybe best to put this in a different function since it feels detour-y.

dorothy-zbornak (commented):
Yeah, I think this approach is fine as an MVP, contingent on the sims looking good.

On another note, I feel like AS is at its breaking point with the amount of twisty edge casing and union types going on but idk how to keep building on it in the short term without making it worse. After this, I'm thinking we should do a feature freeze until a refactor happens.

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
moodlezoup: yep

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
moodlezoup: no, but we have to implement market buys as two FQT sells –– we don't know how much intermediate token we'll need, so we sell the full taker token amount, and after that we might as well sell the full intermediate token amount

contracts/erc20-bridge-sampler/contracts/src/TwoHopSampler.sol
moodlezoup: yup

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
moodlezoup: Oasis, Uniswap, Kyber, LP, MB all return a single `SourceQuoteOperation`

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
moodlezoup: yeah, no X/ETH pairs because we pretty much always want to try ETH as the intermediate token. This is what I'm going with for now: 
```
const TOKEN_ADJACENCY_GRAPH = {
    // renBTC: wBTC
    '0xeb4c2781e4eba804ce9a9803c67d0893436bb27d': ['0x2260fac5e5542a773aa44fbcfedf7c193bc2c599'],
    // TUSD: USDC, DAI
    '0x0000000000085d4780B73119b644AE5ecd22b376': ['0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', '0x6b175474e89094c44da98b954eedeac495271d0f'],
    // USDT: USDC, DAI
    '0xdac17f958d2ee523a2206206994597c13d831ec7': ['0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', '0x6b175474e89094c44da98b954eedeac495271d0f'],
    // COMP: USDC
    '0xc00e94cb662c3520282e6f5717214004a7f26888': ['0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'],
    // LEND: USDC
    '0x80fB784B7eD66730e8b1DBd9820aFD29931aab03': ['0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'],
    // SNX: USDC
    '0xc011a73ee8576fb46f5e1c5751ca3b9fe0af2a6f': ['0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'],
};
```

moodlezoup:
> Yeah, I think this approach is fine as an MVP, contingent on the sims looking good.
> 
> On another note, I feel like AS is at its breaking point with the amount of twisty edge casing and union types going on but idk how to keep building on it in the short term without making it worse. After this, I'm thinking we should do a feature freeze until a refactor happens.

agreed. we should definitely refactor once we freeze v0

contracts/test-utils/src/test_with_reference.ts
```diff
@@ -144,7 +144,7 @@ function _getParameterNames(func: (...args: any[]) => any): string[] {
         .replace(/[/][/].*$/gm, '') // strip single-line comments
         .replace(/\s+/g, '') // strip white space
         .replace(/[/][*][^/*]*[*][/]/g, '') // strip multi-line comments
-        .split('){', 1)[0]
+        .split(/\){|\)=>/, 1)[0]
```
moodlezoup, line 147: es6 things I guess

packages/asset-swapper/contracts/src/BalancerSampler.sol
```diff
@@ -0,0 +1,143 @@
...
+pragma solidity ^0.5.9;
+pragma experimental ABIEncoderV2;
+
+import "./interfaces/IBalancer.sol";
+
+
+contract BalancerSampler {
+
+    /// @dev Base gas limit for Balancer calls.
+    uint256 constant private BALANCER_CALL_GAS = 300e3; // 600k
```
dekz:
```suggestion
    uint256 constant private BALANCER_CALL_GAS = 300e3; // 300k
```

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -42,10 +46,52 @@ export class BalancerPoolsCache {
...
+        }
+        const minTimestamp = Date.now() - cacheExpiryMs;
+        if (value === undefined || value.timestamp < minTimestamp) {
+            return undefined;
+        } else {
+            return value.pools.map(pool => pool.id);
+        }
+    }
+
+    public howToSampleBalancer(
```
moodlezoup, line 67: we will sometimes sample Balancer off-chain, sometimes on-chain. We need to make a GraphQL query once in a while so that the cached pool addresses are kept reasonably fresh, and if we're going to be making the query anyways we might as well compute the quotes off-chain (as we're doing now). Then as long as the cached pool addresses are reasonably fresh, we go with the on-chain sampling approach. 
For the purpose of two-hop, we only use the on-chain sampling approach in order to avoid making two queries (GraphQL + eth RPC) in sequence. As a result it takes a few queries for the Balancer cache to "warm up" and be usable for two-hop

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -109,18 +107,46 @@ export class ExchangeProxySwapQuoteConsumer implements SwapQuoteConsumerBase {
...
-                signatures: quote.orders.map(o => o.signature),
-            }),
-        });
+        if (quote.isTwoHop) {
+            const [firstHopOrder, secondHopOrder] = quote.orders;
+            transforms.push({
+                deploymentNonce: this.transformerNonces.fillQuoteTransformer,
+                data: encodeFillQuoteTransformerData({
+                    sellToken,
+                    buyToken: getTokenFromAssetData(firstHopOrder.makerAssetData),
```
dekz: Are all of the following equivalent? If so might be worth just assigning it to `intermediateToken` and using it in the 3 places.
```typescript
getTokenFromAssetData(firstHopOrder.makerAssetData) 
=== getTokenFromAssetData(secondHopOrder.takerAssetData)
=== getTokenFromAssetData(quote.orders[0].makerAssetData)
```

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -42,10 +46,52 @@ export class BalancerPoolsCache {
         return Promise.race([this._getPoolsForPairAsync(takerToken, makerToken), timeout]);
     }
 
-    protected async _getPoolsForPairAsync(takerToken: string, makerToken: string): Promise<BalancerPool[]> {
+    public getCachedPoolAddressesForPair(
+        takerToken: string,
+        makerToken: string,
+        cacheExpiryMs?: number,
+    ): string[] | undefined {
+        const key = JSON.stringify([takerToken, makerToken]);
```
dekz, line 54: Would we have two caches for `[DAI,USDC]` and `[USDC,DAI]`? Would sorting the keys help a cache hit?

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -155,6 +155,22 @@ function dexQuotesToPaths(
...
+    twoHopQuote: DexSample<MultiHopFillData>,
+    targetInput: BigNumber,
+    ethToOutputRate: BigNumber,
+    fees: FeeSchedule = {},
+): BigNumber {
+    const { output, input, fillData } = twoHopQuote;
+    if (input.isLessThan(targetInput) || output.isZero()) {
+        return ZERO_AMOUNT;
+    }
+    const penalty = ethToOutputRate.times(fees[ERC20BridgeSource.MultiHop]!(fillData));
```
dekz, line 169: Why isn't the penalty just the sources in the hop combined?

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
```diff
@@ -42,10 +46,52 @@ export class BalancerPoolsCache {
...
+        // Sample Balancer off-chain (i.e. via GraphQL query + `computeBalancerBuy/SellQuote`)
+        // if cached values are stale
+        const offChain = cachedBalancerPools === undefined;
+        return { onChain, offChain };
+    }
+
+    protected async _getPoolsForPairAsync(
+        takerToken: string,
+        makerToken: string,
+        cacheExpiryMs: number = FIVE_SECONDS_MS,
```
dekz, line 90: Is this still intentionally low?

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
```diff
@@ -0,0 +1,45 @@
...
+    return intermediateTokens;
+}
+
+/**
+ * Returns the best two-hop quote and the fee-adjusted rate of that quote.
+ */
+export function getBestTwoHopQuote(
+    marketSideLiquidity: MarketSideLiquidity,
+    feeSchedule?: FeeSchedule,
+): { quote: DexSample<MultiHopFillData> | undefined; rate: BigNumber } {
```
dekz:
```suggestion
): { quote: DexSample<MultiHopFillData> | undefined; adjustedRate: BigNumber } {
```
just to avoid any potential bugs in assuming it is the rate rate.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
+        protected readonly _samplerContract: ERC20BridgeSamplerContract,
+        public readonly bancorService?: BancorService,
+        public readonly balancerPoolsCache: BalancerPoolsCache = new BalancerPoolsCache(),
+    ) {}
+
+    public getOrderFillableTakerAmounts(orders: SignedOrder[], exchangeAddress: string): BatchedOperation<BigNumber[]> {
+        return new SamplerContractOperation({
+            source: ERC20BridgeSource.Native,
+            contract: this._samplerContract,
+            function: this._samplerContract.getOrderFillableTakerAssetAmounts,
```
dekz, line 59: Isn't this annoying since you lose type information?

dekz (commented):
Going to review this in 2 passes

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
moodlezoup: o that's what I had it as initially, but I moved it over to the fee schedule for consistency with the other sources –– I also figure there's a chance we mess with it, e.g. apply the gas limit buffer multiplier to the "non-deterministic" hops

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
moodlezoup: fixed

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
moodlezoup: the two aren't exactly interchangeable because we sort the pools by maker token balance. I'm not too concerned about cache misses, bc at least anecdotally (running simbot) the cache warms up pretty quick

packages/asset-swapper/src/utils/market_operation_utils/balancer_utils.ts
moodlezoup: yeah, in theory we should almost never be getting a cache hit for this function because we will only be calling it if `howToSampleBalancer` decides that the cache is too stale to use on-chain sampling. Leaving it in just in case something goes awry

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -152,45 +172,33 @@ export class MarketOperationUtils {
...
-                this._sampler.bancorService,
-            )
-            .then(async r => this._sampler.executeAsync(r));
+        const offChainBalancerPromise = sampleBalancerOffChain
+            ? this._sampler.getBalancerSellQuotesAsync(makerToken, takerToken, sampleAmounts)
+            : Promise.resolve([]);
+
+        const offChainBancorPromise = _opts.excludedSources.includes(ERC20BridgeSource.Bancor)
+            ? Promise.resolve([])
+            : this._sampler.getBancorSellQuotesAsync(makerToken, takerToken, sampleAmounts);
```
dekz: To be safe (I forgot to exclude this). If it's not excluded, we should just perform one sample (2 requests). Until we can sample efficiently.
```suggestion
            : this._sampler.getBancorSellQuotesAsync(makerToken, takerToken, [takerAmount]);
```

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
```diff
@@ -0,0 +1,49 @@
...
+        intermediateTokens = _.get(tokenAdjacencyGraph, takerToken, [] as string[]);
+    } else if (takerToken === wethAddress) {
+        intermediateTokens = _.get(tokenAdjacencyGraph, makerToken, [] as string[]);
+    } else {
+        intermediateTokens = _.union(
+            _.intersection(_.get(tokenAdjacencyGraph, takerToken, []), _.get(tokenAdjacencyGraph, makerToken, [])),
+            [wethAddress],
+        );
+    }
+    return intermediateTokens;
```
dekz: I imagine we'll probably generate this `TokenAdjacencyGraph` programatically. Might be best to filter out bad combos (filter `makerToken` and `takerToken` out of `intermediateTokens`)

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
-    getMStableSellQuotes(makerToken: string, takerToken: string, takerFillAmounts: BigNumber[]): SourceQuoteOperation {
-        return {
+            fillData: { poolAddress },
+            contract: this._samplerContract,
+            function: this._samplerContract.sampleBuysFromBalancer,
+            params: [poolAddress, takerToken, makerToken, makerFillAmounts],
+        });
+    }
+
+    public async getBalancerSellQuotesAsync(
```
dekz: suggestion: make this quite clear that it is an off-chain lookup, perhaps by function name.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
-        };
-    },
-    getMooniswapSellQuotes(
+            contract: this._samplerContract,
+            function: this._samplerContract.sampleBuysFromMStable,
+            params: [takerToken, makerToken, makerFillAmounts],
+        });
+    }
+
+    public async getBancorSellQuotesAsync(
```
dekz: suggestion: rename/comment this similar to balancer to indicate it's pretty dangerous to call often.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
+                takerToken,
+                [new BigNumber(0)],
+                wethAddress,
+                liquidityProviderRegistryAddress,
+            );
+            const secondHopOps = this._getSellQuoteOperations(
+                sources,
+                makerToken,
+                intermediateToken,
+                [new BigNumber(0)],
```
dekz:
```suggestion
                [ZERO_AMOUNT],
```

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
+        wethAddress: string,
+        liquidityProviderRegistryAddress?: string,
+    ): BatchedOperation<Array<DexSample<MultiHopFillData>>> {
+        const intermediateTokens = getIntermediateTokens(makerToken, takerToken, tokenAdjacencyGraph, wethAddress);
+        const subOps = intermediateTokens.map(intermediateToken => {
+            const firstHopOps = this._getSellQuoteOperations(
+                sources,
+                intermediateToken,
+                takerToken,
+                [new BigNumber(0)],
```
dekz:
```suggestion
                [ZERO_AMOUNT],
```

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -1,470 +1,630 @@
...
+                source: ERC20BridgeSource.MultiHop,
+                function: this._samplerContract.sampleTwoHopSell,
+                params: [firstHopOps.map(op => op.encodeCall()), secondHopOps.map(op => op.encodeCall()), sellAmount],
+                fillData: { intermediateToken } as MultiHopFillData, // tslint:disable-line:no-object-literal-type-assertion
+                callback: async (callResults: string, fillData: MultiHopFillData): Promise<BigNumber[]> => {
+                    const [firstHop, secondHop, buyAmount] = this._samplerContract.getABIDecodedReturnData<
+                        [HopInfo, HopInfo, BigNumber]
+                    >('sampleTwoHopSell', callResults);
+                    if (buyAmount.isZero()) {
+                        return Promise.resolve([buyAmount]);
```
dekz: personal nit
```suggestion
                        return Promise.resolve([ZERO_AMOUNT]);
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -273,17 +278,65 @@ export interface GetMarketOrdersOpts {
...
+    TFillData extends FillData = FillData
+> {
+    contract: ERC20BridgeSamplerContract;
+    function: TFunc;
+    params: Parameters<TFunc>;
+    callback?: (callResults: string, fillData: TFillData) => Promise<BigNumber[]>;
 }
 
-export interface OptimizedOrdersAndQuoteReport {
+export class SamplerContractOperation<
```
dekz: nit: move this to its own file

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -273,17 +278,65 @@ export interface GetMarketOrdersOpts {
...
+        this._params = opts.params;
+        this._callback = opts.callback;
+    }
+
+    public encodeCall(): string {
+        return this._samplerFunction
+            .bind(this._samplerContract)(...this._params)
+            .getABIEncodedTransactionData();
+    }
+    public async handleCallResultsAsync(callResults: string): Promise<BigNumber[]> {
```
dekz: do we have anything that is async anymore? Offchain Bancor/Balancer are out of the usual sampler operations?

It does scare me if this async fn can execute for a long time :P

dekz (commented):
A few nits but looking really good to me!

I'll get this on staging and do a proper simbot run tomorrow

packages/asset-swapper/src/utils/market_operation_utils/types.ts
moodlezoup: yep, you're right

dekz (approved):

PR #2650, opened 2020-07-30
Feat/bancor v2

xianny:
## Description

Add integration for Bancor V2.

**contracts/asset-proxy**
* BancorBridge contract and tests

**asset-swapper**
* BancorService (wrapper for the Bancor SDK)
* tests for BancorService against mainnet and real Bancor network
* refactor of SourceQuoteOperation to return one fillData per sample instead of one fillData per source

**contract-addresses**
* added bancorBridge field

We're waiting on https://github.com/bancorprotocol/sdk/pull/20 to merge and publish before we enable Bancor as a source. After the SDK is updated, we need to do the following: 

- [ ] Deploy BancorBridge contract and then update @0x/contract-addresses
- [ ] Add Bancor to list of valid Sell sources in `.../market_operation_utils/constants.ts`
- [ ] Privately publish and AB test on staging/prod

## Testing instructions

I tested with simbot using https://github.com/dorothy-zbornak/0x-swap-server. Charts are in a comment below. 

Unit tests are updated.

## Types of changes

## Checklist:


contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+import "../interfaces/IBancorNetwork.sol";
+
+
+contract BancorBridge is
+    IERC20Bridge,
+    IWallet,
+    DeploymentConstants
+{
+
+    IContractRegistry contractRegistry = IContractRegistry(0x52Ae12ABe5D8BD778BD5397F99cA900624CfADD4);
```
dekz: we should have these addresses in DeploymentConstants so they can be faked in tests (with `address(this)`)

contracts/asset-proxy/contracts/src/interfaces/IBancorNetwork.sol
```diff
@@ -0,0 +1,50 @@
...
+contract IContractRegistry {
+    function addressOf(
+        bytes32 contractName
+    ) external returns(address);
+}
+
+
+contract IBancorNetwork {
+    function convertByPath(
+        IERC20Token[] calldata _path,
```
dekz: replacing these with `address[]` then we can avoid `state.tokenPath[i] = IERC20Token(state.addressPath[i]);`

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+
+        // Just transfer the tokens if they're the same.
+        if (state.addressPath[0] == toTokenAddress) {
+            LibERC20Token.transfer(state.addressPath[0], to, amount);
+            return BRIDGE_SUCCESS;
+        }
+
+        // Otherwise use Bancor to convert
+        // Convert the addresses in the path to IERC20Tokens
+        for (uint i=0; i < state.addressPath.length; i++) {
```
dekz: we should be able to avoid this by  changing the interface of Bancor Contracts to be `address` or `address[]` instead of `IERC20Token`

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+        LibERC20Token.approveIfBelow(state.addressPath[0], contractRegistry.addressOf(bancorNetworkName), fromTokenBalance);
+
+        // Convert the tokens
+        IBancorNetwork bancorNetwork = getBancorNetworkContract();
+        uint256 boughtAmount = bancorNetwork.convertByPath(
+            state.tokenPath, // path originating with source token and terminating in destination token
+            fromTokenBalance, // amount of source token to trade
+            amount, // minimum amount of destination token expected to receive
+            to, // beneficiary
+            address(0x0), // affiliateAccount; no fee paid
```
dekz:
```suggestion
            address(0), // affiliateAccount; no fee paid
```

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+        for (uint i=0; i < state.addressPath.length; i++) {
+            state.tokenPath[i] = IERC20Token(state.addressPath[i]);
+        }
+
+        // // Grant an allowance to the Bancor Network to spend `fromTokenAddress` token.
+        uint256 fromTokenBalance = state.tokenPath[0].balanceOf(address(this));
+        LibERC20Token.approveIfBelow(state.addressPath[0], contractRegistry.addressOf(bancorNetworkName), fromTokenBalance);
+
+        // Convert the tokens
+        IBancorNetwork bancorNetwork = getBancorNetworkContract();
```
dekz: Worth putting this in the `bridgeData` to save the lookup, especially if it is essentially constant (e.g doesn't change depending on the tokens). Would save a chunk of gas.

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+
+        // Otherwise use Bancor to convert
+        // Convert the addresses in the path to IERC20Tokens
+        for (uint i=0; i < state.addressPath.length; i++) {
+            state.tokenPath[i] = IERC20Token(state.addressPath[i]);
+        }
+
+        // // Grant an allowance to the Bancor Network to spend `fromTokenAddress` token.
+        uint256 fromTokenBalance = state.tokenPath[0].balanceOf(address(this));
+        LibERC20Token.approveIfBelow(state.addressPath[0], contractRegistry.addressOf(bancorNetworkName), fromTokenBalance);
```
dekz: same with comment as below

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -0,0 +1,36 @@
...
+export class BancorService {
+    // Bancor recommends setting this value to 2% under the expected return amount
+    public minReturnAmountBufferPercentage = 0.98;
+    private _sdk?: SDK;
+
+    constructor(public ethereumNodeEndpoint: string) {}
+
+    public async getSDKAsync(): Promise<SDK> {
+        if (!this._sdk) {
+            this._sdk = await SDK.create({ ethereumNodeEndpoint: this.ethereumNodeEndpoint });
```
dekz: is there a way to give it a `provider` instead of having to pass down the RPC url?

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+        return this._sdk;
+    }
+
+    public async getQuoteAsync(fromToken: string, toToken: string, amount: BigNumber): Promise<BancorQuoteData> {
+        const sdk = await this.getSDKAsync();
+        const { path, rate } = await sdk.pricing.getPathAndRate(token(fromToken), token(toToken), amount.toString());
```
dekz: does this make an RPC request for every amount? That seems like a lot of (13) requests?

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -0,0 +1,36 @@
...
+        return {
+            fillData: { path: path.map(p => p.blockchainId) }, // ethereum token addresses
```
dekz: this may also give different paths given different amounts?
E.g $100 gives A->B, $500 gives A->B->C. The sampler expects these to all be the same "route/source" so can encounter errors trying to fill for $600 if A->B or A->B->C do not have sufficient liquidity.

We should see if there are SDK functions that allow you to efficiently query with a hint of a path, or like Balancer, select a few candidates and query those.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -287,6 +291,42 @@ export const samplerOperations = {
...
+        bancorService: BancorService,
+    ): SourceQuoteOperation<FillData, BancorQuoteData> {
+        return {
+            source: ERC20BridgeSource.Bancor,
+            encodeCall: _contract => {
+                return '0x';
+            },
+            handleCallResultsAsync: async (_contract, _callResults) => {
+                return Promise.all(
+                    makerFillAmounts.map(amt => bancorService.getQuoteAsync(takerToken, makerToken, amt)), // tslint:disable-line:promise-function-async
```
dekz: Without a side this would be ambiguous and incorrect. 

Buying is `receive 100 of makerToken by selling X takerToken`. 
Selling is `receive X of makerToken by selling 100 takerToken`.

This is performing `receive X of takerToken by selling 100 makerToken`.

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+    {
+        // hold variables to get around stack depth limitations
+        TransferState memory state;
+
+        // Decode the bridge data.
+        (
+            state.addressPath,
+            state.minReturn
+        // solhint-disable indent
+        ) = abi.decode(bridgeData, (address[], uint));
```
dorothy-zbornak: Should be able to just do `abi.decode(IERC20Token[], ...` so it fits with all the types

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,135 @@
...
+
+
+contract BancorBridge is
+    IERC20Bridge,
+    IWallet,
+    DeploymentConstants
+{
+
+    IContractRegistry contractRegistry = IContractRegistry(0x52Ae12ABe5D8BD778BD5397F99cA900624CfADD4);
+    bytes32 bancorNetworkName = bytes32(uint(0x42616e636f724e6574776f726b)); // "BancorNetwork"
```
dorothy-zbornak:
```suggestion
    bytes32 internal constant bancorNetworkName = bytes32(uint(0x42616e636f724e6574776f726b)); // "BancorNetwork"
```

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dorothy-zbornak: I wonder if we can just use the fillData from the last fill node when constructing the order.

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: Haven't found one unfortunately :/

dorothy-zbornak: It looks like you can pass in a web3 provider, might try just wrapping our provider in something like `{ sendAsync(payload) => provider.send(payload) }` or whatever works. https://github.com/bancorprotocol/sdk/blob/master/src/types.ts

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
dorothy-zbornak: Doesn't `getPathAndRate()` take a string for `amt`? https://github.com/bancorprotocol/sdk/blob/master/src/pricing.ts#L17

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dorothy-zbornak: I *think* it actually caches the reserve graph and finds paths based off that. https://github.com/bancorprotocol/sdk/blob/453a13bf6d7d61e7136a004eabf6bc1f242b6f0e/src/blockchains/ethereum/index.ts#L77

This does seem to imply that we need a timer to routinely call `sdk.refresh()` though.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
dorothy-zbornak: Yeah let's just backlog buy support for bancor right now?

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: yeah it does :/ paths are cached but each call to `getPathAndRate` refreshes the available rates in liquidity pools. basically every call to `getPathAndRate` results in 1 RPC call.

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: we can query with a full path or with source + target token, but I don't think we can provide a partial path. 

I'm making changes so `fillData.path` is unique to the amount quoted, so that should resolve the fill data problem. But perhaps still inefficient with the number of RPC calls per quote (1 per quote amount).

xianny: did a deeeeep dive with @dorothy-zbornak in current implementation of path optimisation. as suggested, we can use the `fillData` from the last node.

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -0,0 +1,122 @@
...
+        (
+            state.path,
+            state.bancorNetworkAddress
+        // solhint-disable indent
+        ) = abi.decode(bridgeData, (address[], address));
+        // solhint-enable indent
+
+        require(state.path.length > 2, "BancorBridge/PATH_LENGTH_MUST_BE_GREATER_THAN_TWO");
+        require(state.path[state.path.length - 1] == toTokenAddress, "BancorBridge/LAST_ELEMENT_OF_PATH_MUST_MATCH_OUTPUT_TOKEN");
+
```
dorothy-zbornak: So the shortest path is `[fromToken, converter, toToken]`, right?

contracts/asset-proxy/contracts/test/TestBancorBridge.sol
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    function setNetworkRevertReason(string calldata revertReason)
+        external
+    {
+        _testNetwork.setRevertReason(revertReason);
+    }
+
+    /// @dev Sets the balance of this contract for an existing token.
+    ///      The wei attached will be the balance.
```
dorothy-zbornak:
```suggestion
```

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
dorothy-zbornak: Does the above hack ^ work? Would also rather not pass in an RPC url.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -241,21 +244,23 @@ export class MarketOperationUtils {
...
+                this._sampler.bancorService,
             ),
             // Get buy quotes for taker -> maker.
             await DexOrderSampler.ops.getBuyQuotesAsync(
                 difference(
                     BUY_SOURCES.concat(
                         this._liquidityProviderRegistry !== NULL_ADDRESS ? [ERC20BridgeSource.LiquidityProvider] : [],
                     ),
-                    _opts.excludedSources.concat(ERC20BridgeSource.Balancer),
+                    _opts.excludedSources.concat([ERC20BridgeSource.Balancer]),
```
dorothy-zbornak, line 255: :see_no_evil:

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -0,0 +1,53 @@
...
+        amount: BigNumber = new BigNumber(1),
+    ): Promise<BancorQuoteData> {
+        const sdk = await this.getSDKAsync();
+        const { path, rate } = await sdk.pricing.getPathAndRate(token(fromToken), token(toToken), amount.toString());
+        return {
+            fillData: {
+                path: path.map(p => p.blockchainId),
+                networkAddress: await this.getBancorNetworkAddressAsync(),
+            }, // ethereum token addresses
+            amount: new BigNumber(rate).multipliedBy(this.minReturnAmountBufferPercentage),
```
dorothy-zbornak: Do we need to round this to an integer?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -456,15 +485,35 @@ export const samplerOperations = {
...
                 ),
             ),
         );
-        const samplerOps = subOps.filter(op => op.source !== ERC20BridgeSource.Balancer);
-        const nonSamplerOps = subOps.filter(op => op.source === ERC20BridgeSource.Balancer);
+        const nonSamplerSources = [ERC20BridgeSource.Balancer, ERC20BridgeSource.Bancor];
+        const samplerOps: Array<SourceQuoteOperation<FillData, QuoteData>> = [];
+        const nonSamplerOps: Array<SourceQuoteOperation<FillData, QuoteData>> = [];
+        subOps.forEach(op => {
+            if (nonSamplerSources.find(s => s === op.source) !== undefined) {
```
dorothy-zbornak:
```suggestion
            if (nonSamplerSources.includes(op.source)) {
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -265,7 +277,8 @@ export interface BatchedOperation<TResult> {
     handleCallResultsAsync(contract: ERC20BridgeSamplerContract, callResults: string): Promise<TResult>;
 }
 
-export interface SourceQuoteOperation<TFillData extends FillData = FillData> extends BatchedOperation<BigNumber[]> {
+export interface SourceQuoteOperation<TFillData extends FillData = FillData, TQuoteData extends QuoteData = BigNumber>
```
dorothy-zbornak: We should just make it so *all* sources return `BancorQuoteData` (but just called `QuoteData`) in `handleCallResultAsync()`. This one edge case makes the code confusing.

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: yeaaa it works! sorry I missed it on the first go round.

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: chatted off-thread. we are going to leave it this way for now, pending bancor SDK upgrade to accept multiple amounts.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
xianny: made changes for `getSellQuote`, plz confirm sensical

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
xianny: yup correct

packages/asset-swapper/src/utils/market_operation_utils/index.ts
xianny: `concat` works on individual elements *and* arrays because types aren't real in node 😭 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

packages/asset-swapper/src/utils/market_operation_utils/types.ts
xianny: refactored

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
xianny: discussed off-thread. amounts have to be converted to/from wei instead.

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
```diff
@@ -69,17 +69,16 @@ contract BancorBridge is
         ) = abi.decode(bridgeData, (address[], address));
         // solhint-enable indent
 
-        require(state.path.length > 2, "BancorBridge/PATH_LENGTH_MUST_BE_GREATER_THAN_TWO");
-        require(state.path[state.path.length - 1] == toTokenAddress, "BancorBridge/LAST_ELEMENT_OF_PATH_MUST_MATCH_OUTPUT_TOKEN");
-
         // Just transfer the tokens if they're the same.
         if (state.path[0] == toTokenAddress) {
```
dorothy-zbornak, line 74: ~Will~ Can revert without the `path.length` check.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -201,7 +201,7 @@ export class SwapQuoter {
                 gas: samplerGasLimit,
             },
         );
-        const bancorService = options.ethereumRpcUrl ? new BancorService(options.ethereumRpcUrl) : undefined;
+        const bancorService = options.ethereumRpcUrl ? new BancorService(provider) : undefined;
```
dorothy-zbornak: :tada:

packages/asset-swapper/src/utils/market_operation_utils/bancor_service.ts
```diff
@@ -33,15 +35,23 @@ export class BancorService {
         fromToken: string,
         toToken: string,
         amount: BigNumber = new BigNumber(1),
-    ): Promise<BancorQuoteData> {
+    ): Promise<Quote<BancorFillData>> {
         const sdk = await this.getSDKAsync();
-        const { path, rate } = await sdk.pricing.getPathAndRate(token(fromToken), token(toToken), amount.toString());
+        const blockchain = sdk._core.blockchains[BlockchainType.Ethereum] as Ethereum;
```
dorothy-zbornak, line 40: Wondering: can we set `sdk._core.blockchains[BlockchainType.Ethereum]` to a custom `Ethereum` instance that batches RPC calls?

contracts/asset-proxy/test/bancor_bridge.ts
```diff
@@ -108,7 +108,10 @@ blockchainTests.resets('Bancor unit tests', env => {
                 // Transfer amount.
                 _opts.amount,
                 // ABI-encode the input token address as the bridge data. // FIXME
```
dorothy-zbornak: `FIXME`?

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -530,13 +561,9 @@ export const samplerOperations = {
...
                         source: op.source,
-                        output:
-                            op.source === ERC20BridgeSource.Bancor
-                                ? (output as BancorQuoteData).amount
-                                : (output as BigNumber), // hack; all QuoteData other than Bancor defaults to BigNumber
+                        output: output.amount,
                         input: takerFillAmounts[j],
-                        fillData:
-                            op.source === ERC20BridgeSource.Bancor ? (output as BancorQuoteData).fillData : op.fillData, // Bancor has unique fillData per sample
+                        fillData: output.fillData,
```
dorothy-zbornak: :+1:

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -621,8 +644,8 @@ export const samplerOperations = {
             ),
         );
         const nonSamplerSources = [ERC20BridgeSource.Balancer, ERC20BridgeSource.Bancor];
-        const samplerOps: Array<SourceQuoteOperation<FillData, QuoteData>> = [];
-        const nonSamplerOps: Array<SourceQuoteOperation<FillData, QuoteData>> = [];
+        const samplerOps: Array<SourceQuoteOperation<FillData>> = [];
+        const nonSamplerOps: Array<SourceQuoteOperation<FillData>> = [];
         subOps.forEach(op => {
             if (nonSamplerSources.find(s => s === op.source) !== undefined) {
```
dorothy-zbornak, line 704:
```suggestion
            if (nonSamplerSources.includes(op.source)) {
```

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -101,11 +101,9 @@ export interface BancorFillData extends FillData {
     networkAddress: string;
 }
 
-export interface QuoteData {}
-
-export interface BancorQuoteData extends QuoteData {
-    fillData: BancorFillData;
```
dorothy-zbornak: :+1:

dorothy-zbornak (approved):
Just minor nits, but wtg!
doublicat

contracts/asset-proxy/test/bancor_bridge.ts
xianny: woops that's outdated

contracts/asset-proxy/contracts/src/bridges/BancorBridge.sol
xianny: If it's the same token, then `path.length = 1`. I'll add a `path.length > 0` check.

xianny:
Simbotted with Bancor enabled in https://github.com/dorothy-zbornak/0x-swap-server and got some data:

Screen Shot 2020-08-21 at 9 29 11 AM
No surprise here.



Screen Shot 2020-08-21 at 9 28 12 AM
Bancor revert rate a bit worrying.


Screen Shot 2020-08-21 at 9 28 45 AM
Bancor slippage also higher than the rest.



Screen Shot 2020-08-21 at 9 30 03 AM
Bancor TOKEN slippage is good though. Maybe we enable Bancor only for BNT for now??



Screen Shot 2020-08-21 at 9 24 57 AM
Just for the record.


---

A-B testing with prod:

Screen Shot 2020-08-21 at 9 37 53 AM
Screen Shot 2020-08-21 at 9 38 25 AM
Screen Shot 2020-08-21 at 9 36 01 AM
Screen Shot 2020-08-21 at 9 37 03 AM
Screen Shot 2020-08-21 at 9 36 36 AM

xianny:
With latest asset-swapper (Bancor disabled)

Screen Shot 2020-08-21 at 1 18 02 PM
Screen Shot 2020-08-21 at 1 18 24 PM
Screen Shot 2020-08-21 at 1 19 00 PM

---
Little bit worried about Balancer reverts looking uneven so I ran it for a while with just Balancer and Uniswap as sources:
Screen Shot 2020-08-21 at 3 06 26 PM

---

Screen Shot 2020-08-21 at 1 19 37 PM
Screen Shot 2020-08-21 at 1 19 59 PM

PR #2657, opened 2020-08-05
EP: MetaTransactions fixes

dorothy-zbornak:
## Description

- Adds a reentrancy guard to `executeMetaTransaction()` and `batchExecuteMetaTransactions()`.
  - This introduced new storage, `LibReentrancyGuardStorage`, and a new fixin `FixinReentrancyGuard`.
    - This is a flag-based reentrancy guard model, so we can have both distinct and overlapping reentrancy guards.
- Adds an ETH refund mechanism to meta-transactions.
- `transformERC20()` will now pass the `msg.sender` to transformers.
  - This is so FQT can refund any unused protocol fee to the sender, which may be distinct from the taker.
    - In the case of mtxs, the sender would be the EP itself, then the mtx refund mechanism would transfer it to the mtx relayer.
- Addresses internal feedback.
- Merged #2670 CD spot check feedback.
- Merged #2624 Fix `TransformerDeployer` so we can `kill()` transformers.
- Merged #2671 feature contract renames + minor cleanup.
- Rework nested revert error decoding to handle some edge cases.
- Fix shadowed variable in `TransformerDeployer`.
- Merged #2692 RFQT taker enforcement, kyber and mooni rollups, deployed addresses, etc.

## Checklist:


contracts/zero-ex/contracts/src/features/SignatureValidator.sol
```diff
@@ -37,6 +37,13 @@ contract SignatureValidator is
     using LibBytesV06 for bytes;
     using LibRichErrorsV06 for bytes;
 
+    /// @dev Exclusive upper limit on ECDSA signatures 'R' values.
+    ///      The valid range is given by fig (282) of the yellow paper.
```
hysz, line 41: +1 on ref!

contracts/zero-ex/contracts/src/features/SignatureValidator.sol
```diff
@@ -160,12 +167,18 @@ contract SignatureValidator is
...
-            recovered = ecrecover(
-                hash,
-                v,
-                r,
-                s
-            );
+            if (v < 27) {
+                // Handle clients that encode v as 0 or 1.
+                v += 27;
+            }
```
hysz: It looks like the valid values for `v` are `{27,28}`, after adjustment. Not sure if there's a strong security reason for verifying, but we could also assert this range.

contracts/zero-ex/contracts/src/features/SignatureValidator.sol
```diff
@@ -179,15 +192,21 @@ contract SignatureValidator is
...
-                )),
-                v,
-                r,
-                s
-            );
+            if (v < 27) {
+                // Handle clients that encode v as 0 or 1.
+                v += 27;
+            }
+            if (uint256(r) < ECDSA_SIGNATURE_R_LIMIT && uint256(s) < ECDSA_SIGNATURE_S_LIMIT) {
```
hysz, line 187: Same as above.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
```diff
@@ -249,6 +257,17 @@ contract FillQuoteTransformer is
                     ).rrevert();
             }
         }
+
+        // Refund unspent protocol fees.
+        if (state.ethRemaining > 0 && data.refundReceiver != address(0)) {
```
hysz, line 270: Would it would indicate an internal error when `state.ethRemaining > 0 && data.refundReceiver  == 0`? Wondering if we should `revert` in that scenario.

hysz (approved):
Looks great, nice work! Left a couple small questions but looks g2g.

contracts/zero-ex/contracts/src/transformers/FillQuoteTransformer.sol
dorothy-zbornak: No. That's the behavior we have now. The `PayTakerTransformer` will funnel the unspent ETH to the taker.

contracts/zero-ex/contracts/src/features/SignatureValidator.sol
dorothy-zbornak: Anything outside that range should just cause `ecrecover()` to fail.

dorothy-zbornak: I honestly think we should remove the `s` and `r` range checks. Signatures don't get included in any of our hashes, so replay isn't an issue. Plus, allowing `v` to be `0-1` or `27-28` introduces a new malleability since you can just take the same signature, flip `v` to `0-1`, and it'll be made valid.

hysz: Oh interesting, that makes sense. Yeah I'm down to nix them. Perhaps just run by Paul real quick, since it was his recommendation.

dorothy-zbornak: Paul's recco is to actually just drop the parity `v` support.

moodlezoup (approved):
good stuff, good tests, what's not to like?

hysz (approved):
Updates LGTM!

moodlezoup (approved):
lgtm

contracts/zero-ex/CHANGELOG.json
```diff
@@ -1,4 +1,49 @@
...
+            {
+                "note": "Address CD post-audit feedback",
+                "pr": 2657
+            },
+            {
+                "note": "Add `LogMetadataTransformer`",
+                "pr": 2657
+            },
+            {
+                "note": "Add `IUniswapV2Feature`",
```
moodlezoup: no longer in this PR

PR #2658, opened 2020-08-05
AS: affiliate fees

dorothy-zbornak:
## Checklist:

### DIFF
--- a/packages/asset-swapper/src/types.ts
+++ b/packages/asset-swapper/src/types.ts
@@ -124,13 +124,21 @@ export interface ForwarderExtensionContractOpts {
     feeRecipient: string;
 }
 
+export interface AffiliateFee {
+    recipient: string;
+    buyTokenFeeAmount: BigNumber;
+    sellTokenFeeAmount: BigNumber;
+}
+
 /**
  * @param isFromETH Whether the input token is ETH.
  * @param isToETH Whether the output token is ETH.
+ * @param affiliateFee Fee denominated in taker or maker asset to send to specified recipient.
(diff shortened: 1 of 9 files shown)


packages/asset-swapper/src/types.ts
```diff
@@ -124,13 +124,21 @@ export interface ForwarderExtensionContractOpts {
     feeRecipient: string;
 }
 
+export interface AffiliateFee {
+    recipient: string;
+    buyTokenFeeAmount: BigNumber;
```
dorothy-zbornak, line 129: MM wants percentage fees.

packages/order-utils/src/transformer_data_encoders.ts
```diff
@@ -147,17 +147,11 @@ export function decodePayTakerTransformerData(encoded: string): PayTakerTransfor
  */
 export const affiliateFeeTransformerDataEncoder = AbiEncoder.create({
```
dorothy-zbornak: All this needs to be reverted to original. I goofed. `AbiEncoder` can't handle this kind of encoding for some reason and it ends up reverting.

packages/asset-swapper/src/types.ts
dorothy-zbornak: Or I guess you're gonna do the conversion in API?

moodlezoup: I'm handling the percentages from the 0x-api side

dekz (commented):
LGTM

PR #2669, opened 2020-08-18
fix: asset-swapper empty batch call

dekz:
## Description

Prevent an empty sampler batch call when all operations are NOOPs

Also returns `1337` ganache addresses back to expected for Mesh/contract-addresses

## Checklist:


moodlezoup (approved):

packages/contract-artifacts/src/index.ts
```diff
@@ -37,6 +38,7 @@ export {
     ERC1155Mintable,
     ERC1155Proxy,
     ERC20BridgeProxy,
+    ERC20BridgeSampler,
```
dorothy-zbornak: Can't whoever just import this and the wrapper from `@0x/asset-swapper`?

dorothy-zbornak (approved):

packages/contract-artifacts/src/index.ts
dorothy-zbornak: Defeats the whole point of bundling the contracts since we have to gitpkg those two again. :cry:

packages/asset-swapper/src/contracts.ts
```diff
@@ -0,0 +1 @@
+export { ERC20BridgeSamplerContract } from '../generated-wrappers/erc20_bridge_sampler';
```
dekz, line 1: this gets around an `Didn't find an exportPath for` error

packages/asset-swapper/contracts/src/ERC20BridgeSampler.sol
```diff
@@ -51,6 +51,9 @@ contract ERC20BridgeSampler is
     {
         callResults = new bytes[](callDatas.length);
         for (uint256 i = 0; i != callDatas.length; ++i) {
+            if (callDatas[i].length == 0) {
```
dekz, line 54: prevents any leakages from causing a revert

packages/asset-swapper/src/contracts.ts
dorothy-zbornak: This is so weird. Where do you see this error?

packages/migrations/src/migration.ts
```diff
@@ -293,27 +292,21 @@ export async function runMigrationsAsync(
         etherToken.address,
     );
 
-    const erc20BridgeSampler = await ERC20BridgeSamplerContract.deployFrom0xArtifactAsync(
-        assetSwapperArtifacts.ERC20BridgeSampler,
+    // JAM
+    // tslint:disable-next-line:no-unused-variable
+    const jamToken = await DummyERC20TokenContract.deployFrom0xArtifactAsync(
```
dorothy-zbornak, line 297: Why do we deploy this?

PR #2677, opened 2020-08-24
feat: asset-swapper DODO

dekz:
## Description

* Adds DODO, but not the best way to interact with it as you can only buyETH or sellETH. No buyUSDC or sellUSDC to it. Makes it less useful, especially if it is out of price parity

## Checklist:


contracts/asset-proxy/contracts/src/bridges/DODOBridge.sol
```diff
@@ -0,0 +1,110 @@
...
+        address to,
+        uint256 amount,
+        bytes calldata bridgeData
+    )
+        external
+        returns (bytes4 success)
+    {
+        // Decode the bridge data to get the `fromTokenAddress`.
+        address fromTokenAddress = abi.decode(bridgeData, (address));
+        IDODO pool = IDODO(IDODOZoo(_getDODOZooAddress()).getDODO(fromTokenAddress, toTokenAddress));
```
dekz, line 60: TODO could chuck this in `bridgeData`

packages/asset-swapper/contracts/src/DODOSampler.sol
```diff
@@ -0,0 +1,99 @@
...
+    function sampleBuysFromDODO(
+        address takerToken,
+        address makerToken,
+        uint256[] memory makerTokenAmounts
+    )
+        public
+        view
+        returns (uint256[] memory takerTokenAmounts)
+    {
+        return takerTokenAmounts;
```
dekz, line 96: Whilst buys (of ETH) are supported, we perform this as a sell of USDC. So not entirely useful.

TODO could support buy USDC by sellETH, but `buyAmount` is less frequently used

PR #2684, opened 2020-08-26
asset-swapper: Stop requiring takerAddress for indicative RFQ-T quotes.

feuGeneA:
## Description

Stop requiring takerAddress for indicative RFQ-T quotes.

## Testing instructions

Verified via automated tests over in https://github.com/0xProject/0x-api/pull/320

## Types of changes


PirosB3 (changes requested):
When the QuoteRequestor makes a request to the RFQ client, they need to pass in a `takerAddress`. With these changes, do we risk an undefined `takerAddress`?

feuGeneA:
> When the QuoteRequestor makes a request to the RFQ client, they need to pass in a `takerAddress`. With these changes, do we risk an undefined `takerAddress`?

That's a great question @PirosB3 , thank you for asking it.  Thinking further, I think what we need to do is modify the `@0x/quote-server` to allow indicative quotes to skip passing a `takerAddress`.  I'll put up a PR for that.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -217,7 +213,10 @@ export class QuoteRequestor {
         options: RfqtRequestOpts,
     ): Promise<RFQTIndicativeQuote[]> {
         const _opts: RfqtRequestOpts = { ...constants.DEFAULT_RFQT_REQUEST_OPTS, ...options };
-        assertTakerAddressOrThrow(_opts.takerAddress);
+
+        if (!_opts.takerAddress) {
```
PirosB3, line 222: Would you mind adding a quick comment explaining why we set indicative quote `takerAddress` to `NULL_ADDRESS`?

PirosB3 (approved):

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2684/commits/a8f3a47240c1a0adabbdd60ca82c8c98b4f07851

PR #2686, opened 2020-08-27
fix: reduce excess order hash in quote report

dekz:
## Description

Reduces the amount of order hash generation in the quote report generator

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/quote_report_generator.ts
+++ b/packages/asset-swapper/src/utils/quote_report_generator.ts
@@ -1,7 +1,6 @@
 import { orderHashUtils } from '@0x/order-utils';
 import { SignedOrder } from '@0x/types';
 import { BigNumber } from '@0x/utils';
-import * as _ from 'lodash';
 
 import { MarketOperation } from '../types';
 
@@ -53,6 +52,11 @@ export interface QuoteReport {
     sourcesDelivered: QuoteReportSource[];
 }
 
+interface SignedOrderWithMetadata extends SignedOrder {
+    hash: string;
+    fillableAmount: BigNumber;
+}
+
 const nativeOrderFromCollapsedFill = (cf: CollapsedFill): SignedOrder | undefined => {
     // Cast as NativeCollapsedFill and then check
     // if it really is a NativeCollapsedFill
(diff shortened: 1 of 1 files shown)


packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -180,11 +180,10 @@ function _multiHopSampleToReportSource(
 }
 
 function _nativeOrderToReportSource(
-    nativeOrder: SignedOrder,
-    fillableAmount: BigNumber,
+    nativeOrderWithMetaData: SignedOrderWithMetadata,
     quoteRequestor?: QuoteRequestor,
 ): NativeRFQTReportSource | NativeOrderbookReportSource {
-    const orderHash = orderHashUtils.getOrderHash(nativeOrder);
+    const { hash: orderHash, fillableAmount, ...nativeOrder } = nativeOrderWithMetaData;
```
PirosB3: isn't it better to just rename `orderHash`?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -205,7 +204,6 @@ function _nativeOrderToReportSource(
         };
         return rfqtSource;
     } else {
-        // if it's not an rfqt order, treat as normal
```
PirosB3: why remove?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -109,11 +110,10 @@ export function generateQuoteReport(
...
             if (foundNativeOrder) {
-                return _nativeOrderToReportSource(
-                    foundNativeOrder,
-                    orderHashesToFillableAmounts[orderHashUtils.getOrderHash(foundNativeOrder)],
-                    quoteRequestor,
-                );
+                const matchedOrder = orderWithMetadata.find(
+                    o => o.signature === foundNativeOrder.signature && o.salt === foundNativeOrder.salt,
+                )!;
+                return _nativeOrderToReportSource(matchedOrder, quoteRequestor);
```
PirosB3, line 116: should you check if it exists?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -82,18 +86,15 @@ export function generateQuoteReport(
         // length mismatch, abort
         throw new Error('orderFillableAmounts must be the same length as nativeOrders');
     }
-    const orderHashesToFillableAmounts: { [orderHash: string]: BigNumber } = {};
-    nativeOrders.forEach((nativeOrder, idx) => {
-        orderHashesToFillableAmounts[orderHashUtils.getOrderHash(nativeOrder)] = orderFillableAmounts[idx];
-    });
+    const orderWithMetadata: SignedOrderWithMetadata[] = nativeOrders.map((o, i) => ({
```
PirosB3, line 89: how many orders are we dealing with here? what is N?

PirosB3: does it make sense to keep this a map?

PirosB3 (approved):

PR #2688, opened 2020-08-28
[WIP] Efficient Uniswap

recmo:
## Checklist:


contracts/zero-ex/contracts/src/features/AsmUniswap.sol
```diff
@@ -60,48 +52,74 @@ contract AsmUniswapFeature is {
...
-            mstore(0x00, SWAP_SELECTOR)
+            // Call PAIR.swap(0, wantAmount, msg.sender, new bytes(0))
             mstore(0x04, div(
-                mul(mul(calldataload(0x04), 997), mload(0x20)),
-                add(mul(calldataload(0x04), 997), mul(mload(0x00), 1000))
-            )
+                mul(mul(calldataload(0x04), 997), mload(0x00)),
+                add(mul(calldataload(0x04), 997), mul(mload(0x20), 1000))
+            ))
+            mstore(0x00, SWAP_SELECTOR)
```
recmo: This partially overwrites the previous value.

contracts/zero-ex/contracts/src/features/DirectUniswap.sol
```diff
@@ -0,0 +1,88 @@
...
+            pair,
+            haveAmount
+        ));
+
+        // Compute `wantAmount` and call pair contract for swap
+        // Formula from `UniswapV2Library.getAmountOut`
+        // <https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/libraries/UniswapV2Library.sol#L43>
+        // TODO: Compare above formula with the assertion in `UniswapV2Pair.swap`
+        // <https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Pair.sol#L179>
+        if (haveToken < wantToken) {
```
dorothy-zbornak, line 73: What does this do?

recmo: There is only one UniswapV2Pair instance per token pair. So the "Weth to Dai" and "Dai to Weth" contract are the same. The way Uniswap does this is by sorting the token addresses when computing the pair address. And then the `getReserves()` and `swap(amountOut0, amountOut1, to)` functions also use the sorted order.

PR #2691, opened 2020-09-01
`@0x/contracts-zero-ex`: LiquidityProviderFeature

moodlezoup:
## Description
- Adds a ZeroEx feature to swap directly with liquidity providers.
- ~Updates `ExchangeProxySwapQuoteConsumer` to leverage the new feature~ moved to https://github.com/0xProject/0x-monorepo/pull/2709
- Refactors market_operation_utils significantly, introducing a `Path` class to replace `Fill[]` and accounts for exchange proxy gas overhead in the path optimizer. EP gas overhead is a function of the sources used by a path, which we need to consider with the advent of these "source VIP" features

## Checklist:


contracts/zero-ex/contracts/src/features/LiquidityProviderSwap.sol
```diff
@@ -0,0 +1,201 @@
...
+    /// @dev Store the WETH address in an immutable.
+    /// @param weth_ The weth token.
+    constructor(IEtherTokenV06 weth_)
+        public
+        FixinCommon()
+    {
+        weth = weth_;
+    }
+
+    function sellToLiquidityProvider(
```
hysz: I think we should probably emit an event in here for parity with the `FillQuoteTransformer`. I believe [this event ](https://github.com/0xProject/0x-monorepo/blob/development/contracts/zero-ex/contracts/src/features/TransformERC20.sol#L280) is used by our data integrations.

hysz (approved):
One cheeky note, other than that draft looks solid!

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
```diff
@@ -0,0 +1,204 @@
...
+    /// @dev Store the WETH address in an immutable.
+    /// @param weth_ The weth token.
+    constructor(IEtherTokenV06 weth_)
+        public
+        FixinCommon()
+    {
+        weth = weth_;
+    }
+
+    function sellToLiquidityProvider(
```
dorothy-zbornak, line 78: There's no way to do a buy right?

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
```diff
@@ -0,0 +1,204 @@
...
+    {
+        address providerAddress = getLiquidityProviderForMarket(makerToken, takerToken);
+        if (recipient == address(0)) {
+            recipient = msg.sender;
+        }
+
+        if (takerToken == ETH_TOKEN_ADDRESS) {
+            // Wrap ETH.
+            weth.deposit{value: sellAmount}();
+            weth.transfer(providerAddress, sellAmount);
```
dorothy-zbornak, line 98: If only WETH had the foresight to implement `depositTo()`.

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
```diff
@@ -0,0 +1,204 @@
...
+                address(weth),
+                address(0),
+                address(this),
+                minBuyAmount,
+                ""
+            );
+            boughtAmount = weth.balanceOf(address(this)).safeSub(balanceBefore);
+            // Unwrap wETH and send ETH to recipient.
+            weth.withdraw(boughtAmount);
+            recipient.transfer(boughtAmount);
```
dorothy-zbornak, line 120: And `withdrawTo()` :disappointed:

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
```diff
@@ -0,0 +1,204 @@
...
+
+    /// @dev Store the WETH address in an immutable.
+    /// @param weth_ The weth token.
+    constructor(IEtherTokenV06 weth_)
+        public
+        FixinCommon()
+    {
+        weth = weth_;
+    }
+
```
dorothy-zbornak, line 64: Don't forget your `migrate()` fn.

dorothy-zbornak (commented):
Yup, yup, looking good.

contracts/zero-ex/contracts/src/features/LiquidityProviderFeature.sol
moodlezoup: yeah no way to do a buy atm

contracts/zero-ex/contracts/src/features/ILiquidityProviderFeature.sol
```diff
@@ -0,0 +1,66 @@
...
+
+/// @dev Feature to swap directly with an on-chain liquidity provider.
+interface ILiquidityProviderFeature {
+    event LiquidityProviderForMarketUpdated(
+        address indexed xAsset,
+        address indexed yAsset,
+        address providerAddress
+    );
+
+    function sellToLiquidityProvider(
```
dorothy-zbornak, line 31: +natspec

contracts/zero-ex/contracts/src/features/ILiquidityProviderFeature.sol
```diff
@@ -0,0 +1,66 @@
...
+
+*/
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+
+/// @dev Feature to swap directly with an on-chain liquidity provider.
+interface ILiquidityProviderFeature {
+    event LiquidityProviderForMarketUpdated(
```
dorothy-zbornak, line 25: +natspec

packages/asset-swapper/src/quote_consumers/exchange_proxy_swap_quote_consumer.ts
```diff
@@ -251,8 +253,11 @@ function isBuyQuote(quote: SwapQuote): quote is MarketBuySwapQuote {
     return quote.type === MarketOperation.Buy;
 }
 
-function shouldGoDirectlyToLiquidityProvider(quote: SwapQuote): boolean {
+function shouldGoDirectlyToLiquidityProvider(quote: SwapQuote, affiliateFee: AffiliateFee): boolean {
     return (
-        quote.orders.length === 1 && _.get(quote, 'orders[0].fills[0].source') === ERC20BridgeSource.LiquidityProvider
+        quote.orders.length === 1 &&
```
dorothy-zbornak: I added an `isMetaTransaction` field to the opts struct in mine so we don't create a direct quote in those cases. Maybe the same should be done here.

dorothy-zbornak (commented):
Feature looks good, still reviewing the refactors.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -87,254 +107,70 @@ function nativeOrdersToPath(
...
-    for (const fill of path) {
-        if (input.gte(targetInput)) {
-            break;
+): Fill[] {
+    const sourcePathId = hexUtils.random();
+    const fills: Fill[] = [];
+    // Drop any non-zero entries. This can occur if the any fills on Kyber were UniswapReserves
+    // We need not worry about Kyber fills going to UniswapReserve as the input amount
+    // we fill is the same as we sampled. I.e we received [0,20,30] output from [1,2,3] input
+    // and we only fill [2,3] on Kyber (as 1 returns 0 output)
```
dorothy-zbornak, line 145: What happens if we try to partial fill the `[1->0,2->20,3->30]` kyber path with `input = 0.5`? Like if it gets included at the end the quote?

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
```diff
@@ -87,254 +107,70 @@ function nativeOrdersToPath(
...
+        const { source, fillData } = sample;
+        const input = sample.input.minus(prevSample ? prevSample.input : 0);
+        const output = sample.output.minus(prevSample ? prevSample.output : 0);
+        const fee = fees[source] === undefined ? 0 : fees[source]!(sample.fillData);
+        let penalty = ZERO_AMOUNT;
+        if (i === 0) {
+            // Only the first fill in a DEX path incurs a penalty.
+            penalty = !ethToOutputRate.isZero()
+                ? ethToOutputRate.times(fee)
+                : ethToInputRate.times(fee).times(output.dividedToIntegerBy(input));
```
dorothy-zbornak, line 159: Do you know when we would not have an `ethToOutputRate`? Token prelaunch?

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -283,6 +274,7 @@ export interface GetMarketOrdersOpts {
      * Estimated gas consumed by each liquidity source.
      */
     gasSchedule: FeeSchedule;
+    exchangeProxyOverhead: ExchangeProxyOverhead;
```
dorothy-zbornak, line 282: Docstring?

packages/asset-swapper/src/utils/market_operation_utils/rate_utils.ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { MarketOperation } from '../../types';
+
+import { ZERO_AMOUNT } from './constants';
+import { DexSample, ERC20BridgeSource, FeeSchedule, MultiHopFillData } from './types';
+
+/**
+ * Returns the fee-adjusted rate of a two-hop quote. Returns zero if the
+ * quote falls short of the target input.
+ */
```
dorothy-zbornak, line 11: tyy for doccing these

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,152 +1,103 @@
...
-                    input.plus(BigNumber.min(remainingInput, fill.input)),
-                    output.plus(
-                        // Clip the output of the next fill to the remaining
-                        // input.
-                        clipFillAdjustedOutput(fill, remainingInput),
-                    ),
-                    flags | fill.flags,
-                    nextRemainingFills,
-                );
+                _walk(Path.clone(path).append(fill), nextRemainingFills);
```
dorothy-zbornak, line 70: Do you think this has a noticeable impact on efficiency?

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
```diff
@@ -1,152 +1,103 @@
...
-            return rateBySource[b.sourcePathId].comparedTo(rateBySource[a.sourcePathId]);
+            return rates[b.sourcePathId].comparedTo(rates[a.sourcePathId]);
         }
         return a.index - b.index;
     });
-    _walk([], ZERO_AMOUNT, ZERO_AMOUNT, 0, sortedFills);
-    if (!isValidPath(bestPath)) {
+    _walk(Path.create(side, [], targetInput, pathA.pathPenaltyOpts), sortedFills);
+    if (!bestPath.isValid()) {
         throw new Error('nooope');
```
dorothy-zbornak, line 85: Whoa, is this mine? lol we can remove this.

packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -0,0 +1,287 @@
...
+                i += 1;
+            } else {
+                this.orders.push(createBatchedBridgeOrder(contiguousBridgeFills, opts));
+                i += contiguousBridgeFills.length;
+            }
+        }
+        return this as CollapsedPath;
+    }
+
+    public size(): PathSize {
```
dorothy-zbornak, line 138: y not `public get size(): PathSize`

packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -0,0 +1,287 @@
...
+    output: BigNumber;
+}
+
+export interface PathPenaltyOpts {
+    ethToOutputRate: BigNumber;
+    ethToInputRate: BigNumber;
+    exchangeProxyOverhead: ExchangeProxyOverhead;
+}
+
+export const DEFAULT_PATH_PENALTY_OPTS: PathPenaltyOpts = {
```
dorothy-zbornak, line 36: I honestly like exporting constants from the files where they're relevant but the unfortunate convention in this package right now is to put it in `constants.ts`.

packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -0,0 +1,287 @@
...
+    }
+}
+
+export interface CollapsedPath extends Path {
+    readonly collapsedFills: ReadonlyArray<CollapsedFill>;
+    readonly orders: OptimizedMarketOrder[];
+}
+
+const MULTIBRIDGE_SOURCES = SOURCE_FLAGS.LiquidityProvider | SOURCE_FLAGS.Uniswap;
+export function doSourcesConflict(flags: number): boolean {
```
dorothy-zbornak, line 284: Does this need to be exported?

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -234,6 +224,7 @@ export interface GetMarketOrdersRfqtOpts extends RfqtRequestOpts {
 
 export type FeeEstimate = (fillData?: FillData) => number | BigNumber;
 export type FeeSchedule = Partial<{ [key in ERC20BridgeSource]: FeeEstimate }>;
+export type ExchangeProxyOverhead = (sourceFlags: number) => BigNumber;
```
dorothy-zbornak, line 227: Is there a scenario where we would pass in a mix of flags?

packages/asset-swapper/src/utils/market_operation_utils/path.ts
```diff
@@ -0,0 +1,287 @@
...
+    exchangeProxyOverhead: ExchangeProxyOverhead;
+}
+
+export const DEFAULT_PATH_PENALTY_OPTS: PathPenaltyOpts = {
+    ethToOutputRate: ZERO_AMOUNT,
+    ethToInputRate: ZERO_AMOUNT,
+    exchangeProxyOverhead: () => ZERO_AMOUNT,
+};
+
+export class Path {
```
dorothy-zbornak, line 42: ya I like the class abstraction of a path

packages/asset-swapper/src/utils/market_operation_utils/types.ts
dorothy-zbornak: I guess we would with a mixed path?

dorothy-zbornak: Do you have an example of how we would implement this callback in production?

dorothy-zbornak (approved):
Refactor looks good, just have some open Qs, and obv we need to simbot these changes.

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
moodlezoup: hmm that's a good point... idk if that would be an issue any more after jacob's kyber rework though? We shouldn't hit the UniswapReserve anymore, right?

packages/asset-swapper/src/utils/market_operation_utils/fills.ts
moodlezoup: yeah not sure tbh –– I saw that `nativeOrdersToPath` was using `ethToInputRate` as a fallback but wasn't entirely sure why. I guess is token prelaunch is the only case, then the dex samples would be all zero anyway?

packages/asset-swapper/src/utils/market_operation_utils/path.ts
moodlezoup: bc i am noob

packages/asset-swapper/src/utils/market_operation_utils/path_optimizer.ts
moodlezoup: I doubt it; I think the bottleneck is probably copying the fills array, which we were doing before as well. but I guess we'll see in the simbot results

packages/asset-swapper/src/utils/market_operation_utils/types.ts
moodlezoup: Once we have the Uniswap/Sushiswap feature + the PLP feature, it'll be cheaper to go 100% to one of those sources, so we can use something like:
```
const exchangeProxyOverhead = (sourceFlags: number) =>
    [SOURCE_FLAGS.LiquidityProvider, SOURCE_FLAGS.UniswapV2, SOURCE_FLAGS.SushiSwap].includes(sourceFlags)
        ? new BigNumber(3e4).times(gasPrice)
        : new BigNumber(1.3e5).times(gasPrice);
```

moodlezoup: in the future where we can split a fill between the Unsiwap/Sushiswap/PLP and still leverage the VIP functionality, we'd use something like:
```
const vipSources = SOURCE_FLAGS.LiquidityProvider | SOURCE_FLAGS.UniswapV2 | SOURCE_FLAGS.SushiSwap;
const exchangeProxyOverhead = (sourceFlags: number) =>
    vipSources | sourceFlags === vipSources
        ? new BigNumber(3e4).times(gasPrice)
        : new BigNumber(1.3e5).times(gasPrice);

```

packages/asset-swapper/src/utils/market_operation_utils/path.ts
moodlezoup: nope, good call

packages/asset-swapper/src/utils/market_operation_utils/path.ts
moodlezoup: 😞

dorothy-zbornak:
Hmm, won't this throw off v0, mtx, and + affiliate fee paths because VIPs aren't available to either?

moodlezoup:
> Hmm, won't this throw off v0, mtx, and + affiliate fee paths because VIPs aren't available to either?

good point. v0 should be fine because we won't be passing in an `exchangeProxyOverhead` parameter. re: mtx and affiliate fees, I guess we'll just have to add some special-casing on the API side 🥴

PR #2695, opened 2020-09-03
[WIP] Assembly optimized proxy

recmo:
## Checklist:


contracts/zero-ex/contracts/src/ZeroEx.sol
```diff
@@ -48,29 +51,52 @@ contract ZeroEx {
 
     /// @dev Forwards calls to the appropriate implementation contract.
     fallback() external payable {
```
recmo, line 50: A plain `fallback() external payable` still creates some overhead in Solidity. We can avoid this if we use Yul directly.

PR #2702, opened 2020-09-10
asset-swapper: Quote report pass back optional fill data

kimpers:
## Description
By passing back the `fillData` from the DEX samples we can estimate the gas usage.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/quote_report_generator.ts
+++ b/packages/asset-swapper/src/utils/quote_report_generator.ts
@@ -8,6 +8,7 @@ import {
     CollapsedFill,
     DexSample,
     ERC20BridgeSource,
+    FillData,
     MultiHopFillData,
     NativeCollapsedFill,
 } from './market_operation_utils/types';
@@ -17,6 +18,7 @@ export interface BridgeReportSource {
     liquiditySource: Exclude<ERC20BridgeSource, ERC20BridgeSource.Native>;
     makerAmount: BigNumber;
     takerAmount: BigNumber;
+    fillData?: FillData;
 }
 
 export interface MultiHopReportSource {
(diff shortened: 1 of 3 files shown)


dekz (approved):

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -17,13 +18,15 @@ export interface BridgeReportSource {
     liquiditySource: Exclude<ERC20BridgeSource, ERC20BridgeSource.Native>;
     makerAmount: BigNumber;
     takerAmount: BigNumber;
+    fillData?: FillData;
```
steveklebanoff, line 21: in what cases is `fillData` undefined for `BridgeReportSource`?

packages/asset-swapper/src/utils/quote_report_generator.ts
```diff
@@ -17,13 +18,15 @@ export interface BridgeReportSource {
...
 }
 
 export interface MultiHopReportSource {
     liquiditySource: ERC20BridgeSource.MultiHop;
     makerAmount: BigNumber;
     takerAmount: BigNumber;
     hopSources: ERC20BridgeSource[];
+    fillData?: FillData;
```
steveklebanoff: in what cases is `fillData` undefined for `MultiHopReportSource `?

steveklebanoff (approved):
small question but LGTM

packages/asset-swapper/src/utils/quote_report_generator.ts
kimpers: I am not sure actually, the `DexSample` type has it as potentially undefined so I marked it like that as well. We only need the `fillData` for gas calculations with some of the integrations.

packages/asset-swapper/src/utils/quote_report_generator.ts
kimpers: This one is a copy-pasta error, it should always be there because it's used in the report code 🤦  Will fix.

steveklebanoff (approved):
lgtm!

PR #2703, opened 2020-09-11
UniswapFeature

dorothy-zbornak:
## Description

:warning:  **DO NOT MERGE until feature is deployed and migrated**

*Based on @Recmo's lovely work #2688*

Adds a feature called `UniswapFeature` with an MVP `sellToUniswap()` function with the following features (as discussed):
  - :white_check_mark: (uniswap) multi-hop.
  - :white_check_mark: WETH un/wrapping.
  - :white_check_mark: :sushi:swap
  - :x: accurate buys
  - :x: affiliate fees
  - :x: mtxs
  - :x: events

Simbot (gross) gas benchmarks:
| path | fn | gas |
|------|----|-----|
| `USDC->ETH` | `uniswap` | `118816` |
| `USDC->ETH` | `sellToUniswap()` | `116051` |
| `USDC->ETH` | `sellToUniswap(sushi=true)` | `116319` |
| `ETH->USDC` | `uniswap` | `130828` |
| `ETH->USDC` | `sellToUniswap()` | `126046` |
| `ETH->USDC` | `sellToUniswap(sushi=true)` | `126314` |
| `USDC->DAI` | `uniswap` | `109628` |
| `USDC->DAI` | `sellToUniswap()` | `108204` |
| `USDC->DAI` | `sellToUniswap(sushi=true)` | `NO_LIQUIDITY` |
| `USDC->ETH->DAI` | `uniswap` | `163124` |
| `USDC->ETH->DAI` | `sellToUniswap()` | `156715` |
| `USDC->ETH->DAI` | `sellToUniswap(sushi=true)` | `169539`  ??? |

### `includedSources`
There is now an `includedSources` option to AS that will *only* use the sources passed in, so we don't have to enumerate a massive `excludedSources` filter to target a few sources.

## Checklist:


contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,313 @@
...
+                            // pair contract using `AllowanceTarget.executeCall()`
+                            mstore(0xB00, ALLOWANCE_TARGET_EXECUTE_CALL_SELECTOR_32)
+                            mstore(0xB04, sellToken)
+                            mstore(0xB24, 0x40)
+                            mstore(0xB44, 0x64)
+                            mstore(0xB64, TRANSFER_FROM_CALL_SELECTOR_32)
+                            mstore(0xB68, caller())
+                            mstore(0xB88, pair)
+                            mstore(0xBA8, sellAmount)
+                            if iszero(call(gas(), mload(0xA60), 0, 0xB00, 0xE4, 0x00, 0x0)) {
```
moodlezoup:
```suggestion
                            if iszero(call(gas(), mload(0xA60), 0, 0xB00, 0xC8, 0x00, 0x0)) {
```
? or is my arithmetic wrong

moodlezoup (commented):
looking good, just one q on the assembly. super excited at the possibilities this will open up 😃

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
dorothy-zbornak: I think you're right.

recmo: I think the original (`0xE4`) is correct. ABI encoding is padded to 32-byte boundaries, excluding the first 4 bytes. This means the size must be `4 + n * 20`. The data ends at `0xC8` and the next valid size is `0xE4`.

But please doublecheck using a known-correct abi encoder.

dorothy-zbornak: Yeah, `0xE4` does mirror what `abi.encode()` would do by padding the data to a be word-aligned. But `0xC8` does also work just fine on simbot. I believe we use unaligned calldata in some other contracts, so the rule doesn't seem to be enforced.

packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts
```diff
@@ -35,6 +36,19 @@ import {
...
+ */
+export const TWO_HOP_SOURCE_FILTERS = SourceFilters.all().exclude([
+    ERC20BridgeSource.MultiHop,
+    ERC20BridgeSource.MultiBridge,
+    ERC20BridgeSource.Native,
+]);
+/**
+ * Source filters for `getSellQuotes()` and `getBuyQuotes()`.
+ */
+export const BATCH_SOURCE_FILTERS = SourceFilters.all().exclude([ERC20BridgeSource.MultiHop, ERC20BridgeSource.Native]);
```
dorothy-zbornak, line 50: So now the operations will do some extra filtering of sources passed in so the consumer doesn't have to worry about it.

packages/asset-swapper/test/market_operation_utils_test.ts
```diff
@@ -18,10 +18,22 @@ import * as TypeMoq from 'typemoq';
 import { MarketOperation, QuoteRequestor, RfqtRequestOpts, SignedOrderWithFillableAmounts } from '../src';
```
dorothy-zbornak, line 18: I had to pop an ativan to avoid a mental breakdown when modifying these tests.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+                            }
+                        }
+                    // No need to check results, if deposit/transfers failed the UniswapV2Pair will
+                    // reject our trade (or it may succeed if somehow the reserve was out of sync)
+                    // this is fine for the taker.
+                }
+
+                // Call pair.getReserves(), store the results at `0xC00`
+                mstore(0xB00, UNISWAP_PAIR_RESERVES_CALL_SELECTOR_32)
+                if iszero(call(gas(), pair, 0, 0xB00, 0x4, 0xC00, 0x40)) {
```
smarx: Perhaps this should be a `staticcall`.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+                let pairSellAmount := buyAmount
+                // Compute the buy amount based on the pair reserves.
+                {
+                    let sellReserve := mload(0xC00)
+                    let buyReserve := mload(0xC20)
+                    // Transpose if pair order is different.
+                    if iszero(pairOrder) {
+                        sellReserve := mload(0xC20)
+                        buyReserve := mload(0xC00)
+                    }
```
smarx: This seems a little wasteful in that it does extra `mload`s in the transposed case instead of just one. Why not `switch pairOrder`?

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+            // If buying ETH, unwrap the WETH first
+            if eq(buyToken, ETH_TOKEN_ADDRESS_32) {
+                // Call `WETH.withdraw(buyAmount)`
+                mstore(0xB00, WETH_WITHDRAW_CALL_SELECTOR_32)
+                mstore(0xB04, buyAmount)
+                if iszero(call(gas(), mload(0xA40), 0, 0xB00, 0x24, 0x00, 0x0)) {
+                    bubbleRevert()
+                }
+                // Transfer ETH to the caller.
+                if iszero(call(2300, caller(), buyAmount, 0xB00, 0x0, 0x00, 0x0)) {
```
smarx: I'm [somewhat famously](https://diligence.consensys.net/blog/2019/09/stop-using-soliditys-transfer-now/) against hardcoding a gas stipend like this. I'd prefer to forward all gas.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+
+            function normalizeToken(token) -> normalized {
+                normalized := token
+                // Translate ETH pseudo-tokens to WETH.
+                if eq(token, ETH_TOKEN_ADDRESS_32) {
+                    normalized := mload(0xA40)
+                }
+            }
+
+            function computePairAddress(tokenA, tokenB) -> pair {
```
smarx, line 309: Could we get some better comments here? (Let me know if you want me to write some.)

In particular, writing to 0xB14 followed by 0xB00 and then hashing starting at 0xB0C won't be obvious to many people. I did some math and drew a diagram to make sure I understood what was where in memory. It would also be nice to explain the rest of the `CREATE2` address derivation, or perhaps just link to where Uniswap does/documents it.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+    uint256 constant private ADDRESS_MASK = 0x000000000000000000000000ffffffffffffffffffffffffffffffffffffffff;
+    uint256 constant private ETH_TOKEN_ADDRESS_32 = 0x000000000000000000000000eeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee;
+
+    uint256 constant private ALLOWANCE_TARGET_EXECUTE_CALL_SELECTOR_32 = 0xbca8c7b500000000000000000000000000000000000000000000000000000000;
+    uint256 constant private UNISWAP_PAIR_RESERVES_CALL_SELECTOR_32 = 0x0902f1ac00000000000000000000000000000000000000000000000000000000;
+    uint256 constant private UNISWAP_PAIR_SWAP_CALL_SELECTOR_32 = 0x022c0d9f00000000000000000000000000000000000000000000000000000000;
+    uint256 constant private TRANSFER_FROM_CALL_SELECTOR_32 = 0x23b872dd00000000000000000000000000000000000000000000000000000000;
+    uint256 constant private WETH_WITHDRAW_CALL_SELECTOR_32 = 0x2e1a7d4d00000000000000000000000000000000000000000000000000000000;
+    uint256 constant private WETH_DEPOSIT_CALL_SELECTOR_32 = 0xd0e30db000000000000000000000000000000000000000000000000000000000;
+    uint256 constant private ERC20_TRANSFER_CALL_SELECTOR_32 = 0xa9059cbb00000000000000000000000000000000000000000000000000000000;
```
smarx: It would be nice to have comments here giving the function signatures so it's easy for someone to check that these are the right selectors. The security of this code does rely on us actually calling the right `swap()` function.

Bonus points for actually asserting all the constants are correct in the constructor. I.e. something like `assert(ERC20_TRANSFER_CALL_SELECTOR_32 == keccak256(...) & 0xFFFFFFFF000000000...)`

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+
+            function bubbleRevert() {
+                returndatacopy(0, 0, returndatasize())
+                revert(0, returndatasize())
+            }
+        }
+
+        // Revert if we bought too little.
+        // TODO: replace with rich revert?
+        require(buyAmount >= minBuyAmount, "UniswapFeature/UnderBought");
```
smarx, line 364: If there are bugs, the check here against `minBuyAmount` can give people confidence that the outcome will be correct. That said, there's no explicit guarantee in this code that the requested amount was actually delivered to the sender. We're relying on Uniswap's implementation (which _does_ have a fairly explicit guarantee). I think this is okay, but it feels like the biggest risk. I haven't been able to think of a better check.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+        return LibMigrate.MIGRATE_SUCCESS;
+    }
+
+    /// @dev Efficiently sell directly to uniswap/sushiswap.
+    /// @param tokens Sell path.
+    /// @param sellAmount of `tokens[0]` Amount to sell.
+    /// @param minBuyAmount Minimum amount of `tokens[-1]` to buy.
+    /// @param isSushi Use sushiswap if true.
+    /// @return buyAmount Amount of `tokens[-1]` bought.
+    function sellToUniswap(
```
smarx, line 101: Uniswap's router includes a check to prevent swaps from being executed beyond a provided deadline. I don't think such a check is necessary, but I want to call it out and make sure it was an explicit decision to skip it.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
```diff
@@ -0,0 +1,320 @@
...
+        uint256 sellAmount,
+        uint256 minBuyAmount,
+        bool isSushi
+    )
+        external
+        payable
+        override
+        returns (uint256 buyAmount)
+    {
+        require(tokens.length > 1, "UniswapFeature/InvalidTokensLength");
```
smarx, line 112: Do we need to check that amounts don't exceed 2**112 as in https://github.com/0xProject/0x-monorepo/blob/feat/efficient-uniswap/contracts/zero-ex/contracts/src/features/DirectUniswap.sol#L43?

smarx (commented):
Looking good! I've added some inline comments/questions.

contracts/zero-ex/contracts/src/features/UniswapFeature.sol
dorothy-zbornak: Ah yeah, nice catch.

contracts/zero-ex/contracts/src/fixins/FixinCommon.sol
```diff
@@ -81,6 +81,6 @@ abstract contract FixinCommon {
         pure
         returns (uint256 encodedVersion)
     {
-        return (major << 64) | (minor << 32) | revision;
+        return (uint256(major) << 64) | (uint256(minor) << 32) | uint256(revision);
```
dorothy-zbornak, line 84: lol, noticed all our deployed feature contracts were reporting `FEATURE_VERSION = 0` :innocent:

smarx (approved):

PR #2706, opened 2020-09-15
asset-swapper: Punish latent RFQT makers

feuGeneA:
## Checklist:


packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Tracks a maker's history of timely responses, and manages whether a given
+ * maker should be avoided for being too latent.
+ */
+export class RfqMakerBlacklist {
+    // tslint:disable-next-line:custom-no-magic-numbers
+    private static readonly _msPerMinute = 1000 * 60;
+    private readonly _makerTimeoutStreakLength: { [makerUrl: string]: number } = {};
+    private readonly _makerBlacklistedUntilDate: { [makerUrl: string]: number } = {};
+    constructor(private readonly _blacklistDurationMinutes: number, private readonly _timeoutStreakThreshold: number) {}
```
PirosB3, line 11: why do constructor args have underscores? I know you are taking a shortcut here which I took in the past, but I was suggested to simply set up constructor the old fashioned way

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Tracks a maker's history of timely responses, and manages whether a given
+ * maker should be avoided for being too latent.
+ */
+export class RfqMakerBlacklist {
+    // tslint:disable-next-line:custom-no-magic-numbers
+    private static readonly _msPerMinute = 1000 * 60;
```
PirosB3: Can you replace `1000` with `ONE_SECOND`?

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
```diff
@@ -0,0 +1,38 @@
...
+                    Date.now() + this._blacklistDurationMinutes * RfqMakerBlacklist._msPerMinute;
+                this._makerTimeoutStreakLength[makerUrl] = 0;
+            }
+        } else {
+            this._makerTimeoutStreakLength[makerUrl] = 0;
+        }
+    }
+    public isMakerBlacklisted(makerUrl: string): boolean {
+        if (this._makerBlacklistedUntilDate.hasOwnProperty(makerUrl)) {
+            if (this._makerBlacklistedUntilDate[makerUrl] > Date.now()) {
```
PirosB3: what is the difference between `new Date().getTime()` and `Date.now()`? we usually use the former afaik

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
```diff
@@ -0,0 +1,38 @@
...
+    private static readonly _msPerMinute = 1000 * 60;
+    private readonly _makerTimeoutStreakLength: { [makerUrl: string]: number } = {};
+    private readonly _makerBlacklistedUntilDate: { [makerUrl: string]: number } = {};
+    constructor(private readonly _blacklistDurationMinutes: number, private readonly _timeoutStreakThreshold: number) {}
+    public logTimeoutOrLackThereof(makerUrl: string, didTimeout: boolean): void {
+        if (!this._makerTimeoutStreakLength.hasOwnProperty(makerUrl)) {
+            this._makerTimeoutStreakLength[makerUrl] = 0;
+        }
+        if (didTimeout) {
+            this._makerTimeoutStreakLength[makerUrl] += 1;
```
PirosB3, line 17: `this._makerTimeoutStreakLength[makerUrl] = (this._makerTimeoutStreakLength[makerUrl] || 0) + 1` could allow you to omit L12-14

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
PirosB3: Should this live in this class? or could it be in the same file of `ONE_SECOND` or another const file?

PirosB3 (changes requested):
Can we write some tests for the new class?

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
feuGeneA: the underscores are there because the linter required me to put them there, because they're not just function arguments, they're also declarations of member variables.

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
feuGeneA: No difference in effect, really.  I prefer `Date.now()`, partly because in theory it's less overhead (no constructing a `Date` object), but more importantly because it's easy to tell at a glance what the intention is, without needing to remember that `new Date()` constructs an object set to `"now"`.

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
feuGeneA: I loved the suggestion at first.  But then, when I wrote it up, I decided that I prefer to keep the simple `...Streak... += 1;`, since that's such a key part of the mechanism here, and I'd like to keep it as simple and straightforward as possible.

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2706/commits/79e0a9ef3701f8bdb2e28dd4a6b83bce88019472

packages/asset-swapper/src/utils/rfq_maker_blacklist.ts
PirosB3: right, could we use the old fashioned way instead? or do u think it's not worth it? underscore arguments usually are an alias for unused arguments

PirosB3 (approved):

PR #2712, opened 2020-09-23
AS: respect max slippage in EP consumer

dorothy-zbornak:
## Description

In the EP consumer we were setting `minBuyAmount = quote.worstCaseQuoteInfo.makerAssetAmount`, which would allow for very high slippage because the quote simulator (who populates this value) finds the actual worst case, irregardless of slippage limits.

:heavy_check_mark: SWRV/WETH pairs that slip >1% now revert as incomplete fills in simbot.

## Checklist:


moodlezoup (approved):
god I hate asset swapper 🤷  🥴

PR #2713, opened 2020-09-24
[Price-aware RFQ requests] Decouple QuoteReporter from _generateOptimizedOrdersAsync

PirosB3:
## Description

Decouples the quote reporter logic from generating an optimized order. This PR is simply a refactor in lieu of greater changes that will happen in separate PRs. There should be no impact on the end user.

## Types of changes

* Refactor

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/index.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/index.ts
@@ -7,7 +7,7 @@ import * as _ from 'lodash';
 import { MarketOperation } from '../../types';
 import { QuoteRequestor } from '../quote_requestor';
 
-import { generateQuoteReport } from './../quote_report_generator';
+import { generateQuoteReport, QuoteReport } from './../quote_report_generator';
 import {
     BUY_SOURCE_FILTER,
     DEFAULT_GET_MARKET_ORDERS_OPTS,
@@ -40,6 +40,7 @@ import {
     OptimizerResult,
     OrderDomain,
     TokenAdjacencyGraph,
+    OptimizerResultWithReport,
 } from './types';
 
 // tslint:disable:boolean-naming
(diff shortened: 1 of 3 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -339,19 +360,29 @@ export class MarketOperationUtils {
...
-            bridgeSlippage: _opts.bridgeSlippage,
-            maxFallbackSlippage: _opts.maxFallbackSlippage,
-            excludedSources: _opts.excludedSources,
-            feeSchedule: _opts.feeSchedule,
-            allowFallback: _opts.allowFallback,
-            shouldBatchBridgeOrders: _opts.shouldBatchBridgeOrders,
-            quoteRequestor: _opts.rfqt ? _opts.rfqt.quoteRequestor : undefined,
-            shouldGenerateQuoteReport: _opts.shouldGenerateQuoteReport,
+    ): Promise<OptimizerResultWithReport> {
+        const defaultOpts = { ...DEFAULT_GET_MARKET_ORDERS_OPTS, ...opts };
```
steveklebanoff: Curious - why rename this `defaultOpts`?  I don't think it's an accurate name because this really is the `default options` merged with the `custom options`.  Would prefer to leave as `_opts` as that seems to be the pattern in this file

steveklebanoff (commented):
One nit but overall lgtm, but it looks like CI is failing because static tests are failing (need to lint), and you also will need to update `market_operation_utils_test.ts`

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: Good point! I'm trying to keep the `_` prefix as private because that's what we have usually used it for. I know that I may have broken a convention here. Shall I roll-back to `_opts`?

PirosB3:
@steveklebanoff which tests are you referring to? `MarketOperationUtils tests` are all passing for me

steveklebanoff (approved):

PR #2715, opened 2020-09-29
feat: Add CREAM as a liquidity source

alexkroeger:
## Description

Add CREAM to asset-swapper.

CREAM (*C*rypto *R*ules *E*verything *A*round *M*e) is a Balancer clone.

## Checklist:


contracts/zero-ex/contracts/src/transformers/bridges/BridgeAdapter.sol
```diff
@@ -170,6 +172,12 @@ contract BridgeAdapter is
                 sellAmount,
                 bridgeData
             );
+        } else if (bridgeAddress == CREAM_BRIDGE_ADDRESS) {
```
dekz: We should join this with the other Balancer code block for now.

```
else if (bridgeAddress == BALANCER_BRIDGE_ADDRESS || bridgeAddress == CREAM_BRIDGE_ADDRESS)

File: packages/asset-swapper/src/utils/market_operation_utils/sampler_operations.ts [hunk 2]
```diff
@@ -1040,6 +1105,12 @@ export class SamplerOperations {
                                 .map(poolAddress =>
                                     this.getBalancerSellQuotes(poolAddress, makerToken, takerToken, takerFillAmounts),
                                 );
+                        case ERC20BridgeSource.Cream:
+                            return this.creamPoolsCache
+                                .getCachedPoolAddressesForPair(takerToken, makerToken)!
+                                .map(poolAddress =>
+                                    this.getCreamSellQuotes(poolAddress, makerToken, takerToken, takerFillAmounts),
```
dekz: nit: maybe we could just use `getBalancerSellQuotes` or nah? Don't know if we need `CreamFillData` when it's the same as `BalancerFillData`

dekz (approved):
Do we need an entry in `contract-addresses` for the `creamBridge` to keep it stored somewhere?

Probably needed for the `migrations` as well.

RLlane2126:
Get Outlook for Android<https://aka.ms/ghei36>
________________________________

packages/asset-swapper/src/utils/market_operation_utils/cream_utils.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// tslint:disable:custom-no-magic-numbers
+const FIVE_SECONDS_MS = 5 * 1000;
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+const DEFAULT_TIMEOUT_MS = 1000;
+const MAX_POOLS_FETCHED = 3;
+// tslint:enable:custom-no-magic-numbers
+
+export class CreamPoolsCache {
```
dorothy-zbornak, line 19: Is there really any meaningful difference between this class and `BalancerPoolsCache`? Does it make sense to just modify and inherit from that class, passing in `getPoolsWithTokens()` and `parsePoolData()` callbacks to the super?

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -150,6 +150,23 @@ export class MarketOperationUtils {
...
+            quoteSourceFilters.isAllowed(ERC20BridgeSource.Cream),
+        );
+
+        let excludedSources: ERC20BridgeSource[] = [];
+        if (!sampleCreamOnChain) {
+            excludedSources = excludedSources.concat(ERC20BridgeSource.Cream);
+        }
+        if (!sampleBalancerOnChain) {
+            excludedSources = excludedSources.concat(ERC20BridgeSource.Balancer);
+        }
```
dorothy-zbornak: idk, what do you think of?
```suggestion
        excludedSources = [
            ...(!sampleCreamOnChain ? [ERC20BridgeSource.Cream] : []),
            ...(!sampleBalancerOnChain ? [ERC20BalancerSource.Balancer] : []),
        ];
```

dorothy-zbornak: If not, we should change the `concat()` to `push()`.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -266,6 +294,23 @@ export class MarketOperationUtils {
...
```
dorothy-zbornak: samesies

packages/asset-swapper/src/utils/market_operation_utils/types.ts
```diff
@@ -104,6 +105,10 @@ export interface BalancerFillData extends FillData {
     poolAddress: string;
 }
 
+export interface CreamFillData extends FillData {
```
dorothy-zbornak: Should we  just extend `BalancerFillData`?

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
```diff
@@ -177,6 +178,8 @@ function getBridgeAddressFromFill(fill: CollapsedFill, opts: CreateOrderFromPath
             return opts.contractAddresses.bancorBridge;
         case ERC20BridgeSource.Balancer:
             return opts.contractAddresses.balancerBridge;
+        case ERC20BridgeSource.Cream:
+            return opts.contractAddresses.balancerBridge;
```
dorothy-zbornak: We deploying the bridge in this PR?

contracts/asset-proxy/contracts/src/bridges/CreamBridge.sol
```diff
@@ -0,0 +1,103 @@
...
+
+import "@0x/contracts-erc20/contracts/src/interfaces/IERC20Token.sol";
+import "@0x/contracts-erc20/contracts/src/LibERC20Token.sol";
+import "@0x/contracts-exchange-libs/contracts/src/IWallet.sol";
+import "@0x/contracts-utils/contracts/src/DeploymentConstants.sol";
+import "../interfaces/IERC20Bridge.sol";
+import "../interfaces/IBalancerPool.sol";
+
+
+contract CreamBridge is
```
dorothy-zbornak, line 31: This is exactly the same as `BalancerBridge`, right? If so, can we just do

```solidity
// solhint-disable no-empty-blocks
contract CreamBridge is BalancerBridge {}
``` 

I guess the `require()` error will say "Balancer" but I don't think that'll matter much in reality.

dorothy-zbornak (approved):
Small nits, but looking gr8!

packages/asset-swapper/src/utils/market_operation_utils/index.ts
alexkroeger: Great suggestion!

contracts/asset-proxy/contracts/src/bridges/CreamBridge.sol
alexkroeger: I deployed as-is to keep the require() error

packages/asset-swapper/src/utils/market_operation_utils/orders.ts
alexkroeger: yes, deployed

packages/asset-swapper/src/utils/market_operation_utils/types.ts
alexkroeger: got rid of CreamFillData altogether

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -150,6 +150,20 @@ export class MarketOperationUtils {
...
+        const {
+            onChain: sampleCreamOnChain,
+            offChain: sampleCreamOffChain,
+        } = this._sampler.creamPoolsCache.howToSampleCream(
+            takerToken,
+            makerToken,
+            quoteSourceFilters.isAllowed(ERC20BridgeSource.Cream),
+        );
+
+        const excludedSources = [
```
dekz: This name is a bit confusing as there is an `opts.excludedSources`.

It's probable duplicated in this file twice as well possibly? Could use a helper

dekz (approved):
Think this might need a rebase ontop of latest development, then a re-publish once on `development` for 0xAPI. So that the `sha` matches the commit.

Once this is merged github basically deletes the `sha`  making it hard to find which AS release was which commit, unless published off development.

So rebase, merge, publish:private

packages/asset-swapper/src/utils/market_operation_utils/index.ts
alexkroeger: Fixed naming, I think it's straightforward enough to leave the code the same for now.

PR #2716, opened 2020-10-01
fix: fixed a small bug with buys and quote reporter

PirosB3:
## Description

Fixes a small bug: all buys in `MarketOperationUtils` will only run quote reporter if `rfqt` is enabled. This bug was patched for sells, but I forgot to patch it for buys.

## Checklist:


steveklebanoff (approved):
pending CI

PR #2718, opened 2020-10-01
[WIP] Reduce gas consumption for ERC20 transfers

smarx:
## Description

This rewrites `AllowanceTarget.exceuteCall()` and `TokenSpenderFeature._spendERC20Tokens()` in assembly to save 2798 gas per `_spendERC20Tokens()` call.

## Testing instructions

No test changes are necessary. `yarn test` tests the new, rewritten versions.

## Types of changes

Specifically, this is just gas optimization. No change in behavior is expected.

## Checklist:

### DIFF
--- a/contracts/zero-ex/contracts/src/external/AllowanceTarget.sol
+++ b/contracts/zero-ex/contracts/src/external/AllowanceTarget.sol
@@ -46,10 +46,23 @@ contract AllowanceTarget is
         onlyAuthorized
         returns (bytes memory resultData)
     {
-        bool success;
-        (success, resultData) = target.call(callData);
-        if (!success) {
-            resultData.rrevert();
+        // Gas savings from assembly: 474
+        assembly {
+            let where := add(calldataload(0x24), 4)
+            let cdlen := calldataload(where)
+            calldatacopy(0, add(where, 0x20), cdlen)
+            let success := call(gas(), target, 0, 0, cdlen, 0, 0)
+
+            if iszero(success) {
+                returndatacopy(0, 0, returndatasize())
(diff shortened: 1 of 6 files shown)


contracts/zero-ex/contracts/src/external/AllowanceTarget.sol
```diff
@@ -46,10 +46,23 @@ contract AllowanceTarget is
         onlyAuthorized
         returns (bytes memory resultData)
```
hysz: I reckon we could shave off some more gas by putting in the fallback and doing the `onlyAuthorized` check in assembly. Here's [sample code](https://github.com/0xProject/0x-monorepo/blob/development/contracts/asset-proxy/contracts/src/ERC20Proxy.sol#L52) from the `ERC20Proxy`.

contracts/zero-ex/contracts/src/external/AllowanceTarget.sol
```diff
@@ -46,10 +46,23 @@ contract AllowanceTarget is
...
     {
-        bool success;
```
hysz, line 49: We can also remove the dependency on `LibRichErrorsV06`.

contracts/zero-ex/contracts/src/external/AllowanceTarget.sol
smarx: I will give that a try. Thanks.

smarx: This is done in the latest commit. It saved an additional 256 gas. Thanks!

PR #2719, opened 2020-10-02
[WIP] Feature/subproviders/lattice support

alex-miller-0:
## Description

Hello! This PR adds support for interfacing with the GridPlus [Lattice1 hardware wallet](https://gridplus.io/lattice). It was developed along with [GridPlus' `web3-react` fork](https://github.com/GridPlus/web3-react/tree/add-lattice-example) in an effort to allow integrations with the Lattice. The addition is a simple subprovider schema that can be used by other modules such as `web3-react`. It was designed to match those of Trezor and Ledger.

### About the Lattice

The [Lattice](https://gridplus.io/lattice) is a brand new, next-generation hardware wallet designed for ease of use. Its large screen makes reading transactions and messages much more user-friendly. Keys are held in an HSM and all logic executes in a secure compute environment, which is physically separated from the network interface and protected by a physical anti-tamper mesh. The separated network interface is used to connect to external applications like this one. 

The initial batch of Lattice devices is currently being manufactured and sent to presale buyers, at which point we will turn the web store back on for more orders.

For more info see https://gridplus.io/lattice

### Product Shots

Here are some shots of my development device! (Data shown is not related to this PR)

**Sign Transaction:**

image

**Sign Personal Message:**

image

## Testing instructions

You can't test this PR without a Lattice device, but I will provide information on how I tested it on my side. If you need to test this yourself, please reach out to justin.leroux@gridplus.io and we can work out how to get you a Lattice for testing.

I was able to test with `web3-react`'s `example` Next.js app which is included in [the repo](https://github.com/NoahZinsmeister/web3-react/tree/v6/example). All changes to the app itself are found in [this commit](https://github.com/NoahZinsmeister/web3-react/commit/ffabb34a46fbe375bdf74f60ec730becbef5d7fb) of our fork. This is a simple app where you just select your wallet and then test a `signPersonalMessage` request. Here are shots of my test:

**1. Message is sent to the device from the app:**

image

**2. Message is signed on device and returned to app, which checks it against the available public key for validity:**

Screen Shot 2020-10-02 at 9 14 38 AM

## Types of changes

## Checklist:

-   [x] Add tests to cover changes as needed. **-- I can't think of how to provide tests here, as this module must be used with others in order to interact with the Lattice**
-   [ ] Update documentation as needed. **-- Need more clarity on this. I didn't want to make manual changes to `docs/reference.mdx` since it looks like that gets generated automatically. Not sure where else to provide docs.**

### DIFF
--- a/packages/subproviders/CHANGELOG.json
+++ b/packages/subproviders/CHANGELOG.json
@@ -1,5 +1,14 @@
 [
     {
+        "timestamp": 1601647297,
+        "version": "6.2.0",
+        "changes": [
+            {
+                "note": "Lattice support added"
+            }
+        ]
+    },
+    {
         "timestamp": 1594788383,
         "version": "6.1.1",
         "changes": [
--- a/packages/subproviders/CHANGELOG.json
+++ b/packages/subproviders/CHANGELOG.json
@@ -1,6 +1,5 @@
 [
     {
-        "timestamp": 1601647297,
         "version": "6.2.0",
         "changes": [
             {
--- a/packages/subproviders/src/subproviders/lattice.ts
+++ b/packages/subproviders/src/subproviders/lattice.ts
@@ -115,7 +115,7 @@ export class LatticeSubprovider extends BaseWalletSubprovider {
         try {
             const data = {
                 protocol: 'eth_signTypedData',
-                data: typedData,
+                payload: typedData,
             };
             const sig = await this._latticeConnectClient.signMessage(address, data);
             return sig;
--- a/packages/subproviders/src/subproviders/lattice.ts
+++ b/packages/subproviders/src/subproviders/lattice.ts
@@ -102,7 +102,6 @@ export class LatticeSubprovider extends BaseWalletSubprovider {
 
     /**
      * Sign a typed data message from the account specified in the `address` param.
-     * NOTE: This method is not implemented on the Lattice in v1, but we plan to support it in future versions.
      * @param address Address from which to sign. Must be the address at `m/44'/60'/0'/0/0` of the current wallet.
      * @param typedData The data to be signed.
      * @return Signature hex string of form `0x{r}{s}{v}
--- a/packages/subproviders/src/subproviders/lattice.ts
+++ b/packages/subproviders/src/subproviders/lattice.ts
@@ -1,0 +1,126 @@
+import { assert } from '@0x/assert';
+import { addressUtils } from '@0x/utils';
+import EthereumTx = require('ethereumjs-tx');
+import * as _ from 'lodash';
+
+import { LatticeSubproviderConfig, PartialTxParams, WalletSubproviderErrors } from '../types';
+
+import { BaseWalletSubprovider } from './base_wallet_subprovider';
+
+const DEFAULT_NUM_ADDRESSES_TO_FETCH = 1;
+const MAINNET_ID = 1;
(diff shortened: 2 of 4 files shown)


alex-miller-0:
I see this build is failing. I tried to setup the circle build [per instructions](https://github.com/0xProject/0x-monorepo/blob/development/CONTRIBUTING.md#fix-submit-coverage-ci-failure), but I wasn't sure how to add that `COVERALLS_REPO_TOKEN` param. I didn't want to commit my repo token here and I have never used circle before (and do not have a paid account that can run the build on a `medium+` resource), so I'm not sure what to do. Would appreciate guidance here as well.

packages/subproviders/CHANGELOG.json
```diff
@@ -1,4 +1,13 @@
 [
+    {
+        "timestamp": 1601647297,
```
dekz:
```suggestion
```

This is automatic on a publish, so need to remove otherwise our infra thinks it has been published :)

packages/subproviders/src/subproviders/lattice.ts
```diff
@@ -0,0 +1,126 @@
...
+            const sig = await this._latticeConnectClient.signPersonalMessage(address, data);
+            return sig;
+        } catch (err) {
+            throw err;
+        }
+    }
+
+    /**
+     * Sign a typed data message from the account specified in the `address` param.
+     * NOTE: This method is not implemented on the Lattice in v1, but we plan to support it in future versions.
```
dekz: Might be best to throw an error whilst it's not implemented? Otherwise the dev may have a hard time tracking down a bug.

If it does throw, some tools will default to `eth_sign` of the hash of the data.

alex-miller-0: So there is an error that gets thrown by the underlying SDK ([here](https://github.com/GridPlus/gridplus-sdk/blob/master/src/ethereum.js#L53)). That error will get sent up to the `_latticeConnectClient` ([here](https://github.com/GridPlus/eth-lattice-keyring/blob/master/index.js#L131) and [here](https://github.com/GridPlus/eth-lattice-keyring/blob/master/index.js#L152)) and will eventually make it to this module (and get thrown on [L123](https://github.com/0xProject/0x-monorepo/pull/2719/files#diff-1dad31a0058965df065352f990f8aa77R123)). We will be adding support in the next few months via the other two modules I linked (which we maintain). At that point, the `signMessage` functionality will be extensible to other signing schemas, should future protocols get ossified (since it will switch on that `protocol` value). 

Happy to remove it and come back, but I figured this would avoid having to go through two sequential PRs. But no problem either way!

packages/subproviders/CHANGELOG.json
alex-miller-0: Haha yes this felt strange to me, but I was trying to work through the contribution guidelines.

Reverted in ~b58b3a6~ cc55c2e (I think you're just saying to remove `timestamp` but keep the rest of the entry -- I can also remove the entire entry).

(By the way, I can squash these commits if/when the PR is finalized/approved.)

packages/subproviders/src/subproviders/lattice.ts
alex-miller-0: @dekz went ahead and removed the comment in ea71cfc

We are actually working on an implementation now. It may be available by the time we get an integration out that uses this module.

PR #2720, opened 2020-10-02
feat: Refactor rfq comparison price integration

PirosB3:
## Description

Calculates a "price to beat" that can be funneled to RFQ makers. For now, this is naively the price of the overall order. Adds an extra `comparisonPrice` parameter to the RFQ GET request.

## Testing instructions

- Added an end-to-end unit test
- Will perform testing locally

## Checklist:


PirosB3:
NOTE: this PR will never build until https://github.com/0xProject/quote-server/pull/8 is merged and released

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -57,6 +59,7 @@ export async function getRfqtIndicativeQuotesAsync(
     takerAssetData: string,
     marketOperation: MarketOperation,
     assetFillAmount: BigNumber,
+    comparisonPrice: BigNumber | undefined,
     opts: Partial<GetMarketOrdersOpts>,
 ): Promise<RFQTIndicativeQuote[]> {
     if (opts.rfqt && opts.rfqt.isIndicative === true && opts.rfqt.quoteRequestor) {
```
feuGeneA, line 67: I'm wondering whether this new `comparisonPrice` shouldn't go in the `opts.rfqt` object here rather than getting its own parameter...IF it's only ever intended to be used for RFQT.  If it became part of that object, I believe it would eliminate a lot of the parameter additions you had to do in `quote_requestor`.  Also it would allow you to avoid clients needing to pass in `undefined` for this parameter if they aren't actually using it (like you had to do in your tests), as you could mark the property `?` (optional) in the type definition in the RFQ options object.

packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -460,7 +472,7 @@ export class MarketOperationUtils {
     }
 
     public async _generateOptimizedOrdersAsync(
-        marketSideLiquidity: MarketSideLiquidity,
+        marketSideLiquidity: Omit<MarketSideLiquidity, 'makerTokenDecimals' | 'takerTokenDecimals'>,
```
feuGeneA, line 549: Why was it necessary to explicitly omit these properties rather than just letting them come along for the ride?

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
```diff
@@ -34,7 +36,7 @@ export function getIntermediateTokens(
  * Returns the best two-hop quote and the fee-adjusted rate of that quote.
  */
 export function getBestTwoHopQuote(
-    marketSideLiquidity: MarketSideLiquidity,
+    marketSideLiquidity: Omit<MarketSideLiquidity, 'makerTokenDecimals' | 'takerTokenDecimals'>,
```
feuGeneA, line 46: Same as another comment, wondering why these needed to be omitted here.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -1,6 +1,7 @@
 import { schemas, SchemaValidator } from '@0x/json-schemas';
 import { assetDataUtils, orderCalculationUtils, SignedOrder } from '@0x/order-utils';
 import { RFQTFirmQuote, RFQTIndicativeQuote, TakerRequest } from '@0x/quote-server';
+import { TakerRequestQueryParams } from '@0x/quote-server/lib/src/types';
```
feuGeneA: I think you want to export this from the `index.ts` file in `quote-server` (see my comment on that PR), and then here drop the `lib/src/types` suffix... no?  I may be wrong, but my understanding is that referring to package sub-paths is an anti-pattern.

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -113,6 +114,39 @@ export class QuoteRequestor {
     private readonly _schemaValidator: SchemaValidator = new SchemaValidator();
     private readonly _orderSignatureToMakerUri: { [orderSignature: string]: string } = {};
 
+    public static makeQueryParameters(
+        takerAddress: string,
+        marketOperation: MarketOperation,
+        makerAssetData: string,
+        takerAssetData: string,
+        assetFillAmount: BigNumber,
+        comparisonPrice?: BigNumber | undefined,
```
feuGeneA: doesn't using `?` imply `|undefined`? (I think you can drop the `|undefined`.)

packages/asset-swapper/src/utils/quote_requestor.ts
```diff
@@ -113,6 +114,39 @@ export class QuoteRequestor {
     private readonly _schemaValidator: SchemaValidator = new SchemaValidator();
     private readonly _orderSignatureToMakerUri: { [orderSignature: string]: string } = {};
 
+    public static makeQueryParameters(
+        takerAddress: string,
+        marketOperation: MarketOperation,
+        makerAssetData: string,
```
feuGeneA, line 94: Consider merging `makeQueryParameters()` with the existing `inferQueryParameters()` function (using whatever name feels appropriate to the unified concept).  Seems like it would be confusing to come into this code and see both of those functions and try to understand the difference between the two.

feuGeneA (changes requested):
I'll need to defer to someone else on the Solidity, and also on the suggested price calculation.

Bravo on those typemoq tests! They look like they were a beast to pin down!

packages/asset-swapper/src/utils/quote_requestor.ts
feuGeneA: Addressed in https://github.com/0xProject/0x-monorepo/pull/2720/commits/2cf31f05a1b402f5e8f7853c60dc760baa0c6b3b

packages/asset-swapper/src/utils/market_operation_utils/index.ts
PirosB3: So you are suggesting adding this parameter as part of `GetMarketOrdersOpts`?

packages/asset-swapper/src/utils/market_operation_utils/multihop_utils.ts
PirosB3: I think this was done for testing purposes. In general, bunny-hop does not care about decimals because it simply works with selling + buying base unit amount. Whatever it gets back, it then sells.

packages/asset-swapper/src/utils/quote_requestor.ts
PirosB3: great suggestion! I refactored

feuGeneA (approved):

PR #2730, opened 2020-10-15
fix: [asset-swapper] MultiHop edge cases

dekz:
## Description

Using `includedSources/excludedSources` can introduce a few edge cases.
* It modified feeSources which made it difficult to sometimes find an maker/ETH or taker/ETH price, so cost based routing could get weird.
* It can result in MultiHop not having anything to MultiHop to (e.g DAI/USDC with Curve only, theres no ability to ETH hop)
* There are cases where `firstHopSource` is not present in the `FillData` but it is required. No valid sources or reverts (didn't dig deep enough to confirm it was a revert)
* If there was no optimal path then we would throw, there's the possibility that there is no optimal path, but there is a MultiHop path available.

[Sims on some pairs which use MultiHop often](https://metabase.spaceship.0x.org/dashboard/139?run_id=fool-squalid-sky&adjusted_amount_bought_win_tolerance__bps_=0.5)

## Checklist:

### DIFF
--- a/packages/asset-swapper/CHANGELOG.json
+++ b/packages/asset-swapper/CHANGELOG.json
@@ -153,6 +153,10 @@
             {
                 "note": "Added `DODO`",
                 "pr": 2701
+            },
+            {
+                "note": "Fix for some edge cases with `includedSources` and `MultiHop`",
+                "pr": 2730
             }
         ]
     },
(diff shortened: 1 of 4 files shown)


moodlezoup (approved):
man this shit is hard, thanks for fixing 😅

PR #2731, opened 2020-10-15
fix: [asset-swapper] excludedFeeSources. DODO Trade Allowed

dekz:
## Description

https://github.com/0xProject/0x-monorepo/pull/2730
Enabled fee sources to be static and not result in Uniswap being excluded. Therefor we could likely get a maker/ETH taker/ETH price.

Unfortunately this means it is queried on Ganache tests and results in a revert. 

I've added the ability to provide `excludedFeeSources` so that they can be disabled in tests or environments where it doesn't exist.

## Checklist:

### DIFF
--- a/packages/asset-swapper/src/utils/market_operation_utils/index.ts
+++ b/packages/asset-swapper/src/utils/market_operation_utils/index.ts
@@ -137,11 +137,12 @@ export class MarketOperationUtils {
         const _opts = { ...DEFAULT_GET_MARKET_ORDERS_OPTS, ...opts };
         const [makerToken, takerToken] = getNativeOrderTokens(nativeOrders[0]);
         const sampleAmounts = getSampleAmounts(takerAmount, _opts.numSamples, _opts.sampleDistributionBase);
+
         const requestFilters = new SourceFilters().exclude(_opts.excludedSources).include(_opts.includedSources);
-        // We don't exclude from the fee sources as we always want to be able to get a price
-        // of taker to Eth or maker to Eth, especially for MultiHop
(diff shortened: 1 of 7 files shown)


packages/asset-swapper/src/utils/market_operation_utils/index.ts
```diff
@@ -137,12 +137,13 @@ export class MarketOperationUtils {
...
         const sampleAmounts = getSampleAmounts(takerAmount, _opts.numSamples, _opts.sampleDistributionBase);
+
         const requestFilters = new SourceFilters().exclude(_opts.excludedSources).include(_opts.includedSources);
-        // We don't exclude from the fee sources as we always want to be able to get a price
-        // of taker to Eth or maker to Eth, especially for MultiHop
-        const feeSourceFilters = this._feeSources;
         const quoteSourceFilters = this._sellSources.merge(requestFilters);
 
+        const feeRequestFilters = new SourceFilters().exclude(_opts.excludedFeeSources);
+        const feeSourceFilters = this._feeSources.merge(feeRequestFilters);
```
dorothy-zbornak: Isn't this the same as just..?
```suggestion
        const feeSourceFilters = this._feeSources.exclude(_opts.excludedFeeSources);
```

dorothy-zbornak (approved):
One Q then :heavy_check_mark:

packages/asset-swapper/src/utils/market_operation_utils/index.ts
dekz: done

PR #2735, opened 2020-10-20
Rfqt comparison feature flag

PirosB3:
## Description

Adds a feature flag that allows us to roll out the "price-aware RFQ" feature once that data proves that we are ready to do so.

Did end-to-end testing on my local copy of 0x API

### DIFF
--- a/packages/asset-swapper/src/constants.ts
+++ b/packages/asset-swapper/src/constants.ts
@@ -128,5 +128,6 @@ export const constants = {
 // data analysis work and address some of the issues with maker fillable amounts
 // in later milestones. Once the feature is fully rolled out and is providing value
 // and we have assessed that there is no user impact, we will proceed in cleaning up
-// the feature flag.
+// the feature flag.  When that time comes, follow this PR to "undo" the feature flag:
+// https://github.com/0xProject/0x-monorepo/pull/2735
 export const IS_PRICE_AWARE_RFQ_ENABLED: boolean = false;
--- a/packages/asset-swapper/src/constants.ts
+++ b/packages/asset-swapper/src/constants.ts
@@ -122,3 +122,5 @@ export const constants = {
     DEFAULT_INFO_LOGGER,
     DEFAULT_WARNING_LOGGER,
 };
+
+export const ENABLE_PRICE_AWARE_RFQ: boolean = false;
--- a/packages/asset-swapper/src/constants.ts
+++ b/packages/asset-swapper/src/constants.ts
@@ -123,4 +123,10 @@ export const constants = {
     DEFAULT_WARNING_LOGGER,
 };
 
-export const ENABLE_PRICE_AWARE_RFQ: boolean = false;
+// This feature flag allows us to merge the price-aware RFQ pricing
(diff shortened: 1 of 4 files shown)


packages/asset-swapper/src/constants.ts
```diff
@@ -122,3 +122,11 @@ export const constants = {
     DEFAULT_INFO_LOGGER,
     DEFAULT_WARNING_LOGGER,
 };
+
+// This feature flag allows us to merge the price-aware RFQ pricing
+// project while still controlling when to activate the feature. We plan to do some
+// data analysis work and address some of the issues with maker fillable amounts
+// in later milestones. Once the feature is fully rolled out and is providing value
+// and we have assessed that there is no user impact, we will proceed in cleaning up
+// the feature flag.
```
feuGeneA: I think it would be good to leave a link to this PR in this comment so that whoever cleans up the feature flag will have a clean point of reference.

packages/asset-swapper/src/swap_quoter.ts
```diff
@@ -696,6 +696,31 @@ export class SwapQuoter {
             opts.rfqt = undefined;
         }
 
+        if (
+            !IS_PRICE_AWARE_RFQ_ENABLED && // Price-aware RFQ is disabled.
+            opts.rfqt && // This is an RFQT-enabled API request
+            opts.rfqt.intentOnFilling && // The requestor is asking for a firm quote
```
feuGeneA, line 702: Feels like a nit, but I would suggest you consider adding a whole extra `if` block around the existing one, like `if (!IS_PRICE_AWARE_RFQ_ENABLED) { if (opts.rfqt && ...) { ... } }`. That way, when we come to rip that out, it'll be totally clear that the whole block needs to go.  I could see someone being confused about the intent, just by looking this code, wondering how the feature flag condition relates to the other conditions in this `if`.

feuGeneA: After seeing your other usages of this flag, they don't all seem like they'd be amenable to this dedicated-`if` approach, so I'll defer to your judgement.  However, I'd maintain that what I had in mind as a design goal should probably still stand: make it SUPER easy for whoever rips this out, and don't worry about making it extra verbose or unclean.  In particular, another suggestion I'd recommend is a simple comment at every usage of the flag, explaining how to rip it out.  `// When cleaning up IS_PRICE_AWARE_RFQ_ENABLED, just remove this whole entire if block` might be the comment for this instance.  For another one here, perhaps `Remove the rfqPromise var declaration, and its usage, and remove the rfqtIndicativeQuotes declaration, replacing its usage with []`.  Etc.

feuGeneA (approved):
Approval contingent on clean CI runs (`test-publish` excepted, of course).

The `static-tests` failures clear up with a merge of the target branch, and the `test-rest` failures clear up with some feature-flag usage additions in the tests, all of which I've got working locally and will push up to the branch when you give me the go-ahead.

PirosB3:
Thank you so much for the help Gene, as discussed earlier I would rather get this in now, and I will be responsible for cutting out the feature-flag code once the feature has been deployed.

PR #2736, opened 2020-10-20
[WIP] add protocol fees fixin

smarx:
## Description

This change adds `FixinProtocolFees`, code that collects protocol fees using per-pool `FeeCollector` contracts. Once per epoch, a function on each `FeeCollector` is called to transfer the held ether (first converted to WETH) to the staking contract. Delaying this transfer and associated accounting yields significant gas savings.

## Testing instructions

`yarn build && yarn test` in `contracts/zero-ex`.

The basic happy path with WETH is tested, but all other tests are missing. Future commits will add more robust testing, including negative testing for failure modes.

## Types of changes

## Checklist:


contracts/zero-ex/contracts/src/external/FeeCollector.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// @param poolId The pool ID this contract is collecting fees for.
+    function initialize(
+        IEtherTokenV06 weth,
+        IStaking staking,
+        bytes32 poolId
+    )
+        external
+        onlyAuthorized
+    {
+        weth.approve(address(staking), type(uint256).max);
```
dorothy-zbornak, line 48: `type(uint256).max` :exploding_head:

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,115 @@
...
+        } else {
+            // ETH
+            (bool success,) = address(feeCollector).call{value: amount}("");
+            require(success);
+        }
+    }
+
+    /// @dev Transfer fees for a given pool to the staking contract.
+    /// @param poolId Identifies the pool whose fees are being paid.
+    function _transferFeesForPool(
```
dorothy-zbornak, line 63: Who will end up calling this?

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
```diff
@@ -0,0 +1,115 @@
...
+
+pragma solidity ^0.6.5;
+pragma experimental ABIEncoderV2;
+
+import "@0x/contracts-erc20/contracts/src/v06/IEtherTokenV06.sol";
+import "../external/FeeCollector.sol";
+
+/// @dev Helpers for collecting protocol fees.
+abstract contract FixinProtocolFees {
+    bytes32 immutable feeCollectorCodeHash;
```
dorothy-zbornak, line 27: I guess this is the advantage over a library.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
smarx: Good question. I think the answer is that we need to expose it publicly and let _anyone_ call it. (0x Labs will likely be responsible for actually doing this at the end of every epoch. No one else has as much incentive to pay the gas fees to do it.)

So I could have perhaps made it an `external`, open function in the fixin itself. But I think it's better to expose it exactly once from a feature.

contracts/zero-ex/contracts/src/fixins/FixinProtocolFees.sol
smarx: Yes. I think this is the only advantage.

PR #2752, opened 2026-09-11
Fix Immunefi 87687

e1Ru1o:


contracts/staking/contracts/src/StakingUpgrade.sol
```diff
@@ -0,0 +1,44 @@
+/*
+
+  Copyright 2019 ZeroEx Intl.
```
duncancmt, line 3: Copyright year needs updating

duncancmt (approved):

